PR #245, opened 2018-07-21
0009916: update SQLite to 3.24.0

patrikjuvonen:
**Mantis Bug Tracker issue:**
[9916](https://bugs.multitheftauto.com/view.php?id=9916)

Requested initially by @Dutchman101

Our current SQLite version is 3.13.0.

**Test resource:**
[test_sqlite_3_24_0.zip](https://github.com/multitheftauto/mtasa-blue/files/2217713/test_sqlite_3_24_0.zip)

**New in 3.24.0:**
https://www.sqlite.org/releaselog/3_24_0.html

**Since our current version:**
SQLite Release 3.14.2 On 2016-09-12: https://www.sqlite.org/releaselog/3_14_2.html
SQLite Release 3.15.2 On 2016-11-28: https://www.sqlite.org/releaselog/3_15_2.html
SQLite Release 3.16.2 On 2017-01-06: https://www.sqlite.org/releaselog/3_16_2.html
SQLite Release 3.17.0 On 2017-02-13: https://www.sqlite.org/releaselog/3_17_0.html
SQLite Release 3.18.0 On 2017-03-30: https://www.sqlite.org/releaselog/3_18_0.html
SQLite Release 3.19.3 On 2017-06-08: https://www.sqlite.org/releaselog/3_19_3.html
SQLite Release 3.18.2 On 2017-06-17: https://www.sqlite.org/releaselog/3_18_2.html
SQLite Release 3.20.1 On 2017-08-24: https://www.sqlite.org/releaselog/3_20_1.html
SQLite Release 3.21.0 On 2017-10-24: https://www.sqlite.org/releaselog/3_21_0.html
SQLite Release 3.22.0 On 2018-01-22: https://www.sqlite.org/releaselog/3_22_0.html
SQLite Release 3.23.1 On 2018-04-10: https://www.sqlite.org/releaselog/3_23_1.html
SQLite Release 3.24.0 On 2018-06-04: https://www.sqlite.org/releaselog/3_24_0.html

PR #913, opened 2019-05-05
Update SQLite from 3.27.1 to 3.28.0

patrikjuvonen:
## Summary
- Contains various bug fixes, optimizations and some other minor changes and additions that can be useful (e.g. some improvements to window functions that we got to use since #818)
- Latest changelog: https://www.sqlite.org/changes.html#version_3_28_0

## Tests
Tested with [test_sqlite.zip](https://github.com/multitheftauto/mtasa-blue/files/2872064/test_sqlite.zip) (cheers @botder)

## Validation
To help validate the integrity of the update I have created the following bash script that diffs between my PR branch and the official package provided from the [SQLite website](https://www.sqlite.org/download.html).
```bash
#!/bin/bash

SQLITE_UPDATE_VERSION=3.28.0
SQLITE_PATH_NAME=sqlite-$SQLITE_UPDATE_VERSION

GIT_REPO_BRANCH=vendor/sqlite-$SQLITE_UPDATE_VERSION
GIT_DEST_DIR=mtasa-blue
GIT_REPO_SQLITE_PATH=$GIT_DEST_DIR/vendor/sqlite/

echo 1. Download and extract $SQLITE_PATH_NAME...
curl https://www.sqlite.org/2019/sqlite-autoconf-3280000.tar.gz | tar -xz

echo 2. Clone the vendor update branch $GIT_REPO_BRANCH from $GIT_REPO_URL into $GIT_DEST_DIR...
git clone --depth 1 -b $GIT_REPO_BRANCH $GIT_REPO_URL $GIT_DEST_DIR

diff -r $GIT_REPO_SQLITE_PATH sqlite-*

echo 4. Completed.
```

## Past SQLite updates in MTA
| Date | From | To | Link |
| :--- | :--- | :--- | :--- |
| February 2019 | 3.24.0 | 3.27.1 (current) | #818 |
| July 2018 | 3.13.0 | 3.24.0 | #245 |
| August 2016 | 3.7.17 | 3.13.0 | 3aa17532867c66818cc3a602c4b1ac2143694066 |
| August 2013 | 3.7.8 | 3.7.17 | fd195b799498f75496fe0c15adac3cab3aecb639 |
| October 2011 | 3.6.14 | 3.7.8 | ca063630631d56d5ba9aba5bf6771fe0299f5a9f |

### 2019-04-16 (3.28.0)

1. Enhanced window functions:
    - Add support the EXCLUDE clause.
    - Add support for window chaining.
    - Add support for GROUPS frames.
    - Add support for "<expr> PRECEDING" and "<expr> FOLLOWING" boundaries in RANGE frames.
6. Added the new sqlite3_stmt_isexplain(S) interface for determining whether or not a prepared statement is an EXPLAIN.
7. Enhanced VACUUM INTO so that it works for read-only databases.
8. New query optimizations:
    - Enable the LIKE optimization for cases when the ESCAPE keyword is present and PRAGMA case_sensitive_like is on.
    - In queries that are driven by a partial index, avoid unnecessary tests of the constraint named in the WHERE clause of the partial index, since we know that constraint must always be true.
11. Enhancements to the TCL Interface:
    - Added the -returntype option to the function method.
    - Added the new bind_fallback method.
14. Enhancements to the CLI:
    - Added support for bound parameters and the .parameter command.
    - Fix the readfile() function so that it returns an empty BLOB rather than throwing an out-of-memory error when reading an empty file.
    - Fix the writefile() function so that when it creates new directories along the path of a new file, it gives them umask permissions rather than the same permissions as the file.
    - Change --update option in the .archive command so that it skips files that are already in the archive and are unchanged. Add the new --insert option that works like --update used to work.
19. Added the fossildelta.c extension that can create, apply, and deconstruct the Fossil DVCS file delta format that is used by the RBU extension.
20. Added the SQLITE_DBCONFIG_WRITABLE_SCHEMA verb for the sqlite3_db_config() interface, that does the same work as PRAGMA writable_schema without using the SQL parser.
21. Added the sqlite3_value_frombind() API for determining if the argument to an SQL function is from a bound parameter.
22. Security and compatibilities enhancements to fts3_tokenizer():
    - The fts3_tokenizer() function always returns NULL unless either the legacy application-defined FTS3 tokenizers interface are enabled using the sqlite3_db_config(SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER) setting, or unless the first argument to fts3_tokenizer() is a bound parameter.
    - The two-argument version of fts3_tokenizer() accepts a pointer to the tokenizer method object even without the sqlite3_db_config(SQLITE_DBCONFIG_ENABLE_FTS3_TOKENIZER) setting if the second argument is a bound parameter
25. Improved robustness against corrupt database files.
26. Miscellaneous performance enhancements
27. Established a Git mirror of the offical SQLite source tree. The canonical sources for SQLite are maintained using the Fossil DVCS at https://sqlite.org/src. The Git mirror can be seen at https://github.com/sqlite/sqlite.

### 2019-02-25 (3.27.2)

1. Fix a bug in the IN operator that was introduced by an attempted optimization in version 3.27.0. Ticket df46dfb631f75694
2. Fix a bug causing a crash when a window function is misused. Ticket 4feb3159c6bc3f7e33959.
3. Fix various documentation typos

### DIFF
--- a/vendor/sqlite/premake5.lua
+++ b/vendor/sqlite/premake5.lua
@@ -1,10 +1,10 @@
 project "sqlite"
-	language "C++"
+	language "C"
 	kind "StaticLib"
 	targetname "sqlite"
 	
 	vpaths { 
-		["Headers/*"] = "**.h",
+		["Headers/*"] = "*.h",
 		["Sources/*"] = "*.c",
 		["*"] = "premake5.lua"
 	}
--- a/vendor/sqlite/sqlite3.h
+++ b/vendor/sqlite/sqlite3.h
@@ -123,9 +123,9 @@ extern "C" {
 ** [sqlite3_libversion_number()], [sqlite3_sourceid()],
 ** [sqlite_version()] and [sqlite_source_id()].
 */
-#define SQLITE_VERSION        "3.27.1"
-#define SQLITE_VERSION_NUMBER 3027001
-#define SQLITE_SOURCE_ID      "2019-02-08 13:17:39 0eca3dd3d38b31c92b49ca2d311128b74584714d9e7de895b1a6286ef959a1dd"
+#define SQLITE_VERSION        "3.28.0"
+#define SQLITE_VERSION_NUMBER 3028000
+#define SQLITE_SOURCE_ID      "2019-04-16 19:49:53 884b4b7e502b4e991677b53971277adfaf0a04a284f8e483e2553d0f83156b50"
 
 /*
 ** CAPI3REF: Run-Time Library Version Numbers
@@ -189,6 +189,9 @@ SQLITE_API int sqlite3_libversion_number(void);
 #ifndef SQLITE_OMIT_COMPILEOPTION_DIAGS
 SQLITE_API int sqlite3_compileoption_used(const char *zOptName);
 SQLITE_API const char *sqlite3_compileoption_get(int N);
+#else
+# define sqlite3_compileoption_used(X) 0
+# define sqlite3_compileoption_get(X)  ((void*)0)
 #endif
 
 /*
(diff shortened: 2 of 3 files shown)


qaisjp (approved):
Verified

PR #914, opened 2019-05-05
Fix #580: createTrayNotification not working for Windows 10

### LINKED ISSUE #580: createTrayNotification not working for Windows 10
**Describe the bug**
function createTrayNotification doesn't pop out notification text in balloon when client runs Windows 10. Only on Windows XP, 7 and 8.1 it works.

The MTA app icon will appear in the taskbar without showing any dialogue, when the function is triggered including a notification text that's supposed to show.

**To reproduce**
crun setTimer(function() createTrayNotification( "Hello World", "default", false ) end, 2000, 1)

**Expected behaviour**

**Screenshots**

**Version**

**Additional context**
We figured out that the problem was the main GTA window is not 100% unicode compatible.
Solutions are:
a) Use non-unicode tray notifications
b) Hook GTA to create main unicode window at startup (might not be possible)
c) Create an extra hidden unicode window to use for tray notifications

Isn't a dummy window created to handle non-blocking window drags? Maybe that window could be used to create notifications?

From https://bugs.mtasa.com/view.php?id=9447

samr46:
Fixes #580 

Added extra hidden unicode window to handle tray notifications.

Note: "Quiet hours" mode should be disabled to receive notifications.

New behavior preview: https://imgur.com/HQe3Dln

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -31,54 +31,6 @@ extern CCore* g_pCore;
 bool          g_bBoundsChecker = false;
 SString       g_strJingleBells;
 
-BOOL AC_RestrictAccess()
-{
-    EXPLICIT_ACCESS NewAccess;
-    PACL            pTempDacl;
-    HANDLE          hProcess;
-    DWORD           dwFlags;
-    DWORD           dwErr;
-
-    ///////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


botder:
Does it work in the Release build?

samr46:
Now it does.

botder (approved):

PR #917, opened 2019-05-07
WIP: Improve rendering performance

### LINKED ISSUE #905: Improve rendering performance (low FPS)

When you have too many entities on screen, your FPS will drop. This happens because draw calls are expensive in DX9. This issue must be resolved before implementing #869. It's related to #635.

#### Solution 1:

Any entity that is created more than once, we can use [instancing](https://docs.microsoft.com/en-us/windows/desktop/direct3d9/efficiently-drawing-multiple-instances-of-geometry). This solution is only applicable to entities which have the same materials. Let's say if you have a normal infernus and a modded infernus (different textures), then this solution cannot be applied to the modded entity. 

#### Solution 2:

I talked to Jusonex, he suggested that we can batch the vertex buffers of entities and use a texture atlas to reduce the draw calls. In my opinion, this is a perfect solution. It makes much more sense than instancing since we have given access to scripters to mod game entities. 

Additionally, we can use a secondary thread for rendering. We'll have two kinds of meshes:

- Static Mesh
- Non-static Mesh

A `non-static` mesh gets updated every frame or frequently, and we'll need to keep 2 copies of the vertex buffer, this means there will be 3 vertex buffers for a non-static mesh which is being rendered. The memory will be freed if it's not in sight.  Examples of a non-static mesh are vehicles, peds, animated objects, etc.

The 1st vertex buffer will be used by the main thread to update, the other will remain in the command list if update completes before the render thread, and the final one will be used by the render thread.

A `static` mesh is static, it's guaranteed to remain the same during run-time, so we'll have only 1 vertex buffer and no extra copies are needed. Majority of the entities in game are static, for example, trees, roads, buildings, interiors, houses, etc. So if we are using 3 vertex buffer for non-static meshes, it's okay because non-static meshes are few in numbers.

#### Synchronization

The render thread will wait for main thread to complete if it completes execution before the main thread. If the render thread takes more time to execute than main, then the main thread will update the second buffer for non-static animations and move on to the next frame, when the render thread is done executing it will copy the data from the second vertex buffer of dynamic entities to the 3rd vertex buffer and pass it to the draw call.

**Additional context**

Since I want to fix this problem using solution 2, we will need to write some code for creating a texture atlas of models. It's probably better to convert them by ourselves and instruct the players to download our custom gta3.img to save CPU time. It should be optional, of course. We can simply check the MD5 of gta3.img to see if it belongs to use or not.

ghost:
~~Fixes #905. Texture atlas generator is complete. It combines multiple images together to create a single image.~~ 

See #1061


Samake:
Nice to see starts working on this. Thumbs up! What do you think which size atlas can handle later, i mean to hold the whole textures wouldnt be possible. Would it support multiple Atlasses by categories or so?

Einheit-101:
Wow, Sam@ke alive? Nice.

@Topic: I would like to make some performance measurements. I have a lot of time recently.

ghost:
> 
> What do you think which size atlas can handle later, i mean to hold the whole textures wouldnt be possible. Would it support multiple Atlasses by categories or so?

You simply need to supply the textures, it will automatically take care of sizes. So basically, nothing will change for scripters except for performance which will make a big difference. Maximum size for a single atlas is 16384x16384. If the textures cannot fit into one atlas then another will be created, so it's up to the program to decide how many atlases are needed, and it will generate them without any issue.

>
>@Topic: I would like to make some performance measurements. I have a lot of time recently.

Sure, once I'm done with the PR. I'll let you know. :P

Citizen01:
You have all the support I can give, just wanted to point out your PR includes 2 commits that are not related to this PR. You should have create a seperate branch on your side to prevent this.

ghost:
> 
> You have all the support I can give, just wanted to point out your PR includes 2 commits that are not related to this PR. You should have create a seperate branch on your side to prevent this.

Yes, I realized it later. I'll squash and merge this PR, so these commits will disappear anyway ;p

WADmitry:
How about externally prepared textures (using TexturePacker, etc)? For example, then I can use offsets between images and edge bleeding (to avoid alpha/interpolation artifacts on scaled textures).

Also, about maximum size of a single atlas. It depends on hardware. Some of "old" (or just low-end) hardware may not support 16K, 8K and even 4K textures. I think would be good add ability to adjust maximum size (16K/8K/4K).

ghost:
> 
> How about externally prepared textures (using TexturePacker, etc)? For example, then I can use offsets between images and edge bleeding (to avoid alpha/interpolation artifacts on scaled textures).

If you're talking about rendering images, then you can always use dxDrawImageSection.  I'm adding the atlas feature only to reduce the state changes (IDirect3DDevice9::SetTexture mainly) for 3D meshes. 

> 
> Also, about maximum size of a single atlas. It depends on hardware. Some of "old" (or just low-end) hardware may not support 16K, 8K and even 4K textures. I think would be good add ability to adjust maximum size (16K/8K/4K).

Yes, we'll need to query the device capabilities and depending on that, we can set the maximum atlas size.

WADmitry:
> I'm adding the atlas feature only to reduce the state changes (IDirect3DDevice9::SetTexture mainly) for 3D meshes.

When textures joined into atlas then it may add graphical artifacts on edges (when a textures are independent then the problem usually is not appear) in some cases (for example, when texture is significally zoomed in, interpolation may pick texels of adjacent texture)

ghost:
> 
> When textures joined into atlas then it may add graphical artifacts on edges (when a textures are independent then the problem usually is not appear) in some cases (for example, when texture is significally zoomed in, interpolation may pick texels of adjacent texture)

The program will automatically fix the issue for texture mipmaps and remap the 3D mesh UVs properly to avoid edge bleeding. Do you want to do something similar for 2D images that are within an atlas? If yes, then please explain it here.

PlatinMTA:
Where this will be launching? On 1.5.7 or on some nightly?

Awesome work btw, MTA got a lot of new cool things these past few months.

ghost:
> 
> Where this will be launching? On 1.5.7 or on some nightly?

It will be out on nightlytest.mtasa.com within a month for testing. 1.5.7 will be released soon, so it will come after 1.5.7

ArranTuna:
Very nice. If there is a img generator does that mean that only default GTA SA textures are optimizable?

ghost:
> 
> 
> Very nice. If there is a img generator does that mean that only default GTA SA textures are optimizable?

No. It will work for all textures and DFF files within the IMG even the custom ones. For DFF, the program will remap the UV coordinates.

ArranTuna:
Server modifications aren't in the img though. So wouldn't this mean a total conversion wouldn't have any performance gain?

ghost:
> 
> Server modifications aren't in the img though. So wouldn't this mean a total conversion wouldn't have any performance gain?

How about adding a Lua function on client side that can optimize the supplied TXD and DFF files?

ArranTuna:
That sounds good. I'm looking forward to testing this BTW.

Client/core/CIMGArchive.h
```diff
@@ -0,0 +1,60 @@
+#ifndef CIMGArchive_H
```
qaisjp: Please use `#pragma once` instead

Client/core/CIMGArchive.cpp
```diff
@@ -0,0 +1,180 @@
...
+}
+
+DWORD GetTotalBlockSize(DWORD blockSize)
+{
+    const uint64_t IMG_BLOCK_SIZE = 2048;
+    blockSize = (blockSize + IMG_BLOCK_SIZE - 1) & ~(IMG_BLOCK_SIZE - 1); // Round up to block size
+    return blockSize / IMG_BLOCK_SIZE;
+}
+
+#include <iostream>
```
qaisjp: These includes should probably be at top of file

Client/core/CIMGArchive.cpp
```diff
@@ -0,0 +1,180 @@
...
+std::vector<EntryHeader> CIMGArchive::GetArchiveDirEntries()
+{
+    return archiveFileEntries_;
+}
+
+CIMGArchiveFile* CIMGArchive::GetFileByID(uint id)
+{
+    if (archiveFileEntries_.size() <= id || id < 0)
+    {
+        return NULL;
```
qaisjp: Please use `nullptr` wherever possible in all new code.

Client/core/CIMGArchive.h
ghost: @qaisjp I think it's too early to review the code. I'm just trying to make it work for now. I'll rewrite it when I'm done. ;p

ghost:
`RpClumpStreamWrite` is completely broken in GTA SA. It only works if you disable the custom plugins which rockstar added to the renderware engine, and we cannot do that. I had to rewrite some functions, still 2 more to go for C2dEffect...

ghost:
Generating DFF files finally works with support for all plugins. Now, I need to write some more code for creating texture atlases from TXD raster and remapping the geometry vertices to it for DFF files that use it.

We can also add new Lua functions to generate a DFF or COL file for those who wish to export modified COL or DFF. That'll be only useful if we have functions which allow scripters to modify geometry of clump or its collision. Somebody can even create a new ingame editor. @CrosRoad95's collision editing PR can make this a reality.

ghost:
I was almost about to finish this and found out that the texture must be unwrapped and positioned properly in atlas. Generating it automatically is a very complicated process, which is why human interaction is required to add seams to the mesh where texture might be overlapping and then allow the program to do its job, just like in blender.

Anyways, after spending countless hours, I finally figured out how to automate it without messing up the texture wrapping. We'll need to use [LSCM](https://en.wikipedia.org/wiki/Least_squares_conformal_map), and there is a library named "xatlas" which uses LSCM. However, xatlas was mainly created for baking lightmap, so it doesn't copy the texture to the atlas, I'll need to write some code for that.

Moreover, I removed DirectXTex. It's not needed anymore.

ghost:
The output mesh from xatlas can have more vertices than supplied; hence, we'll need to recreate the geometry for every 3D model. This was a pain because the copying functions for geometry plugins have been optimized out as they are not used, and I had to write my own. It's finally done. 

The new vertices should be copied to the new geometry when it's created, and then we'll have a working build for testers.

PlatinMTA:
I really want to test this right now, can't wait.

ghost:
Let's just say that I'm very close to finish this. I tried generating atlases for all models and they all worked fine except for 3-4 due to some problems with xatlas. No big deal, I'll contact the author, and hopefully he'll fix it. He's actively developing xatlas. You can check my to-do list to see what's left.

I came across low LOD models, and it seems that we cannot use the LOD models from GTA SA if we are going to use atlases. Technically, we can use them, but when the model is not fully visible, the low LOD model will be displayed, and it will use multiple draw calls since every low LOD model uses a different TXD file which has textures with extremely low resolution. We cannot generate atlases for these low LOD models because then the textures will look like huge pixelated color boxes.  

What are we going to do then? I'm going to use another library for generating low LOD models from the new atlas DFF and TXD files. We need LOD support for our players who are running GTA on low-end computers. I'll do this when I finish working on atlases. 

To-do list:

- [x] Avoid adding repeated textures to atlas to reduce atlas size (needs automation)
- [ ] Patch some GTA code to fix messed up color bug with vehicle models. FYI, a vehicle model on screen will take 15 to 24 draw calls per frame. We are reducing it to 1.
- [ ] Rewrite some renderware rendering functions to support atlases, since without doing that, we'll see absolutely no change. VERY IMPORTANT!
- [ ] Use a 3rd party library to generate low lod models instead of using the GTA ones. Low priority.

ArranTuna:
Maybe low LOD's won't even be necessary, when those renderware functions are rewritten and you've seen the performance improvement, maybe it will be so high that the difference of having low LOD's will be insignificant and instead the full model could just be displayed?

Einheit-101:
GTA LOD system is just annoying anyways, and even the default models are so low poly, they are insignificant.

ghost:
@ArranTuna It's not just about performance. It also saves you lots of memory if a model is far away, GTA unloads it and shows the low LOD. Like I said, LOD models have their own TXD files. These TXD files will take up only a few KBs and save memory if you are not using custom models. 

I do have plans to add LOD support for custom models after this PR. 

@Einheit-101 Maybe you're right, but one can argue that if you are flying a hydra and you position your camera in a way where you see thousands of objects on your screen, there will be some performance issue on weak computers.

ArranTuna:
> I do have plans to add LOD support for custom models after this PR.

If you mean objects that servers have applied mods to, there is already some functions for that like setLowLODElement which I have used on modified land masses (but I just use the exact same texture instead of wasting the time and another model ID slot for a low quality texture).

ghost:
I was able to generate a custom img file with atlas models. It was working, however, I could not load more than 4500 TXD files since that's the limit. The game kept crashing when I tried to exceed the limit, so I'll have to fix this by merging TXD files. I wanted to lift the model ID limit, but I got busy with this PR.

Just two more important things to actually benefit from the atlases: 

- Rewrite some renderware rendering functions to support atlases
- Patch some GTA code to fix messed up color bug with vehicle models

ghost:
All models seem to work, but merging TXD files is not a good idea. It results in TXD files being large. I limited the atlas TXD size to 2mb, and I still got 1200 atlas TXDs. I cannot have more than 1500 custom TXD files since GTA internal TXD names will take up the remaining 3500. When you are flying hydra, the streaming memory usage will be very high and objects will start to disappear.  

It happens because loading a single DFF model and its TXD will result in unwanted textures being loaded along with the TXD, and it will eat lots of memory. That's not really what we are looking for. There are 8000 DFF models which are worth optimizing, so we need to uplift the limit to **12000 TXD files** to make the streamer run smooth and use less memory. I'm going to use some code from fastman92 limit adjuster to make sure every DFF with atlas can have its own TXD.

ghost:
I've decided to stop working on this, and go for DX11 instead. It's going to take a lot of work to implement the limit adjuster and fix the bugs/crashes that might appear. Moreover, the atlases textures will eat some good amount of RAM, and the only way to fix that is to reduce the texture sizes, but that's not what we want. Like I said, too much work to make it happen, why not spend that time on implementing upgrade to DX11? 🤔🤔🤔🤔🤔🤔

The atlases code is completely useless now. However, there is another feature in this PR, and that is the ability to output COL, DFF, and TXD files. Basically, we can add some new functions to allow scripters to create their own ingame editors for modifying DFF clumps, textures, and COL models, and then output them to a file.

Let's talk in #1061

PR #926, opened 2019-05-10
Fix #429: setCloudsEnabled affects moon and stars but not actual clouds

### LINKED ISSUE #429: Disabling clouds also disables the moon and stars
**Describe the bug**
If you disable clouds via setCloudsEnabled (client/server), the moon and stars disappear. The sun seems to be unaffected.

**To Reproduce**
1. Set night time or when the moon and stars are visible
2. Use setCloudsEnabled(false)
If you set it to true, the moon and stars appear again.

**Expected behavior**
Moon should not disappear.

**Screenshots**
N/A

1.5.6

1.5.6

**Additional context**
https://bugs.mtasa.com/view.php?id=9905

samr46:
Fixes #429 

Fixed incorrect memory address of clouds render (2 lines)

**New behavior preview:**

<details><summary>Night</summary>

</details>

<details><summary>Day</summary>

</details>

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1997,12 +1997,11 @@ void CMultiplayerSA::SetCloudsEnabled(bool bDisabled)
     else
         MemPut<BYTE>(0x716380, 0xC3);
 
-    // normal clouds
-    // 0071395A     90             NOP
+    // bottom clouds
     if (bDisabled)
-        MemPut<BYTE>(0x713950, 0x83);
+        MemPut<BYTE>(0x7154B0, 0x83);
     else
-        MemPut<BYTE>(0x713950, 0xC3);
+        MemPut<BYTE>(0x7154B0, 0xC3);
 
     // plane trails (not really clouds, but they're sort of vapour)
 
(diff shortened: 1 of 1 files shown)


patrikjuvonen:
Nice one, however high/normal clouds are now no longer affected (e.g. weather ID 10).

mta-screen_2019-05-10_23-46-12

I think the upside about this change however is the fact that we get to keep stars and the moon.

samr46:
Interesting. I checked all weathers and missed it.

patrikjuvonen:
I think one way would be to introduce a new parameter or even a new function maybe since the old way affects the moon and stars (which have nothing to do with our traditional thinking of clouds per se, other than cloud density blocking line of sight, but that doesn't make sense here).

samr46:
It seems that it's another type of clouds. Looks like some kinda "horizon / skyline clouds".

samr46:
Fixed it.
mta-screen_2019-05-11_01-24-08

qaisjp (approved):
Thanks!

patrikjuvonen:
Thanks for seeing it through, seems to work fine now. Cheers!

patrikjuvonen (commented):
Thanks for seeing it through, seems to work fine now. Cheers!

PR #933, opened 2019-05-12
Set missing spawned and dead state for cloned peds

TheNormalnij:
Some functions (setElementDimension, setElementHealth, setPedArmor, ... ) do not work with cloned peds. This is fix for it.

### DIFF
--- a/Server/mods/deathmatch/logic/CPed.cpp
+++ b/Server/mods/deathmatch/logic/CPed.cpp
@@ -112,7 +112,8 @@ CElement* CPed::Clone(bool* bAddEntity, CResource* pResource)
         pTemp->SetHealth(GetHealth());
         pTemp->SetArmor(GetArmor());
         pTemp->SetSyncable(IsSyncable());
-        pTemp->SetSpawned(true);
+        pTemp->SetSpawned(IsSpawned());
+        pTemp->SetIsDead(IsDead());
     }
 
     return pTemp;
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CPed.cpp
```diff
@@ -112,6 +112,7 @@ CElement* CPed::Clone(bool* bAddEntity, CResource* pResource)
         pTemp->SetHealth(GetHealth());
         pTemp->SetArmor(GetArmor());
         pTemp->SetSyncable(IsSyncable());
+        pTemp->SetSpawned(true);
```
qaisjp:
```suggestion
        pTemp->SetSpawned(IsSpawned());
        pTemp->SetIsDead(IsDead());
```
What happens if the ped is dead and then cloned?

I think it might make sense to use `IsSpawned` here.

You may also want to add the same for `SetIsDead` using `IsDead`

qaisjp (commented):
Thank you very much for your pull request!

I've reviewed a potential area for improvement.

TheNormalnij:
Thanks you very much also.

I will be careful here.

Server/mods/deathmatch/logic/CPed.cpp
TheNormalnij: Good suggestion. It did not work properly for dead peds also.

PR #935, opened 2019-05-12
Add dxGetTextSize function

StrixG:
This function retrieves the width and height of a certain piece of text.

Syntax:
```lua
float, float dxGetTextSize ( string text, [float width=0, float scaleXY=1.0, float scaleY=1.0, mixed font="default", bool wordBreak=false, bool colorCoded=false] )
vector2 DxFont:getTextSize ( string text, [float width=0, float scaleXY=1.0, float scaleY=1.0, bool wordBreak=false, bool colorCoded=false] )
```

### TODO
- [x] Fix scaling


Pirulax:
So we basically draw a freaking string to get it's height? Thats waste of drawcalls which are expensive in dx9

StrixG:
@Pirulax dxGetTextWidth works the same way. I don't know how we can do this another way.
Maybe we should add dxGetTextSize/BoundingBox to reduce drawcalls.

ccw808:
As DT_CALCRECT is specified, only bounds are calculated and there should be no draw calls
https://docs.microsoft.com/en-us/windows/desktop/direct3d9/id3dxfont--drawtext

ccw808:
What is the advantage of dxGetTextHeight over dxGetFontHeight?

StrixG:
@ccw808 dxGetTextHeight can calculate height of dxDrawText with wordBreak = true.

qaisjp:
(What if we did dxGetTextSize and returned a vec?)

qaisjp (dismissed):
Unintended line ending changes must be removed before merge.

StrixG:
dxGetTextWidth counts trailing space characters when calculating the width. dxGetTextSize doesn't do that, cause it's not possible when word wrapping is enabled.
Should we count space characters when using dxGetTextSize without word wrapping?

qaisjp:
we should put a note on dxGetTextWidth saying "this assumes word break is off. Use dxGetTextSize instead"

then the difference (with wordBreak true) is fine

dmi7ry:
Still not merged?

qaisjp (approved):

Issue #821, opened 2019-02-17
Pin down return value on failure


qaisjp:
Recently there seems to be some confusion about what values functions should return when they encounter an error.

Historically, the following has been the general policy; **(1)**:

- On failure, all functions should return false.

Somewhere along the way we realised functions that ordinarily return booleans (isPedDead, hasObjectPermissionTo, etc) should not return meaningful values (`false`) when errors are encountered, so we started doing something else for new functions; **(2)**:

- If a successful function call would return a boolean value, a failed call should return `nil`.
- Otherwise, all other failed function calls return `false`.

Some new functions / pull requests have started to do something else; **(3)**: 

- Always return `nil` on failure, no matter the return type.

Things to pin down:

- Pick a system
    1. Stick to the current system **(2)**
    1. Begin adopting **(3)**?
    1. Something else?
- Do we want to modify old functions or should this only apply to new functions?

qaisjp:
Okay, so my current POV is that **(3)** is a good way to go. I think it's nice because it's consistent for every function.

Nevertheless, I have been asking for PRs to stick to **(2)** since that's what is currently done.

I am unsure as to whether existing functions should change. I imagine a lot of existing code does `if result == false` instead of `if not result`.

botder:
What do you think about this one?
https://github.com/multitheftauto/mtasa-blue/blob/a547e4bad58a7475f4323ba0d12a92d866a59de5/Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp#L890-L891 `...`
https://github.com/multitheftauto/mtasa-blue/blob/a547e4bad58a7475f4323ba0d12a92d866a59de5/Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp#L963-L966

I prefer to return **nil** on failure, no matter the return type **(3)**.
I am not sure if changing existing functions, to return nil, would break scripts.

qaisjp:
hasObjectPermissionTo is a getter that returns boolean values, so fits into **(2)**.

botder:
It's not actually a getter, because by definition, it doesn't simply return a private member variable of the player/resource/..., which gets passed to the function. It has to lookup the permission in the ACL for the object and then returns a boolean.

qaisjp:
**(2)** isn't restricted to getters in that traditional sense, it is more about `false` having a special meaning.

[...]

PR #938, opened 2019-05-13
Add areVehicleLightsOn function

### LINKED ISSUE #478: Function to determine current vehicle light state
**Description**
Surprisingly it is currently impossible to get the vehicle light state (Enabled/disabled) in MTA.
Vehicles will always automatically disable their lights between 06:00 and 07:00 and enable them between 20:00 and 21:00. The exact point of time is random for every vehicle, it looks like it is a property of each vehicle once it gets created.

A function that gets the current vehicle light status.

A function that sets the automatic light enable/disable time of a specified / all vehicles to a specified value.

**Additional context**
It can be done with scripting by simply overwriting headlight state, yes.

StrixG:
This function gets the vehicle lights state (on/off).
Works with automatically enabled lights at night.

_Closes #478_


StrixG:
Waits for #821.

botder:
The decision in pull request #821 (**Pin down return value on failure**) is now settled.

patrikjuvonen:
Great! Thank you for your PR.

Einheit-101:
Nice nice!

PR #939, opened 2019-05-13
Fix #479: setAccountName does not save name to database

### LINKED ISSUE #479: setAccountName() does not work properly
**Describe the bug**
This function only changes the name of the account while the server is on. When the server is turned off the account just reverts to the last name it had. I think this happens because the 'internal.db' doesn't get affected by it. 

**To Reproduce**
You can run this resource: https://pastebin.com/p4htLKia

1. Use /create to create an account. 
2. Check if it was created on 'internal.db' (you can see this using a SQLite Browser). 
3. You could also run /checkacc to see if this account was properly created. 
4. If it was, then run /change. 
5. This will change the name of the account. If you run /checkacc you will see that the main account now "does not exists". 
6. You can now recheck the 'internal.db' to see if the account really changed or run /shutdown and reopen the server. 
7. Once reopen use /checkacc to check if the account really changed it's name.

**Expected behavior**
The name of the account should change in the .db aswell, thing which does not occur. So for servers that restart periodicly, changing names using this function isn't posible. You will need to use MySQL or run a custom setAccountName() function that creates and deletes an account to make that change.

 - Version v1.5.6-release-14370
 - Running in Wine? No

 - OS: Windows 7 and Linux Debian 9
 - Version v1.5.6-release-14370 (both)

StrixG:
_Fixes #479_

### DIFF
--- a/Server/mods/deathmatch/logic/CAccountManager.cpp
+++ b/Server/mods/deathmatch/logic/CAccountManager.cpp
@@ -272,9 +272,9 @@ void CAccountManager::Save(CAccount* pAccount, bool bCheckForErrors)
     SString strQuery;
     strQuery += m_pDatabaseManager->PrepareStringf(m_hDbConnection, "UPDATE accounts SET ip=?", SQLITE_TEXT, *strIP);
     if (!strSerial.empty())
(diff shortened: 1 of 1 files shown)


qaisjp (approved):
Thanks for the PR. LGTM. :shipit:

patrikjuvonen (approved):
Nice find, thanks for the PR! I've now tested this and all seems good.

PR #945, opened 2019-05-17
downloadFile: downloading from another resource

### LINKED ISSUE #711: Using downloadFile to download a file of another resource

You can't download a file of another resource through downloadFile. Currently, the only way to do this is to export the downloadFile function in that resource.

I suggest using the _:resourceName/path_ format, like fileOpen.

TheNormalnij:
Fix issue #711

Event [onClientFileDownloadComplete](https://wiki.multitheftauto.com/wiki/OnClientFileDownloadComplete) has parameter requestResource: resource that called downloadFile. The source of this event is the root element of the resource that download file.


qaisjp:
@ccw808 — if I recall correctly, you once mentioned that you were unhappy with `downloadFile`. Do you think we should add new features to `downloadFile` in favour of encouraging other methods of sending files?

TheNormalnij:
[Resources](https://drive.google.com/open?id=1m1v5IhwTfJ3iUt14qXsRx2Rev4yhcjTY) for testing.

<sub>_Note from admin: please try to upload files directly. Thanks!_ [resources.zip](https://github.com/multitheftauto/mtasa-blue/files/3193353/resources.zip)</sub>

TheNormalnij:
I need a more flexible and convenient tool for downloading files. Now I have to use exports for this, but I don't see that this is a good way. Using `downloadFile` directly is more convenient.

onikyannn:
Any updates?

ghost:
Why dont you use fetchRemote ? I recently implemented abortRemoteRequest and getRemoteRequestInfo which both can handle downloads pretty well.

TheNormalnij:
fetchRemote don't math CRC hash. Also, downloadFile use specific HTTP server from mtaserver.conf and don't use ACL right (function is safe).

ccw808:
Looks good to me

ccw808 (approved):

Client/mods/deathmatch/logic/CSingularFileDownload.cpp
```diff
@@ -65,6 +66,15 @@ void CSingularFileDownload::CallFinished(bool bSuccess)
...
         Arguments.PushString(GetShortName());            // file name
         Arguments.PushBoolean(bSuccess);                 // Completed successfully?
+        if (m_pRequestResource)
+        {
+            Arguments.PushResource(m_pRequestResource);  // Resource that called downloadFile
+        }
+        else
+        {
+            Arguments.PushBoolean(false);                 // or false
+        }
```
qaisjp, line 76: @TheNormalnij, when is `m_pRequestResource` null?

TheNormalnij: I think that m_pRequestResource can be deleted during downloading

PR #957, opened 2019-05-24
Fix #798: Pin down startResource and stopResource behaviour

### LINKED ISSUE #798: Pin down startResource and stopResource behaviour
The latest release has this behaviour:

- stopResource returns false and warns if the resource is already stopped
- startResource returns true if the resource is already stopped

In master (via 2527293138927224979f40ba77bd6251964bf514) this behaviour is changed. (This change in API is not obvious from or mentioned in the commit message.)

A default MTA resource relies on undocumented behaviour that never worked (it expects stopResource to return true if already stopped).

We need to discuss and pin down this part of the Lua API.

StrixG:
startResource and stopResource functions now return `false` if already in the target state.
Closes #798


qaisjp (approved):
:shipit:

botder (approved):

PR #960, opened 2019-05-25
Don't queue a resource restart if resource is stopping

StrixG:
myonlake on Discord:
> since it goes to the end of queue and resource is stopped at that point, no, it should return false
> you can't restart a stopped resource
> basically that's the end result

### DIFF
--- a/Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
@@ -607,7 +607,7 @@ int CLuaResourceDefs::restartResource(lua_State* luaVM)
         return 1;
     }
 
-    if (pResource->IsActive())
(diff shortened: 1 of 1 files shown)


patrikjuvonen:
Just to clarify: the quote StrixG included in the PR description is just my opinion I answered to his question on Discord, I haven't requested any change. I mainly came to the conclusion that it should return false because of the fact that [restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) function returns false on a stopped resource. End state of a stopping resource should be "stopped", but is there a scenario where a resource does not reach that end state?

If this PR is something we wish to have merged, we will need scripting feedback because I have a feeling this commit, once applied, will break some scripts depending on this old logic. More so, why does this logic need to be changed now?

StrixG:
A stopping resource always reaches the `loaded` state.
restartResource doesn't work on a stopped resource so, I don't know any reason why restart should be queued.

qaisjp (approved):

PR #961, opened 2019-05-25
Fix #948: cancelEvent works onElementModelChange

### LINKED ISSUE #948: onElementModelChange can't be cancelled using setElementModel
**Describe the bug**
Using the script below doesn't work.

**Steps to reproduce**
Create a server-side script with this: (sorry, this website doesn't support indentation nor highlights)

```lua
addEventHandler ("onElementModelChange", root, function (oldModel, newModel)
        if (getElementType (source) == "player") then
            if (newModel == 12) then
                outputChatBox ("Skin locked! You can't use it.", source, 255, 30, 30)
                setElementModel (source, oldModel) -- According to Wiki, this should work.
            end
        end
end)
```

**Expected behaviour**
Skin should change back.

**How to "fix" the issue**
Just make setElementModel inside a timer, like this:

```lua
addEventHandler ("onElementModelChange", root, function (oldModel, newModel)
        if (getElementType (source) == "player") then
            if (newModel == 12) then
                outputChatBox ("Skin locked! You can't use it.", source, 255, 30, 30)
                setTimer (function (thePlayer)
                    setElementModel (thePlayer, oldModel)
                end, 100, 1, source)
            end
        end
end)
```

**Version**
Both v1.5.6-release-16616

TheNormalnij:
Test resource
[onElementChangeTest.zip](https://github.com/multitheftauto/mtasa-blue/files/3219368/onElementChangeTest.zip)


Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1594,28 +1600,37 @@ bool CStaticFunctionDefinitions::SetElementModel(CElement* pElement, unsigned sh
...
-            {
-                // Got a player in this seat and is it bigger than the supported amount
-                // of seats in this new vehicle
-                CPed* pPed = pVehicle->GetOccupant(i);
-                if (pPed && IS_PLAYER(pPed) && (i > ucMaxPassengers))
+            if (!pVehicle->CallEvent("onElementModelChange", Arguments))
+            {
+                // Change canceled
+                pVehicle->SetModel(usOldModel);
+                return false;
```
qaisjp: the `else` below is not required because of short-circuiting here

TheNormalnij: Yes, this is unnecessary

TheNormalnij:
In this solution, you can't use `setElementModel` inside `onElementModelChange`
It is possible to call `onElementModelChange` before actual change, but it can break existing scripts (`getElementModel` will return old value).

qaisjp:
Nice catch

Pirulax:
Doesn't this cause recursion? if you call setElementModel inside onElementModelChange, without any condition?

TheNormalnij:
setElementModel will return false without recursion if model will be changed inside onElementModelChange.

TheNormalnij:
Yeah, there will be recursion but it will be old behavior. Existing scripts will not be broken.

qaisjp (changes requested):
I've tested with the following script which has the advantage of not needing any player to be connect.

<details><summary>script</summary>

```lua
addCommandHandler("oect", function()
	local o = createObject(1337, 0, 0, 0)
	iprint("Created object at model 1337", inspect(o), "now adding event, and changing model")

	local event
	function event(oldModel, newModel)
		iprint("onElementModelChange received for source", inspect(source), ", oldModel=", oldModel, ", newModel=", newModel)
		iprint("Model is currently set to", getElementModel(o), "will recheck in 500ms")
		iprint("cancelEvent results:", inspect(cancelEvent()))
		setTimer(function()
			iprint("Model is currently set to", getElementModel(o))
			iprint("Destroying the element now!", inspect(destroyElement(o)))
		end, 500, 1)
	end
	addEventHandler("onElementModelChange", o, event)

	setElementModel(o, 1336)
end)
```
</details>

Just run the `oect` command and you should see some output like this in master:

image

With the pull request, it looks like this:

image

That looks good.


However, this could do with a little bit more work before it's ready.

1. What about the client event?
2. Could you provide test code for recursive calls ("Check for another call to setElementModel")

TheNormalnij:
```lua
	addEventHandler( "onElementModelChange", localPlayer, function( oldModel, newModel )
		if newModel < 100 or newModel > 105 then
			outputChatBox( "NOPE. Only 100-105" )
                        --cancelEvent()
			setElementModel( localPlayer, oldModel)
		end
	end )
```

Pirulax:
I don't think checking for recursive stuff is sensible. 
If it crashes, it crashes.
(it can crash with `setElementData` as well for example, yet, we don't check)

botder (approved):

Issue #909, opened 2019-05-02
setVehicleModelDummyPosition crash


Icensow:
**Describe the bug**
Using setVehicleModelDummyPosition on a vehicle with a replaced model causes a crash when leaving the server.

**To reproduce**
1. Start [this resource](https://github.com/multitheftauto/mtasa-blue/files/3139765/dummy_bug.zip)
2. Quit server.
3. See crash.

qaisjp:
This issue still occurs even after #1021

```
Version = 1.5.6-release-18729.0.000
Time = Thu Jul 18 12:08:20 2019
```

This must be fixed before 1.5.7 is released

ccw808:
I noticed all the vehicle dummy functions except `CModelInfoSA::ResetAllVehicleDummies` have a ` if (!IsVehicle())` check at the beginning.
Maybe add the check to `ResetAllVehicleDummies` as well?

forkerer:
The IsVehicle check has to be passed in other functions before the entry in map is even added, so all the changed dummies stored in it have passed that check already

Dutchman101:
Refer to this comment: [https://github.com/multitheftauto/mtasa-blue/issues/1009#issuecomment-516689379](https://github.com/multitheftauto/mtasa-blue/issues/1009#issuecomment-516689379) for additional information sources

PR #963, opened 2019-05-26
SetVehicleModelDummyPosition crash fix

Neproify:
Fixes issue #909. It is simple nullpointer check, because on disconnect the pVisualInfo is null.


Neproify:
BTW, I'm not sure if this is correct way to fix it, but it works. Maybe we should track why the pointer is null?

qaisjp:
Thank you for the pull request, but in this case we must fix the underlying issue instead of patching the crash.

qaisjp:
As suggested by @Dutchman101 and @forkerer — lets just merge this hacky fix and push 1.5.7 through the door.

Issue #908, opened 2019-05-02
Player / vehicle desync after car jacking, if jacker die


StifflersMom:
**Describe the bug**
If player A drives a car and player B starts jacking the car and player B dies at the moment, where he starts to pull player A out of the vehicle, player A become desynced:
From sight of player A: He is on foot, near the car.
From sight of player B and all others: Player a is stuck in the middle of vehicle.

Clientside player A:
getPedOccupiedVehicle(localPlayer) results to his vehicle
isPedInVehicle(localPlayer) results to false (!!!)

Serverside player A and all other clients:
getPedOccupiedVehicle(Player A) results to his vehicle
isPedInVehicle(Player A) results to true

While player A is on foot, he can run out of the streamig range. This results to being immortal for other player, while they see him in his car.

**To reproduce**
You need 2 players.
- start play
- as player A get in a vehicle as driver.
- as player B press F near the vehicle
- at the moment, where B opens the door, kill B

**Expected behaviour**
No desync. If A performs the vehicle exit animation, he should be on foot for all :)

**Screenshots**
Player A before jacked
![grafik](https://user-images.githubusercontent.com/12788736/57102752-8474ad00-6d24-11e9-9b05-c1362fa528c2.png)
Player A after jacked
![grafik](https://user-images.githubusercontent.com/12788736/57102866-b0902e00-6d24-11e9-97bf-c930c5a52e5d.png)
Player B before jacking
![grafik](https://user-images.githubusercontent.com/12788736/57103089-25636800-6d25-11e9-93d7-d848885fe129.png)

Player B after killed respawned while jacking
![grafik](https://user-images.githubusercontent.com/12788736/57102970-e0d7cc80-6d24-11e9-8ff8-4d61bc31f190.png)

**Version**
MTA:SA Server v1.5.6-release-16612 , Client v1.5.6-release-16705

**Additional context**
**In case this is being used for fun to "ruin" players game, this should be fixed with priority, I think.**

Neproify:
IMO it is related to issue #928. I think if dead players would send sync packets, then it would work as it should. Is it working after you spawn player A again? @StifflersMom

Neproify:
@StifflersMom Can you check if this problem still exists on latest master?

StifflersMom:
Checked with
MTA:SA Server v1.5.6-release-18572
Multi Theft Auto v1.5.6-release-18560
The bug still exists like described.

StifflersMom:
[...]

PR #964, opened 2019-05-26
Fixed desync of position on server side when player is dead

### LINKED ISSUE #928: Position desync with dead players
**Describe the bug**
getElementPosition() returns wrong values server side when a player is dead.

**To reproduce**
1. Go to editor
2. Use Jetpack, fly into the air
3. Use srun getElementPosition(me) -> Notice how your z position is pretty high
4. Kill the player
5. The player will fall to the earth (Client side Z position ~13.3)
6. Server-side getElementPosition(me) will still return the coordinate where the player died

**Expected behaviour**
getElementPosition() should not return the death position, it should return the current position because it can happen that a dead player moves away from his death position, for example when falling.

Neproify:
Does what it says. Not tested with more than local player, however it fixes #928 and probably(not tested) #908. Should fix most of desyncs when player is dead, because right now client doesn't send any sync when player died.


qaisjp (approved):
Thanks!

qaisjp:
Btw, instead of making pull requests on your fork, you can follow the instructions [here](https://github.com/multitheftauto/mtasa-blue/wiki/Git-cheat-sheet#pulling-the-latest-changes-from-multitheftautomtasa-blue).

AlexTMjugador:
Microoptimization: you could append the body of the if statement that starts on line 411 to the end of the if statement you changed, because they now check the same condition. That way you surely avoid checking the same condition twice, and reduce 2 lines of code.

PR #970, opened 2019-05-28
Addendum to #964: Remove redundant check on CNetAPI::DoPulse

Neproify:
Addendum to #964: Remove redundant check on CNetAPI::DoPulse


qaisjp (commented):
cheers

PR #971, opened 2019-05-28
Add engineResetModelLODDistance function

### LINKED ISSUE #894: engineSetModelLODDistance can break the default map
There are many objects which have a LOD distance > 325.
If you do any engineSetModelLODDistance to these objects, you can't restore the default until reconnect.

engineResetModelLODDistance would be nice.

one line runcode to see the desaster:
`crun for i=1337,18000 do test=engineGetModelLODDistance(i)  if test then engineSetModelLODDistance(i,test) end end
`

Lpsd:
Resolves #894 

Some default LOD distances are above 325. When changing their LOD distances, you have no way to go back to their default without reconnecting. (using `engineSetModelLODDistance` you are unable to go above 325)

Example usage code (sets all LODs to 325 or below, then resets the values after 5 seconds)

```lua
addCommandHandler("testlod", function()
	local changedLODs = {}
	
	for i=1337,18000 do
		local distance = engineGetModelLODDistance(i)
		if distance then
			engineSetModelLODDistance(i,distance)
			changedLODs[#changedLODs+1] = i
		end
	end

	setTimer(function()
		for i, model in ipairs(changedLODs) do
			local reset = engineResetModelLODDistance(model)
		end
	end, 5000, 1)	
end)
```

Currently this function will return `false` if the LOD distance has not yet been changed via `engineSetModelLODDistance`, or if the current LOD distance is equal to the default LOD distance. I'm fairly certain this falls in line with other "similar" functions.


qaisjp (approved):
Looks good thank you please squash

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -705,6 +707,35 @@ int CLuaEngineDefs::EngineSetModelLODDistance(lua_State* luaVM)
     return 1;
 }
 
+int CLuaEngineDefs::EngineResetModelLODDistance(lua_State* luaVM)
+{
+    SString          strModel = "";
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadString(strModel);
+
+    if (!argStream.HasErrors())
```
patrikjuvonen: As per #821 we should do a hard error here like so:
```cpp
if (argStream.HasErrors())
```
and then move the passing code after that.

patrikjuvonen (dismissed):
Thanks for the PR! See to my comment about the new error behavior that should be adopted and this is good to go.

Lpsd:
Requested changes have been made. This is ready to go

PR #972, opened 2019-05-28
Getting player language on the server

StrixG:
# Summary

1. Added `getPlayerLocalization` server function which returns language code.
2. Added `localization` argument to the `onPlayerConnect` event.

I'm not sure language name is needed on the server, so I made available only language code.

## Examples
`getPlayerLocalization` function:
```lua
local messages = {
  en_US = "Welcome!",
  ru = "Добро пожаловать!"
}
addEventHandler("onPlayerJoin", root,
  function ()
    local language = getPlayerLocalization(source)
    local message = messages[language] or messages["en_US"]
    outputChatBox(message, source, 255, 255, 255)
  end
)
```
`localization` argument:
```lua
local messages = {
  en_US = "You are not whitelisted!",
  ru = "Вы не занесены в белый список!"
}
addEventHandler("onPlayerConnect", root,
  function (nick, ip, _, serial, versionNumber, versionString, localization)
    -- some checks here
    local message = messages[localization] or messages["en_US"]
    cancelEvent(true, message)
  end
)
```


qaisjp:
> I'm not sure about backwards compatibility.

Yes you need bitstream version guards and to bump bitstream version

ccw808:
Should we put all player settings/options info into something like `getPlayerSettings`

PR #973, opened 2019-05-30
Initial version of new style Lua argument parsing

sbx320:
Initial stuff for automated Lua argument parsing by applying template magic.

Basically this allows us to directly use C++ functions without needing to handle the argument parsing via argstreams. For example
```cpp
int CLuaCryptDefs::Hash(lua_State* luaVM)
{
    //  string hash ( string type, string data )
    EHashFunctionType hashFunction;
    SString           strSourceData;

    argStream.ReadEnumString(hashFunction);
    argStream.ReadString(strSourceData);

    if (!argStream.HasErrors())
    {
        SString strResult = GenerateHashHexString(hashFunction, strSourceData);
        lua_pushstring(luaVM, strResult.ToLower());
        return 1;
    }
    else
        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());

    return 1;
}
```
becomes 
```cpp
std::string CLuaCryptDefs::Hash(EHashFunctionType hashFunction, std::string strSourceData)
{
    return GenerateHashHexString(hashFunction, strSourceData);
}
```
which avoids error and reduces the required amount of code massively. 

This PR also allows registering functions via the ArgumentParser template (rather than the ArgumentParserWarn template), which causes usage mistakes to be considered a hard error in accordance with #821. As a usage example I mostly converted CLuaCryptDefs to use the new style. 

## Supported Types:

- `bool`
- `int`, (todo: ushort, char, uchar)
- `std::string` (todo: SString?)
- `CLuaFunctionRef` (argument only)
- enums (with StringToEnum)
- Script entities (via UserDataCast) 
- `std::vector<T>` if T is supported (expects a table with consistent value types) (todo: as result)
- `std::unordered_map<K, V>` if K and V are supported (todo: as result)
- `std::optional<T>` if T is supported (expects either a T or nil/none) (todo: as result)
- `std::variant<Ts...>` if all Ts... are supported
- `nullptr` (return type only) to indicate `nil`

Additionally a lua_State* may be taken as an argument, which will be passed through directly. (We use that in some places to access things based on the running script VM)

## Restrictions:

- Currently this only supports default-constructible complex types. For example one can not read a `std::unique_ptr` (unlikely to change)
- ~~No full support for overloads yet. Using `std::variant` and `std::optional` we have some support for overloading, but a function that expects very different types is not fun to handle~~
- No vararg support (yet)

## Notes:
- See CLuaCryptDefs for usage examples
- Functions may throw a `std::invalid_argument` exception to indicate argument failures inside the functions. 
- This requires VS2019 on Windows with the v142 toolset (VS2017 does *not* work due to compiler bugs)

## Todos:

- [x] Add debug assertions for Lua stack manipulation (compare gettop vs the expected value)
- [x] Support additional types (see above, maybe more?)
- [x] Figure out a way to handle overloads nicely
- [x] Error messages are not generated yet
- [x] Test on Linux
- [x] Figure out why LuaBasic.h is detected as ObjectiveC by clang format (help?)

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -74,10 +74,14 @@ protected:
         return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
     }
 
-    template <auto T, auto U>
+    // Special cases for overloads
+    template <auto T, auto U, auto... Ts>
     static inline int ArgumentParserWarn(lua_State* L)
     {
-        return CLuaFunctionParser<false, CLuaOverloadParser<T, U>::Call>()(L, m_pScriptDebugging);
+        if constexpr (sizeof...(Ts) == 0)
+            return ArgumentParserWarn<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
+        else
+            return ArgumentParserWarn<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
     }
 
     // New style: hard error on usage mistakes
@@ -88,9 +92,12 @@ protected:
     }
 
     // Overload variant
-    template <auto T, auto U>
+    template <auto T, auto U, auto... Ts>
     static inline int ArgumentParser(lua_State* L)
     {
-        return CLuaFunctionParser<true, CLuaOverloadParser<T, U>::Call>()(L, m_pScriptDebugging);
+        if constexpr (sizeof...(Ts) == 0)
+            return ArgumentParser<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
+        else
+            return ArgumentParser<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
     }
 };
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -64,4 +64,41 @@ public:
     static CClientDFFManager*         m_pDFFManager;
     static CClientColModelManager*    m_pColModelManager;
     static CRegisteredCommands*       m_pRegisteredCommands;
+
+protected:
+    // Old style: Only warn on failure. This should
+    // not be used for new functions
+    template <auto T>
+    static inline int ArgumentParserWarn(lua_State* L)
+    {
+        int iReturnValue = CLuaFunctionParser<T>()(L);
+        if (iReturnValue < 0)
+        {
+            m_pScriptDebugging->LogCustom(L, "todo fix error message");
+            lua_pushboolean(L, false);
+            return 1;
+        }
+        return iReturnValue;
+    }
+
+    // New style: hard error on usage mistakes
+    template <auto T>
+    static inline int ArgumentParser(lua_State* L)
+    {
+        try
+        {
+            int iReturnValue = CLuaFunctionParser<T>()(L);
+            if (iReturnValue < 0)
+            {
+                luaL_error(L, "todo fix error message");
+                return 1;
+            }
+            return iReturnValue;
+        }
+        catch (CLuaError& e)
+        {
+            luaL_error(L, e.what());
+            return 0;
+        }
+    }
 };
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -71,34 +71,13 @@ protected:
     template <auto T>
     static inline int ArgumentParserWarn(lua_State* L)
     {
-        int iReturnValue = CLuaFunctionParser<T>()(L);
-        if (iReturnValue < 0)
-        {
-            m_pScriptDebugging->LogCustom(L, "todo fix error message");
-            lua_pushboolean(L, false);
-            return 1;
-        }
-        return iReturnValue;
+        return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
     }
 
     // New style: hard error on usage mistakes
     template <auto T>
     static inline int ArgumentParser(lua_State* L)
     {
-        try
-        {
-            int iReturnValue = CLuaFunctionParser<T>()(L);
-            if (iReturnValue < 0)
-            {
-                luaL_error(L, "todo fix error message");
-                return 1;
-            }
-            return iReturnValue;
-        }
-        catch (CLuaError& e)
-        {
-            luaL_error(L, e.what());
-            return 0;
-        }
+        return CLuaFunctionParser<true, T>()(L, m_pScriptDebugging);
     }
 };
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -74,10 +74,23 @@ protected:
         return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
     }
 
+    template <auto T, auto U>
+    static inline int ArgumentParserWarn(lua_State* L)
+    {
+        return CLuaFunctionParser<false, CLuaOverloadParser<T, U>::Call>()(L, m_pScriptDebugging);
+    }
+
     // New style: hard error on usage mistakes
     template <auto T>
     static inline int ArgumentParser(lua_State* L)
     {
         return CLuaFunctionParser<true, T>()(L, m_pScriptDebugging);
+    }
+
+    // Overload variant
+    template <auto T, auto U>
+    static inline int ArgumentParser(lua_State* L)
+    {
+        return CLuaFunctionParser<true, CLuaOverloadParser<T, U>::Call>()(L, m_pScriptDebugging);
     }
 };
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -87,7 +87,7 @@ protected:
         // Combine functions
         using Overload = CLuaOverloadParser<PaddedFunctionA::Call, PaddedFunctionB::Call>;
 
-        return ArgumentParserWarn<Ret, Overload::Call, Functions...>(L);
+        return ArgumentParserWarn<ReturnOnError, Overload::Call, Functions...>(L);
     }
 
     // New style: hard error on usage mistakes
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -67,28 +67,29 @@ public:
 
 protected:
     // Old style: Only warn on failure. This should
-    // not be used for new functions
-    template <auto T>
+    // not be used for new functions. First template argument
+    // Is a value used as result on invalid argument
+    template <auto Ret, auto T>
     static inline int ArgumentParserWarn(lua_State* L)
     {
-        return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
+        return CLuaFunctionParser<false, Ret, T>()(L, m_pScriptDebugging);
     }
 
     // Special cases for overloads
-    template <auto T, auto U, auto... Ts>
+    template <auto Ret, auto T, auto U, auto... Ts>
     static inline int ArgumentParserWarn(lua_State* L)
     {
         if constexpr (sizeof...(Ts) == 0)
-            return ArgumentParserWarn<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
+            return ArgumentParserWarn<Ret, CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
         else
-            return ArgumentParserWarn<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
+            return ArgumentParserWarn<Ret, CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
     }
 
     // New style: hard error on usage mistakes
     template <auto T>
     static inline int ArgumentParser(lua_State* L)
     {
-        return CLuaFunctionParser<true, T>()(L, m_pScriptDebugging);
+        return CLuaFunctionParser<true, nullptr, T>()(L, m_pScriptDebugging);
     }
 
     // Overload variant
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
@@ -67,22 +67,27 @@ public:
 
 protected:
     // Old style: Only warn on failure. This should
-    // not be used for new functions. First template argument
-    // Is a value used as result on invalid argument
-    template <auto Ret, auto T>
+    // not be used for new functions. ReturnOnError 
+    // must be a value to use as result on invalid argument
+    template <auto ReturnOnError, auto T>
     static inline int ArgumentParserWarn(lua_State* L)
     {
-        return CLuaFunctionParser<false, Ret, T>()(L, m_pScriptDebugging);
+        return CLuaFunctionParser<false, ReturnOnError, T>()(L, m_pScriptDebugging);
     }
 
-    // Special cases for overloads
-    template <auto Ret, auto T, auto U, auto... Ts>
+    // Special case for overloads
+    // This combines multiple functions into one (via CLuaOverloadParser)
+    template <auto ReturnOnError, auto FunctionA, auto FunctionB, auto... Functions>
     static inline int ArgumentParserWarn(lua_State* L)
     {
-        if constexpr (sizeof...(Ts) == 0)
-            return ArgumentParserWarn<Ret, CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
-        else
-            return ArgumentParserWarn<Ret, CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
+        // Pad functions to have the same number of parameters by 
+        // filling both up to the larger number of parameters with dummy_type arguments
+        using PaddedFunctionA = pad_func_with_func<FunctionA, FunctionB>;
+        using PaddedFunctionB = pad_func_with_func<FunctionB, FunctionA>;
+        // Combine functions
+        using Overload = CLuaOverloadParser<PaddedFunctionA::Call, PaddedFunctionB::Call>;
+
+        return ArgumentParserWarn<Ret, Overload::Call, Functions...>(L);
     }
 
     // New style: hard error on usage mistakes
@@ -92,13 +97,18 @@ protected:
         return CLuaFunctionParser<true, nullptr, T>()(L, m_pScriptDebugging);
     }
 
-    // Overload variant
-    template <auto T, auto U, auto... Ts>
+    // Special case for overloads
+    // This combines multiple functions into one (via CLuaOverloadParser)
+    template <auto FunctionA, auto FunctionB, auto... Functions>
     static inline int ArgumentParser(lua_State* L)
     {
-        if constexpr (sizeof...(Ts) == 0)
-            return ArgumentParser<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
-        else
-            return ArgumentParser<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call, Ts...>(L);
+        // Pad functions to have the same number of parameters by
+        // filling both up to the larger number of parameters with dummy_type arguments
+        using PaddedFunctionA = pad_func_with_func<FunctionA, FunctionB>;
+        using PaddedFunctionB = pad_func_with_func<FunctionB, FunctionA>;
+        // Combine functions
+        using Overload = CLuaOverloadParser<PaddedFunctionA::Call, PaddedFunctionB::Call>;
+
+        return ArgumentParser<Overload::Call, Functions...>(L);
     }
 };
--- a/Server/mods/deathmatch/logic/CGame.h
+++ b/Server/mods/deathmatch/logic/CGame.h
@@ -57,6 +57,7 @@ class CGame;
 #include "lua/CLuaManager.h"
 
 #include "CLightsyncManager.h"
+#include "CBanManager.h"
 
 // Forward declarations
 class ASE;
--- a/Server/mods/deathmatch/logic/CGame.h
+++ b/Server/mods/deathmatch/logic/CGame.h
@@ -63,7 +63,6 @@ class CGame;
 class ASE;
 class CAccessControlListManager;
 class CAccountManager;
-class CBanManager;
 class CBlipManager;
 class CClock;
 class CColManager;
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -12,6 +12,7 @@
 #include <variant>
 #include <SharedUtil.Template.h>
 #include "lua/CLuaFunctionParseHelpers.h"
+#include "lua/CLuaStackChecker.h"
 #include "lua/LuaBasic.h"
 
 template <bool, auto*>
@@ -246,6 +247,8 @@ struct CLuaFunctionParser<ErrorOnFailure, Func>
     template <typename T>
     inline T PopUnsafe(lua_State* L, std::size_t& index)
     {
+        // Expect no change in stack size
+        LUA_STACK_EXPECT(0); 
         // trivial types are directly popped
         if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, int> || std::is_same_v<T, float> || std::is_same_v<T, double> ||
                       std::is_same_v<T, short> || std::is_same_v<T, unsigned int> || std::is_same_v<T, unsigned short> || std::is_same_v<T, bool>)
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -16,13 +16,7 @@
 #include "lua/CLuaStackChecker.h"
 #include "lua/LuaBasic.h"
 
-template <bool, auto*>
-struct CLuaFunctionParser
-{
-};
-
-template <bool ErrorOnFailure, typename Ret, typename... Args, auto (*Func)(Args...)->Ret>
-struct CLuaFunctionParser<ErrorOnFailure, Func>
+struct CLuaFunctionParserBase
 {
     std::size_t iIndex = 1;
     std::string strError = "";
@@ -91,7 +85,8 @@ struct CLuaFunctionParser<ErrorOnFailure, Func>
                 std::string strValue(szValue, iLen);
                 if (strValue.length() > 10)
                 {
-                    strValue.resize(10);            // Limit to 10 characters
+                    // Limit to 10 characters
+                    strValue.resize(10);
                     strValue[9] = '.';
                     strValue[8] = '.';
                     strValue[7] = '.';
@@ -216,9 +211,9 @@ struct CLuaFunctionParser<ErrorOnFailure, Func>
         // and can be fetched from a userdata
         if constexpr (std::is_pointer_v<T> && std::is_class_v<std::remove_pointer_t<T>>)
             return iArgument == LUA_TUSERDATA || iArgument == LUA_TLIGHTUSERDATA;
-        
+
         // dummy type is used as overload extension if one overload has fewer arguments
-        // thus it is only allowed if there are no further args on the Lua side 
+        // thus it is only allowed if there are no further args on the Lua side
         if constexpr (std::is_same_v<T, dummy_type>)
             return iArgument == LUA_TNONE;
     }
@@ -260,9 +255,20 @@ struct CLuaFunctionParser<ErrorOnFailure, Func>
         if constexpr (std::is_same_v<T, dummy_type>)
             return dummy_type{};
         // trivial types are directly popped
-        else if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, int> || std::is_same_v<T, float> || std::is_same_v<T, double> ||
-                           std::is_same_v<T, short> || std::is_same_v<T, unsigned int> || std::is_same_v<T, unsigned short> || std::is_same_v<T, bool>)
+        else if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, int> || std::is_same_v<T, short> || std::is_same_v<T, bool> ||
+                           std::is_same_v<T, unsigned int> || std::is_same_v<T, unsigned short>)
             return lua::PopTrivial<T>(L, index);
+        // floats/doubles may not be NaN
+        else if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double>)
+        {
+            T value = lua::PopTrivial<T>(L, index);
+            if (std::isnan(value))
+            {
+                SString strMessage("Expected number at argument %d, got NaN", index);
+                strError = strMessage;
+            }
+            return value;
+        }
         else if constexpr (std::is_enum_v<T>)
         {
             // Enums are considered strings in Lua
@@ -368,7 +374,16 @@ struct CLuaFunctionParser<ErrorOnFailure, Func>
             return static_cast<T>(result);
         }
     }
+};
 
+template <bool, auto*>
+struct CLuaFunctionParser
+{
+};
+
+template <bool ErrorOnFailure, typename Ret, typename... Args, auto (*Func)(Args...)->Ret>
+struct CLuaFunctionParser<ErrorOnFailure, Func> : CLuaFunctionParserBase
+{
     template <typename... Params>
     inline auto Call(lua_State* L, Params&&... ps)
     {
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -22,21 +22,23 @@ struct CLuaFunctionParserBase
     std::string strError = "";
     std::string strErrorFoundType = "";
 
+    // Translates a variant type to a list of names separated by slashes
+    // std::variant<bool, int, float> => bool/int/float
     template <typename T>
-    inline void typeToNameVariant(SString& accumulator)
+    inline void TypeToNameVariant(SString& accumulator)
     {
         using param = typename is_variant<T>::param1_t;
         if (accumulator.length() == 0)
-            accumulator = typeToName<param>();
+            accumulator = TypeToName<param>();
         else
-            accumulator += "/" + typeToName<param>();
+            accumulator += "/" + TypeToName<param>();
 
         if constexpr (is_variant<T>::count != 1)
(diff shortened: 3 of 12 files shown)


StrixG (approved):

azure-pipelines.yml
```diff
@@ -2,7 +2,7 @@ resources:
   containers:
   - container: ci
     image: sbx320/mtasa-blue-azure:latest
-
+
```
qaisjp: Achtung
```suggestion

```

ghost:
Brilliant! So, will we be using `ArgumentParserWarn` for old functions and `ArgumentParser` for newer ones?

ghost (approved):

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -57,6 +57,7 @@ class CGame;
 #include "lua/CLuaManager.h"
 
 #include "CLightsyncManager.h"
+#include "CBanManager.h"
```
qaisjp: Forward declaration `class CBanManager;` is no longer necessary (see line 65)

sbx320:
Yes that is the plan. We need to be careful with converting existing code though, as the behavior of this new stuff may be different in some edge cases.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+        if (strError.length() != 0)
+        {
+            if constexpr (ErrorOnFailure)
+            {
+                luaL_error(L, strError.c_str());
+            }
+            else
+            {
+                pScriptDebugging->LogCustom(L, strError.c_str());
+                lua_pushboolean(L, false);
```
jushar: If I understood it correctly, in "warning-mode" this always returns _false_.   
As mentioned in https://github.com/multitheftauto/mtasa-blue/issues/821#issue-411212784 we probably need yet another case (in which _nil_ is returned) though.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+
+        // std::unordered_map<k,v> is used for maps built from tables
+        if constexpr (is_5specialization<T, std::unordered_map>::value)
+            return iArgument == LUA_TTABLE;
+
+        // CLuaFunctionRef is used for functions
+        if constexpr (std::is_same_v<T, CLuaFunctionRef>)
+            return iArgument == LUA_TFUNCTION;
+
+        // lua_State* can be taken as argument anywhere
```
qaisjp: For the sake of clarity, consider only allowing `lua_State*` to be the first argument

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -33,247 +34,114 @@ void CLuaCryptDefs::LoadFunctions()
...
-        SharedUtil::TeaDecode(result, key, &str);
-        lua_pushstring(luaVM, str);
-        return 1;
-    }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
-
-    lua_pushboolean(luaVM, false);
-    return 1;
+    SString fixme = str;
```
qaisjp: `fixme` introduced

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+            return typeToName<param_t>();
+        }
+        else if constexpr (is_2specialization<T, std::vector>::value)
+            return "table";
+        else if constexpr (is_5specialization<T, std::unordered_map>::value)
+            return "table";
+        else if constexpr (std::is_same_v<T, CLuaFunctionRef>)
+            return "function";
+        else if constexpr (std::is_same_v<T, lua_State*>)
+            return "";            // not reachable
```
qaisjp, line 64: Can we enforce this at compile time?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+#include "lua/CLuaStackChecker.h"
+#include "lua/LuaBasic.h"
+
+struct CLuaFunctionParserBase
+{
+    std::size_t iIndex = 1;
+    std::string strError = "";
+    std::string strErrorFoundType = "";
+
+    template <typename T>
```
qaisjp, line 27: Not immediately obvious that this method is supposed to translate `std::variant<int, bool>` to `"int/bool"`. Could do with a comment here.

```suggestion
    // typeToNameVariant translates `std::variant<int, bool>` to `"int/bool"`
    template <typename T>
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+        else if constexpr (is_variant<T>::value)
+        {
+            SString strTypes;
+            typeToNameVariant<T>(strTypes);
+            return strTypes;
+        }
+        else if constexpr (std::is_pointer_v<T> && std::is_class_v<std::remove_pointer_t<T>>)
+            return GetClassTypeName((T)0);
+        else if constexpr (std::is_same_v<T, dummy_type>)
+            return "";
```
qaisjp, line 74: What happens if none of these are met?

```suggestion
    else {
            dassert("Missing type name in typeToName" && 0)
            return "";
    }
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    static SString ResolveParameter(lua_State* L, std::size_t index)
```
qaisjp: Could also do with a brief explanation. Before reading the source, I interpreted `Resolve` to mean the same thing as `TypeMatch`.

```suggestion
    // ResolveParameter translates an argument (by index) to its typed string representation
    // This is used in messages like `expected string at argument 5, got number (5)`
    //                                                                   ^^^^^^^^^^
    static SString ResolveParameter(lua_State* L, std::size_t index)
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+            accumulator = typeToName<param>();
+        else
+            accumulator += "/" + typeToName<param>();
+
+        if constexpr (is_variant<T>::count != 1)
+            return typeToNameVariant<is_variant<T>::rest_t>(accumulator);
+    }
+
+    template <typename T>
+    inline SString typeToName()
```
qaisjp: Can we make this (and other methods) `constexpr`? (This compiles.)
```suggestion
    constexpr inline SString typeToName()
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+            case LUA_TUSERDATA:
+                return GetUserDataClassName(*((void**)lua_touserdata(L, index)), L);
+            case LUA_TLIGHTUSERDATA:
+                return GetUserDataClassName(lua_touserdata(L, index), L);
+        }
+        return "";
+    }
+
+    // Pop should remove a T from the Lua Stack after verifying that it is a valid type
+    // Pop may also throw a LuaArgumentError to indicate failure
```
qaisjp: I cannot find any instance of `LuaArgumentError` anywhere in this PR.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+        }
+        return PopUnsafe<T>(L, index);
+    }
+
+    // Special type matcher for variants. Returns -1 if the type does not match
+    // returns n if the nth type of the variant matches
+    template <typename T>
+    inline int TypeMatchVariant(lua_State* L, std::size_t index)
+    {
+        if constexpr (std::is_same_v<T, std::variant<>>)
```
qaisjp, line 149:
```suggestion
        // Error out if the variant contains 0 types.
        if constexpr (std::is_same_v<T, std::variant<>>)
```

Consider short-circuiting in this method as well.

Nice use of recursion. Make it obvious by:
- short-circuiting
- using comments like `// Base case: when variant is empty`, `// Base case: when matching with first type in the variant`, `// General case: attempt next type in variant`

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+
+        // dummy type is used as overload extension if one overload has fewer arguments
+        // thus it is only allowed if there are no further args on the Lua side
+        if constexpr (std::is_same_v<T, dummy_type>)
+            return iArgument == LUA_TNONE;
+    }
+
+    // Special PopUnsafe for variants
+    template <typename T, std::size_t currIndex = 0>
+    inline T PopUnsafeVariant(lua_State* L, std::size_t& index, int vindex)
```
qaisjp, line 231: This method would benefit from short-circuiting.

Follow other recommendations for recursion.

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
```diff
@@ -64,4 +64,40 @@ class CLuaDefs
...
+    {
+        return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
+    }
+
+    // Special cases for overloads
+    template <auto T, auto U, auto... Ts>
+    static inline int ArgumentParserWarn(lua_State* L)
+    {
+        if constexpr (sizeof...(Ts) == 0)
+            return ArgumentParserWarn<CLuaOverloadParser<pad_func_with_func<T, U>::Call, pad_func_with_func<U, T>::Call>::Call>(L);
```
qaisjp: This is a little complicated to read. Consider splitting across multiple lines.

At first I thought the `sizeof [...] == 0` was some sort of optimisation. I was wrong!

If I am reading your code correctly, this a pretty fantastic use of recursion. I think it could be made more obvious.

To make the recursion more obvious:
- Consider annotating the `if` with something like `// This is the base case for two overloads` (and similarly for the second one
- Consider guarding off (dropping the `else`) to highlight the base case

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
```diff
@@ -64,4 +64,40 @@ class CLuaDefs
...
+    // Old style: Only warn on failure. This should
+    // not be used for new functions
+    template <auto T>
+    static inline int ArgumentParserWarn(lua_State* L)
+    {
+        return CLuaFunctionParser<false, T>()(L, m_pScriptDebugging);
+    }
+
+    // Special cases for overloads
+    template <auto T, auto U, auto... Ts>
```
qaisjp: Even though `Special cases for overloads` is quite obvious, the fact that each of these template parameters are functions was not obvious to me.

Perhaps you could do
```suggestion
    template <auto F, auto G, auto... Fs>
```
or
```suggestion
    template <auto F, auto G, auto... Hs>
```
or even,
```suggestion
    template <auto F1, auto F2, auto... Fs>
```

I think the last one is the most obvious to me.

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
```diff
@@ -64,4 +64,40 @@ class CLuaDefs
...
+    }
+
+    // New style: hard error on usage mistakes
+    template <auto T>
+    static inline int ArgumentParser(lua_State* L)
+    {
+        return CLuaFunctionParser<true, T>()(L, m_pScriptDebugging);
+    }
+
+    // Overload variant
```
qaisjp: Try to keep the comment consistent with `ArgumentParserWarn` (and try to avoid referring to this as a "variant" as it could be confused with `std::variant`).

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+{
+    using param1_t = Arg1;
+    using rest_t = std::variant<Args...>;
+    static constexpr auto count = sizeof...(Args) + 1;
+};
+
+/**
+    nth_element
+
+    Returns the nth element of a parameter pack
```
qaisjp: This is also another very cool use of recursion.

Consider making this obvious by:
- describing the implementation briefly in the main comment
- moving the base case _above_ the general case
- commenting both implementations with either `// Base case (I == 0)` or `// General case (I > 0)`

```suggestion
    Returns the nth element of a parameter pack.

    It does this by recursively selecting the `n - 1`th item.
    
    For example, to get the Nth item of parameter pack `Args` you:
        nth_element_t<N, Args...>>
```

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+template <typename Test, template <typename, typename, typename, typename, typename> class Ref>
+struct is_5specialization : std::false_type
+{
+};
+
+template <template <typename, typename, typename, typename, typename> class Ref, typename Arg1, typename Arg2, typename Arg3, typename Arg4, typename Arg5>
+struct is_5specialization<Ref<Arg1, Arg2, Arg3, Arg4, Arg5>, Ref> : std::true_type
+{
+    using param1_t = Arg1;
+    using param2_t = Arg2;
```
qaisjp, line 63: If there's a specific reason you exclude `param3_t` (etc.), consider including the reason as a comment.

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+    is_Nspecialization
+
+    These structs allow testing whether a type is a specialization of a
+    class template
+
+    Note: Take care of optional parameters. For example std::unordered_map
+    has 5 template arguments, thus is_5specialization needs to be used, rather
+    than is_2specialization (Key, Value).
```
qaisjp, line 11: Consider briefly explaining how `is_2specialization` (both `false` and `true`) works.

I'm finding the implementation of the `specialization` templates a little tricky to read. If you dumb it down it would be easier for future developers to consume.

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+    using type = std::variant<Ts...>;
+};
+// Empty + Variant = Variant
+template <typename... Ts>
+struct common_variant<std::variant<>, std::variant<Ts...>>
+{
+    using type = std::variant<Ts...>;
+};
+
+template <typename T, typename... Ts>
```
qaisjp, line 165: Comments for the following `common_variant` definitions would be useful. The previous comments here are great!

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+// dummy_type
+// generic dummy type
+struct dummy_type
+{
+};
+
+
+// n_tuple: Constructs a tuple of size N
+//  n_tuple<2>::type == std::tuple<dummy_type, dummy_type>
+template <std::size_t, bool B = false, typename... Args>
```
qaisjp: What does `B` mean?

At a deeper look I think it might mean `Base case`. If that's correct, consider making this use of recursion obvious.

```suggestion
template <std::size_t, bool BaseCase = false, typename... Args>
```

Shared/sdk/SharedUtil.Template.h
```diff
@@ -0,0 +1,193 @@
...
+
+template <std::size_t N, typename... Args>
+struct n_tuple<N, false, Args...>
+{
+    using type = typename n_tuple<N, sizeof...(Args) + 1 >= N, Args..., dummy_type>::type;
+};
+
+// pad_func_with_func
+// pads Func with as many dummy_type arguments as needed to 
+// have the same number of arguments as FuncB has
```
qaisjp, line 218: The rest of this file would benefit from an explanation (and I have not really attempted to read it).

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -0,0 +1,99 @@
...
+    int Push(lua_State* L, float val);
+    int Push(lua_State* L, double val);
+    int Push(lua_State* L, unsigned int val);
+    int Push(lua_State* L, unsigned short val);
+
+
+    template <typename... Ts>
+    int Push(lua_State* L, const std::variant<Ts...>&& val)
+    {
+        return std::visit([L](auto&& value) -> int { return Push(L, value); }, val);
```
qaisjp, line 74: What's going on here?

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -0,0 +1,99 @@
...
+    {
+        lua_newtable(L);
+        int i = 1;
+        for (auto&& v : val)
+        {
+            Push(L, i++);
+            Push(L, v);
+            lua_settable(L, -3);
+        }
+        return 1;
```
qaisjp, line 99: Seeing `return 1` raised red flags. Consider including a comment here like this:
```suggestion

        // We only leave the table reference on the stack
        return 1;
```

This applies to `unordered_map` as well.

You can make this more obvious by editing your description of `Push` near the start of this file:

```diff
- // The return value must be the amount of items pushed to the stack, which should
+ // The return value must be the *net* amount of items pushed to the stack, which should
                                  ^^^^^
```

Or "_relative_", or "_number of items \*left\* on the stack_". I prefer the first or last suggested wordings here (i.e, not "_relative_").

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -0,0 +1,99 @@
...
+        for (auto&& [k, v] : val)
+        {
+            Push(L, k);
+            Push(L, v);
+            lua_settable(L, -3);
+        }
+        return 1;
+    }
+
+    template <typename T>
```
qaisjp, line 118: `enable_if` is very cool! It's obvious, but could do with a comment.
```suggestion
    // Overload just for types that are enums
    template <typename T>
```

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -0,0 +1,99 @@
...
+    typename std::enable_if_t<std::is_enum_v<T>, int> Push(lua_State* L, const T&& val)
+    {
+        return Push(L, EnumToString(val));
+    }
+
+    template <typename T>
```
qaisjp, line 125:
```suggestion
    // Overload just for types that are pointers to classes (and not primitives)
    template <typename T>
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+            case LUA_TNONE:
+                return "none";
+            case LUA_TTABLE:
+                return "table";
+            case LUA_TFUNCTION:
+                return "function";
+            case LUA_TTHREAD:
+                return "coroutine";
+            case LUA_TUSERDATA:
+                return GetUserDataClassName(*((void**)lua_touserdata(L, index)), L);
```
qaisjp, line 119: Why does this call to `lua_touserdata` need `*((void**)` casts but not the next one?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+
+    // TypeMatch<T> should return true if the value on top of the Lua stack can be popped via
+    // PopUnsafe<T>. This must accurately reflect the associated PopUnsafe. Note that TypeMatch
+    // should only check for obvious type violations (e.g. false is not a string) but not
+    // for internal type errors (passing a vehicle to a function expecting a ped)
+    template <typename T>
+    inline bool TypeMatch(lua_State* L, std::size_t index)
+    {
+        int iArgument = lua_type(L, index);
+        // trivial types
```
qaisjp: General question: `trivial` vs `primitive`?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+                      std::is_same_v<T, unsigned int> || std::is_same_v<T, unsigned short>)
+            return (iArgument == LUA_TSTRING || iArgument == LUA_TNUMBER);
+        if constexpr (std::is_same_v<T, bool>)
+            return (iArgument == LUA_TBOOLEAN);
+
+        // Advanced types
+        // Enums are represented as strings to Lua
+        if constexpr (std::is_enum_v<T>)
+            return iArgument == LUA_TSTRING;
+        // std::optional is used for optional parameters
```
qaisjp, line 190:
```suggestion

        // std::optional is used for optional parameters
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+
+        // Catch all for class pointer types, assume all classes are valid script entities
+        // and can be fetched from a userdata
+        if constexpr (std::is_pointer_v<T> && std::is_class_v<std::remove_pointer_t<T>>)
+            return iArgument == LUA_TUSERDATA || iArgument == LUA_TLIGHTUSERDATA;
+
+        // dummy type is used as overload extension if one overload has fewer arguments
+        // thus it is only allowed if there are no further args on the Lua side
+        if constexpr (std::is_same_v<T, dummy_type>)
+            return iArgument == LUA_TNONE;
```
qaisjp, line 226: Sweet

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+    }
+
+    template <typename T>
+    inline T PopUnsafe(lua_State* L, std::size_t& index)
+    {
+        // Expect no change in stack size
+        LUA_STACK_EXPECT(0);
+        // the dummy type is not read from Lua
+        if constexpr (std::is_same_v<T, dummy_type>)
+            return dummy_type{};
```
qaisjp, line 260: Not really related here, but what happens if you somehow _push_ an instance of `dummy_type`?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,438 @@
...
+    inline int operator()(lua_State* L, CScriptDebugging* pScriptDebugging)
+    {
+        int iResult = 0;
+        try
+        {
+            iResult = Call(L);
+        }
+        catch (std::invalid_argument& e)
+        {
+            strError = e.what();
```
qaisjp, line 429: What happens if `strError` already has an error message? Is that even possible? If so, can we somehow emit both error messages (if useful)?

qaisjp (changes requested):
Looks good.

Almost all of these suggestions are readability/documentation related - I've gone with the approach that if I am having trouble skim-reading something, it should probably be explained in a comment. I hope that's OK.

I've also pointed out some methods that would benefit from guarding (or "short-circuiting"... I'm not really sure what it's called). I think it reduces cognitive load. Let me know what you think about this style.

Also, note that my suggestions to `Client/.../CLuaDefs.h` also apply to `Server/.../CLuaDefs.h`.

Still to review:
- I have **not** (yet) checked to make sure you exhaust all types and include the full range of types in all areas of `TypeMatch`/`Push`/`Pop`.
- Deeper read into `CLuaFunctionParser::PopUnsafe`, `CLuaFunctionParser::Call`, and `CLuaOverloadParser`.

Optional:
- Include a short guide somewhere that tells you what you need to edit when adding a new type. This is probably unnecessary as you cover types in the most general case.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: You can't do that, as we do not have a `Push` overload for `dummy_type`

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: The invalid argument exception can only be thrown from the called function (e.g. invalid hash type parameter for `passwordHash`), thus at that point `strError` is empty, as all arguments presented were valid

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: If none of those are met, you're greeted by a compile error, as you're not returning anything in a non-void function

sbx320: With your suggested change, that'd not trigger until runtime

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Yes

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
qaisjp: Consider presenting that tautology in the form of a code comment:
```suggestion
            // This exception can only be thrown from the called function
            // (e.g. invalid hash type parameter for passwordHash)
            // therefore, as all arguments presented were valid, we can
            // guarantee `strError` is currently unset and is safe to overwrite 
            strError = e.what();
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Should be `std::invalid_argument`. Initially I had a custom exception class, but that got scrapped when I moved from exceptions to reference based errors

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
qaisjp: > you're greeted by a compile error

neato

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
sbx320: >This is a little complicated to read. Consider splitting across multiple lines.

I tried that, VS decided to format it away. :( 

>Consider guarding off (dropping the else) to highlight the base case

Dropping the else is not possible there, as it's a `if constexpr` block. For example 
```
template<typename T>
T foo() {
    if constexpr(std::is_same_v<T, void>) {
         bar();
         return;
    }
    return bar();
}
```
fails to compile, as the `T = void` case compiles down to
```
bar();
return;
return bar();
```

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
qaisjp: Is this actually necessary, or is the compiler smart enough to `constexpr` this method itself?

(it's obviously better to be explicit anyway, so later modifications don't mess it up.)

Shared/sdk/SharedUtil.Template.h
sbx320: This is a case of applied laziness. I just needed Key and Value from `std::unordered_map` (which has 5 parameters)

I'll add the other types.

Shared/sdk/SharedUtil.Template.h
sbx320: Yeah, it's somewhat of a base case. Basically it is used to figure out whether more Arguments must be added in the next step.

Shared/mods/deathmatch/logic/lua/LuaBasic.h
sbx320: That's the glorious `std::visit`. Basically what `std::visit(func, variant)` does, is calling func with whatever is currently inside the variant. So if you have a `std::variant<float, int>` which currently contains a `float`, `func` is called with that float. In a way it is "unpacking" the variant into whatever it currently contains.

The lambda simply forwards the contained value to the responding Push overload

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: That's the code difference between OOP and non-OOP. For OOP we use full userdatas (to allow metatables on individual elements). For non OOP we use lightuserdatas, which are only a single pointer on the Lua side.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: primitive is better

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Actually on second glance I don't think it's possible. SString (and std::string) are not constexpr.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Not in a clean manner. As `lua_State*` is used as a generic parameter, we must - in theory - be able to construct an error message if `TypeMatch` fails (which it doesn't, as it's meant to always return true).

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Just noticed that comment is outdated anyway, Pop doesn't throw

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
qaisjp: :(

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Partially applied. You can now only take lua_State* at the beginning of a function. However you are - in theory - still allowed to do `void foo(lua_State*, lua_State*)`

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Fixed (additional template parameter for ArgumentParserWarn added). You can now either do 

- `ArgumentParserWarn<nullptr, func>` to return nil on failure
- `ArgumentParserWarn<false, func>` to return false on failure

I don't think we need more, do we?

sbx320:
Thanks for the review. I applied some changes to resolve most of it.

>I have not (yet) checked to make sure you exhaust all types and include the full range of types in all areas of TypeMatch/Push/Pop.

I most likely do not cover all of them yet. Adding further types is fairly simple though, so I'd suggest we just add them as we go. 

>Deeper read into CLuaFunctionParser::PopUnsafe, CLuaFunctionParser::Call, and CLuaOverloadParser.

If you have questions, ask me :)

>Include a short guide somewhere that tells you what you need to edit when adding a new type. This is probably unnecessary as you cover types in the most general case.

Adding new types is rare, as we usually only introduce new script entities which are all handled already. But as a quick reference for the future:

- New arguments would need to be added in `TypeMatch` (check if type is valid here), `TypeToName` (for error messages) and `PopUnsafe` (read type from Lua stack). 
- New return types would need to be added as a Push overload. 

Details of these operations depend heavily on the type that is to be added.

Shared/mods/deathmatch/logic/lua/CLuaOverloadParser.h
```diff
@@ -0,0 +1,90 @@
...
+    template <std::size_t N, typename... T>
+    static ChosenFunction MakeChoice(std::variant<T...> var)
+    {
+        return std::visit(
+            [](auto&& f) {
+                using ft = decltype(f);
+                if constexpr (!std::is_convertible_v<ft, nth_element_t<N, Args...>>)
+                    return ChosenFunction::FUNCB;            // if it cannot match A, B it is
+                if constexpr (!std::is_convertible_v<ft, nth_element_t<N, Args2...>>)
+                    return ChosenFunction::FUNCA;            // if it cannot match A, B it is
```
qaisjp:
```suggestion
                    return ChosenFunction::FUNCA;            // if it cannot match B, A it is
```
?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -0,0 +1,443 @@
...
+                return 0;
+            }
+            else
+            {
+                return lua::Push(L, Func(std::forward<Params>(ps)...));
+            }
+        }
+        else
+        {
+            return Call(L, ps..., Pop<typename nth_element_impl<sizeof...(Params), Args...>::type>(L, iIndex));
```
qaisjp, line 414: What is this for?

qaisjp (approved):

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: Call recursively calls itself until it has reached the desired number of arguments. In each step we pop the next parameter from the Lua stack and append it to the list of arguments.

qaisjp: And then Push the return value?

sbx320: Yup. So if we're trying to call `int foo(std::string, float)` we get the following calls:
`Call(L);`
`Call(L, str);` (with str being the string returned from Pop)
`Call(L, str, flt)` (flt being the float from Pop)
`Push(Func(str, flt));`

StrixG (approved):

qaisjp:
What's blocking this?

ghost:
> 
> 
> What's blocking this?

Some DX rendering stuff gets broken, if you compile MTA with VS 2019. Other than that, this PR looks complete.

sbx320:
>What's blocking this?

VS2019 on the build server is the big thing. 

But if @saml1er knows about some rendering issues we should investigate those first before upgrading. Probably some new optimization breaking some wrong code. Could you file an issue for those @saml1er so we can track/investigate them?

qaisjp:
> Some DX rendering stuff gets broken

I think @botder fixed one of them in 53121a3b3084b7e503d57be122f8649dd0bb09a8. I don't know what the others are

Dutchman101:
Now that VS2019 is on the build server, and no one created new issues for any DX issues (nor did I observe any bugs in recent nightlies as a result of using vs2019), can we get a fresh look on what's required to proceed with this?

Edit: wait for #1135 results

sbx320:
I'd want to get proper handling for `std::string_view` in (should be significantly faster than using `std::string` as it avoids copying the Lua string) as well as some useful tests for the crypt defs before merging it. 

But no real blockers other than that anymore.

qaisjp:
clang complains

```cpp
/mtasa/blue/Build/../Shared/mods/deathmatch/logic/lua/LuaBasic.h:25:14: warning: inline function 'lua::PopPrimitive<std::__1::basic_string<char> >' is not defined
      [-Wundefined-inline]
    inline T PopPrimitive(lua_State* L, std::size_t& index);
             ^
../Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h:263:25: note: used here
            return lua::PopPrimitive<T>(L, index);
                        ^
```

ghost:
>
> clang complains

Any other issue? I want to merge this.

ghost:
> 
> 
> I'd want to get proper handling for `std::string_view` in (should be significantly faster than using `std::string` as it avoids copying the Lua string) as well as some useful tests for the crypt defs before merging it.
> 

After reading more about std::string_view, looks like we can easily replace `const string &` with `std::string_view`. Should I do it for this PR? @sbx320

Shared/mods/deathmatch/logic/lua/CLuaOverloadParser.h
```diff
@@ -0,0 +1,90 @@
...
+            return Func(std::get<Args>(args)...);
+        else if (choice == ChosenFunction::FUNCB)
+            return Func2(std::get<Args2>(args)...);
+        else if (choice == ChosenFunction::FAIL)
+        {
+            // User provided incorrect arguments
+            throw std::invalid_argument("Overload resolution failed.");
+        }
+        // If this exception is thrown, there is a bug in the overload, as they are not unique
+        throw std::invalid_argument("MTA:SA Bug. Please file an issue with the parameters used for calling.");
```
qaisjp:
```suggestion
        throw std::invalid_argument("MTA:SA Bug. Please file an issue at https://git.io/blue-issues with the parameters used for calling.");
```

Citizen01:
🚀

PR #999, opened 2019-06-17
Disable forboden programs checks in debug mode

CrosRoad95:
The problem:
Every time i use cheat engine in debug mode, and i restart mta i must save everything, close cheat engine, run mta and open cheat engine again. Simply waste of time.

image

### DIFF
--- a/Client/loader/Main.cpp
+++ b/Client/loader/Main.cpp
@@ -78,7 +78,7 @@ MTAEXPORT int DoWinMain(HINSTANCE hLauncherInstance, HINSTANCE hPrevInstance, LP
 
     // Stuff
     HandleCustomStartMessage();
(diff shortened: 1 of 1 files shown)


Client/loader/Main.cpp
```diff
@@ -78,7 +78,9 @@ MTAEXPORT int DoWinMain(HINSTANCE hLauncherInstance, HINSTANCE hPrevInstance, LP
 
     // Stuff
     HandleCustomStartMessage();
-    ForbodenProgramsMessage();
+    #if !defined(MTA_DEBUG)
```
qaisjp:
```suggestion
    #ifndef MTA_DEBUG
```

I think `ifdef`/`ifndef` is preferred (where possible)

qaisjp (approved):

PR #1020, opened 2019-07-04
Implement messageType parameter to onClientChatMessage

### LINKED ISSUE #1014: Add messageType parameter to onClientChatMessage
**Is your feature request related to a problem? Please describe.** Please add the `messageType` parameter in the `onClientChatMessage` event.

**Additional context** I would like this to get added because it will be usefull f.e if someone wants to cancel a team chat message or a /me message from showing up, instead of making a string check which is not safe nor reliable.

Lpsd:
Resolves #1014 

This PR adds the `messageType` parameter to the `onClientChatMessage` event. The possible types are as follows:

```
0 = player
1 = action
2 = team
3 = private
4 = internal
```

0, 1 and 2 follow the same structure as the `onPlayerChat` event.

`private` type is when using the `/msg` command
`internal` is for any messages sent internally by MTA (such as "login: You successfully logged in")

I have included a small test script below (clientside)

```lua
local messageTypes = {
	{id = 0, type = "player"},
	{id = 1, type = "action"},
	{id = 2, type = "team"},
	{id = 3, type = "private"},
	{id = 4, type = "internal"}
}

function getMessageTypeName(messageType)
	for i,v in ipairs(messageTypes) do
		if (tonumber(messageType) == v.id) then
			return v.type
		end
	end
	return false
end

function handleChatMessage(text, r, g, b, messageType)
	print(text, "messageType: "..messageType, getMessageTypeName(messageType))
end
addEventHandler("onClientChatMessage", root, handleChatMessage)
```

Also fixed a couple of typos already present in the code.


CrosRoad95:
as someone said some time ago: enum should be returned as string, not a number

Lpsd:
@CrosRoad95 I have kept this consistent with `onPlayerChat`, which returns an integer

qaisjp:
> as someone said some time ago: enum should be returned as string, not a number

Agreed. I'd also suggest renaming to the following:

```
0 = player
4 = internal
```

moon91210:
I need this for a gamemode I'm working on. Could someone review this PR?

qaisjp (changes requested):
bitstream needs bump

Pirulax:
Soo? @qaisjp

patrikjuvonen (approved):
Looks good, tested working. Thanks @Lpsd and @StrixG.

PR #1021, opened 2019-07-05
Add reference counter increment/decrement on vehicle dummies changes

### LINKED ISSUE #1009: Crash in CModelInfoSA::ResetAllVehicleDummies
**Describe the bug**
There is a suspicious crash (which looks like a regression/recently introduced bug rather than a memory issue) that occurs in **CModelInfoSA::ResetAllVehicleDummies**. It appears NULL is being passed. 

I believe that PR #390 may have caused it, because there's been a myriad of dummy-related crashes ever since which were never seen in dummy functions before. In this case, suspicion arises also because the crash line is one that @forkerer changed in his diff.

Crash trace: [https://pastebin.com/Nf3pQ6cU](https://pastebin.com/Nf3pQ6cU)

Crash line i referred: [https://github.com/multitheftauto/mtasa-blue/blob/master/Client/game_sa/CModelInfoSA.cpp#L976](https://github.com/multitheftauto/mtasa-blue/blob/master/Client/game_sa/CModelInfoSA.cpp#L976)

The crash is in game_sa.dll and the offsets shift around 0035xxx.
Examples: 
r16588 = 00035641 (crash popularity into the thousands)
r18676 = 00035751

The nature of said crashing line `(pVehicleModel->pVisualInfo->vecDummies[dummy.first] = dummy.second;` and having a nullpointer passed there made me want to dig a little deeper. It may aswell be a bad model case, someone who knows what he's doing can find out more, but upon first asking about this crash in #development I got asked to open an issue immediately.

I certainly think it's worth further investigation.

**To reproduce**
Unknown

**Version**
Crash still exists in r18676

forkerer:
Should fix #1009 as per [this ](https://github.com/multitheftauto/mtasa-blue/issues/1009#issuecomment-506022817)Saml1er comment, the issue must have been present before the dummies altering ability was added, as the old system didn't change the reference counter either. Big thanks to @Saml1er for suggesting what could be the cause, as the crash has no easily reproducible reason, still can't be 100% sure if it's this, but can't think of anything else that could be causing it.

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -955,6 +955,8 @@ void CModelInfoSA::SetVehicleDummyPosition(eVehicleDummies eDummy, const CVector
     if (iter == ms_ModelDefaultDummiesPosition.end())
     {
         ms_ModelDefaultDummiesPosition.insert({pVehicleModel, std::map<eVehicleDummies, CVector>()});
+        // Increment this model references count, so we don't unload it before we have a chance to reset the positions
+        m_pInterface->usNumberOfRefs++;
     }
 
     if (ms_ModelDefaultDummiesPosition[pVehicleModel].find(eDummy) == ms_ModelDefaultDummiesPosition[pVehicleModel].end())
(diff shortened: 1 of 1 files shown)


forkerer:
Also #963 tries to fix the same issue, but different way

PR #1023, opened 2019-07-09
Remove ref & deref functions

ghost:
This PR removes the `ref` and `deref` lua functions.

I doubt that the function was ever useful and i also think that it's no big deal in removing it entirely as they were never documented on the wiki.

Luxy.c

### DIFF
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
@@ -15,8 +15,6 @@ void CLuaUtilDefs::LoadFunctions()
     std::map<const char*, lua_CFunction> functions{
         // Util functions to make scripting easier for the end user
         // Some of these are based on standard mIRC script funcs as a lot of people will be used to them
-        {"deref", Dereference},
-        {"ref", Reference},
         {"getTickCount", GetTickCount_},
         {"getRealTime", GetCTime},
         {"split", Split},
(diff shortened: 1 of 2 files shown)


Pirulax:
I agree.

botder:
I would prefer a deprecation warning and then substitute the functions with no-op functions.

ghost:
> I would prefer a deprecation warning and then substitute the functions with no-op functions.

~~i thought the same but then i've seen that lua_ref is throwing an error anyway:~~
https://github.com/multitheftauto/mtasa-blue/blob/master/vendor/lua/src/lauxlib.h#L162

qaisjp:
My thought was "But why?" but reading over this, there's probably no reason for it to be exposed.

No default resource uses `ref` or `deref`. I will take a look at community resources.

> i also think that it's no big deal in removing it entirely as they were never documented on the wiki

I like this approach of backwards compatibility — if it's not documented, or if the documentation does not promise certain behaviour, we should be more willing to "break compatibility".

> I would prefer a deprecation warning and then substitute the functions with no-op functions.

Agreed.

TheNormalnij:
Can be used for some hacks between resources, i think.

qaisjp:
> Can be used for some hacks between resources, i think.

Hmm, yes, if you need to refer to a coroutine from another resource you could use `ref` and pass it around (instead of managing your own table).

It seems like a reasonable use case. In that case, I think we should only remove these functions if (future versions of) Lua has them marked as deprecated.

ghost:
Well, sorry for the late reply but how would you for example get that reference back in lua ? I mean you get a number (ref) back from `ref` but how would you actually get for example the table or function holding that ID in LUA ?

Example:

**Resource 1**
```lua
local refID = false

-- exported
function doSomething(func)
	refID = ref(func)
end

function getRef()
	return refID
end
```

**Resource 2**
```lua
	local funcRef = exports.resource1:getRef()
	
	-- how would get the function from refID ?
	
	funcRef()
end)
```

qaisjp:
In your example it would not be possible. The ref can only be deref'd in resource 1.

TheNormalnij:
Interfeace between resources.
[ref_test.zip](https://github.com/multitheftauto/mtasa-blue/files/3638984/ref_test.zip)

Issue #721, opened 2018-12-01
Discord Rich Presence


JQuinchel:
It would be good to know which server my friends are on, and to be able to enter through discord, and know the played time.

botder:
I tried implementing Discord Rich Presence, but Discord would still show the presence after quitting the game sometimes. Please use **Feature request** issue template next time.

Wolfee-J:
You gotta send a signal that tells discord the game has ended when the client disconnects / game closes.

kereis:
Is there still something going on or is there a possibility to contribute to solving this issue?

Fernando-A-Rocha:
PLEASE implement Discord Rich Presence for MTA! It would be amazing. Definitely worth coding.

kereis:
> PLEASE implement Discord Rich Presence for MTA! It would be amazing. Definitely worth coding.

Well, in case there's not much movement and as it hasn't high priority, I'd like to take a look at it. Can't say any ETA right now.

qaisjp:
Go for it! Might be worth expanding on or taking a look at https://github.com/botder/mtasa-blue/tree/feature/discord-richpresence

It's also worth sharing your ideas here before implementing them.

kereis:
> Go for it! Might be worth expanding on or taking a look at https://github.com/botder/mtasa-blue/tree/feature/discord-richpresence
> 
> It's also worth sharing your ideas here before implementing them.

Sure. Gotta take a look at the repo you've provided. Sorry in advance for my lacking experience in c++. Just in case you can take a look at it and suggest any improvements.

kereis:
> Go for it! Might be worth expanding on or taking a look at https://github.com/botder/mtasa-blue/tree/feature/discord-richpresence
> 
> It's also worth sharing your ideas here before implementing them.

@qaisjp 
Can I simply merge their branch to my fork?

qaisjp:
> Can I simply merge their branch to my fork?

@dopauthor: not exactly merging:

```bash
# Pull botder's branch into your own local repository
git fetch https://github.com/botder/mtasa-blue feature/discord-richpresence:feature/discord-richpresence

# Checkout the feature branch
git checkout feature/discord-richpresence

# Push that up to your own
git push
```

For help with pushing to your own fork (the final step), please read [this page](https://git-scm.com/book/en/v2/Git-Basics-Working-with-Remotes).

kereis:
Do I need any additional extension in order to solve these CMake errors?

[...]

PR #1031, opened 2019-07-24
Add initial Discord Game SDK Integration

0x416c69:
As discussed in #721 I've implemented it.

There is a new library added to the mta path, called discord_game_sdk.dll
I think there should be client verification for this, I'm not sure.

Also it's best to request Discord for the official MTA:SA game so we can use spectate feature too.

My friend @Deihim007 suggested this in the first place.

Ali Gerami.


Client/core/CDiscordManager.cpp
```diff
@@ -0,0 +1,345 @@
...
+    {
+        SString      ipaddr;
+        unsigned int port;
+        CVARS_GET("host", ipaddr);
+        CVARS_GET("port", port);
+
+        std::lock_guard<std::mutex> guardian(m_threadSafety);
+        m_storedActivity.SetDetails("Retrieving server name...");
+        m_storedActivity.GetAssets().SetSmallText(SString("Connected to %s:%i", *ipaddr, port));
+        m_storedActivity.GetAssets().SetSmallImage("a-server"); // TODO: Maybe contact with MTA:SA servers and check if this ip is a premium one containing a small image to set using this function
```
0x416c69: Maybe from https://mtasa.com/toplist/ ?

Discord have a limit of 150 images that you can upload.

Client/core/premake5.lua
```diff
@@ -18,11 +18,13 @@ project "Client Core"
 			"../../vendor/jpeg-9b",
 			"../../vendor/pthreads/include",
 			"../../vendor/sparsehash/src/",
-			"../../vendor/hwbrk"
+			"../../vendor/hwbrk",
+            "../../vendor/discordgsdk/cpp"
```
patrikjuvonen: Spaces and tabs mixed here (also check the rest of the file).

Client/core/CCore.cpp
```diff
@@ -124,12 +124,19 @@ CCore::CCore()
 
     // Create tray icon
     m_pTrayIcon = new CTrayIcon();
+
+    // Initialize discord manager
+    m_pDiscordManager = new CDiscordManager();
```
jushar: Please use a `std::unique_ptr` for new things.

Client/core/CCore.h
```diff
@@ -40,6 +40,7 @@ class CCore;
 #include <ijsify.h>
 #include <core/CWebCoreInterface.h>
 #include "CTrayIcon.h"
+#include "CDiscordManager.h"
```
jushar: To improve compile speed, please use a forward declaration for this and include `CDiscordManager.h` in `CCore.cpp`

Client/core/CDiscordManager.cpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+void CDiscordManager::OnActivityJoin(const char* joinSecret)
+{
+    if (!joinSecret)
+        return;
+
+    // Parse it
+    SString ipport, secret;
+    ushort  port;
```
jushar, line 95: Better declare and define these variables in a single statement (to reduce the scope).

Client/core/CDiscordManager.h
```diff
@@ -0,0 +1,69 @@
...
+
+    bool NeedsSuicide() const { return m_bSuicide; }
+    void SetDead() { m_bSuicide = false; }
+    void DisconnectNotification();
+
+private:
+
+    void Restore();
+
+    discord::Core*    m_discordCore;
```
jushar: Please use UpperCamcelCase for member variable names.

Client/core/CQueryReceiver.cpp
```diff
@@ -0,0 +1,235 @@
+/*****************************************************************************
```
jushar, line 1: Is it a lot of effort for you to do this `QueryReceiver` refactor in a separate pull request?

Client/core/StdInc.h
```diff
@@ -69,3 +70,4 @@
 #include "CNickGen.h"
 #include <core/CWebViewInterface.h>
 #include "CTrayIcon.h"
+#include "CDiscordManager.h"
```
jushar: There's no need to have this in `StdInc.h` as it's only used in a few files.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1343,6 +1353,15 @@ void CGame::InitialDataStream(CPlayer& Player)
 
     marker.Set("onPlayerJoin");
 
+    SString joinSecret = Player.GetDiscordJoinSecret();
+    if (joinSecret.length())
+    {
+        CLuaArguments Arguments;
+        Arguments.PushBoolean(true);
+        Arguments.PushString(joinSecret);
+        Player.CallEvent("onPlayerDiscordJoin", Arguments);
```
jushar, line 1362: I'd maybe generalize this a bit to something like `onPlayerMessengerJoin` to be not too tightly coupled to Discord.

The signature may be something like: `string messengerType, bool justConnected, string secret`.

Shared/sdk/SharedUtil.Thread.h
```diff
@@ -11,6 +11,8 @@
 #include <sys/timeb.h>
 #include <pthread.h>
 
+#pragma once
```
jushar: This should be at the beginning of the file.

jushar (commented):
Thank you very much for the pull request! Overall, it looks very good.

I commented on a few minor issues, but haven't fully reviewed it yet.

Server/mods/deathmatch/logic/CGame.cpp
qaisjp: Perhaps the third argument could be a table that can change depending on the "messenger"? Is "messenger" correct terminology here? Not sure.

Client/core/CCore.cpp
sbx320: In this case `CDiscordManager` should just be a member of `CCore`. There is no need to use a pointer here at all.

Client/core/CCore.cpp
```diff
@@ -1957,6 +1968,24 @@ uint CCore::GetMaxStreamingMemory()
     return m_fMaxStreamingMemory;
 }
 
+//
+// ResetDiscordRichPresence
+//
+void CCore::ResetDiscordRichPresence()
+{
+    if (m_pDiscordManager)
```
sbx320: The else branch can't happen (unless you delete `CCore` and then call this funciton, which is going to blow up anyway).

Client/core/CDiscordManager.cpp
```diff
@@ -0,0 +1,345 @@
...
+    Reconnect(true); // Try to interact with discord on construction
+    m_pThread = new CThreadHandle(CDiscordManager::DiscordThread, this);
+
+    m_storedActivity.GetAssets().SetLargeImage("multi-theft-auto"); // Always thing
+    m_storedActivity.GetAssets().SetLargeText("Playing MTA:SA");
+}
+
+CDiscordManager::~CDiscordManager()
+{
+    m_bSuicide = true;
```
sbx320: This look very hacky. Not sure how to do this more nicely though

Client/core/CDiscordManager.cpp
```diff
@@ -0,0 +1,345 @@
...
+}
+
+// Establish connection with discord
+void CDiscordManager::Reconnect(bool bOnInitialization)
+{
+    bool discordRichPresence;
+    CVARS_GET("discord_rich_presence", discordRichPresence);
+    if (!discordRichPresence) return; // Disabled
+
+    m_threadSafety.lock();
```
sbx320: Just use a lock guard here. `lock()` and `unlock()` are very easy to miss and could be moved around accidentally.

Client/core/ServerBrowser/CServerList.cpp
```diff
@@ -371,29 +371,20 @@ std::string CServerListItem::Pulse(bool bCanSendQuery, bool bRemoveNonResponding
     }
     else
     {
-        // Poll the socket
-        sockaddr_in clntAddr;
-        int         addrLen = sizeof(clntAddr);
-        int         len = recvfrom(m_Socket, szBuffer, SERVER_LIST_QUERY_BUFFER, MSG_PARTIAL, (sockaddr*)&clntAddr, &addrLen);
-        int         error = WSAGetLastError();
-        if (len >= 0)
+        // Poll the socket and parse the packet
```
sbx320, line 374: Why are you changing the server list for Discord Rich Presence here? I'd prefer this to be in a separate pull request tbh.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -303,6 +303,9 @@ CClientGame::CClientGame(bool bLocalPlay)
     // Start async task scheduler
     m_pAsyncTaskScheduler = new SharedUtil::CAsyncTaskScheduler(2);
 
+    // Server info construction
+    m_pServerInfo = new CServerInfo();
+
```
sbx320: Also here: Just make it a regular member instead of a pointer. Adding another level of indirection doesn't gain us anything.

sbx320 (changes requested):

Client/core/CCore.cpp
0x416c69: Alright.

Client/core/CDiscordManager.cpp
0x416c69: The only other way is signals I guess

Client/core/CDiscordManager.cpp
0x416c69: I can't unless I move that line to another scope, it will result in dead lock.

Client/core/ServerBrowser/CServerList.cpp
0x416c69: I can make it in another PR but this is necessary for Rich Presence, server won't share its name with clients and I use query to fetch the name, which was a mess and duplicating the code is painful and messy.

Client/core/CQueryReceiver.cpp
0x416c69: That way this PR cannot be compiled unless I copy the old code from server list for query in CDiscordManager.cpp which is painful and messy.

Server/mods/deathmatch/logic/CGame.cpp
0x416c69: onPlayerExternalJoin maybe?

Client/core/CCore.cpp
ccw808: I think a pointer is better here because it avoids dependency on CDiscordManager.h

qaisjp: @ccw808, would that still be the case with the fwd decl mentioned [here](https://github.com/multitheftauto/mtasa-blue/pull/1031#discussion_r307152804)?

ccw808: Yes. The fwd decl mentioned [there](https://github.com/multitheftauto/mtasa-blue/pull/1031#discussion_r307152804) requires a pointer

0x416c69: So class CDiscordManager; and no more including the header in CCore.h

But I still need to go with smart pointers right?

Shared/mods/deathmatch/logic/Enums.cpp
```diff
@@ -161,4 +161,8 @@ ADD_ENUM1(PACKET_ID_WEAPON_BULLETSYNC)
 ADD_ENUM1(PACKET_ID_PED_TASK)
 ADD_ENUM1(PACKET_ID_PLAYER_NO_SOCKET)
 ADD_ENUM1(PACKET_ID_PLAYER_NETWORK_STATUS)
+ADD_ENUM1(PACKET_ID_PLAYER_ACINFO)
```
0x416c69, line 164: These packet IDs weren't in enum info struct, I was not sure when I was adding the newly added missing ones including mine.

Client/core/CDiscordManager.cpp
jushar: You could create a new scope.

E.g.:
```cpp
// ...
if (!discordRichPresence) return; // Disabled

{
  std::lock_guard<std::mutex>{m_threadSafety};
  discord::Result res = discord::Core::Create(DISCORD_CLIENT_ID, DiscordCreateFlags_NoRequireDiscord, &m_discordCore);
}
// ...
```

0x416c69: So another scope it is, alright.

0x416c69:
So as you guys stopped reviewing this, I'll adjust the necessary modifications to the PR.

Edit: There you have it.

qaisjp:
Seems to work well, with a few minor catches!

Doesn't seem to compile on Windows without modification (i.e. copying a lib file from https://dl-game-sdk.discordapp.net/latest/discord_game_sdk.zip into `vendor/discordgsdk/lib`).

Please can you fix this?

**Things that worked great**

- activity message when loading game / in menu
- activity message when joining a server
- activity message when ingame
- activity message disappears when you disable rich presence whilst ingame
- activity message disappears when you disable rich presence whilst in menu

**Something odd**

The discord activity message does not reappear after you enable rich presence, it only appears when you join a server. (Does not say "In Main Menu" when you disconnect from a server either!)

0x416c69:
> The discord activity message does not reappear after you enable rich presence, it only appears when you join a server.

Well, it will be enabled in 15 seconds or less, it will use the "reconnecting" feature for reappearing which will be done in every 15 sec. (Thought it's better to prevent user spamming enable/disable)

> (Does not say "In Main Menu" when you disconnect from a server either!)

Are you sure? It works for me.

> Please can you fix this?

Why not.

Also things that you may have missed:

- Closing and opening discord after the game is opened
- Backward compatibility check

qaisjp:
> Well, it will be enabled in 15 seconds or less, it will use the "reconnecting" feature for reappearing which will be done in every 15 sec. (Thought it's better to prevent user spamming enable/disable)
> 
> >

Okay!

> (Does not say "In Main Menu" when you disconnect from a server either!)

I meant after you turned the feature off and on again. So this doesn't matter anymore.

> Also things that you may have missed:
> 
> - Closing and opening discord after the game is opened
> - Backward compatibility check

Yep, can confirm the first one works. Haven't tried backwards compat.

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -987,6 +995,11 @@ void CPacketHandler::Packet_PlayerList(NetBitStreamInterface& bitStream)
             // Call the onClientPlayerJoin event
             CLuaArguments Arguments;
             pPlayer->CallEvent("onClientPlayerJoin", Arguments, true);
+
+            if (g_pCore->GetNetwork()->GetServerBitStreamVersion() >= 0x06D)
+                g_pCore->GetDiscordManager()->SetState(SString("%i/%i Players", g_pClientGame->GetPlayerManager()->Count(), g_pClientGame->GetServerInfo()->GetMaxPlayers()), [](EDiscordRes) {});
+            else
+                g_pCore->GetDiscordManager()->SetState(SString("%i Players", g_pClientGame->GetPlayerManager()->Count()), [](EDiscordRes) {});
```
qaisjp: This (and other places) needs to say `1 Player` if there is only 1 player. 

Are you able to do something about the code duplication?

0x416c69: Resolved

0x416c69:
May I know which version is this PR targeted for?

qaisjp:
@0x416c69 i want to merge this ~~tomorrow~~ asap

p.s. i tested backwards compat and it works 👍

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -452,6 +457,9 @@ void CPacketHandler::Packet_ServerJoined(NetBitStreamInterface& bitStream)
     g_pClientGame->m_pLocalPlayer->CallEvent("onClientPlayerJoin", Arguments, true);
 
     g_pCore->UpdateRecentlyPlayed();
+
+    if (g_pCore->GetNetwork()->GetServerBitStreamVersion() < 0x06D)
+        g_pCore->GetDiscordManager()->SetState(SString("%i Players", ucNumberOfPlayers), [](EDiscordRes) {});
```
qaisjp: This needs to be updated too (I meant the duplication of SetState between here, [`Packet_ServerDisconnected`](https://github.com/multitheftauto/mtasa-blue/pull/1031/files#diff-3ed6112654809f7609a791c26259f645R1004) and [`CServerInfo`](https://github.com/multitheftauto/mtasa-blue/pull/1031/files#diff-be4af7a17233d412f46d4e3cecb371feR22))

0x416c69:
Latest commit needs a retest.

I don't remember when Packet_PlayerList packet arrives (and I'm a little busy to check)

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -7100,3 +7100,16 @@ void CClientGame::VehicleWeaponHitHandler(SVehicleWeaponHitEvent& event)
...
+void CClientGame::UpdateDiscordState()
+{
+    // Set discord state to players[/slot] count
+    uint playerCount = g_pClientGame->GetPlayerManager()->Count();
+    SString state(std::to_string(playerCount));
+
+    if (g_pCore->GetNetwork()->GetServerBitStreamVersion() >= 0x06D)
+        state += "/" + std::to_string(g_pClientGame->GetServerInfo()->GetMaxPlayers());
+
+    state += (playerCount == 1 ? " Player" : " Players");
```
qaisjp: Should still say `1/38 Players` 😄 (only use singular form if we do not know the max number of players)

soz

Deihim007:
> @0x416c69 i want to merge this tomorrow
> 
> p.s. i tested backwards compat and it works 👍

Yaaaay!
p.s: Is it cancelled? 🤦‍♂

qaisjp:
@botder would u be so kind to please merge https://github.com/multitheftauto/mtasa-blue/commit/c292216945cc0e5d041c60052982d3f72be9eac6 into this? pretty please?

qaisjp (dismissed):

Client/mods/deathmatch/CClient.cpp
```diff
@@ -306,3 +297,62 @@ void CClient::GetPlayerNames(std::vector<SString>& vPlayerNames)
...
+void CClient::TriggerDiscordJoin(SString strSecret)
+{
+    g_pClientGame->TriggerDiscordJoin(strSecret);
+}
+
+CClient::InitializeArguments CClient::ExtractInitializeArguments(const char* arguments)
+{
+    // Format: "nickname [password] [secret]"
+    // Examples: "GloriousToaster99 secret", "RandomPainter10", "RandomNickname pw123 abc123def"
+    std::string_view view(arguments);
```
0x416c69: What am I looking at? Is there a problem with SString::Split?
@botder

qaisjp:
The "join game" stuff is interesting. Just tested it now. It works fine if the player clicking the "join" button is on the server already, but doesn't trigger any event if they weren't in the server (in my case, with MTA already open but not connected to any server). But they join the game!!

0x416c69:
So they join the server when they press "Join Game" and no mta event will be triggered but the event will be triggered if they are already on the server?

qaisjp:
> So they join the server when they press "Join Game" and no mta event will be triggered but the event will be triggered if they are already on the server?

that is what currently happens, yes.

- when `justConnected` is expected to be true, no events are sent
- `justConnected = false` events work fine

although i'm tempted to disable the lua functions / events, and merge what we have right now. (and you can try fixing this in a follow up PR)

0x416c69:
This looks like an easy to fix bug, prolly caused by the new argument parser which @botder made. I'll look into it first time at tomorrow morning (gotta sleep now, had a busy day)

0x416c69:
156cdf947223b9f493567e6a2d7470d516454768 should have done the job.

Can you test it for once more?

Client/mods/deathmatch/CClient.cpp
0x416c69: Well, unless you want nickname     [some jumping on space key]           password

0x416c69:
@qaisjp I confirm that the bug you reported is now fixed. (Thanks to @StrixG)

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -1187,6 +1189,38 @@ int CLuaPlayerDefs::SetPlayerBlurLevel(lua_State* luaVM)
...
+        LogWarningIfPlayerHasNotJoinedYet(luaVM, pElement);
+
+        if (CStaticFunctionDefinitions::SetPlayerDiscordJoinParams(pElement, strKey, strPartyId, uiPartySize, uiPartyMax))
+        {
+            lua_pushboolean(luaVM, true);
+            return 1;
+        }
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
```
StrixG: As qaisjp said: 
> New API functions are now expected throw errors instead of print warnings.

```suggestion
        return luaL_error(luaVM, argStream.GetFullErrorMessage());
```

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -1187,6 +1189,38 @@ int CLuaPlayerDefs::SetPlayerBlurLevel(lua_State* luaVM)
...
+
+    lua_pushboolean(luaVM, false);
```
StrixG:
```suggestion
```

qaisjp (approved):
This looks ready to merge.

The Lua API will need changing but I'm happy for us to say something like this on the wiki:

> This is function is experimental and you should first check that the function exists before using it. The function can be removed at any time.
>
> ```lua
> if onPlayerDiscordJoin then
>     onPlayerDiscordJoin(...)
> end
> ```

I'll give it a little bit of time for one of @multitheftauto/blue-collaborators to send a secondary approval, because it's many lines of code.

# If someone merges this, please **squash merge**, a couple of large files have been deleted and shouldn't end up in `master`.

utils/buildactions/install_discord.lua
```diff
@@ -20,19 +20,16 @@ newaction {
 		if os.host() ~= "windows" then return end
 
 		-- Check md5
-		local download_discord = true
 		local archive_path = DISCORD_VENDOR.."/"..DISCORD_FILENAME
 		if os.isfile(archive_path) and os.md5_file(archive_path) == DISCORD_MD5 then
 			print("Discord Game SDK is up to date.")
-			download_discord = false
+			return
```
qaisjp, line 26: It was like this before, but I found that if you delete the other files but _keep_ the `.zip`, it won't re-add the other files.

I think CEF has the same issue, so I don't mind if both do.

qaisjp:
repeating for loudness

# If someone merges this, please squash merge, a couple of large files have been deleted and shouldn't end up in `master`.

qaisjp:
I am merging this in 1 week if nobody else reviews this

ghost:
@qaisjp The PR looks good. Let's go for the merge!

ghost (approved):

PR #1038, opened 2019-08-03
Gridlist setting image for item

CrosRoad95:
Syntax: `bool guiGridListSetItemImage(element gridList, int row, int column, element staticImage)`
usage:
```lua
bl = guiCreateStaticImage( 0,0,0,0, "bl.png", false) -- create some invisible image ( in this case it works as dxCreateTexture )
guiGridListSetItemImage(gridlist, 1, 1, bl) -- assign to item
```

### Example result:
image

### Todo:
1. when image gets destroyed, row keep stretched
2. sometimes this crash appear in `mtasa-blue\vendor\cegui-0.4.0-custom\src\CEGUIImageset.cpp`
image

### Test resource

[gridlist.zip](https://github.com/multitheftauto/mtasa-blue/files/3464173/gridlist.zip)


patrikjuvonen:
Closed as a stale draft.

Issue #992, opened 2019-06-12
guiCreateStaticImage does not output an error if the file does not exist


qaisjp:
**Describe the bug**

guiCreateStaticImage will silently fail if the image does not exist.

**To reproduce**

```lua
crun guiCreateStaticImage(0.05, 0.68, 0.60, 0.003, "asdf123.png", true)
```

**Expected behaviour**

A warning should be printed

**Screenshots**

**Version**

302b83f16c9dfc2936310e4eaa7993c97a4cf2e7

**Additional context**

PR #1041, opened 2019-08-06
Fix of issue #992

CrosRoad95:
It fix issue #992, now guiCreateStaticImage throw warning if image don't exists

Also this will throw error for default admin panel due path `gui\\images\\dot.png` is wrong, correct is `client\\images\\dot.png` 

image

```lua
guiCreateStaticImage(0.35, 0.68, 0.1, 0.1, "consider.png", true)
guiCreateStaticImage(0.35, 0.78, 0.1, 0.1, "invalidImage.png", true) // throw error
```

Full test resource:
[static image.zip](https://github.com/multitheftauto/mtasa-blue/files/3473065/static.image.zip)

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
@@ -795,14 +795,22 @@ int CLuaGUIDefs::GUICreateStaticImage(lua_State* luaVM)
             SString    strPath;
             if (CResourceManager::ParseResourcePathInput(path, pResource, &strPath))
             {
-                CClientGUIElement* pGUIElement = CStaticFunctionDefinitions::GUICreateStaticImage(*pLuaMain, position, size, strPath, relative, parent);
-                if (pGUIElement != NULL)
+                if (FileExists(strPath))
                 {
-                    lua_pushelement(luaVM, pGUIElement);
-                    return 1;
+                    CClientGUIElement* pGUIElement = CStaticFunctionDefinitions::GUICreateStaticImage(*pLuaMain, position, size, strPath, relative, parent);
+                    if (pGUIElement != NULL)
+                    {
+                        lua_pushelement(luaVM, pGUIElement);
+                        return 1;
+                    }
+                    else
+                        argStream.SetCustomError(path, "Failed to create static image");
                 }
+                else 
+                    argStream.SetCustomError(path, "File not found");
             }
-            argStream.SetCustomError(path, "Bad file path");
+            else
+                argStream.SetCustomError(path, "Bad file path");
         }
     }
     if (argStream.HasErrors())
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -795,11 +795,13 @@ int CLuaGUIDefs::GUICreateStaticImage(lua_State* luaVM)
             if (CResourceManager::ParseResourcePathInput(path, pResource, &strPath))
             {
                 CClientGUIElement* pGUIElement = CStaticFunctionDefinitions::GUICreateStaticImage(*pLuaMain, position, size, strPath, relative, parent);
-                lua_pushelement(luaVM, pGUIElement);
-                return 1;
+                if (pGUIElement != NULL)
```
Citizen01: IIRC, we now try to use `nullptr` instead of `NULL`

CrosRoad95: i don't know which variant i should use

qaisjp:
Why didn't this work before?

CrosRoad95:
`m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());`
wasn't called due `return 1`

qaisjp:
But [the first `if`](https://github.com/multitheftauto/mtasa-blue/pull/1041/files#diff-16d0409384a35f36d21666b5ced6b2caL795) shouldn't let it get to that point

CrosRoad95:
it checks if path is valid, not does file exists

qaisjp:
what happens with your pull request if you try to create a staticimage with a file that is empty?

i believe your code currently blames `CStaticFunctionDefinitions::GUICreateStaticImage` returning null for any reason on a bad filepath

you should add an if block to check if the file exists instead of doing a null check. you could even output a generic "failed to create static image" warning when null is returned

CrosRoad95:
now it working as present:
```lua
guiCreateStaticImage(0.35, 0.68, 0.1, 0.1, "c.lua", true) -- Failed to create static image
guiCreateStaticImage(0.35, 0.68, 0.1, 0.1, "consider.png", true) -- no errors, image got created
guiCreateStaticImage(0.35, 0.78, 0.1, 0.1, "doesntExists.png", true) -- File not found
guiCreateStaticImage(0.35, 0.78, 0.1, 0.1, ":asd:asd:asd", true) -- Bad file path
```

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
qaisjp: never use `NULL`, always use `nullptr`

CrosRoad95: done

qaisjp:
thank you

PR #1051, opened 2019-08-12
Upgrade radar image quality

Dezash:
This PR replaces the radar image with a higher quality one, the size of which is 6000x6000. This is the highest resolution version of this image that I could find.
The file size is considerably larger, however the benefit is that you can use the zoom-in feature and still view the map in detail.

Here are some comparisons:

### DIFF
--- a/Client/mods/deathmatch/logic/CRadarMap.cpp
+++ b/Client/mods/deathmatch/logic/CRadarMap.cpp
@@ -56,7 +56,7 @@ CRadarMap::CRadarMap(CClientManager* pManager)
 
     // Create the radar and local player blip images
     m_pRadarImage =
-        g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\radar.jpg"), NULL, false, 1024, 1024, RFORMAT_DXT1);
+        g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\radar.jpg"), NULL, false, 6000, 6000, RFORMAT_DXT1);
     m_pLocalPlayerBlip = g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\radarset\\02.png"));
 
     // Create the marker textures


CrosRoad95:
i don't think replacing streight to significantly higher quality is good idea, it can increase time of loading is lower spec computers or even decrease fps.

For me best solution is add option in settings like "high quality map" by default disabled

Dezash:
@CrosRoad95 I don't think that higher resolution textures would decrease FPS, it would use a little bit more VRAM though. I'm not sure about the loading time, can we measure it somehow?

CrosRoad95:
dxCreateTexture with getTickCount before and after. I had simillar map on my server, and dxCreateTexture took ~2 seconds to load.

Haxardous:
before (If this got pulled), I recommend using **zopflipng** for image optimization, this will increase the performance significantly.
- **Tool source code:** https://github.com/google/zopfli
- **Tutorial:** https://forum.mtasa.com/topic/119032-tool-png-optimizer/

you can also use **UI Up-scaling** to improve actual image quality instead of using an image that went through ton of image shrinking tools. 
- **Checkout this video:** https://www.youtube.com/watch?v=rl1pbEG2kgA&t
- **Waifu2x:** http://waifu2x.udp.jp/

ccw808:
Large textures require a large amount of contiguous memory. e.g. 6000x6000 (as a 8192x8192 DXT1) would require 32MB of RAM + another 32MB block of VRAM when being displayed. Also, to avoid problems with Intel GMA etc. the size of a single texture cannot be larger than 2048x2048

Dezash:
So would having an optional setting to use a 6000x6000 image map be considered or is having anything beyond 2048x2048 out of MTA bounds?

Haxardous:
ba11eb3a-fc2a-4d8e-a9e4-962ceb6e63e3

Original/UI upscaling comparison: https://imgur.com/a/FdUPATk

Dutchman101:
> So would having an optional setting to use a 6000x6000 image map be considered or is having anything beyond 2048x2048 out of MTA bounds?

while I don't think this is in the scope of this PR (would require additional code), MTA favours to provide an ideal set-up.. so when we offer that, you can bet that plenty of users with less-than-ideal PC specs will also (try to) use it, degrading their own experience due to additional performance load.

Anyways, i graphically worked on the radar file and found that these edits are really the best possible result (quality increase/filesize/resolution required for preservation) within the intended boundaries:

1) 2048 x 2048 JPG: [https://imgur.com/a/jr8B9iR](https://imgur.com/a/jr8B9iR)
This is the midway for the quality improvement you're requesting.. we cannot go to 6000 x 6000, but perhaps we can do this. This is already pretty damn HQ, so I think it can as well satisfy you, @Dezash, a zoomed-in example of detail: **[link](https://i.imgur.com/YeIZBth.png)** 

2) 1024 x 1024 JPG: [https://imgur.com/a/x7o0GKo](https://imgur.com/a/x7o0GKo)
The increase in quality (while the filesize decreased) is barely noticable, but that's because any higher quality isn't preservable under this resolution.

Although the current format used for radar is JPG, using PNG just like the rest of image files is better because JPG is slower, the PNG format & library are more optimized; rendering equal images is less heavy on GPU and decoding when using PNG than with JPG. Despite that, filesize of equal PNG (so, affects only total data files size) is a little higher though, but this isn't a priority when we're trying to optimize client performance. So for the aforementioned reasons, if we picked my images to move forward in this PR, then here are PNG versions of my samples:

1) 1024 x 1024 PNG: [https://imgur.com/a/CMELzV5](https://imgur.com/a/CMELzV5)
2) 2048 x 2048 PNG: [https://imgur.com/a/EAfLH27](https://imgur.com/a/EAfLH27)

Also, even though I know, (in a desktop browser only) imgur doesn't touch (or compress) original images as long you use the .png / .jpg source links, here is a .zip file with my resulting files:

[radar_img.zip](https://github.com/multitheftauto/mtasa-blue/files/3494213/radar_img.zip)

Note to @Haxardous: even though AI upscaling is interesting, it's of no use in this scenario because it is for when you don't have a higher quality version and need to work from something low-res; in this case we had the ultra HQ radar image that @Dezash provided to begin with, so we don't need to interpolate something equal with AI upscaling from a lower image.

If you have a HQ image like that to begin with, the only obstacle you'll face (if the person is skilled enough with graphical design) is the bottleneck of what you can preserve on the respective image resolutions, like I demonstrated in my overview (between 1024 and 2048)

Einheit-101:
I like the idea of that 6000x6000 image, and i dont like all these Toaster PCs that are preventing MTA from being what it **COULD** be.

Dutchman101:
> I like the idea of that 6000x6000 image, and i dont like all these Toaster PCs that are preventing MTA from being what it **COULD** be.

Don't forget that MTA is working with an old game from 2005 which isn't designed for heavier stuff (we feel the technical limitations constantly), also with the reality that a significant part of our userbase is on a toaster, we cannot abandon them just because of favouritism/people that feel it's OK because it is OK for them.. we care about all users equally, thinking like that would end up killing MTA

So really, anything like 6000 x 6000 just can't happen.. if you need to, then use a script to replace it to your likings

Dezash:
@Dutchman101 I don't think that having an optional setting is playing favoritism and I never heard of any game that would withhold certain graphical enhancements just because some people with low-end devices might degrade their performance by enabling them.

With that being said, it is important to know exactly what kind of impact does using textures above 2048x2048 have and whether it is specific to user's hardware or the game engine itself.

qaisjp:
> So would having an optional setting to use a 6000x6000 image map be considered or is having anything beyond 2048x2048 out of MTA bounds?

Having a setting would be essential, yes.

I think it's fine if it's a setting. We could maybe add an extra disclaimer for Intel users with certain cards.

Maybe one of these days someone can revamp the video menu so people can see exactly how much each setting impacts gameplay, a bit like GTA V's :-)

qaisjp (changes requested):
This is a good idea, but this feature can't be merged in its current form because of the reasons mentioned here: https://github.com/multitheftauto/mtasa-blue/pull/1051#issuecomment-520516776

I think it's still fine to add this feature as long as it can be turned off. We could maybe add an extra disclaimer for Intel users with certain cards.

It should only be turned on by default if the user's computer can handle the improvement. Just keeping this feature turned off by default is also acceptable.

qaisjp:
I am closing this due to inactivity. Feel free to continue committing / request for this PR to be re-opened when you're ready to continue working on this.

PR #1052, opened 2019-08-15
Add dxDrawWiredSphere function

CrosRoad95:
I mantioned about it long time ago, sometimes you want draw sphere for some kind of purpuse like debugging ( you can notice it on my videos where i drew something simillar to sphere https://www.youtube.com/watch?v=_QV2FkASLas ) but writting script for that isn't trivial, also can cause too high lags but mta already supports doing wire spheres when showcol enabled.

Syntax:
```lua
bool dxDrawWiredSphere( float x, float y, float z, float radius, color theColor, float fLineWidth, uint density )
```
Density must be between 1 and 4 integer.
Negative radius doesn't matter, sphere will draw as normal as positive radius

Line width difference:
image

How different density levels looks like
image

And test resource which i forgot to attach
[sphere.zip](https://github.com/multitheftauto/mtasa-blue/files/3504967/sphere.zip)


CrosRoad95:
i'm still waiting for marge, or reject

ccw808 (approved):

qaisjp:
What is `density`? Is it `segments`?

tut001:
> What is `density`? Is it `segments`?

That's correct.

CrosRoad95:
i renamed it

qaisjp (approved):

PR #1055, opened 2019-08-16
Add element data subscription

tederis:
The intention is to extend the existing element data's synchronization. I suggest to add a new synchronization mode based on the subscription model.

In some situations we are facing a necessity of synchronization between just a couple of machines like server and client. So instead of broadcasting data toward all the players we can manually select only a few of them and save a lot of bandwidth.

It can be done using the following functions:

**setElementData**( element, key, value, "subscribe" )
**subscribeElementData**( element, key, player )

To unsubscribe the data:

**unsubscribeElementData**( element, key, player )

To _compatibility_ purposes I have left the old syntax:

**setElementData**( element, key, value, true or nil ) for broadcasting
**setElementData**( element, key, value, false ) for local use

**setElementData**( element, key, value, nil ) will using a last synchronization option, so there no need to specifying it everytime.

List of functions:
**setElementData**( element theElement, string key, var value, var syncMode=true,false,nil,"broadcast","local","subscribe" )
**subscribeElementData**( element theElement, string key, player thePlayer )
**unsubscribeElementData**( element theElement, string key, player thePlayer )


ArranTuna:
It could take a table as thePlayer for when sending it to a group of players like triggerClientEvent allows as it's much more efficient.

Lpsd:
@ArranTuna This works more efficiently by **not** having to pass a table of players every time. You only have to define them once, or when that group of users receiving the data changes.

ArranTuna:
Which you'd have to do in a for loop, correct? If there were 50 players you wanted subscribed then that would be 50 function calls, versus 1 function call if a table could be passed. Lua function calls are expensive which is why triggerClientEvent allows a table.

Lpsd:
My mistake, I misread your original comment and thought you were suggesting to add a table as input to `setElementData`, not `subscribeElementData`/`unsubscribeElementData`

qaisjp:
This is seemingly really popular, so let's push it through.

My understanding is that `syncType="subscribe"` will cause the data to only be broadcasted to players that have subscribed to that key, correct? Sounds good to me.

> `setElementData( element, key, value, nil )` will using a last synchronization option, so there no need to specifying it everytime.

Sounds interesting.  I think being explicit will make it less confusing when reading code, as the reader will be able to quickly determine if it's broadcast, local or subscription based. That is to say, they won't need to remember the `syncType`.

What's the use case for not wanting to be explicit every time?

tederis:
> My understanding is that `syncType="subscribe"` will cause the data to only be broadcasted to players that have subscribed to that key, correct? Sounds good to me.

Yes, when the parameter is passed data will be broadcasted to subscribed players only.

> Sounds interesting. I think being explicit will make it less confusing when reading code, as the reader will be able to quickly determine if it's broadcast, local or subscription based. That is to say, they won't need to remember the syncType.

I was trying to make the feature simple as much as possible. But I agree this can be confusing. Well, okay, I will remove this ambiguity.

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1637,6 +1650,82 @@ int CLuaElementDefs::removeElementData(lua_State* luaVM)
...
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadUserData(pElement);
+    argStream.ReadString(strKey);
+    argStream.ReadUserData(pPlayer);
+
+    if (!argStream.HasErrors())
+    {
+        CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+        if (pLuaMain)
+        {
```
qaisjp: I don't think you need `pLuaMain` checks here

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1637,6 +1650,82 @@ int CLuaElementDefs::removeElementData(lua_State* luaVM)
...
+            if (IS_PLAYER(pPlayer))
+            {
+                if (CStaticFunctionDefinitions::SubscribeElementData(pElement, strKey, pPlayer))
+                {
+                    lua_pushboolean(luaVM, true);
+                    return 1;
+                }
+            }
+            else
+                argStream.SetTypeError("player", 1);
```
qaisjp: You also don't need IS_PLAYER checks - this is done by CArgStream
```suggestion
            if (CStaticFunctionDefinitions::SubscribeElementData(pElement, strKey, pPlayer))
            {
                lua_pushboolean(luaVM, true);
                return 1;
            }

```

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1637,6 +1650,82 @@ int CLuaElementDefs::removeElementData(lua_State* luaVM)
...
+        }
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
```
qaisjp:
```suggestion
        return luaL_error(luaVM, argStream.GetFullErrorMessage());

```

we now return errors for usage mistakes

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1637,6 +1650,82 @@ int CLuaElementDefs::removeElementData(lua_State* luaVM)
...
+
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
```
qaisjp: same stuff here about: `pLuaMain`, `IS_PLAYER` and throw errors for usage mistakes

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -162,15 +162,12 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
             {
                 const char*         szName = iter->first.c_str();
                 const CLuaArgument* pArgument = &iter->second.Variable;
-                bool                bSynchronized = iter->second.bSynchronized;
+                ESyncType           syncType = iter->second.syncType;
```
qaisjp: you're declaring syncType here but not using it. also if you're changing something here you probably want backwards compatible code https://github.com/multitheftauto/mtasa-blue/wiki/BitStream

Server/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
```diff
@@ -271,6 +271,12 @@ ADD_ENUM(CAccountPassword::SHA256, "sha256")
 ADD_ENUM(CAccountPassword::MD5, "md5")
 IMPLEMENT_ENUM_END("account-password-type")
 
+IMPLEMENT_ENUM_CLASS_BEGIN(ESyncType)
+ADD_ENUM(ESyncType::BROADCAST, "broadcast")
+ADD_ENUM(ESyncType::LOCAL, "local")
+ADD_ENUM(ESyncType::SUBSCRIBE, "subscribe")
+IMPLEMENT_ENUM_CLASS_END("sync-type")
```
qaisjp:
```suggestion
IMPLEMENT_ENUM_CLASS_END("sync-mode")
```
minor nit: we can keep it the way it is in the code, but i think it's clearer if we document this as a "sync mode"

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -309,6 +328,26 @@ bool CPlayerManager::IsValidPlayerModel(unsigned short usPlayerModel)
...
+    {
+        CPlayer* pPlayer = *iter;
+        pPlayer->UnsubscribeElementData(pElement, name);
+    }
+}
+
+void CPlayerManager::ClearElementData(CElement* pElement)
+{
+    list<CPlayer*>::const_iterator iter = m_Players.begin();
+    for (; iter != m_Players.end(); iter++)
```
qaisjp: You probably want to use a C++ range-based for loop here and above

qaisjp (changes requested):
Looks good, but I've highlighted where a few improvements could be made

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
tederis: But `GetVirtualMachine(luaVM)` can return `NULL` in some cases, isn't it? Moreover functions above also have this check.

qaisjp: I actually don't know for sure, but I think that's because CLuaElementDefs has some _really_ old definitions. Plus you're not using `pLuaMain` anywhere anyway

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
tederis: It's just a misprint. The PR is completely backward compatible.

tederis:
I have applied the highlighted improvements and added a new function `isElementSubscribed( element theElement, string key, player thePlayer )`

prnxdev:
Can't it be achieved using observer pattern 🤔?

tederis:
> Can't it be achieved using observer pattern 🤔?

It's what I actually implemented. It's not usual observer pattern as The Gang of Four taught us but it still using this model.

tederis:
Well, it's done with naming conventions

qaisjp (approved):

qaisjp:
Sorry, a couple more questions:

1. what happens to existing subscriptions when you switch an element data key between "local", "broadcast" and "subscribe"?
2. what happens when you add subscribers to an element data key that is currently in "broadcast" or "local" mode?
3. what's the difference between using "broadcast" without adding subscribers, and "local"?

tederis:
> 1. what happens to existing subscriptions when you switch an element data key between "local", "broadcast" and "subscribe"?

When switching from "subscribe" mode to "local"(or "broadcast") all subscriptions are cancelled.

**local** -> **broadcast** - nothing happening
**local**  -> **subscribe** - nothing happening
**broadcast**  -> **local** - nothing happening
**broadcast**  -> **subscribe** - nothing happening
**subscribe**  -> **local** - all subscribtions are cancelled
**subscribe**  -> **broadcast** - all subscribtions are cancelled

> 2. what happens when you add subscribers to an element data key that is currently in "broadcast" or "local" mode?

In case of an attempt to add subscriber to an element data that is currentely in "subscribe" mode the element data will be immediately sended to a new subscriber. When data is in "local"(or "broadcast") mode the attempt to add a new subscriber will be discarded.

> 3. what's the difference between using "broadcast" without adding subscribers, and "local"?

Do you mean "subscribe" mode? When using "subscribe" mode without subscribers all going the same way as "local". I think we can rid off dedicated "subscribe" mode and allow to subscribe element data when "local" mode is selected.

qaisjp:
Thanks for the quick response. Sounds good!

> Do you mean "subscribe" mode?

Yes 🙃 

> I think we can rid off dedicated "subscribe" mode and allow to subscribe element data when "local" mode is selected.

I was thinking that as well, but I'm not 100% sure.

Reasons in favour of keeping "subscribe" (what you have now):
1. **predictable**: it's obvious what happens when read the script ("local" vs "subscribe") - someone won't be confused because their "local" element data is actually being sent to a client
2. I think switching to "local" and back to "subscribe" would be a quick way to remove all subscribers (even though we could add a function in a future PR that sets the subscriber list)

Reasons in favour of removing "subscribe" (potential upcoming change):

3. existing `setElementData` API doesn't change, we only get the new functions



Actually I think reason (1) is pretty important, so I'd say we should keep it as it is. What do you think?

tederis:
I have modified the PR and fixed a logic of switching between modes. There is also an improvement of sending data when new subscriber is appears. I think everything is ready to merge by the now.

tederis:
> Actually I think reason (1) is pretty important, so I'd say we should keep it as it is. What do you think?

Yes, we should keep things predictable. People should use new parameters for new features whereas old sync behaviour is just for backwards compatibility.

Pirulax:
And there it goes, my fix for speeding the set/get functions up is messed up now. At least this is done as well. I can redo it.

Dutchman101:
After getting merged, it hasn't been documented on the wiki yet. Documenting your contributions is part of the coding guidelines, and our list of undocumented functionality is increasing, leading to no one using useful features.
[https://wiki.multitheftauto.com/wiki/SetElementData](https://wiki.multitheftauto.com/wiki/SetElementData)

@tederis

PR #1058, opened 2019-08-18
Add engineGetModelTextures function

### LINKED ISSUE #1040: Game TXD files access
Everyone knows that texutres are very important thing in mapping, but for this time we need to upload our own texutre-images to the server instead of using game texutres like in SAMP (there is SetObjectMaterial() function to sort this out.

The second advantage of having similar function in MTA is smaller amount of files which players have to download because textures (TXD files) are basic part of the game.

As I were talking with LopSided on discord there should be a function like `engineGetGameTexture( txdFileName ,  textureName)` which would return the texture element which we need. The example usage of this would be:
`engineGetGameTexture("rock_coastsfw", "Grass_128HV")` to get the grass texture.

Lpsd:
Resolves #1040 (although implementation is different from suggested)

This function allows you to get the textures of any model.

`engineGetModelTextures(model)`

**Arguments:**
model: either the model ID or model name

**Returns:**
Returns a table of texture elements _[textureName, texture]_, false otherwise.

**Example:**
_Get the textures for model ID 3722 and draw them with dxDrawImage_
```lua
function init()
	textures = engineGetModelTextures(3722)
end

function render()
	local offset = 0
	for name,texture in pairs(textures) do
		local size = dxGetPixelsSize(dxGetTexturePixels(texture))
		dxDrawImage(0+offset, 0, size, size, texture)
		offset = offset + size
	end
end
```
Example code


Client/game_sa/CRenderWareSA.cpp
```diff
@@ -755,6 +756,56 @@ void CRenderWareSA::GetModelTextureNames(std::vector<SString>& outNameList, usho
...
+// Get textures associated with the model
+// Will try to load the model if needed
+//
+////////////////////////////////////////////////////////////////
+void CRenderWareSA::GetModelTextures(std::vector<std::tuple<std::string, CPixels>>& outTextureList, ushort usModelId)
+{
+    outTextureList.clear();
+
+    // Special case for CJ
+    //if (usModelId == 0)
```
sbx320: Is this intentionally commented out?

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -755,6 +756,56 @@ void CRenderWareSA::GetModelTextureNames(std::vector<SString>& outNameList, usho
...
+        // Try model load
+        bLoadedModel = true;
+        pGame->GetModelInfo(usModelId)->Request(BLOCKING, "CRenderWareSA::GetModelTextures");
+        pTXD = CTxdStore_GetTxd(usTxdId);
+    }
+
+    std::vector<RwTexture*> rwTextureList;
+    GetTxdTextures(rwTextureList, pTXD);
+
+    for (std::vector<RwTexture*>::iterator iter = rwTextureList.begin(); iter != rwTextureList.end(); iter++)
```
sbx320: Just use a range-based for here for readability

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -755,6 +756,56 @@ void CRenderWareSA::GetModelTextureNames(std::vector<SString>& outNameList, usho
...
+    {
+        RwD3D9Raster* pD3DRaster = (RwD3D9Raster*)(&(*iter)->raster->renderResource);
+        CPixels       texture;
+        g_pCore->GetGraphics()->GetPixelsManager()->GetTexturePixels(pD3DRaster->texture, texture);
+        outTextureList.emplace_back((*iter)->name, texture);
```
sbx320: Using `outTextureList.emplace_back((*iter)->name, std::move(texture));` here would allow some  further compiler optimizations by using moves.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -928,6 +930,40 @@ int CLuaEngineDefs::EngineGetVisibleTextureNames(lua_State* luaVM)
...
+        return luaL_error(luaVM, argStream.GetFullErrorMessage());
+
+    ushort usModelID = CModelNames::ResolveModelID(strModelName);
+    if (usModelID != INVALID_MODEL_ID)
+    {
+        std::vector<std::tuple<std::string, CPixels>> textureList;
+        g_pGame->GetRenderWare()->GetModelTextures(textureList, usModelID);
+
+        lua_newtable(luaVM);
+        for (auto pair : textureList)
```
sbx320: should be `const auto&`

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -928,6 +930,40 @@ int CLuaEngineDefs::EngineGetVisibleTextureNames(lua_State* luaVM)
...
+        {
+            CClientTexture* pTexture = g_pClientGame->GetManager()->GetRenderElementManager()->CreateTexture("", &std::get<1>(pair), RDEFAULT, RDEFAULT,RDEFAULT, RFORMAT_UNKNOWN, TADDRESS_WRAP);
+            lua_pushstring(luaVM, (const char*)&std::get<0>(pair));
```
sbx320: This does not look correct. Just do `std::get<0>(pair).c_str()` here instead of weird casts.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -928,6 +930,40 @@ int CLuaEngineDefs::EngineGetVisibleTextureNames(lua_State* luaVM)
...
+    SString strModelName;
+
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadString(strModelName, "");
+
+    if (argStream.HasErrors())
+        return luaL_error(luaVM, argStream.GetFullErrorMessage());
+
+    ushort usModelID = CModelNames::ResolveModelID(strModelName);
+    if (usModelID != INVALID_MODEL_ID)
```
sbx320: I'd prefer `if (usModelID == INVALID_MODEL_ID) return 1;` here to reduce indentation depth and to keep consistent with the `HasErrors` call above.

sbx320 (dismissed):
Looks good, added some minor style comments. I'm not 100% sure on the lifetimes of the textures though. Might be worth trying to call engineGetModelTextures a few thousand times and seeing what happens to memory usage.

CrosRoad95:
does it work for custom models? you should also add filter parametr - which texture you want in case models has many, high quality textures and you want just one.

you should also do test by loop all models 1 to 20000 and check if it crash

Lpsd:
@sbx320 I have implemented the suggested changes, aside from the "range-based for loop" in CRenderWareSA - this seems to crash the client and I can't figure out why.

I have tested all model ID's to 20000 and there is no crash, also as long as you `destroyElement` on the texture then you are fine, no memory leak.

Next I'll implement "filter" parameter suggested by @CrosRoad95 , so you can get specific texture(s) from the result instead of returning all. After that I think this PR is ready.

Lpsd:
Added `textureNames` parameter. Allows you to choose which textures you want from the model by name. Supports a single string or a table of strings. Examples:

```lua
engineGetModelTextures(3722, "Was_sd")
engineGetModelTextures(3722, {"Was_sd", "Was_side"})
```

forkerer:
I would suggest using pattern matching similar to the one used in `engineApplyShaderToWorldTexture`, that supports wildcards and is already used in codebase

JobiwanD:
For the "range-based loop", I've changed these lines and it works well.

```C++
for (RwTexture* pTexture : rwTextureList)
{
    SString strTextureName = pTexture->name;

    /* ... */

    if (bValidTexture)
    {
        RwD3D9Raster* pD3DRaster = (RwD3D9Raster*)(&pTexture->raster->renderResource);
        /* ... */
    }

    /* ... */
}
```

I have restarted many times (~5000) a resource containing an improved version of your example. The memory usage remains stable **but**, as you have explained, the resource has to call `destroyElement` for each texture and for the table itself.

```lua
function init()
	textures = engineGetModelTextures(3722)
end

function terminate()
	for name,texture in pairs(textures) do
		destroyElement(texture)
	end
	destroyElement(textures)
end
addEventHandler("onClientResourceStop", resourceRoot, terminate)

function render()
	local offset = 0
	for name,texture in pairs(textures) do
		local pixels = dxGetTexturePixels(texture)
		local size = dxGetPixelsSize(pixels)
		dxDrawImage(0+offset, 0, size, size, texture)
		offset = offset + size
	end
end

```

JobiwanD:
@CrosRoad95 if you try it with a custom model, you will get the textures of the original model and the imported ones.

image

How can I help for the pattern part?

Lpsd:
I have implemented the suggestion for wildcard matching and cleaned up the ugly loop in `CRenderWareSA::GetModelTextures`

Requesting review

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1068,6 +1070,49 @@ int CLuaEngineDefs::EngineGetVisibleTextureNames(lua_State* luaVM)
...
+    else if (argStream.NextIsTable())
+        argStream.ReadStringTable(vTextureNames);
+
+    if (argStream.HasErrors())
+        return luaL_error(luaVM, argStream.GetFullErrorMessage());
+
+    ushort usModelID = CModelNames::ResolveModelID(strModelName);
+    if (usModelID == INVALID_MODEL_ID)
+    {
+        lua_pushboolean(luaVM, false);
```
qaisjp, line 1126: print warning here?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1068,6 +1070,49 @@ int CLuaEngineDefs::EngineGetVisibleTextureNames(lua_State* luaVM)
...
+        return 1;
+    }
+
+    std::vector<std::tuple<std::string, CPixels>> textureList;
+    g_pGame->GetRenderWare()->GetModelTextures(textureList, usModelID, vTextureNames);
```
qaisjp: so your `GetModelTextures` will simply leave `vTextureNames` empty if the texture names do not exist.

have you considered returning a bool, and returning false in failure scenarios? just a thought.

qaisjp (approved):
Requested a change or two but looks good

Lpsd:
The requested changes have been made.

PR #1059, opened 2019-08-18
Fix resetting dummies in vehicles with replaced models

### LINKED ISSUE #1057: Find out why a vehicle model's visual info may be nullptr
**Describe the bug**
Find out why [this](https://github.com/multitheftauto/mtasa-blue/blob/master/Client/game_sa/CModelInfoSA.cpp#L978) may be a nullptr sometimes, and fix underlying bug.
See #963 for the original hacky workaround fix.

**Expected behaviour**
Visual info should not be nullptr.

**Version**
Multi Theft Auto v1.5.6-release-18769

forkerer:
tries to fix #1057, i specified in comments there what is the most likely cause of the issues, this pull request adds an additional step in the process of unloading custom DFFs, in which it restores the default dummies if the model is a vehicle. I had to change the structure of dummies cache map a bit, as DFF cleaning does only have access to CModelInfo structure, and not underlying specializations. Also indexing the dummies cache by CModelInfo seems like a better choice in general IMO.


Client/game_sa/CModelInfoSA.cpp
```diff
@@ -950,39 +950,50 @@ void CModelInfoSA::SetVehicleDummyPosition(eVehicleDummies eDummy, const CVector
...
-    for (auto& info : ms_ModelDefaultDummiesPosition)
+    if (!IsVehicle())
+        return;
+
+    auto iter = ms_ModelDefaultDummiesPosition.find(this);
+    if (iter == ms_ModelDefaultDummiesPosition.end())
+        return; // Early out in case the model doesn't have any dummies modified
+
+    auto pVehicleModel = reinterpret_cast<CVehicleModelInfoSAInterface*>(m_pInterface);
+    for (auto& dummy : ms_ModelDefaultDummiesPosition[this])
```
qaisjp:
```suggestion
    for (const auto& dummy : ms_ModelDefaultDummiesPosition[this])
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -950,39 +950,50 @@ void CModelInfoSA::SetVehicleDummyPosition(eVehicleDummies eDummy, const CVector
...
-        for (auto& dummy : ms_ModelDefaultDummiesPosition[pVehicleModel])
-        {
-            // TODO: Find out why this is a nullptr, and fix underlying bug
-            if (pVehicleModel->pVisualInfo != nullptr)
-                pVehicleModel->pVisualInfo->vecDummies[dummy.first] = dummy.second;
-        }
-        ms_ModelDefaultDummiesPosition[pVehicleModel].clear();
-        // Decrement reference counter, since we reverted all position changes, the model can be safely unloaded
-        info.first->usNumberOfRefs--;
+        if (pVehicleModel->pVisualInfo != nullptr)
```
qaisjp: I can confirm that this fix works without this line!

forkerer: Nice, should i remove it then? I can't be 100% sure that there are no other edge cases that turn this field into null.

qaisjp: Perhaps we can log a line here instead. You can pick a number that isn't mentioned in this list: https://gist.github.com/qaisjp/fb9b2e78a04836c66344888a0161a003

Dutchman101:
Can this be merged?

qaisjp:
No, not in its current form.

Dutchman101:
> No

What else needs to be done?

I am curious about the effects of this on an atomics crash that I feel must have been introduced with the dummy features.

qaisjp (dismissed):
please fix whitespace changes & merge conflicts

qaisjp:
Sorry it took so long, @Dutchman101

qaisjp:
And thank you, forkerer!

PR #1066, opened 2019-08-21
Fix mem leak in LuaManager

pentaflops:
After disconnecting from the server, the CLuaManager should cleanup LuaVM, but does not

_With love from NEXTRP Team_

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
@@ -47,6 +47,9 @@ CLuaManager::~CLuaManager()
         delete (*iter);
     }
 
+	// Close and remove LVM from memory
+	ProcessPendingDeleteList();
+
     // Clear the C functions
     CLuaCFunctions::RemoveAllFunctions();
 }


pentaflops:
This is a critical leak :)

qaisjp:
Thank you for the fix!

patrikjuvonen:
I'm just curious to understand how https://github.com/multitheftauto/mtasa-blue/blob/master/Client/mods/deathmatch/logic/CClientGame.cpp#L1126 fails to free the memory in this case. I'm not exactly sure as to how the pulses work, but isn't `ProcessPendingDeleteList()` getting called? Do you have some stats that can show how significant the memory leak is as you're describing it to be "critical"? I suppose it makes sense to call the function in destructor though...

botder:
Leaving a server deletes CClientGame including CLuaManager. You don't run the pulse anymore to clean up afterwards.

patrikjuvonen:
> Leaving a server deletes CClientGame including CLuaManager. You don't run the pulse anymore to clean up afterwards.

Ok, we should cross check for other missing cleanups elsewhere too.

pentaflops:
> I suppose it makes sense to call the function in destructor though...

This is exactly what was done.

> Do you have some stats that can show how significant the memory leak is as you're describing it to be "critical"?

In the discord channel, I send screenshots showing a leak when disconnecting (reconnecting) and when restarting resources (when the player is on the server).

botder:
I don't see any reason not to merge this pull request and I have tested it. Thanks for the work.

qaisjp:
###### Copied from @pentaflops' messages on Discord

**before**

image

**after**

image

> Before the fix, the game crashed after 5 reconnections to the server. Now this is not observed (if you do not take into account the memory leak RW).

PR #1072, opened 2019-08-24
WIP: Make the quick connect button a reconnect button

qaisjp:
Someone said:

image

Currently that button only opens the Server Browser and says "_yo, you can copy it into the box up here_".

It's a relic from when the quick connect button used to open a small window and ask you for the host/port.

So now the button simply reconnects.

### DIFF
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -56,7 +56,6 @@ CServerBrowser::CServerBrowser()
     m_PrevServerBrowserType = ServerBrowserTypes::INTERNET;
 
     m_bFocusTextEdit = false;
-    m_uiShownQuickConnectHelpCount = 0;
     m_uiIsUsingTempTab = 0;
     m_BeforeTempServerBrowserType = ServerBrowserTypes::INTERNET;
     m_llLastGeneralHelpTime = 0;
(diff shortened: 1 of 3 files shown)


Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -2196,13 +2156,6 @@ void CServerBrowser::SetNextHistoryText(bool bDown)
 
 void CServerBrowser::OnQuickConnectButtonClick()
 {
-    // Show help text
```
qaisjp, line 2199: todo

ccw808:
For me, pressing the Quick Connect button shows a list of recent servers with the last connected server at the top

qaisjp:
Hmm. Yes, it does seem to do that too.

We should probably hook this up and determine how frequently quick connect is used. Not urgent

xerox8521:
Maybe make the reconnect as seperate menu entry and only visible if you connected atleast once to any server **successfully** (as in not rejected by the server due to eg wrong password or stuff like that). 

Would mean one would have to temporary save the most recent connection which updates automatically everytime you connect **successfully** to a server.

ArranTuna:
Or right clicking on quick connect would connect to the IP/port of an edit box in settings in the Multiplayer tab.

qaisjp:
> right clicking on quick connect

hmm i suppose rightclick is a something that can be leveraged here. what exactly are you hoping to connect to?

i'm guessing you connect to both your development server and actual server frequently, and would like to be able to just right click to connect to one of them?

ArranTuna:
That's right. Whichever IP/port is specified in an edit box would be useful in my case, but I guess a lot of players only play on one server so another option is that right clicking would simply /reconnect though nobody would be aware of this feature which is why an edit box in multiplayer tab of settings would also be useful as it could say something like: "Right clicking on quick connect will connect to:"

qaisjp:
I don't really want to continue with this. A keybind should work for now, to `connect` to whatever server. (Or we should make a change that lets it work.)

PR #1079, opened 2019-09-04
Fix dxDrawCircle segments argument

ApeLsiN4eG:
Fixed a bug when it drew `number of segments-1` segments instead of `number of segments` that you pass in the `segments` argument of the `dxDrawCircle` function.

**For example:** 
`dxDrawCircle(500, 500, 100, 0, 360, tocolor(255,255,255,255), tocolor(255,255,255,255), 4)` draws a triangle with 3 segments instead of a square.
`dxDrawCircle(500, 500, 100, 0, 360, tocolor(255,255,255,255), tocolor(255,255,255,255), 3)` draws nothing.


patrikjuvonen:
:+1:

Nice one!
Thank you for the PR and congratulations on your first contribution!  =)

Small sidenote: in the future when you make a patch, I suggest you create a new branch such as `bugfix/wrong-segment-count-in-dxDrawCircle` and create a pull request from there instead of directly from your `master` branch, as this way you can keep different bugfixes and features separated from each other and makes the whole process troubleless for both sides. Your `master` branch should basically always be identical to our `master` branch.

PR #1084, opened 2019-09-07
Increase max characters for kick and ban reason to 128

### LINKED ISSUE #1068: Increase max characters for reason in kickPlayer
Character limit is set to 64 for the reason message using kickPlayer function. Writing a larger string will be cut off in the message box. 

For example: "You have been kicked/banned by X for X days - Reason: You didn't behav....."

Increase max characters for reason string to 128 or 256.

Shorten the actual reason string to fit the max characters resulting in weird grammar/sentences.

StrixG:
Before:
image
After:
image

Fixes #1068

### DIFF
--- a/Client/mods/deathmatch/logic/Config.h
+++ b/Client/mods/deathmatch/logic/Config.h
@@ -84,14 +84,6 @@ public:
 #define MIN_DISCONNECT_REASON_LENGTH 1
 #define MAX_DISCONNECT_REASON_LENGTH 127
 
-// Max kick string length that can be sent
-#define MIN_KICK_REASON_LENGTH 1
-#define MAX_KICK_REASON_LENGTH 64
-
-// Max ban string length that can be sent
-#define MIN_BAN_REASON_LENGTH 1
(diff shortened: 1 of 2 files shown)


ccw808:
Might as well increase it a lot while we're here. What about 500? Also I think unused BAN/KICK_REASON_LENGTH in \Client\mods\deathmatch\logic\Config.h should be removed.

PlatinMTA:
But, what about the ban/kick message window? Isn't the limit made so the text doesn't break?

Haxardous:
> But, what about the ban/kick message window? Isn't the limit made so the text doesn't break?

it'll get extended (look closely, you'll notice that the length is different)

qaisjp:
500 sounds good.

Lpsd:
can we say 512 for OCD reasons

qaisjp:
> can we say 512 for OCD reasons

Now we must define it as `(int)511.13`

StrixG:
500 will be too much. The window has a fixed height.

qaisjp (approved):

PR #1093, opened 2019-09-13
Support for FBX files

CrosRoad95:
# General
This pull request introducing support for .FBX files natively supported by 3ds max, blender and i think zmodeler. It will develop in next updates for more features, for now are only required to work with it.

Whole idea is to make alternative way to render models with 100% control how they are drawn. It can almost replace .dff files. I created that pull request already to gather opinion, ideas, fix some issues and give more time to review all 16k+ lines.

This feature isn't done and is unstable (mainly lack of functions that clean up everything)
It is easy for me to control what left via the todo list below

# Overall idea how it should work

Everything is rendered base on `rendering templates`, you define what meshes and how should be rendered over specified elements and models ( same way as shaders already doing - apply and remove ). You can define up to 4 294 967 295 templates in context of one FBX element. Each template and template models has own properties such as offset position, draw distance, cull mode etc

# Advantages over DFF

- Almost non-blocking loading thanks to doing it in background on another thread makes you almost can't feel any lags.
- Smaller files with higher density of data (I get 3x smaller fbx file than dff from my tests)
- Plugins from 3ds max, blender no longer needed, those applications support it by default
- Support for multiple models in one file with hierarchy support
- Embedded textures, everything kept in one file (but still you will able to load textures from external sources)
- No limits in volume, complexity , amount of textures, models (up to 18 446 744 073 709 551 615 objects inside 1 .fbx file)
- Replace feature doesn't exists, you just select on which model/element you want to render template.
- Applying and removing from model/element has no cost in performance ( except first few frames while textures and other stuff can be created )
- Maximum possible draw distance is float limit ( three and 38 zeros ), trilions of trilions of trilions times greater than distance across gta world.
- Sky's the limit, whole process from beginning to the end can be modified without unnecessary reversing gta

# Functions and events

### Events:
- `onFBXLoaded` without parameters, source is our fbx element. called when get successfully loaded.

### Functions:
- `fbx fbxLoadFile( string raw data / file path )` - load provided fbx file
- `table(status) fbxGetLoadingStatus( fbx )` - used after fbxLoadFile and before `onFBXLoaded` returns current status of loading or error in occur
- `table(objects) fbxGetAllObjects( element fbx [ objectType filter] )` returns all objects ids and his type in pair table `[id] = type`, use filter to get ids with selected type
- `table fbxGetAllTemplates( fbx )` returns a table with all templates ids which can be use to render
- `table fbxGetAllTemplateModels( fbx, templateId )` returns all models ids in template
- `mixed fbxGetObjectProperties( fbx, objectId, object property )` - returns a properties of any object inside such a position, color, bounding box, parent, face count, vertices count, texture, patch of texture, hierarchy, type etc.
-  `mixed fbxGetTemplateProperties( fbx, templateId, template property )` simillar as above but releted to template
- `bool fbxSetTemplateProperties(fbx, templateId, property, value)` set properties
- `mixed fbxGetTemplateModelProperties( fbx, templateId, modelId, property)` as above
- `bool fbxSetTemplateModelProperties( fbx, templateId, modelId, property, value)` as above
- `int fbxAddTemplate( fbx ) ` - adds new template and return his id
- `bool fbxRemoveTemplate( fbx, templateId )` - remove template
- `int fbxTemplateAddModel( fbx, objectId ( mesh type ), parent, offset position )` adds mesh to template
- `bool fbxTemplateRemoveModel( fbx, modelId )` removes mesh from template
- `bool fbxApplyTemplateToModel(fbx, templateId, gtaModelId)` makes template renders over specified gta model id
- `bool fbxRemoveTemplateFromModel( fbx, templateId, gtaModelId)` removes template render from specified gta model id
- `bool fbxApplyTemplateToElement( fbx, templateId, element)` - same as fbxApplyTemplateToModel but per element
- `bool fbxRemoveTemplateFromElement( fbx, templateId, element)` - same as fbxRemoveTemplateFromModel but per element
- `bool fbxRenderTemplate( fbx, templateId, position, rotation, scale)` used in render event let you draw them just like dxDraw functions
- `table(textures) fbxGetAllTextures( fbx )` returns all created textures
- `table fbxGetObjectRawData( fbx, dataType, objectId )` return raw data such a vertices positions, indices, colors, uv maps etc
- `bool fbxTemplateExists( fbx, templateId)` - checks if template exists and return true or false

# Todo

- [x] make template 0 as default with exact copy of 3ds max/ blender scene
- [x] support for file path parametr in load function
- [x] add frustum culling ( avoid rendering meshes out of screen )
- [x] make something if user remove texture ( i think the best will add black white checkered texture used if something wrong with other )
- [x] template and template object draw distance property
- [x] make remove functions working.
- [x] meshes should disappear in same way as currently gta models doing
- [x] interiors and dimensions support
- [x] /showfbx simillar to /showcol what shows stats in the world
- [x] global lighting properties
- [ ] finish some functions, missing arguments ( example position which you set in before exporting ) or not working one
- [ ] fix lighting feature
- [ ] check for support for shaders and make if you set normal map and other simillar in 3d modeling tool, it will atomatically works without any script needed
- [ ] fix blending ( texture alpha ) issues if possible
- [ ] make something like optimized renderer queue, group buffers by textures
- [ ] add command like /showfbxstats what shows you gui simillar to showmemstats with stats of fbx
- [ ] add oop
- [ ] add function to clean up memory of .fbx file from raw data if you don't use it
- [ ] check if information about hide/show are exported and respect it in default template
- [ ] clean up code
- [ ] make something in mutex in duplicate line ( qaisjp or other person said it will added in some commit )
- [ ] night/day lighting
- [ ] levels of distance, property with template object id which will rendered if some other will too far
- [ ] remove getAllTextures and replace it into something like getAllElements which return all mta elements ( textures, shaders and other ) useded or belonged to fbx
- [ ] turn on/off property for templates to disable them 
- [ ] check remove functions against memory leaks, require fix of getElementsByType fix

# Issues

- From unknown reason, getElementsByType leaking memory
- Alpha blending

# Future

- global settings for fbx such a:
- - way of calculating draw distance - depends on bounding sphere or not
- sharing resources between fbx elements ( rendering one model from two individual .fbx files )
- parametr to use content of other template
- custom occlusion zones
- if flag enabled ( by default false ), sun/moon color, size, position should affect models
- Support for animations, bones, blending vertices
- Add new properties for texture element to control them even more
- make `fbxAddTemplate` second parametr is id of other template, if used copy will created, same with `fbxTemplateAddModel` but you can provide `template id, model id` to copy from other template, or just `model id` to do it from same template
- fbxRenderTemplate support extra argument "element" type which can be used to do extra transformations based of that element
- use multithreading to increase even more rendering power
- filter for load function to skip parts you don't need
- creating own meshes from lua
- add function to build .col file from fbx mesh
- supports for animations like rotating parts
- add function to replace textures and shaders inside fbx and properties to do it per template/ template object
# Misc

- almost all functions will return false if you use them before file gets loaded
- doing files above some size will take a lot of time to verify file by mta, avoid 1GB .fbx file which will work but loading can take so long, better to split it into parts
- fbxRenderTemplate is useful for things like attachments for vehicles, weapons, backpacks and other you need only visual effect
- fbxGetObjectRawData i introduced to make possible get for example positions of water planes
- each 1MB of .fbx file can store about 288000 triangles
- don't try to load .fbx files greater than 20MB of raw mesh data, it typicaly is ~5,76 milions of vertices and simillar data
- if you rendering over model/element, most values like positions, scale, interior are used
- fbxApplyTemplateToModel can be used with different templates over one gta model id, same with ToElement version
- not everything is supported ( example animations ids will not returned by fbxGetAllObjects ), same with properties of them.
- not all objects supports all properties, example geometry supports vertices count but not color
- by object i meant object which is inside fbx file, texture, material, mesh, geometry etc

Test resource i'm using with 2 .fbx files
[example resource.zip](https://github.com/multitheftauto/mtasa-blue/files/3611210/example.resource.zip)

Example videos of current progress
https://www.youtube.com/watch?v=hbUCIeYiAHE
https://www.youtube.com/watch?v=j6qhzwaGy1M

lag caused by loading 2MB file ( from zip above )
image

This is how looks default template
image

Sources of project i used for reading/writing fbx files https://github.com/nem0/OpenFBX
i didn't describe everything :) mta team will have a fun with testing and reviewing the code


Client/game_sa/CGameSA.h
```diff
@@ -467,6 +468,7 @@ class CGameSA : public CGame
     CWeaponStatManager*             m_pWeaponStatsManager;
     CPointLights*                   m_pPointLights;
     CObjectGroupPhysicalProperties* m_pObjectGroupPhysicalProperties;
+    CFBXManager*        m_pFBXManager;
```
qaisjp: formatting !

Client/mods/deathmatch/StdInc.h
```diff
@@ -50,6 +50,9 @@
 #include "CIdArray.h"
 #include "pcrecpp.h"
 
+// fbx
+#include "fbx/ofbx.h"
```
qaisjp, line 54: I think we want to avoid adding new header files to StdInc.h wherever possible.

the others in this file are fine, since we haven't established a new convention for luadefs or CClientBlah, but you should try to remove this line specifically from this file

Client/mods/deathmatch/logic/CClientManager.cpp
```diff
@@ -177,6 +178,9 @@ CClientManager::~CClientManager()
 
     delete m_pModelManager;
     m_pModelManager = nullptr;
+
+    delete m_pFBXManager;
+    m_pFBXManager = NULL;
```
qaisjp:
```suggestion
    m_pFBXManager = nullptr;
```

Einheit-101:
Are you telling me that this can bypass GTAs rendering limit of ~1500 meters (LODs) and i can draw an FBX model through the whole GTA world if i want to??

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -437,4 +437,5 @@ void CLuaManager::LoadCFunctions()
     CLuaWaterDefs::LoadFunctions();
     CLuaWeaponDefs::LoadFunctions();
     CLuaXMLDefs::LoadFunctions();
+    CLuaFBXDefs::LoadFunctions();
```
qaisjp: This is sorted alphabetically, so please move this higher

Client/mods/deathmatch/logic/luadefs/CLuaDefs.cpp
```diff
@@ -30,6 +30,7 @@ CClientPickupManager*      CLuaDefs::m_pPickupManager = NULL;
 CClientDFFManager*         CLuaDefs::m_pDFFManager = NULL;
 CClientColModelManager*    CLuaDefs::m_pColModelManager = NULL;
 CRegisteredCommands*       CLuaDefs::m_pRegisteredCommands = NULL;
+CClientFBXManager*         CLuaDefs::m_pFBXManager = NULL;
```
qaisjp:
```suggestion
CClientFBXManager*         CLuaDefs::m_pFBXManager = nullptr;
```

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
```diff
@@ -26,6 +26,7 @@ class CClientPlayerManager;
 class CClientRadarMarkerManager;
 class CClientTeamManager;
 class CClientVehicleManager;
+class CClientFBX;
```
qaisjp: Shouldn't this say this instead?
```suggestion
class CClientFBXManager;
```

Client/sdk/core/CCoreInterface.h
```diff
@@ -23,6 +23,7 @@
 #include "CWebCoreInterface.h"
 #include "CTrayIconInterface.h"
 #include "CChatInterface.h"
+#include "CFBXInterface.h"
```
qaisjp: Is it possible for you to just do a forward declaration here?

Client/sdk/game/CGame.h
```diff
@@ -80,6 +80,7 @@ typedef void(InRenderer)();
 #include "CWorld.h"
 #include "TaskCarAccessories.h"
 #include "CObjectGroupPhysicalProperties.h"
+#include "CFBXManager.h"
```
qaisjp, line 83: Is it possible for you to just do a forward declaration here?

Shared/sdk/CMatrix.h
```diff
@@ -27,6 +27,15 @@ class CMatrix
         vUp = CVector(0.0f, 0.0f, 1.0f);
         vPos = CVector(0.0f, 0.0f, 0.0f);
     }
+    
+    CMatrix(float* floats)
```
qaisjp: This looks very unsafe

Client/core/CFBX.h
```diff
@@ -9,15 +9,24 @@
  *
  *****************************************************************************/
 
+#include "CFrustum.h"
```
qaisjp, line 12: forward decl please

Client/core/CFrustum.cpp
```diff
@@ -0,0 +1,182 @@
+#include "StdInc.h"
```
qaisjp, line 1: does this really need the entire StdInc.h?

ccw808: I believe `StdInc.h` is required for precompiled headers

CrosRoad95: I tried without but vs required stdinc.h

ArranTuna:
> alternative way to render models

Please could you elaborate on this, like would it be more efficient to have mods in this FBX format? Would their rendering have less impact on performance?

vendor/fbx/ofbx.cpp
```diff
@@ -2159,7 +2159,7 @@ namespace ofbx
             {
                 vec = vertices[geom->to_old_vertices[i]];
                 geom->vertices[i] = CVector(vec.x, vec.y, vec.z);
-                geom->indices[i] = codeIndex(i, i % 3 == 2);
+                geom->indices[i] = std::abs(codeIndex(i, i % 3 == 2));
```
qaisjp: please don't add any modifications to vendor files

qaisjp:
There are a few commits that modify the `vendor/fbx` folder. Please consider sending some of those changes as improvements upstream (i.e. send pull requests to [nem0/OpenFBX](https://github.com/nem0/OpenFBX)).

vendor/fbx/ofbx.h
```diff
@@ -0,0 +1,485 @@
...
+    };
+
+    struct Cluster : Object
+    {
+        static const Type s_type = Type::CLUSTER;
+
+        Cluster(const Scene& _scene, const IElement& _element);
+
+        virtual const int*    getIndices() const = 0;
+        virtual int           getIndicesCount() const = 0;
```
qaisjp, line 271: `getPlurals()` and `getSingularCount()`. So `getIndexCount` in the original code is correct.

CrosRoad95:
i'll send some pull request to his repo after i finish this pull request and will work stable with changes i did

> Please could you elaborate on this, like would it be more efficient to have mods in this FBX format? Would their rendering have less impact on performance?

with optimization i'm going to do ( culling, render distances, grouping by textures, buffers ) it will handle up to few milions of triangles at once.

but if want to render scene with thousends of different textures, it will lag anyway due `setTextures` issues

nbredikhin:
Awesome work! 
Will it be compatible with `engineApplyShaderToWorldTexture`? 
Will it be possible to access model hierarchy (something like `setVehicleComponentVisible`, but for models)?
Will it be possible to add support for animations in future?

CrosRoad95:
> Will it be compatible with engineApplyShaderToWorldTexture?

no, but will simillar functionality
> Will it be possible to access model hierarchy 

yes
> Will it be possible to add support for animations in future?

yes

CrosRoad95:
Project starts to become just HUGE, what makes me to change the plans.
1. project will splitted into a lot of smaller commits, pull request.
2. If someone has suggestions how to rename functions, i'm open at suggestions. ( template mean: what models of .fbx should be used to create other, individual model base on sub models .fbx has )
3. Initial release will contain 2 functions: load and draw with position, rotation and scale argument.

After that, i will start to marge other feature i already did.
For now i'm closing this pull request.

mrxenginner:
Considering this PR about Bullet Physics, are there any examples of converting a character's bones to the world of GTA?

tryangelic:
There are many PRs like this one that close, very well-described and well-elaborated PRs that end up not being very relevant to some; it would be fascinating if this were added

PR #1098, opened 2019-09-20
Fix texture blending

### LINKED ISSUE #1097: Blend mode is ignored in TileBatcher
**Describe the bug**
The blend mode is forced to `blend` when drawing a shader or a texture with any addressing except `wrap`.

**To reproduce**
1. Download [blendtest.zip](https://github.com/multitheftauto/mtasa-blue/files/3623857/blendtest.zip) and start it.
2. Change blend mode using the `blend overwrite` command.
3. Blending won't change on `clamp`, `mirror`, `border` textures.

**Expected behaviour**
The blend mode should change.

**Screenshots**
image

**Version**
Multi Theft Auto v1.5.7-release-20210

**Additional context**
PrimitiveBatcher may also be affected.

StrixG:
Fixes #1097 

- [x] Fix Tile blending
- [x] Fix Primitive blending

### DIFF
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -488,6 +488,9 @@ void CGraphics::CheckModes(EDrawModeType newDrawMode, EBlendModeType newBlendMod
     // Draw mode changing?
     if (bDrawModeChanging || bBlendModeChanging)
     {
+        SetBlendModeRenderStates(newBlendMode);
+        m_CurBlendMode = newBlendMode;
+
         // Flush old
         if (m_CurDrawMode == EDrawMode::DX_SPRITE)
         {
@@ -521,16 +524,9 @@ void CGraphics::CheckModes(EDrawModeType newDrawMode, EBlendModeType newBlendMod
         {
             m_pLineInterface->Begin();
         }
-    }
 
-    // Blend mode changing?
-    if (bDrawModeChanging || bBlendModeChanging)
-    {
-        SetBlendModeRenderStates(newBlendMode);
+        m_CurDrawMode = newDrawMode;
     }
-
-    m_CurDrawMode = newDrawMode;
-    m_CurBlendMode = newBlendMode;
 }
 
 void CGraphics::CalcWorldCoors(CVector* vecScreen, CVector* vecWorld)
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -1733,7 +1733,7 @@ void CGraphics::DrawQueueItem(const sDrawQueueItem& Item)
                 fU2 *= fUScale;
                 fV2 *= fVScale;
             }
-            CheckModes(EDrawMode::TILE_BATCHER);
+            CheckModes(EDrawMode::TILE_BATCHER, Item.blendMode);
             m_pTileBatcher->AddTile(t.fX, t.fY, t.fX + t.fWidth, t.fY + t.fHeight, fU1, fV1, fU2, fV2, t.pMaterial, t.fRotation, t.fRotCenOffX, t.fRotCenOffY,
                                     t.ulColor);
             RemoveQueueRef(Item.Texture.pMaterial);
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -488,8 +488,6 @@ void CGraphics::CheckModes(EDrawModeType newDrawMode, EBlendModeType newBlendMod
     // Draw mode changing?
     if (bDrawModeChanging || bBlendModeChanging)
     {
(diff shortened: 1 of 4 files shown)


Client/core/Graphics/CGraphics.cpp
```diff
@@ -488,6 +488,9 @@ void CGraphics::CheckModes(EDrawModeType newDrawMode, EBlendModeType newBlendMod
     // Draw mode changing?
     if (bDrawModeChanging || bBlendModeChanging)
     {
+        SetBlendModeRenderStates(newBlendMode);
```
ccw808: Why did you move `SetBlendModeRenderStates` to here?

Client/core/Graphics/CPrimitiveBatcher.cpp
```diff
@@ -100,24 +100,10 @@ void CPrimitiveBatcher::UpdateMatrices(float fViewportSizeX, float fViewportSize
 void CPrimitiveBatcher::SetDeviceStates()
 {
     // Set states
-    m_pDevice->SetRenderState(D3DRS_ZENABLE, D3DZB_FALSE);
-    m_pDevice->SetRenderState(D3DRS_ZFUNC, D3DCMP_LESSEQUAL);
-    m_pDevice->SetRenderState(D3DRS_ZWRITEENABLE, FALSE);
-    m_pDevice->SetRenderState(D3DRS_CULLMODE, D3DCULL_NONE);
```
forkerer, line 106: Why do you remove stuff not related to blending? Deleting this one will result in out of order primitives not rendering because of backface culling

StrixG: Oops

ccw808: @forkerer Those states are already correctly set in `CGraphics::SetBlendModeRenderStates` https://github.com/multitheftauto/mtasa-blue/blob/4c3bbb91468205d8e55c53e6a2cfe96ca306f2c7/Client/core/Graphics/CGraphics.cpp#L344

Client/core/Graphics/CGraphics.cpp
StrixG: The first shader or primitive in the queue didn't change the blend mode for some reason.

ccw808: Changing the render states here will mean buffered draws will use the incorrect blend mode. (The following `->End()` and `->Flush()` calls ensure rendering using the previous blend mode is complete)

I think the first shader problem is caused by something else

Client/core/Graphics/CPrimitiveBatcher.cpp
forkerer: That's interesting, i remember i had some issues with culling when writing it, that's why it was set explicitly there. Maybe I'm misremembering something, sorry for false alarm

Client/core/Graphics/CTileBatcher.cpp
```diff
@@ -141,23 +141,6 @@ void CTileBatcher::Flush()
         if (g_pDeviceState->AdapterState.bRequiresClipping)
             m_pDevice->SetRenderState(D3DRS_CLIPPING, TRUE);
         m_pDevice->SetRenderState(D3DRS_ZENABLE, m_bUseCustomMatrices ? D3DZB_TRUE : D3DZB_FALSE);
```
ccw808, line 147: D3DRS_ZENABLE will need restoring after the draw

Client/core/Graphics/CTileBatcher.cpp
```diff
@@ -142,6 +146,13 @@ void CTileBatcher::Flush()
             m_pDevice->SetRenderState(D3DRS_CLIPPING, TRUE);
         m_pDevice->SetRenderState(D3DRS_ZENABLE, m_bUseCustomMatrices ? D3DZB_TRUE : D3DZB_FALSE);
 
+        m_pDevice->SetTextureStageState(0, D3DTSS_COLOROP, D3DTOP_MODULATE);
+        m_pDevice->SetTextureStageState(0, D3DTSS_COLORARG1, D3DTA_TEXTURE);
+        m_pDevice->SetTextureStageState(0, D3DTSS_COLORARG2, D3DTA_DIFFUSE);
+        m_pDevice->SetTextureStageState(0, D3DTSS_ALPHAOP, D3DTOP_MODULATE);
+        m_pDevice->SetTextureStageState(0, D3DTSS_ALPHAARG1, D3DTA_TEXTURE);
+        m_pDevice->SetTextureStageState(0, D3DTSS_ALPHAARG2, D3DTA_DIFFUSE);
+
```
ccw808: Why are these `SetTextureStageState` calls required?

StrixG: These calls aren't required since blend modes set them.

PR #1128, opened 2019-10-20
Add getElementOffsetPosition function

CrosRoad95:
function: `x,y,z = getElementOffsetPosition( element, offsetX, offsetY, offsetZ)`
oop: `vector3 = element:getOffsetPosition( offsetX, offsetY, offsetZ)`
It is the "useful function" getOffsetFromXYZ already availiable at wiki https://wiki.multitheftauto.com/wiki/GetOffsetFromXYZ.

I propose to add this function due is commonly used on many servers and example on wiki not working without matrix class. It is also faster because of c++ implementation.

Test resource draws cross for each vehicle and object. download: [offsetposition.zip]
(https://github.com/multitheftauto/mtasa-blue/files/3748052/offsetposition.zip)

Green lines are from oop variant.
image

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -369,6 +369,14 @@ bool CStaticFunctionDefinitions::GetElementPosition(CClientEntity& Entity, CVect
     return true;
 }
 
+bool CStaticFunctionDefinitions::GetElementOffsetPosition(CClientEntity& Entity, CVector vecOffset, CVector& vecPosition)
+{
+    CMatrix matrix;
+    Entity.GetMatrix(matrix);
+    vecPosition = matrix.TransformVector(vecOffset);
+    return true;
+}
+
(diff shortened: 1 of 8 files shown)


Einheit-101:
I suggest adding getPositionFromElementOffset() aswell, which gets a 3D position from an offset of an element.

```lua
function getPositionFromElementOffset(element, offX, offY, offZ)
	local m = getElementMatrix ( element )
	local x = offX * m[1][1] + offY * m[2][1] + offZ * m[3][1] + m[4][1]
	local y = offX * m[1][2] + offY * m[2][2] + offZ * m[3][2] + m[4][2]
	local z = offX * m[1][3] + offY * m[2][3] + offZ * m[3][3] + m[4][3]
	return x, y, z
end
```

moon91210:
Don't we already have the transform method?

```lua
function getElementOffsetPosition(elem, ox, oy, oz)
	return elem.matrix:transformPosition(ox, oy, oz)
end
```

qaisjp:
I'm not convinced this is necessary. https://github.com/multitheftauto/mtasa-blue/pull/1128#issuecomment-552210070 should work fine. Leaving open for 1 week for further comments.

PR #1133, opened 2019-10-25
Add dxCreateTexture multithreading support

CrosRoad95:
It adds `useMultithreading` at last argument ( boolean ) for dxCreateTexture what cause almost non-blocking load of textures and function `dxIsTextureCreated` to check did multithreading loading done.

Here's comparison
image

Problems i found:
1. Owner of server is able to freeze user pc by creating many big textures. My solution is to add limit in settings in "advanced" with multithreading limits, combobox with limits 0 up to 10 and unlimited
2. I don't think i did it correct due multithreading require twices many ram for a while to create texture
image

3. Scripters would know how much multithreading can they use, thats why i propose to add some function to get how many threads is in use, how many left for each category ( for texture, and others in future ). `int inUse, int limit = getMultithreadingStatus( string category = "texture" )`
4. Should functions fail if multithreading is limited? or just force to false when limit reached?
5. From unknown reason, sometimes texture just don't load.

Test resource:
[multithreadingdxcreatetexture.zip](https://github.com/multitheftauto/mtasa-blue/files/3772882/multithreadingdxcreatetexture.zip)


Pirulax:
You can deny an application to use a thread(s) in the task manager, no need for the checkbox.

0x416c69:
Why not just use CAsyncTaskScheduler for this? All that matters is for textures to be loaded in a different thread than the main thread which causes the game to halt.

You start creating a lot of threads and things gonna get weird inside windows, and if you are going to set a limit on threads, then scripters need to handle things by their own.

qaisjp:
I agree with @0x416c69. Exposing multithreading in this way feels kind of icky to me, and unnecessarily complicated for the scripter.

I also think dxCreateTexture should take a callback with the texture in the callback argument - this is consistent with everything else that touches multiple threads.

thisdp:
> I agree with @0x416c69. Exposing multithreading in this way feels kind of icky to me, and unnecessarily complicated for the scripter.
> 
> I also think dxCreateTexture should take a callback with the texture in the callback argument - this is consistent with everything else that touches multiple threads.

Maybe nodejs will be a good example in this situation. I think there is no need to check whether the texture is being created use "getMultithreadingStatus". Just give a call back of success or not.

0x416c69:
@qaisjp @thisdp Yes. I also had this idea to set a callback for it but there will be a slight problem with mass loading and that is the scripter gonna deal with a lot of callbacks execution.

I was gonna say let's go with both callback and dxIsTextureCreated.

qaisjp:
@0x416c69 I was thinking that they would do something like this:

```lua
local textures = {}
local loaded = 0
local texturePaths = {"a.png", "b.png", "c.png", } -- etc

function store(key)
	return function(texture)
		loaded = loaded + 1
		textures[key] = texture

		if loaded == #texturePaths then
			loaded = true
		end
	end
end

addEventHandler("onClientResourceStart", resourceRoot, function()
	for _, path in ipairs(texturePaths) do
		-- This can return false if the path is wrong
		-- or some other problem with queueing up the creation
		-- (for actual creation in another thread)
		local success = dxCreateTexture(store(path), path, "argb")
		if not success then
			outputDebugString("Failed to create texture")
			cancelEvent()
			return
		end
    end
end)

function render()
	if loaded ~= true then
		dxDrawText("Loading", 200, 200)
		return
	end

	for i, path in ipairs(texturePaths) do
		local texture = textures[path]
		dxDrawImage(i * 30, 0, 30, 30, texture)
	end
end
	
```

(This assumes the callback is the first argument to `dxCreateTexture`.)

(Code is obviously untested.)

0x416c69:
@qaisjp Although the same code with dxIsTextureCreated will take less LOC, I say let's go with only callbacks and avoid calling dxIsTextureCreated every frame/timeSlice to check if all of the textures are loaded which will have a cost on lua VM and performance.

CrosRoad95:
i refactored code to make it work with callback function but i stuck. In general it working, but very often whole mta hanging and consume a lot of ram even total textures size i 10MB, red line is when i started resource.,
image

thisdp:
I'd like to let dxCreateTexture returns an empty texture element even it is loading, and fill the texture after loaded.

qaisjp:
> I'd like to let dxCreateTexture returns an empty texture element even it is loading, and fill the texture after loaded.

What's the use case? Sounds like this is susceptible to a race condition

CrosRoad95:
can race condition cause both threads hangs?

qaisjp:
Race condition can do anything, it's undefined behaviour. Well actually it can probably just be solved by a lock, but I still don't understand the use case.

thisdp:
> Race condition can do anything, it's undefined behaviour. Well actually it can just be solved by a lock, but I still don't understand the use case.

some thing like
https://wiki.multitheftauto.com/wiki/DgsRequestQRCode
which doesn't require a call back function

Pirulax:
I think using std::async wouldve been the smartest idea, because in some instances it takes more time to create a thread than to create the texture, especially when creating ARGB tex.

well, MTA's thread pool does the same actually, but i guess its the best to not use proprietary things.

Pirulax:
> can race condition cause both threads hangs?

im pretty sure its casued by interfering with the d3d9 device when ctoring the underlaying data of the texture.

Edit: yes, this PR will never work this way. Its using non-thread safe functions, and its pure luck if it does.

Pirulax:
This PR will never work in its current state, it uses non thread-safe functions(especially when creating the texuture with the d3d9 devoce, in 'CreateUnderlayingData')

Pirulax:
@CrosRoad95 I think you can close this PR. I tried to do the same, but the biggest bottleneck here is the texture conversion most of the time(or just creating the texture in the d3d9 device with 'CreateTexture'). Since d3d9 isnt thread safe by default, (as said before), this only works by pure luck.

If Sam implements dx11, we can do this easily, because dx is thread safe.

PR #1145, opened 2019-11-02
Primitive elements for fast 2d and 3d rendering

CrosRoad95:
It adds element type "primitive-buffer" which store data, needed to rendering huge amount of primitives at once. It is piece of FBX support.

Main reason why i created this reason is issue with textures, it just don't rendering at all, instead the sky color is showing.

It provides 3 functions:
```lua
primitiveBufferCreate( primitiveFormat, primitiveType, [ table of indices, ], table vertex1, table vertex2, table vertex3 )
```

`primitiveFormat` - comma separeted enum strings. Defines format of vertices and does indices are in use.
```
1. `xyz` - 3d position.
2. `xy` - 2d position, `z` is set to 0 automatically.
3. `uv` - texture coords, makes rendering require material.
4. `diffuse` - color, result of `tocolor` function.
5. `16bitindices` - 16 bit indices are in use, first table before vertices describe them, must contain divisable by 3 amount of non-negative numbers, index can not be larger than amount of vertices. can adress up to ~21845 faces. Take just 6 bytes for each face.
6. `32bitindices` - 32 bit version, as above but allow to address milions of faces. 12 bytes per face.
```

Restrictions: can't use `xyz` and `xy` at once, can't use `16bitindices` and `32bitindices` at once.
Example: `xyz,diffuse` or `diffuse,xyz` will make vertex format must be `{x,y,z, color}`, define ordering doesn't matter, but vertex format is order restricted: `position, diffuse, uv`
More examples: indices  {1,2,3, 1,2,4} with 4 vertex defined and "trianglelist" primitive type will draw for example square, otherwise without indices you have to change primitive type or add two another, duplicated vertices
```lua
primitiveBufferDraw( primitive-buffer, [material, ] positionXYZ, rotationXYZ, scaleXYZ, postgui, view)
primitiveBufferDraw3D( primitive-buffer, [material, ] positionXYZ, rotationXYZ, scaleXYZ, postgui, view)
```
`view` can be `perspective` or `orthographic` ( i think 3d version should not have view argument )

Other:
1. i did not created structs for each variant of vertex due in future it will accept more vertex format data what will cause i have to create dozens of struct for each case.
2. Multiple stream are cause due above reason and will useful for future eg.: switching between day/night lighting or edit things
3. You can pass `z` coord even for gui things, that's purposeful - you can use it to make preview of model in gui just like object_preview does but much faster without fps drop.
4. I introduces indices due it can reduce amount of ram usage a lot, for example terrain grid can reduce amount of ram usage by even 70% and i'll make you don't need to add duplicated vertices what can cause rendering problem ( eg. 1 pixel gap)

Test resource: 
[primitivebuffers.zip](https://github.com/multitheftauto/mtasa-blue/files/3800954/primitivebuffers.zip)

Screenshots:

image
primitive buffers
image


CrosRoad95:
I refactore it and add another checks while creating, right now i'm writing unit tests to test all possible cases. Problem with texture still appear.

Now new issue has appeared, these two memory leak, i need help with them:
image
image
image

Pirulax:
God damn.
Use smart pointers, so you wont have any memory leaks....

CrosRoad95:
i'm still waiting for help

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -236,6 +236,8 @@ CClientGame::CClientGame(bool bLocalPlay)
     // Singular file download manager
     m_pSingularFileDownloadManager = new CSingularFileDownloadManager();
 
+    m_pPrimitiveBufferManager = new CClientPrimitiveBufferManager(m_pManager);
```
ghost, line 240: Where are you deallocating `m_pPrimitiveBufferManager`?

CrosRoad95: i think i doesn't deallocate it

ghost: It's causing a memory leak. It's better to use `std::unique_ptr` here. It's a smart pointer, it will automatically get deleted when it goes out of scope, i.e., when client.dll gets unloaded. 

```
static std::unique_ptr<CClientPrimitiveBufferManager> m_pPrimitiveBufferManager= nullptr;

// In CClientGame::CClientGame
m_pPrimitiveBufferManager = std::make_unique<CClientPrimitiveBufferManager>(m_pManager);
```

Client/core/Graphics/CPrimitiveBuffer3dBatcher.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    // Draw
+    m_pDevice->SetTexture(0, nullptr);
+
+    // Cache last used material, so we don't set directx parameters needlessly
+    CMaterialItem* pLastMaterial = nullptr;
+
+    for (auto& primitive : m_primitiveBufferMap)
+    {
+        if (primitive.first != nullptr && primitive.second.size() > 0)
```
sbx320: It might be nicer to invert these ifs to reduce indentation depth (`if(...) continue;`)

Client/core/Graphics/CPrimitiveBufferBatcher.cpp
```diff
@@ -0,0 +1,308 @@
...
+ *  PURPOSE:
+ *
+ *
+ *****************************************************************************/
+
+#include <StdInc.h>
+#include "../Client/mods/deathmatch/logic/CClientPrimitiveBufferInterface.h"
+#include "CPrimitiveBufferBatcher.h"
+
+////////////////////////////////////////////////////////////////
```
sbx320: Is there a specific reason for these large (and mostly empty) comment blocks?

Client/mods/deathmatch/logic/CClientPrimitiveBuffer.cpp
```diff
@@ -0,0 +1,73 @@
...
+void CClientPrimitiveBuffer::Unlink()
+{
+    g_pClientGame->GetPrimitiveBufferManager()->Delete(this);
+}
+
+CClientPrimitiveBuffer::~CClientPrimitiveBuffer()
+{
+    delete m_pIndexBuffer;
+    for (int i = 0; i < 8; i++)
+        if (m_arrayVertexBuffer[i] != nullptr)
```
sbx320, line 40: This if isn't actually required, as `delete nullptr` is allowed (and has no effect)

Client/mods/deathmatch/logic/CClientPrimitiveBuffer.h
```diff
@@ -0,0 +1,116 @@
...
+            FVF = D3DFVF_TEX1;
+            break;
+        case PRIMITIVE_DATA_DIFFUSE:
+            m_vecVertexElements.push_back({2, 0, D3DDECLTYPE_D3DCOLOR, D3DDECLMETHOD_DEFAULT, D3DDECLUSAGE_COLOR, 0});
+            index = 2;
+            FVF = D3DFVF_DIFFUSE;
+            break;
+    }
+
+    VOID* pVoid;            // POINTER TO POINTER, remember forkerer
```
sbx320: this should be `void`, `VOID` is stupid windows stuff

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+CClientPrimitiveBufferManager::~CClientPrimitiveBufferManager()
+{
+    DeleteAll();
+}
+
+CClientPrimitiveBuffer* CClientPrimitiveBufferManager::Get(unsigned long ulID)
+{
+    // Find the PrimitiveBuffer with the given id
+    list<CClientPrimitiveBuffer*>::const_iterator iter = m_List.begin();
```
sbx320: You could just use a range based for here:
```
for(conat auto& iter : m_List) 
{
    if(iter->GetID() == ulID)
       return iter;
}
return nullptr;
```

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+    for (; iter != m_List.end(); iter++)
+    {
+        if ((*iter)->GetID() == ulID)
+        {
+            return *iter;
+        }
+    }
+
+    return NULL;
```
sbx320: `nullptr`

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+}
+
+CClientPrimitiveBuffer* CClientPrimitiveBufferManager::Create()
+{
+    CClientPrimitiveBuffer* pPrimitiveBuffer = new CClientPrimitiveBuffer(m_pManager, INVALID_ELEMENT_ID);
```
sbx320: Are new & delete really required here? A `std::unique_ptr` would be much nicer.

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+void CClientPrimitiveBufferManager::Delete(CClientPrimitiveBuffer* pPrimitiveBuffer)
+{
+    RemoveFromList(pPrimitiveBuffer);
+    delete pPrimitiveBuffer;
+}
+
+void CClientPrimitiveBufferManager::DeleteAll()
+{
+    list<CClientPrimitiveBuffer*>::const_iterator iter = m_List.begin();
```
sbx320: Bonus: If `m_List` contains unique_ptrs only, this stuff isn't required, as `m_List.clear()` will delete everything automatically

sbx320 (commented):
Some minor comments about some code style stuff. Not sure if the DX stuff is correct as it has been years since I looked at DX9

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
CrosRoad95: is `conat` some new c++ 20 standard keyword which speed up everything by 0.0001% ? i'm joking ofc

CrosRoad95:
image
now just tests required against leaks, crashes, various material input ect.

CrosRoad95:
i ran test resource, restarted it few times, and here's result from memory usage
image
looks fine for me. Here's a bit different script i used ( command /testprimitives ): 
[bufs.zip](https://github.com/multitheftauto/mtasa-blue/files/4055895/bufs.zip)

Kiriller12:
Isn't it better to use a table to store all the vertex data due to limitations of the unpack function, or can some function be added to add vertices to the primitive instead of creating a new one? The unpacking limit is approximately 8000 items.

ghost:
Looks good. I will test this tomorrow.

ghost:
I merged your PR into the latest master branch, and I compiled it in release mode. Every time I execute _testleaks_, the memory goes up, and it never goes back even if you reconnect or restart the resource. After executing _testleaks_ command 6-7 times, the memory went up to 820mb from 500mb.

CrosRoad95:
i don't know why ;/

qaisjp (changes requested):
not a full review, but the memory leak that saml1er reports needs to be fixed (at least)

Client/mods/deathmatch/logic/CClientPrimitiveBufferManager.cpp
Pirulax: const doesnt do anything, well, nearly anything. It mostly is for `visuals`, that is, it indicates, that the data wont be changed. Here you can't use it, since im pretty sure `GetID()` isn't `const`, so, just use `auto& iter`.

Pirulax:
I'll refactor the code to use smart ass pointers. Hope that'll fix the problems

Pirulax:
I think its because the index and vertex buffers aren't `release`-d. Not quite 

> testleaks

Seems like the buffers weren't freed in the destructor(`SAFE_RELEASE` wasn't called). Now after adding that, it seems like to be better. 
Ran the command 4-5 times, the mem usage only increased by 20 mb. 
After disconnecting it went down to 411(from 438). I was at 420 when starting the test. 
There are probably a few minor leaks as well.

PR #1157, opened 2019-11-20
Add GitHub Actions support

qaisjp:
See https://github.com/qaisjp/mtasa-blue/commit/6f885c54f9a01f911a98d106410521f952546ff0/checks for an example of what this looks like.


qaisjp:
would be nice to:
- [x] [actions/upload-artifact](https://github.com/actions/upload-artifact)
- [ ] get rid of the docker image being on dockerhub and put it on [multitheftauto/mtasa-blue/packages](https://github.com/multitheftauto/mtasa-blue/packages)
- [x] [actions/cache](https://github.com/actions/cache) directx
- [ ] [actions/cache](https://github.com/actions/cache) cef
- [x] delete azure pipelines

qaisjp:
Build artifacts are in action here: https://github.com/qaisjp/mtasa-blue/runs/312502546

image

qaisjp:
DXFiles caching has been added. It only shaves five seconds, but it is a good example of this working well!

**[Cache miss](https://github.com/qaisjp/mtasa-blue/runs/312967480)**

image

**[Cache hit](https://github.com/qaisjp/mtasa-blue/runs/312994634)**

image

qaisjp:
CEF caching needs a little bit of work since we first need to emit something like this: `echo ::set-output name=cef_version::HERE_CEF_VERSION_HERE`

And use `${{ steps.STEP_ID.outputs.cef_version }}` in our cache key + cache path.

Because of the cache size limit we'll also want to cache the compressed version, not the unpacked version, so we'll need to update the cef buildaction to support that.

qaisjp:
- GitHub Packages requires authentication to pull images and so is a little bit of extra work. Not necessary right now and can be done in a separate PR.
- CEF cache can also be done in a separate PR.

I'm happy with what we've got now. Please can I have a review @sbx320?

sbx320 (approved):
Looks good to me

PR #1164, opened 2019-11-24
Add analog control sync for accelerate and brake_reverse

Addlibs:
This pull adds analog control state sync for ButtonSquare (_brake_reverse_) and ButtonCross (_accelerate_) to correct replication of analog control values when using a controller. 

In previous version, if the driver applied an analog value of 0.2 to accelerate, the actual sync packet sent to the server was simply a bool of whether accelerate is pressed or not, and the server would pass that onto other clients who interpreted it as analog value of 1.0 (full acceleration). In this pull, the analog value is correctly replicated and applied onto remote player control states. A video demonstration of the change's effect: https://youtu.be/4u5caoSxQqQ (from non-driver client). 

Additionally, these changes mean the size of the _SSmallKeysyncSync_ and _SFullKeysyncSync_ grows from 40 to 58 bits in length when sending both analog values. When controller isn't being used, the size will be 42 bits in length.

Since this pull changes bitstream messages, I've increased the MTA_DM_BITSTREAM_VERSION and made it backward-compatible.


qaisjp:
Awesome work!

ccw808:
To help reduce bandwidth usage I suggest only sending the analog values when they are in the range 1-254.
Example:
```
    if (data.ucButtonSquare >= 1 && data.ucButtonSquare =< 254)
    {
        bitStream.WriteBit(true);
        bitStream.Write(data.ucButtonSquare);
    }
    else
        bitStream.WriteBit(false);
```

qaisjp:
@ccw808 please can you review this if you're happy with it?

qaisjp (dismissed):
after we merge master in this, bitstream needs bump. [see diff without `Merge branch 'master'`.](https://github.com/multitheftauto/mtasa-blue/pull/1164/files/d0431fe7a3c11640df62eacb002bada1b93cbedf)

ccw808 (approved):

Issue #567, opened 2018-09-23
Fix setAnalogControlState


LosFaul:
Edit: i cleaned up this topic, because it was a mess before

**Is your feature request related to a problem? Please describe.**
setAnalogControlState is not working when there is actual player input.

```lua
addEventHandler("onClientPreRender", root,
    function()
        local accelerate = getAnalogControlState("accelerate")
        setAnalogControlState("accelerate", accelerate * 0.1)
    end
)
```
in this example i try to modify the player input,
but setAnalogControlState has no effect

**Describe the solution you'd like**
Fix setAnalogControlState, so that it allows to modify player input

LosFaul:
Today i made some testing again, to ensure how things work
(getAnalogControlState/setAnalogControlState)

`addEventHandler("onClientPreRender", root,
	function()
		local accelerate = getAnalogControlState("accelerate")
		setAnalogControlState("accelerate", accelerate * 0.1)
	end,
	false
)`

the new set value by setAnalogControlState is always overwritten and the use of this function has no effect
i'm not sure if it is inteded to work this way
but then there should be an additional argument to setAnalogControlState to not allow an additional overwrite for the current frame from input:
setAnalogControlState(controlName, value[, disableOverwriteForCurrentFrame = false])

LosFaul:
this is a bump with another example:
reversing controls

```lua
addEventHandler("onClientPreRender", root,
    function()
        -- get controll values
        local left = getAnalogControlState("vehicle_left")
        local right = getAnalogControlState("vehicle_right")
        -- reverse controll values
        setAnalogControlState("vehicle_left", right)
        setAnalogControlState("vehicle_right", left)
    end
)
```
this does not work either

you can it get working for keyboards but **not for gamepads** as gamepads will bypass it
keyboard solution would be currently to unbind controll keys and make and own check if the keys are pressed

but if the issue would be fixed keyboard and gamepads would share the same solution

qaisjp:
Remind me to self-assign 1 month from now. Issue is open to someone else until then.

LosFaul:
I thought a bit how a potential solution could be done.
Keeping the game process order in mind: https://wiki.multitheftauto.com/wiki/Game_Processing_Order
and not trying to break existing scripts

a solution could look like this:
```lua
setAnalogControlState(string controll, float value[, forced = false])
```
the solution would be to add a third optional argument which is per default false
this is in order to not break existing scripts
if forced is true
[...]

PR #1165, opened 2019-11-24
Add ability to get raw controller analog value

Addlibs:
Adds a bool parameter to [getAnalogControlValue](https://wiki.multitheftauto.com/wiki/GetAnalogControlState) and [getPedAnalogControlValue](https://wiki.multitheftauto.com/wiki/GetPedAnalogControlState) to poll for raw controller input for a given control value, ignoring scripted overrides, control toggles, etc.

This allows scripts to do things like mirroring analog input for accelerate onto `brake_reverse`. This will ignore non-analog inputs for the control, in other words, it'll return 0 for `accelerate` when pressing `W` on the keyboard unless you're actually pressing on a controller trigger (or whichever axis is bound to "accelerate").

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPad.cpp
+++ b/Client/mods/deathmatch/logic/CClientPad.cpp
@@ -473,7 +473,7 @@ bool CClientPad::GetAnalogControlState(const char* szName, CControllerState& cs,
     unsigned int uiIndex;
     if (GetAnalogControlIndex(szName, uiIndex))
     {
-        // Do we have a script override?
+        // If we are not ignoring overrides and have a script override
         if (!bIgnoreOverrides && m_sScriptedStates[uiIndex] != CS_NAN)
             switch (uiIndex)
             {
--- a/Client/mods/deathmatch/logic/CClientPad.cpp
+++ b/Client/mods/deathmatch/logic/CClientPad.cpp
@@ -468,13 +468,13 @@ bool CClientPad::GetAnalogControlIndex(const char* szName, unsigned int& uiIndex
 }
 
 // Get the analog control state directly from a pad state.  Use for players.
-bool CClientPad::GetAnalogControlState(const char* szName, CControllerState& cs, bool bOnFoot, float& fState)
+bool CClientPad::GetAnalogControlState(const char* szName, CControllerState& cs, bool bOnFoot, float& fState, bool bIgnoreOverrides)
 {
     unsigned int uiIndex;
     if (GetAnalogControlIndex(szName, uiIndex))
     {
         // Do we have a script override?
-        if (m_sScriptedStates[uiIndex] != CS_NAN)
+        if (!bIgnoreOverrides && m_sScriptedStates[uiIndex] != CS_NAN)
             switch (uiIndex)
             {
                 case 0:
--- a/Client/mods/deathmatch/logic/CClientPed.h
+++ b/Client/mods/deathmatch/logic/CClientPed.h
@@ -573,7 +573,7 @@ public:
     CStatsData*                              m_stats;
     CControllerState*                        m_currentControllerState;
     CControllerState*                        m_lastControllerState;
-    CControllerState                         m_rawControllerState;
+    CControllerState                         m_rawControllerState; // copy of lastControllerState before CClientPed::ApplyControllerStateFixes is applied (modifies states to prevent stuff like rapid input glitch)
     CRemoteDataStorage*                      m_remoteDataStorage;
     unsigned long                            m_ulLastTimeFired;
     unsigned long                            m_ulLastTimeBeganAiming;
--- a/Client/mods/deathmatch/logic/CClientPed.h
+++ b/Client/mods/deathmatch/logic/CClientPed.h
@@ -573,6 +573,7 @@ public:
     CStatsData*                              m_stats;
     CControllerState*                        m_currentControllerState;
     CControllerState*                        m_lastControllerState;
+    CControllerState                         m_rawControllerState;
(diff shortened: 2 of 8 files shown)


Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -573,6 +573,7 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
     CStatsData*                              m_stats;
     CControllerState*                        m_currentControllerState;
     CControllerState*                        m_lastControllerState;
+    CControllerState                         m_rawControllerState;
```
qaisjp: A future programmer might be confused about the difference between `m_currentControllerState` and `m_rawControllerState` (specifically, what `current` has that `raw` doesn't have). I would suggest you include comments here that explain this.

Client/mods/deathmatch/logic/CClientPad.cpp
```diff
@@ -468,13 +468,13 @@ bool CClientPad::GetAnalogControlIndex(const char* szName, unsigned int& uiIndex
 }
 
 // Get the analog control state directly from a pad state.  Use for players.
-bool CClientPad::GetAnalogControlState(const char* szName, CControllerState& cs, bool bOnFoot, float& fState)
+bool CClientPad::GetAnalogControlState(const char* szName, CControllerState& cs, bool bOnFoot, float& fState, bool bIgnoreOverrides)
 {
     unsigned int uiIndex;
     if (GetAnalogControlIndex(szName, uiIndex))
     {
         // Do we have a script override?
```
qaisjp: This comment should be updated to say something like:

```suggestion
        // If we are not ignoring overrides and have a script override
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1660,17 +1660,21 @@ bool CStaticFunctionDefinitions::GetPedControlState(CClientPed& Ped, const char*
...
 {
     if (Ped.GetType() == CCLIENTPLAYER)
     {
         CControllerState cs;
         Ped.GetControllerState(cs);
         bool         bOnFoot = (!Ped.GetRealOccupiedVehicle());
         unsigned int uiIndex;
+
+        if (bRawInput)
+            cs = Ped.m_rawControllerState;
```
qaisjp: If `bRawInput` is `true` then line 1668 is useless. I would suggest this:

```suggestion
            cs = Ped.m_rawControllerState;
        else
            Ped.GetControllerState(cs);
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1660,17 +1660,21 @@ bool CStaticFunctionDefinitions::GetPedControlState(CClientPed& Ped, const char*
     return false;
 }
 
-bool CStaticFunctionDefinitions::GetPedAnalogControlState(CClientPed& Ped, const char* szControl, float& fState)
+bool CStaticFunctionDefinitions::GetPedAnalogControlState(CClientPed& Ped, const char* szControl, float& fState, bool bRawInput)
 {
     if (Ped.GetType() == CCLIENTPLAYER)
     {
         CControllerState cs;
         Ped.GetControllerState(cs);
```
qaisjp:
```suggestion
```

(see above)

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -7005,13 +7009,17 @@ bool CStaticFunctionDefinitions::GetControlState(const char* szControl, bool& bS
...
+bool CStaticFunctionDefinitions::GetAnalogControlState(const char* szControl, float& fState, bool bRawInput)
 {
     CControllerState cs;
     CClientPlayer*   pLocalPlayer = m_pPlayerManager->GetLocalPlayer();
     pLocalPlayer->GetControllerState(cs);
     bool bOnFoot = (!pLocalPlayer->GetRealOccupiedVehicle());
-    if (CClientPad::GetAnalogControlState(szControl, cs, bOnFoot, fState))
+
+    if (bRawInput)
+        cs = pLocalPlayer->m_rawControllerState;
```
qaisjp:
```suggestion
        cs = pLocalPlayer->m_rawControllerState;
    else
        pLocalPlayer->GetControllerState(cs);
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -7005,13 +7009,17 @@ bool CStaticFunctionDefinitions::GetControlState(const char* szControl, bool& bS
     return false;
 }
 
-bool CStaticFunctionDefinitions::GetAnalogControlState(const char* szControl, float& fState)
+bool CStaticFunctionDefinitions::GetAnalogControlState(const char* szControl, float& fState, bool bRawInput)
 {
     CControllerState cs;
     CClientPlayer*   pLocalPlayer = m_pPlayerManager->GetLocalPlayer();
     pLocalPlayer->GetControllerState(cs);
```
qaisjp:
```suggestion
```

qaisjp (approved):
Thanks for the pull request! Your proposal makes sense. I've left some comments.

qaisjp (changes requested):
Whoops, meant to _request changes_.

LosFaul:
just to keep things on track
this might be connected to this stuff here:
https://github.com/multitheftauto/mtasa-blue/issues/567

qaisjp (approved):
lgtm, thank you! :shipit:

PR #1167, opened 2019-11-25
Fix #1158: add outputDebugString level 4 to omit debug info

### LINKED ISSUE #1158: Adjustment for outputDebugString - level 0
I've always used outputDebugString in my server, this is very nice tool, which imho could be improved a bit, by applying some changes. Thing i miss it's hiding path to script for level 0 (**Custom message**), which isn't fully customizable, because it doesn't allow to hide/remove path to script file. I'm aware of that, i can use level 3, which doesn't output path (but still adds prefix **INFO:**), but for this level, colors aren't working, which isn't good for me (used them to distinguish debugs - for client/server, and for different scripts), besides path extends debug message, which for me it's very unnecessary.

Let user decide if path to script should be shown, by removing it (**script/server.lua:**) so it will output just text that user want. Or by adding 6th argument which will decide if path should be hidden. By default it should be false. I think 2nd solution it's much better.

Creating custom script for debugscript.

**Additional context**
-

Addlibs:
This pull adds a new parameter to outputDebugString for omitting file path and line number info from a custom debug message. This pull closes #1158


Shared/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -156,12 +157,12 @@ void CScriptDebugging::LogCustom(lua_State* luaVM, const char* szMessage)
...
-                                 unsigned char ucRed, unsigned char ucGreen, unsigned char ucBlue)
+                                 unsigned char ucRed, unsigned char ucGreen, unsigned char ucBlue, bool bOmitDebugInfo)
 {
     SString strText = ComposeErrorMessage(szPrePend, luaDebugInfo, szMessage);
 
-    // Create a different message if type is "INFO"
-    if (uiMinimumDebugLevel > 2)
+    // Create a different message if type is "INFO" or it's a custom message with debug lines output omitted
+    if (uiMinimumDebugLevel > 2 || bOmitDebugInfo)
         strText = SString("%s%s", szPrePend, szMessage);
```
Citizen01: Should be if-else here so we don't call `ComposeErrorMessage` if the if statement below (line 165) is true.

Shared/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -55,7 +56,7 @@ void CScriptDebugging::LogCustom(lua_State* luaVM, unsigned char ucRed, unsigned
     VSNPRINTF(szBuffer, MAX_STRING_LENGTH, szFormat, marker);
     va_end(marker);
 
-    LogString("", GetLuaDebugInfo(luaVM), szBuffer, 0, ucRed, ucGreen, ucBlue);
+    LogString("", GetLuaDebugInfo(luaVM), szBuffer, 0, ucRed, ucGreen, ucBlue, bOmitDebugInfo);
```
Citizen01: What about not changing the `LogString` method definition and sending `nullptr` as the 2nd argument if `bOmitDebugInfo` is `true` ?
I mean, what's the point of getting infos from the LuaVM and sending it if you already know you will just discard that data ?
(Ofc, you will need to check for the `luaDebugInfo` before using it)

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Output.cpp
```diff
@@ -133,12 +133,14 @@ int CLuaFunctionDefs::OutputClientDebugString(lua_State* luaVM)
     unsigned char    ucRed = 255;
     unsigned char    ucGreen = 255;
     unsigned char    ucBlue = 255;
+    bool             bOmitDebugInfo = false;
```
Citizen01: No more hungarian notation as per the new coding style.
(Should be confirmed first because [the wiki page](https://wiki.multitheftauto.com/wiki/Coding_guidelines#Style) still hasn't change since the decision was made)

ccw808: Why are we ditching Hungarian notation? Won't it make the codebase inconsistent?

Citizen01: I really like hungarian notation to be honest. If it was only me, I would have keep it.
I can't remember if we decided to keep it depending of where the modifications are made to keep it consistent with the legacy code around.
@qaisjp and/or @botder ?

Shared/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -55,7 +56,8 @@ void CScriptDebugging::LogCustom(lua_State* luaVM, unsigned char ucRed, unsigned
     VSNPRINTF(szBuffer, MAX_STRING_LENGTH, szFormat, marker);
     va_end(marker);
 
-    LogString("", GetLuaDebugInfo(luaVM), szBuffer, 0, ucRed, ucGreen, ucBlue);
+    // Send empty SLuaDebugInfo if omitDebugInfo is true
+    LogString("", omitDebugInfo ? SLuaDebugInfo() : GetLuaDebugInfo(luaVM), szBuffer, 0, ucRed, ucGreen, ucBlue);
```
Citizen01: Instantiating an empty `SLuaDebugInfo` to act as if we were not sending debug info looks very hacky-ish to me.

Addlibs: Well, the function takes a reference, so I can't use a `nullptr` and I'd rather avoid redefining it to use a pointer. `SLuaDebugInfo` constructor actually produces a pretty well defined no-debug-info object. https://github.com/multitheftauto/mtasa-blue/blob/0d0d60c7df86413d03066903549c565083a716c4/Client/mods/deathmatch/logic/lua/LuaCommon.h#L93 https://github.com/multitheftauto/mtasa-blue/blob/0d0d60c7df86413d03066903549c565083a716c4/Server/mods/deathmatch/logic/lua/LuaCommon.h#L87

I could change it to something like 
```cpp
SLuaDebugInfo debugInfo;
if (!omitDebugInfo)
    debugInfo = GetLuaDebugInfo(luaVM);
    
LogString("", debugInfo, szBuffer, 0, ucRed, ucGreen, ucBlue);
```
if that looks less hackish. Thoughts?

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Output.cpp
qaisjp: I don't remember, the oldest thing I can find is https://discordapp.com/channels/278474088903606273/278474088903606273/531797552102768641

> I can't remember if we decided to keep it depending of where the modifications are made to keep it consistent with the legacy code around.

That sounds like it makes sense.

cc @multitheftauto/blue-contributors

qaisjp:
If this is only available for level `0`, why not just add a level `4` with this behaviour? (This would rename `0` from `custom` to `debug`. And make `4` the new "custom".)

Alternatively, we could keep your feature behind an extra boolean, but also make it work for any level. 

What do you think?

Addlibs:
I do like the idea of moving it to its own type/level. Much cleaner solution than an additional bool I guess.

qaisjp (changes requested):
Waiting for new level

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -368,7 +368,7 @@ int CLuaFunctionDefs::DownloadFile(lua_State* luaVM)
                         }
                     }
                 }
-                m_pScriptDebugging->LogCustom(luaVM, 255, 255, 255, "%s: File doesn't exist", lua_tostring(luaVM, lua_upvalueindex(1)));
+                m_pScriptDebugging->LogCustom(luaVM, SString("%s: File doesn't exist", lua_tostring(luaVM, lua_upvalueindex(1))));
```
Addlibs, line 371: I've noticed this message is unique across the whole codebase in that it used non-prefixed white-colored LogCustom instead of "WARNING: " prepended orange message. I've changed that to be consistent with the rest of the codebase.

Shared/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -55,6 +55,21 @@ void CScriptDebugging::LogCustom(lua_State* luaVM, unsigned char ucRed, unsigned
     VSNPRINTF(szBuffer, MAX_STRING_LENGTH, szFormat, marker);
     va_end(marker);
 
+    SLuaDebugInfo luaDebugInfo;
+    LogString("", luaDebugInfo, szBuffer, 0, ucRed, ucGreen, ucBlue);
```
Addlibs, line 59: The new LogCustom still uses 0 as the uiMinimumDebugLevel so that we don't have to add a 4th level to /debugscript and instead simply show them in the same level as previous custom-level messages. The differenciation between LogCustom and LogDebug is based on whether `DEBUG_INFO_NONE` of SLuaDebugInfo is used instead of real debug info.

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
qaisjp: okay

qaisjp:
Thank you!

qaisjp (commented):
Damn, I hoped this would get rid of the "Changes requested" message

PR #1177, opened 2019-12-06
Add "remember this option" checkbox to NVidia Optimus dialog

Lpsd:
Mantis: https://bugs.multitheftauto.com/view.php?id=9450

This adds a "remember this option" checkbox to the optimus dialog - as this dialog pops up every time optimus is detected, which can be annoying for the user! 

The remember/skip option resets with `HandleResetSettings()` (when user didn't reach load screen 5 times in a row). Maybe there are other cases in which this option needs to be reset?

Preview:

NVidia Optimus dialog preview

Edit: The wording is actually rather confusing, if you know how the whole process works. Your selection on this dialog is actually saved already - maybe it should be "Don't show again" rather than "Remember this option", as this implies that your selection isn't being saved in the first place.


Einheit-101:
Holy moly, thanks. Please add this asap. This annoys me for years now.

MrDadosz:
Nice. Same thing should be added for window when you have more than one monitor.

qaisjp:
Please can we swap the ordering of 'If you get desperate, force windowed mode' and 'Remember this option'? And maybe also add a caption like this: "Once you have picked the best option that works, this might help:"?

Lpsd:
The requested changes have been implemented

Dialog Preview

qaisjp (approved):
lgtm :shipit:

PR #1163, opened 2019-11-23
Add hasElementData Lua feature to client and serverside

Simi23:
Issue #1060 suggests that there should be an 'isElementData' feature which would look for element data on a given key just like 'getElementData'.

`bool hasElementData ( element theElement, string key [, bool inherit = true] )`

### **Required Arguments**

- **theElement:** This is the element with data you want to test for.
- **key:** The name of the element data entry you want to test for.

### **Optional Arguments**

- **inherit:** - toggles whether or not the function should go up the hierarchy to find the requested key in case the specified element doesn't have it.

Resolves #1060 

##### edit by qaisjp: updated syntax to use new function name

PR #1203, opened 2020-01-07
Add "ok" result to getElementData

### LINKED ISSUE #1060: isElementData
If I set `false` using `setElementData()` then it is not possible to find out if there is data with the required key or not, because both cases will return `false`.

add function `isElementData()`

It is possible to get all data using `getAllElementData()` and then check a required key, but it may seriously affect the performance.

**Additional context**
no

qaisjp:
This resolves #1060 and replaces #1163. 

A simple case:

image

Which is useful if you've set something to `false`:

image

/cc @patrikjuvonen

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -326,47 +326,6 @@ int CLuaElementDefs::GetElementByIndex(lua_State* luaVM)
     return 1;
 }
 
-int CLuaElementDefs::GetElementData(lua_State* luaVM)
-{
-    //  var getElementData ( element theElement, string key [, inherit = true] )
-    CClientEntity* pEntity;
-    SString        strKey;
-    bool           bInherit;
-
-    CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pEntity);
-    argStream.ReadString(strKey);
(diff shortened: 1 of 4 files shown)


qaisjp:
Force pushed to fix a typo in the commit messages.

patrikjuvonen:
I am not sure what I've said before, but I prefer #1163 over this after all.

So I'm giving my vote for `isElementData` rather than a second returned value from `getElementData`.

I still like the clean-up you've done otherwise however by merging them under shared.

qaisjp:
An extra return value will break lots of scripts. Rejected.

patrikjuvonen:
For future reference we had a discussion on Discord about this PR at  https://discordapp.com/channels/278474088903606273/366384007535001612/666672959829835777 onwards.

We found that this would potentially break existing scripts and introduce inconvenient behavior.

Calling tonumber directly will result in an error.
```lua
x = tonumber(getElementData(player, "posX"))
-- bad argument #2 to 'tonumber' (number expected, got boolean)
```

Using getElementData directly inside a table will insert a boolean value which may break some code (e.g. concatenation, loops) if boolean values are not handled correctly.
```lua
position = {
    getElementData(player, "posX"),
    getElementData(player, "posY"),
    getElementData(player, "posZ")
}
-- example of resulting table: {1, 1, 1, true}
```
Probably a few more examples could be provided but these were enough to reject the proposal.

PR #1213, opened 2020-01-17
Fix #1174: accept team & table of players in outputChatBox

### LINKED ISSUE #1174: outputChatBox - expand visibleTo possibility
outputChatBox doesn't accept table nor team element (well, for team this function returns true, but still doesn't output it.)

Expand visibleTo argument by adding send list (like triggerClientEvent has as 1st argument; array table. I think that feature isn't there without a reason?). Besides, it would be nice if user pass team element as visibleTo argument, then player which belongs to this team would receive message.

Sending message separately to player using loop (not good idea, but one and only, if i am wrong correct me.)

**Additional context**
-

StrixG:
Closes #1174

### DIFF
--- a/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -9847,10 +9847,7 @@ void CStaticFunctionDefinitions::OutputChatBox(const char* szText, const std::ve
 {
     assert(szText);
 
-    for (const auto& pPlayer : sendList)
-    {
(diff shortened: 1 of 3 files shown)


qaisjp (approved):

qaisjp (approved):
LGTM, thank you!

Dutchman101:
This change breaks existing scripts, especially gamemodes: if there's a chathandler replacement (like `freeroam` has), all chat messages will be output twice if the player is in a team.

Issue #1223

Citizen01:
Mentionned #1223 issue now fixed in 24a233f6e897771504a270940e05ccc4652bce53

PR #1221, opened 2020-01-23
Optimised images

ricksterhd123:
credits goes to imgbot,
not sure if it's worth much but thought it could help


Woovie:
I like this idea, but for .png, I'd like to see if using https://github.com/google/zopfli would be better if we are going to bother doing this. (Thanks @Dutchman101)

Any idea what they use to compress PNG in this?

ricksterhd123:
Not sure but it worked on my fork so i thought maybe i should pull request, i was thinking of doing the same thing to the resources repository too.

ricksterhd123:
I'll definitely look at that zopfli - maybe one could write a script to automate image compression

Woovie:
Yes, I could write a simple bash script to do zopfli to the repo. I'll do a few and do image size comparisons to see if it's worth it.

Woovie:
Replaced with #1732

PR #1224, opened 2020-01-24
Fix #1223: double chat messages if player is in team

### LINKED ISSUE #1223: Regression: double chat messages if player is in team
**Describe the bug**
As a result of d2c53c7, a player that is part of a team will see all chat messages being duplicated, if there is a custom chathandler on the server (cancelling `onPlayerChat` and processing the messages itself). Therefore, the change breaks existing scripts.

You can say that it's a niche issue, but one of the default MTA resources (`freeroam`) uses a custom chathandler and therefore is affected. MTA should never break any existing scripts.

**To reproduce:**
1. Start a resource with a custom chathandler, e.g:

```lua
addEventHandler("onPlayerChat", root,
    function(msg, type)
        if type == 0 then
            cancelEvent()

            if isElement(source) then
                local r, g, b = getPlayerNametagColor(source)
                outputChatBox(getPlayerName(source) .. ": #FFFFFF" .. msg:gsub("#%x%x%x%x%x%x", ""), root, r, g, b, true)
                outputServerLog("CHAT: " .. getPlayerName(source) .. ": " .. msg)
            end
        end
    end
)
```

2. Create and enter a team
3. Send a message in the global chat
4. Observe the double messages (of everyone on the server, including yourself):
dup

**Version**
Server: v1.5.7-release-r20391

**Additional context**
Not reproducible on Server: v1.5.7-release-r20386 so it's clearly a result of d2c53c7

ghost:
Fixes #1223 ~~and also adds support for `vehicle` element~~. `RUN_CHILDREN` was calling `g_pTeamManager` which ran `OutputChatBox` again with `player` element.

download

Test resource: [1223.zip](https://github.com/multitheftauto/mtasa-blue/files/4106913/1223.zip)

### DIFF
--- a/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -9837,21 +9837,6 @@ bool CStaticFunctionDefinitions::OutputChatBox(const char* szText, CElement* pEl
         }
         return true;
     }
-    else if (IS_VEHICLE(pElement))
-    {
-        CVehicle* pVehicle = static_cast<CVehicle*>(pElement);
-        
-        for (int i = 0; i < MAX_VEHICLE_SEATS; ++i)
-        {
-            CPed* pPed = pVehicle->GetOccupant(i);
-            CPlayer* pPlayer = (pPed ? static_cast<CPlayer*>(pPed) : nullptr);
-
-            if (pPlayer)
-            {
-                pPlayer->Send(CChatEchoPacket(szText, ucRed, ucGreen, ucBlue, bColorCoded));
-            }
-        }
-    }
     else
     {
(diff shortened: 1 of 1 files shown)


qaisjp:
Thanks for sending the fix. I don't think supporting vehicles should be added. Are there any other functions that do this?

ghost:
Well, i thought about that you might want to send messages to all passengers. For example to display next bus station name in rp gamemodes - spares a for loop. :P plus I dont think that it would hurt but can remove if you like. I think that MTA should provide what's possible rather than not adding them.

And I dunno about other functions. Maybe `triggerClientEvent` ?

qaisjp:
Yeah, I can't find any function that does this. Please can you remove that from this PR?

Feel free to create a separate _proposal_ instead of packaging this new feature alongside a bugfix

ghost:
done @qaisjp

PR #1234, opened 2020-01-27
Add ability to play sounds from raw data to playSound(3D)

Dezash:
This adds the ability to load sounds from raw data in playSound and playSound3D functions similarly to engineLoadCOL/TXD/DFF/IFP.

I would appreciate some feedback on URL checking and whether it would be appropriate to replace regex with something else like checking the string length.


qaisjp:
> I would appreciate some feedback on URL checking and whether it would be appropriate to replace regex with something else like checking the string length.

For engine functions we currently do something like this: https://github.com/multitheftauto/mtasa-blue/blob/b21322ce5f112a51136cedd7d17c99b5c971c584/Client/mods/deathmatch/logic/CClientColModel.cpp#L156-L160

That's insufficient for URLs. And we can't check for a magic header here (without checking for all possible kinds).

I would say just assume it's a raw string if the string contains a new line, I wouldn't use a regex here.

Dezash:
> I would say just assume it's a raw string if the string contains a new line, I wouldn't use a regex here.

Is it always the case that raw data would contain new lines?
If not, maybe we could use the fact that URLs have a de facto limit of 2048 characters. Any sound file will probably be larger than 2KB.

AlexTMjugador:
No sound container format I'm aware of starts with the "http" string, and I highly doubt a sound container format would ever want to start with a confusing signature like "http", so I think that checking whether the first 4 bytes of the parameter are case-insensitive equal to "http" is enough. Personally, I find a regex to be overkill for this purpose, as it involves a compilation process under the hood and such.

Edit: I have realized that a valid M3U playlist file can be plain text URLs separated by line breaks. I think combining the signature check I mentioned with line-feed search as proposed by @qaisjp would be the best approach. After all, playing a M3U playlist with a single URL is the same as playing that URL directly :)

Edit 2: the algorithm I'm thinking on would roughly be like this:
- Compare the first 4 bytes with "http", case insensitively. If they are not equal, it can't be an URL, so treat it as raw.
- If the first 4 bytes are equal to "http", search for the first line feed character. If there is a line feed character, then the data is possibly a raw M3U playlist (URLs can't contain line feed characters). In other case, it is a URL (or a M3U playlist with a single URL, so it doesn't matter if we treat it as raw or not, as the end result would be the same). I assume no sound container format starts with "http".

Edit 3: as an optimization, if during the line feed character search a boundary is exceeded (like the 2 KiB @Dezash mentioned), the algorithm may give up and treat the input as raw, as it's very unlikely that such a large input would ever be a valid URL. It'd be nice to adjust this boundary to the size of the internal buffer that BASS uses to store URLs.

Edit 4: if more URL schemas are to be supported, the comparison with "http" can be replaced with a comparison with "http" or any of the interesting URL schemas.

Dezash:
Done.
The algorithm extracts the first 4 bytes and lowers its characters. It then compares whether they match "http" or "ftp". If they do then it checks whether the length is less than 2048 characters or if it doesn't contain any line feed characters.

Client/mods/deathmatch/logic/CClientSoundManager.cpp
```diff
@@ -123,14 +123,22 @@ void CClientSoundManager::SetDimension(unsigned short usDimension)
...
     {
         pSound->PlayStream(strSound, bLoop, bThrottle);
         return pSound;
     }
+    else if (bIsRawData)
+    {
+        size_t size = strSound.size();
+        void* pMemory = new char[size];
+        memcpy(pMemory, strSound.data(), size);
+        if (pSound->Play((void*)pMemory, size, bLoop))
```
qaisjp, line 139: Just dropping this note here so that I (or another reviewer) doesn't forget about it.

Blindly trusting the input in this way scares me. Without looking at all into what `pSound->Play`, I hope it first uses some kind of decoder to verify that the sound is actually valid.

Dezash: I assumed that BASS_StreamCreateFile and BASS_MusicLoad do handle input properly but then again I can't be sure since the BASS library isn't open source.

AlexTMjugador: According to the [documentation](http://www.un4seen.com/doc/#bass/BASS_MusicLoad.html), both of those BASS functions can return a BASS_ERROR_FILEFORM error code if the file format is not recognised, so I agree with BASS verifying the memory buffer contents being a reasonable assumption. It'd definitely be scary if it didn't.

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -153,16 +153,24 @@ int CLuaAudioDefs::PlaySound(lua_State* luaVM)
             {
                 SString strFilename;
                 bool    bIsURL = false;
+                bool    bIsRawData = false;
                 if (CResourceManager::ParseResourcePathInput(strSound, pResource, &strFilename))
                     strSound = strFilename;
                 else
-                    bIsURL = true;
+                {
+                    SString signature = strSound.SubStr(0, 4).ToLower();
```
Citizen01: `SubStr(0, x)` is `Left(x)`
Also you can get rid of `.ToLower()` if you then use `stricmp` but I don't know if you should.

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -153,16 +153,24 @@ int CLuaAudioDefs::PlaySound(lua_State* luaVM)
...
+                    if ((memcmp(signature, "http", 4) == 0 || memcmp(signature, "ftp", 3) == 0)
+                        && (strSound.length() <= 2048 || strSound.find('\n') == SString::npos))
+                        bIsURL = true;
```
Citizen01: Hmmm so `httplmao` is considered as a valid url ?
From what I see, I assume we were already relying on the internals of **PlaySound** to have the last word on the validity of the input. Do we consider that check as "good enough" ? (I can't see it producing crashes from what the previous code looked like, I'm fine with that then)

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
Dezash: Done

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
AlexTMjugador: > Hmmm so `httplmao` is considered as a valid url ?

Just a candidate string for BASS to interpret it successfully as a URL. After all, it seems reasonable that, as you said, BASS should have the last word on its validity. Duplicating the checks that BASS already should do is a waste of time, and adds extra complexity if more URL types are to be fully parsed and checked for in the future.

Of course, I assume BASS is good enough that it reliably discards invalid input without crashing. Things like "httplmao" will be garbage no matter whether they are treated as URL or not.

qaisjp (approved):

PR #1236, opened 2020-01-28
Open Asset Import Library

CrosRoad95:
Extended version of my previous project with suggestion from Jusonex. It introduces not only .fbx but all popular model file types possible to load and display as regular object ( without collision ) without any gta limitations and conversion needed.

I created this draft pull request right now because it even is initial commit, will contain a lot of content i want to describe and can be hard to do it in one take, also it let me get feedback from you.

List of supported file types: http://assimp.sourceforge.net/lib_html/index.html
most common: .blend from blender, 3ds from 3ds max, .fbx from both and other modeling programs, .obj as model from plain text, .glb from paint3d

Test resource ( 28.01.2020 ), zip contains few test models and assimp viewer - standalone application to test models outside game
[assimp.zip](https://github.com/multitheftauto/mtasa-blue/files/4122713/assimp.zip)

Right now it provides new element

- `asset-model` - treat as asset-node ( root node ), contains all information about asset you loaded 
- `asset-node` - "sub model", 
- `asset-mesh` - 

Functions:
```lua
asset-model loadAssetModel( string rawData/file path, function callback, table options)
```
loads model, it is very recommended to provide callback, if provided, model will loaded at another thread without noticable lag
callback provides two argument: `bool loadedSuccessfully, string error`
options:
```lua
type | key | default value | meaning
bool "rawdata" = false -- indicate you provided raw data
string "hint" = "rawdata" -- append custom name to root node if raw data provided
float "globalScale" = 1 -- import scale
```

```lua
mixed getAssetProperties( element mixed, string key)
```
return verious information about all asset elements, some are supported only by some asset elements
keys:

- `animationscount` - how many animations contains asset-model
- `camerascount` - how many cameras contains asset-model
- `lightscount` - how many lights contains asset-model
- `materialscount` - how many materials contains asset-model
- `meshescount` - how many meshes contains asset-model or asset-node
- `texturescount` - how many textures contains asset-model
- `nodescount` - how many nodes contains asset-model
- `name` - gets name of asset-model ( root node ) or asset-node
- `position` - node offset position of parent node
- `rotation` - node offset rotation of parent node
- `scale` - node offset scale of parent node
- `boundingbox` - summed up all node meshes to return bbox
- `childnodescount` - gets number of child nodes
- `parentnode` - returns parent node, of asset-node, doesn't work for asset-model

```lua
table assetGetChilldrenNodes( asset-model / asset-node )
```
returns all child nodes, some file format support hierarchy

Planned functions: getting meshes and its data, creating instance of model, modyfing it, 

Screenshots
blend file:
image

raw data:
image

70MB .obj file with
image

Future plans:
1. Debugging features??
http://assimp.sourceforge.net/lib_html/class_assimp_1_1_logger.html
http://assimp.sourceforge.net/lib_html/class_assimp_1_1_importer.html#a9bb793072c84c784279d0f6e870bb42d
http://assimp.sourceforge.net/lib_html/class_assimp_1_1_importer.html#a6a4d830ffb3f77a3c7c919e0af006920
2. get/set raw data

Last update: 28.01.2020 - pr opened


Client/mods/deathmatch/logic/CClientAssetModel.cpp
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    CacheNodes(m_pScene->mRootNode);
+    m_bModelLoaded = true;
+    return "";
+}
+
+const char* CClientAssetModel::LoadFromFile(std::string strPath)
+{
+    m_pScene = importer.ReadFile(strPath, m_uiImportFlags);
```
CrosRoad95: should i put it into try catch block? in some cases it can crash, this part of code is executed at another thread

qaisjp: why does it crash? is it a memory issue? yes, suggest catching and returning an error string, would be ideal (if possible)

CrosRoad95: callback if second argument will == "crash", at third could return crash error string.
Model i used was .fbx file exported from 3ds max
image
image
image

qaisjp: yeah i'd let that crash MTA. does this happen in regular use?

CrosRoad95: assimp viewer crashes too. If you try to load some models they cause crash but in regular use with other models it doesn't crash

Einheit-101:
"Without GTA Limitations" - the only limitations are draw distance / streaming limit, am i right?

CrosRoad95:
> "Without GTA Limitations" - the only limitations are draw distance / streaming limit, am i right?

it completly bypass all gta limitations what cause only limit is how many resources has your pc

CrosRoad95:
~~should i use ogre as library for rendering? https://github.com/OGRECave/ogre
it supports directx9 which mta uses, and directx11 which saml1er going to implement.
both, ogre and assimp complement each other. i'm not going to reinventing rendering functions or copy fragments of other code~~

quiret:
Pretty interesting stuff. I think I will try this stuff once it is ready. I hope that your stuff will nicely integrate into the GTA:SA environment.

ghost:
> 
> should i use ogre as library for rendering? https://github.com/OGRECave/ogre

Yes, that would be nice.

CrosRoad95:
because idea has grown up, i come with better idea how make this feature even more powerful
few days ago i started working on implementing ogre ( https://www.ogre3d.org/ ) into mta, but i faced with multiple problems such a blending ogre rendering and gta, so i found i could just add option to switch between renderware and ogre.
Server could have set what graphics engine use, original renderware, or ogre.
because saml1er already going to implement directx11, and ogre supports both, directx9 and 11, should i make it at dx11 or 9?
or even better, should i use this https://github.com/Microsoft/DirectXTK/wiki/Getting-Started ?

with this feature, assimp will support only newer engine, and newer engine will came up with new set of functions for shaders, textures, objects, ect

Einheit-101:
An ogre switch sounds good, but can ogre use the .dff models and .fx shaders that renderWare uses or do we need to convert models and shaders then? A different engine is not a trivial thing.

CrosRoad95:
it can't use .dff just like that, but i can make function which convert .dff to ogre-friendly format,
about .fx, it supports HLSL as well just like mta right now http://wiki.ogre3d.org/HLSL but i don't know how it will work with directx11

prnxdev:
Can't wait for that. Nice work! 😊

PR #1247, opened 2020-02-10
Trailer sync fix

tederis:
This is a small fix brings us the temporary solution for a trailer's desync issue. It was tested well on a real project for about six months. It's not a comprehensive solution but fixes the problem at the moment. Video: https://youtu.be/iFClCa_xaG4


MegadreamsBE:
Does it also handle a trailer that is carrying a trailer itself?

tederis:
> Does it also handle a trailer that is carrying a trailer itself?

Haven't tested it. But I think it should.

qaisjp (approved):
Thank you, looks fine to me!

PlatinMTA:
Nice! Gonna check this on my server as soon as possible.

fresholia:
a

Client: Multi Theft Auto v1.5.7-release-20358.5.001 - Server: Multi Theft Auto v1.5.7-release-20411

When more than one player goes with a truck, it looks like this by other players.

dmi7ry:
> Client: Multi Theft Auto v1.5.7-release-20358.5.001 - Server: Multi Theft Auto v1.5.7-release-20411

It was fixed in r20402, so your 20358 has not the fix.

PlatinMTA:
> > Client: Multi Theft Auto v1.5.7-release-20358.5.001 - Server: Multi Theft Auto v1.5.7-release-20411
> 
> It was fixed in r20402, so your 20358 has not the fix.

AKA: The fix just changes clientside code, just update to the last nightly and you will see the trailers synced, at least it works just fine in my server 👍

Xenius97:
> When more than one player goes with a truck, it looks like this by other players.

This still an issue, trailers are broken.

PR #1254, opened 2020-02-21
Add isResourceProtected function

### LINKED ISSUE #1253: Add `isResourceProtected`
**Describe the bug**

We know you can protect your resources by active `protected` flag in [mtaserver.conf](https://wiki.multitheftauto.com/wiki/Server_mtaserver.conf#resource) 
so admins/mods cannot stop/restart it via `stop`/`restart` command,
but you can with [stopResource](https://wiki.multitheftauto.com/wiki/StopResource),[restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) functions
so admins/mods can stop it by resources like admin panel or other moderation resources
**To reproduce**

1. open admin panel
2. stop/restart a protected resource

**Expected behaviour**
[stopResource](https://wiki.multitheftauto.com/wiki/StopResource),[restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) 
mustn't stop/restart the resource because is protected
**Version**
MTA:SA Server v1.5.7-release-18957
**Additional context**
I have two ideas to solve the problem
1. if the resource is protected [stopResource](https://wiki.multitheftauto.com/wiki/StopResource),[restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) functions should return false
or 2. Add `isResourceProtected` function to prevent moderation resources from stop/restart protected resources

StrixG:
Closes #1253

### DIFF
--- a/Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
@@ -53,6 +53,7 @@ void CLuaResourceDefs::LoadFunctions()
         {"getResourceExportedFunctions", getResourceExportedFunctions},
         {"getResourceOrganizationalPath", getResourceOrganizationalPath},
         {"isResourceArchived", isResourceArchived},
+        {"isResourceProtected", isResourceProtected},
 
         // Set stuff
         {"setResourceInfo", setResourceInfo},
@@ -121,6 +122,7 @@ void CLuaResourceDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getState", "getResourceState");
     lua_classfunction(luaVM, "getACLRequests", "getResourceACLRequests");
     lua_classfunction(luaVM, "isArchived", "isResourceArchived");
+    lua_classfunction(luaVM, "isProtected", "isResourceProtected");
 
     lua_classvariable(luaVM, "dynamicElementRoot", NULL, "getResourceDynamicElementRoot");
     lua_classvariable(luaVM, "exportedFunctions", NULL, "getResourceExportedFunctions");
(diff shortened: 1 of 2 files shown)


ccw808:
'protected' seems slightly ambiguous. Should we make the name more indicative of the base functionality? e.g. isResourceStoppable or something better

StrixG:
> 'protected' seems slightly ambiguous. Should we make the name more indicative of the base functionality? e.g. isResourceStoppable or something better

I think isResourceStoppable would be more ambiguous than isResourceProtected. Because protected resource is stoppable, but only with stopResource, restartResource functions.

Dutchman101:
I don't think this is the right solution at all, the runcode resource (and admin panel code executor) would first need to check if the functions `stopResource` or `restartResource` are entered, and then perform the `isResourceProtected` check.. as admins can otherwise just run these functions (in a resource that can run code) without the checks. 
We can do it for default resources (runcode and admin panel) but otherwise that's a hassle, not all scripters are that intuitive (if we can do better by default, then why not..) and you could theoretically even implement runcode functionality through a runcode query, with the 'new' interface lacking these checks.

I believe that the mtaserver.conf value 'protected' should rather be authoritative.. just return false from any functions capable of manipulating the running state (`stopResource` and `restartResource`) and cancel it. Server owners setting this, from its naming since years, generally assume(d) it's safe without further action needed in their scripts.. if we just correct it (make it authoritative like it's supposed to be), an server update will solve it all at once, instead of the owner having to take notice and update scripts manually. That's another benefit of doing it how I consider it would be properly done..

qaisjp (approved):

qaisjp:
Hm, protection can indeed be slightly confusing but I can't think of something other than `stoppable`, which is also confusing to me.

Adding this function is fine because scripts should be able to get this attribute. Alternatively we could extend https://wiki.multitheftauto.com/wiki/GetResourceInfo to support `".protected"` as a key instead of adding another function?

See [`NameStartChar`](https://www.w3.org/TR/2006/REC-xml11-20060816/#NT-NameStartChar) — I think `.` is an invalid start character, so it could never appear in the meta XML file.

I'm not sure about whether scripts should be able to start or stop protected resources. But I generally find the existence of resource "protection" a little weird and it would be useful to get more info about why (what scenarios) people protect resources. Discussion about changing the behaviour is a separate issue though, since we'd still need `isResourceProtected`.

qaisjp:
My suggestion above about using `getResourceInfo` is kind of ridiculous - in the end there's no big deal in adding an extra function. Especially considering not that many resource-related functions have been added in the last ten years.

I'm going to merge this now - if we come up with a better name, we can change it

PR #1261, opened 2020-02-22
Fix #1258: Add mixed hex and RGB(A) parsing to XMLColorToInt

### LINKED ISSUE #1258: Add hex and RGB(A) support to all map color attributes
It seems various color formats like hex or RGB(A) aren't supported for many element types like `radararea` and `blip` but throw an error instead.

This is a breaking change. For example the race gamemode depends on this old behavior as described in #1393

Make color format support consistent across element types.

Use of a third-party script to convert unsupported color values in map files to a supported format.

**Additional context**
- Past attempt to fix: #1261

patrikjuvonen:
## Summary
`XMLColorToInt` is used when parsing i.e. `color` attribute from map files on various elements such as vehicles, blips, radarareas, teams and so on.

Currently it only supports a single format for multi-value attributes (i.e. vehicles have many colors) and many elements don't support hex color format right now anyway, so this PR adds consistent color attribute behavior across all element types.

So now it is possible to specify separate colors like so:
```
#ff0, 255, 0, 0, #ff0000, #ffffff55
```
Which corresponds to:
```
yellow, red, red, transparent white
```
And works straight for vehicles.

But most importantly all elements now have consistent `color` behavior, so for example blips, radarareas, teams can now be passed a hex color value and it will work.

This PR closes #1258.

## Tests
A test map file is available for download [here](https://github.com/multitheftauto/mtasa-blue/files/4239960/test.txt).


qaisjp (approved):
LGTM. :shipit:

Server/mods/deathmatch/logic/Utils.cpp
```diff
@@ -535,18 +535,104 @@ bool XMLColorToInt(const char* szColor, unsigned long& ulColor)
...
+    {
+        // Ambiguous value before a comma
+        SString strValue;
+        getline(ss, strValue, ',');
+
+        // Remove spaces
+        ReplaceOccurrencesInString(strValue, " ", "");
+
+        // Is the value looking like a hexadecimal?
+        if (strValue[0] == '#')
```
qaisjp, line 578: can strValue be empty?

patrikjuvonen: > can strValue be empty?

Yes, but there is a catch.
If you enter `255, , #f00`, it would read it as palette colors: 255 (doesn't exist), 0 (black), red (255, 0, 0).

I could introduce a fix on line 587 for this so that you cannot mix palette and RGB colors:
```c++
ucCount += ucLength - uiRGBAIndex;
```

With this fix you can do the above, but it will force the colors to RGB if they aren't complete sets. So, above example would become:
`255, , #f00` -> red (255, 0, 0), red (255, 0, 0).

I suppose this fix is a good idea. What do you think?

patrikjuvonen: And if you enter an empty hexadecimal value (a single `#` anywhere) it will error out in console:
`ERROR: Bad 'color' value specified in <vehicle> (line x)`.

qaisjp: so strValue will always have at least one char?

patrikjuvonen: Actually you know what, the fix makes most sense since this function is shared across elements. Will fix.

patrikjuvonen: > so strValue will always have at least one char?

Yup.

patrikjuvonen: Now it's been changed so that empty values just skip as RGB. This also changes it so that you can't have both palette and RGB in the same color string.

qaisjp (approved):
Still lgtm

PR #1265, opened 2020-02-28
Fix #359: engineReplaceModel memory leak

### LINKED ISSUE #359: engineReplaceModel memory leak
**Describe the bug**
Can not recover the memory taken by engineLoadDFF/engineRestoreModel/engineReplaceModel

**To Reproduce**
MemoryLeak1 shows destroying the dff element created by engineLoadDFF can not recover the memory.
MemoryLeak2 gives a increasing memory usage and it takes more memory with engineRestoreModel than engineLoadDFF only
```lua
----------------------------Memory Leak 1
function MemoryLeak1()
	function reloadDFF()
		if isElement(DFF) then
			destroyElement(DFF)
		end
		DFF = engineLoadDFF("rectangle.dff")
	end

	for i=1,10000 do
		reloadDFF()
	end
end
----------------------------Memory Leak 2
function MemoryLeak2()
	function reloadDFF()
		if isElement(DFF) then
			destroyElement(DFF)
		end
		DFF = engineLoadDFF("rectangle.dff")
	end
	
	addEventHandler("onClientPreRender",root,function()
		engineRestoreModel(3458)
	end)
	
	addEventHandler("onClientHUDRender",root,function()
		for i=1,50 do
			reloadDFF()
			engineReplaceModel(DFF,3458)
		end
	end)
end
```
Hex code of the dff i used
```Hex
1000000050020000FFFF0318010000000C000000FFFF03180100000000000000000000000E0000006C000000FFFF0318010000003C000000FFFF0318010000000000803F0000000000000000000000000000803F0000000000000000000000000000803F000000000000000000000000FFFFFFFF030002000300000018000000FFFF0318FEF253020C000000FFFF031852656374616E676C653030311A00000074010000FFFF03180100000004000000FFFF0318010000000F00000058010000FFFF03180100000098000000FFFF03187200000002000000040000000100000001000000000002000300020000000000000000000000000017B751383046E24001000000010000000000A0400000A040000000000000A0C00000A040000000000000A0C00000A0C0000000000000A0400000A0C00000000000000000000000000000803F00000000000000000000803F00000000000000000000803F00000000000000000000803F0800000054000000FFFF03180100000008000000FFFF031801000000FFFFFFFF0700000034000000FFFF0318010000001C000000FFFF0318000000000054FFFFDCA3FE00000000000000803F0000803F0000803F0300000000000000FFFF03180300000048000000FFFF03180E0500002C000000FFFF03180000000001000000060000000600000000000000000000000100000002000000020000000300000000000000FDF2530204000000FFFF0318000000001400000028000000FFFF03180100000010000000FFFF0318000000000000000005000000000000000300000000000000FFFF03180300000000000000FFFF0318
```
**Expected behavior**
engineRestoreModel will recover the memory
destroyElement will recover the memory taken by dff

 - Version 1.5.5
 - Running in Wine? No

ghost:
This memory leak comes from GTA SA for object models. It is located in `CFileLoader::SetRelatedModelInfoCB`. In GTA SA, models are not replaced, so the memory leak never happens there, but in MTA SA when we replace the atomic of the model, the original atomic is never deleted, and the memory usage keeps on getting higher and higher as object models are replaced and unloaded.

The reversed code of `CFileLoader::SetRelatedModelInfoCB` can be found [HERE](https://github.com/saml1er/gta-reversed/blob/3955cdda80e8c9c8b130c1f4ecf66c23f8fb4780/source/game_sa/CFileLoader.cpp#L59). I have plans to move my reversed code to MTA, but of course, we can't do it this way as it is ugly, and we haven't decided yet. Is the current code ok, or should I make it an asm hook for the time being? I can refactor it later when I start moving the reversed code.

Test resource: [leaktest.zip](https://github.com/multitheftauto/mtasa-blue/files/4269249/leaktest.zip)

commands: `leak1` and `leak2`. 

The testing code was provided by @thisdp.

### DIFF
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -1,0 +1,90 @@
+#include "StdInc.h"
+#include "gamesa_renderware.h"
+#include "CFileLoaderSA.h"
+
+CFileLoaderSA::CFileLoaderSA()
+{
+    InstallHooks();
+}
+
+CFileLoaderSA::~CFileLoaderSA()
+{
+
+}
+
+void CFileLoaderSA::InstallHooks()
+{
+    HookInstall(0x537150, (DWORD)CFileLoader_SetRelatedModelInfoCB, 5);
+}
+
+class CAtomicModelInfo
+{
+public:
+
+    void CAtomicModelInfo::DeleteRwObject()
+    {
+        ((void(__thiscall *)(CAtomicModelInfo*))(*(void ***)this)[8])(this);
+    }
+
+    void CAtomicModelInfo::SetAtomic(RpAtomic* atomic)
+    {
+        ((void(__thiscall *)(CAtomicModelInfo*, RpAtomic*))(*(void ***)this)[15])(this, atomic);
+    }
+
+};
+
+class CDamagableModelInfo
+{
+public:
+    void CDamagableModelInfo::SetDamagedAtomic(RpAtomic* atomic)
+    {
+        ((void(__thiscall *)(CDamagableModelInfo*, RpAtomic*))0x4C48D0)(this, atomic);
+    }
+};
+
+static char* GetFrameNodeName(RwFrame* frame) {
+    return ((char* (__cdecl*)(RwFrame*))0x72FB30)(frame);
+}
+
+static void GetNameAndDamage(char const* nodeName, char* outName, bool& outDamage) {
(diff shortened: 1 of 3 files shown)


qaisjp (approved):
I think it's fine.

Obviously using a hook would be cleaner right now, but I think what you've got right now is quite similar to the techniques used [here](https://github.com/saml1er/mtasa-blue/blob/engineReplaceModel-leak-fix/Client/game_sa/CControllerConfigManagerSA.cpp#L33-L44), except yours is better in that it doesn't use `_asm`.

Issue #756, opened 2019-01-04
Motion Blur into Video Settings


LosFaul:
**Is your feature request related to a problem? Please describe.**

A lot people hate motion blur and would like to have it disabled on any server. This would be usefull because not every server offers options for players to change their motion blur setting.

**Describe the solution you'd like**

There should be added a bar for the strengh of motion blur effect in video settings and a tick box which allows scripts to overwrite the motion blur value and resets motion blur value on server leave.
Bar:
0 |-----------| 100
tick box:
☐ Allow scripts to overwrite Motion Blur setting.

Related script function:
setPlayerBlurLevel / setBlurLevel
https://wiki.multitheftauto.com/wiki/SetPlayerBlurLevel

qaisjp:
I think servers should be in control of blur level, and there to just be a single checkbox "Enable blur" to set the default on-join (servers should be able to override this freely). Thoughts?

qaisjp:
A problem here is that there is no `resetBlurLevel` function, so existing servers will not be able to get the default. Perhaps we could add `resetBlurLevel` (and then also add a blur level bar, as originally described in the issue).

Blur level could be updated whilst ingame, but only apply if the blur level is currently "reset". That way servers can use resetBlurLevel to relinquish control to the user's preference, or use `setBlurLevel(defaultValue)` to force it to a certain value.

Guillaume57:
The blur level is restored to its default value when starting a game (joining a server or map editor), so why need resetBlurLevel function ? If server want to reset player's default blur level, it can first save the value of getPlayerBlurLevel when player join and then setPlayerBlurLevel to this value to reset it.

qaisjp:
Wait. Do we really need `resetBlurLevel` if we had something like `getClientSetting` https://github.com/multitheftauto/mtasa-blue/issues?q=is%3Aissue+sort%3Aupdated-desc+getClientSetting

Sorry for kind of breaking #1266

ghost:
Well, i added `resetBlurLevel` since i felt like it was kinda missing. I also added it cause i thought about `resetBlurLevel` to reset to be used in this PR to reset blur level to clients preference or default GTA value.

Quite unsure about `getClientSetting` tbh

PR #1266, opened 2020-03-01
Add clientside lua function resetBlurLevel

ghost:
This PR adds clientside method named `resetBlurLevel` which is a starting point for https://github.com/multitheftauto/mtasa-blue/issues/756


Woovie (approved):

Woovie:
Any chance we could get resetPlayerBlurLevel as well?

ghost:
> Any chance we could get resetPlayerBlurLevel as well?

Do you mean for server-sided scripts ? I did add this method mainly cause of the above mentioned issue regarding blur level setting. Plus: all (?) `reset*` methods are clientside only.

Woovie:
Yes, server-side too. Nothing about the original issue to me indicates it should be a client method only. Correct me if I'm wrong.

ghost:
Well, no, it didnt indicate that it's meant to be clientside only but i thought it since like i said all `reset*` methods are clientside.

Gonna add `resetPlayerBlurLevel` clientside which is an alias to `resetBlurLevel`.

Imo, we dont need `resetPlayerBlurLevel` on server-side since the server is not responsible for visual stuff unless it's required when it comes to sync (such as weather, time, minute duration)

LosFaul:
resetBlurLevel on server side doesn't make much sence since it's only a visual effect

qaisjp:
why is `resetPlayerBlurLevel` added as just an alias?

ghost:
> why is `resetPlayerBlurLevel` added as just an alias?
Made it an alias because the logic behind it is the same. Kinda hate duplicating code.

qaisjp:
Why do we need both?

ghost:
> Why do we need both?

I dont really think that we need it but @Woovie wanted to have it added as you can see above.

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -274,7 +274,6 @@ void CLuaManager::LoadCFunctions()
         {"getGaragePosition", CLuaFunctionDefs::GetGaragePosition},
         {"getGarageSize", CLuaFunctionDefs::GetGarageSize},
         {"getGarageBoundingBox", CLuaFunctionDefs::GetGarageBoundingBox},
-        {"getBlurLevel", CLuaFunctionDefs::GetBlurLevel},
```
qaisjp: nit: gets and sets are supposed to be placed separately

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.World.cpp
```diff
@@ -549,6 +549,13 @@ int CLuaFunctionDefs::SetBlurLevel(lua_State* luaVM)
     return 1;
 }
 
+int CLuaFunctionDefs::ResetBlurLevel(lua_State* luaVM)
+{
+    g_pGame->SetBlurLevel(static_cast<unsigned char>(36));
```
qaisjp:
```suggestion
    g_pGame->SetBlurLevel(36);
```
just 36 should just work here

ghost:
done @qaisjp

qaisjp (approved):
:shipit:

qaisjp:
sorry, please see https://github.com/multitheftauto/mtasa-blue/issues/756#issuecomment-599557039

PR #1272, opened 2020-03-03
Add `capslock` & `numlock` to `getKeyState`

### LINKED ISSUE #1271: Add functions to check for capslock & numlock key states
We've currently no way of finding out wheter capslock or numlock are active or not. It could be useful for e.g show drawings when caps lock are active.

Add the following functions (client only):
`bool isCapsLockEnabled()`
`bool isNumLockEnabled()`

**Additional context**
I wanted to add those functions but i'm unsure about naming - any ideas ?

ghost:
This issue closes https://github.com/multitheftauto/mtasa-blue/issues/1271

Test script used:

```lua
	local caps = getKeyState("capslock")
	local caps2 = getKeyState("capslock", true)
	
	local num = getKeyState("numlock")
	local num2 = getKeyState("numlock", true)
	
	local scroll = getKeyState("scroll")
	local scroll2 = getKeyState("scroll", true)
	
	local key = getKeyState("a")
	local key2 = getKeyState("a", true)
	
	dxDrawText(("caps: %s - %s\nnum: %s - %s\nscroll: %s - %s\nkey: %s - %s"):format(tostring(caps), tostring(caps2), tostring(num), tostring(num2), tostring(scroll), tostring(scroll2), tostring(key), tostring(key2)), 250, 250)
end)
```


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6972,16 +6972,22 @@ bool CStaticFunctionDefinitions::GetKeyState(const char* szKey, bool& bState)
...
-            bState = (::GetKeyState(pKey->ulCode) & 0x8000) ? true : false;
-
-            return true;
-        }
+        DWORD dwFlags = (keyCode == VK_CAPITAL || keyCode == VK_NUMLOCK) ? 0x0001 : 0x8000;
+        bState = (::GetKeyState(keyCode) & dwFlags) ? true : false;
+        return true;
     }
-
+
```
qaisjp:
```suggestion

```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6972,16 +6972,22 @@ bool CStaticFunctionDefinitions::GetKeyState(const char* szKey, bool& bState)
...
+    
+    if (g_pCore->IsFocused() && keyCode > 0)
     {
-        if (g_pCore->IsFocused())
-        {
-            bState = (::GetKeyState(pKey->ulCode) & 0x8000) ? true : false;
-
-            return true;
-        }
+        DWORD dwFlags = (keyCode == VK_CAPITAL || keyCode == VK_NUMLOCK) ? 0x0001 : 0x8000;
```
qaisjp:
```suggestion
        DWORD dwFlags = pKey ? 0x8000 : 0x1;
```

qaisjp (changes requested):
thanks

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
ghost: Your change wouldnt work since pKey is always set for `capslock` (It's added to bindable keys)

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6972,16 +6972,21 @@ bool CStaticFunctionDefinitions::GetKeyState(const char* szKey, bool& bState)
 
     CKeyBindsInterface* pKeyBinds = g_pCore->GetKeyBinds();
     const SBindableKey* pKey = pKeyBinds->GetBindableFromKey(szKey);
+    unsigned long keyCode = 0;
+
     if (pKey)
+        keyCode = pKey->ulCode;
+    else if (strcmp(szKey, "capslock") == 0)
+        keyCode = VK_CAPITAL;
```
qaisjp: since `VK_CAPITAL` is a bindable key, will this ever execute?

xLuxy: yeah, it does

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
qaisjp: Wait — does `0x1` return the system-wide toggle state, and `0x8000` the current keydown state? 

If that is the case, we cannot make this change and have to do something different

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
qaisjp: Why? `capslock` is listed here https://wiki.multitheftauto.com/wiki/Key_names and you've stated it's a bindable key, so this should not get executed.

edit: this was in response to a comment deleted by @xLuxy, "yeah it does"

ghost: Yeah, just seen, sec

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
ghost: How would you do it differently ? I mean you can check whatever they're pressed or not bot i was talking about the system wide toggle state.

thisdp:
Will this affect the existing key state check of "capslock"? I mean whether you hold it or release..

ghost:
> Will this affect the existing key state check of "capslock"? I mean whether you hold it or release..

Yes, it will. We could introduce an extra argument to the function like `getKeyState(string key[, bool toggle = false])` to overcome this. Opinions ?

qaisjp:
`getKeyState(string key[, bool toggle = false])` sounds good to me

ghost:
Added toggle parameter. I tested both flags (0x8000 and 0x1) against other keys such as A, G and so on and they all worked even with 0x1 set. Keeping it there because i dont know how it behaves on other pcs.

Changed test script:
```lua
	local caps = getKeyState("capslock")
	local caps2 = getKeyState("capslock", true)
	
	local num = getKeyState("numlock")
	local num2 = getKeyState("numlock", true)
	
	local scroll = getKeyState("scroll")
	local scroll2 = getKeyState("scroll", true)
	
	local key = getKeyState("a")
	local key2 = getKeyState("a", true)
	
	dxDrawText(("caps: %s - %s\nnum: %s - %s\nscroll: %s - %s\nkey: %s - %s"):format(tostring(caps), tostring(caps2), tostring(num), tostring(num2), tostring(scroll), tostring(scroll2), tostring(key), tostring(key2)), 250, 250)
end)
```

qaisjp:
I am thinking we just make `numlock` a bindable key instead of making that key only work with `getKeyState`. Since `"scroll"` is accepted on [WIKI-Key_names](https://wiki.multitheftauto.com/wiki/Key_names), there's no harm for `"numlock"` to be supported too.

I also tried to find out what the current `"scroll"` key name _actually_ means. See: https://github.com/multitheftauto/mtasa-blue/blob/d5722d5ac5c0ed0210849bf03c263ec88ea98c2a/Client/core/CKeyBinds.cpp#L110

`GTA_KEY_SCROLL` refers to some GTA-specific magic number, but 0x91 corresponds to [`VK_SCROLL`](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes).

It shouldn't be too difficult to add support for `VK_NUMLOCK` but, if you don't want to make `numlock` a bindable key right now, we can just leave numlock support out of this PR and do it another time.

thisdp:
lol scroll lock does even not exist on my keybord

ghost:
> I am thinking we just make `numlock` a bindable key instead of making that key only work with `getKeyState`. Since `"scroll"` is accepted on [WIKI-Key_names](https://wiki.multitheftauto.com/wiki/Key_names), there's no harm for `"numlock"` to be supported too.
> 
> I also tried to find out what the current `"scroll"` key name _actually_ means. See:
> 
> https://github.com/multitheftauto/mtasa-blue/blob/d5722d5ac5c0ed0210849bf03c263ec88ea98c2a/Client/core/CKeyBinds.cpp#L110
> 
> `GTA_KEY_SCROLL` refers to some GTA-specific magic number, but 0x91 corresponds to [`VK_SCROLL`](https://docs.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes).
> 
> It shouldn't be too difficult to add support for `VK_NUMLOCK` but, if you don't want to make `numlock` a bindable key right now, we can just leave numlock support out of this PR and do it another time.

Tried to add
```cpp
{"numlock", 0x9, NO_KEY_DEFINED, NO_DATA, 0}
```
but it somehow doesnt seem to be working. I'm also unable to use breakpoints to debug parts of the code - it keeps saying that no debug symbols have been loaded - might be due an old VS2019 installation. Gonna update and see if i'm able to debug this afterwards.

qaisjp:
It needs to be 0x90 not 0x9

qaisjp (changes requested):
Your problem will be fixed when you:

> use 0x90 not 0x9

xLive:
What will happen with this PR? since the owner is a @ghost

qaisjp:
Someone will need to resubmit a new PR

PR #1277, opened 2020-03-05
Fix duplicate enum usage in #1177 (NVidia Optimus Dialog)

Lpsd:
This fixes a small issue with PR #1177 (merged) where an enum was being used twice for two different labels.


qaisjp (approved):

qaisjp:
thanks

Issue #1151, opened 2019-11-11
Strange behavior of the Skimmer (e.g. crashes on approach)


Rob3681:
**Describe the bug**
- When approaching a Skimmer, the game can freeze or crash.
- When approaching a Skimmer, the Skimmer may or may not be invisible.
- When entering such an invisible Skimmer, the screen flickers very strong.

**Version**
Multi Theft Auto v1.5.7-release-20344

saesrpgmta:
I'm glad you reported this, our server has been experiencing client crashes lately and couldn't put our finger on it, so we just removed all skimmers on the server with runcode and now, no one is crashing.

Server and clientmin both Multi Theft Auto v1.5.7-release-20344

ccw808:
Please try with latest client nightly

Rob3681:
The problem still occurs in v1.5.7-release-20347.

ccw808:
Please try 20348

Rob3681:
It seems that it is now fixed.

saesrpgmta:
Can confirm that it is also fixed on our server with the 20348 build

Dutchman101:
This is the result of undefined behavior somewhere, to which vs2019 is less tolerant due to stricter conformance demands (which is why vs2019 was reverted for now in f22306b)

So far we have only been able to confirm Skimmers always being invisible, and then when you press enter near it (which would make you enter it instantly without anim, due to specialised Skimmer entering handling in MTA) the game will glitch out by screen becoming white and recursing contrast. 

(yes, the thing that's also considered as a very rare GTA bug: 

> sbx320 1:06 AM
> screen turning white is usually caused by some NaN / Inf crap propagating through everything
> 
> dutchman101 1:07 AM
> the same thing as can happen randomly (but is very rare) like when you're flying a hydra out of bounds, with the game tabbed out
> an infinite loop..

> sbx320 1:08 AM
> well the issue is related to something going to NaN
> and NaN + anything is NaN, same with the other operations
> so it just propagates through the entire code that does anything with floating point numbers

[...]

PR #1285, opened 2020-03-09
Disable Skimmer

ghost:
Temporary "fix" for #1151. ~~I'm still able to create skimmer in map editor. Does anybody know why?~~


theSarrum:
I think this would break existing scripts.

ccw808:
What does "Temporary "fix"" mean?

Lpsd:
I don't think this is needed, why would we completely disable the Skimmer? The next step after this "temporary fix" is fixing the glitch, so I don't see the purpose of this.

ghost:
> 
> 
> What does "Temporary "fix"" mean?

I should've explained it better. It's not a fix. We're just removing the skimmer completely. I don't know how to reproduce the bug. By removing skimmer, it will avoid the client from crashing when approaching it. Maybe we should just revert back to VS2017?

qaisjp:
Technically speaking, neither reverting or removing the skimmer are 'fixes'.

Reverting back to vs2017 is probably more preferable in _getting 'nightly releases' suitable for public use_, so we can keep developing elsewhere.

See further info in https://github.com/multitheftauto/mtasa-blue/issues/1151#issuecomment-597128153

PR #1215, opened 2020-01-18
Add functions to manipulate colshapes parameters

StrixG:
- [x] **Radius** *(circle, sphere, tube)*
```lua
float getColShapeRadius(colshape colShape)
bool setColShapeRadius(colshape colShape, float radius)
```
- [x] **Size**

**Tube (height)**
```lua
float getColShapeSize(colshape colShape)
bool setColShapeSize(colshape colShape, float size)
```
**Rectangle (width, height)**
```lua
float, float getColShapeSize(colshape colShape)
bool setColShapeSize(colshape colShape, Vector2 size)
bool setColShapeSize(colshape colShape, float width, float height)
```
**Cuboid (width, depth, height)**
```lua
float, float, flaot getColShapeSize(colshape colShape)
bool setColShapeSize(colshape colShape, Vector3 size)
bool setColShapeSize(colshape colShape, float width, float depth, float height)
```
- [x] **Polygon points**
```lua
table getColPolygonPoints(colshape colPolygon)
Vector2 getColPolygonPointPosition(colshape colPolygon, int pointIndex)
bool setColPolygonPointPosition(colshape colPolygon, int pointIndex, Vector2 position)
bool addColPolygonPoint(colshape colPolygon, Vector2 position[, int pointIndex])
bool removeColPolygonPoint(colshape colPolygon, int pointIndex)
```

PR #1287, opened 2020-03-10
Fix signed/unsigned mismatch warnings introduced in #1215

StrixG:


Client/mods/deathmatch/logic/CClientColPolygon.cpp
```diff
@@ -69,28 +69,20 @@ void CClientColPolygon::SetPosition(const CVector& vecPosition)
...
+    CalculateRadius(vecPoint);
+    return true;
+}
 
-    CVector2D vecDistance = vecPoint - m_vecPosition;
-    float     fDistance = vecDistance.Length();
+bool CClientColPolygon::AddPoint(CVector2D vecPoint, unsigned int uiPointIndex)
+{
+    if (uiPointIndex > m_Points.size())
+        return false;
```
qaisjp, line 82: Does this now need to return false if `uiPointIndex < 0`?

Server/mods/deathmatch/logic/luadefs/CLuaColShapeDefs.cpp
```diff
@@ -627,14 +627,21 @@ int CLuaColShapeDefs::AddColPolygonPoint(lua_State* luaVM)
...
+        if (iPointIndex == 0)
         {
-            lua_pushboolean(luaVM, true);
+            lua_pushboolean(luaVM, CStaticFunctionDefinitions::AddColPolygonPoint(pColPolygon, vecPoint));
         }
         else
         {
-            m_pScriptDebugging->LogWarning(luaVM, "Invalid point index");
-            lua_pushboolean(luaVM, false);
+            if (CStaticFunctionDefinitions::AddColPolygonPoint(pColPolygon, iPointIndex - 1, vecPoint))
```
qaisjp: I think this can be collapsed into a single `if / else if / else` construct

Client/mods/deathmatch/logic/CClientColPolygon.cpp
ghost: The type of `uiPointIndex` is unsigned int. Let's say if the value of `uiPointIndex` is -1 when calling AddPoint method, in hex, it will be 0xFFFFFFFF, so the logic is correct. 🙂

qaisjp (approved):

Client/mods/deathmatch/logic/rpc/CColShapeRPCs.cpp
```diff
@@ -55,17 +55,24 @@ void CColShapeRPCs::UpdateColPolygonPoint(CClientEntity* pSource, NetBitStreamIn
 void CColShapeRPCs::AddColPolygonPoint(CClientEntity* pSource, NetBitStreamInterface& bitStream)
 {
     SPosition2DSync size(false);
-    int             iPointIndex;
-    if (bitStream.Read(&size) && bitStream.Read(iPointIndex))
```
qaisjp, line 59: I suspect polygon point update function calls will now fail against older clients, but I think it's mostly ok because this is a beta feature anyway.

PR #1290, opened 2020-03-13
Fix #1279: driveby for peds

### LINKED ISSUE #1279: Fix driveby for peds
**Describe the bug**
**Primary issue**
No control over what the ped shoots at in driveby.
Using setPedDoingGangDriveby, "vehicle_fire" control state and setPedAimTarget the ped does aim at something but he shoots at a completely different location.

See video: https://streamable.com/vh0bh
The ped aim target is set to local player target end.

The white line is from ped position to his target end, which never seems to update.
The red line is from ped weapon muzzle to the aim target. Which is the expected bullet trajectory.

Also tried with different weapons and vehicles and seats.

**Secondary issue**
The ped is very limited in his movements in driveby, see below screenshots. He can only aim right or forwards compared to player. In single player peds seem to be able to aim like the player does here.
mta-screen_2020-03-05_22-08-20
mta-screen_2020-03-05_22-08-12

**To reproduce**
[drivebytest.zip](https://github.com/multitheftauto/mtasa-blue/files/4295205/drivebytest.zip)

**Expected behaviour**
The ped should shoot where he's aiming.

**Version**
1.5.7

**Additional context**

Zangomangu:
Fixes #1279

- Driveby shooting didn't work for peds because ped target vectors are not updated (Ped::Interpolate) when he's in a vehicle. This is fixed by using Ped::SetAimingData instead of Ped::SetTargetTarget.

- All 4 driveby aim animations now work.

- Ped target origin is currently ped position instead of gun muzzle, so bullets appear to shoot from center of ped. I've changed it to gun muzzle, I don't think this should break any scripts?

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -2444,23 +2444,31 @@ bool CStaticFunctionDefinitions::SetPedAimTarget(CClientEntity& Entity, CVector&
                 fRotDiff = fRotDiff + PI * 2;
             }
 
-            // Find the aim anim and correct fArmX/fArmY (if not default anim)
+            // Find the aim anim and correct fArmX/fArmY
             if (fRotDiff > PI * 0.25 && fRotDiff < PI * 0.75)
             {
+                // Facing left
                 cInVehicleAimAnim = 1;
                 fArmX = fArmX - PI / 2;
                 fArmY = -fArmY;
             }
             else if (fRotDiff > PI * 0.75 || fRotDiff < -PI * 0.75)
             {
+                // Facing backwards
                 cInVehicleAimAnim = 2;
                 fArmX = fArmX + PI;
                 fArmY = -fArmY;
             }
             else if (fRotDiff < -PI * 0.25 && fRotDiff > -PI * 0.75)
             {
+                // Facing right
                 cInVehicleAimAnim = 3;
                 fArmX = fArmX + PI / 2;
+            }
+            else
+            {
+                // Facing forwards
+                // Do nothing, initial values are fine
             }
 
             // Set aim and target data without interpolation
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2402,21 +2402,74 @@ bool CStaticFunctionDefinitions::SetPedAimTarget(CClientEntity& Entity, CVector&
...
+            if (fRotDiff > PI)
+            {
+                fRotDiff = -(PI - (fRotDiff - PI));
+            }
+            else if (fRotDiff < -PI)
+            {
+                fRotDiff = fRotDiff + PI * 2;
+            }
+
+            // Find the aim anim and correct fArmX/fArmY (if not default anim)
```
qaisjp: Does "if not default anim" refer to the initial value of `cInVehicleAimAnim`? Might be worth rephrasing this to "if not facing forwards".

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2402,21 +2402,74 @@ bool CStaticFunctionDefinitions::SetPedAimTarget(CClientEntity& Entity, CVector&
...
         // Arm direction
         float fArmX = -atan2(vecTarget.fX - vecOrigin.fX, vecTarget.fY - vecOrigin.fY),
               fArmY = -atan2(vecTarget.fZ - vecOrigin.fZ, DistanceBetweenPoints2D(vecTarget, vecOrigin));
 
-        // TODO: use gun muzzle for origin
-        Ped.SetTargetTarget(TICK_RATE, vecOrigin, vecTarget);
-        Ped.SetAim(fArmX, fArmY, 0);
+        if (Ped.IsInVehicle())
+        {
+            // Driveby aim animation
```
qaisjp, line 2426:
```suggestion
            // Driveby aim animation
            // 0 = forwards, 1 = left, 2 = back, 3 = right
```

qaisjp (approved):
Left a few minor comments, but it works and the code looks good.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Zangomangu: Yeah makes sense

PR #1295, opened 2020-03-15
Fix #1294: enable enter_exit for peds to fix alternative attack

### LINKED ISSUE #1294: Alternative attack for peds doesnt work
**Describe the bug**
Can't trigger alternative attacks for peds
The alternative attack is dependent on the fighting style (setPedFightingStyle).

**To reproduce**
Set control state aim_weapon true.
Set control state enter_exit true.

**Expected behaviour**
Expected ped to do the same as local player.

**Screenshots**

**Version**
v1.5.7

**Additional context**

Zangomangu:
This PR simply enables enter_exit control for peds on foot, which allows them to perform alternative attacks.

Tested with vehicles around to make sure nothing funky happens.


qaisjp (approved):
thank you!

> Tested with vehicles around to make sure nothing funky happens.

please can you confirm this means that peds don't enter vehicles? from my tests it doesn't seem that they do

Zangomangu:
Yeah, they don't enter vehicles.

PR #1298, opened 2020-03-16
Download CEF from opensource.spotify.com

qaisjp:
New old builds should be fine because Spotify hosts CEF archives for years:

1. http://opensource.spotify.com/cefbuilds/index.html
2. wait a couple seconds for the entire page to load, javascript will prettify the page (yes actually wait)
3. scroll to "Windows 32bit" and click 'Show more builds'
4. oldest build available is `05/26/2016 - CEF 3.2704.1414.g185cd6c / Chromium 51.0.2704.47`

Might be worth adding the sha1 to the config section, so that someone can't change the binaries underneath us (and spotify) and replace them with malicious files. Let me know if I should do this.


AlexTMjugador:
If malicious builds are a concern, and the hash expected life time is high, then I think that we should be using SHA256 instead of SHA1 for checking for file integrity. See: [https://en.wikipedia.org/wiki/File_verification#File_formats](https://en.wikipedia.org/wiki/File_verification#File_formats)

jushar:
If we use SHA256 instead of SHA1, could you also add a build action that downloads the archive, calculates the hash and prints it to the console?

E.g. something like
```
premake5 calc-cef-hash "80.0.8+gf96cd1d+chromium-80.0.3987.132"
```

Or alternatively, we could also automate the update process entirely by writing the url + hash to an additional file (e.g. a json file or something else premake can parse) and use that for downloading CEF.

Other than that, the patch looks fine to me and is ready for merging.

jushar (approved):

qaisjp:
sorry i didn't read, i'll do that now

qaisjp:
i'll do that now

qaisjp:
requested feature done in #1332

PR #1324, opened 2020-03-26
Fix crash when we can't connect to Discord

qaisjp:
urgent - please merge today if you're happy with the code


patrikjuvonen (approved):
Looks good 👍

PR #889, opened 2019-04-25
setWaterLevel: added bIncludeWorldSeaLevel and bIncludeOutsideWorldLevel

TheNormalnij:
This makes it possible to change the water level outside the game world.

PR #1342, opened 2020-04-02
setWaterLevel: added bIncludeWorldSeaLevel and bIncludeOutsideWorldLevel

TheNormalnij:
Recreated  #889


Zangomangu:
I've tested this and it works well.
It keeps default behaviour so there are no issues with existing scripts.

ghost (approved):

qaisjp:
@saml1er we are on a feature freeze :(

ghost:
Oh shit. Revert?

qaisjp:
Yes please - I don't think there is enough time for people to use and test `setWaterLevel`.

Also, @sbx320 asked this question in the original PR, which I do not think was answered:

> Is there any particular reason why we'd not just apply it to everything anyway? I think setWaterLevel not applying outside world boundaries is more of a bug rather than a feature.

ffsPLASMA:
This feature would become handy for some sort of tsunami scenario.

TheNormalnij:
@qaisjp, with this way you can replace all water and replace water level seamlessly with new water. Also this is workaround for  some streamer bugs when all map is replaced

qaisjp:
sbx's comment is not talking about rejecting the PR, it's talking about making setWaterLevel just always work for all water and not needing to specifically set `bIncludeOutsideWorldLevel`.

Why would you ever set `bIncludeOutsideWorldLevel` to false?

TheNormalnij:
To keep old behaviour

qaisjp:
It's a bug, not a feature. I can't think of a reason why you'd ever want that old (broken) behaviour.

ffsPLASMA:
Map scripts that use setWaterLevel might get broken, thats why we should keep the broken way for legacy reasons.

TheNormalnij:
Some [scripts](https://forum.mtasa.com/topic/101142-jsconverted-maps/) will broken if we change behaviour

TheNormalnij:
Reopen?

qaisjp:
Yeah github doesn't support reopening of merged PRs. please can you submit another one 🙃

PR #1372, opened 2020-04-11
Fix #1362: xmlLoadString doesn't work

### LINKED ISSUE #1362: xmlLoadString doesn't work
**Describe the bug**
xmlLoadString doesn't work on client or server atm.
It yields "Unable to load XML string".

**To reproduce**
[test.zip](https://github.com/multitheftauto/mtasa-blue/files/4455090/test.zip)

**Version**
Both client and server
Latest 20502
Doesn't work in 20447 either.
Windows 10

EDIT:
Tested in 19626 where it was added, doesnt work there either.

qaisjp:
Fixes #1362.

The change in `CXMLImpl::ParseString` allows `<animals> <f></f> <g/> </animals>` to work.

The change to `CXMLImpl::BuildNode` allows `<f>hi</f>` to work.

There is probably a better fix. I guess we are creating our XML node the wrong way - otherwise opening normal XML files would maybe be broken.

<details><summary>Tested with this code</summary>

```lua
function print_tree(x)
	print(x.name, x.value)
	for i, n in ipairs(x.children) do
		print(i)
		print_tree(n)
	end
end

local rootNode = xmlLoadString("<animals> <f>hi</f> <g/> </animals>")
if rootNode then
	print("Success")
	print_tree(rootNode)
else
	print("RIP")
end
```

</details>

In resource form: [x.zip](https://github.com/multitheftauto/mtasa-blue/files/4465204/x.zip)

### DIFF
--- a/Shared/XML/CXMLImpl.cpp
+++ b/Shared/XML/CXMLImpl.cpp
@@ -46,7 +46,8 @@ CXMLNode* CXMLImpl::ParseString(const char* strXmlContent)
     TiXmlDocument* xmlDoc = new TiXmlDocument();
     if (xmlDoc)
     {   
-        if (xmlDoc->Parse(strXmlContent, 0, TIXML_ENCODING_UTF8))
+        xmlDoc->Parse(strXmlContent, 0, TIXML_ENCODING_UNKNOWN);
+        if (!xmlDoc->Error())
         {
             TiXmlElement* xmlDocumentRoot = xmlDoc->RootElement();
             CXMLNodeImpl* xmlBaseNode = new CXMLNodeImpl(nullptr, nullptr, *xmlDocumentRoot);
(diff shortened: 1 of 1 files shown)


Shared/XML/CXMLImpl.cpp
```diff
@@ -46,7 +46,8 @@ CXMLNode* CXMLImpl::ParseString(const char* strXmlContent)
     TiXmlDocument* xmlDoc = new TiXmlDocument();
     if (xmlDoc)
     {   
-        if (xmlDoc->Parse(strXmlContent, 0, TIXML_ENCODING_UTF8))
+        xmlDoc->Parse(strXmlContent, 0, TIXML_ENCODING_UNKNOWN);
```
qaisjp: Actually the change from `UTF8` to `UNKNOWN` is not necessary. But I suppose our scripters will appreciate us not assuming the input is UTF8.

Lpsd (approved):
Looks good to me

Shared/XML/CXMLImpl.cpp
Lpsd: Our file implementation uses `TIXML_DEFAULT_ENCODING` - maybe follow this for consistency?

qaisjp: wtf is this garbage
image

qaisjp:
[Oh, you wrote `CXMLImpl::BuildNode`](https://github.com/multitheftauto/mtasa-blue/commit/3b2ed1caceacb84ad8635fab37fc0c0221d976c2). I thought it was ancient code. Then I think this fix is probably fine then.

PR #1380, opened 2020-04-12
Add getElementCount function

CrosRoad95:
Just adds function `int getElementCount(element-type)` client and server side, count types: `player,ped,water,vehicle,object,pickup,marker,colshape,blip,radararea,team`. Doesn't loop through all elements, just return size of list. it replaces `getMarkerCount` and  `getPlayerCount`

Test resource: compare result from getElementsByType and my function for all types client and serverside
[count.zip](https://github.com/multitheftauto/mtasa-blue/files/4466952/count.zip)

In depracated list i don't know what number should i set, and did i set this list properly


Einheit-101:
i just use #getElementsByType("whatever"), isnt that good enough?

Kenix157:
> i just use #getElementsByType("whatever"), isnt that good enough?

C++ loop is faster than lua, also it replaces getMarkerCount and getPlayerCount to a single one.

CrosRoad95:
#getElementsByType loop all elements, this function juz return single number

Pirulax:
> > i just use #getElementsByType("whatever"), isnt that good enough?
> 
> C++ loop is faster than lua, also it replaces getMarkerCount and getPlayerCount to a single one.



Creating a new Lua table every time you want to count the number of elements is a waste, considering it takes a fair amount of time to push all the data into the table.

This function is probably way faster, it also replaces 2 other functions, which is great!

Server/mods/deathmatch/logic/CResourceChecker.Data.h
```diff
@@ -295,6 +295,7 @@ namespace
         {"getRemoteRequests", "1.5.7-9.20307"},
         {"getRemoteRequestInfo", "1.5.7-9.20307"},
         {"abortRemoteRequest", "1.5.7-9.20307"},
+        {"getElementCount", "1.5.7-9.99999"}, // idk what number should be here
```
Pirulax, line 298: To anyone who merges this: Please update this.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -834,6 +836,12 @@ int CLuaElementDefs::GetElementsByType(lua_State* luaVM)
     return 1;
 }
 
+int CLuaElementDefs::GetElementsCount(lua_State* luaVM)
+{
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
+
```
Pirulax: Why would you do this? Why dont you just register this Lua function as `getElementCount`? I feel like silently pushing a false is a very bad idea.

Shared/mods/deathmatch/logic/Enums.cpp
```diff
@@ -166,3 +166,38 @@ ADD_ENUM1(PACKET_ID_CHAT_CLEAR)
...
+ADD_ENUM(EElementCountType::ELEMENT_WATER, "water")
+ADD_ENUM(EElementCountType::ELEMENT_VEHICLE, "vehicle")
+ADD_ENUM(EElementCountType::ELEMENT_OBJECT, "object")
+ADD_ENUM(EElementCountType::ELEMENT_PICKUP, "pickup")
+ADD_ENUM(EElementCountType::ELEMENT_MARKER, "marker")
+ADD_ENUM(EElementCountType::ELEMENT_COLSHAPE, "colshape")
+ADD_ENUM(EElementCountType::ELEMENT_BLIP, "blip")
+ADD_ENUM(EElementCountType::ELEMENT_RADARAREA, "radararea")
+ADD_ENUM(EElementCountType::ELEMENT_TEAM, "team")
+IMPLEMENT_ENUM_END("element-type")
```
Pirulax, line 203: I think you could use the existing 'CElement::EElementType` here instead?

Shared/mods/deathmatch/logic/Enums.h
```diff
@@ -68,3 +68,21 @@ DECLARE_ENUM_CLASS(PasswordHashFunction);
 DECLARE_ENUM_CLASS(StringEncryptFunction);
 
 DECLARE_ENUM(ePacketID);
+
+enum EElementCountType
```
Pirulax, line 72: "Prefer `enum class` over `enum`. If you use `enum class`, remove the `ELEMENT_` prefix.

Client/mods/deathmatch/logic/CClientColManager.h
```diff
@@ -29,6 +29,9 @@ class CClientColManager
     bool Exists(CClientColShape* pShape);
     void DeleteAll();
 
+    std::vector<CClientColShape*>::const_iterator IterBegin() { return m_List.begin(); }
+    std::vector<CClientColShape*>::const_iterator IterEnd() { return m_List.end(); }
+
```
Pirulax, line 34: Not sure what others think about it, but just return `const auto& GetColShapes() const {  return m_List }`, and iterate thru it with range based for loop

Client/mods/deathmatch/logic/CClientPedManager.h
```diff
@@ -41,6 +41,8 @@ class CClientPedManager
     void RestreamPeds(unsigned short usModel);
     void RestreamWeapon(unsigned short usModel);
 
+    unsigned int Count() { return static_cast<unsigned int>(m_List.size()); };
+
```
Pirulax, line 45: Return size_t instead, no casting needed.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -2584,3 +2592,30 @@ int CLuaElementDefs::IsElementWaitingForGroundToLoad(lua_State* luaVM)
...
+        }
+    }
+
+    if (argStream.HasErrors())
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+
+    // Failed
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
```
Pirulax: Use the new parser. See other PRs for examples. (Search for ArgumentParser in CLuaDefs)

Shared/mods/deathmatch/logic/Enums.h
Pirulax: Btw, what about EElementType? It's already there, and has these types..

PR #1381, opened 2020-04-12
Fix #1106: add the possibility to remove server from recent list

### LINKED ISSUE #1106: Removing servers from recent list
I think we should have a button to remove the server from the "recent" list, because some people like me end up accumulating a lot of servers in the list and often are servers that you don't even want to get on again.

**Additional context**
I know, we can remove it manually in coreconfig.xml, but it's not an easy thing for newbies.

ecastro98:
Fix #1106

### DIFF
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -2284,6 +2284,7 @@ bool CServerBrowser::OnServerListKeyDown(CGUIKeyEventArgs Args)
         }
         // Remove selected server from recent list when pressing on backspace
         case DIK_BACK:
+        case DIK_DELETE:
         {
             RemoveSelectedServerFromRecentlyPlayedList();
             break;
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -19,20 +19,20 @@ extern CCore* g_pCore;
 template <>
 CServerBrowser* CSingleton<CServerBrowser>::m_pSingleton = NULL;
 
-#define SB_SPAN             0.85f  //How much % of the screen the server browser should fill
-#define SB_NAVBAR_SIZE_Y    40 // Navbar button size
-#define SB_BUTTON_SIZE_X    26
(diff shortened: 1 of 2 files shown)


Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -2247,6 +2272,29 @@ bool CServerBrowser::OnServerListChangeRow(CGUIKeyEventArgs Args)
...
+    switch (Args.scancode)
+    {
+        case DIK_UPARROW:
+        case DIK_DOWNARROW:
+        {
+            OnServerListChangeRow(Args);
+            break;
+        }
+        // Remove selected server from recent list when pressing on backspace
+        case DIK_BACK:
```
Addlibs, line 2286: I'd add DIK_DELETE here too since its very common on Windows. Backspace is more common in macOS but both should be fine here imo

StrixG (approved):
Works great

Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -19,20 +19,20 @@ extern CCore* g_pCore;
...
+#define SB_SPACER 10            // Spacer between searchbar and navbar
+#define SB_SMALL_SPACER 5
+#define SB_SEARCHBAR_COMBOBOX_SIZE_X 45            // Mow much the search type combobox occupies of searchbar
+#define SB_SEARCHBAR_COMBOBOX_SIZE_Y 22
+#define SB_PLAYERLIST_SIZE_X 200            // Width of players list [NB. adjusted for low resolutions in CServerBrowser::CreateTab]
+#define SB_BACK_BUTTON_SIZE_Y 40            // Size of the back butt
+#define COMBOBOX_ARROW_SIZE_X 23            // Fixed CEGUI size of the 'combobox' arrow
+#define TAB_SIZE_Y 25                       // Fixed CEGUI size of the Tab in a tab panel
+
+#define CONNECT_HISTORY_LIMIT 20
```
Citizen01, line 35: Why are we allowing this unwanted changes ?

Citizen01: Why are we allowing these unwanted changes ? @StrixG

PR #1397, opened 2020-04-21
Update CEF to 81.2.19+g3b56636+chromium-81.0.4044.113

qaisjp:


jushar:
Sorry, I'm currently not able to test it.

If it works fine with the webbrowser default resource (just open it and check if things work on the loaded page), feel free to merge it. If it turns out to cause problems, we can always revert it anyway.

qaisjp:
Okay 👍

Please can someone check that this works? You can get a copy here: https://github.com/multitheftauto/mtasa-blue/suites/616036046/artifacts/4787565 (you need to open link in new tab)

qaisjp:
> If it works fine with the webbrowser default resource (just open it and check if things work on the loaded page), feel free to merge it. If it turns out to cause problems, we can always revert it anyway.

I just tested and that works here. Going ahead with merge.

PR #1398, opened 2020-04-21
Allow to use peds and vehicles as a camera target

TheNormalnij:
Allow to use peds and vehicles as a camera target in setCameraTarget function.
- [x] Clientside
- [x] Serverside


Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4515,23 +4515,30 @@ bool CStaticFunctionDefinitions::SetCameraTarget(CElement* pElement, CElement* p
             pTarget = pPlayer;
 
         // Make sure our target is a player element
```
Citizen01: Useless comment now.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4515,23 +4515,30 @@ bool CStaticFunctionDefinitions::SetCameraTarget(CElement* pElement, CElement* p
...
+                pPlayer->Send(CLuaPacket(SET_CAMERA_TARGET, *BitStream.pBitStream));
+                break;
+            }
+            default:
+                return false;
         }
     }
 
-    return false;
+    return true;
```
Citizen01: No, don't brake it and keep it to `false` there.

Citizen01 (changes requested):
Thanks for the PR. However, if it was this easy to add support for new element types as camera target, what blocks us to support all entity types ?
(I think there might be a valid reason for it to only support players in the 1st place but I don't know what it is then)

TheNormalnij:
I tried with objects but it didn't work.
Now i have bug with getCameraTarget. It need to be fixed.

TheNormalnij:
setCameraViewMode works only with 1-4 modes. Maybe it can be fixed later.

sbx320 (approved):
We'd need to lock this behaviour behind a min MTA version check, as otherwise this will cause issues with new vs old clients. Aside from that it looks good.  

So **before merging** (whoever does it): Please add a version check with an applicable version

TheNormalnij:
@sbx320 do you mean MinClientReqCheck and MinServerReqCheck or pBitStream->Version?

Citizen01 (approved):

Citizen01:
Requires a min MTA version check still.
Do not merge yet.

TheNormalnij:
Added

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4490,20 +4490,27 @@ bool CStaticFunctionDefinitions::SetCameraTarget(CElement* pElement, CElement* p
...
+                pCamera->SetFOV(70.0f);
 
-            CBitStream BitStream;
-            if (pPlayer->GetBitStreamVersion() >= 0x5E)
-                BitStream.pBitStream->Write(pCamera->GenerateSyncTimeContext());
-            BitStream.pBitStream->Write(pTarget->GetID());
-            pPlayer->Send(CLuaPacket(SET_CAMERA_TARGET, *BitStream.pBitStream));
-            return true;
+                CBitStream BitStream;
+                if (pPlayer->GetBitStreamVersion() >= 0x5E)
```
qaisjp, line 4506: Wrong bitstream you need to find the latest one

PR #1413, opened 2020-04-26
Remove DEBUG guards for OutputDebugLine

qaisjp:
OutputDebugLine is a macro that does nothing in debug mode.


qaisjp:
Just wanted to see if build passed or not

PR #1414, opened 2020-04-26
Fix #1090 Replaced weapon_crouch anim does not play if retainPedState is true

### LINKED ISSUE #1090: Replaced weapon_crouch anim does not play if retainPedState is true
**Describe the bug**
When the "weapon_crouch" animation is replaced and you play the replaced animation in setPedAnimation() with retainPedState set to true, the animation will play and as soon as its stopped the player will return to his default "weapon_crouch" animation.

**To reproduce**
https://github.com/multitheftauto/mtasa-blue/files/3096769/bug.zip

-> This resource replaces "weapon_crouch" with "prone_idle"
-> Crouch the player, he will be lying on the ground
-> crun setPedAnimation(localPlayer, "prone", "prone_idle", 50, false, true, true, false, 50, true)
-> The player will crouch when the animation is finished

**Expected behaviour**
retainPedState should not play the default "weapon_crouch" animation, it should play the replaced "prone_idle" animation when finished, it seems that an additional check is needed here.

**Windows Version 1.5.7-release-20157**

ghost:


Client/sdk/game/CAnimBlendAssocGroup.h
```diff
@@ -19,6 +19,160 @@ struct RpClump;
...
+    ANIM_GROUP_PRO = 133,
+    ANIM_GROUP_OLDWOMAN = 134,
+    ANIM_GROUP_FATWOMAN = 135,
+    ANIM_GROUP_JOGWOMAN = 136,
+    ANIM_GROUP_OLDFATWOMAN = 137,
+    ANIM_GROUP_SKATE = 138,
+    TOTAL_ANIM_GROUPS = 139
+};
+
+enum eMoveAnims
```
qaisjp: are we able to use use enum class?

ghost: Yes, that's definitely better. I forgot.

ghost:
I think it's probably better to create a merge commit when merging this PR. The second commit modifies a lot of code, and it has nothing to do with the issue.

qaisjp:
> I think it's probably better to create a merge commit when merging this PR. The second commit modifies a lot of code, and it has nothing to do with the issue.

agreed

qaisjp (approved):

PR #1418, opened 2020-04-27
Allow to change player camera mode

TheNormalnij:
getCameraViewMode will return 2 values: vehicle camera mode and ped camera mode
`int, int getCameraViewMode( )`
setCameraViewMode accepts 2 arguments:
`bool setCameraViewMode( int vehicleCameraMode [, int pedCameraMode ] )`
Ped camera modes:
1 - close
2 - middle
3 - far


Client/sdk/game/CCamera.h
```diff
@@ -95,6 +95,13 @@ enum eVehicleCamMode
     MODE_CINEMATIC
 };
 
+enum ePedCamMode
```
qaisjp: Cannot use `enum` in new code. Please use  `enum class`.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -106,11 +106,13 @@ int CLuaCameraDefs::GetCamera(lua_State* luaVM)
...
+    if (CStaticFunctionDefinitions::GetCameraViewMode(ucVehicleMode, ucPedMode))
     {
-        lua_pushnumber(luaVM, ucMode);
-        return 1;
+        lua_pushnumber(luaVM, ucVehicleMode);
+        lua_pushnumber(luaVM, ucPedMode);
+        return 2;
     }
 
     lua_pushboolean(luaVM, false);
```
qaisjp: Please convert to use new argument parser. See `Shared/.../CLuaCryptoDefs.cpp` for an example.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -443,13 +445,18 @@ int CLuaCameraDefs::GetCameraClip(lua_State* luaVM)
...
     if (!argStream.HasErrors())
     {
-        CStaticFunctionDefinitions::SetCameraViewMode(usViewMode);
+        CStaticFunctionDefinitions::SetCameraVehicleViewMode(usVehicleViewMode);
+
+        if (usPedViewMode != 0)
+            CStaticFunctionDefinitions::SetCameraPedViewMode(usPedViewMode);
 
         lua_pushboolean(luaVM, true);
         return 1;
```
qaisjp: Please convert to use new argument parser. See `Shared/.../CLuaCryptoDefs.cpp` for an example.

qaisjp (changes requested):
Thanks for the PR!

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: After all changes i have code like this
```
bool CLuaCameraDefs::SetCameraViewMode(eVehicleCamMode usVehicleViewMode, ePedCamMode usPedViewMode)
{
    CStaticFunctionDefinitions::SetCameraVehicleViewMode(usVehicleViewMode);

    if (usPedViewMode != 0) // here is wrong,  i know
        CStaticFunctionDefinitions::SetCameraPedViewMode(usPedViewMode);

    return true;
}
```
How to make second argument optional here?

qaisjp: `std::optional<ePedCamMode>`

```
if (usPedViewMode)
    SomeCall(usPedViewMode.value())
```

I think that will work

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: I think i can't return 2 values here.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: This way works, but there is another issue. I can put any value in second argument and function will return true, also i can put any string in first argument and function will return true. I need to use _int_ instead _unsigned short_ otherwise the parser does not work. New parser will broke old behavior.
```
bool CLuaCameraDefs::SetCameraViewMode(int usVehicleViewMode, std::optional<int> usPedViewMode)
{
    CStaticFunctionDefinitions::SetCameraVehicleViewMode((eVehicleCamMode)usVehicleViewMode);

    if (usPedViewMode)
        CStaticFunctionDefinitions::SetCameraPedViewMode((ePedCamMode)usPedViewMode.value());
 
    return true;
}
```

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
qaisjp: Have you tried `std::pair`?

TheNormalnij: I tied std::pair and std:tuple

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4922,15 +4922,22 @@ bool CStaticFunctionDefinitions::FadeCamera(bool bFadeIn, float fFadeTime, unsig
...
+    m_pCamera->SetCameraPedViewMode(ucMode);
+    return true;
+}
+
+bool CStaticFunctionDefinitions::GetCameraViewMode(unsigned short& ucVehicleMode, unsigned short& ucPedMode)
+{
+    ucVehicleMode = (unsigned short)m_pCamera->GetCameraVehicleViewMode();
+    ucPedMode     = (unsigned short)m_pCamera->GetCameraPedViewMode();
     return true;
 }
```
qaisjp: Unless those methods are being used in more than one place, you can just get rid of the methods in CStaticFunctionDefinitions and just move the content to the actual luadef.

Client/sdk/game/CCamera.h
```diff
@@ -95,6 +95,13 @@ enum eVehicleCamMode
     MODE_CINEMATIC
 };
 
+enum class ePedCamMode
+{
+    MODE_CLOSE,
+    MODE_MIDDLE,
+    MODE_FAR,
+};
```
qaisjp: enum classes don't need a `MODE_` prefix, because they can only be accessed like `ePedCamMode::MODE_CLOSE`.

I don't know if we want `ePedCamMode::Close` or `ePedCamMode::CLOSE`, so you can just pick one

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -104,17 +105,14 @@ int CLuaCameraDefs::GetCamera(lua_State* luaVM)
...
-        lua_pushnumber(luaVM, ucMode);
-        return 1;
-    }
+    CClientCamera* pCamera = g_pClientGame->GetManager()->GetCamera();
 
-    lua_pushboolean(luaVM, false);
-    return 1;
+    unsigned short ucVehicleMode = (unsigned short)pCamera->GetCameraVehicleViewMode();
+    unsigned short ucPedMode = (unsigned short)pCamera->GetCameraPedViewMode();
+
```
Pirulax: I don't think those casts are needed. They're implicitly converted.
Also, it's `us` not `uc`

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -104,17 +105,14 @@ int CLuaCameraDefs::GetCamera(lua_State* luaVM)
...
+    return std::make_tuple(ucVehicleMode, ucPedMode);
```
Pirulax: I think you can return `{ucVehicleMode, ucPedMode}`

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -441,25 +439,17 @@ int CLuaCameraDefs::GetCameraClip(lua_State* luaVM)
     return 2;
 }
 
-int CLuaCameraDefs::SetCameraViewMode(lua_State* luaVM)
+bool CLuaCameraDefs::SetCameraViewMode(int usVehicleViewMode, std::optional<int> usPedViewMode)
```
Pirulax: same here, it should be `unsigned `short as well, not int, since you've used it before

TheNormalnij: I think should be `unsigned char` there but new lua parser did not work with this type. I spend a lot of time here. I will change in `unsigned chrar` if it will works.

Pirulax: ohh, yeah. chars aren't supported. You could use unsigned short. as `us` stands for taht. Btw, tbh, I'd just leave out the hungarian notation, but since you've already included it, just fix it.

TheNormalnij: I will do this after merging #1555

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: It will broke old behavior.

Pirulax: It does the same. It constructs a tuple.
I didn't mean to return a table in Lua.
It's called an initializer list.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
Pirulax: If that ever gets merged...

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: Good stuff, thx.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
TheNormalnij: It should be.

TheNormalnij:
Changes requested but i don't see where.

StrixG:
I think you should make `vehicleCameraMode` optional so that you can individually set vehicle and ped camera modes.
`setCameraViewMode(nil, 1) -- sets only ped camera mode`

Or we can add new functions:
```lua
setCameraVehicleViewMode(int mode)
setCameraPedViewMode(int mode)
```
And deprecate the old one.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -439,25 +437,17 @@ int CLuaCameraDefs::GetCameraClip(lua_State* luaVM)
...
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+    if (ucPedViewMode)
+        pCamera->SetCameraPedViewMode((ePedCamMode)ucPedViewMode.value());
 
-    lua_pushboolean(luaVM, false);
-    return 1;
-}
+    return true;
+    }
```
StrixG: Wrong indentation.
```suggestion
}
```

Pirulax:
@qaisjp it says you've requested changes. Are they fulfilled?

patrikjuvonen (approved):

StrixG (approved):

Dutchman101:
@TheNormalnij please update the wiki pages accordingly:
[https://wiki.multitheftauto.com/wiki/SetCameraViewMode](https://wiki.multitheftauto.com/wiki/SetCameraViewMode)
[https://wiki.multitheftauto.com/wiki/GetCameraViewMode](https://wiki.multitheftauto.com/wiki/GetCameraViewMode)

Pirulax:
I dont think he has the time, so someone else has to.

stoneage-mta:
Wiki pages updated, now this PR can be closed.

prnxdev:
IMHO it should be separated to `getCameraViewMode` and `getVehicleCameraViewMode`

stoneage-mta:
This would break current scripts, because `getCameraViewMode` already returns the vehicle's camera view mode.

Pirulax:
It has been merged already.

Issue #1337, opened 2020-03-31
engineReplaceModel kicks the player out of the vehicle 


PlatinMTA:
**Describe the bug**
If you use engineReplaceModel() to model ID while you are on a vehicle with the same model ID, you will be warped out the vehicle.

**To reproduce**
Get inside a Perennial (404)

Run [this resource](https://github.com/multitheftauto/mtasa-blue/files/4409556/enginereplacebug.zip)
```lua
--clientside
function onStart()
        local txd,dff
	txd = engineLoadTXD("vehiculos/fiatuno.txd", 404) 
	engineImportTXD(txd, 404)
	dff = engineLoadDFF("vehiculos/fiatuno.dff", 404)
	engineReplaceModel(dff, 404)
end
addEventHandler("onClientResourceStart", resourceRoot, onStart)
```

**Expected behaviour**
The vehicle shouldn't kick you out

**Version**
Clientside: Windows 10 - v1.5.7-release-20462
Serverside: Windows 10 - v1.5.7-release-20462

**Additional context**
This also happens when using engineRestoreModel()

l33t

botder:
It's worth mentioning this long-standing bug might already count as feature and fixing it _could_ break existing scripts.

Lpsd:
In that case would it be possible to have a flag/option somewhere for this (defaulting to original behaviour)?

Not sure where that would be set though.

PlatinMTA:
> In that case would it be possible to have a flag/option somewhere for this (defaulting to original behaviour)?
> 
> Not sure where that would be set though.

The api could be changed to be something like this i think:

engineReplaceModel(dff, int model, [warp ped out of vehicle = true])

Lpsd:
> > In that case would it be possible to have a flag/option somewhere for this (defaulting to original behaviour)?
> > Not sure where that would be set though.
> 
> The api could be changed to be something like this i think:
> 
> engineReplaceModel(dff, int model, [warp ped out of vehicle = true])

I’m not sure about that, it seems kind of weird and messy (since engineReplaceModel doesn’t only deal with vehicles) - plus if a similar issue arises in future that requires backwards-compatibility breaking fixes to be implemented, we can’t just keep adding parameters, it’s ugly!

It probably calls for a separate issue to be opened, but maybe there needs to be some kind of flag system for each bugfix which would break backwards compatibility entirely (and doesn’t seem fit for simply adding an extra parameter).

qaisjp:
[...]

PR #1431, opened 2020-05-09
Fix #863 warpPedIntoVehicle fails when there are too many vehicles streamed in

### LINKED ISSUE #863: warpPedIntoVehicle doesn't work when there are too many vehicles streamed in
**Describe the bug**
I have a map that contains over 500+ vehicles and players can spawn vehicles themselves with _/vehicle_ command. This command also warps players into newly created vehicle. In any area which has vehicles created with the .map file, players can not warp into this newly created vehicle because warpPedIntoVehicle() doesn't do its job.

**To reproduce**
1. Create a map file that contains too many cars.
2. Test warpPedIntoVehicle to a vehicle that's created with createVehicle(...)
3. You'll see it's not going to warp into the vehicle.

**Expected behaviour**
I expect my ped warp into vehicle even though the area is full of vehicles or the .map file has 60000 vehicles.

**Screenshots**
 - no screenshots. -

**Version**
I'm using MTA 1.5.6 on Linux with Wine.

ghost:
**Context**

When the vehicle's memory pool reaches its limit, then the streamer needs to stream out a vehicle to make room for a new vehicle. This causes a small delay when we call `warpPedIntoVehicle`, and the game detects that the ped is supposed to be in a vehicle that doesn't exist, so it sets the occupied state of ped to false. When that new vehicle is finally created after the small delay, too bad the ped doesn't occupy the vehicle anymore and isn't warpped back into the vehicle.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -1793,7 +1793,7 @@ void CClientGame::UpdateVehicleInOut()
             }
 
             // Are we supposed to be in a vehicle? But aren't?
-            if (pOccupiedVehicle && !pVehicle)
+            if (pOccupiedVehicle && !pVehicle && !m_pLocalPlayer->IsWrapInToVehicleRequired())
             {
                 // Jax: this happens when we try to warp into a streamed out vehicle, including when we use CClientVehicle::StreamInNow
                 // ..maybe we need a different way to detect bike falls?
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -1793,7 +1793,7 @@ void CClientGame::UpdateVehicleInOut()
             }
 
             // Are we supposed to be in a vehicle? But aren't?
-            if (pOccupiedVehicle && !pVehicle && !m_pLocalPlayer->IsWrapInToVehicleRequired())
+            if (pOccupiedVehicle && !pVehicle && !m_pLocalPlayer->IsWarpInToVehicleRequired())
             {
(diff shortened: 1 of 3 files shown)


ghost:
Oh, it turns out #863 and #1337 are related. Should I fix it? @botder

qaisjp (approved):
Looks good. What if the player is being warped into *a different* vehicle whilst it's previous vehicle is momentarily unloaded? (Is that even possible?)

ghost:
> 
> 
> Looks good. What if the player is being warped into a different vehicle whilst it's previous vehicle is momentarily unloaded? (Is that even possible?)

I believe that scenario is handled by warpPedIntoVehicle function. 
```
// Toss the previous player out of it if neccessary
if (CClientPed* pPreviousOccupant = pVehicle->GetOccupant(uiSeat))
    RemovePedFromVehicle(pPreviousOccupant);
```
The vehicle occupant at a specific seat is kicked out of the vehicle when warpPedIntoVehicle is called and the occupant array of the previous vehicle is updated in CClientVehicle, so if the previous vehicle streams in before the new vehicle, it won't make a difference since it won't have the occupants in the array. And finally, when the new vehicle streams in, it will go through its occupant array and warp all peds into the vehicle.

The code at the bottom of `CClientGame::UpdateVehicleInOut` function looks like a hack to me, and it's responsible for the bug. Someone probably tried fixing a bug and didn't realize that it caused other bugs to pop up.

If you mean actual GTA internal warp, then obviously no. GTA SA gameplay logic runs on a single thread ;d

PlatinMTA:
I'm gonna test tomorrow (or later) if this fixes #1337, i hope it does.

Thanks.

ghost:
@PlatinMTA It doesn't fix #1337. I'll write a separate fix for it tomorrow.

PR #1433, opened 2020-05-10
Fix #1337 engineReplaceModel kicks the player out of the vehicle 

### LINKED ISSUE #1337: engineReplaceModel kicks the player out of the vehicle

ghost:
Fixes #1337 and adds a new property to setGlitchEnabled "kickoutofvehicle_onmodelreplace" to avoid breaking existing scripts. 

~~P.S Yeah, the git history is a mess, fixing it.~~


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2981,14 +2981,19 @@ void CClientVehicle::Destroy()
...
         // Remove all the passengers physically
         for (unsigned int i = 0; i < 8; i++)
         {
-            if (m_pPassengers[i])
+            CClientPed* pPassenger = m_pPassengers[i];
+            if (pPassenger)
             {
-                m_pPassengers[i]->InternalRemoveFromVehicle(m_pVehicle);
+                pPassenger->InternalRemoveFromVehicle(m_pVehicle);
+                if (!g_pClientGame->IsGlitchEnabled(CClientGame::GLITCH_KICKOUTOFVEHICLE_ONMODELREPLACE))
```
qaisjp: Does the compiler optimise this call and only call it once?

ghost: Good catch. This needs to be fixed.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2981,14 +2981,19 @@ void CClientVehicle::Destroy()
...
+        bool bWarpInToVehicleRequired = !g_pClientGame->IsGlitchEnabled(CClientGame::GLITCH_KICKOUTOFVEHICLE_ONMODELREPLACE);
         for (unsigned int i = 0; i < 8; i++)
         {
-            if (m_pPassengers[i])
+            CClientPed* pPassenger = m_pPassengers[i];
+            if (pPassenger)
             {
-                m_pPassengers[i]->InternalRemoveFromVehicle(m_pVehicle);
+                pPassenger->InternalRemoveFromVehicle(m_pVehicle);
+                pPassenger->SetWarpInToVehicleRequired(bWarpInToVehicleRequired);
```
qaisjp, line 2996: Should this be forced false?

qaisjp (approved):

Client/mods/deathmatch/logic/CClientVehicle.cpp
ghost: No, it's not forced to false. I removed the check because I added a logical NOT  `!` 
Capture
If IsGlitchEnabled returns true, then bWarpInToVehicleRequired is false, and if it returns false, then bWarpInToVehicleRequired is true.

qaisjp: I mean, before the most recent commit, it would ONLY set true if the glitch was enabled. Now it sets it sets to true or false, depending on whether glitch was enabled. Wasn't sure if that was 100% intended / has side effects. I don't know the big picture of the vehicle entry process

ghost: You're right, but it still won't make any difference because when `InternalRemoveFromVehicle` is called, it will set m_bWarpInToVehicleRequired to false for the ped. 
https://github.com/multitheftauto/mtasa-blue/blob/master/Client/mods/deathmatch/logic/CClientPed.cpp#L4186

And in this case, we're calling InternalRemoveFromVehicle before SetWarpInToVehicleRequired. Btw, SetWarpInToVehicleRequired only affects the bugged part of code that was causing issues with engineReplaceModel. It has no effect on other ped warping code.

PR #1437, opened 2020-05-13
Fix #1436: getValidPedModels() to include engineRequestModel skins as well

### LINKED ISSUE #1436: getValidPedModels() won't check if and id was allocated as a ped skin after ID 312
**Describe the bug**
getValidPedModels checks every ID from 0 to 312 to see if they are in fact ped skins. However it doesn't check if, for example, ID 3188 is a ped skin or not.

**Expected behaviour**
getValidPedModels() should also check if the allocable skins ID had been in fact allocated.

**Version**

Pirulax:
Okay, so in theory this fixes #1436 just fine.

Another way couldve been to just do this:
```cpp
    for (int i = 0; i <= MAX_MODEL_ID; i++)
    {
        if (CClientPlayerManager::IsValidModel(i))
        {
            lua_pushnumber(luaVM, ++iIndex);
            lua_pushnumber(luaVM, i);
            lua_settable(luaVM, -3);
        }
    }
```

but i felt like that would have  way bigger overhead than doing it this way. But correct me if im wrong.


PlatinMTA:
As a hotfix maybe you want it to go until ID 5389, because that is the last allocable skin for peds. It will skip like 1500 loops so i guess it's worth it

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -24,6 +26,18 @@ int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
         }
     }
 
+    // gather our, engineRequestModel skins
```
sbx320: that comma looks a bit out of place

Client/mods/deathmatch/logic/CClientModelManager.cpp
```diff
@@ -72,6 +72,22 @@ CClientModel* CClientModelManager::FindModelByID(int iModelID)
     return nullptr;
 }
 
+std::vector<CClientModel*> CClientModelManager::GetModelsByType(const eClientModelType type, const unsigned int minModelID)
+{
+    std::vector<CClientModel*> found;
+    found.reserve(16); // just pre-allocate some space, because we know there might be a few models
```
sbx320: 16 seems very low for a reserve. Ideally we'd just use the actual count of gta_sa ped models here

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -24,6 +26,18 @@ int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
...
+    // start at 313, so we wont reinsert the same skin ids(sometimes 3 is allocated, and that would appear twice in the list if we would start at 0)
+    for (const CClientModel* model : m_pManager->GetModelManager()->GetModelsByType(eClientModelType::CCLIENTMODELPED, 313)) 
+    {
+        lua_pushnumber(luaVM, ++iIndex);
+        lua_pushnumber(luaVM, model->GetModelID());
+        lua_settable(luaVM, -3);
+    }
+
+    // Another option could be to loop thru all MAX_MODELS skins,
```
sbx320: That comment doesn't really fit in here, just stating it in a pull request is perfectly fine

Pirulax:
> As a hotfix maybe you want it to go until ID 5389, because that is the last allocateable skin for peds. It will skip like 1500 loops so i guess it's worth it

Are you a 100 percent sure? Another approach would be to just:
```cpp
std::vector<std::vector<CClientModel*>> m_modelsByType;
```
or an std::map, for the first one. But Im not sure if thats necessary. To be quite honest, that seems overcomplicated to me, because nobody uses this function in render(if so, then they're plain stupid).

sbx320 (approved):
Looks good to me, just a few minor comments. Another alternative would be to only track added models (and hardcode the rest), but that'd be fragile and prone to future breakage if other model types are added.

Client/mods/deathmatch/logic/CClientModelManager.cpp
Pirulax: Its not that low actually. I could allocate 32, but more than that is unnecessary, because only custom models are stored here, the GTA ones arent.
But I could reserve 512, I mean thats only a few KB, so, doesnt really matter, and the value is RVOd anyways, so in theory there should be no copy of memory at all.

Pirulax:
Oh god, im so stupid at git, please help 🤣

> error: pathspec 'MTASA.sln.lnk' did not match any file(s) known to git

while trying to run:
` git commit *.* -m "made changes as suggested by sbx"`

Update: Ive googled around a bit more, and found out about -a. Now it works 😅

` git commit -a -m "made changes as suggested by sbx"`

Client/mods/deathmatch/logic/CClientModelManager.cpp
sbx320: RVO doesn't neccesarily apply due to `push_back`. But yeah, you're right here. I was thinking about `CClientModelManager` containing all (including GTA) models. 16 is probably fine here.

ccw808: Doesn't the first push_back effectively do a reserve anyway?

Pirulax:
Is this fine as it is?

My professional test method was:
```lua
local testCount = 100;

local requesteds = {}
for i = 1, testCount do
    requesteds[engineRequestModel("ped")] = true;
end

local found = 0;
for _, id in pairs(getValidPedModels()) do
    if (requesteds[id]) then 
        found = found + 1
     end
end
assert(found == testCount, "oh shit... here we go again!")
```

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -24,6 +26,15 @@ int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
         }
     }
 
+    // gather our custom skin model IDs allocated with engineRequestModel
+    // (there might be some < 313 as well, and since we dont want duplicates, we start at 313, others are already included by the loop above)
```
StrixG:
```suggestion
    // Gather our custom skin model IDs allocated with engineRequestModel
    // (there might be some < 313 as well, and since we don't want duplicates, we start at 313, others are already included by the loop above)
```

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -14,6 +14,8 @@ int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
 {
     int iIndex = 0;
     lua_newtable(luaVM);
+
+    // gather GTASA default skins
```
StrixG:
```suggestion
    // Gather GTASA default skins
```

Pirulax:
Issues:
- Because of the way engineFreeModel works ID's arent actually unreferenced (Removed) in the model manager, which causes this function to return `free-d` Ids
- Perhaps just add a new function: `engineGetModelsByType` as suggested by @qaisjp

PR #1442, opened 2020-05-15
Fix #1440 Driveby aiming is inverted in some cases

### LINKED ISSUE #1440: Driveby aiming is inverted in some cases
**Describe the bug**
When aiming left or behind in driveby, the gun pitch is inverted. When you aim down, the player points the gun up.
It only applies to the local player as MTA only syncs 1 aim rotation value for remote players in vehicles (Maybe this should be changed).

**To reproduce**
Get in a car or bike, perform driveby and point left or behind.

**Expected behaviour**
Gun should point where you aim.

**Screenshots**
image

**Version**
v1.5.7-release-20520
Bug has been in MTA for a long time

**Additional context**

Zangomangu:
Fixed by inverting fGunDirectionY if doing driveby and aiming left or behind.
Fixes #1440

### DIFF
--- a/Client/mods/deathmatch/logic/CNetAPI.cpp
+++ b/Client/mods/deathmatch/logic/CNetAPI.cpp
@@ -625,8 +625,8 @@ void CNetAPI::ReadKeysync(CClientPlayer* pPlayer, NetBitStreamInterface& BitStre
             BitStream.Read(&aim);
 
             // Read out the driveby direction
-            unsigned char ucDriveByAim;
-            BitStream.Read(ucDriveByAim);
+            eVehicleAimDirection ucDriveByAim;
+            BitStream.Read(*reinterpret_cast<char*>(&ucDriveByAim));
 
             // Set the aim data (immediately if in vehicle, otherwize delayed/interpolated)
             if (pVehicle)
@@ -763,7 +763,7 @@ void CNetAPI::WriteKeysync(CClientPed* pPlayerModel, NetBitStreamInterface& BitS
 
                 // Write the driveby direction
                 CShotSyncData* pShotsyncData = g_pMultiplayer->GetLocalShotSyncData();
-                BitStream.Write(pShotsyncData->m_cInVehicleAimDirection);
+                BitStream.Write(static_cast<char>(pShotsyncData->m_cInVehicleAimDirection));
             }
         }
         else
@@ -928,7 +928,7 @@ void CNetAPI::ReadPlayerPuresync(CClientPlayer* pPlayer, NetBitStreamInterface& 
             BitStream.Read(&aim);
 
             // Interpolate the aiming
-            pPlayer->SetAimInterpolated(TICK_RATE_AIM, rotation.data.fRotation, aim.data.fArm, flags.data.bAkimboTargetUp, 0);
+            pPlayer->SetAimInterpolated(TICK_RATE_AIM, rotation.data.fRotation, aim.data.fArm, flags.data.bAkimboTargetUp, eVehicleAimDirection::FORWARDS);
 
             // Read the aim data only if he's shooting or aiming
             if (aim.isFull())
@@ -1484,9 +1484,10 @@ void CNetAPI::ReadVehiclePuresync(CClientPlayer* pPlayer, CClientVehicle* pVehic
             // Read out the driveby direction
             SDrivebyDirectionSync driveby;
             BitStream.Read(&driveby);
+            eVehicleAimDirection ucDirection = static_cast<eVehicleAimDirection>(driveby.data.ucDirection);
 
             // Set the aiming
-            pPlayer->SetAimingData(TICK_RATE, aim.data.vecTarget, aim.data.fArm, 0.0f, driveby.data.ucDirection, &aim.data.vecOrigin, false);
+            pPlayer->SetAimingData(TICK_RATE, aim.data.vecTarget, aim.data.fArm, 0.0f, ucDirection, &aim.data.vecOrigin, false);
         }
         else
         {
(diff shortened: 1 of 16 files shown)


qaisjp (changes requested):
The nested `if`s are a bit intense - please can you use `return` where possible?

Client/multiplayer_sa/multiplayer_shotsync.cpp
```diff
@@ -168,6 +174,16 @@ VOID WriteGunDirectionDataForPed(CPedSAInterface* pPedInterface, float* fGunDire
             // Make sure our pitch is updated (fixes first-person weapons not moving)
             *fGunDirectionY = pGameInterface->GetCamera()->Find3rdPersonQuickAimPitch();
 
+            // Are we doing driveby?
+            if (pAimingPed->IsDoingGangDriveby())
```
ghost: Comments like these are not needed. They just take up space but are not helpful. `IsDoingGangDriveby` is self-explanatory.

Client/multiplayer_sa/multiplayer_shotsync.cpp
```diff
@@ -168,6 +174,16 @@ VOID WriteGunDirectionDataForPed(CPedSAInterface* pPedInterface, float* fGunDire
...
+            {
+                // Fix pitch in driveby when facing left or backwards
+                if (LocalShotSyncData.m_cInVehicleAimDirection == 1 || LocalShotSyncData.m_cInVehicleAimDirection == 2)
+                {
+                    *fGunDirectionY = -*fGunDirectionY;
+                }
```
ghost: I can see magic numbers here. It would be nice if you can define an enum in CMultiplayer.h. 

```
enum class eVehicleAimDirection : char
{
    FORWARDS = 0,
    LEFT, 
    BACKWARDS,
    RIGHT
};
```
CShotSyncData:
```
 eVehicleAimDirection m_cInVehicleAimDirection;
```

The fix:
```
switch (LocalShotSyncData.m_cInVehicleAimDirection)
{
    case eVehicleAimDirection::LEFT:
    case eVehicleAimDirection::BACKWARDS:
    *fGunDirectionY = -*fGunDirectionY;
    break;
}
```
You would also need to change a few references in multiplayer_shotsync.cpp from char to eVehicleAimDirection in multiplayer_shotsyn.cpp, CClientGame, and CClientPed. The compiler will throw errors and let you know where to change them.

Client/mods/deathmatch/logic/CNetAPI.cpp
```diff
@@ -631,12 +631,12 @@ void CNetAPI::ReadKeysync(CClientPlayer* pPlayer, NetBitStreamInterface& BitStre
             // Set the aim data (immediately if in vehicle, otherwize delayed/interpolated)
             if (pVehicle)
             {
-                pPlayer->SetAimingData(TICK_RATE_AIM, aim.data.vecTarget, aim.data.fArm, 0.0f, ucDriveByAim, &(aim.data.vecOrigin), false);
+                pPlayer->SetAimingData(TICK_RATE_AIM, aim.data.vecTarget, aim.data.fArm, 0.0f, static_cast<EVehicleAimDirection>(ucDriveByAim), &(aim.data.vecOrigin), false);
```
ghost: Remove static_cast and change the type of the parameter. Using casts for enums defeats the purpose of having enums in the first place. I think only using it for bitstreams is fine or in some cases where the type of enum is different from the local variable/function parameter and changing it to enum might possibly break things.

Client/mods/deathmatch/logic/CNetAPI.cpp
```diff
@@ -631,12 +631,12 @@ void CNetAPI::ReadKeysync(CClientPlayer* pPlayer, NetBitStreamInterface& BitStre
...
             }
             else
             {
                 pPlayer->SetTargetTarget(TICK_RATE_AIM, aim.data.vecOrigin, aim.data.vecTarget);
-                pPlayer->SetAimInterpolated(TICK_RATE_AIM, fPlayerRotation, aim.data.fArm, flags.data.bAkimboTargetUp, ucDriveByAim);
+                pPlayer->SetAimInterpolated(TICK_RATE_AIM, fPlayerRotation, aim.data.fArm, flags.data.bAkimboTargetUp, static_cast<EVehicleAimDirection>(ucDriveByAim));
             }
```
ghost: remove static_cast

Client/sdk/multiplayer/CMultiplayer.h
```diff
@@ -37,6 +37,14 @@ struct SClothesCacheStats
     uint uiNumRemoved;
 };
 
+enum class EVehicleAimDirection : char
+{
```
ghost: Hmm, I'm not sure why you changed the prefix to `E`. It's not consistent with the coding style. If you don't like using Hungarian notation, that's completely fine, but it's important to be consistent. I don't think we have any enums with the capital letter "E". I'd rename it to `VehicleAimDirection` or `eVehicleAimDirection`.

qaisjp: image
and more


But yes there's 256 `enum e` and 58 `enum E`, so indeed we should prefer lowercase.

qaisjp: Should it be `char` or `unsigned char`, though?

ghost: Oh my.. 

> Should it be `char` or `unsigned char`, though?

It doesn't make any difference if the range is between 0-127

Zangomangu: I looked around and the first enum I found had E. But I changed it now.

Zangomangu:
I've tested it again after the commits.

qaisjp (approved):
Refactor looks good. Force pushed to keep the bugfix and the refactor in distinct commits, so that we can do a rebase-merge or a merge commit.

PR #1450, opened 2020-05-20
Add CLuaArgument support to new Lua arg parser

Pirulax:
adds ability to read CLuaArguments.
Will be useful for setElementData, and probably other functions as well.

PR #1419, opened 2020-04-27
Fix CLuaFunctionParserBase lua_State* index check

qaisjp:

PR #1451, opened 2020-05-20
Use new arg parser for element data functions

Pirulax:
Requires #1450 and #1419


Pirulax:
Yeah, i failed miserably with merging the other PR, and accidentally leaving it in there.

**Im ashamed of my stupidity.**

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
```diff
@@ -8,6 +8,8 @@
  *
  *****************************************************************************/
 
+#include "lua/CLuaFunctionParser.h"
```
Pirulax: Note sure if this is needed. But otherwise it doesnt compile, because `CLuaFunctionParser` isnt a function.(In theory it should work just fine as the clientside CLuaCryptDefs uses the new parser as well)

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.h
```diff
@@ -10,6 +10,8 @@
 
 #pragma once
 #include "CLuaDefs.h"
+#include <optional>
+#include <variant>
```
Pirulax, line 14: Note sure about these two either, on server side it compiled just fine without them. 
Didnt just forward decl them, because they need a template.

qaisjp:
argument parser warn is only for pre 1.5.8 functions. so hasElementData + the fancy new element data functions should use a regular argument parser `ArgumentParser<F>`

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
```diff
@@ -1,87 +1,78 @@
...
-    {
-        // Warn and truncate if key is too long
-        m_pScriptDebugging->LogCustom(luaVM, SString("Truncated argument @ '%s' [%s]", lua_tostring(luaVM, lua_upvalueindex(1)),
-                                                     *SString("string length reduced to %d characters at argument 2", MAX_CUSTOMDATA_NAME_LENGTH)));
-        strKey = strKey.Left(MAX_CUSTOMDATA_NAME_LENGTH);
-    }
+#ifdef MTA_CLIENT
+    return element->GetCustomData(key.c_str(), inherit.value_or(true)) != nullptr;
+#else
+    return CStaticFunctionDefinitions::GetElementData(element, key.c_str(), inherit.value_or(true)) != nullptr;
```
qaisjp: `inherit.value_or(true)`

very cool! didn't know you could do this

Pirulax: Yeah, indeed. I went thru the doc of `optional` and `variant`. They're pretty clever things.

Pirulax (commented):
> argument parser warn is only for pre 1.5.8 functions. so hasElementData + the fancy new element data functions should use a regular argument parser `ArgumentParser<F>`

Yeah, I know that, totally forgot about `hasElementData` being a new function, and everything. Ill go ahead and make the changes.

Client/mods/deathmatch/logic/luadefs/CLuaDefs.h
qaisjp: Needs to be at the top of `CLuaElementDefs.cpp`, as shown here https://github.com/multitheftauto/mtasa-blue/blob/c99423b4421d4fa9e924e2e5163949a7b73e5fc9/Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp#L12

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.h
qaisjp: Yep, that's fine here.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1719,80 +1719,41 @@ int CLuaElementDefs::SetElementID(lua_State* luaVM)
...
+
+    //        if (CStaticFunctionDefinitions::RemoveElementData(*pEntity, strKey))
+    //        {
+    //            lua_pushboolean(luaVM, true);
+    //            return 1;
+    //        }
+    //    }
+    //}
+    //else
+    //    m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
```
qaisjp: what is happening here?

Pirulax: The function isnt even registered, so i commented it out to not cause any confusion(`CStaticFuncsDefs::RemoveElementData` is "implemented" as `return false;`)

qaisjp: just delete the code

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
```diff
@@ -1,87 +1,78 @@
 #include "StdInc.h"
 
-int CLuaElementDefs::GetElementData(lua_State* luaVM)
-{
-    //  var getElementData ( element theElement, string key [, inherit = true] )
+// Warn and truncate if key is too long(luaVM must be in a var luaVM)
```
sbx320: I'd prefer this to be a (private static) function rather than a macro

ghost: This. C++ doesn't like macros.

Pirulax: yeah, but then i need to pass in m_pScriptDebugging, which seemed a bit odd. Or i could define it a static CLuaElemetnDefs function, which would be confusing.
Which is the better way?

ghost: >  Or i could define it a static CLuaElemetnDefs function, which would be confusing.

I'd say it's better to make it a static function outside of CLuaElementDefs::GetElementData class. This function will be inaccessible to other .cpp files, and that's a good thing. By design, `CLuaElementDefs` is meant to have Lua functions, adding utility functions just doesn't look right. There are other ways to properly refactor the code, but this is a simple solution, and I think it's good.

Pirulax: You mean adding a function to `CLuaElementDefsShared.cpp` like, 
```cpp
void TruncateStringAndWarn(std::string& key, lua_State* const luaVM, CBlah const scriptDebugging);
```

ghost: Yes, and don't forget to add the `static` keyboard before the function.  :)

```
static void TruncateStringAndWarn(std::string& key, lua_State* const luaVM, CBlah const scriptDebugging)
{
// code
}
```

Pirulax:
*Do changes to the macro in shared element defs as suggested*
Actually, I've deleted it, and replaced it with a function.

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
Pirulax: Yeah, thanks for reminding!

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1721,47 +1721,6 @@ int CLuaElementDefs::SetElementID(lua_State* luaVM)
...
-    //    {
-    //        if (strKey.length() > MAX_CUSTOMDATA_NAME_LENGTH)
-    //        {
-    //            // Warn and truncate if key is too long
-    //            m_pScriptDebugging->LogCustom(luaVM, SString("Truncated argument @ '%s' [%s]", lua_tostring(luaVM, lua_upvalueindex(1)),
-    //                *SString("string length reduced to %d characters at argument 2", MAX_CUSTOMDATA_NAME_LENGTH)));
-    //            strKey = strKey.Left(MAX_CUSTOMDATA_NAME_LENGTH);
-    //        }
-
-    //        if (CStaticFunctionDefinitions::RemoveElementData(*pEntity, strKey))
```
qaisjp: what about the static function definition? that must be dead code too, right?

qaisjp (dismissed):
Linux server doesn't compile

Pirulax:
Well, rebasing was totally unnecessary... I could've cherry picked... Well, nvm now I guess.

Pirulax:
Turns out my super rebase didnt even include the PR merge of adding CLuaArgument to the parser...

Pirulax:
Yeah, i should rebase and remove these idiotic sommits, you're right.

Pirulax:
lets just get this working, and i dont care, we'll forget what happened here....

Pirulax:
Alright. Im not quite sure if its my fault, or the parser is being weird?
Because it compiles just fine on Windows...

buepas:
> Linux server doesn't compile

I think since CI has passed, that CR has been resolved?

Pirulax:
Thanks for fixing up the PR @qaisjp

ghost (dismissed):

patrikjuvonen:

PR #1453, opened 2020-05-21
Refactor LuaDef::LoadFunctions() to use array instead of map for inlining

Pirulax:
As the name states it, this PR modifies all 
`std::map<const char*, lua_CFunction> functions` to
`constexpr static const std::pair<const char*, lua_CFunction> functions[]`
so it can be inlined. 
Also the last commit modifies all `idk what it was` to 
```cpp
for (const auto& [name, func] : functions)
    CLuaCFunctions::AddFunction(name, func);
```


qaisjp (approved):

ghost:
This is really good.

ghost (approved):

StrixG (approved):

Pirulax:
> This is really good.

Thanks!

PR #1476, opened 2020-05-30
Fix std::optional popping and error messages in new parser

qaisjp:
**Issue 1: Pop Lua argument when invalid optional encountered**

Example:

- Signature: `test(string name, opt<int> bornAt, opt<int> diedAt)`
- Call (1): `test("1", nil, 1000)`
- Call (2): `test("2", 1000, 2000)`

In (1), only `name` will be set. In (2), both `bornAt` and `diedAt` are
set.

Thanks to @Pirulax for identifying the fix.

**Issue 2: Bad types to optionals are ignored (new parser)**

No warning is shown.


sbx320 (approved):
Good catch

PR #1467, opened 2020-05-26
Refactors to the code 

Pirulax:
Alright, here I'll try to refactor a lot of code, while in the same time having meaningful commits.
Im creating this PR, so the changes can be reviewed if needed.
Also, Im trying not to refactor the code too much, so I dont introduce bugs, because that woudnt be fun.

Mostly I refactor loop to use 'range based for loop' I also add a lot of 'auto', and 'constexpr' where needed.

PR #1478, opened 2020-05-31
Refactor CVector/2D/4D to use constexpr + noexcept everywhere

Pirulax:
Basically this PR makes the above mentioned classes use constexpr, and noexcept everywhere they can.

All functions other than the constructors were probably inlined before anyways, but declaring the constructors as `constexpr` allows me(or someone else), to declare constexpr variables with vectors. 
Its kinda useful, just gives me that _constifying_ feel. 

These edits are cherry-picked from #1467

### DIFF
--- a/Shared/sdk/CVector.h
+++ b/Shared/sdk/CVector.h
@@ -26,53 +26,63 @@
 class CVector
 {
 public:
-    float fX, fY, fZ;
+    float fX = 0.0f;
+    float fY = 0.0f;
+    float fZ = 0.0f;
 
-    CVector()
-    {
-        this->fX = 0;
-        this->fY = 0;
-        this->fZ = 0;
-    };
+    constexpr CVector() = default;
 
-    CVector(float fX, float fY, float fZ)
+    constexpr CVector(float x, float y, float z) :
+        fX(x),
+        fY(y),
+        fZ(z)
     {
-        this->fX = fX;
-        this->fY = fY;
-        this->fZ = fZ;
     }
 
-    CVector(const CVector4D& vec)
+    constexpr CVector(const CVector4D& vec) noexcept :
+        fX(vec.fX),
+        fY(vec.fY),
+        fZ(vec.fZ)
     {
-        this->fX = vec.fX;
-        this->fY = vec.fY;
-        this->fZ = vec.fZ;
     }
 
+    constexpr CVector& operator=(const CVector4D& vec) noexcept
+    {
+        fX = vec.fX;
+        fY = vec.fY;
+        fZ = vec.fZ;
+        return *this;
+    }
+
+
+    constexpr CVector Clone() const { return *this; }
+
+    // returns the length of the vector and normalizes it
     float Normalize()
     {
-        float t = sqrt(fX * fX + fY * fY + fZ * fZ);
+        const float t = Length();
         if (t > FLOAT_EPSILON)
         {
-            float fRcpt = 1 / t;
-            fX *= fRcpt;
-            fY *= fRcpt;
-            fZ *= fRcpt;
+            fX /= t;
+            fY /= t;
+            fZ /= t;
+
+            return t;
         }
         else
-            t = 0;
-        return t;
+            return 0;
     }
 
-    float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
+    inline float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
 
-    float LengthSquared() const { return (fX * fX) + (fY * fY) + (fZ * fZ); }
+    // returns just the squared length(eg.: x*x* + y*y + z*z)
+    inline float LengthSquared() const { return (fX * fX) + (fY * fY) + (fZ * fZ); }
 
-    float DotProduct(const CVector* param) const { return fX * param->fX + fY * param->fY + fZ * param->fZ; }
+    inline float DotProduct(const CVector* param) const { return fX * param->fX + fY * param->fY + fZ * param->fZ; }
 
     void CrossProduct(const CVector* param)
     {
-        float _fX = fX, _fY = fY, _fZ = fZ;
+        const float _fX = fX, _fY = fY, _fZ = fZ;
         fX = _fY * param->fZ - param->fY * _fZ;
         fY = _fZ * param->fX - param->fZ * _fX;
         fZ = _fX * param->fY - param->fX * _fY;
@@ -83,7 +93,7 @@ public:
     {
         CVector vecRotation;
         vecRotation.fZ = atan2(fY, fX);
-        CVector vecTemp(sqrt(fX * fX + fY * fY), fZ, 0);
+        CVector vecTemp(std::hypotf(fX, fY), fZ, 0);
         vecTemp.Normalize();
         vecRotation.fY = atan2(vecTemp.fX, vecTemp.fY) - PI / 2;
         return vecRotation;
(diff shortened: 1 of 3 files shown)


Pirulax:
OMG IT WORKS ON THE FIRST TRY jeez m8 thats gr8

sbx320 (approved):

Shared/sdk/CVector.h
```diff
@@ -26,53 +26,63 @@
...
     float Normalize()
     {
-        float t = sqrt(fX * fX + fY * fY + fZ * fZ);
+        const float t = Length();
         if (t > FLOAT_EPSILON)
         {
-            float fRcpt = 1 / t;
-            fX *= fRcpt;
-            fY *= fRcpt;
-            fZ *= fRcpt;
```
qaisjp, line 60: I thought muls were less expensive than divs, so this is an optimisation to div once, and mul multiple times.

It's possible that the compiler optimises this for us already, though.

Shared/sdk/CVector.h
```diff
@@ -26,53 +26,63 @@
...
-            t = 0;
-        return t;
+            return 0;
     }
 
-    float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
+    inline float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
 
-    float LengthSquared() const { return (fX * fX) + (fY * fY) + (fZ * fZ); }
+    // returns just the squared length(eg.: x*x* + y*y + z*z)
```
qaisjp:
```suggestion
    // LengthSquared returns Length() without sqrt applied (i.e. x*x* + y*y + z*z)
```

Shared/sdk/CVector.h
```diff
@@ -26,53 +26,64 @@
...
-        return t;
+            return 0;
     }
 
-    float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
+    inline float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
 
-    float LengthSquared() const { return (fX * fX) + (fY * fY) + (fZ * fZ); }
+    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y + z*z)
+    // For more info see CMaterialLine3DBatcher::Flush()
```
qaisjp: I don't think it's necessary to include the second line here because it's common to use `LengthSquared` for comparisons, which is when you don't strictly care about the square root.

If you still believe it's worth mentioning, then it might be clearer to have a comment like this, which is more descriptive and doesn't require the user to dig into potentially unrelated code.

```suggestion
    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y + z*z).
    // This can be useful if you only want to compare lengths.
```

Shared/sdk/CVector.h
```diff
@@ -192,96 +195,88 @@ class CVector
...
-    CVector operator-() const { return CVector(-fX, -fY, -fZ); }
+    constexpr CVector operator*(const float fRight) const noexcept { return CVector(fX * fRight, fY * fRight, fZ * fRight); }
 
-    void operator+=(float fRight)
+
+    constexpr CVector operator/(const CVector& vecRight) const noexcept { return CVector(fX / vecRight.fX, fY / vecRight.fY, fZ / vecRight.fZ); }
+
+    constexpr CVector operator/(const float fRight) const noexcept { return CVector(fX / fRight, fY / fRight, fZ / fRight); }
+
+
```
qaisjp:
```suggestion
```

Shared/sdk/CVector.h
```diff
@@ -192,96 +195,88 @@ class CVector
...
-    void operator+=(const CVector& vecRight)
+    constexpr void operator+=(const CVector& vecRight) noexcept
     {
         fX += vecRight.fX;
         fY += vecRight.fY;
         fZ += vecRight.fZ;
     }
 
-    void operator-=(float fRight)
+
```
qaisjp:
```suggestion
```

Shared/sdk/CVector.h
```diff
@@ -192,96 +195,88 @@ class CVector
...
-    void operator-=(const CVector& vecRight)
+    constexpr void operator-=(const CVector& vecRight) noexcept
     {
         fX -= vecRight.fX;
         fY -= vecRight.fY;
         fZ -= vecRight.fZ;
     }
 
-    void operator*=(float fRight)
+
```
qaisjp:
```suggestion
```

Shared/sdk/CVector.h
```diff
@@ -192,96 +195,88 @@ class CVector
...
-    void operator*=(const CVector& vecRight)
+    constexpr void operator*=(const CVector& vecRight) noexcept
     {
         fX *= vecRight.fX;
         fY *= vecRight.fY;
         fZ *= vecRight.fZ;
     }
 
-    void operator/=(float fRight)
+
```
qaisjp:
```suggestion
```

Shared/sdk/CVector4D.h
```diff
@@ -71,100 +61,83 @@ class CVector4D
...
+    constexpr CVector4D operator-(const CVector4D& vecRight) const noexcept { return CVector4D(fX - vecRight.fX, fY - vecRight.fY, fZ - vecRight.fZ, fW - vecRight.fW); }
 
-    CVector4D operator/(const CVector4D& vecRight) const { return CVector4D(fX / vecRight.fX, fY / vecRight.fY, fZ / vecRight.fZ, fW / vecRight.fW); }
 
-    void operator+=(float fRight)
+    constexpr CVector4D operator/(const CVector4D& vecRight) const noexcept { return CVector4D(fX / vecRight.fX, fY / vecRight.fY, fZ / vecRight.fZ, fW / vecRight.fW); }
+
+    constexpr CVector4D operator/(const float fRight) const noexcept { return CVector4D(fX / fRight, fY / fRight, fZ / fRight, fW / fRight); }
+
+
```
qaisjp:
```suggestion
```

qaisjp (changes requested):
Thank you! I've left a couple style comments, a comment comment, and a suggestion to keep one potential optimisation.

Shared/sdk/CVector.h
```diff
@@ -26,53 +26,64 @@
...
+        fX = vec.fX;
+        fY = vec.fY;
+        fZ = vec.fZ;
+        return *this;
     }
 
+
+    constexpr CVector Clone() const { return *this; }
+
+    // returns the length of the vector and normalizes it
```
qaisjp:
```suggestion
    // Normalize returns the normalized length of the vector.
```

Pirulax:
I feel like that "potential optimization" is quite unnecessary, as, by default MSVC uses SSE2, which will probably dump all the 4 floats into a m128 reg, and divide them. Which wont be much slower than just doing this optimization trick(which we shoudnt, as the compiler will optimize it, especially if we enable fast-math(which will pretty much guarantee us that the compiler will use SSE2 with this trick we've used:
``` cpp
#pragma fast-math push
#pragma fast-math on
// code
#pragma fast-math pop
```

Shared/sdk/CVector2D.h
```diff
@@ -20,128 +20,123 @@
...
 
-    float DotProduct(CVector2D& other) const { return fX * other.fX + fY * other.fY; }
+    constexpr float DotProduct(CVector2D& other) const { return fX * other.fX + fY * other.fY; }
 
-    float Length() const { return sqrt(fX * fX + fY * fY); }
+    float Length() const { return std::hypotf(fX, fY); }
 
-    float LengthSquared() const { return (fX * fX) + (fY * fY); }
+    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y)
+    // For more info see CMaterialLine3DBatcher::Flush()
```
qaisjp:
```suggestion
    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y).
    // This can be useful if you only want to compare lengths.
```

Shared/sdk/CVector4D.h
```diff
@@ -20,48 +20,38 @@
...
+    // Warning, this function is returning the wrong value(fW is missing), its kept because nothing uses it, only
+    // CLuaVector4DDefs.
+    constexpr float DotProduct(const CVector4D& other) const noexcept { return fX * other.fX + fY * other.fY + fZ * other.fZ; }
 
-    float Length() const { return sqrt(fX * fX + fY * fY + fZ * fZ + fW * fW); }
+    float Length() const noexcept { return sqrt(fX * fX + fY * fY + fZ * fZ + fW * fW); }
 
-    float LengthSquared() const { return (fX * fX) + (fY * fY) + (fZ * fZ) + (fW * fW); }
+    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y + z*z + w*w)
+    // For more info see CMaterialLine3DBatcher::Flush()
```
qaisjp:
```suggestion
    // LengthSquared returns Length() without sqrt applied (i.e. returns x*x* + y*y + z*z + w*w).
    // This can be useful if you only want to compare lengths.
```

qaisjp (approved):

ghost:
@Pirulax 
> tbh I agree, but nobody reads CVector code anyways. Constexpr is needed(at least for the constructor) in CGraphics refactor(somewhere in the end, the sphere thing)

Why is constexpr needed there? The main purpose of constexpr is to compute the result in compile time. That can only happen if the compiler knows about the values used in the method at compile-time, otherwise, it will just inline it at best and won't do any good optimizations.

Pirulax:
As stated somewhere, I need this for the CGraphics refactor thingy

PR #1485, opened 2020-06-01
Fix getModuleInfo

qaisjp:
It would just complain about giving it a bad type as an argument, even
though you were giving it a perfectly valid string.


ghost:
Is it possible to use the new argument parser for this function?

qaisjp:
CLuaFunctionDefs needs to inherit CLuaDefs to use the new parser. Should I do that?

ghost:
> 
> 
> CLuaFunctionDefs needs to inherit CLuaDefs to use the new parser. Should I do that?

I don't see any problem with that. Yes. :)

qaisjp:
I've given it a go but it's a too complicated refactor that I think we should not do for 1.5.8.

ghost:
I see. Sure, it can be done after 1.5.8 release.

ghost (approved):

PR #1488, opened 2020-06-01
Remove unused variable from COpenPortsTester.h

AlexTMjugador:
The `strServerIP` variable is not used in the `Start()` method of the `COpenPortsTester` class, so it can (and probably should) be removed. I don't think that this variable is meant to do anything meaningful anyway, as I couldn't find any commit or issue relating to it.


qaisjp (approved):
Thanks!

PR #1496, opened 2020-06-06
Enable GPU for CEF

### LINKED ISSUE #1493: Transparent attribute in createBrowser/guiCreateBrowser doesn't work
**Describe the bug**
If you set the transparent attribute to true in createBrowser/guiCreateBrowser and load a page without a background, the background in MTA is white and not transparent. This problem did not occur in 20561. It is probably new since 20566. Maybe this is related to commit 42cb6691357adb135e91f50b2355836f2c5a9e07?

**Version**
Multi Theft Auto v1.5.7-release-20566

**Reproduction steps**

1. Apply following patch to `webbrowser` resource to enable browser transparency

   ```diff
	diff --git a/[gameplay]/webbrowser/WebBrowserGUI.lua b/[gameplay]/webbrowser/WebBrowserGUI.lua
	index 6e82888..32b2a75 100644
	--- a/[gameplay]/webbrowser/WebBrowserGUI.lua
	+++ b/[gameplay]/webbrowser/WebBrowserGUI.lua
	@@ -20,7 +20,7 @@ function WebBrowserGUI:constructor()
	        self.m_ButtonClose:setProperty("HoverTextColour", "FF990909")
	        self.m_ButtonClose:setFont("default-bold-small")
	
	-       self.m_Browser = GuiBrowser(5, 62, sizeX - 10, sizeY - 67, false, false, false, self.m_Window)
	+       self.m_Browser = GuiBrowser(5, 62, sizeX - 10, sizeY - 67, false, true, false, self.m_Window
	```
2. Go to `mtasa.com/asdfasdf` to see 404 page
3. Background should be transparent. It is white!

qaisjp:
This removes `--disable-gpu`. I haven't checked performance, though. How 2 check performance @Jusonex?

-----

Fix #1493 'browser transparency does not work' (tested fix)

Upstream issue:
https://bitbucket.org/chromiumembedded/cef/issues/2938/osr-transparent-background-broken-since
/
https://bitbucket.org/chromiumembedded/cef/issues/2944/osr-no-transparency-when-gpu-disabled

Partially reverts 2b3998456b3180c2f40da6055bf307e470e6f457 "Tweak CEF
rendering performance"

### DIFF
--- a/Client/cefweb/CWebApp.cpp
+++ b/Client/cefweb/CWebApp.cpp
@@ -27,8 +27,6 @@ void CWebApp::OnRegisterCustomSchemes(CefRawPtr<CefSchemeRegistrar> registrar)
 
 void CWebApp::OnBeforeCommandLineProcessing(const CefString& process_type, CefRefPtr<CefCommandLine> command_line)
 {
-    command_line->AppendSwitch("disable-gpu-compositing");
-    command_line->AppendSwitch("disable-gpu");
     // command_line->AppendSwitch("disable-d3d11");
     command_line->AppendSwitch("enable-begin-frame-scheduling");
 


qaisjp:
~~I'm doing an additional test to see if it actually fixes the issue, reading the bitbucket issues again it seems that it might still be broken when background:white~~

~~(i only tested on mta's 404 page, which has no background)~~

EDIT: I misread the issue, `background` must be unset or set to `transparent` - white should not cause a transparent background.

(Completely disregard this comment)

jushar (changes requested):
tl;dr: GPU cannot be enabled without extra effort.

### Why is GPU disabled?
With GPU rendering enabled, CEF renders frames into a D3D11 texture (on Windows) and displays it directly (this is the non-offscreen rendering approach that is used in browsers).   
The "legacy" offscreen-rendering approach (which MTA is using) is that we copy over the texture from  CEF into our own D3D9 texture - this is done more or less byte-by-byte (`memcpy` accelerates it a bit thanks to the CPU's vector extensions).
If GPU rendering is enabled, the texture memory is copied from the graphics card over the bus to the CPU. As lots of bus communication is involved for that operation, copying these textures is very slow. It's even slower, because all that has to be done the other way round too as we write the texture back into the D3D9 texture which also resides in GPU memory.

If we instead turn GPU rendering off, everything is calculated on the CPU using a software renderer. The rendering alone is ofc slower, but we can observe that this approach is still significantly faster as expensive copy operations are reduced.

### How can we fix that?
As said earlier, we use the "legacy" approach. That means, there's a new approach that allows us to access the D3D11 texture that CEF uses internally directly. We can't use that though without extra effort as GTA is built on top of D3D9.

Luckily, there's a compatibility API that allows us to create a D3D9 texture that shares memory with a D3D11 texture (it's a parameter of the [`IDirect3DDevice9::CreateTexture`](https://docs.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-createtexture) function).   
It turns out though that this API can only be used if the device was created by the `CreateDeviceEx` (GTA invokes `CreateDevice`).   
Unfortunately, this is not just a drop-in replacement, but comes with a few breaking changes in the DirectX API, such as some dropped texture types/usages (there's a changelog somewhere), so these changes have to be dealt with first (I remember that @ccw808 did some promising tests on a fork).

### Notes on the issue
If there's no easy workaround, I suggest reverting back to the last working version (is there an older CEF build that is newer than MTA's?) and waiting for a fix by the CEF authors.

qaisjp:
Nice explanation! Really interesting context.

> If there's no easy workaround, I suggest reverting back to the last working version (is there an older CEF build that is newer than MTA's?) and waiting for a fix by the CEF authors.

Yes, there is one. I've just created a new pull request for this, #1498.

PR #1499, opened 2020-06-06
Fix playSound3D actually playing 2D for raw data

correaAlex:


qaisjp (approved):
Makes sense but I haven't tested.

@Dezash would you be able to take a look at this? Line introduced in f215edb545212de7aba2a42acaa4389649b82c5a

correaAlex:
> Makes sense but I haven't tested.
> 
> @Dezash would you be able to take a look at this? Line introduced in [f215edb](https://github.com/multitheftauto/mtasa-blue/commit/f215edb545212de7aba2a42acaa4389649b82c5a)

I tested it, it seems to me is was a typo. The sound did not play in 3D from raw data

qaisjp:
Thank you so much!

PR #1500, opened 2020-06-06
Add functionality to build the debug version with docker

buepas:
This is a QoL improvement. With this change Ubuntu users have less problems building the docker image (since apt-get install needs to be ran as root), as well as building the debug target without needing to rebuild the image every time the target has to be changed


qaisjp (approved):
Looks great! Thank you so much :)

PR #1506, opened 2020-06-08
Fix #1291: getRotation returns incorrect values for ped when set with createPed

### LINKED ISSUE #1291: Ped rotation is always zero on client side if ped has been frozed after creating it
**Describe the bug**
Ped rotation is always zero on client side if ped has been frozed after creating it.

**To reproduce**
```lua
--- Serverside!
local peds = {
	{150, 1464.1395263672, -1002.5094604492, 26.828125, 90},
}

for i, v in ipairs(peds) do
	local ped = createPed(v[1], v[2], v[3], v[4], v[5], false)
	setElementFrozen(ped, true)
        setElementID(ped, "testPED")
end
```

Server side runcode: elem:ped[150]00000194E4B44F28 [element:ped]
Result: vector3: { x = 0.000, y = 0.000, z = 90.000 } [userdata]

Client side runcode: getElementByID("testPED").rotation
Result: vector3: { x = 0.000, y = 0.000, z = 0.000 } [userdata]

**Version**
Multi Theft Auto v1.5.7-release-20406

Pirulax:
Basically, when you created a ped with `createPed` server side, and set a rotation there, the client side `getElemRot` returned incorrect values, but the ped was visually at the given rot.
This was caused by the matrix not being updated by GTA. We only set current/target rotations in the interface.

This PRr fixes the issue. I dont think it ruins backwards compatibility, since the only case this issue occurs is when the ped is created after setting the rotation values. This probably only occurs when creating the ped with `EntityAdd` packet.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -3614,9 +3614,15 @@ void CClientPed::_CreateModel()
 
         // Restore any settings
         m_pPlayerPed->SetLanding(false);
-        m_pPlayerPed->SetMatrix(&m_Matrix);
+        // update matrix to contain current rotation values - fix github: #1291
+        // as setting it thru the interface(SetCurrent/TargetRotation) doesnt update the matrix 
+        {
+            m_Matrix.SetRotation({ 0, 0, m_fCurrentRotation });
+            SetMatrix(m_Matrix); // also sets the matrix on m_pPlayerPed
+        }
         m_pPlayerPed->SetCurrentRotation(m_fCurrentRotation);
         m_pPlayerPed->SetTargetRotation(m_fTargetRotation);
+
         m_pPlayerPed->SetMoveSpeed(&m_vecMoveSpeed);
         m_pPlayerPed->SetTurnSpeed(&m_vecTurnSpeed);
         Duck(m_bDucked);
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -3614,12 +3614,15 @@ void CClientPed::_CreateModel()
 
         // Restore any settings
         m_pPlayerPed->SetLanding(false);
-        // update matrix to contain current rotation values - fix github: #1291
-        // as setting it thru the interface(SetCurrent/TargetRotation) doesnt update the matrix 
-        {
-            m_Matrix.SetRotation({ 0, 0, m_fCurrentRotation });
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -3614,9 +3614,15 @@ void CClientPed::_CreateModel()
 
         // Restore any settings
         m_pPlayerPed->SetLanding(false);
-        m_pPlayerPed->SetMatrix(&m_Matrix);
+        // update matrix to contain current rotation values - fix github: #1291
+        // as setting it thru the interface(SetCurrent/TargetRotation) doesnt update the matrix 
+        {
+            m_Matrix.SetRotation({ 0, 0, m_fCurrentRotation });
+            SetMatrix(m_Matrix); // also sets the matrix on m_pPlayerPed
+        }
```
qaisjp: Style comments:

- Please use correct capitalisation (`Update`), spelling (`through`) and punctuation (`doesn't`). Also, leave a space before your opening parenthesis (`interface (Set...)`).
- Please leave a blank line above comments.
- Please format your code to abide with .clang-format (brace spacing, comment spacing)

```suggestion

        m_pPlayerPed->SetCurrentRotation(m_fCurrentRotation);
        m_pPlayerPed->SetTargetRotation(m_fTargetRotation);

        // Fix GitHub #1291: also update the matrix to contain the current rotation
        // as just setting it through the interface above doesn't update our matrix.
        m_Matrix.SetRotation({0, 0, m_fCurrentRotation});
        SetMatrix(m_Matrix);            // also sets the matrix on m_pPlayerPed

```

qaisjp (changes requested):
Firstly -- thanks for the pull request!

There's some style comments here - I know they can seem nitpicky but I hope you can bare with them!

But I am not certain this is a legit fix -- _I_ would expect the SetCurrentRotation method to update the matrix. I'm not 100% sure if that's the correct behaviour, though.

Pirulax:
Well, when creating a ped server side with `createPed`, and the entity add packet arrives, and sets the rotation using `SetCurrentRotation` `m_pPlayerPed` isnt yet created, thats why the matrix isnt updated. Its pretty interesting as to why it doesnt update the matrix when the rotation is just fine.

ghost (approved):

qaisjp:
> and sets the rotation using SetCurrentRotation

why doesn't SetCurrentRotation update the matrix?

patrikjuvonen:
I would like to merge this fix but is there an answer as to why SetCurrentRotation doesn't update the matrix and should that be the case or not?

PR #1511, opened 2020-06-09
Feature/async checksum

Pirulax:
tldr: adds async resource file checksum generation, also gets rid of redundant ones(todo).

### DIFF
--- a/Client/mods/deathmatch/logic/CDownloadableResource.cpp
+++ b/Client/mods/deathmatch/logic/CDownloadableResource.cpp
@@ -61,6 +61,11 @@ CChecksum CDownloadableResource::GetServerChecksum()
     return m_ServerChecksum;
 }
 
+CChecksum CDownloadableResource::GetLastClientChecksum()
+{
+    return m_LastClientChecksum;
+}
+
 int CDownloadableResource::GetDownloadPriorityGroup()
 {
     return m_pResource->GetDownloadPriorityGroup();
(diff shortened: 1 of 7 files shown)


Pirulax:
Please test this on your PC as well, as mine is a pos, but someone with a better CPU should clearly see the differences, when joining a server with 1 gb of files.

Also, as this is a daft, its not complete(there are still redundant checksum generations, randomly across the project for some reason...)

Pirulax:
Okay, so I have a question: Should we regen file checksums when loading the resource? Because, right now, its already checked in `CClientGame::Packet_StartResource` before start. Its also checked in `CResourceFileDownloadManager::OnDownloadFinished`, if the files needed to be downloaded.

Pirulax:
This PR will be deleted, as I have a better solution to this. But I still need opinions about it, so please leave them if u have any

qaisjp:
Some feedback:

- `CCs` should be Checksums
- Be careful just getting rid of the 'redundant' checksum checks - have you tried using `git blame` to find out which commit they were added in? The commit message might explain why it's necessary / lead to a bug report that explains why it is necessary.

  The [`.git-blame-ignore-revs` file](https://github.com/multitheftauto/mtasa-blue/blob/master/.git-blame-ignore-revs) helps make the blame process easier by excluding commits with whitespace/formatting mass-changes. Read the file for documentation on how to use that file.

PR #1516, opened 2020-06-11
Add async checksum generation for resource files

Pirulax:
Basically, resource file checksums were checked on the main thread(problem one), also, they were checked, in worst case scenario, 3 times(on the first load), and best case scenario 2.

This PR adds async checksum generation, as well as removes redundant checks. It works perfectly for me, I had no issues whatsoever with it. 

Some notes to who ever tests this PR:
- `Packet_StartResource` is called every time a resources needs to be started. No matter whenever its started with a command / is in `mtaconfig`

- After a resource is stopped all its associated `CResourceFiles` are deleted, which means, after restarting the resource their checksums are rechecked

- Adding the additional `const bool bCanDownload` was necessary to avoid a 'download loop' [as i call it], which occurs when i dont know how / when, but if it does, the resource is stuck in a loop, and tries to download the file over and over again

- `bIsSehrGut` is serious, and without it the code would explode

- Im sorry for the 'addendums', but I just totally forgot about a few things, but I wanted to keep the commits as clean as possible. 

- A rebase will be needed to remove debug stuff

Edit:
Bugs discovered:
- Seems like after updating a resource file, it doesnt re-download it. Needs further investigation
- Make sure there's enough RAM to load the whole file into memory, if not use the fallback, old method(65 kb stack buffer)

### DIFF
--- a/Client/mods/deathmatch/logic/CDownloadableResource.cpp
+++ b/Client/mods/deathmatch/logic/CDownloadableResource.cpp
@@ -10,6 +10,8 @@
  *****************************************************************************/
 
 #include <StdInc.h>
+#include <functional> // std::bind
+#include <future>     // return value from SharedUtil::async
 
 CDownloadableResource::CDownloadableResource(CResource* pResource, eResourceType resourceType, const char* szName, const char* szNameShort, uint uiDownloadSize,
                                              CChecksum serverChecksum, bool bAutoDownload)
@@ -39,6 +41,14 @@ CDownloadableResource::~CDownloadableResource()
 bool CDownloadableResource::DoesClientAndServerChecksumMatch()
 {
     return (m_LastClientChecksum == m_ServerChecksum);
+}
+
+std::future<CChecksum> CDownloadableResource::GenerateClientChecksumAsync()
+{
+    // Force us to use a buffer and not re-read the file twice,
+    // because HDD bandwidth is more important than multi-core speed(we're fast enough already in terms of CPU)
+    return SharedUtil::async([this]() { return GenerateClientChecksum(CBuffer()); });
+    //return SharedUtil::async(std::bind(&CDownloadableResource::GenerateClientChecksum, this), CBuffer());
 }
 
 CChecksum CDownloadableResource::GenerateClientChecksum()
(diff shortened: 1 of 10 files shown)


Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -226,35 +226,127 @@ bool CResource::CallExportedFunction(const char* szFunctionName, CLuaArguments&
...
+    }
+
+    // Make sure all checksums are generated
+    // We do this here, so it can run async
+    for (auto* pFile : m_ResourceFiles)
+        if (pFile->HasClientChecksumGenerated())
+            pFile->GenerateClientChecksumAsync();
+
+    // Files should be downloaded by now, let's check their hashes, and re-download if needed
+    bool bAllWasSehrGut = true;
```
Lpsd, line 277: Can you use an English var name?

Pirulax: :(
It was meant to be a joke. 
But sure, if it bothers anyone I can change it, no problem.

Client/mods/deathmatch/logic/CDownloadableResource.h
```diff
@@ -20,11 +20,12 @@
 
 #include <bochs_internal/crc32.h>
 #include "CChecksum.h"
+#include <future>
 
 class CDownloadableResource
 {
 public:
-    enum eResourceType
+    enum eResourceType : unsigned char // Must be unsigned char, so in CClientGame::Packet_StartResource it's read properly from the packet.
```
qaisjp, line 28: put the comment above on a new line above

Client/mods/deathmatch/logic/CDownloadableResource.cpp
```diff
@@ -10,23 +10,20 @@
  *****************************************************************************/
 
 #include <StdInc.h>
+#include <future>     // return value from SharedUtil::async
+
+#define INVALIDATE_OLD_FUTURE if (m_GenerateChecksumFuture.valid()) m_GenerateChecksumFuture = std::future<CChecksum>();
```
qaisjp, line 15: what's wrong with a private function?

Pirulax: I felt like this is neater, than declaring a 4-5 line function.

qaisjp: macros are undesirable

ghost: > macros are undesirable

agreed

Pirulax: agreed

PR #1517, opened 2020-06-11
Improve export function call performance, and a small refactor to CResource

Pirulax:
As we all know(or at least me xd), exported function calls are pretty slow. 

Currently, we use an std::list with a heap allocated object(not even Terry A. Davis knows why). This PR aims to improve its performance by using a hash set. 

I think the benefits are obvious: **no cache misses, log(n)** VS **cache misses(a lot) and O(n)**

### DIFF
--- a/Client/mods/deathmatch/logic/CResource.cpp
+++ b/Client/mods/deathmatch/logic/CResource.cpp
@@ -202,11 +202,10 @@ void CResource::AddExportedFunction(const char* szFunctionName)
     m_ExportedFunctionsSet.insert(szFunctionName); 
 }
 
-bool CResource::CallExportedFunction(const char* szFunctionName, CLuaArguments& args, CLuaArguments& returns, CResource& caller)
+bool CResource::CallExportedFunction(const SString& strFunctionName, CLuaArguments& args, CLuaArguments& returns, CResource& caller)
 {
-    // Check if we actually have this function exported
-    if (m_ExportedFunctionsSet.find(szFunctionName) != m_ExportedFunctionsSet.end())
-        return args.CallGlobal(m_pLuaVM, szFunctionName, &returns);
+    if (m_ExportedFunctionsSet.find(strFunctionName) != m_ExportedFunctionsSet.end()) // Is it actually exported?
+        return args.CallGlobal(m_pLuaVM, strFunctionName, &returns);
     return false;
 }
 
--- a/Client/mods/deathmatch/logic/CResource.cpp
+++ b/Client/mods/deathmatch/logic/CResource.cpp
@@ -466,3 +466,11 @@ void CResource::HandleDownloadedFileTrouble(CResourceFile* pResourceFile, bool b
     g_pClientGame->TellServerSomethingImportant(bScript ? 1002 : 1013, strMessage, 4);
     g_pCore->GetConsole()->Printf("Download error: %s", *strMessage);
 }
+
+CResourceFile* CResource::GetFileFromShortName(const char* szShortName)
+{
+    for (auto* pFile : m_ResourceFiles)
+        if (strcmp(szShortName, pFile->GetShortName()) == 0)
+            return pFile;
+    return nullptr;
+}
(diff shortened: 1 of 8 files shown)


Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -482,3 +466,12 @@ void CResource::HandleDownloadedFileTrouble(CResourceFile* pResourceFile, bool b
     g_pClientGame->TellServerSomethingImportant(bScript ? 1002 : 1013, strMessage, 4);
     g_pCore->GetConsole()->Printf("Download error: %s", *strMessage);
 }
+
+// Check if resource has a file by meta path(also known as short name)
```
StrixG:
```suggestion
// Check if resource has a file by meta path (also known as short name)
```

Shared/sdk/SharedUtil.FastHashMap.h
```diff
@@ -146,4 +146,14 @@ inline SString GetDeletedMapKey(SString*)
...
+inline std::string GetEmptyMapKey(std::string*)
+{
+    return std::string("\xFF\xFE");
+}
+
+inline std::string GetDeletedMapKey(std::string*)
+{
+    return std::string("\xFF\xFF");
+}
+
```
qaisjp: ?

Pirulax:
If everything seems fine to you guys, let me know, and I'll rebase.

Client/mods/deathmatch/logic/CResource.h
```diff
@@ -110,6 +106,7 @@ class CResource
     int            GetDownloadPriorityGroup() { return m_iDownloadPriorityGroup; }
     void           SetDownloadPriorityGroup(int iDownloadPriorityGroup) { m_iDownloadPriorityGroup = iDownloadPriorityGroup; }
 
+    CResourceFile* GetFileFromShortName(const char* szShortName);
```
qaisjp:
```suggestion
    CResourceFile* GetFileFromShortName(const char* szShortName);

```

Pirulax:
Well, actually, I'm pretty sure this doesn't do much. The only way it could, is to store an std::string somewhere, and put std::string_view into the set, because, now I keep allocating strings left and right, which is way worse than the linear complexity of the old list approach.

Pirulax:
Well, actually, I'm pretty sure this doesn't do much. The only way it could, is to store an `std::string` somewhere, and put `std::string_view` into the `set`, because, now I keep allocating strings left and right, which is way worse than the linear complexity of the old list approach. 

Ooor, use `std::vector<std::string>` to store all names, and do linear complexity search, but without allocating strings. 

Edit: [Stackoverflow](https://stackoverflow.com/questions/20317413/what-are-transparent-comparators) helped me, as usual.

Pirulax:
This commit in theory fixes the above mentioned `std::string` copy issue, and allows for direct comparison of `std::string` and `const char*`

Pirulax:
Rebased.

PlatinMTA:
Damn that seems like a really good improvement. Really looking forward to this.

Pirulax:
It's not that big of a deal to be honest. But, if it compiles, you can download the release version, and run a benchmark. The improvement is especially big if there's of exported functions.

Pirulax:
My git `skills` 😂😂
Non existent. 
Also, VSC's git was throwing errors at me, so that's a part of it.

Also, remember, this performance improvement is only on the client. I didn't touch the server.. We'll I've tried, but, tbh, there it's not that noticeable, while here, ppl do call these things every frame..

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -393,17 +393,19 @@ int CLuaResourceDefs::GetResourceExportedFunctions(lua_State* luaVM)
...
-        for (; iterd != pResource->IterEndExportedFunctions(); iterd++)
+        unsigned int uiIndex = 0;
+
+        for (const auto& strName : pResource->GetExportedFunctionNames())
         {
             lua_pushnumber(luaVM, ++uiIndex);
-            lua_pushstring(luaVM, (*iterd)->GetFunctionName());
-            lua_settable(luaVM, -3);
+            lua_pushstring(luaVM, strName.c_str());
+            lua_settable(luaVM, -3); // Pack it into a key-value pair
```
qaisjp: this line isn't formatted correctly. Select the line -> Edit -> Advanced -> Format Selection

PR #1518, opened 2020-06-12
Add getRoofPosition

Pirulax:
Since there's `getGroundPosition` I added this as well. I'd say its pretty neat.

I coudn't use the new parser because it can only be used in classes which are derived from `CLuaDefs`

### DIFF
--- a/Client/game_sa/CWorldSA.cpp
+++ b/Client/game_sa/CWorldSA.cpp
@@ -443,6 +443,14 @@ float CWorldSA::FindGroundZFor3DPosition(CVector* vecPosition)
     return fReturn;
 }
 
+float CWorldSA::FindRoofZFor3DCoord(CVector* pvecPosition, bool* pbOutResult)
+{
+    DEBUG_TRACE("FLOAT CWorldSA::FindRoofZFor3DCoord(float x, float y, float z, bool * pbOutResult)");
+
+    auto CWorld_FindRoofZFor3DCoord = (float(__cdecl*)(float, float, float, bool*))0x569750;
+    return CWorld_FindRoofZFor3DCoord(pvecPosition->fX, pvecPosition->fY, pvecPosition->fZ, pbOutResult);
+}
+
 void CWorldSA::LoadMapAroundPoint(CVector* vecPosition, FLOAT fRadius)
 {
     DEBUG_TRACE("VOID CWorldSA::LoadMapAroundPoint(CVector * vecPosition, FLOAT fRadius)");
--- a/Client/game_sa/CWorldSA.h
+++ b/Client/game_sa/CWorldSA.h
@@ -68,6 +68,7 @@ public:
     BYTE  GetLevelFromPosition(CVector* vecPosition);
     float FindGroundZForPosition(float fX, float fY);
     float FindGroundZFor3DPosition(CVector* vecPosition);
+    float FindRoofZFor3DCoord(CVector* pvecPosition, bool* pbOutResult);
     void  LoadMapAroundPoint(CVector* vecPosition, float fRadius);
     bool  IsLineOfSightClear(const CVector* vecStart, const CVector* vecEnd, const SLineOfSightFlags flags);
     bool  HasCollisionBeenLoaded(CVector* vecPosition);
@@ -93,7 +94,6 @@ public:
      * \todo Add ExtinguishAllCarFiresInArea (0x566950)
      * \todo Add FindLowestZForCoord (0x5697F0)
      * \todo Add FindNearestObjectOfType (see 0x46D5FD)
-     * \todo Add FindRoofZFor3DCoord (0x569750)
      * \todo Add GetIsLineOfSightClear (0x56A490)
      * \todo Add ProcessVerticalLine (0x5674E0)
      * \todo Add RemoveReferencesToDeletedObject (0x565510)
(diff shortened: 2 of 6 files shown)


Lpsd:
I think `getCeilingPosition` is a better name (any references to `roof` should also be updated to `ceiling`)

Pirulax:
Idk, the name in GTA source code is roof, and it sounds about right, since 'ceiling' is the inside, 'roof' is the outside

CrosRoad95:
use new lua parser

Pirulax:
> use new lua parser

Did you read the description? 🙂

Client/game_sa/CWorldSA.cpp
```diff
@@ -443,6 +443,31 @@ float CWorldSA::FindGroundZFor3DPosition(CVector* vecPosition)
...
+    DEBUG_TRACE("FLOAT CWorldSA::FindRoofZFor3DCoord(CVector * vecPosition, bool * pbOutResult)");
+
+    DWORD dwFunc  = FUNC_FindRoofZFor3DCoord;
+    FLOAT fReturn = 0;
+
+    FLOAT fX = pvecPosition->fX;
+    FLOAT fY = pvecPosition->fY;
+    FLOAT fZ = pvecPosition->fZ;
+
+    _asm
```
qaisjp: In new code, we no longer allow asm blocks in methods that already have C++ code. Cast `FUNC_FindRoofZFor3DCoord` to a "function pointer" and then call that function.

qaisjp (dismissed):

Einheit-101:
What does this exactly get? The next surface above the player?

Client/game_sa/CWorldSA.cpp
ghost: Yes.
```
auto CWorld_FindRoofZFor3DCoord = (float(__cdecl*)(CVector, bool*))0x0569750;
return CWorld_FindRoofZFor3DCoord(*pvecPosition, pbOutResult);
```

Pirulax:
> What does this exactly get? The next surface above the player?

From my tests its the same as a 'processLineOfSight' from above you. 
Actually, yes:

```cpp
float
CWorld::FindRoofZFor3DCoord(float x, float y, float z, bool *found)
{
	CColPoint point;
	CEntity *ent;
	if(ProcessVerticalLine(CVector(x, y, z), 1000.0f, point, ent, true, false, false, false, true, false, nil)) {
		if(found) *found = true;
		return point.point.z;
	} else {
		if(found == nil)
			printf("THERE IS NO MAP BELOW THE FOLLOWING COORS:%f %f %f. (FindGroundZFor3DCoord)\n", x, y,
			       z);
		if(found) *found = false;
		return 20.0f;
	}
}
```

Client/game_sa/CWorldSA.cpp
Pirulax: :(
I was so happy I learned to copy paste ASM code
Sure, no problem btw. I'll do it.

Pirulax: Why is it `__cdecl*`, instead of `__cdecl`? Im just curious

Pirulax:
@Lpsd suggested me to change the name to "getCeilingPositon". What others think?

qaisjp:
> @Lpsd suggested me to change the name to "getCeilingPositon". What others think?

Does it get the roof or the ceiling?

Lpsd:
> > @Lpsd suggested me to change the name to "getCeilingPositon". What others think?
> 
> Does it get the roof or the ceiling?

The ceiling (position of nearest solid surface directly above the player).

Client/game_sa/CWorldSA.cpp
sbx320: Since `0x0569750` is the address of the function, we need to construct a function pointer and not the function itself. You can't put a function inside a variable, but a pointer works fine.

Pirulax:
> > > @Lpsd suggested me to change the name to "getCeilingPositon". What others think?
> > 
> > 
> > Does it get the roof or the ceiling?
> 
> The ceiling (position of nearest solid surface directly above the player).

It gets the top Z coord of the closet solid above the player, as it can be seen in the code snippet.

Lpsd:
> It gets the top Z coord of the closet solid above the player, as it can be seen in the code snippet.

So let's say the player is at `0, 0, 0` and there is a cube with a width & height of 2, with its central point being at `0, 0, 2` - would this function return 1 or 3 for the z coordinate?

Pirulax:
3.

Client/game_sa/CWorldSA.h
```diff
@@ -16,6 +16,7 @@
 #define FUNC_ProcessLineOfSight                             0x56BA00 // ##SA##
 #define FUNC_FindGroundZFor3DCoord                          0x5696C0 // ##SA##
 #define FUNC_FindGroundZForCoord                            0x569660 // ##SA##
+#define FUNC_FindRoofZFor3DCoord                            0x569750 // ##SA##
 #define FUNC_CTimer_Stop                                    0x561AA0 // ##SA##
 #define FUNC_CTimer_Update                                  0x561B10 // ##SA##
```
ghost: This is just an old bad design. There's no need to define a macro for function address, you can just hardcode it when calling the function in `*SA.cpp` file.

This is enough and readible:
```
auto CWorld_FindRoofZFor3DCoord = (float(__cdecl*)(CVector, bool*))0x0569750;
```
Anyways, the PR looks done, so let's just merge it.

ghost (approved):

Client/game_sa/CWorldSA.h
Pirulax: I just did it because of consistency. I know its the same as just putting it there, but it seemed like a better idea to follow the convention.

Client/game_sa/CWorldSA.cpp
Pirulax: Ah, that makes sense, thanks!

Pirulax:
Rebased, and fixed up.

PR #1520, opened 2020-06-14
Fix #1228 - Move exe patching to loader

### LINKED ISSUE #1228: Move exe patching to loader

If a user install GTA after MTA, MTA will not patch their GTA exe and will fail to start.

Move (or copy) exe patching from the installer and to the loader.

**Additional context**

We have received a lot of support messages relating to this. It would be better to instead just attempt to patch on failure instead of presenting some error message.

Current source is here: https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/installer/nightly.nsi

[This line](https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/installer/nightly.nsi#L1212) refers to "vpatch", which might be this: https://www.tibed.net/vpatch/

ccw808:
fixes #1228

### DIFF
--- a/Client/loader/CInstallManager.cpp
+++ b/Client/loader/CInstallManager.cpp
@@ -99,6 +99,16 @@ void CInstallManager::InitSequencer()
         CR "update_end: "                                                          ////// End of 'update game' //////
         CR "            CALL SwitchBackFromTempExe "                               //
         CR " "                                                                     //
+        CR "gta_version_check:"                                                    ////// Start of 'gta version check' //////
+        CR "            CALL ProcessGtaVersionCheck "                              //
+        CR "            IF LastResult == ok GOTO gta_version_end: "                //
+        CR "            IF LastResult == quit GOTO do_quit: "                      //
+        CR " "                                                                     //
+        CR "            CALL ChangeToAdmin "                                       // If changes failed, try as admin
+        CR "            IF LastResult == ok GOTO gta_version_check: "              //
+        CR "            GOTO do_quit: "                                            // If changes as admin failed, then quit
+        CR "gta_version_end: "                                                     ////// End of 'gta version check' //////
+        CR " "                                                                     //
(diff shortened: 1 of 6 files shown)


qaisjp:
Haven't read through the code properly yet, but I noticed that the original patcher code is still there - is that intentional? 

Thought about fetching the patch from a server instead of shipping it with every update? (Granted, it is only 4.5 MB.)

Client/loader/CInstallManager.cpp
```diff
@@ -527,6 +542,81 @@ void MigrateFile(const SString& strFilenameOld, const SString& strFilenameNew)
...
+        return "ok";
+    }
+
+    // Check required GTA file is correct
+    if (GenerateHashHexStringFromFile(EHashFunctionType::MD5, strPatchBase) != strPatchBaseMd5)
+    {
+        AddReportLog(5053, SString("_ProcessGtaVersionCheck: Incorrect file '%s'", *strPatchBase));
+        SString strMessage(_("MTA:SA cannot continue because the following files are incorrect:"));
+        strMessage += "\n\n" + strPatchBase;
+        BrowseToSolution("gengta_pakfiles", ASK_GO_ONLINE, strMessage);
```
qaisjp, line 582: I'm guessing https://updatesa.mtasa.com/sa/trouble/?tr=gengta_pakfiles and the one below will be created when this is merged?

qaisjp (approved):
LGTM

ccw808:
> Haven't read through the code properly yet, but I noticed that the original patcher code is still there - is that intentional?

Yes
 
> Thought about fetching the patch from a server instead of shipping it with every update? (Granted, it is only 4.5 MB.)

Reasons for not downloading are: 1) Launcher currently doesn't depend on internet access, so there might be firewall issues and 2) data files are only in the full update. Slim update is used if client already has certain files.

Client/loader/CInstallManager.cpp
ccw808: Correct

PR #1530, opened 2020-06-19
Allow hourOn to be greater than hourOff for engineSetModelVisibleTime

StrixG:
Some San Andreas night objects appear at 23 and disappear "next morning". It was impossible to set similar behavior.
```lua
engineSetModelVisibleTime(9934, 23, 6) -- Before: visible from 6 to 23. After: visible from 23 to 6.
```


qaisjp (approved):
makes sense, lgtm, :shipit:

PR #1534, opened 2020-06-22
Fix #1443 - trim teaEncode trailing zeroes

### LINKED ISSUE #1443: Tea encoding bug
**Describe the bug**

My version: Multi Theft Auto v1.5.7-release-20520
The problem occurs when using the encodeString, decodeString, teaDecode and teaEncode functions. The characters "\0" are added to the decrypted string. Exists only on the client side, and when the number of characters is not a multiple of 4.

**To reproduce**
Use the `iprint(decodeString("tea",encodeString("tea",1,{key="123"}),{key="123"}))`

**Expected behaviour**
Return the correct value

**Screenshots**

none

**Version**

Client Windows
Copyright (C) 2003 - 2020 Multi Theft Auto

**Additional context**

Example:
`iprint(decodeString("tea",encodeString("tea",1,{key="123"}),{key="123"}))`
returns "1\0\0\0" instead of "1"

More examples:
String before encoding - String after decoding

```
"1" - "1\0\0\0"
"12" - "12\0\0"
"123" - 123\0"
"1234" - "1234"
"1234567" - "1234567\0"
"12345678" - "12345678"
```

Pirulax:
So basically, with the introduction of sbx's new parser, we started returning strings with `lua_pushlstring`, which takes a length as an argument, thus pushes the whole string along with padding. The old method used `lua_pushstring` which used `strlen`, thus god rid of the padding.
This PR fixes #1443 .

Edit: test resource: [utf1443.zip](https://github.com/multitheftauto/mtasa-blue/files/4813800/utf1443.zip)

### DIFF
--- a/Shared/sdk/SharedUtil.Hash.hpp
+++ b/Shared/sdk/SharedUtil.Hash.hpp
@@ -772,7 +772,7 @@ namespace SharedUtil
         delete[] strbuf;
     }
 
-    void TeaDecode(const SString& str, const SString& key, SString* out)
+    void TeaDecode(const SString& str, const SString& key, SString* out, const bool trimPaddingAtEnd)
     {
         unsigned int v[2];
         unsigned int w[2];
@@ -818,5 +818,22 @@ namespace SharedUtil
 
         out->assign((char*)buffer, numPasses * 4);
         delete[] buffer;
+
+        if (trimPaddingAtEnd)
+        {
+            // Check last block for 0 padding, and delete
+
+            auto iter = out->end() - 1;
+            if (*iter != 0)
+                return;
+
+            const auto blockEnd = out->end() - 4;
+            while (iter > blockEnd && *--iter == 0);
+
+            // Loop goes past the non-0 char, so make sure we only delete 0
+            iter++;
+
+            out->erase(iter, out->end()); // At this point its 100% theres a 0 padding, so we can delete that
+        }
     }
 }            // namespace SharedUtil
--- a/Shared/sdk/SharedUtil.Hash.hpp
+++ b/Shared/sdk/SharedUtil.Hash.hpp
@@ -820,18 +820,23 @@ namespace SharedUtil
         delete[] buffer;
 
 
-        // Check last block for 0 padding, and erase if there's any
+        // Delete all 0's from the end
+        // As of now, even if the user passed it in
+        // We'll delete it, because there's no way
+        // We can know if the user added the padding or not.
         {
             auto iter = out->end() - 1;
             if (*iter != 0)
                 return;
 
-            const auto blockEnd = out->end() - 4;
-            while (iter > blockEnd && *--iter == 0);
+            // Traverse string until the beginning,
+            // Stop at the first non-zero char
+            while (iter > out->begin() && *--iter == 0);
 
-            // Loop goes past the non-0 char, so make sure we only delete 0
+            // Go back to the zero char
             iter++;
 
+            // Erase 0s
             out->erase(iter, out->end());
         }
     }
--- a/Shared/sdk/SharedUtil.Hash.hpp
+++ b/Shared/sdk/SharedUtil.Hash.hpp
@@ -824,20 +824,20 @@ namespace SharedUtil
         // As of now, even if the user passed it in
         // We'll delete it, because there's no way
         // We can know if the user added the padding or not.
+
+        // Note: If we get there, there's at least 1 block(4 chars)
+        //       before out->end()
         {
-            auto iter = out->end() - 1;
-            if (*iter != 0)
-                return;
+            // Traverse the string until the beginning, or until the fist non-0 char
+            auto iter = out->end();
+            for (; iter != out->begin(); iter--)
+            {
+                if (*std::prev(iter) != 0)
+                    break;
+            }
 
-            // Traverse string until the beginning,
-            // Stop at the first non-zero char
-            while (iter > out->begin() && *--iter == 0);
-
-            // Go back to the zero char
-            iter++;
-
-            // Erase 0s
-            out->erase(iter, out->end());
+            if (iter != out->end()) 
+                out->erase(iter, out->end()); // Erase all 0s
         }
     }
 }            // namespace SharedUtil
--- a/Shared/sdk/SharedUtil.Hash.hpp
+++ b/Shared/sdk/SharedUtil.Hash.hpp
@@ -836,7 +836,7 @@ namespace SharedUtil
                     break;
             }
 
-            if (iter != out->end()) 
+            if (iter != out->end())
                 out->erase(iter, out->end()); // Erase all 0s
         }
     }
--- a/Shared/sdk/SharedUtil.Hash.hpp
+++ b/Shared/sdk/SharedUtil.Hash.hpp
@@ -696,7 +696,7 @@ namespace SharedUtil
             return GenerateHashHexString(hashFunction, NULL, 0);
     }
 
-    void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
+    inline void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
     {
         unsigned int v0 = v[0], v1 = v[1], i, sum = 0;
         unsigned int delta = 0x9E3779B9;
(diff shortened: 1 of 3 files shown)


Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -818,5 +818,22 @@ namespace SharedUtil
...
+        if (trimPaddingAtEnd)
+        {
+            // Check last block for 0 padding, and delete
+
+            auto iter = out->end() - 1;
+            if (*iter != 0)
+                return;
+
+            const auto blockEnd = out->end() - 4;
+            while (iter > blockEnd && *--iter == 0);
```
qaisjp: Please use more lines

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -772,7 +772,7 @@ namespace SharedUtil
         delete[] strbuf;
     }
 
-    void TeaDecode(const SString& str, const SString& key, SString* out)
+    void TeaDecode(const SString& str, const SString& key, SString* out, const bool trimPaddingAtEnd)
```
qaisjp: Is this necessary? The only place where this is false is in `Shared/sdk/SharedUtil.Tests.hpp`, so it seems like this is unnecessary.

qaisjp (dismissed):

Shared/sdk/SharedUtil.Hash.hpp
Pirulax: I feel like this is perfectly readable. Not that anyone reads that code anyways.

Shared/sdk/SharedUtil.Hash.hpp
Pirulax: No, it's not.

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -818,5 +818,22 @@ namespace SharedUtil
 
         out->assign((char*)buffer, numPasses * 4);
         delete[] buffer;
+
+        if (trimPaddingAtEnd)
+        {
+            // Check last block for 0 padding, and delete
+
+            auto iter = out->end() - 1;
```
sbx320: What if `out.begin() == out.end()` (aka. the input is empty)?

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -818,5 +818,22 @@ namespace SharedUtil
...
+            if (*iter != 0)
+                return;
+
+            const auto blockEnd = out->end() - 4;
```
sbx320: This doesn't seem correct. If I encode `abc\0\0\0`, then the decode result (pre-trim) would be `abc\0\0\0\0\0`. After trimming it ends up being `abc\0`, which is incorrect. We should trim the exact amount of excess padding that was added previously..

sbx320 (changes requested):

Shared/sdk/SharedUtil.Hash.hpp
ccw808: Previous version did the same thing. I think that should be looked at in a new pull request

Shared/sdk/SharedUtil.Hash.hpp
Pirulax: The block size is 4. If its smaller than that the function returns before ever reaching this point.

See code:
```cpp
        // Count the number of passes that we need
        int numBlocks = str.length() / 4;
        int numPasses = numBlocks - 1;

        if (numPasses <= 0)
            return;
```

Shared/sdk/SharedUtil.Hash.hpp
Pirulax: I feel like doing `strlen` on a 10-15 mb string isn't particularly efficient.

Pirulax: So it this the correct behavior?

Pirulax: @sbx320  I don't see a way to achieve what you want. We can't possible figure out if it was padding, or the user has added the `\0\0\0`'s. Please correct me if there's a way to do so.

Pirulax: I can trim all `0`'s, not just the ones in the last block. That would leave us with a string that has no padding. But if the user added `0`'s we can't figure out on the other end if that was padding, or it was originally in the string

Pirulax:
I can't connect to my 1.5.6 server, for some stupid reason(a resource sets min client version, and I cant figure out which one)

Please, could someone run a few tests, and post the results here on 1.5.6 server and client?

Test values:
- 123
- 1234
- \0\0Hello
- 123\0
- hithere\0\0\0\0
- \0hi\0
- \0\0\0\0\0hi\0\0
- 123\0\0\0\0

They key can be anything. Let it be.. "c++20"

ccw808:
@Pirulax mtaserver.conf:
```
    <minclientversion></minclientversion>
    <minclientversion_auto_update>0</minclientversion_auto_update>
```
Edit: Oh a resource sets it - Nevermind

Shared/sdk/SharedUtil.Hash.hpp
qaisjp: This requires less thinking
```
while (iter > blockEnd || *iter == 0)
    iter--;
```

Pirulax:
btw, the problem is that, now I've fixed `en/decodeString` as well, which has been returning wrong values since it was introduced. Should we keep backwards compatibility? Or just hope nobody tries comparing decoded strings?

Pirulax:
Btw, since the old one used `strnlen` it strips all padding from the end, so I'll modify this one to be like that.

Pirulax:
> @Pirulax mtaserver.conf:
> 
> ```
>     <minclientversion></minclientversion>
>     <minclientversion_auto_update>0</minclientversion_auto_update>
> ```
> 
> Edit: Oh a resource sets it - Nevermind

Actually, I'm not sure if it was a resource. I think this'll help, thanks!

Pirulax:
New resource to test: [utf1443.zip](https://github.com/multitheftauto/mtasa-blue/files/4819898/utf1443.zip)

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -818,5 +818,26 @@ namespace SharedUtil
 
         out->assign((char*)buffer, numPasses * 4);
         delete[] buffer;
+
+
```
qaisjp, line 822:
```suggestion
```

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -696,7 +696,7 @@ namespace SharedUtil
             return GenerateHashHexString(hashFunction, NULL, 0);
     }
 
-    void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
+    inline void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```
qaisjp, line 699: This inlining is not relevant to the fix.

Pirulax:
We need to fix the padding in teaEncode, because with binary files 0's at the file end is totally acceptable.

Shared/sdk/SharedUtil.Hash.hpp
qaisjp:
```suggestion
   void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -696,7 +696,7 @@ namespace SharedUtil
             return GenerateHashHexString(hashFunction, NULL, 0);
     }
 
-    void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
+   void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```
qaisjp:
```suggestion
    void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```

Shared/sdk/SharedUtil.Hash.hpp
```diff
@@ -710,7 +710,7 @@ namespace SharedUtil
         w[1] = v1;
     }
 
-    void decodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
+    inline void decodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```
qaisjp, line 713:
```suggestion
    void decodeXtea(unsigned int* v, unsigned int* w, unsigned int* k)
```

Pirulax:
Perhaps, maybe add a noTrim argument?

qaisjp:
Lets try and fix the regression in this PR and we can discuss other problems with teaEncode in a separate issue.

ccw808:
> 
> 
> Lets try and fix the regression in this PR and we can discuss other problems with teaEncode in a separate issue.

Agreed. Let's get this fixed yesterday

PR #1535, opened 2020-06-22
Refactor teaEncode/teaDecode functions to use std::string_view

Pirulax:
This PR replaces teaEncode/teaDecode input from `std::string` to `std::string_view`. Also adds a new macro `SDK_WITH_TEA`, for compatibility reasons (Game and multiplayer SA still use c++14, and `string_view` is c++17).

Also, now it uses `std::move` in the lambda capture list for `en/decodeString` for improved performance.

### DIFF
--- a/Shared/sdk/CScriptArgReader.h
+++ b/Shared/sdk/CScriptArgReader.h
@@ -593,21 +593,21 @@ public:
     {
         switch (lua_type(m_luaVM, m_iIndex))
         {
-        case LUA_TNUMBER:
-        case LUA_TSTRING:
-        {
-            outValue = {
-                lua_tostring(m_luaVM, m_iIndex),
-                lua_strlen(m_luaVM, m_iIndex)
-            };
+            case LUA_TNUMBER:
+            case LUA_TSTRING:
+            {
+                outValue = {
+                    lua_tostring(m_luaVM, m_iIndex),
+                    lua_strlen(m_luaVM, m_iIndex)
+                };
 
-            break;
-        }
-        default:
-        {
-            outValue = { nullptr, 0 };
-            SetTypeError("string");
-        }
+                break;
+            }
+            default:
+            {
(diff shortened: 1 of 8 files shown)


Pirulax:
Fail.. I totally forgot about me pushing the commit, and I soft reseted.. nvm...

Shared/sdk/CScriptArgReader.h
```diff
@@ -586,6 +586,33 @@ class CScriptArgReader
...
+            };
+
+            break;
+        }
+        default:
+        {
+            outValue = { nullptr, 0 };
+            SetTypeError("string");
+        }
+        }
```
StrixG: Corrected formatting:
```suggestion
        switch (lua_type(m_luaVM, m_iIndex))
        {
            case LUA_TNUMBER:
            case LUA_TSTRING:
            {
                outValue = {
                    lua_tostring(m_luaVM, m_iIndex),
                    lua_strlen(m_luaVM, m_iIndex)
                };

                break;
            }
            default:
            {
                outValue = { nullptr, 0 };
                SetTypeError("string");
            }
        }
```

qaisjp:
Is it possible to `#if` based on the the C++ version instead of a new macro?

Pirulax:
Yes indeed, it is. Good idea, not sure why I didnt do it that way.

Pirulax:
Update: It's not, as MSVC's cpp version macro is fucked on some older versions, and I dont want to introduce inconsistencies across compilers.

botder:
> Update: It's not, as MSVC's cpp version macro is fucked on some older versions, and I dont want to introduce inconsistencies across compilers.

Can you update this pull request, please? We've had the commit 6d4e8adb843adbb77d9627c1c74c3d90d43c96a9 (https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170) since.

PR #1537, opened 2020-06-22
Refactor Lua manager to use std::vector for m_VirtualMachines

Pirulax:
Improves performance by using `std::vector` instead of a list.
as you can see from the commits, I first opted to use `CFastHashSet`, but, it's unnecessary, since we only use `find()` once. (When deleting the VM).
Most of the time we iterate thru the vector (`DoPulse()`), where the vector clearly wins in any scenario, and by a big margin.

Tldr: use `vector` instead of `list`, to make things faster, for the expense of `RemoveVM` getting slower.


Pirulax:
I tested the server, works fine. The client should work just as fine.

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -72,7 +71,9 @@ bool CLuaManager::RemoveVirtualMachine(CLuaMain* pLuaMain)
         m_pRegisteredCommands->CleanUpForVM(pLuaMain);
 
         // Remove it from our list
-        m_virtualMachines.remove(pLuaMain);
+        const auto iter = std::find(m_VirtualMachines.begin(), m_VirtualMachines.end(), pLuaMain);
+        assert(iter != m_VirtualMachines.end()); // Make sure it exists
```
qaisjp: Our auto-formatting does unusual things to inline comments, and this comment position is too sane, so I think the code is formatted incorrectly. Try formatting selection.

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -130,23 +128,25 @@ CLuaMain* CLuaManager::GetVirtualMachine(lua_State* luaVM)
...
 
     // Find a matching VM in our map
-    CLuaMain* pLuaMain = MapFindRef(m_VirtualMachineMap, luaVM);
-    if (pLuaMain)
-        return pLuaMain;
+    {
+        CLuaMain* pLuaMain = MapFindRef(m_VirtualMachineMap, luaVM);
+        if (pLuaMain)
+            return pLuaMain;
+    }
```
qaisjp:
```suggestion
    CLuaMain* pLuaMain = MapFindRef(m_VirtualMachineMap, luaVM);
    if (pLuaMain)
        return pLuaMain;
```
Is the change in indentation+scoping necessary here?

Client/mods/deathmatch/logic/lua/CLuaManager.h
```diff
@@ -33,8 +33,7 @@ class CLuaManager
     void      OnLuaMainOpenVM(CLuaMain* pLuaMain, lua_State* luaVM);
     void      OnLuaMainCloseVM(CLuaMain* pLuaMain, lua_State* luaVM);
 
-    std::list<CLuaMain*>::const_iterator IterBegin() { return m_virtualMachines.begin(); };
-    std::list<CLuaMain*>::const_iterator IterEnd() { return m_virtualMachines.end(); };
+    auto GetLuaMains() const { return m_VirtualMachines; }
```
qaisjp: Why did you pick a return `auto` here? Seems like an unusual place to use it, especially since `m_VirtualMachines` doesn't have a particularly difficult type to write. I'm happy to keep `auto` if you think we should keep it, though.

Server/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -79,12 +79,13 @@ bool CLuaManager::RemoveVirtualMachine(CLuaMain* pLuaMain)
...
         // Delete it unless it is already
         if (!pLuaMain->BeingDeleted())
-        {
             delete pLuaMain;
-        }
 
         // Remove it from our list
-        m_virtualMachines.remove(pLuaMain);
+        const auto iter = std::find(m_VirtualMachines.begin(), m_VirtualMachines.end(), pLuaMain);
+        assert(iter != m_VirtualMachines.end()); // Make sure it exists
```
qaisjp: Format selection, as mentioned above

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -190,19 +190,17 @@ void CLuaModule::_RegisterFunctions(lua_State* luaVM)
...
+    for (CLuaMain* luaMain : m_pLuaModuleManager->GetLuaManager()->GetLuaMains())
     {
-        lua_State*                luaVM = (*liter)->GetVM();
-        vector<SString>::iterator iter = m_Functions.begin();
-        for (; iter != m_Functions.end(); ++iter)
+        lua_State* luaVM = luaMain->GetVM();
+        for (const SString& strFuncName : m_Functions)
         {
-            // points function to nill
+            // Remove function global variable
```
qaisjp:
```suggestion
            // Set function global variable to nil
```

Client/mods/deathmatch/logic/lua/CLuaManager.h
```diff
@@ -52,6 +51,6 @@ class CLuaManager
     CRegisteredCommands* m_pRegisteredCommands;
 
     CFastHashMap<lua_State*, CLuaMain*> m_VirtualMachineMap;
-    std::list<CLuaMain*>                m_virtualMachines;
+    std::vector<CLuaMain*>              m_VirtualMachines;
```
qaisjp, line 54: You are changing type from `std::list` to `std::vector`.

http://www.cplusplus.com/reference/vector/vector/erase/ says this:

> Because vectors use an array as their underlying storage, erasing elements in positions other than the vector end causes the container to relocate all the elements after the segment erased to their new positions. This is generally an inefficient operation compared to the one performed for the same operation by other kinds of sequence containers (such as list or forward_list).

Have you taken this into consideration?

qaisjp (changes requested):
Left a few comments, mostly just a few small nits and two questions. It looks fine, otherwise.

Client/mods/deathmatch/logic/lua/CLuaManager.h
Pirulax: Actually I shouldve returned `const auto&`.
The reason is that it doesnt matter what the type of `m_VirtualMachines` is. Its known to be an STL container, and I find this easier to read and maintain than:
```cpp
const std::vector<CLuaMain*>& GetLuaMains() const { return m_VirtualMachines; }
```

Client/mods/deathmatch/logic/lua/CLuaManager.h
Pirulax: I did.
But, `std list` has a huge overhead on small datatypes, and heap allocates every.
While deliting from a list is `O(n)`, in reality that `O(n)` is far slower than vectors `O(n^2)`, because while std vector will easily be cached, iterating thru it is very fast, compared to std::list, thanks to the CPU cache.
Also, copying it is really fast as well, since we're talking about pointers, which are 4 bytes. Even in the worst case, this will take maybe n * 10 clock cycles to copy, which is very fast, compared to `std::list` where we need to wait a lot more.

Tldr, yes I did, and std::vector will always be faster (if `sizeof(T)` is small (like a pointer is 4 bytes))

Client/mods/deathmatch/logic/lua/CLuaManager.h
qaisjp: Okay

Client/mods/deathmatch/logic/lua/CLuaManager.h
qaisjp: Okay 👍

Server/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -93,30 +94,29 @@ bool CLuaManager::RemoveVirtualMachine(CLuaMain* pLuaMain)
 
 void CLuaManager::OnLuaMainOpenVM(CLuaMain* pLuaMain, lua_State* luaVM)
 {
-    MapSet(m_VirtualMachineMap, pLuaMain->GetVirtualMachine(), pLuaMain);
```
qaisjp, line 96: It seems like MapSet and MapRemove is the standard way to interact with CFastHashMap. What's the purpose of the change here and in OnLuaMainCloseVM?

Pirulax: Because `MapSet`, and `MapInsert` as useless. Their only purpose was to provide a better interface for setting/insertion, but since C++11 there are initializer lists, which make those functions totally useless.
That is, pre c++11:
```cpp
m_VirtualMachineMap.insert(std::pair<....>(...))
``` 
Now:
```cpp
m_VirtualMachineMap.insert({...})
```

As for `MapSet`.. I really dont see why it exists in the first place, it is just a simple:
```cpp
m_VirtualMachineMap[key] = value
```

I strongly disagree with the usage of these functions in new code.

qaisjp: Isn't `m_VirtualMachineMap[key] = value` cleaner than `m_VirtualMachineMap.insert({key, value})`?

Pirulax: It is, thanks!

Pirulax:
TODO: (Theory) List was originally used because it might modify itself when deleting the VM or something.. Must check.
Well, had a quick look: It might happen if a timer callback stops a resource. Sadly this can't be solved easily.

PR #1538, opened 2020-06-23
Add translate and discord badges to readme

Disinterpreter:
GitHub is a social coding (as they say) so we must to make an effort for make it really social.

Small badges make the repo beautiful and contain important things.

Now it looks that:
image

P.S squash me if you will pull it.


qaisjp:
Looks great although I think we could do without the website and wiki badges. Translate is a great idea as that is something that a user can directly contribute to. (Users can obv contribute to wiki but it doesn't feel like it fits.)

Disinterpreter:
The first idea was add only Discord. But I thought to add more badges. I was ready with removing many badges, I made a choice for contributors. But we left one badge and it's good.

You can look changes here: https://github.com/Disinterpreter/mtasa-blue

qaisjp:
Can we have the header on top (and the badges underneath the header)

Disinterpreter:
> 
> 
> Can we have the header on top (and the badges underneath the header)

I looked away to other repos and I saw, they use badges under the header. Yes, you are right.

qaisjp:
one last thing - please can we have a blank line below the header (just an empty line) so that it's vaild markdown too?

Disinterpreter:
> 
> 
> one last thing - please can we have a blank line below the header (just an empty line) so that it's vaild markdown too?

Sorry, but I think no. Or I don't understand. 
You can look at other repos like:
image
image

qaisjp:
it is in markdown only, it won't affect the render

Disinterpreter:
like that?

StrixG (approved):

qaisjp (approved):

PR #1541, opened 2020-06-26
Add prerequisites to README.md for Windows build

Deihim007:
I added Prerequisites for windows build. Although removing all that and pointing a link to https://wiki.multitheftauto.com/wiki/Compiling_MTASA Could also be a better solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,12 @@ If not using Visual Studio 2017, download and install the [EditorConfig](https:/
 
 ### Build Instructions
 #### Windows
+
+Prerequisites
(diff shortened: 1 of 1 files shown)


qaisjp:
I agree that maybe moving all this elsewhere might be better, but for now we can still keep these changes.

qaisjp (approved):

mirh:
Why are you linking an older SDK?
[This](https://www.microsoft.com/en-us/download/details.aspx?id=6812) is the last one.

PR #1542, opened 2020-06-27
Optimize split: Use string_view, and don't buffer if delimiter is a char

Pirulax:
This PR removes the redundant buffer(`SString`, and then `new char[]`).
Now it only buffers `if delimenter.length() > 1`
Now it uses std::make_unique(for `char[]` buffer)

Unit tests [here](https://repl.it/@JohnLugo/SimpleTinyOpengroup#main.cpp) for the no-buffer case


Pirulax:
Commit messages aren't perfect.
I'll rebase before merging

qaisjp:
Nice! And thank you for including the link to your tests.

Did you measure to see the difference in speed in Lua?

botder:
This pull request must be updated to use the new argument parser. Afterwards we can merge it.

PR #1547, opened 2020-06-29
Add social buttons to main menu

Deihim007:
Adds social buttons to main menu:
new

Thanks to @Dutchman101 for the idea and @turret001, @DNL291 and @Tut for their suggestions.


Pirulax:
Use `std::unique_ptr` instead of 'new' and 'delete'
(you might need to `#include <memory>` for that)

I'd suggest you replacing the copy paste code with this:
```cpp
{
    const std::pair<const char*, GUI_CALLBACK>socialButtonDatas[]{
        {"discord", GUI_CALLBACK(&CMainMenu::OnDiscordButtonClick, this)},
        {"github",  GUI_CALLBACK(&CMainMenu::OnGithubButtonClick, this)},
        {"youtube", GUI_CALLBACK(&CMainMenu::OnYoutubeButtonClick, this)},
        {"twitter", GUI_CALLBACK(&CMainMenu::OnTwitterButtonClick, this)},
    };

    float offsetX = 0.01f;
    size_t i = 0;
    const CVector2D buttonSize((NATIVE_SOCIAL_ICON_X / NATIVE_RES_X) * m_iMenuSizeX, (NATIVE_SOCIAL_ICON_Y / NATIVE_RES_Y) * m_iMenuSizeY);
    for (const auto& [name, buttonClickHandler] : socialButtonDatas)
    {

        std::unique_ptr<CGUIStaticImage> button(pManager->CreateStaticImage(m_pCanvas))

        button->LoadFromFile(SString("cgui\\images\\socialset\\%s", name).c_str());
        button->SetProperty("InheritsAlpha", "False");
        button->SetAlpha(0.35f);
        button->SetSize(buttonSize, false);
        button->SetPosition({offsetX * m_iMenuSizeX - buttonSize.fX / 2, 0.97f * m_iMenuSizeY - buttonSize.fY / 2}, false);
        button->SetZOrderingEnabled(false);
        button->SetMouseEnterHandler(GUI_CALLBACK(&CMainMenu::OnSocialButtonHover, this));
        button->SetMouseLeaveHandler(GUI_CALLBACK(&CMainMenu::OnSocialButtonUnhover, this));
        button->SetClickHandler(buttonClickHandler);

        m_socialButtons[i++] = std::move(button);
        offsetX += 0.03f;
    }
}
```
Also, add this to the header:
```cpp
std::unique_ptr<CGUIStaticImage> m_socialButtons[4] = { nullptr };
```

Deihim007 (commented):
Thanks, Done. @Pirulax

T-MaxWiese-T:
Nothing has been uploaded to the YouTube channel for 4 years. How about the MTA website/forum and Facebook?

Pirulax:
Does it work as expected? Im fairly surprised it works. Did you have to fix any errors?
Thanks btw!

FileEX:
I think we should replace youtube button with e.g forum.mtasa.com or mtasa.com, because like @T-MaxWiese-T sayed - Nothing has been uploaded to the youtube channel for a very long time and it's nonsense in my opinion.

qaisjp:
Facebook would be nice

Deihim007:
> Does it work as expected? Im fairly surprised it works. Did you have to fix any errors?
> Thanks btw!

Worked just fine!, only had to add `.png` to `"cgui\\images\\socialset\\%s"` @Pirulax

Deihim007:
> I think we should replace youtube button with e.g forum.mtasa.com or mtasa.com, because like @T-MaxWiese-T sayed - Nothing has been uploaded to the youtube channel for a very long time and it's nonsense in my opinion.

> Facebook would be nice

> Nothing has been uploaded to the YouTube channel for 4 years. How about the MTA website/forum and Facebook?

Just in-case, i added all of them.
image

Deihim007:
new2

Disinterpreter:
> 
> 
> new2

what about reddit?

qaisjp:
I think we should keep only Discord, Twitter, Facebook, and GitHub.

Reasons being:

- website: this link isn't very useful and there's no real call to action
- discord: someone can join the server and chat (an explicit action)
- twitter: someone can follow us for updates (an explicit action)
- facebook: same as twitter
- YouTube: the most dead platform (although there's an explicit action to subscribe there's no real benefit to a user following here)
- GitHub: maybe a good reminder we're open source (for people that know what GitHub is) - I suppose an explicit action would be starring the repo, and subscribing for releases (watching)

(Usually Facebook is updated more or less whenever Twitter is updated. Although in general these days we only use discord announcements and unfortunately have forgotten about other social media.)

I still think the design of these icons are a bit programmer arty and could use a bit of work

Woovie:
I think YouTube is still important and we should have it. When working with platforms like Discord, YouTube subscribers are a number people look at.

qaisjp:
Okay, lets keep YT then.

Off topic for this issue: it would be good to put someone in charge of comms, i.e. to maintain YouTube, the upkeep of our social media in general, and various news channels.

prnxdev:
IMHO this should be loaded from external source (API). Just in case that some link stopped working or something.

Woovie:
No API needed, but yeah we could store a flat file possibly on mtasa.com of a JSON array containing each social link, its associated text, and what image to render. Simple enough to alter in the future. If it fails though, we should have a state that doesn't break the client.

qaisjp:
I wonder if it's possible to use a single imageset file instead of multiple

Client/core/CMainMenu.cpp
```diff
@@ -981,6 +1015,68 @@ bool CMainMenu::OnQuitButtonClick(CGUIElement* pElement)
...
+    return true;
+}
+
+bool CMainMenu::OnGithubButtonClick(CGUIElement* pElement)
+{
+    // Visit the github url
+    ShellExecuteNonBlocking("open", "https://github.com/multitheftauto/mtasa-blue");
+
+    return true;
+}
```
qaisjp, line 1078: There is probably a nicer solution than having many functions. E.g. setting some data on the element that maps to some index, and then you can grab the link out of `socialButtonDatas`.

Pirulax: Probably yes.
iirc `SubscribeEvent` takes a lambda as well (needs testing tho), if thats the case we dont even need these static functions

Pirulax:
> No API needed, but yeah we could store a flat file possibly on mtasa.com of a JSON array containing each social link, its associated text, and what image to render. Simple enough to alter in the future. If it fails though, we should have a state that doesn't break the client.

We should do this, shoudn't be too hard.
```json
[
    {
        "url": "www.youtube.com/user/mta",
        "imgset": "gui/youtube.png"
    }, 
    // ....
]
```

Issue #321, opened 2018-08-10
Vectors are incredibly slow


Pirulax:
**Describe the bug**
So, using Vectors is a really good way to make your script more maintainable, but...
They are incredibly slow.
Here are some results:
```
------------[Vector2]------------
Create: 12348 ms.
sub: 16146 ms.
mul: 12616 ms.
eq: 456 ms.
pow: 12112 ms.
div: 13508 ms.
add: 11024 ms.
------------[Vector3]------------
Create: 13820 ms.
sub: 12819 ms.
mul: 11932 ms.
eq: 635 ms.
pow: 11516 ms.
div: 11353 ms.
add: 12621 ms.
------------[Vector4]------------
Create: 14891 ms.
sub: 11640 ms.
mul: 12306 ms.
eq: 385 ms.
pow: 13031 ms.
div: 11626 ms.
add: 10854 ms.
------------------------
Sub: 95
Function calls: 283 ms.

```
Sub: 95 (Time it takes to do a simple i = i-i expression)
Function calls: 283 ms. (This is the time it takes to call a function.)

**Expected behavior**
I'd expect Vectors to be far more faster.
 ~~The main reason they are slow is they use `lua_classmetamethod`, you may ask why:~~
~~Because I made a little Vector class in Lua and ran a performance test, got nearly the same results~~
~~the difference was 300ms/1m runs(with integers, with floats it was a little bit less)~~
Looks like comparation is pretty fast, maybe modification of the value is the slow part.

**Additional context**
Here's the test resource: 
[vectest.zip](https://github.com/multitheftauto/mtasa-blue/files/2278549/vectest.zip)

(It's redundant a little and such, but really I just made it to ran these tests.)

CrosRoad95:
I agree! vectors are 10 times slower then normal list of coords

Pirulax:
On the other side, internal Lua functions are pretty fast, so, maybe we need a custom version of lua, which has vectors by default, or something, because this way only MTA's source code benefits from Vectors. 
They are literally 100x times slower(not considering the time it takes to create them.)

Pirulax:
Actually, OOP in MTA is 10x slower than PP..
Ridiculous

qaisjp:
There is also a memory "problem" caused by infrequent garbage collection. See [mantis 9840](https://bugs.mtasa.com/view.php?id=9480).

It would be useful to document this online.

Suggestions:

[...]

PR #1554, opened 2020-07-05
Add getElementSpeed function

HexonDev:
Based on the useful function. Most people use this rather than get velocity

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
@@ -112,6 +112,12 @@ ADD_ENUM(HUD_CROSSHAIR, "crosshair")
 ADD_ENUM(HUD_ALL, "all")
 IMPLEMENT_ENUM_END("hud-component")
 
+IMPLEMENT_ENUM_CLASS_BEGIN(eSpeedUnit)
+ADD_ENUM(eSpeedUnit::KMPH, "kmph")
+ADD_ENUM(eSpeedUnit::MPH, "mph")
+ADD_ENUM(eSpeedUnit::MPS, "mps")
+IMPLEMENT_ENUM_CLASS_END("speed-unit")
+
 IMPLEMENT_ENUM_BEGIN(eAmbientSoundType)
 ADD_ENUM(AMBIENT_SOUND_GENERAL, "general")
 ADD_ENUM(AMBIENT_SOUND_GUNFIRE, "gunfire")
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
@@ -116,6 +116,7 @@ IMPLEMENT_ENUM_CLASS_BEGIN(eSpeedUnit)
 ADD_ENUM(eSpeedUnit::KMPH, "kmph")
 ADD_ENUM(eSpeedUnit::MPH, "mph")
 ADD_ENUM(eSpeedUnit::MPS, "mps")
+ADD_ENUM(eSpeedUnit::KT, "kt")
 IMPLEMENT_ENUM_CLASS_END("speed-unit")
 
 IMPLEMENT_ENUM_BEGIN(eAmbientSoundType)
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
@@ -116,7 +116,7 @@ IMPLEMENT_ENUM_CLASS_BEGIN(eSpeedUnit)
 ADD_ENUM(eSpeedUnit::KMPH, "kmph")
 ADD_ENUM(eSpeedUnit::MPH, "mph")
 ADD_ENUM(eSpeedUnit::MPS, "mps")
-ADD_ENUM(eSpeedUnit::KT, "kt")
+ADD_ENUM(eSpeedUnit::KN, "kn")
 IMPLEMENT_ENUM_CLASS_END("speed-unit")
 
 IMPLEMENT_ENUM_BEGIN(eAmbientSoundType)
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
@@ -117,6 +117,7 @@ ADD_ENUM(eSpeedUnit::KMPH, "kmph")
 ADD_ENUM(eSpeedUnit::MPH, "mph")
 ADD_ENUM(eSpeedUnit::MPS, "mps")
 ADD_ENUM(eSpeedUnit::KN, "kn")
+ADD_ENUM(eSpeedUnit::GTASA, "gtasa")
 IMPLEMENT_ENUM_CLASS_END("speed-unit")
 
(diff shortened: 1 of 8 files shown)


LosFaul:
i think it's a bad idea to hardcode something like this
since there are also other speed units
for example there is knots for boats and planes

another point:
i'm also not sure if it's a good practice to hardcode non SI-Units since mph is an imperial speed unit

qaisjp:
Hi, thanks for the pull request.

I'm not convinced this is necessary when [the Lua equivalent](https://wiki.multitheftauto.com/wiki/GetElementSpeed) seems to work perfectly fine.

What's the reasoning for adding this to MTA?

HexonDev:
> What's the reasoning for adding this to MTA?

1. In my opinion, since the MTA does not have a "require” or "import" function, it is simpler than copying and pasting code parts a hundred times where you want to use.

2. Exports are very slow to add an exported function.

Pirulax:
I agree with him.
Exports can be very slow(something which hopefully will be improved with my PR), but to give you exact numbers:
10 000export calls, can take around 80 ms, and now, thats a lot.

And to be honest, who uses `getVelocity`? Everyone needs the speed, why not add it as a default function?

Pirulax:
> i think it's a bad idea to hardcode something like this
> since there are also other speed units
> for example there is knots for boats and planes
> 
> another point:
> i'm also not sure if it's a good practice to hardcode non SI-Units since mph is an imperial speed unit

Thats a good point actually. But since people already use them, why not?

Shuubaru:
> i think it's a bad idea to hardcode something like this
> since there are also other speed units
> for example there is knots for boats and planes
> 
> another point:
> i'm also not sure if it's a good practice to hardcode non SI-Units since mph is an imperial speed unit

since this is mainly used for vehicles, knots should be included too.

HexonDev:
> since this is mainly used for vehicles, knots should be included too.

I added knots (kn) to the speed units

LosFaul:
if this gets accepted or not
but:
please, do not write KMPH but KPH for km/h
for knots use factor: 97.19222454 instead of 180 * 0.539956803

LosFaul:
may make it so, if no string for speed unit is passed
that it returns GTA SA speed unit

qaisjp:
> In my opinion, since the MTA does not have a "require” or "import" function, it is simpler than copying and pasting code parts a hundred times where you want to use.

I'm still not convinced that adding this function to MTA is the solution to that problem, esp. since there are so many different units that users may want to use.

I'm going to close this PR by the end of the week, unless another contributor is convinced this is worth merging?

Pirulax:
I think if they really want to use those special units they want, they can always google up something like "m/s to <my special unit>" and just apply a multiplier to it
The unit thing is there for convenience, and more readable code, but if users want to use some specific units they can always just multiply any of the given units

AlexTMjugador:
In addition to what @qaisjp said, why limit this function to peds, players and vehicles? Objects, projectiles, weapons and even sounds also have velocities (and thus speed), according to this function: [https://github.com/multitheftauto/mtasa-blue/blob/1e2d9ae5683249cac9594e86817897a1546ece69/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L438](https://github.com/multitheftauto/mtasa-blue/blob/1e2d9ae5683249cac9594e86817897a1546ece69/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L438)

Overall, I feel that this function is just a convenience for the what may be the top 10 uses of GetElementVelocity, that by its design will no matter what leave some use cases out, and I don't think that its perceived performance advantage really justifies having one more API to perform the same operation. Right now, to get the speed of an element, we first get the velocity vector, and then compute its length by either manual exponentiation operations or the Vector class (which IMHO is very nice from a readability and mathematical point of view). This PR would add another way, making the Lua API unnecessarily more complex.

If the performance advantage is really that big, then I think that some actual benchmarks comparing real scripts that need to compute a lot of speeds would be interesting to see. I also feel that most script developers just want this function to draw some kind of speedometer on screen, where exporting it for use between resources is not needed (computing an element speed is like two lines of code using the Vector class, and you don't really keep your code much more DRY by exporting it) and maybe they can get away with any performance issue by using a digital speedometer and lowering the velocity sampling frequency. As a wise man once said, _premature optimization is the root of all evil_.

Edit: in case someone misinterpreted the previous quote as against optimizations: I believe there was/is/were/are some PR about optimizing the Vector class performance. In fact, I think that optimizations should aim to make that generic and highly reusable class faster, instead of addressing these individual use cases like this. And AFAIK the Vector class performance can still be improved in some ways, perhaps very noticeably.

Pirulax:
That PR is mine, the vector one.
The best solution to the Vector thing wouldve been to implement it in pure Lua. But afaik it wasn't so any faster really, and reading it would become slower as we'd need to traverse the array.

Pirulax:
Btw, Im not sure if that * 50 for the m/s is accurate
One GTA unit is exactly a meter, but I'm pretty sure velocity uses `precision unit` which is 1 / 16 of a meter, so 6,25 cm. 
[Source](https://gta.fandom.com/wiki/Unit)

But we left it this way, because the useful function had used it like this already.

Pirulax:
Oh and, last time I've tried in-game, Ped and other's returned `0` as their velocity.

LosFaul:
> 
> 
> Btw, Im not sure if that * 50 for the m/s is accurate
> One GTA unit is exactly a meter, but I'm pretty sure velocity uses `precision unit` which is 1 / 16 of a meter, so 6,25 cm.
> [Source](https://gta.fandom.com/wiki/Unit)
> 
> But we left it this way, because the useful function had used it like this already.

since i'm working a lot with setElementVelocity
i can tell you 1/50 is the correct factor for speed units

AlexTMjugador:
> Oh and, last time I've tried in-game, Ped and other's returned `0` as their velocity.

Does it happen with `getElementVelocity` too? If not, I think something inconsistent is going on there.

qaisjp:
@AlexTMjugador comment goes into a lot of detail, and mirrors my thoughts: https://github.com/multitheftauto/mtasa-blue/pull/1554#issuecomment-655078544

> If the performance advantage is really that big, then I think that some actual benchmarks comparing real scripts that need to compute a lot of speeds would be interesting to see.

We've not seen this, so I'm going to go ahead and mark this as decline and close it. I'm, of course, happy to reopen this if we see some useful metrics. Thank you very much for taking the initiative and creating this pull request, @HexonDev :)

Pirulax:
The speed difference is actually pretty big, at least compared to the Lua version of the function on Wiki.
Especially because of #321.
Would you guys agree to add this function, but without hardcoding units?
We could return `m/s`, as thats a SI unit, and can be converted to anything on demand.

qaisjp:
The original version never used OOP https://wiki.multitheftauto.com/index.php?title=GetElementSpeed&direction=prev&oldid=42919

Someone replaced it with OOP in ~~2017~~ 2014: https://wiki.multitheftauto.com/index.php?title=GetElementSpeed&direction=next&oldid=36719

Please feel free to replace the implementation on the wiki with a non-OOP implementation

Pirulax:
Can't argue with that

qaisjp:
In fact, it was @AlexTMjugador that replaced it in *2014

LosFaul:
i personally would just start with returning speed in GTA units
since getElementVelocity also returns speed in GTA units

AlexTMjugador:
> In fact, it was @AlexTMjugador that replaced it in *2014

Good old times when I didn't know what a class was and thought that short-circuiting every check was cool and readable... 😂

Anyway, anybody, feel free to improve that function. I'm sorry if because of it anybody thought that getting an element speed is slower than it needs to be.

PR #1555, opened 2020-07-06
Add short and char support to the new parser

StrixG:
For example, setPedFightingStyle requires `unsigned char` parameter.


Pirulax:
I think we could just change all that mess to `std::is_integral_v<T> && !std::is_same_v<T, bool> || std::is_floating_point_v<T>`
Note: The `&& !std::is_same_v<T, bool>` part is needed, as `bool` is considered an integral type.

sbx320 (approved):

PR #1573, opened 2020-07-22
Support for passing macros to a shader

### LINKED ISSUE #1220: An opportunity to pass preprocessor definitions for shaders
D3DX effects offer us an argument for passing preprocessor definition. It's would be great to add an opportunity to pass these defines to shader simply by calling `dxCreateShader(string filepath / string raw_data [, float priority = 0, float maxDistance = 0, bool layered = false, string elementTypes = "world,vehicle,object,other", string defines = "" ])`. It also would be helpful because in some cases shaders contain reusable parts. We can write just one shader for many cases simply putting correspondig macros among the codelines. It also opens a simple way to activate debug parts within a shader without any shader modifications.

I suggest to add a new argument to function dxCreateShader. In example `dxCreateShader(string filepath / string raw_data [, float priority = 0, float maxDistance = 0, bool layered = false, string elementTypes = "world,vehicle,object,other", string defines = "" ])`.

There is no alternative at the moment

tederis:
An implementation of idea suggested in the issue #1220

**Syntax:**
`element, string dxCreateShader( string filepath / string raw_data [, table macros = {}, float priority = 0, float maxDistance = 0, bool layered = false, string elementTypes = "world,vehicle,object,other" ] )`
where the **macros** is an optional parameter. For sake of backward compatibility it can be omitted.

Sample resource:
[shader_macro.zip](https://github.com/multitheftauto/mtasa-blue/files/4958577/shader_macro.zip)

### DIFF
--- a/Client/core/Graphics/CRenderItem.EffectCloner.cpp
+++ b/Client/core/Graphics/CRenderItem.EffectCloner.cpp
@@ -32,19 +32,22 @@ CEffectCloner::CEffectCloner(CRenderItemManager* pManager)
 //
 ////////////////////////////////////////////////////////////////
 CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus,
-    bool bDebug, const std::vector<std::pair<SString, SString>>& macros)
+    bool bDebug, const EffectMacroList& macros)
 {
-    const std::size_t h0 = CalculateInvariantHash(strFile, macros);
+    SEffectInvariant invariant{
+        ConformPathForSorting(strFile),
+        macros
+    };
 
     // Do we have a match with the initial path
-    CEffectTemplate* pEffectTemplate = MapFindRef(m_ValidMap, h0);
+    CEffectTemplate* pEffectTemplate = MapFindRef(m_ValidMap, invariant);
     if (pEffectTemplate)
     {
         // Have files changed since create?
         if (pEffectTemplate->HaveFilesChanged())
         {
             // EffectTemplate is no good for cloning now, so move it to the old list
-            MapRemove(m_ValidMap, h0);
+            MapRemove(m_ValidMap, invariant);
             m_OldList.push_back(pEffectTemplate);
             pEffectTemplate = NULL;
         }
@@ -78,7 +81,7 @@ CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SStrin
         else
         {
             // Add to active map
-            MapSet(m_ValidMap, h0, pEffectTemplate);
+            m_ValidMap[std::move(invariant)] = pEffectTemplate;
         }
 
         if (!strReport.empty())
@@ -160,7 +163,7 @@ void CEffectCloner::MaybeTidyUp(bool bForceDrasticMeasures)
 #endif
 
     // Valid Effect not used for a little while can go
-    for (std::map<std::size_t, CEffectTemplate*>::iterator iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
+    for (auto iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
     {
         CEffectTemplate* pEffectTemplate = iter->second;
         if (pEffectTemplate->GetTicksSinceLastUsed() > (bForceDrasticMeasures ? 0 : iTicks))
@@ -177,18 +180,4 @@ void CEffectCloner::MaybeTidyUp(bool bForceDrasticMeasures)
     {
         CGraphics::GetSingleton().GetDevice()->EvictManagedResources();
     }
-}
-
-std::size_t CEffectCloner::CalculateInvariantHash(const SString& strFile, const std::vector<std::pair<SString, SString>>& macros)
-{
-    std::size_t h0 = std::hash<SString>{}(ConformPathForSorting(strFile));
-
-    for (const auto& entry : macros)
-    {
-        // Hash function from boost
-        h0 ^= std::hash<SString>{}(entry.first)  + 0x9e3779b9 + (h0 << 6) + (h0 >> 2);
-        h0 ^= std::hash<SString>{}(entry.second) + 0x9e3779b9 + (h0 << 6) + (h0 >> 2);
-    }
-
-    return h0;
 }
--- a/Client/core/Graphics/CRenderItem.EffectCloner.cpp
+++ b/Client/core/Graphics/CRenderItem.EffectCloner.cpp
@@ -31,17 +31,20 @@ CEffectCloner::CEffectCloner(CRenderItemManager* pManager)
 //
 //
 ////////////////////////////////////////////////////////////////
-CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus, bool bDebug)
+CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus,
+    bool bDebug, const std::vector<std::pair<SString, SString>>& macros)
 {
+    const std::size_t h0 = CalculateInvariantHash(strFile, macros);
+
     // Do we have a match with the initial path
-    CEffectTemplate* pEffectTemplate = MapFindRef(m_ValidMap, ConformPathForSorting(strFile));
+    CEffectTemplate* pEffectTemplate = MapFindRef(m_ValidMap, h0);
     if (pEffectTemplate)
     {
         // Have files changed since create?
         if (pEffectTemplate->HaveFilesChanged())
         {
             // EffectTemplate is no good for cloning now, so move it to the old list
-            MapRemove(m_ValidMap, ConformPathForSorting(strFile));
+            MapRemove(m_ValidMap, h0);
             m_OldList.push_back(pEffectTemplate);
             pEffectTemplate = NULL;
         }
@@ -57,7 +60,7 @@ CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SStrin
         HRESULT hr;
         for (uint i = 0; i < 2; i++)
         {
-            pEffectTemplate = NewEffectTemplate(m_pManager, strFile, strRootPath, bIsRawData, strOutStatus, bDebug, hr);
+            pEffectTemplate = NewEffectTemplate(m_pManager, strFile, strRootPath, bIsRawData, strOutStatus, bDebug, macros, hr);
             if (pEffectTemplate || hr != E_OUTOFMEMORY || i > 0)
             {
                 if (pEffectTemplate && i > 0)
@@ -75,7 +78,7 @@ CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SStrin
         else
         {
             // Add to active map
-            MapSet(m_ValidMap, ConformPathForSorting(strFile), pEffectTemplate);
+            MapSet(m_ValidMap, h0, pEffectTemplate);
         }
 
         if (!strReport.empty())
@@ -157,7 +160,7 @@ void CEffectCloner::MaybeTidyUp(bool bForceDrasticMeasures)
 #endif
 
     // Valid Effect not used for a little while can go
-    for (std::map<SString, CEffectTemplate*>::iterator iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
+    for (std::map<std::size_t, CEffectTemplate*>::iterator iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
     {
         CEffectTemplate* pEffectTemplate = iter->second;
         if (pEffectTemplate->GetTicksSinceLastUsed() > (bForceDrasticMeasures ? 0 : iTicks))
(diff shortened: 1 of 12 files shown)


Client/core/Graphics/CRenderItem.EffectCloner.cpp
```diff
@@ -31,17 +31,20 @@ CEffectCloner::CEffectCloner(CRenderItemManager* pManager)
 //
 //
 ////////////////////////////////////////////////////////////////
-CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus, bool bDebug)
+CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus,
+    bool bDebug, const std::vector<std::pair<SString, SString>>& macros)
 {
+    const std::size_t h0 = CalculateInvariantHash(strFile, macros);
+
```
Pirulax: The `h0` variable name is very bad. Could you please change it to something meaningful?

Client/core/Graphics/CRenderItem.EffectCloner.cpp
```diff
@@ -157,7 +160,7 @@ void CEffectCloner::MaybeTidyUp(bool bForceDrasticMeasures)
 #endif
 
     // Valid Effect not used for a little while can go
-    for (std::map<SString, CEffectTemplate*>::iterator iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
+    for (std::map<std::size_t, CEffectTemplate*>::iterator iter = m_ValidMap.begin(); iter != m_ValidMap.end();)
```
Pirulax: `for (auto [unused, pEffectTemplate) : m_ValidMap)` should work just fine. Or if you want to omit the ugly `unused` part, you could use
`for (auto pair : m_ValidMap)`
then just `pair->second`, which feels worse actually, since its not clear at first glance what `second` is

Client/core/Graphics/CRenderItem.EffectCloner.cpp
```diff
@@ -175,3 +178,17 @@ void CEffectCloner::MaybeTidyUp(bool bForceDrasticMeasures)
         CGraphics::GetSingleton().GetDevice()->EvictManagedResources();
     }
 }
+
+std::size_t CEffectCloner::CalculateInvariantHash(const SString& strFile, const std::vector<std::pair<SString, SString>>& macros)
+{
+    std::size_t h0 = std::hash<SString>{}(ConformPathForSorting(strFile));
+
+    for (const auto& entry : macros)
```
Pirulax: `macroEntry` would be a better name maybe?
Sorry for being nitpicky

Client/core/Graphics/CRenderItem.EffectCloner.h
```diff
@@ -19,12 +19,14 @@ class CEffectCloner
...
     void         MaybeTidyUp(bool bForceDrasticMeasures = false);
+    std::size_t  CalculateInvariantHash(const SString& strFile, const std::vector<std::pair<SString, SString>>& macros);
 
-    CElapsedTime                        m_TidyupTimer;
-    CRenderItemManager*                 m_pManager;
-    std::map<SString, CEffectTemplate*> m_ValidMap;            // Active and files not changed since first created
-    std::vector<CEffectTemplate*>       m_OldList;             // Active but files changed since first created
+    CElapsedTime                            m_TidyupTimer;
+    CRenderItemManager*                     m_pManager;
+    std::map<std::size_t, CEffectTemplate*> m_ValidMap;            // Active and files not changed since first created
```
Pirulax: Please add a comment about `size_t` being a hash here, since its not clear at all.
I'd suggest to use `std::unordered_map` as it's faster (it also hash a hash function, which isn't much of a use here, sadly).

Client/core/Graphics/CRenderItem.EffectTemplate.cpp
```diff
@@ -214,6 +216,13 @@ void CEffectTemplate::CreateUnderlyingData(const SString& strFile, const SString
     macroList.back().Name = "IS_DEPTHBUFFER_RAWZ";
     macroList.back().Definition = bUsesRAWZ ? "1" : "0";
 
+    for (const auto& entry : macros)
+    {
+        macroList.push_back(D3DXMACRO());
+        macroList.back().Name = entry.first.c_str();
+        macroList.back().Definition = entry.second.c_str();
+    }
+
```
Pirulax: In theory you can use `emplace_back` here:
```cpp
typedef struct D3DXMACRO {
  LPCSTR Name;
  LPCSTR Definition;
} D3DXMACRO, *LPD3DXMACRO;
```
It would be: `macroList.emplace_back(entry.first.c_str(), entry.second.c_str());`
Or a more readable way:
```cpp
for (const auto& [name, definition] : macros)
   macroList.emplace_back(name.c_str(), definition.c_str());
```
Should work just fine.

Client/core/Graphics/CRenderItem.EffectTemplate.cpp
```diff
@@ -351,6 +360,9 @@ void CEffectTemplate::CreateUnderlyingData(const SString& strFile, const SString
         }
     }
 
+    // Copy macros
+    m_Macros = macros;
```
Pirulax: Not sure if its used anywhere else, but `std::move(macros)`.
Pass `std::vector` by value until here, so you can std::move it(a vector is like 8 bytes @ 32 bit, its worth the effort probably, if you can std::move a few strings in the end)

Client/core/Graphics/CRenderItem.EffectCloner.cpp
```diff
@@ -31,17 +31,20 @@ CEffectCloner::CEffectCloner(CRenderItemManager* pManager)
 //
 //
 ////////////////////////////////////////////////////////////////
-CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus, bool bDebug)
+CEffectWrap* CEffectCloner::CreateD3DEffect(const SString& strFile, const SString& strRootPath, bool bIsRawData, SString& strOutStatus,
+    bool bDebug, const std::vector<std::pair<SString, SString>>& macros)
```
Pirulax: I thought about this `macros` thing, why dont we use a struct like 
```cpp
struct ShaderMacros 
{
   std::string name;
   std::string definition;
};
```
As its used on a lot of places.
Or, just typdef the whole thing to something like 
```cpp
typedef std::vector<std::pair<SString, SString>>& macros) ShaderMacroList_t;
```

Shared/sdk/CScriptArgReader.h
```diff
@@ -1007,6 +1007,25 @@ class CScriptArgReader
...
+        {
+            InternalReadPair(outPairs, m_iIndex);
+            ++m_iIndex;
+            return;
+        }
+
+        SetTypeError("table");
+        ++m_iIndex;
+    }
+
```
Pirulax: Every time I see a vector, I want to use `reserve()`
2 strings are 16 bytes, so if you reserve like 128, then `shrink_to_fit()` you should be good.
Also, you could just use `lua_rawlen`, which may or may not return the correct value. [See here](https://stackoverflow.com/questions/26643285/get-lua-table-size-in-c) for more info

Shared/sdk/CScriptArgReader.h
```diff
@@ -1052,6 +1071,31 @@ class CScriptArgReader
...
 
 protected:
+    void InternalReadPair(std::vector<std::pair<SString, SString>>& outPairs, int iIndex)
+    {
+        lua_pushnil(m_luaVM);
+        while (lua_next(m_luaVM, iIndex) != 0)
+        {
+            int keyType = lua_type(m_luaVM, -2);
+            int valueType = lua_type(m_luaVM, -1);
+            if (keyType == LUA_TSTRING)
```
Pirulax: Don't forget about the fact, that `LUA_TNUMBER` can be converted to a string as well. As this is a common function, I think the conversion to string should be accepted here.

Shared/sdk/CScriptArgReader.h
```diff
@@ -1052,6 +1071,31 @@ class CScriptArgReader
...
+                SString value;
+                if (valueType == LUA_TSTRING || valueType == LUA_TNUMBER)
+                {
+                    uint uiLength = lua_strlen(m_luaVM, -1);
+                    value.assign(lua_tostring(m_luaVM, -1), uiLength);
+                }
+                else if (valueType == LUA_TBOOLEAN)
+                {
+                    value = (lua_toboolean(m_luaVM, -1) ? "1" : "0");
+                }
```
Pirulax: Feel like a switch case would be more readable?

Pirulax (commented):
These are my thoughts, take them with a grain of salt, others may think otherwise.

Client/core/Graphics/CRenderItem.EffectTemplate.cpp
tederis: D3DXMACRO have no constructor for two arguments. emplace_back uses std::allocator_traits so there is no way to pass the arguments this way. Are you sure MTA supporting C++17 currently? Structured binding was introduced since C++17.

Client/core/Graphics/CRenderItem.EffectTemplate.cpp
tederis: Opps. I forgot to delete this.

Client/core/Graphics/CRenderItem.EffectCloner.cpp
tederis: The iterator in this particular case is needed for erasing effect. So I think it's OK.

Client/core/Graphics/CRenderItem.EffectTemplate.cpp
qaisjp: > MTA supporting C++17 currently? Structured binding was introduced since C++17.

Yes!

https://github.com/multitheftauto/mtasa-blue/blob/6d662f3f8f49a07c351be842ae27148e3cdbb407/premake5.lua#L31

But not in `game_sa` and `multiplayer_sa`:

https://github.com/multitheftauto/mtasa-blue/blob/6d662f3f8f49a07c351be842ae27148e3cdbb407/Client/game_sa/premake5.lua#L7-L9 https://github.com/multitheftauto/mtasa-blue/blob/3ab4d05e8d62278d81dcab76bd1dc89cb383e10a/Client/multiplayer_sa/premake5.lua#L7-L13

Client/core/Graphics/CRenderItem.EffectCloner.cpp
ghost: yes, please

Shared/sdk/CScriptArgReader.h
tederis: It's more a matter of taste. I personally think a pure branching in this case is more readable.

tederis:
I have switched to std::unordered_map so any possible hash collisions have been eliminated.

Client/core/Graphics/CRenderItem.EffectCloner.h
```diff
@@ -7,6 +7,37 @@
...
+            // Hash function from boost
+            uiHash ^= std::hash<SString>{}(name)       + 0x9e3779b9 + (uiHash << 6) + (uiHash >> 2);
+            uiHash ^= std::hash<SString>{}(definition) + 0x9e3779b9 + (uiHash << 6) + (uiHash >> 2);
+        }
+
+        return uiHash;
+    }
+};
+
 ////////////////////////////////////////////////////////////////
```
ghost, line 41: The old code was a lot simpler. Explicit template specialization is very helpful when you want different code to execute for different data types for the same struct or function. In this case, `CalculateInvariantHash` was only referenced in CEffectCloner::CreateD3DEffect. I don't see any benefit of using a template for something simple like this.  

In my opinion, it would be nice to bring back `CEffectCloner::CalculateInvariantHash` and  set the type of **m_ValidMap** to 
```
std::unordered_map<std::size_t, CEffectTemplate*> m_ValidMap;
```
The rest of the PR is good.

tederis: It was a lot simpler, I agree. But without explicit element comparison(what unordered_map actually do) there is a risk of hash collision. This is why I added such a specialization. Complicated but reliable.

ghost (approved):

qaisjp:
[non-blocking] I wish there was a nicer solution for ordered macros, but there isn't, is there?

I suppose it could be `{ tuple... }` Rather than `{ {k=v}, ... }`. But I don't know if that's better.

tederis:
> [non-blocking] I wish there was a nicer solution for ordered macros, but there isn't, is there?
> 
> I suppose it could be `{ tuple... }` Rather than `{ {k=v}, ... }`. But I don't know if that's better.

I have added a nice solution for this. It now supports a **table of tables** and **table of key-values**. It can be combined also. Example:
```
macros = {
    -- The first approach
    { "DEF", "VALUE" }, 
    { "DEF2", "DEF VALUE2" },

    -- The second approach
    KEY = "VALUE",
    [ "key" ] = "KEY + VALUE"
}
```
`{ key, value }` is always processing first and only sequentially. `[ key ] = value ` is just for flags.

Dutchman101:
@tederis Please document these changes on scripting wiki

patrikjuvonen:
I think it is a serious oversight to have let macros table go in as the first optional parameter. That is a breaking change...

Pirulax:
Why?

patrikjuvonen:
> Why?

Because argument order has changed. Or does the new argument parser skip too?

tederis:
> > Why?
> 
> Because argument order has changed. Or does the new argument parser skip too?

Like you said, it is an optional argument. So the order doesn't matter.

Pirulax:
It only reads the argument if it's a table. Otherwise it's skipped.

patrikjuvonen:
Ok my bad. I was confused because wiki page had two versions with and without the argument so looked like backwards compatibility issue.

tederis:
I have changed the wiki page to emphasize the optionality of this argument.

Issue #981, opened 2019-06-07
Support for model-integrated UV anim


tut98:
**Is your feature request related to a problem? Please describe.**
Custom models using UV animation does not seem to be featured in MTA. The animation data is stored in the model and does not require IDE or IPL changes.

**Describe the solution you'd like**
I'd like to see the UV anim data be supported for MTA.

**Describe alternatives you've considered**
See [this page](https://wiki.multitheftauto.com/wiki/Shader_examples), it has a resource that lets you achieve the same result as if the UV animation was integrated in the model.

**Additional context**
This has, from my understanding never been possible on MTA. Had a look around threads from 2009-2016.

saml1er:
Does it work in SA if you replace the DFF file in gta3.img?

tutgreco:
Hello @saml1er, thank you very much for the reply.

I have tested it on single player with gta3.img modification and it works fine. On MTA the data is not read. I have recorded a video on SA and MTA respectively:
https://streamable.com/2017f
https://streamable.com/a98fu

PR #1575, opened 2020-07-22
Fix detection of some dff's

### LINKED ISSUE #1444: engineLoadDFF can't load some dff's from buffer.
**Describe the bug**
Some DFF's ( all have 0x2B anim section in beginning ) can't be loaded from buffer. 
From file these dff's load well.

**To reproduce**
[testDff.zip](https://github.com/multitheftauto/mtasa-blue/files/4639181/testDff.zip)
Use command /testdff from this test resouce

**Expected behaviour**
I can load this dff from buffer.

**Version**
Multi Theft Auto v1.5.7-release-20537

TheNormalnij:
Fix issue #1444 
Some DFF's have UV Animation Dictionary RW Section (0x2B). IsDFFData check only 0x10 section and you can't load DFF's with 0x2B section from buffer using engineLoadDFF.


ghost (approved):
Thanks for the PR. Does it also fix #981?

TheNormalnij:
No. It only helps to load a model from a string.

PR #1577, opened 2020-07-24
Further xmlLoadString memleak fixes

qaisjp:
I recommend that the reviewer reads commit by commit.

~~Assistance needed - see fffadce "test destructors" - please can someone explain why it only says `delete node;` and not `delete doc;` too?~~

~~When a resource is stopped, `CLuaMain` is deleted, and that automatically deletes this unordered set:~~

```cpp
std::unordered_set<std::unique_ptr<SXMLString>> m_XMLStringNodes;
```

~~And we have destructors defined `SXMLString` and `SXMLStringImpl`.~~

Only tested serverside on windows.

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaMain.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaMain.cpp
@@ -57,11 +57,6 @@ CLuaMain::~CLuaMain()
 
     // Delete the timer manager
     delete m_pLuaTimerManager;
-    
-    for (auto& xmlNode : m_XMLNodes)
-    {
-        delete xmlNode;
-    }
 
     CClientPerfStatLuaMemory::GetSingleton()->OnLuaMainDestroy(this);
     CClientPerfStatLuaTiming::GetSingleton()->OnLuaMainDestroy(this);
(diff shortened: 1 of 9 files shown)


qaisjp (commented):
What happens if resource `x` uses `xmlLoadString`, passes it to resource `b` (via export), and then resource `x` is stopped?

ghost:
> 
> 
> What happens if resource `x` uses `xmlLoadString`, passes it to resource `b` (via export), and then resource `x` is stopped?

This will probably crash the program. When I get some time, I'll try to study how the xml implementation works, and I'll write a fix for it.

qaisjp:
Just wanted to drop that down last night as something that needs to be tested. Now that I've tested, I can confirm it works fine!

```
run f = exports.xml:getFileNode()
run s = exports.xml:getStringNode()
run f.children[1].name -- outputs something
run s.children[1].name -- outputs something
stop xml
run f.children -- bad argument, expected xml-node
run s.children -- bad argument, expected xml-node
```

Presumably because our Lua/XML implementation works entirely off CXMLNode, which is shared by both file-based nodes and string-based nodes. Not investigated too much into _why_, but it makes sense.

ghost (approved):

PR #1582, opened 2020-07-29
Audit all OOP definitions

### LINKED ISSUE #1240: Behaviour of some OOP methods depend on min client version setting
The following OOP methods change their behaviour depending on the min client version setting. The min client requirement can change for various reasons (e.g. compiling) and therefore cause the script to break.
```
getElementBoundingBox
getVehicleComponentPosition
getVehicleComponentRotatio
```

qaisjp:
This pull request fixes #1240 by removing three instances of version-dependent API functionality. It also fixes a mistake in getColShapeSize where a vector is always returned -- multiple floats should have been returned for procedural calls.

This pull request also led to the creation of #1583. 

Note to reviewer: commits are atomic, so you can review commit-by-commit.


StrixG (approved):

qaisjp:
ty for review!

PR #1597, opened 2020-08-05
Fix #1594: getPedMoveState returns incorrect move state

### LINKED ISSUE #1594: getPedMoveState returns incorrect move state while entering a vehicle
**Describe the bug**
[getPedMoveState](https://wiki.multitheftauto.com/wiki/GetPedMoveState) returns incorrect state while entering a vehicle
like powerwalk, fall

**To reproduce**
1. enter a vehicle 
2. while entering a vehicle 
do crun getPedMoveState(localPlayer)
3. you will probably see "powerwalk" or "fall"

or you can use a script I made it to detect any [getPedMoveState](https://wiki.multitheftauto.com/wiki/GetPedMoveState) changes and output it to the chatbox
[moveState.zip](https://github.com/multitheftauto/mtasa-blue/files/5018167/moveState.zip)

**Expected behaviour**
it should return false or specific state

**Screenshots**

mta-screen_2020-08-03_22-34-40
**Version**
v1.5.7-release-20582

Unde-R:
Fixed getPedMoveState while entering a vehicle and added when leaving a vehicle.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -193,6 +193,8 @@ void CClientPed::Init(CClientManager* pManager, unsigned long ulModelID, bool bI
     m_MovementStateNames[MOVEMENTSTATE_STAND] = "stand";
     m_MovementStateNames[MOVEMENTSTATE_WALK] = "walk";
     m_MovementStateNames[MOVEMENTSTATE_POWERWALK] = "powerwalk";
+    m_MovementStateNames[MOVEMENTSTATE_ENTERINGVEHICLE] = "enteringvehicle";
+    m_MovementStateNames[MOVEMENTSTATE_LEAVINGVEHICLE] = "leavingvehicle";
     m_MovementStateNames[MOVEMENTSTATE_JOG] = "jog";
     m_MovementStateNames[MOVEMENTSTATE_SPRINT] = "sprint";
     m_MovementStateNames[MOVEMENTSTATE_CROUCH] = "crouch";
@@ -2378,7 +2380,7 @@ eMovementState CClientPed::GetMovementState()
             return MOVEMENTSTATE_JUMP;
         // Is he entering a vehicle?
         else if (IsEnteringVehicle())
-            return false;
+            return MOVEMENTSTATE_ENTERINGVEHICLE;
         // Is he falling?
         else if (!IsOnGround() && !GetContactEntity())
             return MOVEMENTSTATE_FALL;
@@ -2424,7 +2426,7 @@ eMovementState CClientPed::GetMovementState()
     {
         // Is he leaving a vehicle?
         if (IsLeavingVehicle())
-            return false;
+            return MOVEMENTSTATE_LEAVINGVEHICLE;
     }
     return MOVEMENTSTATE_UNKNOWN;
 }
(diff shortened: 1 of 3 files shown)


qaisjp (dismissed):
Incorrectly formatted: mixed tabs/spaces.

Use `Advanced -> Format Selection` to fix the formatting.

Unde-R:
> Incorrectly formatted: mixed tabs/spaces.
> 
> Use `Advanced -> Format Selection` to fix the formatting.

Like this?

Unde-R:
Sorry, I didn't check before commiting.

Einheit-101:
Does this only fix the return value when entering/leaving a vehicle or does this also fix other issues with getPedMoveState() ?

Unde-R:
> Does this only fix the return value when entering/leaving a vehicle or does this also fix other issues with getPedMoveState() ?

only when leaving and entering.
I still have to change it to return false, instead of "leavingvehicle" "enteringvehicle".

Lpsd:
It seems we have resolved all issues relating to move state by another method (awaiting PR).

Regardless, I think this could be a useful feature by itself - there may be cases where you need to know if a ped is entering/leaving a vehicle without using events like `onClientVehicleStartEnter` and `onClientVehicleExit` - this would actually reduce the amount of code needed in some cases (since you can check "now" rather than relying on events).

All looks good to me, I'm happy to merge unless anyone has issues with this before I'm home (~3 hours) 😄

Unde-R:
> It seems we have resolved all issues relating to move state by another method (awaiting PR).
> 
> Regardless, I think this could be a useful feature by itself - there may be cases where you need to know if a ped is entering/leaving a vehicle without using events like `onClientVehicleStartEnter` and `onClientVehicleExit` - this would actually reduce the amount of code needed in some cases (since you can check "now" rather than relying on events).
> 
> All looks good to me, I'm happy to merge unless anyone has issues with this before I'm home (~3 hours) 😄

According to qaisjp he said that we have:

> We currently have "stand", "walk", "jog", "sprint", etc. So "enteringvehicle" is not using the correct tense.
>
So It'll be better to return false instead of "leavingvehicle" "enteringvehicle"

qaisjp:
> We currently have "stand", "walk", "jog", "sprint", etc. So "enteringvehicle" is not using the correct tense.

To clarify I meant that, _if we kept this addition_, it should be `enter_vehicle` (or `entervehicle`) rather than `enteringvehicle`.

But also, as @Unde-R said, in https://github.com/multitheftauto/mtasa-blue/issues/1594#issuecomment-669286857 I suggested to not have the enter/leaving stuff:

> Basically, I'm not 100% sure whether `enter_vehicle` and `leave_vehicle` fits well with the rest of these move states. What do you think?
> 
> What does this function return when we're flying in a jetpack or parachuting? (I'm guessing it returns `fall`?)

I did not mention why I had some uncertainty though.

(I commented on the wrong issue, I thought I was commenting on this PR.)

> there may be cases where you need to know if a ped is entering/leaving a vehicle without using events like `onClientVehicleStartEnter` and `onClientVehicleExit`

The reason I had uncertainty is because I think the same thing can be achieved using [`getPedTask`](https://wiki.multitheftauto.com/wiki/GetPedTask) (+ [List of player tasks](https://wiki.multitheftauto.com/wiki/List_of_player_tasks)).

And basically it's sort of slippery slope as we don't include things like "jetpack_fly" and similar stuff. So vehicle entry/leaving (and jetpack flying) feels sort of "offtopic" for `getPedMoveState`.

Does my thought process make sense? (Of course, feel free to push back if you have other preferences.)

Lpsd:
~~Yeah, on second thoughts, it's probably best to just expose the `IsEnteringVehicle` and `IsLeavingVehicle` methods to the Lua API, rather than adding this to move state.~~

Forgot about `getPedTask` - that already provides the functionality.

Unde-R:
Yeah sure, but at least instead of "Fall" It'll return false

Lpsd:
Agreed. I think returning false here is the best solution.

Unde-R:
#1603

Unde-R:
I think we can close this Pr and move to the new one, since It has everything.

Lpsd:
You should update this PR to do the "return false" thing.

Unde-R:
Oh sure.

Lpsd:
Sorry, I actually didn't see your other PR returns `MOVEMENTSTATE_UNKNOWN` - I'll close this 😆

Issue #1322, opened 2020-03-25
debugscript returns "debugscript: Your debug mode was set to 0" if using a character value.


Haxardous:
debugscript will set it's status to "0" whenever typing "debugscript (any *word* character)".

how to reproduce the bug:
1. type /debugscript 1 or 2 or 3 for e.g. (debugscript 3)
2. type /debugscript word for e.g. (debugscript a)
and it'll return "debugscript: Your debug mode was set to 0"

PlatinMTA:
I guess this happens bcuz atoi will return 0 in failure to read a number, too lazy to check

Btw I can reproduce the bug

qaisjp:
maybe this bug is actually a feature

Pirulax:
Im pretty sure its intended, because when you fuck up, you can just type in "debugscript ", then smash ur keyboard, and press enter in the smashing process, and then alt+f4

Unde-R:
Dont mind this, I did something wrong.

qaisjp:
Fixed in 8f25c22 via #1602

PR #1602, opened 2020-08-06
Fix debugscript setting to 0 when using invalid characters

Unde-R:
Fix for issue #1322

### DIFF
--- a/Server/mods/deathmatch/logic/CConsoleCommands.cpp
+++ b/Server/mods/deathmatch/logic/CConsoleCommands.cpp
@@ -1232,34 +1232,42 @@ bool CConsoleCommands::DebugScript(CConsole* pConsole, const char* szArguments, 
         {
             CPlayer* pPlayer = static_cast<CPlayer*>(pClient);
 
-            // Convert to number
-            int iLevel = atoi(szArguments);
-            if (iLevel != (int)pPlayer->GetScriptDebugLevel())
+            // To prevent characters before converting
+            if (strcmp(szArguments, "0") == 0 || strcmp(szArguments, "1") == 0 || strcmp(szArguments, "2") == 0 || strcmp(szArguments, "3") == 0)
             {
-                // Between 0 and 3?
-                if (iLevel >= 0 && iLevel <= 3)
+                // Convert to number
+                int iLevel = atoi(szArguments);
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CConsoleCommands.cpp
```diff
@@ -1232,34 +1232,42 @@ bool CConsoleCommands::DebugScript(CConsole* pConsole, const char* szArguments,
...
-            // Convert to number
-            int iLevel = atoi(szArguments);
-            if (iLevel != (int)pPlayer->GetScriptDebugLevel())
+            // To prevent characters before converting
+            if (strcmp(szArguments, "0") == 0 || strcmp(szArguments, "1") == 0 || strcmp(szArguments, "2") == 0 || strcmp(szArguments, "3") == 0)
             {
-                // Between 0 and 3?
-                if (iLevel >= 0 && iLevel <= 3)
+                // Convert to number
+                int iLevel = atoi(szArguments);
```
qaisjp: What if you put this here instead?

```suggestion
                int iLevel = atoi(szArguments);
                if (iLevel == 0 && strcmp(szArguments, "0") != 0) {
                    pEchoClient->SendEcho("debugscript: Syntax is 'debugscript <mode>'");
                    return false;
                }
```

We generally try to avoid adding loads of nested if statements if we can, but some of our old (like this!) doesn't do that.

Unde-R:
Oh that seems better, It should work

Server/mods/deathmatch/logic/CConsoleCommands.cpp
Unde-R: Done, reverted the old commit and added this instead.

qaisjp (approved):
Tested, works fine! Thank you for this PR :)

Pirulax:
Why not use std::stoi?
Also, `strcmp` is unnecessary here, because its a single character. Id recommend `*szArguments == '0'` instead (or, the more familiar `szArguments[0] == '0'`, the `'` (slash or whatever) is important, as you want the ASCII keycode, not the number 0 (which represents the null terminator))

Pirulax:
Nvm, it already got merged

qaisjp:
Good suggestions, feel free to send a follow up PR cleaning this up, @Pirulax! Cheers

Issue #1598, opened 2020-08-05
getPedMoveState can returns "fall" while crawling


xLive:
**Describe the bug**
[getPedMoveState](https://wiki.multitheftauto.com/wiki/GetPedMoveState) can returns "fall" while crawling

**To reproduce**
1- crawl
2- do `crun getPedMoveState(localPlayer)`
3- sometimes the function will return "fall"

or use the script I made it to detect the movement change every frame and output it on the chatbox
[moveState.zip](https://github.com/multitheftauto/mtasa-blue/files/5031705/moveState.zip)

**Expected behaviour**
should return only "crawl"

**Screenshots**
![mta-screen_2020-08-06_02-17-11](https://user-images.githubusercontent.com/30420446/89474298-e6c12680-d78d-11ea-8829-616265ed48e7.png)

**Version**
v1.5.7-release-20582

Unde-R:
while crawling and moving or what

xLive:
> while crawling and moving or what

crouch and move

xLive:
Okay, after some tests
I think is not only for crawling it happens with jog,sprint even stand

so I think the surface is the reason

for example this screenshot
![mta-screen_2020-08-06_03-03-35](https://user-images.githubusercontent.com/30420446/89475846-c5fad000-d791-11ea-8162-88d07189b985.png)
I'm just standing at this position and the movement state changes to "fall" every frame

so I think it's the same issue with #1590?

Unde-R:
yeah seems not very accurate

Einheit-101:
Exactly related with #1590

Xenius97:
https://streamable.com/sqjrit

Issue #1590, opened 2020-08-02
getPedMoveState can return "fall" when standing on certain surfaces


Einheit-101:
**Describe the bug**
Sometimes i get "fall" even when the player is standing solid and fine.

**To reproduce**
1. Go to the editor
2. Move on the roof of this house 
    ![image](https://user-images.githubusercontent.com/923242/89134925-0acc0000-d521-11ea-92c2-82bc57b09aa0.png)

3. crun getPedMoveState(localPlayer)
4. Profit

**Expected behaviour**
The function should return "stand"

**Additional information**
It also returns "fall" even when sprinting or doing whatever else. And the bug does not appear on the whole roof surface. 

**Multi Theft Auto v1.5.7-release-20597**

Lpsd:
Could be related to the behaviour of `CClientPed::IsOnGround()` (#534)

See https://github.com/multitheftauto/mtasa-blue/blob/7399bf620b7d5846a87642a4f154d8414a58f4a5/Client/mods/deathmatch/logic/CClientPed.cpp#L2381-L2382 and https://github.com/multitheftauto/mtasa-blue/blob/7399bf620b7d5846a87642a4f154d8414a58f4a5/Client/mods/deathmatch/logic/CClientPed.cpp#L4624-L4630

I'll investigate later.

Unde-R:
> Could be related to the behaviour of `CClientPed::IsOnGround()` (#534)
> 
> See
> 
> https://github.com/multitheftauto/mtasa-blue/blob/7399bf620b7d5846a87642a4f154d8414a58f4a5/Client/mods/deathmatch/logic/CClientPed.cpp#L2381-L2382
> 
> and
> https://github.com/multitheftauto/mtasa-blue/blob/7399bf620b7d5846a87642a4f154d8414a58f4a5/Client/mods/deathmatch/logic/CClientPed.cpp#L4624-L4630
> 
> I'll investigate later.

Yeah, i've just checked it.

Issue #534, opened 2018-09-21
isPedOnGround issues


ArranTuna:
**Describe the bug**
Current isPedOnGround function is very unstable and sometimes doesn't even show the ped is on ground while he is.

**To reproduce**
```lua
function checkDistanceBetweenPlayerAndGround()
local x, y, z = getElementPosition(localPlayer)
local z2 = getGroundPosition(x, y, z)
local dist = getDistanceBetweenPoints3D(x, y, z, x, y, z2)
local ground = tostring(isPedOnGround(localPlayer))
outputChatBox("Distance: "..dist.." IsOnGround: "..ground)
end

crun setTimer(checkDistanceBetweenPlayerAndGround, 1000, 60)

-- Some where I found that returns false when it shouldn't:
crun setElementPosition(localPlayer, -230.175, -1107.408, 16.42)
-- If you don't get it there just move around in that area and eventually you'll see it return false.
```

**Expected behaviour**
Return true when on ground.

**Screenshots**
Here's the video illustrating the problem: http://youtu.be/-qMeC5M4pks

**MTA Client:** 
Unstated.

**MTA Server:** 
Unstated.

**Additional context**
I'm stood in the road and `crun isPedOnGround(localPlayer)` returns false.
If you don't get it first time, keep moving and it will eventually return false when it shouldn't.

Edit: It seems that isPedOnGround fails if your feet aren't actually touching the floor because you're stood on a slope and your feet are slightly off the ground but you're still on the ground though so this function shouldn't be returning false.

From https://bugs.mtasa.com/view.php?id=8378

PlatinMTA:
isPedOnGround() seems to fail with custom objects or custom mapping on interiors aswell.

PR #1603, opened 2020-08-07
Fix getPedMoveState returns incorrect move states

Unde-R:
This should fix the following issues:
#1598 
#1594 
#1590

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -4646,6 +4646,9 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
+    // We don't care if the ped is underground
+    if (definitelyLessThan(vecPosition.fZ, fGroundLevel))
+        return false;
     // Check (vecPosition.fZ - fGroundLevel) <= 1.0f
     return definitelyLessThan((vecPosition.fZ - fGroundLevel), 1.0f) || essentiallyEqual((vecPosition.fZ - fGroundLevel), 1.0f);
 }
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -2377,6 +2377,10 @@ eMovementState CClientPed::GetMovementState()
         else if (strcmp(szComplexTaskName, "TASK_COMPLEX_JUMP") == 0)
             return MOVEMENTSTATE_JUMP;
 
+        // Is he entering a vehicle?
+        else if (IsEnteringVehicle())
+            return MOVEMENTSTATE_UNKNOWN;
+
         // Is he falling?
         else if (!IsOnGround() && !GetContactEntity())
             return MOVEMENTSTATE_FALL;
@@ -2416,6 +2420,15 @@ eMovementState CClientPed::GetMovementState()
                 return MOVEMENTSTATE_CRAWL;
         }
     }
+
+    // Do we have a player, and are we in a car?
+    if (m_pPlayerPed && GetRealOccupiedVehicle())
+    {
+        // Is he leaving a vehicle?
+        if (IsLeavingVehicle())
+            return MOVEMENTSTATE_UNKNOWN;
+    }
+
     return MOVEMENTSTATE_UNKNOWN;
 }
 
@@ -4626,7 +4639,7 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
-    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel) <= 1.0f);
+    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel) <= 1.01f);
 }
 
 bool CClientPed::IsClimbing()
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -10,6 +10,7 @@
 
 #include "StdInc.h"
 #include "game/CAnimBlendAssocGroup.h"
+#include "Utils.h"
 
 using std::list;
 using std::vector;
@@ -4645,8 +4646,8 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
-    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel - 1.0f) <=
-                                                 ((fabs(vecPosition.fZ) < fabs(fGroundLevel) ? fabs(fGroundLevel) : fabs(vecPosition.fZ)) * FLOAT_EPSILON));
+    // Check (vecPosition.fZ - fGroundLevel) <= 1.0f
+    return definitelyLessThan((vecPosition.fZ - fGroundLevel), 1.0f) || essentiallyEqual((vecPosition.fZ - fGroundLevel), 1.0f);
 }
 
 bool CClientPed::IsClimbing()
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -4645,7 +4645,13 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
-    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel) <= 1.01f);
+
+    // We don't care if the ped is underground
+    if ((vecPosition.fZ - fGroundLevel) <= FLOAT_EPSILON)
+        return false;
+        
+    // Check that (zPos - groundLevel) <= 1, but also account for epsilon
+    return (vecPosition.fZ - fGroundLevel - 1.0f) <= FLOAT_EPSILON;
 }
 
 bool CClientPed::IsClimbing()
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -4645,13 +4645,8 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
-
-    // We don't care if the ped is underground
-    if ((vecPosition.fZ - fGroundLevel) <= FLOAT_EPSILON)
-        return false;
-        
-    // Check that (zPos - groundLevel) <= 1, but also account for epsilon
-    return (vecPosition.fZ - fGroundLevel - 1.0f) <= FLOAT_EPSILON;
+    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel - 1.0f) <=
+                                                 ((fabs(vecPosition.fZ) < fabs(fGroundLevel) ? fabs(fGroundLevel) : fabs(vecPosition.fZ)) * FLOAT_EPSILON));
 }
 
 bool CClientPed::IsClimbing()
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -2416,15 +2416,9 @@ eMovementState CClientPed::GetMovementState()
         {
             // Is he moving the contoller at all?
             if (cs.LeftStickX == 0 && cs.LeftStickY == 0)
-            {
                 return MOVEMENTSTATE_CROUCH;
-            }
             else
-            {
-                if ((cs.LeftStickX == -128 || cs.LeftStickX == 128) && cs.RightShoulder1)
-                    return MOVEMENTSTATE_ROLL;
                 return MOVEMENTSTATE_CRAWL;
-            }
         }
     }
 
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -2415,9 +2415,15 @@ eMovementState CClientPed::GetMovementState()
         {
             // Is he moving the contoller at all?
             if (cs.LeftStickX == 0 && cs.LeftStickY == 0)
+            {
                 return MOVEMENTSTATE_CROUCH;
+            }
             else
+            {
+                if ((cs.LeftStickX == -128 || cs.LeftStickX == 128) && cs.RightShoulder1)
+                    return MOVEMENTSTATE_ROLL;
                 return MOVEMENTSTATE_CRAWL;
+            }
         }
     }
 
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -2378,7 +2378,6 @@ eMovementState CClientPed::GetMovementState()
         else if (strcmp(szComplexTaskName, "TASK_COMPLEX_JUMP") == 0)
             return MOVEMENTSTATE_JUMP;
 
-        // Is he entering a vehicle?
         else if (IsEnteringVehicle())
             return MOVEMENTSTATE_UNKNOWN;
 
@@ -2422,10 +2421,8 @@ eMovementState CClientPed::GetMovementState()
         }
     }
 
-    // Do we have a player, and are we in a car?
     if (m_pPlayerPed && GetRealOccupiedVehicle())
     {
-        // Is he leaving a vehicle?
         if (IsLeavingVehicle())
             return MOVEMENTSTATE_UNKNOWN;
     }
--- a/Shared/mods/deathmatch/logic/Utils.cpp
+++ b/Shared/mods/deathmatch/logic/Utils.cpp
@@ -1134,3 +1134,23 @@ CVector ConvertEulerRotationOrder(const CVector& a_vRotation, eEulerRotationOrde
         return a_vRotation;
     }
 }
+
+bool approximatelyEqual(float a, float b, float epsilon)
+{
+    return fabs(a - b) <= ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * FLOAT_EPSILON);
+}
+
+bool essentiallyEqual(float a, float b, float epsilon)
+{
+    return fabs(a - b) <= ((fabs(a) > fabs(b) ? fabs(b) : fabs(a)) * FLOAT_EPSILON);
+}
+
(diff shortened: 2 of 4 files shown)


Lpsd:
Thanks for your PR. I'm sure this will be fine but I'll need to do some extensive testing later just to make sure it doesn't unintentionally break anything else. Good work though!

Unde-R:
> Thanks for your PR. I'm sure this will be fine but I'll need to do some extensive testing later just to make sure it doesn't unintentionally break anything else. Good work though!

Thanks mate

Unde-R:
I tested a code to get groundposition and compared it with player's Z position in the areas which return "fall" and here is the results ( Z - groundposition ) :
1,00000095367
1.00000095367
1.00000095368

qaisjp:
thanks for submitting this. quick question for future documentation / reproduction purposes, please can you provide the coordinates you tested with? thank you :)

Unde-R:
> thanks for submitting this. quick question for future documentation / reproduction purposes, please can you provide the coordinates you tested with? thank you :)

sure one moment

Unde-R:
```
Ground Position: 14.250931739807
Player Position: 2221.7590332031, -1653.4652099609, 15.250932693481
-------------
Ground Position: 14.252799987793
Player Position: 2221.814453125, -1653.5106201172, 15.252800941467
-------------
Ground Position: 14.287864685059
Player Position: 2223.142578125, -1653.3934326172, 15.287865638733
```

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -4626,7 +4639,7 @@ bool CClientPed::IsOnGround()
     CVector vecPosition;
     GetPosition(vecPosition);
     float fGroundLevel = static_cast<float>(g_pGame->GetWorld()->FindGroundZFor3DPosition(&vecPosition));
-    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel) <= 1.0f);
+    return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel) <= 1.01f);
```
qaisjp:
```suggestion

    // We don't care if the ped is underground
    if ((vecPosition.fZ - fGroundLevel) <= FLOAT_EPSILON)
        return false;
        
    // Check that (zPos - groundLevel) <= 1, but also account for epsilon
    return (vecPosition.fZ - fGroundLevel - 1.0f) <= FLOAT_EPSILON;
```

Does this do the same thing?

Unde-R: let me test

Unde-R: since FLOAT_EPSILON expands to 0.0001f it should work, gonna test it rn.

qaisjp: My suggested code seems theoretically incorrect (even if it might happen to work).

Looking at https://stackoverflow.com/a/77735/1517394:

> It all depends on context.
>
> I have spent a long time tracing a bugs in a system that presumed `a==b` if `|a-b|<epsilon`. The underlying problems were: [...]

And https://stackoverflow.com/a/253874/1517394:

> The following definitions are from [The art of computer programming by Knuth](http://books.google.ca/books?ei=KRkLSYLZEJ7ktQOX3-ChBQ&id=T89QAAAAMAAJ&dq=%22essentially+equal%22+%22approximately+equal%22+knuth&q=%22essentially+equal%22+%22approximately+equal%22+%22definitely+less+than%22&pgis=1#search):
> 
> ```cpp
> bool approximatelyEqual(float a, float b, float epsilon)
> {
>     return fabs(a - b) <= ( (fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
> }
> 
> bool essentiallyEqual(float a, float b, float epsilon)
> {
>     return fabs(a - b) <= ( (fabs(a) > fabs(b) ? fabs(b) : fabs(a)) * epsilon);
> }
> 
> bool definitelyGreaterThan(float a, float b, float epsilon)
> {
>     return (a - b) > ( (fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
> }
> 
> bool definitelyLessThan(float a, float b, float epsilon)
> {
>     return (b - a) > ( (fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
> }
> ```

Unde-R: Yeah probably, well we can lower the "1.01f" more according to the values that I got above, "1.01f" is already alot

Einheit-101: Why not use 1.001 instead of 1.01? 0.001 is very very low and no one will notice an issue with this, pretty sure

Unde-R: yeah I thought about it, just waiting for lpsd to test it

Unde-R:
I added `MOVEMENTSTATE_ROLL` because It was declared but no code for It.

- [x] crouch-roll
- [x] crawl-roll

Result: [`https://pastebin.com/vz4zQb3D`](https://pastebin.com/vz4zQb3D)

Einheit-101:
This is absolutely amazing!

Unde-R:
Looks like this issue should be fixed too #534

qaisjp (changes requested):
I'd like to see more correct handling, via `FLOAT_EPSILON`. (Assuming that this is actually the correct way to handle it.)

Unde-R:
Thanks for the help!
Tested It's working fine now.

qaisjp:
The suggested code I provided is not correct. Please read https://github.com/multitheftauto/mtasa-blue/pull/1603#discussion_r467206465

Unde-R:
> The suggested code I provided is not correct. Please read [#1603 (comment)](https://github.com/multitheftauto/mtasa-blue/pull/1603#discussion_r467206465)

`return (vecPosition.fZ > fGroundLevel && (vecPosition.fZ - fGroundLevel - 1.0f) <=
                                                 ((fabs(vecPosition.fZ) < fabs(fGroundLevel) ? fabs(fGroundLevel) : fabs(vecPosition.fZ)) * FLOAT_EPSILON));` Thoughts on this?

qaisjp:
Almost there! Remember, there's two checks here. One that checks `vecPosition.fZ > fGroundLevel`, and another that checks `(vecPosition.fZ - fGroundLevel) <= 1.0f`.

I think this should be split up into two if-s, with correct float comparisons.

To be honest, the code is a bit hard to read, so perhaps you should add helpers like [these](https://swift.im/git/swift-contrib/diff/Swiften/Base?id=c8ced8597ab28b928862d21554c0a68c238b1165&context=4)* to [`Utils.cpp`](https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/mods/deathmatch/logic/Utils.cpp) or [`Utils.h`](https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/mods/deathmatch/logic/Utils.h).

Then you can do something like this:

```cpp
    // We don't care if the ped is underground
    if (definitelyGreaterThan(vecPosition.fZ, fGroundLevel))
        return false;
        
    // Check (vecPosition.fZ - fGroundLevel) <= 1.0f
    return definitelyLessThan(vecPosition.fZ - fGroundLevel, 1.0f) || essentiallyEqual(vecPosition.fZ - fGroundLevel, 1.0f);
```

I'm not sure if this best practice, but it's the best thing I can come up, and I feel like it's better than just adding 0.001 and calling it a day.


[*] this link is from "Swift", but it is actually some chat software, and not the Swift Programming Language

Unde-R:
> ```c++
>     // We don't care if the ped is underground
>     if (definitelyGreaterThan(vecPosition.fZ, fGroundLevel))
>         return false;
>         
>     // Check (vecPosition.fZ - fGroundLevel) <= 1.0f
>     return definitelyLessThan(vecPosition.fZ - fGroundLevel, 1.0f) || essentiallyEqual(vecPosition.fZ - fGroundLevel, 1.0f);
> ```
> 
I tried this ^^ but somehow It's not working properly.

Unde-R:
> 
```c++
// We don't care if the ped is underground
    if (definitelyGreaterThan(vecPosition.fZ, fGroundLevel))
        return false;
```
>
Looks like It's because of this.

Pirulax:
Do this (like in the CVector class): 
```cpp
abs(vec.z) - abs(groundposz) >= FLOAT_EPSILON
```
This is totally acceptable, as `float epsilon` is a thing in CS

qaisjp:
```cpp
    // We don't care if the ped is underground
    if (definitelyGreaterThan(vecPosition.fZ, fGroundLevel))
        return false;
```

The code I suggested had a small mistake.

It doesn't work because we're returning `false` if our z-position is greater than the ground level. That does not match up to the comment that says "we don't care if the ped is underground"?

It should return false if the ped is underground. But we're not checking if the ped is underground correctly.

you should use `definitelyLessThan` instead :)

Unde-R:
Yeah I'll update it soon.
What about crouch-roll should I keep it here or make another PR for it?

qaisjp:
> What about crouch-roll should I make another PR for it?

Good idea

Unde-R:
> Good idea

I'll open a PR rn

ghost (dismissed):

Einheit-101:
This is tested a lot and seems to be fine, why is this not merged yet?

patrikjuvonen:
> This is tested a lot and seems to be fine, why is this not merged yet?

No accepted reviews, and no testing beyond the stale ones has been made. I am interested to hear if this actually resolves the three linked issues and how does it do that since it is not apparent from the description. This leaves me to have to thoroughly delve into the code to figure out whats going on, how to test it and how to try to break it. None of which I have time to do right now personally.

Einheit-101:
I could test functionality with a working build if i can get one. My resource uses these functions.

ghost:
@patrikjuvonen 

Tested on nightly (Multi Theft Auto v1.5.9-release-21251 - but the version shouldn't matter, since it exists on stable as well)

#1598 - [fixed build](https://www.youtube.com/watch?v=Pun1AZwban0) & [nightly build](https://www.youtube.com/watch?v=vAMUaDXrHvY)
#1594 - [fixed build](https://www.youtube.com/watch?v=oqmr-yYHdnk) & [nightly build](https://www.youtube.com/watch?v=ueNFJcby9xM)
#1590 - [fixed build](https://www.youtube.com/watch?v=jDmoqu4tnaM) & [nightly build](https://www.youtube.com/watch?v=7x_N6ldlvvw)

Besides those locations, i've tested whether it happens in the other [places](https://imgur.com/a/k9y8UEc), which haven't been mentioned. And it seems that fix works well.

patrikjuvonen:
> @patrikjuvonen
> 
> Tested on nightly (Multi Theft Auto v1.5.9-release-21251 - but the version shouldn't matter, since it exists on stable as well)
> 
> #1598 - [fixed build](https://www.youtube.com/watch?v=Pun1AZwban0) & [nightly build](https://www.youtube.com/watch?v=ueNFJcby9xM) #1594 - [fixed build](https://www.youtube.com/watch?v=oqmr-yYHdnk) & [nightly build](https://www.youtube.com/watch?v=ueNFJcby9xM) #1590 - [fixed build](https://www.youtube.com/watch?v=jDmoqu4tnaM) & [nightly build](https://www.youtube.com/watch?v=7x_N6ldlvvw)
> 
> Besides those locations, i've tested whether it happens in the other [places](https://imgur.com/a/k9y8UEc), which haven't been mentioned. And it seems that fix works well.

I think you shared the wrong nightly build video for 1598, it's the same as 1594. Tests should be equal.

patrikjuvonen:
I don't think 1594 is fixed by this PR (based on the video above). Why is it returning false for move state when the move state is supposed to be something specific? It will still be causing the same headache.

Unde-R:
> I don't think 1594 is fixed by this PR (based on the video above). Why is it returning false for move state when the move state is supposed to be something specific? It will still be causing the same headache.

because of this
https://github.com/multitheftauto/mtasa-blue/pull/1603/files#diff-25862afc0e3088d4619017944046a0d75c9a75a9ffbdb666216bb3a9478453fdR2381

Unde-R:
> I don't think 1594 is fixed by this PR (based on the video above). Why is it returning false for move state when the move state is supposed to be something specific? It will still be causing the same headache.

basically in the video of the fixed build, he was standing at first then he's trying to enter the vehicle thats why its showing false, because he's entering the vehicle
```c++
        else if (IsEnteringVehicle())
            return MOVEMENTSTATE_UNKNOWN;
```

patrikjuvonen:
Exactly. I don't understand why it has been specified as unknown. If the ped is jogging to enter a vehicle, shouldn't it return jogging?

ghost:
@patrikjuvonen i guess qaisjp [agreed upon making it false](https://github.com/multitheftauto/mtasa-blue/issues/1594#issuecomment-669576930)

He also suggested it:

```
Vehicle-related movement states

Note that getPedMoveState documents what happens when the player is in a vehicle:

    Returns a string indicating the ped's move state, or false if the ped is not streamed in, the movement type is unknown, the ped is in a vehicle or the ped is invalid.

Basically, I'm not 100% sure whether enter_vehicle and leave_vehicle fits well with the rest of these move states. What do you think?
```

ghost:
> I think you shared the wrong nightly build video for 1598, it's the same as 1594. Tests should be equal.

Oh sorry, you're right, i added wrong url. Fixed it (and in previous post as well) - https://www.youtube.com/watch?v=vAMUaDXrHvY

It might be hard to see, but it changes to "fall" once in a time (0:12 on this screen)
obraz

patrikjuvonen:
> > I think you shared the wrong nightly build video for 1598, it's the same as 1594. Tests should be equal.
> 
> Oh sorry, you're right, i added wrong url. Fixed it (and in previous post as well) - https://www.youtube.com/watch?v=vAMUaDXrHvY
> 
> It might be hard to see, but it changes to "fall" once in a time (0:12 on this screen) obraz

Thanks.

> @patrikjuvonen i guess qaisjp [agreed upon making it false](https://github.com/multitheftauto/mtasa-blue/issues/1594#issuecomment-669576930)
> 
> He also suggested it:
> 
> ```
> Vehicle-related movement states
> 
> Note that getPedMoveState documents what happens when the player is in a vehicle:
> 
>     Returns a string indicating the ped's move state, or false if the ped is not streamed in, the movement type is unknown, the ped is in a vehicle or the ped is invalid.
> 
> Basically, I'm not 100% sure whether enter_vehicle and leave_vehicle fits well with the rest of these move states. What do you think?
> ```

That is not his suggestion, the quote is what is documented in Wiki under https://wiki.multitheftauto.com/wiki/GetPedMoveState#Returns

In #1594 the ped is valid, streamed in, is not in a vehicle and I would expect the movement state to be known as well, since it returned non-false before. So I just don't understand why we're now downgrading it to return `false`. If it's now a limitation for some reason, then that's fine, but it's not fixing the issue. Alternatively we can close that issue and create a new one about an unknown state when entering a vehicle, because I believe that still needs to be worked out.

Unde-R:
> > > I think you shared the wrong nightly build video for 1598, it's the same as 1594. Tests should be equal.
> > 
> > 
> > Oh sorry, you're right, i added wrong url. Fixed it (and in previous post as well) - https://www.youtube.com/watch?v=vAMUaDXrHvY
> > It might be hard to see, but it changes to "fall" once in a time (0:12 on this screen) obraz
> 
> Thanks.
> 
> > @patrikjuvonen i guess qaisjp [agreed upon making it false](https://github.com/multitheftauto/mtasa-blue/issues/1594#issuecomment-669576930)
> > He also suggested it:
> > ```
> > Vehicle-related movement states
> > 
> > Note that getPedMoveState documents what happens when the player is in a vehicle:
> > 
> >     Returns a string indicating the ped's move state, or false if the ped is not streamed in, the movement type is unknown, the ped is in a vehicle or the ped is invalid.
> > 
> > Basically, I'm not 100% sure whether enter_vehicle and leave_vehicle fits well with the rest of these move states. What do you think?
> > ```
> 
> That is not his suggestion, the quote is what is documented in Wiki under https://wiki.multitheftauto.com/wiki/GetPedMoveState#Returns
> 
> In #1594 the ped is valid, streamed in, is not in a vehicle and I would expect the movement state to be known as well, since it returned non-false before. So I just don't understand why we're now downgrading it to return `false`. If it's now a limitation for some reason, then that's fine, but it's not fixing the issue. Alternatively we can close that issue and create a new one about an unknown state when entering a vehicle, because I believe that still needs to be worked out.

before it did return fall when entering or leaving I assume

patrikjuvonen:
> before it did return fall when entering or leaving I assume

Before this patch was initially introduced, that is (2020)? Because it's correct in this video srslyyyy posted: https://youtu.be/ueNFJcby9xM

patrikjuvonen:
I think we might be talking about two different things. There's ped moving to the "door"/entrypoint and actually physically entering the vehicle.

The actual entering animation is ok to be false.

For the moving to the "door"/entrypoint it shouldn't be false.

Unde-R:
We can do when he's entering the vehicle then we get the task name and see?

Unde-R:
"TASK_GROUP_ENTER_CAR" what about this?
but idk about group task

Einheit-101:
I just tested this. It works better than before, but 3 issues:
- Move state returns "fall" while wielding a jetpack (Maybe not an issue, but noteworthy)
- Move state returns false while trying to enter / exit a vehicle and the Ped is still running towards the vehicle
- Move state returns "fall" while standing on certain dynamic objects like this car wreck: https://imgur.com/a/zEB0Ljx

The last issue is the only problematic one because the other two can be fixed with scripting. It looks like all dynamic objects cause "fall".

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -2377,6 +2378,9 @@ eMovementState CClientPed::GetMovementState()
         else if (strcmp(szComplexTaskName, "TASK_COMPLEX_JUMP") == 0)
             return MOVEMENTSTATE_JUMP;
 
+        else if (IsEnteringVehicle())
```
theSarrum, line 2416:
```suggestion
        else if (IsEnteringVehicle() && strcmp(szSimpleTaskName, "TASK_SIMPLE_GO_TO_POINT") != 0)
```
Did a lot of tests and the `TASK_SIMPLE_GO_TO_POINT` task in MTA is only performed when the player walks to a vehicle to enter it and this is what is needed to solve #1594.

Shared/mods/deathmatch/logic/Utils.cpp
```diff
@@ -1134,3 +1134,23 @@ CVector ConvertEulerRotationOrder(const CVector& a_vRotation, eEulerRotationOrde
...
+
+bool definitelyGreaterThan(float a, float b, float epsilon)
+{
+    return (a - b) > ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * FLOAT_EPSILON);
+}
+
+bool definitelyLessThan(float a, float b, float epsilon)
+{
+    return (b - a) > ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * FLOAT_EPSILON);
+}
```
theSarrum, line 1152:
```suggestion
bool approximatelyEqual(float a, float b, float epsilon)
{
    return fabs(a - b) <= ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}

bool essentiallyEqual(float a, float b, float epsilon)
{
    return fabs(a - b) <= ((fabs(a) > fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}

bool definitelyGreaterThan(float a, float b, float epsilon)
{
    return (a - b) > ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}

bool definitelyLessThan(float a, float b, float epsilon)
{
    return (b - a) > ((fabs(a) < fabs(b) ? fabs(b) : fabs(a)) * epsilon);
}
```

theSarrum (changes requested):

PR #1604, opened 2020-08-08
Hide encode function's secret key in debughooks

gptrk0:
Prevent to catch secret key, what passed to encode function.
- encodeString
- decodeString
- teaEncode
- teaDecode


ccw808 (approved):

patrikjuvonen (approved):

qaisjp:
Other comments start from arg 0, how come these start from 1 or 2?

gptrk0:
> Other comments start from arg 0, how come these start from 1 or 2?

It's the index of the argument list. Secret key at the 3rd place in encodeString. (index 2, because counts from 0)

qaisjp:
My point is that the other comments include details about the previous parameters, whereas yours skips and starts from the secret parameter. E.g. this exists:

```
{"logIn", {{EArgType::Password, 2}}},                           // player, account, 2=PASSWORD
```

But yours is:

```cpp
{"encodeString", {{EArgType::MaxArgs, 2}}},                     // 2=SECRETKEY, CALLBACK
```

I think it should be something like:

```cpp
{"encodeString", {{EArgType::MaxArgs, 2}}},                     // algorithm, input, 2=SECRETKEY, ...
```

qaisjp (approved):

StrixG (approved):

patrikjuvonen:
Thank you for the pull request!

PR #1607, opened 2020-08-08
Rename setCameraShakeLevel to setCameraDrunkLevel & getCameraDrunkLevel

### LINKED ISSUE #723: setCameraShakeLevel(0) setting wrong camera shake level
**Describe the bug**
When setting setCameraShakeLevel(0) camera will not always stop shaking, instead it will act like if you used setCameraShakeLevel(255).

**To reproduce**
1. crun setCameraShakeLevel(255)
2. crun setCameraShakeLevel(0)
3. Repeat points 1 and 2 multiple times until when setting camera shake level 0 the camera will start shaking again.
4. crun getCameraShakeLevel() will return 0 when it's obvious the camera isn't steady, but still shaking.

**Expected behaviour**
I'm expecting the camera to stop shaking when doing crun setCameraShakeLevel(0)

**Version**
Executing client-side command: for i, v in pairs(getVersion( )) do outputChatBox(tostring(i)..": "..tostring(v)) end
number: 342
netcode: 474
name: MTA:SA Client
sortable: 1.5.6-9.16068.0
mta: 1.5.6
type: Release
os: Windows
tag: 1.5.6-release-16068
Command executed!

Unde-R:
-This may fix multitheftauto#723 since it prevents overwritting the drunk level incase entering same value
-Also in addition to what Lpsd said in multitheftauto#1197 I renamed it
-Both ( setCameraShakeLevel & getCameraShakeLevel ) to ( setCameraDrunkLevel & getCameraDrunkLevel )
@Patrick2562 If you can test It, since I couldn't reproduce the bug

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
@@ -22,7 +22,7 @@ void CLuaCameraDefs::LoadFunctions()
         {"getCameraTarget", GetCameraTarget},
         {"getCameraInterior", GetCameraInterior},
         {"getCameraGoggleEffect", GetCameraGoggleEffect},
-        {"getCameraShakeLevel", GetCameraShakeLevel},
+        {"getCameraDrunkLevel", GetCameraDrunkLevel},
         {"getCameraFieldOfView", GetCameraFieldOfView},
 
         // Cam set funcs
@@ -35,7 +35,7 @@ void CLuaCameraDefs::LoadFunctions()
         {"getCameraClip", GetCameraClip},
         {"setCameraViewMode", SetCameraViewMode},
         {"setCameraGoggleEffect", SetCameraGoggleEffect},
-        {"setCameraShakeLevel", SetCameraShakeLevel},
+        {"setCameraDrunkLevel", SetCameraDrunkLevel},
     };
 
     // Add functions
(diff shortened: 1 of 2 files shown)


gptrk0:
Still bad, sadly.

Einheit-101:
maybe add new function "setCameraShakeLevel" that does actually what it is supposed to do, shake the camera like an explosion does.

Unde-R:
yeah maybe, but we need to fix this first

Disinterpreter:
In SAMP it calls `GetPlayerDrunkLevel` =/ But I think the drunk camera is more understandable

https://github.com/multitheftauto/amx/blob/9dfc9108c3abf2a72867f606a8dcb21b5f6f2a11/amx/server/natives/a_players.lua#L112-L115

qaisjp:
Can't give a proper review right now, but this is not backwards compatible.

Unde-R:
agree

qaisjp:
Please can we have a rename in a separate PR?

Unde-R:
> Please can we have a rename in a separate PR?

Sure.

Unde-R:
what about this, we can revert it?

qaisjp:
What I mean is, yes, don't include the rename in this PR (you should remove it). Each PR should solve a distinct feature

Unde-R:
Oh Alright

Unde-R:
Maybe we can close or rename this PR

qaisjp:
Yep rename

patrikjuvonen:
Closed in favor of #1631

PR #1609, opened 2020-08-10
Fix isPlayerHudComponentVisible incompatible with command `showhud`

### LINKED ISSUE #547: isPlayerHudComponentVisible only works with setPlayerHudComponentVisible
**Describe the bug**

Examples when it is wrong:

**To reproduce**
- `showhud`

**Expected behaviour**

**Screenshots**

**MTA Client:** 
 - Version: 
 - Running in Wine? 

**MTA Server:** 
 - Version 
 - OS: 

**Additional context**

Unde-R:
this should Fix #547 since the command `showhud` disables the hud and `isPlayerHudComponentVisible` only checks if the component is visible so It makes no sense. 
This should be enough but feel free to say anything.

### DIFF
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -276,6 +276,8 @@ bool CHudSA::IsComponentVisible(eHudComponent component)
     SHudComponent* pComponent = MapFind(m_HudComponentMap, component);
     if (pComponent)
     {
+        if (g_pCore->GetGame()->GetHud()->IsDisabled())
+            return false;
         // Determine if invisible by matching data with disabled values
         uchar* pSrc = (uchar*)(&pComponent->disabledData);
         uchar* pDest = (uchar*)(pComponent->uiDataAddr);
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -271,13 +271,16 @@ void CHudSA::SetComponentVisible(eHudComponent component, bool bVisible)
 //
 // CHudSA::IsComponentVisible
 //
-bool CHudSA::IsComponentVisible(eHudComponent component)
+bool CHudSA::IsComponentVisible(eHudComponent component, bool bOutIsEnabled)
 {
     SHudComponent* pComponent = MapFind(m_HudComponentMap, component);
     if (pComponent)
     {
-        if (g_pCore->GetGame()->GetHud()->IsDisabled())
-            return false;
+        if (bOutIsEnabled)
+        {
+            if (g_pCore->GetGame()->GetHud()->IsDisabled())
+                return false;
+        }
         // Determine if invisible by matching data with disabled values
         uchar* pSrc = (uchar*)(&pComponent->disabledData);
         uchar* pDest = (uchar*)(pComponent->uiDataAddr);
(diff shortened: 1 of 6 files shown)


patrikjuvonen:
> This is a weird one. There needs to be a distinction between "visible" and "enabled"...

_Originally posted by @qaisjp in https://github.com/multitheftauto/mtasa-blue/issues/547#issuecomment-423599054_

If I understand correctly this pull request would make it impossible to know whether a component has been disabled by script or user. I think this is a negative side effect that must be solved before merge. Also, maybe allow getting `showhud` state using another function or make new function to know if it's toggled by the user (maybe there is a function already?)

Unde-R:
Uhh.. maybe we can work on `CCommandFuncs::HUD`

qaisjp (changes requested):
I agree we can't merge the current version of this. I've explained why at https://github.com/multitheftauto/mtasa-blue/issues/547#issuecomment-671595861.

Let's discuss the solution on #547 and we've settled on a solution you can implement it here if you want.

Unde-R:
Here is the result:
[`https://pastebin.com/mAqf4hmU`](https://pastebin.com/mAqf4hmU)

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1951,9 +1951,9 @@ bool CStaticFunctionDefinitions::ShowPlayerHudComponent(eHudComponent component,
     return true;
 }
 
-bool CStaticFunctionDefinitions::IsPlayerHudComponentVisible(eHudComponent component, bool& bOutIsVisible)
+bool CStaticFunctionDefinitions::IsPlayerHudComponentVisible(eHudComponent component, bool bOutIsEnabled, bool& bOutIsVisible)
```
qaisjp, line 1954: This, and elsewhere, should be `bCheckEnabled`.

Client/game_sa/CHudSA.cpp
```diff
@@ -271,11 +271,16 @@ void CHudSA::SetComponentVisible(eHudComponent component, bool bVisible)
...
+bool CHudSA::IsComponentVisible(eHudComponent component, bool bOutIsEnabled)
 {
     SHudComponent* pComponent = MapFind(m_HudComponentMap, component);
     if (pComponent)
     {
+        if (bOutIsEnabled)
+        {
+            if (g_pCore->GetGame()->GetHud()->IsDisabled())
+                return false;
+        }
```
qaisjp, line 283: Why not use `if (bIsEnabled && ...->IsDisabled())`?

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1951,9 +1951,9 @@ bool CStaticFunctionDefinitions::ShowPlayerHudComponent(eHudComponent component,
...
 {
-    bOutIsVisible = g_pGame->GetHud()->IsComponentVisible(component);
+    bOutIsVisible = g_pGame->GetHud()->IsComponentVisible(component, bOutIsEnabled);
     return true;
 }
```
ghost, line 1958: Since you're modifying this function, it's probably better to remove it. We don't put code in `CStaticFunctionDefinitions` anymore. You can call `g_pGame->GetHud()->IsComponentVisible` directly. The rest of the PR looks okay.

qaisjp (commented):
Had these two comments from a draft review ages ago, I guess I forgot to submit

Client/game_sa/CHudSA.cpp
Unde-R: yeah sure

PR #1612, opened 2020-08-12
Fix #1474: Implement getAllElementData client-side

### LINKED ISSUE #1474: Implement getAllElementData client-side
Currently, [getAllElementData](https://wiki.multitheftauto.com/wiki/GetAllElementData) is only available serverside. Since you are able to set element data clientside (without server or remote client sync) it would make sense for this function to also be available clientside.

Identified by Liberty#5062 (Discord)

Unde-R:
Fixes #1474


Client/mods/deathmatch/logic/CClientEntity.h
```diff
@@ -198,6 +198,7 @@ class CClientEntity : public CClientEntityBase
 
     CCustomData*  GetCustomDataPointer() { return m_pCustomData; }
     CLuaArgument* GetCustomData(const char* szName, bool bInheritData, bool* pbIsSynced = nullptr);
+    CLuaArguments* GetAllCustomData(CLuaArguments* table);
     bool          GetCustomDataString(const char* szKey, SString& strOut, bool bInheritData);
     bool          GetCustomDataFloat(const char* szKey, float& fOut, bool bInheritData);
```
qaisjp: Please format the document! (I can tell this is not formatted correctly because the types will all line up.)

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -303,6 +303,22 @@ CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData
     return NULL;
 }
 
+CLuaArguments* CClientEntity::GetAllCustomData(CLuaArguments* table)
+{
+    assert(table);
+
+    // Grab it and return a pointer to the variable
+    std::map<std::string, SCustomData>::const_iterator iter = m_pCustomData->IterBegin();
+    for (; iter != m_pCustomData->IterEnd(); iter++)
```
qaisjp: Use `for (auto it = m_pCustomData->IterBeing();`.

I recognise this is copied from `CLuaArguments* CElement::GetAllCustomData` (server) but when we unify the code (if ever), we can just select the better version (this one).

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -303,6 +303,22 @@ CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData
...
+    {
+        table->PushString(iter->first.c_str());                // key
+        table->PushArgument(iter->second.Variable);            // value
+    }
+
+    return table;
+}
+
+
```
qaisjp:
```suggestion

```

Please only leave exactly one blank line between methods.

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -69,6 +68,7 @@ void CLuaElementDefs::LoadFunctions()
 
         // Element data
         {"getElementData", GetElementData},
+        {"getAllElementData", GetAllElementData},
```
qaisjp: Is it possible for you to use `ArgumentParser`? See https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp for an example.

Let me know if you run into difficulties with the CLuaArguments return type.

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
```diff
@@ -50,6 +50,41 @@ int CLuaElementDefs::GetElementData(lua_State* luaVM)
...
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadUserData(pElement);
+
+    if (!argStream.HasErrors())
+    {
+        CLuaArguments Args;
+#ifdef MTA_CLIENT
+        CLuaArguments* pVariable = pElement->GetAllCustomData(&Args);
+#else
+        CLuaArguments* pVariable = CStaticFunctionDefinitions::GetAllElementData(pElement, &Args);
```
qaisjp: You can just get rid of `CStaticFunctionDefinitions::GetAllElementData` server-side. It's only used here and is defined as this (below), which is essentially the same thing:

https://github.com/multitheftauto/mtasa-blue/blob/6941e0b0700a6d8c2775cc65de8022496b55e4d4/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L403-L409

qaisjp (changes requested):
Firstly, I didn't mention this earlier, I'd just like to thank you so much for getting involved with mtasa-blue development :)

And thank you so much for this PR and working on our backlog :)

I've left a few comments with the changes I'd like to see addressed.

Client/mods/deathmatch/logic/CClientEntity.h
Unde-R: Yeah sure

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Unde-R: I'd love to do It with ArgumentParser but It's confusing, couldn't do It sadly

Unde-R: I'm using ArgumentParser rn but I don't know the return type, you can help?

qaisjp: There does not seem to be an obvious way.

This WON'T work because, while it will put the table on the stack, it will tell Lua that 0 arguments were provided.
```cpp
void GetAllElementData(lua_State* L, CElement* pElement) {
    // logic..
    pVariable->PushAsTable(L);
}
```

Might be nice to create a special return type that automatically wraps returned CLuaArguments, and PushesAsTable when appropriate.

Or support a `LuaReturn<N>` so that we can push our own stuff on the stack and tell the argument parser what we want to do. 

Sorry for wasting your time! I won't block your PR on this change. I'll make a separate requested change that will solve the surface problem.

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
```diff
@@ -50,6 +50,37 @@ int CLuaElementDefs::GetElementData(lua_State* luaVM)
...
+        CLuaArguments Args;
+        CLuaArguments* pVariable = pElement->GetAllCustomData(&Args);
+        if (pVariable)
+        {
+            pVariable->PushAsTable(luaVM);
+            return 1;
+        }
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
```
qaisjp: As mentioned in https://github.com/multitheftauto/mtasa-blue/pull/1612#discussion_r470167380, we no longer log warnings for usage mistakes. These are now errors! See #821.




```suggestion
        return luaL_error(luaVM, argStream.GetFullErrorMessage());
```

qaisjp (changes requested):

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -303,6 +303,20 @@ CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData
...
 }
 
+CLuaArguments* CClientEntity::GetAllCustomData(CLuaArguments* table)
+{
+    assert(table);
+
+    // Grab it and return a pointer to the variable
+    for (auto it = m_pCustomData->IterBegin(); it != m_pCustomData->IterEnd(); it++)
+    {
+        table->PushString(it->first.c_str());                // key
```
qaisjp: is `.c_str()` actually required? I have a hunch it will work without.

Unde-R: We can remove it and see

Unde-R: Yeah, It's working fine

qaisjp: select the function and then do ADvanced -> Format Selection :)

Unde-R: Done, we need just the return type I guess now.

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -303,6 +303,20 @@ CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData
     return NULL;
 }
 
+CLuaArguments* CClientEntity::GetAllCustomData(CLuaArguments* table)
+{
+    assert(table);
+
+    // Grab it and return a pointer to the variable
```
Pirulax: Comment makes no sense. The `Variable` is a value by itself, so you're passing a value to `PushArgument`, so you copy it.
The `key` is an `SString` so you copy it as well.
Correct me if im wrong.

Shared/mods/deathmatch/logic/luadefs/CLuaElementDefsShared.cpp
```diff
@@ -50,6 +50,37 @@ int CLuaElementDefs::GetElementData(lua_State* luaVM)
...
 
+int CLuaElementDefs::GetAllElementData(lua_State* luaVM)
+{
+    //  table getAllElementData ( element theElement )
+
+#ifdef MTA_CLIENT
+    CClientEntity* pElement;
+#else
+    CElement* pElement;
+#endif
```
Pirulax: Just use CElement, its #defined to CClientEntity on client side.

Pirulax (dismissed):

Client/mods/deathmatch/logic/CClientEntity.cpp
Unde-R: Yeah that's the same code from https://github.com/multitheftauto/mtasa-blue/blob/927509565ab3a8d48df006440654c38e0942aa89/Server/mods/deathmatch/logic/CElement.cpp#L521

Pirulax:
This should be refactored to use the new parser.

Unde-R:
@StrixG what about the speed of this is better or less?

StrixG:
> @StrixG what about the speed of this is better or less?

I haven't tested it yet, but I think the new parser is slower.

patrikjuvonen:

patrikjuvonen (approved):

PR #1613, opened 2020-08-13
Workaround for custom map collision crashes

ghost:
There is a crash that happens due to problems with collision files, reported by @Dutchman101. All players on a server will start crashing until the resource with the problematic collision file has been restarted. Crash dumps show that there is an infinite loop that happens in CPhysical class. This infinite loop happens because `CPhysical::GetBoundRect` and `CEntity::GetBoundRect` function returns a bounding box with incorrect values sometimes. The minimum value is greater than the maximum value, and this causes an infinite loop. I'm not sure what causes this corruption, but if this PR can fix the problem, then I'm hopeful I can get to the bottom of it.

### DIFF
--- a/Client/game_sa/CEntitySA.cpp
+++ b/Client/game_sa/CEntitySA.cpp
@@ -45,7 +45,7 @@ CRect* CEntitySAInterface::GetBoundRect_(CRect* pRect)
     TransformFromObjectSpace(point, vecMax);
     rect.StretchToPoint(point.fX, point.fY);
     *pRect = rect;
-    pRect->FixIncorrectTopLeft(); // Fix #1613: custom map collision crashes in CPhysical class (infinite loop)
+    pRect->FixIncorrectTopLeft();            // Fix #1613: custom map collision crashes in CPhysical class (infinite loop)
     return pRect;
 }
 
--- a/Client/game_sa/CEntitySA.cpp
+++ b/Client/game_sa/CEntitySA.cpp
@@ -16,6 +16,39 @@ extern CGameSA* pGame;
 unsigned long CEntitySA::FUNC_CClumpModelInfo__GetFrameFromId;
 unsigned long CEntitySA::FUNC_RwFrameGetLTM;
 
+void CEntitySAInterface::TransformFromObjectSpace(CVector& outPosn, CVector const& offset)
+{
+    ((void(__thiscall*)(CEntitySAInterface*, CVector&, CVector const&))0x533560)(this, outPosn, offset);
+}
+
+CVector* CEntitySAInterface::GetBoundCentre(CVector* pOutCentre)
+{
+    return ((CVector * (__thiscall*)(CEntitySAInterface*, CVector*))0x534250)(this, pOutCentre);
+}
+
+CRect* CEntitySAInterface::GetBoundRect_(CRect* pRect)
+{
+    CColModelSAInterface* colModel = CModelInfoSAInterface::GetModelInfo(m_nModelIndex)->pColModel;
+    CVector               vecMin = colModel->boundingBox.vecMin;
+    CVector               vecMax = colModel->boundingBox.vecMax;
+    CRect                 rect;
+    CVector               point;
+    TransformFromObjectSpace(point, vecMin);
+    rect.StretchToPoint(point.fX, point.fY);
+    TransformFromObjectSpace(point, vecMax);
+    rect.StretchToPoint(point.fX, point.fY);
+    float maxX = vecMax.fX;
+    vecMax.fX = vecMin.fX;
+    vecMin.fX = maxX;
+    TransformFromObjectSpace(point, vecMin);
+    rect.StretchToPoint(point.fX, point.fY);
+    TransformFromObjectSpace(point, vecMax);
+    rect.StretchToPoint(point.fX, point.fY);
+    *pRect = rect;
+    pRect->FixIncorrectTopLeft();            // Fix custom map collision crashes in CPhysical class (infinite loop)
+    return pRect;
+}
+
 CEntitySA::CEntitySA()
 {
     // Set these variables to a constant state
@@ -25,6 +58,12 @@ CEntitySA::CEntitySA()
     DoNotRemoveFromGame = false;
     m_pStoredPointer = NULL;
     m_ulArrayID = INVALID_POOL_ARRAY_ID;
+    InstallHooks();
+}
+
+void CEntitySA::InstallHooks()
+{
+   HookInstall(0x534120, &CEntitySAInterface::GetBoundRect_);
 }
 
 /*VOID CEntitySA::SetModelAlpha ( int iAlpha )
--- a/Client/game_sa/CEntitySA.cpp
+++ b/Client/game_sa/CEntitySA.cpp
@@ -45,7 +45,7 @@ CRect* CEntitySAInterface::GetBoundRect_(CRect* pRect)
     TransformFromObjectSpace(point, vecMax);
     rect.StretchToPoint(point.fX, point.fY);
     *pRect = rect;
-    pRect->FixIncorrectTopLeft();            // Fix custom map collision crashes in CPhysical class (infinite loop)
+    pRect->FixIncorrectTopLeft(); // Fix #1613: custom map collision crashes in CPhysical class (infinite loop)
     return pRect;
 }
 
--- a/Client/game_sa/CEntitySA.cpp
+++ b/Client/game_sa/CEntitySA.cpp
@@ -49,6 +49,11 @@ CRect* CEntitySAInterface::GetBoundRect_(CRect* pRect)
     return pRect;
 }
 
+void CEntitySAInterface::StaticSetHooks()
+{
+    HookInstall(0x534120, &CEntitySAInterface::GetBoundRect_);
+}
+
 CEntitySA::CEntitySA()
 {
     // Set these variables to a constant state
@@ -58,12 +63,6 @@ CEntitySA::CEntitySA()
     DoNotRemoveFromGame = false;
     m_pStoredPointer = NULL;
     m_ulArrayID = INVALID_POOL_ARRAY_ID;
-    InstallHooks();
-}
-
-void CEntitySA::InstallHooks()
-{
-   HookInstall(0x534120, &CEntitySAInterface::GetBoundRect_);
 }
 
 /*VOID CEntitySA::SetModelAlpha ( int iAlpha )
--- a/Client/game_sa/CPhysicalSA.cpp
+++ b/Client/game_sa/CPhysicalSA.cpp
@@ -17,7 +17,7 @@ CRect* CPhysicalSAInterface::GetBoundRect_(CRect* pRect)
     CEntitySAInterface::GetBoundCentre(&boundCentre);
     float fRadius = CModelInfoSAInterface::GetModelInfo(m_nModelIndex)->pColModel->boundingBox.fRadius;
     *pRect = CRect(boundCentre.fX - fRadius, boundCentre.fY - fRadius, boundCentre.fX + fRadius, boundCentre.fY + fRadius);
-    pRect->FixIncorrectTopLeft(); // Fix #1613: custom map collision crashes in CPhysical class (infinite loop)
+    pRect->FixIncorrectTopLeft();            // Fix #1613: custom map collision crashes in CPhysical class (infinite loop)
     return pRect;
 }
 
--- a/Client/game_sa/CPhysicalSA.cpp
+++ b/Client/game_sa/CPhysicalSA.cpp
@@ -21,13 +21,7 @@ CRect* CPhysicalSAInterface::GetBoundRect_(CRect* pRect)
     return pRect;
 }
 
-
-CPhysicalSA::CPhysicalSA()
-{
-    InstallHooks();
-}
-
-void CPhysicalSA::InstallHooks()
+void CPhysicalSAInterface::StaticSetHooks()
 {
     HookInstall(0x5449B0, &CPhysicalSAInterface::GetBoundRect_);
 }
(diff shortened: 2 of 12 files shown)


tederis:
I found this bug in 2010. And it was a pain to replace models on custom mods. Сongratulations, comrades!

Client/game_sa/CEntitySA.cpp
```diff
@@ -16,6 +16,39 @@ extern CGameSA* pGame;
...
+    rect.StretchToPoint(point.fX, point.fY);
+    float maxX = vecMax.fX;
+    vecMax.fX = vecMin.fX;
+    vecMin.fX = maxX;
+    TransformFromObjectSpace(point, vecMin);
+    rect.StretchToPoint(point.fX, point.fY);
+    TransformFromObjectSpace(point, vecMax);
+    rect.StretchToPoint(point.fX, point.fY);
+    *pRect = rect;
+    pRect->FixIncorrectTopLeft();            // Fix custom map collision crashes in CPhysical class (infinite loop)
```
qaisjp: Nit:

```suggestion
    pRect->FixIncorrectTopLeft();            // Fix #1613 custom map collision crashes in CPhysical class (infinite loop)
```


Essentially, it would be good for a future reader to know why this line was added. Since we don't have an issue out for this bug, linking to this PR (#1613) since it has a sufficient enough description.

If we incorporate https://github.com/multitheftauto/mtasa-blue/pull/1613#discussion_r470236477 then we could perhaps just include a comment in the function definition (and perhaps a hint "read ThisFuncName for more information on why this is necessary").

Client/game_sa/CPhysicalSA.cpp
```diff
@@ -11,6 +11,27 @@
 
 #include "StdInc.h"
 
+CRect* CPhysicalSAInterface::GetBoundRect_(CRect* pRect)
+{
+    CVector boundCentre;
+    CEntitySAInterface::GetBoundCentre(&boundCentre);
+    float fRadius = CModelInfoSAInterface::GetModelInfo(m_nModelIndex)->pColModel->boundingBox.fRadius;
+    *pRect = CRect(boundCentre.fX - fRadius, boundCentre.fY - fRadius, boundCentre.fX + fRadius, boundCentre.fY + fRadius);
+    pRect->FixIncorrectTopLeft(); // Fix custom map collision crashes in CPhysical class (infinite loop)
```
qaisjp:
```suggestion
    pRect->FixIncorrectTopLeft(); // Fix #1613 custom map collision crashes in CPhysical class (infinite loop)
```

Same feedback as above, just including it here for completeness.

Client/game_sa/HookSystem.h
```diff
@@ -16,7 +16,33 @@
...
+    {
+        MemCpy((PVOID)dwInstallAddress, JumpBytes, iJmpCodeSize);
+        return TRUE;
+    }
+    else
+    {
+        return FALSE;
+    }
+}
+
```
qaisjp, line 45: Correct me if I am wrong, but this seems like a pretty core change that isn't related to this crash fix. Is there a particular reason it's here instead of in a separate PR?

Client/game_sa/CEntitySA.cpp
```diff
@@ -16,6 +16,44 @@ extern CGameSA* pGame;
...
+    TransformFromObjectSpace(point, vecMax);
+    rect.StretchToPoint(point.fX, point.fY);
+    float maxX = vecMax.fX;
+    vecMax.fX = vecMin.fX;
+    vecMin.fX = maxX;
+    TransformFromObjectSpace(point, vecMin);
+    rect.StretchToPoint(point.fX, point.fY);
+    TransformFromObjectSpace(point, vecMax);
+    rect.StretchToPoint(point.fX, point.fY);
+    *pRect = rect;
```
qaisjp, line 47: Optional: I'm aware this is just the original code reimplemented, but it would be good to have some code comments explaining what GetBoundRect_ does, and if necessary, what is going on here.

Client/game_sa/CRadarSA.cpp
```diff
@@ -175,16 +175,14 @@ VOID CRadarSA::ClearMarkerForEntity(CPed* ped)
...
-
 VOID CRadarSA::DrawAreaOnRadar(float fX1, float fY1, float fX2, float fY2, const SColor color)
 {
     // Convert color to required abgr at the last moment
     unsigned long abgr = color.A << 24 | color.B << 16 | color.G << 8 | color.R;
-    CRect         myRect;
+    struct
+    {
+        float fX1, fY1, fX2, fY2;
+    } myRect;
```
qaisjp: Side question: how come this is a custom struct and not a CRect? Is a `CRect` just too heavyweight?

Shared/sdk/CRect.h
```diff
@@ -0,0 +1,72 @@
...
+        right = 0.0f;
+        top = 0.0f;
+    }
+
+    // custom function. Used for fixing custom map collision crashes in CPhysical class (infinite loop)
+    void FixIncorrectTopLeft()
+    {
+        if (left > right) left = right - 1.0f;
+        if (top > bottom) top = bottom - 1.0f;
+    }
```
qaisjp: This seems like a really specific function that will probably not be used again. Maybe it makes sense to just have a static method `CEntitySA::FixRectIncorrectTopLeft(CRect &rect)` and call that, instead of including this logic inside the `CRect` class itself?

(Come to this review comment from another review comment? [Go back](https://github.com/multitheftauto/mtasa-blue/pull/1613#discussion_r470230671), if you want.)

Shared/sdk/CRect.h
```diff
@@ -0,0 +1,72 @@
...
+        left = 0.0f;
+        bottom = 0.0f;
+        right = 0.0f;
+        top = 0.0f;
+    }
+
+    // custom function. Used for fixing custom map collision crashes in CPhysical class (infinite loop)
+    void FixIncorrectTopLeft()
+    {
+        if (left > right) left = right - 1.0f;
```
qaisjp, line 88: Why was `1.0f` chosen here instead of `0.1337f` or `1.337f` or `0.5f`? Seems like an arbitrary number? A code comment explaining this would be good. Or maybe I'm just missing the obvious significance.

qaisjp (changes requested):
Nice fix. Hopefully we can get this into 1.5.8. I've left a couple cleanliness and structural comments, please don't mind my terseness, I'm in business mode 😎.

Also, if possible, it would also be good to have a zip of an example collision file. This way we can reproduce the issue and confirm that the fix works. And if a similar problem comes happens in the future, we would have something to test with so we can actually check if the bug came back somehow.

Client/game_sa/CRadarSA.cpp
ghost: Ah correct, this should be CRect. MTA has its own version of CRect, it's called CRect2D. I confused this with CRect2D.

Shared/sdk/CRect.h
ghost: It's a bit confusing, but I'll try my best to explain. The value must be greater than or equal to 1.0f because let's say if the value of `left` is 15.5, when it's casted to int, it will be rounded off to 15, so we need to subtract at least 0.6 to make make it 14 when it's casted to int . It doesn't really matter which value we subtract because the bounding box values are already corrupted at this point, and we're just trying to avert the crash by making sure that the top and left values are always less than the bottom and right values.

Client/game_sa/HookSystem.h
ghost: The change is needed by this PR because, in C++, you can't get the address of class methods and cast it to unsigned int or DWORD, so this is basically a dirty hack that allows us to get the address of class methods. If I remove this change, then this code will not compile:
```
HookInstall(0x534120, &CEntitySAInterface::GetBoundRect_);
```

Shared/sdk/CRect.h
ghost: Yeah, we can do that, but I don't see any benefit. I think this function belongs in CEntitySAInterface class because it's totally related to it and *SA classes are mainly used because we can't add new members to the actual gta classes.

Client/game_sa/CEntitySA.cpp
ghost: This code would've made sense if we had the source code for TransformFromObjectSpace function. `TransformFromObjectSpace` will multiply the matrix of the entity with the offset vector in 2nd paramter and store the result in _point_.  
`StretchToPoint` performs sanity checks and clamps the bounding box (rect variable) min and max values to the point.

Client/game_sa/CEntitySA.cpp
```diff
@@ -45,7 +45,7 @@ CRect* CEntitySAInterface::GetBoundRect_(CRect* pRect)
     TransformFromObjectSpace(point, vecMax);
     rect.StretchToPoint(point.fX, point.fY);
     *pRect = rect;
-    pRect->FixIncorrectTopLeft();            // Fix custom map collision crashes in CPhysical class (infinite loop)
```
qaisjp: Btw the comment indentation change here is probably not in line with the auto-formatter. Try Format Selection? (I think as a shorthand if you type a semicolon at Line 50 it will just format the entire method, and then you can remove the semicolon)

ghost: I modified the indentation manually. Fixed now.

ghost:
@qaisjp any more improvements needed?

Shared/sdk/CRect.h
qaisjp: Please can we have this as a code comment?

Shared/sdk/CRect.h
qaisjp: The benefit is that, since this code isn't really relevant for all rectangles in general, it shouldn't be in CRect.

qaisjp: > . I think this function belongs in CEntitySAInterface class because it's totally related to it and *SA classes are mainly used because we can't add new members to the actual gta classes.

Feel free to do that. Unless this is a typo and you meant CRect?

ghost: > Feel free to do that. Unless this is a typo and you meant CRect?

Ah, yes. I meant CRect, like the way it is right now.

Client/game_sa/CRadarSA.cpp
qaisjp: Might be useful to include a comment on the CRect class that explains when and why it should be used

qaisjp (approved):
left comments in the threads that are currently open but i'll leave that up to you

Dutchman101:
Relevant crash offsets (all suffering from the same infinite loop, causing POINTER_DOUBLE_LINK_POOL usage to be 100%, for the mentioned reason):

0x0013368b (`CPtrListDoubleLink::Add` - main crash, high incidence at all times)
0x00144bc8 (`Physical::Add`)
0x00142721 (`CPhysical::RemoveAndAdd`)

Custom maps are just risky business without a fix.. your server could die when something random happens, which is unknown, and players start crashing on connect. The crashes are top crashes in MTA, like the first offset had thousands of victims on a single build.

Dutchman101:
Just for the record, this wasn't a (complete) fix.. it is possible that it reduces crash prevalence, so we're better off to keep it in after all. But maybe @tederis has some ideas, and i will get around to creating an actual issue for it soon.

Note: saml1er and I were aware it wasn't really fixed (should have posted it here), way before leaving MTA he sort of understood the problem, described it, and had plans on what to do next. But last week he also deleted his Github and discord accounts so that conversation is gone and i cannot look up the details. I forgot what it was. 
But it may have been something like "there are more functions/calls that can return the bad data which leads to a crash.. we need to add these measures in more places" (don't take my word for it)

It's definately not caused by _corrupt collisions inside a map object model_ (unless some GTA object with such a problem is the trigger) as I can get it to happen by loading extended custom maps using only default GTA objects (no custom ones), waiting for players to potentially start crashing. I do believe that something predisposes a server to have it happen, though. It is just an ancient MTA bug.

tederis:
I have an idea and possibly can fix the crash. But I cannot reproduce the crash at the moment. I would be grateful if anyone gives me a custom map example(it can be just a IPL if the map consists of default GTA objects).

Dutchman101:
> I have an idea and possibly can fix the crash. But I cannot reproduce the crash at the moment. I would be grateful if anyone gives me a custom map example(it can be just a IPL if the map consists of default GTA objects).

It's totally random, and the chance to get it triggered (a point where many or all players start crashing) is tied to using custom maps. Even when no custom models are involved, so i seriously doubt it has much to do with corruption in a world model or its collision, unless GTA messed up with a default object somewhere. But that wouldn't explain why it triggers randomly, more like MTA may be corrupting something internal (to GTA) under specific circumstances.

Since the issue is so obscure, the closest you can get to reproduction is to have a server (preferably one that encountered it before) run a set of custom maps, waiting patiently until the unknown thing triggers and starts the client crashes. Then you can come with a debugger attached or run other tests to pinpoint it.

In the past i was talking to an affected guy, and he made this post: [https://pastebin.com/GxYcDjq7](https://pastebin.com/GxYcDjq7) which implies it's even more obscure and non-understandable. That hints at a chance vehicles (or resources using or allocating vehicles - that guy told me it happened straight after he changed freeroam vehicle spawning script code.. and during that support session I even validated that restarting 'freeroam' would stop the crashes immediately) rather than maps, are involved - I also recall some mantis issue that was about a crash with 1 of these related offsets, but mentions it was due to the "suspension ballast" on a Sandking (vehicle). It might be a coincidence that it triggered a crash on the same offset (being a different crash) but idk.

It's all very confusing, but I am certain that even the above column is somehow relevant to the underlying issue.

Also, crash popularity (on 0x0013368b) is the highest on these servers:

```
⚝ BRASILEIROS VIVENDO NO STYLO ⚝ | FREEROAM | ➠ 『 Gangs - Policias - Eventos - Drift 』

SeeMTA v3 LV - Connect Server || Ide kell csatlakozni! || lv.see-game.com

[BCG #1] ❖ BRASIL CIDADE GRANDE ❖ ROLEPLAY ➠ 『 Policias • Facções • Assaltos 』
```
You could try to contact the server owners and work out what makes them special in this regards (what they are doing).

tederis:
So, I tried to fix it. The PR: https://github.com/multitheftauto/mtasa-blue/pull/2198

PR #1619, opened 2020-08-17
Revert "Fix #1443 - trim teaEncode trailing zeroes"

### LINKED ISSUE #1443: Tea encoding bug

StrixG:
#1534 trims all trailing zeroes including those that should be there.
```lua
local str = encodeString("tea", "Hel\0lo\0", {key = "1234"})
iprint(decodeString("tea", str, {key = "1234"}) -- "Hel\0lo"
```
We need to find another solution for #1443.
Reverts multitheftauto/mtasa-blue#1534

### DIFF
--- a/Shared/sdk/SharedUtil.Hash.h
+++ b/Shared/sdk/SharedUtil.Hash.h
@@ -80,6 +80,9 @@ namespace SharedUtil
         unsigned char m_digest[16];
     };
 
+    void encodeXtea(unsigned int* v, unsigned int* w, unsigned int* k);
+    void decodeXTea(unsigned int* v, unsigned int* w, unsigned int* k);
+
     void TeaEncode(const SString& str, const SString& key, SString* out);
     void TeaDecode(const SString& str, const SString& key, SString* out);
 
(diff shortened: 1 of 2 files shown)


ccw808 (approved):

Pirulax:
We should use cryptopp or something, instead of this home made stuff.
But I'm not sure whenever Crypto++'s TEA will behave the same.

PR #1625, opened 2020-08-19
Fix for breaking backwards compatibility with oop functions returning vectors

sbx320:
In the resolution to #1240 we've changed oop functions to consistently return vectors, rather than three numbers per vector in `Vehicle.getComponentPosition`, `Vehicle.getComponentRotation` and `Element.getBoundingBox`. It appears that the incorrect variants have been used in some scripts though, which are now broken. 

This PR adds a workaround for the most common use case. Rather than always returning a vector, we conditionally return a vector based on what the caller expected at the callsite (see d3257e4 for the Lua API changes for that). If the caller expected three values, we return three numbers otherwise a vector. This allows both of the following call variants to work
```
local veh = Vehicle(411, 0, 0, 0)
local x, y, z = veh:getCompontentPosition("[...]")
-- x, y, z are now numbers
-- but also:
local vec = veh:getCompontentPosition("[...]")
```

It should be noted that this doesn't fully resolve the backwards incompatible change, as users may have attempted to do one of the following:
```
-- I do not care about the z axis
local x, y = veh:getComponentPosition("[...]") 
-- old: x, y numbers
-- new: x vector y nil

-- Capture it for storage
local compPos = {veh:getComponentPosition("[...]")
-- old: compPos is a table containing three numbers
-- new: compPos is a table containing a single vector
```
But I suspect that those aren't all that commonly found.

### DIFF
--- a/vendor/lua/src/lua.h
+++ b/vendor/lua/src/lua.h
@@ -220,7 +220,13 @@ LUA_API int   (lua_load) (lua_State *L, lua_Reader reader, void *dt,
                                         const char *chunkname);
 
 LUA_API int (lua_dump) (lua_State *L, lua_Writer writer, void *data);
-
+// MTA specific: Returns the number of expected results in a C call.
+// Note that this will no longer be reliable if 
+// It will also not be reliable in case of incorrectly called functions
+// e.g. 
+//   local a, b = tostring(3) 
+// will return 2, despite tostring only returning one number
+LUA_API int (lua_ncallresult) (lua_State* L);
 
 /*
 ** coroutine functions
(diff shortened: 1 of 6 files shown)


ccw808 (approved):

ghost (approved):

vendor/lua/src/lua.h
```diff
@@ -220,7 +220,13 @@ LUA_API int   (lua_load) (lua_State *L, lua_Reader reader, void *dt,
                                         const char *chunkname);
 
 LUA_API int (lua_dump) (lua_State *L, lua_Writer writer, void *data);
-
+// MTA specific: Returns the number of expected results in a C call.
+// Note that this will no longer be reliable if
```
qaisjp, line 224: if

sbx320: 85f0e03e4f241a9d9b230dc014be6e4810d6cb60 👀

qaisjp: :ok_hand:

Issue #1197, opened 2019-12-29
Raise setCameraShakeLevel limit from 255


guillaume-dorczynski:
Hi I'm guix, you may remember me, years ago I made a patch to add function "shakeCamera" but it was never added into MTA. Instead, someone else added function setCameraShakeLevel.

My version didn't have this limit of 255 ( and I'm almost sure it didn't have this bug #723 ), it was really good for making a realistic speed effect like in some racing games, now with such limit it just looks like the driver is drunk. It was also good for a damage effect like in most FPS games.

Please allow this function to use a much greater value than 255, and fix the bug :)

This version simply set the drunk level, mine was calling function sub_50A970, in my opinion this is the right thing to do.

Lpsd:
Feel free to submit your own PR for camera shake, as you described above (although with that you'd more than likely need to have another function name, for backwards compatibility reasons). 

Alternatively you're more than welcome to modify the existing function and fix the issues mentioned, but the implementation (using GTA's drunk level) would have to stay the same (again, for backwards compat. reasons).

On that note, it would probably make sense for `setCameraShakeLevel` to be deprecated and renamed to `setCameraDrunkLevel` - if you did then decide to implement your alternative "camera shake" there would be no confusion with a function such as `shakeCamera`

PR #1631, opened 2020-08-22
Rename set/getCameraShakeLevel to set/getCameraDrunkLevel 

Unde-R:
In addition to what Lpsd said in #1197 I renamed:
Both ( setCameraShakeLevel & getCameraShakeLevel ) to ( setCameraDrunkLevel & getCameraDrunkLevel )


qaisjp:
There are other things you need to do when renaming functions, like:

- keeping the original name around
- marking the old name as deprecated, so it can be auto upgraded

I'll write some documentation for this and send you a link here.

Unde-R:
Sure thing mate

qaisjp:
Instructions added here: https://github.com/multitheftauto/mtasa-blue/wiki/Lua-API#how-to-rename-functions

Unde-R:
Feel free to request any other changes ^^

Unde-R:
https://github.com/multitheftauto/mtasa-blue/blob/15b2ec72367c1a4226b5afebb7d38b113ca34022/Server/mods/deathmatch/logic/CResourceChecker.Data.h#L211
How I can get the current version so I can add it here?

qaisjp:
This is the number that appears on https://buildinfo.mtasa.com.

The number increments per-commit, but only appears on the website after the next nightly is built.

So, unfortunately this is multi-step. (Which is why everyone forgets about it.)

This one needs to get merged, then you need to wait a day, and then you need to send a PR with the version number.

(But we probably won't merge this before 1.5.8? I don't know.)

Unde-R:
Ah alright ty

ghost (dismissed):

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.h
```diff
@@ -24,7 +24,7 @@ class CLuaCameraDefs : public CLuaDefs
     LUA_DECLARE(GetCameraTarget);
     LUA_DECLARE(GetCameraInterior);
     LUA_DECLARE(GetCameraGoggleEffect);
-    LUA_DECLARE(GetCameraShakeLevel);
+    LUA_DECLARE(GetCameraDrunkLevel);
```
ghost: Consider using the new argument parser for this.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.h
```diff
@@ -37,7 +37,7 @@ class CLuaCameraDefs : public CLuaDefs
     LUA_DECLARE(GetCameraClip);
     LUA_DECLARE(SetCameraViewMode);
     LUA_DECLARE(SetCameraGoggleEffect);
-    LUA_DECLARE(SetCameraShakeLevel);
+    LUA_DECLARE(SetCameraDrunkLevel);
```
ghost: +

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.h
Unde-R: Sure mate, tmrw

StrixG (dismissed):

Pirulax:
I'm not quite sure whats the problem with `shake level`?

Unde-R:
> I'm not quite sure whats the problem with `shake level`?

Well gta named it drunkLevel

Unde-R:
thnx for the commits @StrixG

Pirulax:
Well okay. I'm not quite sure if the old function needs to be deprecated tho.
They both achieve the same thing. Its like `std::string::length()` and `std::string::size()` they achieve the same thing, but have different names for convenience.
Perhaps we could keep both functions?

patrikjuvonen (approved):

PR #1634, opened 2020-08-24
Refactor CConsoleCommands::Debugscript

Pirulax:
In a previous commit ( 8f25c22 ) the `debugscript` command got fixed, but I think it needs a refactor as well, so here we are.
I didnt want to be consistent with the existing style, because I think this is more readable, but correct me if im wrong.

### DIFF
--- a/Client/game_sa/CAEVehicleAudioEntitySA.cpp
+++ b/Client/game_sa/CAEVehicleAudioEntitySA.cpp
@@ -11,6 +11,8 @@
 
 #include "StdInc.h"
 
+//static auto s_aEngineDummySlots = (CAEVehicleAudioEntitySA::DummyEngineSlot*)(VAR_CAEVehicleAudioEntity__s_aDummyEngineSlots);
+
 CAEVehicleAudioEntitySA::CAEVehicleAudioEntitySA(CAEVehicleAudioEntitySAInterface* pInterface)
 {
     m_pInterface = pInterface;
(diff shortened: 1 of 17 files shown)


Pirulax:
> yeah I see, Im meant the || strlen..

Ah well, thats necessary to make sure I actually have a valid char when doing that `atoi` like thing.

Client/game_sa/CAEVehicleAudioEntitySA.cpp
```diff
@@ -53,15 +55,80 @@ void CAEVehicleAudioEntitySA::TurnOnRadioForVehicle()
...
+
+void CAEVehicleAudioEntitySA::CancelVehicleEngineSound(short engineSoundStateID)
+{
+    return ((void(__thiscall*)(void*, short))FUNC_CAEVehicleAudioEntity__CancelVehicleEngineSound)((void*)m_pInterface, engineSoundStateID);
+}
+
+void CAEVehicleAudioEntitySA::MakeSureEngineSoundsAreLoaded()
+{
+    // TODO: Check if this is the right way to do this
+    //       (aka should we write a function to load the sound bank, and all that, or this hack is just fine)
```
ghost, line 86:
```suggestion
    // CAEVehicleAudioEntity::JustGotInVehicleAsDriver loads the engine sound banks
```
If it works and has no side effects, then it's totally fine.

Client/game_sa/CAEVehicleAudioEntitySA.cpp
```diff
@@ -53,15 +55,80 @@ void CAEVehicleAudioEntitySA::TurnOnRadioForVehicle()
...
+    ((void(__thiscall*)(void*))FUNC_CAEVehicleAudioEntity__JustGotInVehicleAsDriver)((void*)m_pInterface);
+}
+
+unsigned int CAEVehicleAudioEntitySA::SetEngineAccelerationSoundBank(unsigned int bankId) // Seems to crash with id 83
+{
+    if (bankId == m_pInterface->m_wEngineOnSoundBankId)
+        return m_pInterface->m_wEngineOnSoundBankId;
+
+    // TODO: Add check for bankid (maybe use CAEAudioHardware or something)
+    auto oldValue = m_pInterface->m_wEngineOnSoundBankId;
```
ghost, line 96:
```suggestion
    short oldValue = m_pInterface->m_wEngineOnSoundBankId;
```
The auto keyword is mostly used to improve readability. This is not good use of auto. I personally only use it when the line takes way too much space and makes the code hard to read.

Client/game_sa/CAEVehicleAudioEntitySA.h
```diff
@@ -122,27 +122,74 @@
 #define FUNC_CAEVehicleAudioEntity__Service                                            0x502280
 
 #define VAR_CAEVehicleAudioEntity__s_pPlayerDriver                                     0xB6B990
+#define VAR_CAEVehicleAudioEntity__s_aDefaultVehicleAudioSettings                      0x860AF0
+#define VAR_CAEVehicleAudioEntity__s_aDummyEngineSlots                                 0xB6B9A0
```
ghost, line 126: There's no reason to define preprocessors when you can just easily write
```
// header file
const int MAX_DUMMY_ENGINE_SLOTS = 10;
class CAEVehicleAudioEntitySA 
{
public:
    static DummyEngineSlot(&s_DummyEngineSlots)[MAX_DUMMY_ENGINE_SLOTS];
};
```
```
// .cpp file
DummyEngineSlot(&CAEVehicleAudioEntitySA::s_DummyEngineSlots)[MAX_DUMMY_ENGINE_SLOTS] = *(DummyEngineSlot(*)[MAX_DUMMY_ENGINE_SLOTS])0xB6B9A0;
```

ghost (commented):
I've pointed out a few things that must be addressed before we can merge. Overall, the PR looks good. We just need to test it.

qaisjp (changes requested):
This PR contains a lot more than the title? I am confusion. I think 3bfb3a87b34fb420b22340340c547dcfc1a6a190 needs to be removed?

Pirulax:
Yeah, i forgot that I have opened a PR for this branch, so I casually pushed my audio entity things here(because I forgot to switch branches)..
Ill revert it. (Well, soon (TM))

Client/game_sa/CAEVehicleAudioEntitySA.h
Pirulax: I know about `constexpr` (and I use it a lot), but I just wanted to be consistent about using macros.
I might as well just copy paste the address directly then.

Pirulax:
Issue: Contains code changes unrelated to PR.

PR #1635, opened 2020-08-24
Add roll state to getPedMoveState while crouching or crawling

Unde-R:
Looks like the movement was declared but no code for It.


ghost (approved):

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -2413,7 +2413,15 @@ eMovementState CClientPed::GetMovementState()
             if (cs.LeftStickX == 0 && cs.LeftStickY == 0)
                 return MOVEMENTSTATE_CROUCH;
             else
+            {
+                // Is he aiming and rolling while crouching?
+                if ((cs.LeftStickX == -128 || cs.LeftStickX == 128) && cs.RightShoulder1)
+                    return MOVEMENTSTATE_ROLL;
```
qaisjp, line 2433: Btw, how did you figure out this if-check?

Unde-R: https://github.com/multitheftauto/mtasa-blue/blob/927509565ab3a8d48df006440654c38e0942aa89/Client/sdk/game/CPad.h#L20
https://github.com/multitheftauto/mtasa-blue/blob/927509565ab3a8d48df006440654c38e0942aa89/Client/sdk/game/CPad.h#L27

ghost: @qaisjp How about this?

```
bool CControllerState::IsMoving()
{
    return cs.LeftStickX == -128 || cs.LeftStickX == 128;
}

bool CControllerState::IsAiming()
{
    return cs.RightShoulder1 != false;
}
```

```
if (cs.IsMoving() && cs.IsAiming())
    return MOVEMENTSTATE_ROLL;
```

qaisjp: That makes me very happy

Unde-R: make us*, thnx saml1er

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -2413,7 +2413,15 @@ eMovementState CClientPed::GetMovementState()
             if (cs.LeftStickX == 0 && cs.LeftStickY == 0)
                 return MOVEMENTSTATE_CROUCH;
             else
```
Goulartee, line 2415: dsadsasadsa

patrikjuvonen (changes requested):
At the moment this shows MOVEMENTSTATE_ROLL is activated when the ped is not actually rolling.

Steps to reproduce:
- Have fists (or any non-roll inducing weapon) equipped
- Hold right mouse and walk right/left, it shows roll while you're crawling

Furthermore: what are potential roll-preventing features in GTA/MTA? At least while you're in vehicle getPedMoveState is not available, but how about when you force an animation while crouched and try to roll? What other potential things can break this? I suggest try to break this check in as many ways as possible, thanks :+1:

A test resource would be good too :) 

Additionally, if you stop pressing any of the keys, the movement state returned is no longer roll, even though the roll animation is still playing.

This doesn't cover all cases it seems.

PR #1639, opened 2020-08-27
Fix #406: improve help command in server console

### LINKED ISSUE #406: Improve help command in server console
The help command in the server is not very useful. You get a list of commands, but using them will get no response unless you use them successfully.

Current `help` shows you a list of commands. I'd like every single server console command to require a piece of help text, shown when you do `help [command]`.

The `help` output (without argument) should also inform you that you can do `help [command]`.

**Additional context**

Replaces https://bugs.mtasa.com/view.php?id=5642

See below screenshot from an old version of MTA:

image

Unde-R:
For #406

### DIFF
--- a/Server/mods/deathmatch/logic/CConsole.cpp
+++ b/Server/mods/deathmatch/logic/CConsole.cpp
@@ -119,10 +119,10 @@ bool CConsole::HandleInput(const char* szCommand, CClient* pClient, CClient* pEc
     return false;
 }
 
-void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, string sConsoleHelp)
+void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, const char* szConsoleHelpText)
 {
     // Make a command class and add it to the list
-    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted, sConsoleHelp);
+    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted, szConsoleHelpText);
     m_Commands.push_back(pCommand);
 }
 
--- a/Server/mods/deathmatch/logic/CConsole.cpp
+++ b/Server/mods/deathmatch/logic/CConsole.cpp
@@ -119,10 +119,10 @@ bool CConsole::HandleInput(const char* szCommand, CClient* pClient, CClient* pEc
     return false;
 }
 
-void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted)
+void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, string sConsoleHelp)
 {
     // Make a command class and add it to the list
-    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted);
+    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted, sConsoleHelp);
     m_Commands.push_back(pCommand);
 }
 
--- a/Server/mods/deathmatch/logic/CConsole.h
+++ b/Server/mods/deathmatch/logic/CConsole.h
@@ -24,7 +24,7 @@ public:
 
     bool HandleInput(const char* szCommand, CClient* pClient, CClient* pEchoClient);
 
-    void             AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted);
+    void             AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, string m_sConsoleHelp);
     void             DeleteCommand(const char* szCommand);
     void             DeleteAllCommands();
     CConsoleCommand* GetCommand(const char* szKey);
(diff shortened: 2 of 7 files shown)


Unde-R:
### Result:
help

Vadya963:
> I need help with definitions if someone can do that

do you need a description of the commands?

Vadya963:
https://wiki.multitheftauto.com/wiki/Server_Commands

Unde-R:
> https://wiki.multitheftauto.com/wiki/Server_Commands

oh thanks

Unde-R:
@StrixG updated! 😃

Server/mods/deathmatch/logic/CConsole.cpp
```diff
@@ -119,10 +119,10 @@ bool CConsole::HandleInput(const char* szCommand, CClient* pClient, CClient* pEc
     return false;
 }
 
-void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted)
+void CConsole::AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, string sConsoleHelp)
 {
     // Make a command class and add it to the list
-    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted);
+    CConsoleCommand* pCommand = new CConsoleCommand(pHandler, szCommand, bRestricted, sConsoleHelp);
     m_Commands.push_back(pCommand);
```
Pirulax, line 126: imho keep using `const char*`, not need for `string` here at all. (if you really want to use one, use `std::move`). Also, wrong hungarian notation, it should be `str` not `s`

Pirulax: imho keep using const char*, not need for string here at all. (if you really want to use one, use std::move). Also, wrong hungarian notation, it should be str not s

Server/mods/deathmatch/logic/CConsole.h
```diff
@@ -24,7 +24,7 @@ class CConsole
 
     bool HandleInput(const char* szCommand, CClient* pClient, CClient* pEchoClient);
 
-    void             AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted);
+    void             AddCommand(FCommandHandler* pHandler, const char* szCommand, bool bRestricted, string sConsoleHelp);
```
Pirulax: Same

Server/mods/deathmatch/logic/CConsoleCommand.cpp
```diff
@@ -20,6 +20,7 @@ CConsoleCommand::CConsoleCommand(FCommandHandler* pHandler, const char* szComman
     m_szCommand = new char[strlen(szCommand) + 1];
     strcpy(m_szCommand, szCommand);
     m_bRestricted = bRestricted;
+    m_sConsoleHelp = sConsoleHelp;
```
Pirulax, line 23: Same.

Server/mods/deathmatch/logic/CConsoleCommands.cpp
```diff
@@ -1278,32 +1278,53 @@ bool CConsoleCommands::DebugScript(CConsole* pConsole, const char* szArguments,
...
-        // Add a new line every third command
-        if (iCount == 3)
+        std::string strHelpText = "Available commands:\n\n";
+        pEchoClient->SendConsole("help [command]");
+
+        // Loop through all added commands
+        int                                    iCount = 0;
+        list<CConsoleCommand*>::const_iterator iter = pConsole->CommandsBegin();
+        for (; iter != pConsole->CommandsEnd(); iter++)
         {
```
Pirulax: Can be:
```cpp
for (auto iter = pConsole->CommandsBegin(); iter != pConsole->CommandsEnd(); iter++)
```
But, my way to do it is to just add a new method to pConsole:
```cpp
const auto& CommandsList() const { return m_commands; }
```
Then just a range based for loop:
```cpp
for (const auto& command : pConsole->CommandsList())
```

Pirulax (changes requested):

Server/mods/deathmatch/logic/CConsoleCommands.cpp
Pirulax: Also, just realized, this: 
```cpp
for (const auto& command : pConsole->CommandsList())
```
could be:
```cpp
for (CConsoleCommand* command : pConsole->CommandsList())
```
A reference to a pointer is really dumb.

Server/mods/deathmatch/logic/CConsole.cpp
Pirulax: no need for it, because its an unnecessary copy. (from sConsoleHelp, into  `CConsoleCommand` ctor, and then into a member variable. Thats actually 3 unnecessary copies. To be honest, i would store it as a `const char*` member variable instead, since these are all static texts)

Unde-R: Ah I see

Server/mods/deathmatch/logic/CConsoleCommands.cpp
Unde-R:
```c++
const auto& CommandsList() { return m_commands; }
```
no need for `const`

Unde-R: works fine anyway, commiting in a sec

Unde-R:
I Think this can't be marked as Draft anymore.

ghost (approved):

Issue #719, opened 2018-11-27
setModelWheelSize to set vehicle model wheel size


Grafu:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I'm replacing car models and the wheel size does not match the original model (wheels stick into the ground).

**Describe the solution you'd like**
I'd love MTA to have a function something like setModelWheelSize (or even setVehicleWheelSize) to set vehicle/model wheel sizes.

**Describe alternatives you've considered**
I do not see any alternatives for this, except matching custom model to original model considering wheel sizes. This is very frustrating and I run out of matching car slots fast. + They are not perfectly matched.

**Additional context**
Attaching googled memory address offsets for vehicle wheel sizes: https://pastebin.com/mt7eLw0Z
According to this source, wheel memory addresses have offset +1112 [Float] from vehicle pointer (ctrl + f Car Wheel Size). Considering this context...

+1108 [DWord] modding data as in garage info
+1112 [Float] Car Wheel Size from vehicle.ide
+1116 [DWord] Time left for car alarm to sound in ms
+1120 [DWord] Pointer to driver

...neighbour values of Car Wheel Size are DWords offset every 4 bytes. But Car Wheel Size is Float instead of DWord I guess it is 2 bytes and there are 2 wheel sizes (front and rear) in that position??

Mantis: https://bugs.multitheftauto.com/view.php?id=7441

Einheit-101:
Original bug report:
https://bugs.mtasa.com/view.php?id=7441

Wolfee-J:
This'd be perfect for the 'Vehicle streamer' system I'm planning.

Dutchman101:
Practical example:
![image](https://user-images.githubusercontent.com/12105539/81773482-9fa5fb00-94e8-11ea-96da-61b489aa2a58.png)

This has been the biggest annoyance for modellers in the past years I guess.. it severely limits their freedom. It's an MTA problem (in GTA, you can easily modify wheel size in vehicles.ide to accomodate your mod) and therefore, a lot of GTA vehicle mods that are turned into server mods will never look like in single-player.

I've seen a lot of modellers, including myself, trying to play around with the wheel dummy position and pivot placement, but it won't help. That is before the person discovers the existence of this bug.

MrSol0:
original suggestion was made in 
2012-12-10 01:08
--
anyone thinks it's time to make this matter urgent? or is it gonna be forgotten about for another 10 years.

[...]

PR #1641, opened 2020-08-28
Add get/setVehicleWheelScale

AlexTMjugador:
These functions, related to issue #719, simply expose the wheel scale size attribute of the `CVehicle` class to clientside Lua scripts.

### Why?
As explained in the related issue above, custom vehicle models may have a different wheel size than the default model they replace. This wouldn't be an issue if GTA: SA somehow calculated ride height, wheel hitbox and possibly other data from the information contained in the model, but it doesn't. Singleplayer vehicle modifications handle differing wheel sizes by modifying the `WheelScale_Front` and `WheelScale_Rear` fields in `data/vehicles.ide`, which tell GTA the size of the front and rear wheel components so that it can calculate things properly, but script developers can't change those.

On the other hand, most replacement vehicle models with differing wheel sizes that I've seen are not too far off from the sizes of the default models. I believe that the most visible effect of this discrepancy is that vehicle mods look "sunken" into the ground in MTA: SA, while in singleplayer they don't. The `setVehicleWheelScale` function that this PR adds, while **it doesn't affect how GTA: SA calculates the mentioned physical parameters for the model**, allows to draw all the wheels of a vehicle taller or shorter, so that these small differences can be compensated and the custom model will look better. It may also add the possibility of enhancing car tunning scripts.

### Why not?
Admittedly, these functions expose relatively unflexible functionality to scripts. Ideally, we should let scripts set the `WheelScale_Front` and `WheelScale_Rear` values independently, or even improve upon that and let them change the size of each wheel somehow, updating the physical parameters of the model too. However, I've heard that doing that is fairly complex, I don't really have the skills to successfully do a deep dive in GTA: SA internals, and maybe these functions are "good enough".

### Isn't this PR already covered by `setVehicleComponentScale`?
It's not! For example, consider this runcode command:
```lua
crun for _, comp in pairs({ "wheel_lf_dummy", "wheel_rf_dummy", "wheel_lb_dummy", "wheel_rb_dummy" }) do localPlayer.vehicle:setComponentScale(comp, 2, 2, 2) end
```
Intuitively, running that while in a car should make the wheels 2x bigger, but it doesn't.

### Why not fix `setVehicleComponentScale` instead?
That would be a good idea, but I don't know why `setVehicleComponentScale` doesn't work with wheels. I guess it may have something to do with the `CVehicle` attribute that this PR exposes, or other GTA stuff.

## Added functions
```lua
float getVehicleWheelScale ( vehicle theVehicle )
```
Returns a decimal number that represents the scale multiplier in the vertical and horizontal axes that is applied for all the wheels of the vehicle. By default, 1.
```lua
bool setVehicleWheelScale ( vehicle theVehicle, float wheelScale )
```
Sets the wheel scale multiplier for all the wheels of the vehicle. A value of 1 draws the wheels with the default scale. A value of 0 is allowed, and it makes wheels disappear.

## Screenshots
mta-screen_2020-08-28_23-25-35
mta-screen_2020-08-28_23-26-29
mta-screen_2020-08-28_23-30-10

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -3664,18 +3664,6 @@ bool CStaticFunctionDefinitions::GetVehicleModelExhaustFumesPosition(unsigned sh
     return false;
 }
 
-bool CStaticFunctionDefinitions::GetVehicleWheelScale(CClientVehicle& Vehicle, float& fWheelScale)
-{
-    fWheelScale = Vehicle.GetWheelScale();
-    return true;
-}
-
-bool CStaticFunctionDefinitions::SetVehicleWheelScale(CClientVehicle& Vehicle, float fWheelScale)
-{
-    Vehicle.SetWheelScale(fWheelScale);
-    return true;
-}
-
 bool CStaticFunctionDefinitions::SetElementCollisionsEnabled(CClientEntity& Entity, bool bEnabled)
 {
     switch (Entity.GetType())
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -3603,6 +3603,23 @@ bool CStaticFunctionDefinitions::SetVehicleWindowOpen(CClientVehicle& Vehicle, u
     return Vehicle.SetWindowOpen(ucWindow, bOpen);
 }
 
+bool CStaticFunctionDefinitions::SetVehicleModelWheelSize(unsigned short usModel, eResizableVehicleWheelGroup eWheelGroup, float fWheelSize)
+{
+    if (CClientVehicleManager::IsValidModel(usModel))
+    {
+        auto pModelInfo = g_pGame->GetModelInfo(usModel);
+        if (pModelInfo)
+        {
+            pModelInfo->SetVehicleWheelSize(eWheelGroup, fWheelSize);
+            // Restream needed to update ride height
+            m_pVehicleManager->RestreamVehicles(usModel);
+
+            return true;
+        }
+    }
+    return false;
+}
+
 bool CStaticFunctionDefinitions::IsVehicleWindowOpen(CClientVehicle& Vehicle, uchar ucWindow)
 {
     return Vehicle.IsWindowOpen(ucWindow);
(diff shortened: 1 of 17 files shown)


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -3664,6 +3664,18 @@ bool CStaticFunctionDefinitions::GetVehicleModelExhaustFumesPosition(unsigned sh
...
+{
+    fWheelScale = Vehicle.GetWheelScale();
+    return true;
+}
+
+bool CStaticFunctionDefinitions::SetVehicleWheelScale(CClientVehicle& Vehicle, float fWheelScale)
+{
+    Vehicle.SetWheelScale(fWheelScale);
+    return true;
+}
```
qaisjp: It looks like these can be removed :) (as well as in the header)

qaisjp:
Mad props to you, @AlexTMjugador — the PR is really high quality and very descriptive! Thank you

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
AlexTMjugador: Sure 👍 

I'll remove them tomorrow, although I may supersede this PR with a better one soon™️.

Einheit-101:
I dont think this is a very good idea, maybe someone can find out why setVehicleComponentScale doesnt work with wheels and otherwise the server owner can still remodel his vehicle wheels, this is not a good fix for the missing vehicle wheel collision issue.

AlexTMjugador:
> I dont think this is a very good idea, maybe someone can find out why setVehicleComponentScale doesnt work with wheels and otherwise the server owner can still remodel his vehicle wheels, this is not a good fix for the missing vehicle wheel collision issue.

I agree with your point, @Einheit-101. This PR, alone, feels like a kinda bad fix for that issue. However, in combination with #1644, it may be better. What do you think?

ghost:
setVehicleComponentScale was designed to only modify the wheel geometry. The reason why it doesn't affect the collision is that there's no code to modify the collision. We shouldn't touch setVehicleComponentScale because it takes 3 parameters for scale, and that's absurd. We're dealing with wheels here, it should only take 1 parameter, but it's too late to fix the function.

ghost (approved):

Dutchman101:
@AlexTMjugador Please add the below wiki pages:

[https://wiki.multitheftauto.com/wiki/getVehicleWheelScale](https://wiki.multitheftauto.com/wiki/getVehicleWheelScale)
[https://wiki.multitheftauto.com/wiki/setVehicleWheelScale](https://wiki.multitheftauto.com/wiki/setVehicleWheelScale)

AlexTMjugador:
> @AlexTMjugador Please add the below wiki pages:
> 
> https://wiki.multitheftauto.com/wiki/getVehicleWheelScale
> https://wiki.multitheftauto.com/wiki/setVehicleWheelScale

Added, thanks for the reminder!

PR #1644, opened 2020-08-29
Add get/setVehicleModelWheelSize

### LINKED ISSUE #719: setModelWheelSize to set vehicle model wheel size

AlexTMjugador:
These functions, related to issue #719, allow getting and setting the front and rear wheel size of a vehicle model that GTA: SA reads from the `vehicle.ide` file in clientside Lua scripts. Moreover, in combination with #1641, all the wheels can then be visually scaled per vehicle.

### Why?
#1641 provides functions that allow customizing the draw scale of all the wheels of a vehicle, which should be enough for compensating typical small differences in sizes and allowing some creative usages.

However, this PR provides another solution, which also has a feature that is desired by some people I talked to on Discord: it influences the related physical parameters of the model. These parameters are the ride height (which is also known as "height above ground", and affects handling) and wheel collision (used to check if bullets hit a tyre to burst it, for example). This approach not only allows to compensate for bigger differences, which relieves 3D artists of having to stick to a ride height that is similar to the default one, but also allows for even more model customization. For instance, race servers may use these functions to give a different feel to a same car model for each race.

As the wheel size parameters do not change the visual scale of the wheels, unless they are different for each axle (in which case the wheels are drawn with a different scale for each axle and rotated), the core idea of #1641 is still relevant for further per-vehicle, all-wheel customization, that takes the model wheel scale influenced by these functions as a basis.

### Potential for improvement
I think that this patchset could be improved with the following ideas:

- Expand the vehicle model loading functions to calculate these values automagically (but the ability to set them manually doesn't need to be removed; there might be some use cases based on deliberately setting different values).
- Allow each vehicle to have different wheel sizes, even if they share the same model.
- Make the wheel size different for each wheel, instead of for each axle.

Nevertheless, those ideas are rather complex to implement. I believe that this PR (and maybe its "companion" #1641) offers substantial functionality for server owners that, perhaps most importantly, shouldn't impose a notorious mantainance or backward compability burden, should these improvements be made.

## Acknowledgements
I'd like to thank @saml1er for sharing over Discord the results of his previous experiments with changing these parameters, which included proof of concept code and some helpful pointers. Without his help this PR probably wouldn't exist.

## Functions added
```lua
float/table getVehicleModelWheelSize ( int vehicleModel[, string wheelGroup ] )
```
Returns a decimal number that represents the size of the wheels in the specified wheel group of a vehicle model ID. The possible wheel groups are "front_axle" (for the wheels in the front axle), "rear_axle" (for the wheels in the rear axle) and "all_wheels" (for all the wheel groups). These sizes affect the ride height and wheel collision, but they also may affect visual wheel scale and rotation if they differ. The default values for a model are read by GTA: SA from the `vehicles.ide` file, and must be greater than 0. If the wheel group is "all_wheels" or not specified, a table like `{ ["front_axle"] = 0.7, ["rear_axle"] = 0.8 }` is returned, with the values for each wheel group.
```lua
bool setVehicleModelWheelSize ( int vehicleModel, string wheelGroup, float wheelSize )
```
Sets the wheel size multiplier for the wheels in a wheel group of a vehicle model ID.

## Screenshots
Custom vehicle model, before and after:
mta-screen_2020-08-29_19-00-21
mta-screen_2020-08-29_19-00-38
A funky quad:
mta-screen_2020-08-29_19-13-52

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -83,8 +83,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"getVehicleComponents", GetVehicleComponents},
         {"getVehicleModelExhaustFumesPosition", GetVehicleModelExhaustFumesPosition},
         {"getVehicleModelDummyPosition", GetVehicleModelDummyPosition},
-        {"getVehicleModelFrontWheelSize", ArgumentParser<GetVehicleModelFrontWheelSize>},
-        {"getVehicleModelRearWheelSize", ArgumentParser<GetVehicleModelRearWheelSize>},
+        {"getVehicleModelWheelSize", ArgumentParser<GetVehicleModelWheelSize>},
 
         // Vehicle set funcs
         {"createVehicle", CreateVehicle},
@@ -141,8 +140,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"setVehicleWindowOpen", SetVehicleWindowOpen},
         {"setVehicleModelExhaustFumesPosition", SetVehicleModelExhaustFumesPosition},
         {"setVehicleModelDummyPosition", SetVehicleModelDummyPosition },
-        {"setVehicleModelFrontWheelSize", ArgumentParser<SetVehicleModelFrontWheelSize>},
-        {"setVehicleModelRearWheelSize", ArgumentParser<SetVehicleModelRearWheelSize>},
+        {"setVehicleModelWheelSize", ArgumentParser<SetVehicleModelWheelSize>},
     };
 
     // Add functions
@@ -225,8 +223,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getUpgradeOnSlot", "getVehicleUpgradeOnSlot");
     lua_classfunction(luaVM, "getModelExhaustFumesPosition", OOP_GetVehicleModelExhaustFumesPosition);
     lua_classfunction(luaVM, "getVehicleModelDummyPosition", OOP_GetVehicleModelDummyPosition);
-    lua_classfunction(luaVM, "getModelFrontWheelSize", "getVehicleModelFrontWheelSize");
-    lua_classfunction(luaVM, "getModelRearWheelSize", "getVehicleModelRearWheelSize");
+    lua_classfunction(luaVM, "getModelWheelSize", "getVehicleModelWheelSize");
 
     lua_classfunction(luaVM, "setComponentVisible", "setVehicleComponentVisible");
     lua_classfunction(luaVM, "setSirensOn", "setVehicleSirensOn");
@@ -270,8 +267,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "setGravity", "setVehicleGravity");
     lua_classfunction(luaVM, "setModelExhaustFumesPosition", "setVehicleModelExhaustFumesPosition");
     lua_classfunction(luaVM, "setVehicleModelDummyPosition", "setVehicleModelDummyPosition");
-    lua_classfunction(luaVM, "setModelFrontWheelSize", "setVehicleModelFrontWheelSize");
-    lua_classfunction(luaVM, "setModelRearWheelSize", "setVehicleModelRearWheelSize");
+    lua_classfunction(luaVM, "setModelWheelSize", "setVehicleModelWheelSize");
 
     lua_classfunction(luaVM, "resetComponentPosition", "resetVehicleComponentPosition");
     lua_classfunction(luaVM, "resetComponentRotation", "resetVehicleComponentRotation");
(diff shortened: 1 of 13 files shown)


ghost:
Get rid of the current Lua functions in the PR and make it `setVehicleModelWheelSize` and `getVehicleModelWheelSize`, then we can merge the PR.

StrixG:
I propose this syntax:
```lua
bool setVehicleModelWheelSize(int vehicleModel, [string wheels="all"], float size)
float getVehicleModelWheelSize(int vehicleModel, string wheels)
-- Examples
setVehicleModelWheelSize(411, "front", 0.7)
setVehicleModelWheelSize(411, "rear", 2.0)
setVehicleModelWheelSize(451, "all", 0.8)
setVehicleModelWheelSize(451, 0.8) -- will change the size for all wheels

getVehicleModelWheelSize(411, "rear") -- 2.0
getVehicleModelWheelSize(411, "all") -- maybe return table like {front=0.7, rear=2.0}
```

LosFaul:
> 
> 
> I propose this syntax:
> 
> ```lua
> bool setVehicleModelWheelSize(int vehicleModel, float size, [string wheels="all"])
> float getVehicleModelWheelSize(int vehicleModel, string wheels)
> -- Examples
> setVehicleModelWheelSize(411, 0.7, "front")
> setVehicleModelWheelSize(411, 2.0, "rear")
> setVehicleModelWheelSize(451, 0.8, "all")
> setVehicleModelWheelSize(451, 0.8) -- will change the size for all wheels
> 
> getVehicleModelWheelSize(411, "rear") -- 2.0
> getVehicleModelWheelSize(411, "all") -- maybe return table like {front=0.7, rear=2.0}
> ```

first string then value
please keep it in a 'key-value' consistent way
and not a 'value-key' way

```lua
setVehicleModelWheelSize(411, "front/rear/all", 0.7)
```

AlexTMjugador:
I've just streamlined the API according to the suggestions I received, thanks!

It took me a bit more than I expected due to some new Lua function argument parser magic, but in the end I think that all works fine. On a side note, in the `setVehicleModelFrontWheelSize ( int vehicleModel[, string wheelGroup ], float wheelSize )` function, the `wheelGroup` parameter is not that optional, because if it is omitted the argument parser will complain of an unexpected type. It works if it is set to `null`, however. Anyway, I believe that I declared my functions correctly, and therefore fixing that issue is a matter for another PR.

qaisjp:
Having `wheelGroup` be optional will cause ambiguity in the future, if the syntax is extended in the future to support wheel IDs.

Client/sdk/game/CModelInfo.h
```diff
@@ -154,6 +154,9 @@ class CModelInfo
     virtual CVector      GetVehicleDummyPosition(eVehicleDummies eDummy) = 0;
     virtual void         SetVehicleDummyPosition(eVehicleDummies eDummy, const CVector& vecPosition) = 0;
     virtual void         ResetVehicleDummies(bool bRemoveFromDummiesMap) = 0;
+    virtual float        GetVehicleWheelSize(eResizableVehicleWheelGroup eWheelGroup) = 0;
+    virtual void         SetVehicleWheelSize(eResizableVehicleWheelGroup eWheelGroup, float fWheelSize) = 0;
+    virtual void         ResetVehicleWheelSizes(std::pair<float, float>* defaultSizes = NULL) = 0;
```
qaisjp: Should this be `nullptr` instead?

AlexTMjugador: Yeah. I just let the old me decide to use that NULL.

AlexTMjugador:
> Having `wheelGroup` be optional will cause ambiguity in the future, if the syntax is extended in the future to support wheel IDs.

I get your point. I think that better forwards compatibility is worth the extra verbosity on the scripts, and scripters would probably need to think if they really want to change all the wheel sizes at once anyway. I'll change that now.

AlexTMjugador:
The CI checks have passed, and I've addressed all the suggestions. I think that this PR is ready for further reviews, or even merging.

ghost (approved):
@AlexTMjugador The PR looks fine to me. I'd like to merge both wheels PR at once. If you're done working on the PR, then let me know and we'll merge it.

AlexTMjugador:
All good for me, I don't see anything I want to change 👍

StrixG (approved):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4052,3 +4056,42 @@ int CLuaVehicleDefs::OOP_GetVehicleModelExhaustFumesPosition(lua_State* luaVM)
...
+    }
+}
+
+bool CLuaVehicleDefs::SetVehicleModelWheelSize(const unsigned short usModel, const eResizableVehicleWheelGroup eWheelGroup, const float fWheelSize)
+{
+    if (fWheelSize <= 0)
+        throw std::invalid_argument("Invalid wheel size");
+
+    return CStaticFunctionDefinitions::SetVehicleModelWheelSize(usModel, eWheelGroup, fWheelSize);
+}
```
qaisjp, line 4097: There is no need to use CStaticFunctionDefinitions in the new parser.

AlexTMjugador: Okay! But where do we put `CStaticFunctionDefinitions::GetVehicleModelWheelSize`, though? It's actually used several times in `CLuaVehicleDefs::GetVehicleModelWheelSize`. Or should I get rid of that function too?

ghost: You can delete `CStaticFunctionDefinitions::GetVehicleModelWheelSize` and refactor the code, like this: https://pastebin.com/2J6qzHUM

Dutchman101:
@AlexTMjugador Please add the below wiki pages:

[https://wiki.multitheftauto.com/wiki/getVehicleModelWheelSize](https://wiki.multitheftauto.com/wiki/getVehicleModelWheelSize)
[https://wiki.multitheftauto.com/wiki/setVehicleModelWheelSize](https://wiki.multitheftauto.com/wiki/setVehicleModelWheelSize)

AlexTMjugador:
> @AlexTMjugador Please add the below wiki pages:
> 
> https://wiki.multitheftauto.com/wiki/getVehicleModelWheelSize
> https://wiki.multitheftauto.com/wiki/setVehicleModelWheelSize

Added, thanks for the reminder!

guillaume-dorczynski:
I had made something similar, able to set wheel scale and width for all vehicles including monster truck, dumper and bikes, for each vehicle IDs (not just models), and for each wheel so it was possible to set a wheel scale to near 0 and it was like the wheel was missing, and automatically scaling the width of the tire skid marks and the amount of smoke, it also allowed to change camber and toe angles, position offsets, also changed the rotation speed of the wheels to look more realistic... It was not ready for a pull request but almost, then I lost all my work and unable to recover it :( ... Good memories anyway, that was the first time I wrote in ASM :)

AlexTMjugador:
These functions exposed string wheel groups to the Lua API with the intention of being forward compatible with changes like adding per-wheel size, so feel free to improve them :)

PR #1648, opened 2020-08-31
Followup to #1644: Add get/setVehicleModelWheelSize

AlexTMjugador:
This PR addresses the review @qaisjp made on PR #1644 after it got merged. I hope it's better now 🙂

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -4077,38 +4077,41 @@ bool CLuaVehicleDefs::SetVehicleWheelScale(CClientVehicle* const pVehicle, const
 std::variant<float, std::unordered_map<std::string, float>> CLuaVehicleDefs::GetVehicleModelWheelSize(
     const unsigned short usModel, const std::optional<eResizableVehicleWheelGroup> eWheelGroup)
 {
-    eResizableVehicleWheelGroup eActualWheelGroup = eWheelGroup.value_or(eResizableVehicleWheelGroup::ALL_WHEELS);
+    CModelInfo* pModelInfo = nullptr;
+    if (CClientVehicleManager::IsValidModel(usModel))
+        pModelInfo = g_pGame->GetModelInfo(usModel);
 
+    if (!pModelInfo)
+        throw std::invalid_argument("Invalid model ID");
+
+    eResizableVehicleWheelGroup eActualWheelGroup = eWheelGroup.value_or(eResizableVehicleWheelGroup::ALL_WHEELS);
     if (eActualWheelGroup == eResizableVehicleWheelGroup::ALL_WHEELS)
     {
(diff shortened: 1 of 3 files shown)


ghost (approved):

StrixG (approved):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4077,38 +4077,41 @@ bool CLuaVehicleDefs::SetVehicleWheelScale(CClientVehicle* const pVehicle, const
...
-        float fRearWheelSize;
-        if (!CStaticFunctionDefinitions::GetVehicleModelWheelSize(usModel, eResizableVehicleWheelGroup::REAR_AXLE, fRearWheelSize))
-            throw std::invalid_argument("Invalid model ID");
-
         // Return a table like { ["front_axle"] = 0.7, ["rear_axle"] = 0.8 }
         return std::unordered_map<std::string, float>{
-            {"front_axle", fFrontWheelSize},
-            {"rear_axle", fRearWheelSize},
+            {"front_axle", pModelInfo->GetVehicleWheelSize(eResizableVehicleWheelGroup::FRONT_AXLE)},
+            {"rear_axle", pModelInfo->GetVehicleWheelSize(eResizableVehicleWheelGroup::REAR_AXLE)},
```
qaisjp, line 4093: these strings `front_axle` and `rear_axle` are defined here https://github.com/multitheftauto/mtasa-blue/blob/c3dc499b923f7540fbbeed0f8ffadbcb837c9ca0/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp#L584-L588 so you should probably use the enum and convert that to a string, instead of hardcoding the string here.

AlexTMjugador: Unfortunately, returning a `std::unordered_map<eResizableVehicleWheelGroup, float>` instead of a `std::unordered_map<std::string, float>` doesn't compile with a C2665 error, about missing `lua::Push` overloads in `LuaBasic.h`, at line 138, so without changes to the new argument parser, which I think that are kind out of scope for this PR, I was forced to use a string instead.

However, I think that using a string directly may be good if for some reason we want to modify the input values of that function without altering the keys of the returned table, so it's not that bad either. What do you think?

qaisjp: I was thinking we keep the return type as `<string, float>` and manually convert from enum to string. 

But I think having `std::unordered_map<eResizableVehicleWheelGroup, float>` would be better. We should create an issue for this, for the future. I think we can keep your code as it is right now.

> However, I think that using a string directly may be good if for some reason we want to modify the input values of that function without altering the keys of the returned table, so it's not that bad either. What do you think?

I suppose that's true but I think _changing_ strings in this way would only happen in very exceptional circumstances.

PR #1650, opened 2020-09-04
Tweak wheel scaling behavior to be more consistent

AlexTMjugador:
#1641, which introduced `setVehicleWheelScale`, doesn't account for GTA: SA changing the wheel scale as a result of wheel upgrades being added or removed, and using different default scales for vehicles with upgraded wheels. Although this does not pose a serious issue, I have just found that it leads to inconsistent behavior, like the one that is shown in the following flow of events:

> - Wheel upgrade is removed from streamed in vehicle
> - Wheel scale is reset by GTA to the default, which is returned by getVehicleWheelScale
> - Vehicle is streamed out and back in
> - The wheel scale is set back to the last value set by a script or 1 (not the same scale that it was streamed out with)

This PR tweaks that behavior by adding logic to synchronize the default wheel scale calculated by GTA with the one present in `CClientVehicle` attributes, for which functionality added in #1644 is used. This way, the previous event flow, for example, becomes:

> - Wheel upgrade is removed from streamed in vehicle
> - Wheel scale is reset by GTA
> - Vehicle is streamed out and back in
> - The wheel scale set by GTA is preserved

Also, as expected, the scale set by a script is now always honored, until it is reset by GTA during a script-initiated wheel upgrade installation or removal.

All in all, I believe that the new behavior is more consistent with the expectations that script developers may have, and as the function was introduced only days ago, it's not too late to fix this. Moreover, I think that GTA resetting the wheel size makes sense because, as upgraded wheel models are reused accross different vehicles, they need different scales to accomodate each vehicle, so it should be changed anyway. Finally, as a side bonus, the new behavior works better for streamed out vehicles.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicle.cpp
+++ b/Client/mods/deathmatch/logic/CClientVehicle.cpp
@@ -171,6 +171,9 @@ CClientVehicle::CClientVehicle(CClientManager* pManager, ElementID ID, unsigned 
 
     // We've not yet been streamed in
     m_bJustStreamedIn = false;
+
+    // We've not changed the wheel scale
+    m_bWheelScaleChanged = false;
 }
 
 CClientVehicle::~CClientVehicle()
@@ -2682,7 +2685,6 @@ void CClientVehicle::Create()
         m_pVehicle->SetGravity(&m_vecGravity);
         m_pVehicle->SetHeadLightColor(m_HeadLightColor);
         m_pVehicle->SetRadioStatus(0);
-        m_pVehicle->SetWheelScale(m_fWheelScale);
 
         if (IsNitroInstalled())
         {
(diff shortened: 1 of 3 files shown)


qaisjp:
> as the function was introduced only days ago, it's not too late to fix this.

Don't worry! I _think_ stuff that we introduce into master can be changed without worry for BC until it is actually released in a numbered version.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2825,10 +2827,21 @@ void CClientVehicle::Create()
             if (m_bHasCustomHandling)
                 ApplyHandling();
         }
+
         // Re-add all the upgrades - Has to be applied after handling *shrugs*
+        bool bPreviousWheelScaleChanged = m_bWheelScaleChanged;
+        float fPreviousWheelScale = m_fWheelScale;
```
qaisjp:
```suggestion
        // Applying upgrades can change these values,
        // so we should keep track of the original values.
        bool bPreviousWheelScaleChanged = m_bWheelScaleChanged;
        float fPreviousWheelScale = m_fWheelScale;

        // Re-add all the upgrades - Has to be applied after handling *shrugs*
```

qaisjp (approved):
I've left a nit, where primarily `// Re-add all the upgrades - Has to be applied after handling *shrugs*` didn't make sense above these two assignments.

This LGTM. This sounds super reasonable, thank you very much for the explanation, and thank you for continuously testing these functions 🎉

ghost (approved):

Client/mods/deathmatch/logic/CClientVehicle.cpp
AlexTMjugador: Thank you! I've just commited a variant of your suggested change, changing the comments a bit to make them more clear and concise.

PR #1651, opened 2020-09-05
Fix #995: add onResourceLoadStateChange event

### LINKED ISSUE #995: onResourceChange

Always when you do refresh or refreshall on server, you have always have to do additional user interaction in your managment resources to update all the changed resources. I would like an event that gets called when the refresh command has done it's job and gives a table of all changes resources.

add an event "onResourceChange" that gives a table of all resources that changed
{ {resName1, changeInfo}, {resName2, changeInfo}, ... }
changeInfo should be a string with an information such as:
"loaded", "unloaded", "meta changes", "file changes", "error loading", ...

you always have to loop through all resources, which is annoying and can block the main thread when there are thousands of resources

TeteX1:
As requested in issue #995, this PR introduces a new event in MTA. The event will enable users to know when a resource loading state has changed.

According to this [discord conversation](https://discordapp.com/channels/278474088903606273/366384007535001612/751333369941327882), the initial intention was to add 3 events; onResourceLoad, onResourceUnload and onResourceChange. However, @sbx320 has suggested a single event with the name 'onResourceStateChange' which will serve the purpose of all the three and it was agreed upon and thus, I started working on it. I tested the event and it works well as will be evidenced later in this PR.

EDIT: After a conversation between me, @qaisjp and @Lpsd on Discord, we've established that it would be better to change the event name from 'onResourceStateChange' to 'onResourceLoadStateChange', in addition to a few other changes.

# onResourceLoadStateChange
### Parameters
`[resource] changedResource, [string] oldState, [string] newState`

- **changedResource**: the resource that was either loaded, reloaded or is unloading.
- **oldState**: the state the resource was in before it changed.
- **newState**: the state the resource has changed to.

### Source
The source of this event is the root element.

### Fixed state combinations
The event will either change from:
```
nil -> loaded (new resource)
loaded -> nil (resource is about to unload)
loaded -> loaded (the resource has just reloaded)
```

P.S. The reason that there is an "unloading" state rather than "unloaded" when a resource unloads is that when the server decides to remove a resource, it will remove it's object and everything related to it internally, disallowing the user to access the `changedResource` parameter in the handler. Therefore I triggered the event right before the resource unloads so that the user could access the resource before it gets completely unloaded.

EDIT: 'unloaded' + 'unloading' was changed to nil and 'reloaded' was removed.

# Testing

1. I created a test resource with the name 'newevent' and included the following code in it:
```Lua
addEventHandler("onResourceLoadStateChange", root,
    function (resource, oldState, newState)
        print("resource '" .. getResourceName(resource) .. "' has changed from '" .. tostring(oldState) .. "' to '" .. tostring(newState) .. "'")
    end
)
```
2. I created a dummy resource and called it 'test_'.
3. I started the 'newevent' resource.
4. I changed the resource name from 'test_' to 'test'.
5. I typed in 'refresh' in the console.
6. I changed the content of the 'test' resource.
7. I typed in 'refreshall' in the console.

ok

### Downside
I consider the sole downside of such event is that it will only be triggered if the server refreshed the resources or the user force-refreshed the resources by either using a console command or a script that uses the `refreshResources` function.


Lpsd:
State names should be consistent with [getResourceState](https://wiki.multitheftauto.com/wiki/GetResourceState)

It may also be worth looking into `CResource::GetState()` and seeing how/when the vars included there are set (may be better to trigger this event under the same circumstances).

qaisjp:
I also have some uncertainty about introducing this new pseudo-enum ("unloaded", "loaded", "reloaded", "unloading").

LosFaul:
"reloaded" state seems unnecessary
since you can assmue if newState is "loaded"
that the resource has been reloaded or added

qaisjp:
Even though we chatted about it last night, I still don't understand why we only want load state change and not all state change.

If you're writing a resource manager, won't the resource manager need to know if the resource has stopped or started?

LosFaul:
> 
> 
> Even though we chatted about it last night, I still don't understand why we only want load state change and not all state change.
> 
> If you're writing a resource manager, won't the resource manager need to know if the resource has stopped or started?

for that there already exists
https://wiki.multitheftauto.com/wiki/OnResourcePreStart
https://wiki.multitheftauto.com/wiki/OnResourceStart
https://wiki.multitheftauto.com/wiki/OnResourceStart

and that could be kinda redundant with these events

only things missing were events that tell you if a resource was loaded/unloaded/failed to load

qaisjp:
hm

patrikjuvonen (approved):
Thanks for the pull request. Looks and works good.

patrikjuvonen:
@TeteX1 Please create a new Wiki page for this new event.

ghost:
@patrikjuvonen
https://wiki.multitheftauto.com/wiki/OnResourceLoadStateChange

~~I hope it should be fine, but i don't know how to add it [here.](https://wiki.multitheftauto.com/wiki/Server_Scripting_Events)~~

Done, thanks to @theSarrum

PR #1653, opened 2020-09-06
Fix crash in CModelInfoSA::GetVehicleWheelSize if model info is not l…

sbx320:
This fixes a crash introduced with the recent addition of wheel scale functionality. Since `CClientVehicle::ResetWheelScale` may be called even if the vehicle is not streamed in, the model info for the vehicle may not be loaded. As a result, accessing the model info may fail, when trying to call `CClientVehicle::ResetWheelScale`, causing a crash. 

To fix, we make `CModelInfo::GetVehicleWheelSize` load the model if required. I also applied this to `SetVehicleWheelSize` as I suspect we might run into similar issues.

--- 

Generally the way we handle `CModelInfo` seems a bit confusing and unreliable. Might be better to wrap it to request models automatically when attempting to use them via a RAII wrapper.

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1020,6 +1020,10 @@ float CModelInfoSA::GetVehicleWheelSize(eResizableVehicleWheelGroup eWheelGroup)
     if (!IsVehicle())
         return 0.0f;
 
+    // Request model load right now if not loaded yet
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
FWIW, the changes LGTM.

qaisjp (approved):

ghost (approved):

PR #601, opened 2018-09-27
Fix nametags,waters,sounds are interiorless, added setAllInteriors function

FileEX:
Fixes #596 - **Needs testing**

**Bug description**
Nametags are visible to players in other interiors. 

**To reproduce nametags bug**
1. player 1 joins the game
2. sp 0 0 2
3. run me.interior = 1
4. run setCameraInterior(me, 1) (just in case)
5. player 2 joins the game
6. sp 1 1 2
7. observe nametag

Now the interior of the players is also checked. But needs testing.

This PR fix waters interiorless and sounds interiorless and nametags. This PR added 'setAllInteriors' function.

PR #1661, opened 2020-09-10
Fix #322: onClientMarkerHit(Leave) ignores element's interior

### LINKED ISSUE #322: onClientMarkerHit ignores element's interior
I have a client-side marker, placed into a non-zero interior, but onClientMarkerHit would still trigger if element in interior 0 enters the marker. I assume onClientColShapeHit has the same issue, though it's more evident with markers since they ignore hit element's height on the client side.

Unde-R:
According to #322 Looks like [onClientMarkerHit](https://wiki.multitheftauto.com/wiki/OnClientMarkerHit) doesn't care about element's interior and when I checked MTA's code looks like [onClientMarkerLeave](https://wiki.multitheftauto.com/wiki/OnClientMarkerLeave) does the same thing, and for [onMakerHit](https://wiki.multitheftauto.com/wiki/OnMarkerHit) and [onMakerLeave](https://wiki.multitheftauto.com/wiki/OnMarkerLeave) they do check for interior.
https://github.com/multitheftauto/mtasa-blue/blob/3b95ff48969474c691a8a4f2276f581377111f06/Server/mods/deathmatch/logic/CMarker.cpp#L315

### DIFF
--- a/Client/mods/deathmatch/logic/CClientMarker.cpp
+++ b/Client/mods/deathmatch/logic/CClientMarker.cpp
@@ -399,7 +399,7 @@ void CClientMarker::StreamOut()
 
 void CClientMarker::Callback_OnCollision(CClientColShape& Shape, CClientEntity& Entity)
 {
-    if (IS_PLAYER(&Entity))
+    if (IS_PLAYER(&Entity) && GetInterior() == Entity.GetInterior())            // Matching interior?
     {
         // Call the marker hit event
         CLuaArguments Arguments;
@@ -411,7 +411,7 @@ void CClientMarker::Callback_OnCollision(CClientColShape& Shape, CClientEntity& 
 
 void CClientMarker::Callback_OnLeave(CClientColShape& Shape, CClientEntity& Entity)
 {
-    if (IS_PLAYER(&Entity))
+    if (IS_PLAYER(&Entity) && GetInterior() == Entity.GetInterior())            // Matching interior?
     {
         // Call the marker hit event
         CLuaArguments Arguments;


patrikjuvonen:
I believe an extra event parameter must be added for backwards compatibility reason, but this is debatable.

Unde-R:
in discord, qaisjp said that we can fix it but without adding new parameter

patrikjuvonen:
> in discord, qaisjp said that we can fix it but without adding new parameter

@qaisjp What's your idea on backwards compatibility or is this a screw-it issue?

qaisjp:
screw it situation (but maybe we should do this after 1.5.8 so we can watch the effects for longer?)

qaisjp (changes requested):
This works fine and is technically safe for merging... but I think, reduce confusion, we should only merge this once we've fixed interiors appearing in all dimensions (#601 needs to be reimplemented in a separate PR).

Lpsd:
Why have we decided to randomly break backwards compatibility here when a reasonable, alternative solution already exists?

I think we should just add `matchingInterior` parameter to the client-side version of these events, as initially suggested.

patrikjuvonen:
Yeah, I think a parameter is good for now.

Client/mods/deathmatch/logic/CClientMarker.cpp
```diff
@@ -399,7 +399,7 @@ void CClientMarker::StreamOut()
 
 void CClientMarker::Callback_OnCollision(CClientColShape& Shape, CClientEntity& Entity)
 {
-    if (IS_PLAYER(&Entity))
+    if (IS_PLAYER(&Entity) && GetInterior() == Entity.GetInterior())            // Matching interior?
```
patrikjuvonen, line 402: For now, let's instead add the matching interior as an extra parameter to onClientMarkerHit call, just like matching dimension.

Client/mods/deathmatch/logic/CClientMarker.cpp
```diff
@@ -411,7 +411,7 @@ void CClientMarker::Callback_OnCollision(CClientColShape& Shape, CClientEntity&
 
 void CClientMarker::Callback_OnLeave(CClientColShape& Shape, CClientEntity& Entity)
 {
-    if (IS_PLAYER(&Entity))
+    if (IS_PLAYER(&Entity) && GetInterior() == Entity.GetInterior())            // Matching interior?
```
patrikjuvonen, line 414: For now, let's instead add the matching interior as an extra parameter to onClientMarkerHit call, just like matching dimension.

patrikjuvonen (changes requested):

PR #1663, opened 2020-09-12
Fix stack overflow when attaching elements in maps

tederis:
When storing elements in maps sometimes we are attaching elements. In that case cross referencing may lead to 
an endless cycle and finally to a stack overflow error. The PR is intended to fix it.

The resource for crashing your own server:
[test.zip](https://github.com/multitheftauto/mtasa-blue/files/5212644/test.zip)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientEntity.cpp
+++ b/Client/mods/deathmatch/logic/CClientEntity.cpp
@@ -1070,6 +1070,29 @@ bool CClientEntity::IsEntityAttached(CClientEntity* pEntity)
     return ListContains(m_AttachedEntities, pEntity);
 }
 
+bool CClientEntity::IsAttachedToElement(CClientEntity* pEntity, bool bRecursive)
+{
+    if (bRecursive)
+    {
+        std::set<CClientEntity*> history;
+
+        for (CClientEntity* pCurrent = this; pCurrent; pCurrent = pCurrent->GetAttachedTo())
+        {
+            if (pCurrent == pEntity)
+                return true;
+
+            if (MapContains(history, pCurrent))
+                break; // This should not be possible, but you never know
+
+            MapInsert(history, pCurrent);
+        }
+
+        return false;
+    }
+
+    return m_pAttachedToEntity == pEntity;
+}
+
 void CClientEntity::ReattachEntities()
 {
     // Jax: this should be called on streamIn/creation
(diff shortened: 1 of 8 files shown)


Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -1070,6 +1070,29 @@ bool CClientEntity::IsEntityAttached(CClientEntity* pEntity)
...
+
+        for (CClientEntity* pCurrent = this; pCurrent; pCurrent = pCurrent->GetAttachedTo())
+        {
+            if (pCurrent == pEntity)
+                return true;
+
+            if (MapContains(history, pCurrent))
+                break; // This should not be possible, but you never know
+
+            MapInsert(history, pCurrent);
```
Pirulax: Why not just use the return value of set::insert? It returns false if theres already a matching element in the set. Much faster than checking, then inserting.
Like this: 
```cpp
if (!std::get<bool>(history.insert(pCurrent)); 
    break; // This should not be possible, but you never know
```

Server/mods/deathmatch/logic/CElement.cpp
```diff
@@ -1126,6 +1126,29 @@ bool CElement::IsElementAttached(CElement* pElement)
...
+
+        for (CElement* pCurrent = this; pCurrent; pCurrent = pCurrent->GetAttachedToElement())
+        {
+            if (pCurrent == pElement)
+                return true;
+
+            if (MapContains(history, pCurrent))
+                break; // This should not be possible, but you never know
+
+            MapInsert(history, pCurrent);
```
Pirulax: Same suggestion here as before.

Pirulax (changes requested):

Client/mods/deathmatch/logic/CClientEntity.cpp
tederis: It is just a copy-paste from the old code

Pirulax: Yeah, I though, but its probably like a decade old, now there's a better way to do it :D

patrikjuvonen (approved):
Thank you for the pull request. Seems good to me.

ghost (approved):

qaisjp:
My instinct is that we should (also|only) deal with this in the map parsing/loading process.

(I have only skimmed the code, so I don't know if this change affects performance, and therefore I don't know if the problem should ONLY be dealt with when loading maps.)

tederis:
> My instinct is that we should (also|only) deal with this in the map parsing/loading process.
> 
> (I have only skimmed the code, so I don't know if this change affects performance, and therefore I don't know if the problem should ONLY be dealt with when loading maps.)

But it's already working that way. The processes are occuring in function LinkupElements() which in turn invoked inside a map loading process.

qaisjp:
Great!

StrixG (approved):

PR #1664, opened 2020-09-13
Add return value to MapInsert

ccw808:
`MapInsert` will now return true if the insertion took place. Can be used to remove use of `MapContains` before `MapInsert`


qaisjp (approved):

Pirulax:
I don't think this function is useful anymore, as `map::emplace` is the way imho. No need to make pairs, or anything, thats what this function was all about, wasn't it?

Issue #1363, opened 2020-04-09
Disconnect crash after PR #1327 (Update colshape and marker hit detection)


Dutchman101:
**Describe the bug**
This change: 3b95ff4 has introduced a disconnect crash.

<details><summary>Stacktrace 1:</summary>

```
STACK_TEXT:  
0177fa64 652121e0 00000000 333b2e62 1c6ae800 client!CClientMarker::AttachTo+0x15
0177fab8 6526049e 333b2e0e 26cf5888 0177fae4 client!CClientEntity::~CClientEntity+0x150
0177fad4 65233daf 333b2e2a 26cf5888 0177fb30 client!CClientStreamElement::~CClientStreamElement+0x3e
0177faf0 65276cda 2561a970 0177fb18 65235e51 client!CClientObject::~CClientObject+0x16f
0177fafc 65235e51 00000001 1c6ae800 1678cab0 client!CDeathmatchObject::`scalar deleting destructor'+0x1a
0177fb18 65230213 333b2fe6 0915ae70 1678cab0 client!CClientObjectManager::~CClientObjectManager+0x21
0177fb3c 6521acc5 333b2fba 01ba1a70 0915ae70 client!CClientManager::~CClientManager+0x363
0177fb60 651eb488 01c23948 690bf2d1 01c23948 client!CClientGame::~CClientGame+0x525
0177fb68 690bf2d1 01c23948 690be53f 01ba1a70 client!CClient::ClientShutdown+0x58
0177fb70 690be53f 01ba1a70 0923fa80 69074590 core!CModManager::Unload+0x71
0177fb7c 69074590 3f21a6bb 092a7398 0923fa80 core!CModManager::DoPulsePostFrame+0xf
0177fc6c 6911fa94 3f21a667 092a7398 00000000 core!CCore::DoPostFramePulse+0x8a0
0177fcb0 69126b40 0923fa80 0410d6e8 007f99b0 core!CDirect3DEvents9::OnPresent+0x164
0177fd14 007f9b12 092a7398 00000000 00000000 core!CProxyDirect3DDevice9::Present+0x30
WARNING: Stack unwind information not available. Following frames may be wrong.
0177fd38 007fb1c3 0410d6e8 001f03bc 00000000 gta_sa+0x3f9b12
00000000 00000000 00000000 00000000 00000000 gta_sa+0x3fb1c3
 
[...]

PR #1327, opened 2020-03-27
Update colshape and marker hit detection when attaching

Lpsd:
Fixes #1317

Issue #1317, opened 2020-03-25
Collision positions for colshapes are not updated when attached to an element


Lpsd:
When attaching a colshape to an object, the collision position for the colshape is not updated, so onColShapeHit will not trigger when walking into the "visual" position of the colshape (viewed with /showcol).

I've included a test resource below. I'm running `v1.5.7-release-20358.5.001`

[colshape.zip](https://github.com/multitheftauto/mtasa-blue/files/4378360/colshape.zip)

Citizen01:
Does it stay where it has been created ? or does it follow the object but at the center of the model ?
Because I remember years ago that it was attached successfully but not taking into account the offset provided so a marker attached to a car and with an offset so it is (let's say) just behind the trunk would not trigger when walking in the marker but was being triggered if I jumped onto the roof (so the col is kinda stuck to the 0, 0, 0 position of the vehicle it is attached to.

Lpsd:
I've tested the following:

- Create an object 3 units in front of the player.
- Create a colshape at 0, 0, 0
- Add an "onColShapeHit" which outputs to debugscript when player hits
- Attach the colshape to the object (using `attachElements`)
- Walk player to object, colshape hit NOT detected
- Teleport player to 0, 0, 0 - colshape hit detected
- Walk player away from 0, 0, 0 and move object to random location on the map
- Teleport back to 0, 0, 0 - colshape hit NOT detected

Without looking into the code, my conclusion from this is that `attachElements` is not updating the actual "hitbox" position for the colshape. However, `setElementPosition` does update the hitbox position for the colshape (even when used on the object it is attached to, since it cascades), which essentially means there's an offset on the hitbox from the object position and the initial colshape position.

Solution: make `attachElements` also update the hitbox position for colshapes.

I'll submit a PR tomorrow, it's late now.

Lpsd:
Following on to my previous message, I added an extra step to the end of my above test, by walking the player back into the object after moving it with `setElementPosition` (when both the object and colshape are attached) and the "onColShapeHit" event triggers. 

This means that there is no offset after using `setElementPosition` as it fixes the "hitbox" position for the colshape.

I'm currently working on this, PR later on.

PR #1666, opened 2020-09-13
Fix #1659: revert fix for #1317

### LINKED ISSUE #1659: attached markers having 2 ids causes server crash
[KWKTest.zip](https://github.com/multitheftauto/mtasa-blue/files/5191508/KWKTest.zip)

this resource will crash the server 6 seconds after you run it

if you attach a marker to an object, 
then destroy all of the object's attached elements other than the marker, 
then try to change anything about the marker the server will crash instantly.
due to the fact there are now 2 ids for a marker, the marker now exists but does not exist with a resource like this.

qaisjp:
- Reverts "Fix crash from #1363 (#1364)" (fee420d70112d6d53e75d798643a92b36837fc81).
- Reverts "Update colshape and marker hit detection when attaching (#1327)" (3b95ff48969474c691a8a4f2276f581377111f06).

Reopens #1317 and fixes #1659.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientColShape.cpp
+++ b/Client/mods/deathmatch/logic/CClientColShape.cpp
@@ -73,19 +73,7 @@ void CClientColShape::SetPosition(const CVector& vecPosition)
         UpdateSpatialData();
         CStaticFunctionDefinitions::RefreshColShapeColliders(this);
     }
-}
-
-void CClientColShape::AttachTo(CClientEntity* pEntity)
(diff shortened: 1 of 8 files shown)


qaisjp:
Not tested whether this actually fixes #1659 but I have a strong suspicion this causes the bug

sbx320:
I think we can go with #1668 instead of reverting

qaisjp:
Agreed.

PR #1668, opened 2020-09-17
Fix #1659: attached markers having 2 ids causes server crash

### LINKED ISSUE #1659: attached markers having 2 ids causes server crash

ghost:
This PR fixes the bug caused by commit 3b95ff48969474c691a8a4f2276f581377111f06. When you attach a marker to an entity, the colshape of the marker also gets attached to the entity. This allowed the scripters to access the colshape via `getAttachedElements` and deleting it caused the server to crash. So now we're attaching the colshape to the marker rather than the entity itself and making the marker colshape indestructible for scripts. ~~and hiding the marker's colshape from `getAttachedElements`.~~

Fix #1659
Replaces #1666

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -1680,7 +1680,7 @@ int CLuaElementDefs::DestroyElement(lua_State* luaVM)
     if (!argStream.HasErrors())
     {
         // Destroy it
-        if (pEntity->CanBeDestroyedByScript() && CStaticFunctionDefinitions::DestroyElement(*pEntity))
+        if (CStaticFunctionDefinitions::DestroyElement(*pEntity))
         {
             lua_pushboolean(luaVM, true);
             return 1;
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -1196,12 +1196,6 @@ int CLuaElementDefs::GetAttachedElements(lua_State* luaVM)
             {
                 CClientEntity* pAttached = pEntity->GetAttachedEntity(i);
                 assert(pAttached->GetAttachedTo() == pEntity);
-                if (IS_MARKER(pEntity))
-                {
-                    CClientMarker* marker = static_cast<CClientMarker*>(pEntity);
-                    if (marker->GetColShape() == pAttached)
-                        continue;
-                }
                 if (!pAttached->IsBeingDeleted())
                 {
                     lua_pushnumber(luaVM, i + 1);
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -1686,7 +1686,7 @@ int CLuaElementDefs::DestroyElement(lua_State* luaVM)
     if (!argStream.HasErrors())
     {
         // Destroy it
-        if (CStaticFunctionDefinitions::DestroyElement(*pEntity))
+        if (pEntity->CanBeDestroyedByScript() && CStaticFunctionDefinitions::DestroyElement(*pEntity))
         {
             lua_pushboolean(luaVM, true);
             return 1;
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -1196,6 +1196,12 @@ int CLuaElementDefs::GetAttachedElements(lua_State* luaVM)
             {
                 CClientEntity* pAttached = pEntity->GetAttachedEntity(i);
                 assert(pAttached->GetAttachedTo() == pEntity);
+                if (IS_MARKER(pEntity))
+                {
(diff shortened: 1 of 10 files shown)


sbx320 (approved):

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1196,6 +1196,12 @@ int CLuaElementDefs::GetAttachedElements(lua_State* luaVM)
             {
                 CClientEntity* pAttached = pEntity->GetAttachedEntity(i);
                 assert(pAttached->GetAttachedTo() == pEntity);
+                if (IS_MARKER(pEntity))
+                {
+                    CClientMarker* marker = static_cast<CClientMarker*>(pEntity);
```
qaisjp: I don't know if the compiler is smart enough to optimise this.

Can we move this code out of the loop?  Something like this:


Above the loop:

```cpp
CClientEntity* pIgnoredElement = nullptr;
if (IS_MARKER(pEntity))
{
    CClientMarker* marker = static_cast<CClientMarker*>(pEntity);
    pIgnoredElement = marker->GetColShape();
}
```

Inside the loop:

```cpp
if (pAttached == pIgnoredElement)
    continue;
```

qaisjp (dismissed):

qaisjp:
Will the colshape still appear in `getElementsByType("colshape")`? (or whatever the element type is called)

qaisjp:
Does the colshape appear in `getElementsByType("colshape")`? (or whatever the element type is called)

----

edit:

Hmm, there's been some discussion on Discord about this issue, and @Lpsd / @StrixG remembered that [getElementColShape](https://wiki.multitheftauto.com/wiki/GetElementColShape) actually exists.

Perhaps we should not add more code to hide it from `getElementsByType "colshape"` (because 3b95ff4 would not have changed any behaviour there)

qaisjp:
Some more chat on Discord:

> @qaisjp
>
> 1. Keep this commit [`40ddd58` (#1668)](https://github.com/multitheftauto/mtasa-blue/pull/1668/commits/40ddd58fdd36973be879876f62a3ddaa20844d4c)
> 2. Don't keep this commit [`3b68da8` (#1668)](https://github.com/multitheftauto/mtasa-blue/pull/1668/commits/3b68da80197ff8b89ce3b5adc153f7e8479043ff)
> 3. Fix any crashes caused by a marker's colshape being deleted
>
> Should we do this?
>
> ----
>
> @Lpsd: yes
>
> ----
>
> @jlillis: yup.

Thought process being:

- No new / weird / confusing behaviour introduced
- If people want to do dodgy shit, they can do dodgy shit without causing a crash


Also, the same thing will happen to pickups. (getElementColShape works for pickups too.)

ghost:
I made two changes to the PR:
- Marker colshapes can be accessed by getAttachedElements (mainly because there many ways to access the colshape)
- pickup and marker colshapes are indestructible for scripts, i.e., destroyElement function will have no effect on them.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
ghost: Although I have removed the code now, I don't think the compiler will optimize this. What you suggested is good.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1680,7 +1680,7 @@ int CLuaElementDefs::DestroyElement(lua_State* luaVM)
     if (!argStream.HasErrors())
     {
         // Destroy it
-        if (CStaticFunctionDefinitions::DestroyElement(*pEntity))
+        if (pEntity->CanBeDestroyedByScript() && CStaticFunctionDefinitions::DestroyElement(*pEntity))
```
sbx320: Wouldn't it make more sense to add this to `CStaticFunctionDefinitions::DestroyElement` instead? Seems a bit odd to me to have it as part of the Lua interface

ghost: done

ghost:
> I think we (Discord ppl) were thinking of changing the marker/pickup code to not crash with destroyed colshapes, rather than limiting destroyElement. But we can change this later if someone really wants destroyElement to work.

I don't see why anyone would want to destroy a marker's colshape.

qaisjp (approved):
> I don't see why anyone would want to destroy a marker's colshape.

Probably to use a marker for decorative purposes only without triggering events / calculating collisions.


But I'm happy to approve what you've got here! :)

StrixG (approved):

qaisjp:
Thinking out loud: if I recall correctly, the colshape collision detection depends on a grid system or other smart algorithm, so theoretically if someone wanted to optimise (& get rid of) the col detection, they could just:

- detach & move the colshapes really far out of the map, so collisions are never calculated
- using the new colshape editing functions, change the size of the colshapes to make them tiny (this might not even necessary)

PR #1672, opened 2020-09-21
Use long long as Lua integer formatting type

### LINKED ISSUE #1667: string.format more than 0x80000000 (32bit limit)
**Describe the bug**
string.format("%X", num) don't work with numbers more than 0x80000000.

**To reproduce**
```lua
iprint("0x7FFFFFFF: ", string.format("%X", 0x7FFFFFFF)) -- 7FFFFFFF
iprint("0x80000000: ", string.format("%X", 0x80000000)) -- 80000000
iprint("0x80000001: ", string.format("%X", 0x80000001)) -- 80000000 (just floored to 0x80000000)
```

**Version**
server: 1.5.7-20642
client: 1.5.7-20632
os: windows 7 32bit

**Additional context**
On older builds it works normally (i.e. server v1.5.7-release-20582)

sbx320:
This changes the integer type Lua uses for formatting to `long long`, rather than just `long` before. Since we have C++11 conforming compilers, all our supported platforms have `long long` anyway. 

Fixes #1667

### DIFF
--- a/vendor/lua/src/luaconf.h
+++ b/vendor/lua/src/luaconf.h
@@ -747,18 +747,8 @@ union luai_Cast { double l_d; long l_l; };
 ** CHANGE them if your system supports long long or does not support long.
 */
 
-#if defined(LUA_USELONGLONG)
-
 #define LUA_INTFRMLEN		"ll"
 #define LUA_INTFRM_T		long long
-
-#else
-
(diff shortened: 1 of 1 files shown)


qaisjp (approved):
Sounds good.

(Usually I recommend against editing vendor files, but I've been told `luaconf.h` is a conf file we can edit 😄)

PR #1673, opened 2020-09-24
New bone functions and onClientPedsProcessed event

### LINKED ISSUE #465: Bone positions and rotations are 1 frame lagging even when using onClientPreRender
**Describe the bug**
(Moved from https://bugs.mtasa.com/view.php?id=8055 ) 
The bone positions seem to be 1 frame laggy. It seems that the ped rotation is set after onClientPreRender.

**To Reproduce**
1. Go to my server, SAAW World War II
2. Spawn with any class at any team
3. Aim your weapon and move the screen
4. Weapon position/rotation is 1 frame delayed
5. Use first Person mode to see it better ( /fps command ) 

I am using modified bone_attach, so you could easily reproduce it with that resource anywhere. 

Resource by Ren712:
https://bugs.mtasa.com/file_download.php?file_id=1105&type=bug

**Expected behavior**
Custom attachements onClientPreRender should not be delayed by 1 frame

**Video**
http://www.youtube.com/watch?v=U95ls5931lE

**MTA version 1.5.6 stable

**Additional context**
Ren712 reported that this bug only happens with AK47/M4 aiming (his report was MTA 1.3), but you can clearly see on my server that all weapons are affected, in fact, i think bone positions are updated too late in any case and it has nothing to do with weapons at all.

### LINKED ISSUE #1646: Access to bone rotations in getPedBonePosition or a new function "getPedBoneRotation"
**Description**
Currently we dont have access to bone rotations. We have the bone_attach resource, but it is not accurate, its calculations are not precise.

I would like to see a function that gives the scripters bone rotations. The best way is to simply add them as additional return value in "getPedBonePosition", alternatively we could add a new function "getPedBoneRotation". Somehow GTA must know how to attach the weapon models to the ped skin so precisely, and that value is what we need.

Maybe someone has a working mathematical way to retrieve correct bone positions/rotations. I have not seen any bone_attach fork that does this correctly, attached weapon models are sliding or jumping around (a few millimeters only, but noticable when looking up close). And i do NOT talk about the 1-frame bone position delay, thats another issue.

**Additional context**
I just investigated the precision problem further and it is simply the fact that the distances of the hand bones are too short to get a smooth rotation value...

ghost:
Video:  https://cdn.discordapp.com/attachments/366384007535001612/758487499517198366/2020-09-24_05-36-42.mp4

This PR adds a new event and 8 new functions:
```
onClientPedsProcessed 
```
This new event also fixes the issue with bone frame delay #465
```
bool setElementBonePosition (element theElement, int boneId, float x, float y, float z)
bool setElementBoneRotation (element theElement, int boneId, float yaw, float pitch, float roll)

float x, float y, float z getElementBonePosition (element theElement, int boneId)
float yaw, float pitch, float roll getElementBoneRotation (element theElement, int boneId)

bool setElementBoneMatrix (element theElement, int boneId, matrix theMatrix)
matrix getElementBoneMatrix (element theElement, int boneId) 

void updateElementRpHAnim(element theElement)
```
`setElementBonePosition` allows you to set the translation values for the bone. 
`setElementBoneRotation` allows you to set the orientation values for the bone.
`setElementBoneMatrix` allows you to set the bone world matrix. 

`updateElementRpHAnim` should be called after setElementBoneRotation. This is a bit heavy function, so it should be only called after you are completely done with rotating bones.

You can rotate a ped head vertically, like this:
```
local pitch = 0 
addEventHandler ( "onClientPedsProcessed", root,
function ()
    pitch = pitch + 3
    if pitch > 360 then pitch = 0 end
    local BONE_HEAD = 5
    setElementBoneRotation(localPlayer, BONE_HEAD, 0, pitch, 0)
    updateElementRpHAnim(localPlayer)
end)
```

<details>
  <summary>complete list of ped bone ids (32 bones)</summary>

```
BONE_NORMAL = 0 // Normal or Root both are same
BONE_PELVIS = 1
BONE_SPINE = 2
BONE_SPINE1 = 3
BONE_NECK = 4
BONE_HEAD = 5
BONE_L_BROW = 6
BONE_R_BROW = 7
BONE_JAW = 8
BONE_L_CLAVICLE = 31
BONE_L_UPPER_ARM = 32
BONE_L_FORE_ARM = 33
BONE_L_HAND = 34
BONE_L_FINGER = 35
BONE_L_FINGER_01 = 36
BONE_R_CLAVICLE = 21
BONE_R_UPPER_ARM = 22
BONE_R_FORE_ARM = 23
BONE_R_HAND = 24
BONE_R_FINGER = 25
BONE_R_FINGER_01 = 26
BONE_L_BREAST = 302
BONE_R_BREAST = 301
BONE_BELLY = 201
BONE_L_THIGH = 41
BONE_L_CALF = 42
BONE_L_FOOT = 43
BONE_L_TOE_0 = 44
BONE_R_THIGH = 51
BONE_R_CALF = 52
BONE_R_FOOT = 53
BONE_R_TOE_0 = 54
```
</details>

I am thinking of creating a new function to get LTMs (Local Transformation Matrix) of bones. This will make it easy for people to implement their own animation system that they use to animate their models from Lua. A shader can be used for matrix multiplication, and this will give a very good performance boost. GTA SA doesn't use a shader for skinning, and that's why multiple animating peds will cause FPS drop when they are streamed in. Adding an additional function to disable GTA SA skinning to save performance would also be nice since it's very slow as it relies on the CPU for skinning rather than the GPU. Nonetheless, it would be only useful for people who want to implement a custom skeletal animation system.


Einheit-101:
My goodness, you are a LORD!

stoneage-mta:
Thanks for this PR 🥇

PlatinMTA:
Damn this is huge, good work

ghost:
I did some more research and figured out a way to rotate all 32 bones. It seems to work fine. I removed setPedNodeOrientation and replaced it with setElementBoneRotation. When you rotate a bone, the child bones will also be affected. The API is much cleaner now. Check the first post for more detailed info. Now I just need to clean up the code and the PR is complete.

Pirulax:
@saml1er Lemme know if I can help with anything.

ghost:
> 
> 
> @saml1er Lemme know if I can help with anything.

Sure, I'll let you know if I need help with something.

ghost:
This took a lot more work than I expected. The PR is finally complete. I modified the argument parser for the PR because I wanted to add support for CMatrix. It works, but I'm not sure if I've done it properly.

Client/core/Core.cpp
```diff
@@ -30,6 +30,12 @@ int WINAPI DllMain(HINSTANCE hModule, DWORD dwReason, PVOID pvNothing)
         WriteDebugEvent(SString("DLL_PROCESS_ATTACH %08x", pvNothing));
         if (IsGTAProcess())
         {
+            if (AllocConsole())
+            {
+                freopen("CONIN$", "r", stdin);
+                freopen("CONOUT$", "w", stdout);
+                freopen("CONOUT$", "w", stderr);
+            }
```
qaisjp: This is really useful we really need to just have this run in master when we're in Debug mode

ghost: yes, and it's also much faster than MTA's GUI console.

qaisjp (approved):

PR #1677, opened 2020-09-25
Allow load custom IMG containers

TheNormalnij:
This feature will save client RAM memory and model replacing time.

**API:**
```lua
-- load IMG file
img = engineLoadIMG( "file.img" )
img = EngineIMG( "file.img" ) -- OOP

-- Get files count in IMG
filesCount = engineImageGetFilesCount( img )
filesCount = img:getFilesCount() -- OOP
filesCount = img.filesCount -- OOP

-- Get file list
fileList = engineImageGetFiles( img )
fileList  = img:getFiles() -- OOP
fileList  = img.files -- OOP

-- fileList  = {
--     [fileID] = "fileName";
--     [nextFileID] = "fileName2";
-- }

for fileID = 1, filesCount  do
   -- actions
end

-- Get any file from IMG
fileContent = engineImageGetFile( img, "model.dff" ) -- You can use fileName
fileContent = engineImageGetFile( img, 1 )                  -- ... or file ID
fileContent = img:getFile( "model.txd" ) -- OOP
fileContent = img:getFile( 2 ) -- OOP

-- Add image to GTA:SA stream
engineAddImage( img )
img:add() -- OOP

-- Remove image from GTA:SA stream
engineRemoveImage( img )
img:remove() -- OOP

-- Link model to image file
-- engineImageLinkDFF( element img, string/number fileName/fileID, number modelID )
-- engineImageLinkTXD( element img, string/number fileName/fileID, number modelID )
engineImageLinkDFF( img, "file.dff", 400 ) -- You can use fileID (faster) or file name, 
engineImageLinkTXD( img, "file.txd", 123 )
img:linkDFF( "file.dff", 400 ) -- OOP
img:linkTXD( "file.txd", 123) -- OOP

-- Restore model default image
engineRestoreDFFImage( 400 )
Engine.restoreDFFImage( 400 ) -- OOP

engineRestoreTXDImage( 123 )
Engine.restoreTXDImage( 123 ) -- OOP

-- Utility function will return model texture ID
textureID = engineGetModelTXDID( 400 )
```

How to replace model with this feature:
```lua
local img = IMG( "infernus.img" )

if not img then
    return false
end

if not img:add() then
    return false
end

local infernusModelID = 411
local infernusModelTxdID = engineGetModelTXDID( infernusModelID )

img:linkDFF( "infernus.dff", infernusModelID )
img:linkTXD( "infernus.txd", infernusModelTxdID )

-- Apply changes
engineRestreamWorld()
```

**Test:**
1. Start [test resource](https://github.com/multitheftauto/mtasa-blue/files/5535781/testImg.zip)
2. Spawn infernus  `cv 411`

**Limits:**
You can load 2 IMG's. It's GTA limitation, but I hope limit can be changed.
IMG size can't bigger that 1GB (MTA limitation)


TheNormalnij:
```
local img = engineLoadIMG( "infernus.img" )
local infernusModelID = 411
local infernusTxdID = 20000 + engineGetModelTXDID( infernusModelID )
engineSetModelIMG( infernusModelID, img, 1, 551 )
engineSetModelIMG( infernusTxdID, img, 552, 571 )
```
A new way to replace model. This way uses separate thread to read model file.

**Edit: it was debug code**

TheNormalnij:
I thought i can allocate new object models only after img functions. I will add another PR only with this feature.

Pirulax:
The thing with `engineLoad*` functions is that they check the checksum of the files twice(or at least one) on the main thread(so they can return immediately).

TheNormalnij:
lol, #981 works via this.

TheNormalnij:
Added test resource

TheNormalnij:
I will change some API after TXD and COL PR's

TheNormalnij:
Ready for suggestions

Pirulax:
One thing I've noticed in your example code is that you have used 0-based indexing for `getFile`. Lua has 1-based indexing, so we should keep it consistent.

TheNormalnij:
It was fixed

Pirulax:
Also the lua class name should be `EngineIMG` (Since col is, for example, `EngineCOL`, dff is `EngineDFF`).

TheNormalnij:
What do you means?

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
 
     reinitStreaming();
 }
+
+void CStreamingSA::SetStreamingInfoForModelId(uint id, unsigned char usStreamID, uint uiOffset, ushort usSize, uint uiNextInImg)
+{
+    CStreamingInfo* pItemInfo = GetStreamingInfoFromModelId(id);
+
+    // Change nextInImg filed for prev model
+    for ( unsigned int i = 0; i < 26316; i++)
```
Pirulax: Do we have a macro for this magic `26316`?

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
     Function_ReInitStreaming reinitStreaming = (Function_ReInitStreaming)(0x40E560);
 
     reinitStreaming();
 }
+
+void CStreamingSA::SetStreamingInfoForModelId(uint id, unsigned char usStreamID, uint uiOffset, ushort usSize, uint uiNextInImg)
+{
+    CStreamingInfo* pItemInfo = GetStreamingInfoFromModelId(id);
+
+    // Change nextInImg filed for prev model
```
Pirulax: I think you've misspelled `fileid` for `filed`?

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+        }
+    }
+
+    pItemInfo->archiveId = usStreamID;
+    pItemInfo->offsetInBlocks = uiOffset;
+    pItemInfo->sizeInBlocks = usSize;
+    pItemInfo->nextInImg = uiNextInImg;
+}
+
+CStreamingInfo* CStreamingSA::GetStreamingInfoFromModelId(uint id)
```
Pirulax: I think this can be changed to `GetStreamingInfo`, the current name is too long.
Also, if changed, `id` should be changed to `modelID`, to make it clear that it should be a `model id`.

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+{
+    return &ms_aInfoForModel[id];
+}
+
+unsigned char CStreamingSA::AddArchive(const char* szFilePath)
+{
+    // Get internal IMG id
+    // By default gta sa uses 6 of 8 IMG archives
+    uchar ucArchiveId = -1;
+    for (int i = 6; i < 8; i++)
```
Pirulax: `size_t` should be use instead of `int`

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+    {
+        CArchiveInfo* info = GetArchiveInfo(i);
+        if (!info->uiStreamHandleId)
```
Pirulax: `if (!GetArchiveInfo(i)->uiStreamHandleId)`

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+        return -1;
+
+    // Get free stream handler id
+    HANDLE hFile = INVALID_HANDLE_VALUE;
+    unsigned char ucStreamID = 0;
+    
+    for (unsigned char ID = 0; ID < VAR_StreamHandlersMaxCount; ++ID)
+    {
+        HANDLE hFile = m_aStreamingHandlers[ID];
+        if (!hFile)
```
Pirulax: `if (!m_aStreamingHandlers[ID])`
(The `hFile` variable inside the loop is confusing, since it has the same name as the one in the outer scope, perhaps you intended to change that one?)

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+            break;
+        }
+    };
+
+    if (ucStreamID == 0)
+        return -1;
+
+    //  Create new stream handler
+    //  Supported state stored in 0x8E3FE0
+    DWORD dOpenFlags = *(DWORD*)(0x8E3FE0) | FILE_ATTRIBUTE_READONLY | FILE_FLAG_RANDOM_ACCESS;
```
Pirulax:
```cpp
const auto streamCreateFlags = *(DWORD*)(0x8E3FE0);
DWORD openFlags = streamCreateFlags | FILE_ATTRIBUTE_READONLY | FILE_FLAG_RANDOM_ACCESS;
```

Client/game_sa/CStreamingSA.cpp
```diff
@@ -140,15 +141,101 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
+    pItemInfo->sizeInBlocks = usSize;
+    pItemInfo->nextInImg = uiNextInImg;
+}
+
+CStreamingInfo* CStreamingSA::GetStreamingInfoFromModelId(uint id)
+{
+    return &ms_aInfoForModel[id];
+}
+
+unsigned char CStreamingSA::AddArchive(const char* szFilePath)
```
Pirulax: Consider using `std::optional` here.
You can use `return std::nullopt` instead of `return -1`.
To return the actual value, just do it the usual way: `return value;

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+{
+    m_pImgManager->RemoveFromList(this);
+
+    if (IsStreamed())
+        StreamDisable();
+
+    Unload();
+}
+
+bool CClientIMG::Load(SString sFilePath)
```
Pirulax: Use `std::string` here, if you want to take advantage of move semantics. SString doesn't have a move constructor currently.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    m_strFilename = std::move(sFilePath);
+
+/*    g_pClientGame->GetResourceManager()->ValidateResourceFile(m_strFilename, nullptr, 0);
+    if (!g_pCore->GetNetwork()->CheckFile("img", m_strFilename))
+        return false;
+*/
+
+    // Open the file
+    m_pFile = File::Fopen(m_strFilename, "rb");
```
Pirulax: I'd personally use an `std::ifstream` instead with `std::ios::binary` flag.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        fclose(m_pFile);
+        return false;
+    }
+
+    m_uiFilesCount = fileHeader.uiFilesCount;
+
+    // Read content info
+
+    m_pContentInfo.resize(m_uiFilesCount);
```
Pirulax: Is it necessary to keep this in memory?
How big can it be? Its okay if its only a few megs for a 200-300 mb img.

Client/mods/deathmatch/logic/CClientIMG.h
```diff
@@ -0,0 +1,69 @@
...
+private:
+    class CClientIMGManager*    m_pImgManager;
+
+    SString                     m_strFilename;
+    FILE*                       m_pFile;
+    unsigned char               m_ucArchiveID;
+    unsigned int                m_uiFilesCount;
+    std::vector<tImgFileInfo>   m_pContentInfo;
+
+    std::list<tLinkedModelRestoreInfo*> m_pRestoreData;
```
Pirulax: Dont you dare to use a list of heap allocated 16 byte structs. 
Please use `std::vector` here. 
Also, a `clear + shrink_to_fit` when you'd otherwise `.clear()` the `list`.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+    m_pRestoreData.push_back(pModelRestoreData);
+
+    g_pGame->GetStreaming()->SetStreamingInfoForModelId(uiModelID, m_ucArchiveID, pFileInfo->uiOffset, pFileInfo->usSize);
+    return true;
+}
+
+bool CClientIMG::UnlinkModel(unsigned int uiModelID)
+{
+    std::list<tLinkedModelRestoreInfo*>::iterator inter = m_pRestoreData.begin();
+    for (; inter != m_pRestoreData.end(); inter++)
```
Pirulax: Use a range loop here, please.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    CStreamingInfo* pCurrentStreamingInfo = g_pGame->GetStreaming()->GetStreamingInfoFromModelId(uiModelID);
+
+    tLinkedModelRestoreInfo* pModelRestoreData = new tLinkedModelRestoreInfo();
+    pModelRestoreData->uiModelID = uiModelID;
+    pModelRestoreData->ucStreamID = pCurrentStreamingInfo->archiveId;
+    pModelRestoreData->uiOffset = pCurrentStreamingInfo->offsetInBlocks;
+    pModelRestoreData->usSize = pCurrentStreamingInfo->sizeInBlocks;
+
+    m_pRestoreData.push_back(pModelRestoreData);
```
Pirulax: You can use `emplace_back` here (when you change `m_pRestoreData` to a `vector`). Just make a constructor for `tLinkedModelRestoreInfo`.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+{
+    if (!IsStreamed())
+        return false;
+
+    if (uiFileID >= m_uiFilesCount)
+        return false;
+
+    tImgFileInfo* pFileInfo = GetFileInfo(uiFileID);
+
+    CStreamingInfo* pCurrentStreamingInfo = g_pGame->GetStreaming()->GetStreamingInfoFromModelId(uiModelID);
```
Pirulax: You really should check if the model is valid.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    long iReadCount = fread(buffer.data(), 1, ulReadCount, m_pFile);
+
+    return iReadCount;
+}
+
+tImgFileInfo* CClientIMG::GetFileInfo(unsigned int usFileID)
+{
+    if (usFileID > m_uiFilesCount)
+        return NULL;
```
Pirulax: `nullptr`

Pirulax:
Nothing anymore, you've changed the example code now :P

Client/mods/deathmatch/logic/CClientIMG.h
```diff
@@ -0,0 +1,69 @@
...
+    void SetPosition(const CVector& vecPosition){};
+
+    eClientEntityType GetType() const { return CCLIENTIMG; };
+    unsigned char     GetArchiveID() { return m_ucArchiveID; };
+    unsigned int      GetFilesCount() { return m_uiFilesCount; };
+    bool              Load(SString sFilePath);
+    void              Unload();
+    tImgFileInfo*     GetFileInfo(unsigned int uiFileID);
+    unsigned int      GetFileID(std::string &sFileName);
+    long              GetFile(unsigned int uiFileID, std::string& buffer);
```
Pirulax: Try adding `const`, and `noexcept` if possible.
Only use `noexcept` where the function itself doesnt throw (Make sure none of the called functions throw. Eg.: Failed vector reallocation throws, unless you handle it with a try catch, the function itself shouldnt be `noexcept`)

Client/mods/deathmatch/logic/CClientIMGManager.cpp
```diff
@@ -0,0 +1,110 @@
...
+ *  PURPOSE:     .img container manager class
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+CClientIMGManager::CClientIMGManager(CClientManager* pManager)
+{
+    // Init
+    m_bRemoveFromList = true;
```
Pirulax, line 17: Use member init list.
Or even better, just set the default value in the header.

Client/mods/deathmatch/logic/CClientIMGManager.cpp
```diff
@@ -0,0 +1,110 @@
...
+        if (pIMG == *iter)
+        {
+            // It exists
+            return true;
+        }
+    }
+
+    // It doesn't
+    return false;
+}
```
Pirulax: `return std::find(m_List.begin(), m_List.end(), pIMG) != m_List.end()`

Client/mods/deathmatch/logic/CClientIMGManager.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+CClientIMG* CClientIMGManager::GetElementThatLinked(unsigned int uiModel)
+{
+    uchar ucArhiveID = g_pGame->GetStreaming()->GetStreamingInfoFromModelId(uiModel)->archiveId;
+    return GetElementFromArchiveID(ucArhiveID);
+}
+
+bool CClientIMGManager::IsLinkableModel(unsigned int uiModel)
+{
+    return uiModel <= 26316; // StreamModelInfoSize
```
Pirulax, line 79: Again the same magic number. Is there a macro for it?

Client/mods/deathmatch/logic/CClientIMGManager.h
```diff
@@ -0,0 +1,44 @@
...
+
+    CClientIMG* GetElementFromArchiveID(unsigned char ucStreamID);
+    CClientIMG* GetElementThatLinked(unsigned int uiModel);
+
+    bool        RestoreModel(unsigned int uiModel);
+    static bool IsLinkableModel(unsigned int uiModel);
+
+    std::list<CClientIMG*>::const_iterator IterBegin() { return m_List.begin(); }
+    std::list<CClientIMG*>::const_iterator IterEnd() { return m_List.end(); }
+
```
Pirulax: Can be replaced by `const auto& GetAll() const noexcept { return m_List; }`

Client/mods/deathmatch/logic/CClientIMGManager.h
```diff
@@ -0,0 +1,44 @@
...
+private:
+    void AddToList(CClientIMG* pIMG) { m_List.push_back(pIMG); }
+    void RemoveFromList(CClientIMG* pIMG);
+
+    std::list<CClientIMG*> m_List;
```
Pirulax, line 41: Use `std::vector<CClientIMG*>`.
Using a `list` for pointers is a really, really, really bad idea.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -118,6 +132,26 @@ void CLuaEngineDefs::AddEngineTxdClass(lua_State* luaVM)
...
 
+void CLuaEngineDefs::AddEngineImgClass(lua_State* luaVM)
+{
+    lua_newclass(luaVM);
+
+    lua_classfunction(luaVM, "create", "engineLoadIMG");
+    lua_classfunction(luaVM, "add", "engineAddImage");
+    lua_classfunction(luaVM, "remove", "engineRemoveImage");
+    lua_classfunction(luaVM, "getFile", "engineImageGetFile");
+    lua_classfunction(luaVM, "getFileList", "engineImageGetFileList");
```
Pirulax: Perhaps `getFiles` would be more correct.
For example its `getElementsByType` not `getElementListByType`.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
     return 1;
 }
 
+CClientIMG* CLuaEngineDefs::EngineLoadIMG(lua_State* const luaVM, std::string strFilePath)
+{
+    CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+    if (!pLuaMain)
+        return false;
```
Pirulax: No need to check this.
Also IIRC you have a `getVirtualMachineResource` function in `CLuaManager`.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
     return 1;
 }
 
+CClientIMG* CLuaEngineDefs::EngineLoadIMG(lua_State* const luaVM, std::string strFilePath)
```
Pirulax: Change `strFilePath` to `strRelativeFilePath`.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+            // Success. Make it a child of the resource img root
+            pImg->SetParent(pRoot);
+
+            return pImg;
+        }
+        else
+        {
+            delete pImg;
+            throw std::invalid_argument("Error loading IMG");
+        }
```
Pirulax, line 586: This is a really bad idea IMHO.
The manager should create the instance, not the Lua function.
Anyways, this can be changed later on.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+}
+
+std::vector<std::string> CLuaEngineDefs::EngineImageGetFileList(CClientIMG* pIMG)
+{
+    uint uiFilesCount = pIMG->GetFilesCount();
+    std::vector<std::string> aOutput;
+    aOutput.reserve(uiFilesCount);
+
+    for (unsigned int i = 0; i < uiFilesCount; i++)
+        aOutput.push_back(pIMG->GetFileInfo(i)->szFileName);
```
Pirulax: `emplace_back` for clarity. They do the same in this case.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+}
+
+uint CLuaEngineDefs::EngineImageGetFilesCount(CClientIMG* pIMG)
+{
+    return pIMG->GetFilesCount();
+}
+
+std::vector<std::string> CLuaEngineDefs::EngineImageGetFileList(CClientIMG* pIMG)
+{
+    uint uiFilesCount = pIMG->GetFilesCount();
```
Pirulax: `GetFilesCount` should be returning `size_t`.
Use `const auto` or `const size_t` here.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+    std::vector<std::string> aOutput;
+    aOutput.reserve(uiFilesCount);
+
+    for (unsigned int i = 0; i < uiFilesCount; i++)
```
Pirulax: Use `size_t` for loop counters.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+
+std::string CLuaEngineDefs::EngineImageGetFile(CClientIMG* pIMG, std::variant<std::string, uint> file)
+{
+    uint  uiFileID = -1;
+    uint* pFileID = std::get_if<uint>(&file);
+
+    if (pFileID)
+        uiFileID = *pFileID - 1;
+    else
+        uiFileID = pIMG->GetFileID(std::get<std::string>(file));
```
Pirulax:
```cpp
const auto fileID = std::holds_alternative<std::string> ? 
    pIMG->GetFileID(std::get<std::string>(file)) : std::get<uint>(file);
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+    std::vector<std::string> aOutput;
+    aOutput.reserve(uiFilesCount);
+
+    for (unsigned int i = 0; i < uiFilesCount; i++)
+        aOutput.push_back(pIMG->GetFileInfo(i)->szFileName);
+
+    return aOutput;
+}
+
+std::string CLuaEngineDefs::EngineImageGetFile(CClientIMG* pIMG, std::variant<std::string, uint> file)
```
Pirulax: If you would change your `GetFileID` function to take an `std::string_view`, this could be changed as well.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -503,6 +537,155 @@ int CLuaEngineDefs::EngineImportTXD(lua_State* luaVM)
...
+    if (uiModelID >= 20000)
+        throw std::invalid_argument("Expected model ID in range [0-19999] at argument 3");
+
+    uint  uiFileID = -1;
+    uint* pFileID = std::get_if<uint>(&file);
+
+    if (pFileID)
+        uiFileID = *pFileID - 1;
+    else
+        uiFileID = pIMG->GetFileID(std::get<std::string>(file));
```
Pirulax: Now that you've used this function more than once, I'd say just make a static function for it:
```cpp
static uint ResolveFileID(const std::variant<std::string, uint>& variant)
{
    if (std::holds_alternative<std::string>)
        return pIMG->GetFileID(std::get<std::string>(file));
    return std::get<uint>(file);
}
```
Put it above the img engine functions.

Pirulax (dismissed):

Pirulax:
Uh, sorry, I wanted to open a PR instead of pushing it here.
Oh well. Hopefully you wont be mad at me😅

Pirulax:
I've resolved some of my comments.
To be honest I don't see the point of `GetFile` (it returns the binary content of the file).
This might lead to people using an archive as an encrypted file storage (eg.: They load the IMG, and use `getFile` to get the content of a file, then decrypt it, and load it using engineLoadDFF) which isn't what we want.
I suggest to remove this function entirely.

Pirulax:
Note: Before merging a check for all TXDs, DFFs in the img needs to be added thru netcode to prevent buffer overflow code injection.

Pirulax:
Issues:
- The world should be re-streamed when the `CClientIMG` is unloaded, otherwise the isn't visually reset to the default GTA one.
- [This assert in CModelInfoSA](https://github.com/multitheftauto/mtasa-blue/blob/1463866b271b12837b4ac557dacb81e261273243/Client/game_sa/CModelInfoSA.cpp#L370) fails when trying to load `411` which is your infernus model. (Keep in mind I had another img loaded, but I dont think it contributed to the problem). 

Thoughts:
- What if another archive tries to link the model?
- What if engineReplaceModel is called onto a model thats linked? 

Issue with the example resource: You forgot to add `engineRestreamWorld()` at the end, which causes the model not to be loaded from the img.

Client/game_sa/CStreamingSA.cpp
Pirulax: GameSA project is c++14, std optional is c++17.

Pirulax:
Please take a look at [this ](https://github.com/TheNormalnij/mtasa-blue/pull/2/commits/8180adc601dbca4c2c25111dd5f17857164814eb) commit, unless you want to experience crashes :D

prnxdev:
IMHO OOP version should be
```lua
Engine.loadIMG(path)
```

Pirulax:
That would be inconsistent with `EngineDFF/COL/TXD`.

prnxdev:
I think whole OOP elements that does something inside constructor (like in this example) are misleading. It would be better if those method be named as their behavior. For example in this situation those `EngineDFF/COL/TXD` should be `Engine.loadDFF/COL/TXD`. 

There is also this situation:
```lua
Player('John Doe')
```
could be
```lua
Player.getFromName('John Doe')
```

Pirulax:
Thats literally the whole point of a constructor, to construct an instance.
`EngineTXD`, etc.. doesnt do anything by itself until you call `import` (in case of txd).
The same thing is true for `EngineIMG`: The IMG isn't actually loaded into GTA unless you call `add()`.

TheNormalnij:
I notice that GTA can't read some big files. It's hard to fix. I waste to many times here. Now i know only one helpful info: `ReadFile` function in GTA SA at 0x4065D9 can fail with ERROR_NOACCESS (998) error code.
**EDIT**: I know why

AlexTMjugador:
@Pirulax:
> This might lead to people using an archive as an encrypted file storage

Although I understand that negates some of (if not all) the performance improvements of IMG archives, why not keep it? Storing a bunch of files in a IMG container definitely seems more organized and performant that storing that bunch of files in a directory hierarchy, as the latter has more filesystem overhead, or reinventing the wheel by implementing your own container in Lua. I'd say it's a nice feature; I'd just make sure that people is aware of the inferior performance via documentation.

Pirulax:
I wonder if thats because of our own filehandle to the file?

ghost:
If the size of a file in the .img archive is greater than `CStreaming::ms_streamingBufferSize` then the game will crash. When you read the .img archive, you'll need to check the size of all files and allocate a new buffer for `CStreaming::ms_pStreamingBuffer` if you find a file of greater size.

Pirulax:
Yes, thats what I thought, I remember seeing a buffer for col files as well, which was around 8-9 kb.  Well, allocating a new buffer shoudn't be too hard.
Actually, I'm not sure how this whole streaming stuff works. Is there a way to lock a mutex, so we can change all the buffer stuff? Because, afaik the streamer runs on a separate thread.

TheNormalnij:
> Well, allocating a new buffer shoudn't be too hard.

Im not sure about this. I don't have idea how i can change pointer to buffer during thread with ReadFileA. I need little help with this.

thisdp:
what about https://wiki.multitheftauto.com/wiki/EngineLoadIMGContainer

TheNormalnij:
Ah, shit, here we go again

TheNormalnij:
Bug fixed

Client/mods/deathmatch/logic/CClientIMGManager.cpp
TheNormalnij: No

Client/mods/deathmatch/logic/CClientIMG.cpp
TheNormalnij: It's unnecessary here. Because mta check model before object creation.

Client/mods/deathmatch/logic/CClientIMGManager.h
TheNormalnij: Hmm we don't need this lines

TheNormalnij:
Ready for re-review and merge

Client/game_sa/CStreamingSA.cpp
```diff
@@ -247,3 +251,47 @@ void CStreamingSA::RemoveArchive(unsigned char ucArhiveID)
...
+
+    // Close old streaming handle
+    TerminateThread(*phStreamingThread, 0);
+
+    // Create new buffer
+    if (uiBlockSize & 1)
+        uiBlockSize = uiBlockSize++ + 1;
+
+    typedef int(__cdecl * Function_CMemoryMgr_MallocAlign)(uint32 uiCount, uint32 uiAlign);
+    int iPointer = ((Function_CMemoryMgr_MallocAlign)(0x72F4C0))(uiBlockSize << 11, 2048);
```
Pirulax: use `void*`

Client/game_sa/CStreamingSA.cpp
```diff
@@ -247,3 +251,47 @@ void CStreamingSA::RemoveArchive(unsigned char ucArhiveID)
...
+
+    // Copy data from old buffer to new buffer
+    MemCpyFast((void*)iPointer, (void*)ms_pStreamingBuffer[0], ms_streamingBufferSize);
+    MemCpyFast((void*)(iPointer + 1024 * uiBlockSize), (void*)ms_pStreamingBuffer[1], ms_streamingBufferSize);
```
Pirulax: You can copy both in one go I'd say.

Client/game_sa/CStreamingSA.cpp
```diff
@@ -247,3 +251,47 @@ void CStreamingSA::RemoveArchive(unsigned char ucArhiveID)
...
+    ms_pStreamingBuffer[0] = iPointer;
+    ms_pStreamingBuffer[1] = iPointer + 2048 * ms_streamingBufferSize;
+
+    int pointer = *(int*)0x8E3FFC;
+    SGtaStream(&streaming)[5] = *(SGtaStream(*)[5])(pointer);
+
+    streaming[0].lpBuffer = ms_pStreamingBuffer[0];
+    streaming[1].lpBuffer = ms_pStreamingBuffer[1];
+
+    free(reinterpret_cast<void*>(ms_pStreamingBuffer[0] - 1));
```
Pirulax: Use `CMemoryManager::AlignFree`

Client/game_sa/CStreamingSA.cpp
```diff
@@ -247,3 +251,47 @@ void CStreamingSA::RemoveArchive(unsigned char ucArhiveID)
...
+
+    // Create new streming handle
+    auto pStreamingThreadId = *(LPDWORD)0x8E4000;
+    HANDLE   hStreamingThread = CreateThread(0, 0x10000u, (LPTHREAD_START_ROUTINE)0x406560, 0, 4u, &pStreamingThreadId);
+
+    HANDLE hCurrentThread = GetCurrentThread();
+    int    iPriority = GetThreadPriority(hCurrentThread);
+    SetThreadPriority(hStreamingThread, iPriority);
+    ResumeThread(hStreamingThread);
```
Pirulax: I think saving the original thread's priority would be the right thing

TheNormalnij: This thread designed without support any good locking stuff

Client/game_sa/CStreamingSA.cpp
TheNormalnij: this is it

Client/game_sa/CStreamingSA.cpp
TheNormalnij: no, we should save alignment here

Client/game_sa/CStreamingSA.cpp
TheNormalnij: yeah, but we do some math operations with `iPointer`

Client/game_sa/CStreamingSA.cpp
TheNormalnij: This thread uses main thread priority

Client/mods/deathmatch/logic/CClientIMGManager.cpp
Pirulax: Even with the new stuff FastMan added?

Client/game_sa/CStreamingSA.cpp
Pirulax: But call the function.

Client/game_sa/CStreamingSA.cpp
Pirulax: Do it like in the original code.
https://github.com/codenulls/gta-reversed/blob/d5e83be9f188daf57902d0199f6c5019ba6f3e8d/source/game_sa/CStreaming.cpp#L2745-L2747

Client/mods/deathmatch/logic/CClientIMGManager.cpp
TheNormalnij: yes

Client/game_sa/CStreamingSA.cpp
TheNormalnij: This code don't uses `void*` also

TheNormalnij: void* used

Client/game_sa/CStreamingSA.cpp
TheNormalnij: used

Pirulax:
The code isn't the cleanest (we could use more SA functions), but I'll let it pass, it can be refactored later.

Pirulax:
World should be restreamed on resource stop (if it had any loaded images).

TheNormalnij:
Oh, now it don't contains bus, i think

Client/mods/deathmatch/logic/CClientIMGManager.cpp
```diff
@@ -95,3 +100,18 @@ void CClientIMGManager::RemoveFromList(CClientIMG* pIMG)
         m_List.remove(pIMG);
     }
 }
+
+void CClientIMGManager::UpdateStreamerBufferSize()
+{
+    ushort usRequestStreamSize = m_uiDefaultStreamerBufferSize;
+
+    for (const auto& pImg : m_List)
```
Pirulax:
```suggestion
    for (CClientIMG* pImg : m_List)
 ```

File: Client/mods/deathmatch/logic/CClientIMG.h [hunk 38]
```diff
@@ -83,6 +84,7 @@ class CClientIMG : public CClientEntity
     std::string                m_filePath;
     unsigned char              m_ucArchiveID;
     std::vector<tImgFileInfo>  m_fileInfos;
+    unsigned short             m_usRequiredBufferSize;
```
Pirulax, line 87: What units is the file size in? Bytes? Please rename the variable. 
Also, just use size_t here imho.

Client/game_sa/CStreamingSA.h
```diff
@@ -66,6 +66,7 @@ class CStreamingSA : public CStreaming
     unsigned char   AddArchive(const char* szFilePath);
     void            RemoveArchive(unsigned char ucStreamHandler);
     void            SetStreamingBufferSize(uint32 uiSize);
+    uint32          GetStreamingBufferSize() { return ms_streamingBufferSize * 2; };
```
Pirulax: Could you add a comment or change the function name? This is really counterintuitive.

Pirulax (commented):
It still not as good as I'd like it to be, but it works, so it can be merged, and later refactored.

Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -0,0 +1,245 @@
...
+        return false;
+    }
+
+    m_ifs.read(reinterpret_cast<char*>(m_fileInfos.data()), sizeof(tImgFileInfo) * (std::streampos)fileHeader.uiFilesCount);
+    if (m_ifs.fail() || m_ifs.eof())
+    {
+        m_ifs.close();
+        return false;
+    }
+
```
Pirulax, line 93: Okay, here we need to call netc to check all binary files.
Here's the code to do it:
`if (!g_pCore->GetNetwork()->CheckFile(<file type>, <filename>, buffer.data(), buffer.size()))`
The types possible are:
- dff
- col
- txd

You will have to load whole files into a buffer. I'd recommend you use  `std::unique_ptr<char[]>(new char[m_usRequiredBufferSize])`. (Note the usage of `new` instead of `make_unique`, this is because we don't want to value initialize the array (eg.: zero it out)).

Actually, maybe defining an additional function would make more sense.

**Also, we need a way to lock the image, since a malicious scripter might modify it after it has been verified.**

Pirulax:
Please do not merge until [this](https://github.com/multitheftauto/mtasa-blue/pull/1677/files#r619783879) has been resolved.

Client/game_sa/CStreamingSA.cpp
```diff
@@ -257,7 +257,13 @@ void CStreamingSA::SetStreamingBufferSize(uint32 uiBlockSize)
     if (uiBlockSize == ms_streamingBufferSize * 2)
         return;
 
-    // Suspend old streaming handle
+    int pointer = *(int*)0x8E3FFC;
+    SGtaStream(&streaming)[5] = *(SGtaStream(*)[5])(pointer);
+
+    // Wait while streaming threads ends tasks
+    while (streaming[0].bInUse && streaming[1].bInUse)
```
Pirulax, line 275: I think you've forgot a ;
```suggestion
    while (streaming[0].bInUse && streaming[1].bInUse);
```

TheNormalnij: Unnecessary there

Client/game_sa/CStreamingSA.h
TheNormalnij: It's streaming buffer size. What can i say there ?

Client/mods/deathmatch/logic/CClientIMG.cpp
TheNormalnij: IMG file uses `FILE_SHARE_READ` asses. I don't know, but maybe it's enough

Client/mods/deathmatch/logic/CClientIMG.h
TheNormalnij: Img uses only size of blocks. So whole class and manager use this type

Client/game_sa/CStreamingSA.h
Pirulax: Yeah, but you return the 2x the buffer size.
Perhaps rename the function to `GetTotalBufferSize` or something.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -626,14 +626,32 @@ bool CLuaEngineDefs::EngineImageLinkDFF(CClientIMG* pIMG, std::variant<size_t, s
...
         throw std::invalid_argument(SString("Expected modelid in range 0 - 19999, got %d", uiModelID));
-    return pIMG->LinkModel(uiModelID, ResolveIMGFileID(pIMG, file));
+
+    size_t fileID = ResolveIMGFileID(pIMG, file);
+    std::string buffer;
+    if (!pIMG->GetFile(ResolveIMGFileID(pIMG, file), buffer))
+        throw std::exception("Failed to read file. Probably EOF reached, make sure the archieve isn't corrupted.");
+
+    if (!g_pCore->GetNetwork()->CheckFile("dff", "", buffer.data(), buffer.size()))
+        throw std::exception("Failed to link file. Make sure the archieve isn't corrupted.");
```
Pirulax: `ArgumentParser` only catches `invalid_argument` and up.

Client/mods/deathmatch/logic/CClientIMG.cpp
TheNormalnij: Another checks added

TheNormalnij: Resolved?

Pirulax:
```lua
-- Restore model default image
engineRestoreDFFImage( 400 )
Engine.restoreDFFImage( 400 ) -- OOP
```
Could we use `engineRestoreModel` instead of `engineRestoreDFFImage`?

```lua
engineRestoreTXDImage( 123 )
Engine.restoreTXDImage( 123 ) -- OOP
```
Maybe `engineRestoreTXD`?

Pirulax:
Also, what about COL's?
I don't think netc accepts dffs with inbuilt cols.

TheNormalnij:
This tools doesn't work with COL's yet, i think,

TheNormalnij:
> Could we use `engineRestoreModel` instead of `engineRestoreDFFImage`?

`engineRestoreModel` restores dff and txd together

Pirulax:
Well, I presume it works with dff's with cols inside them?

TheNormalnij:
Dff checker in netc should have this check. I think this checker contains check for inbuilt col in vehicles

Pirulax:
And it probably explicitly disallows cols for other types of models.
@ccw808 should be able to help us with this.

Pirulax:
We have to increase the stream memory limit, otherwise this is pretty much useless.
Once the stream memory limit is reached (which is easy with big models), streaming stops working, and everything flickers horribly.

Dutchman101:
> We have to increase the stream memory limit, otherwise this is pretty much useless.
> Once the stream memory limit is reached (which is easy with big models), streaming stops working, and everything flickers horribly.

Am i right to say that issue #2221 blocks this PR then?

Pirulax:
The memory limit can be changed with a single variable, its a 1 liner fix. We could just add a function to do so.

Dutchman101:
> The memory limit can be changed with a single variable, its a 1 liner fix. We could just add a function to do so.

I agree, we can just add that function now to clear the road for this PR. See my [latest post](https://github.com/multitheftauto/mtasa-blue/issues/2221#issuecomment-862439378) on #2221 for reasoning on why I agree.

Pirulax:
Theres an issue, called `CClientModelCacheManager`. It keeps lots of vehicle/ped/weapon/upgrade models in memory 
The issue is that it doesnt account for the size of the models now (as it wasn't designed to handle custom engine images). For example: By default a vehicle model is ~100 kbytes, but now, with custom models its, more like 4-5 MB / model, and you get... 250-300 MB that wont be freed, and since thats > 256 MB streaming limit the game starts glitching out (tries to unload models to free up memory, but it cant because vehicles are referenced by the cache manager).

Heres the stream memory => cache memory mapping. It's totally wrong, as vehicle models can be a few megabytes in size now.
https://github.com/multitheftauto/mtasa-blue/blob/3f9348106feb1e64757fe92a06f6973bf3decd23/Client/core/CModelCacheManager.cpp#L222-L228

The solution is to either:
- Remove it. (As I dont see the point of keeping it, we should let the streamer handle model loading)
- Refactor it to account for model size. (Currently its designed to use ~100 megs max)

I did a test (removed all the code of it), and everything seems to work just fine.

@Dutchman101 [there is](https://github.com/multitheftauto/mtasa-blue/blob/3f9348106feb1e64757fe92a06f6973bf3decd23/Client/core/CModelCacheManager.cpp#L130-L187) some crash fixing code (related to vehicle upgrades), do you know what it fixes?

@ccw808 Any idea what it was built for, what issues does it fix? Edit - Used git blame in VS, and found commit the commit in which it was added (fefd0887). I really want to just remove it, but it seems like there's a good reason for its existence. Sadly it uses way too much RAM with this PR. (And also glitches out the game, as these models basically can't be freed by GTA)

Dutchman101:
> @Dutchman101 [there is](https://github.com/multitheftauto/mtasa-blue/blob/3f9348106feb1e64757fe92a06f6973bf3decd23/Client/core/CModelCacheManager.cpp#L130-L187) some crash fixing code (related to vehicle upgrades), do you know what it fixes?

I don't really know, but because that crash fix uses report ID 8544 i can tell (from metrics) that approximately 70 users have had it trigger during the past month. Maybe it's very old but after all it's still firing for those users, so i guess if it was cut out they would've crashed there.

But as if it appears to be the report ID is only used in a certain condition (not the entire crashfix - just "if LowestUnsafeUpgrade has fallen"), true averted crash count will probably be higher. So unless you understand the crash reason and how to fix it in a better way I'd advise not to touch it.

Pirulax:
Yeah, but the issue is that, people now may replace these upgrades with the img, and if the upgrades are big (in file size), then this will use a lot of memory (as from what I understand all upgrades are pre-streamed).
Anyways, I'll leave that crash fix in, and if someone has issues with it, then it could be investigated further.

Disinterpreter:
Is it ready? I think it is a useful feature.

Xenius97:
> Is it ready? I think it is a useful feature.

No, but on the wiki is an usefun function that works as well.
https://wiki.multitheftauto.com/wiki/EngineLoadIMGContainer

TheNormalnij:
> > Is it ready? I think it is a useful feature.
> 
> No, but on the wiki is an usefun function that works as well. https://wiki.multitheftauto.com/wiki/EngineLoadIMGContainer

This feature is ready now

patrikjuvonen:

TheNormalnij:
Fixed

patrikjuvonen (approved):

Xenius97:
I've tested it, everything looks great. Memory usage is only 700mb when 668mb car pack img container is loaded.
This will solve most of out of memory crashes by models. Good job.
image

Xenius97:
@TheNormalnij For objects seems not working properly.
When you replace collision you need full reconnect instead of engineRestreamWorld, but when you load collision too, model will not have textures.

Without collisions:
image


Collisions loaded, used only restream
image


After reconnect:
image


Test resource:
[buildings.zip](https://github.com/multitheftauto/mtasa-blue/files/10345431/buildings.zip)

TheNormalnij:
Can be avoided via `engineReplaceCOL` is `setTimer`

Xenius97:
> Can be avoided via `engineReplaceCOL` is `setTimer`

Yes, it's working fine with timer, tested. Maybe put a warning/infobox into wiki when this will be merged.

But i've found another thing, when you load vehicle model to incomatible model (eg. car mod to bike) game will instant crash.
Don't know this happens via regular model replacement too, or not.

TheNormalnij:
> Yes, it's working fine with timer, tested. Maybe put a warning/infobox into wiki when this will be merged.

#2866 makes timer unnecessary

> But i've found another thing, when you load vehicle model to incomatible model (eg. car mod to bike) game will instant crash.
> Don't know this happens via regular model replacement too, or not.

I think, we can't fix all cases. Any additional checks will slow down performance, when you link IMG files to models.

Xenius97:
I've noticed something, all engine* functions are supporting raw data except this. Possible to add raw data support to engineLoadIMG? Cause if you wan't to protect your models, you need to compile it with some method like tea, aes128 and impossible to load if function doesn't supporting it.

Pirulax:
What do you mean by "raw data"?
The point of all of this is to stream data from the disk, rather than load it into memory.
If we were to keep it in memory, then it would be just a worse `engineLoad*` replacement.
I suggest that instead of trying to encrypt your models, just do DRM.
While, it would be possible to somehow integrate encryption into this function I don't think it would solve the core problem.

PR #1684, opened 2020-09-28
Allow allocating new object models using engineRequestModel

TheNormalnij:
Use first argument `object` in `engineRequestModel` to allocate new model. New model will based on trashcan (1337).
Changed valid objects list and this can cause bugs for old scripts. For example, you can't create object with ID 1. New method think  that it's a ped.
Valid objects are [here](https://pastebin.com/H8rNeqaC)

**API:**
```
engineRequestModel( "object" )
```

**TEST:**
1. Start [test resource](https://github.com/multitheftauto/mtasa-blue/files/5431350/testAddObjects.zip)
2. Press "home" button.
3. This part of map use new allocated ID's


qaisjp:
> Changed valid objects list and this can cause bugs for old scripts.

If this is a separate bugfix, please can you submit it in a separate pull request?

qaisjp:
Excellent work, by the way :)

TheNormalnij:
I can't. Without this fix you can't create new objects and change models and textures.

Client/mods/deathmatch/logic/CClientObjectManager.cpp
```diff
@@ -142,23 +92,18 @@ CClientObject* CClientObjectManager::Get(ElementID ID)
...
+    if (ulObjectModel < 20000)
     {
-        unsigned int uiChunk = ulObjectModel / (sizeof(unsigned int) * 8);
-        unsigned int shift = ulObjectModel - (uiChunk * sizeof(unsigned int) * 8);
-        unsigned int bit = 1 << shift;
-        bIsValid = !!(g_iValidObjectModels[uiChunk] & bit);
-
-        // Return whether we can grab the model information or not
-        if (bIsValid && !g_pGame->GetModelInfo(ulObjectModel))
+        CModelInfo* pModelInfo = g_pGame->GetModelInfo(ulObjectModel);
```
qaisjp: Does this work if the model isn't loaded?

TheNormalnij: Yes

Client/game_sa/CStreamingSA.cpp
```diff
@@ -133,4 +133,10 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
         add     esp, 0xC
     }
-}
\ No newline at end of file
+}
+
+CStreamingInfo* CStreamingSA::GetStreamingInfoFromModelId(ushort id)
+{
+    CStreamingInfo* pItemInfo = (CStreamingInfo*)(ARRAY_StreamModelInfo);
+    return pItemInfo + id;
```
Pirulax: I'd suggest using `return pItemInfo[id]`, as its an array

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -70,6 +80,21 @@ bool CClientModel::Deallocate(void)
...
+            {
+                if ((*iter)->GetModel() == m_iModelID)
+                {
+                    if ((*iter)->IsStreamedIn())
+                    {
+                        (*iter)->StreamOutForABit();
+                    }
+                    (*iter)->SetModel(1337);
+                }
+            }
```
Pirulax: Id suggest using `auto`, and all the new features.
```cpp
for (auto iter = pObjectManager->IterBegin(); iter != pObjectManager->IterEnd(); iter++)
{
    const CClientObject* pObject = *iter;
    // reset of the code here
}
```
A more cleaner solution would be to add a method to the object manager:
```cpp
const auto& GetObjects() { return m_objects; } // or whatever the member variable std::vector is. 
```
And just use a `range based for loop`:
```cpp
for (auto* pObject : pObjectManager->GetObjects())
{
    // code
}
```

Client/mods/deathmatch/logic/CClientObjectManager.h
```diff
@@ -25,6 +25,9 @@ class CClientObjectManager
 
     void DeleteAll();
 
+    std::vector<CClientObject*>::const_iterator IterBegin() { return m_Objects.begin(); };
+    std::vector<CClientObject*>::const_iterator IterEnd() { return m_Objects.end(); };
```
Pirulax: Ah, so you just added these methods? Then just change this to my suggestion I guess:
```cpp
const auto& GetObjects() { return m_Objects; }
```

Client/sdk/game/CModelInfo.h
```diff
@@ -28,6 +28,17 @@ class CBoundingBox
...
+{
+    MODEL_INFO_TYPE_ATOMIC = 1,
+    MODEL_INFO_TYPE_TIME = 3,
+    MODEL_INFO_TYPE_WEAPON = 4,
+    MODEL_INFO_TYPE_CLUMP = 5,
+    MODEL_INFO_TYPE_VEHICLE = 6,
+    MODEL_INFO_TYPE_PED = 7,
+    MODEL_INFO_TYPE_LOD_ATOMIC = 8,
+};
+
```
Pirulax: I'd suggest using `enum class` here:
```cpp
enum class eModelInfoType : unsigned char
{
    ATOMIC = 1,
    TIME = 3,
    WEAPON = 4,
    CLUMP = 5,
    VEHICLE = 6,
    PED = 7,
    LOD_ATOMIC = 8,
};

```

Client/sdk/game/CStreaming.h
```diff
@@ -11,11 +11,24 @@
...
+{
+    WORD  prevId;
+    WORD  nextId;
+    WORD  nextInImg;
+    unsigned char flg;
+    unsigned char archiveId;
+    DWORD offsetInBlocks;
+    DWORD sizeInBlocks;
+    DWORD loadState;
+};
```
Pirulax: Is `unsigned char` meant to be a `BYTE`? Not sure which one we should use here. (afaik unsigned char is guaranteed to be exactly a byte. But for the sake of consistency)

Pirulax (changes requested):

Client/mods/deathmatch/logic/CClientObjectManager.h
qaisjp: I think they want to keep `m_Objects` private, assuming it's already currently private.

Client/sdk/game/CStreaming.h
qaisjp: If this is changed I think we prefer "non-Windows constants"

ghost:
Thanks for the PR. I skimmed through the code, and it looks good. Testing the PR is going to take some time. I suggest we merge this after 1.5.8.

Client/mods/deathmatch/logic/CClientObjectManager.h
Pirulax: With a `const auto&` it'll remain basically private. (you can do the same with 2 iterators as with a `const std::vector&`. But a range based loop is way more readable than an iter one)

Client/game_sa/CStreamingSA.cpp
TheNormalnij: This can't be compiled. Idk how to replace this

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: > E1087	no instance of function "SharedUtil::CMappedContainer<T, LIST_TYPE>::begin [with T=CClientObject *, LIST_TYPE=std::vector<CClientObject *, std::allocator<CClientObject *>>]" matches the argument list and object (the object has type qualifiers that prevent a match)

Client/game_sa/CStreamingSA.cpp
qaisjp: Does this work?

```cpp
CStreamingInfo[] pItemInfo = (CStreamingInfo[])(ARRAY_StreamModelInfo);
return pItemInfo[id]
```

TheNormalnij: Invalid syntax 🤷‍♀️

tederis: It's just an array of pointers, guys
```
CStreamingInfo** pItemInfo = (CStreamingInfo**)(ARRAY_StreamModelInfo);
return pItemInfo[id]
```

tederis:
```
CStreamingInfo* GetStreamingInfoFromModelId(uint id)
{
    CStreamingInfo* pItemInfo = (CStreamingInfo*)(0x8E4CC0);
    return &pItemInfo[id];
}
```

Client/mods/deathmatch/logic/CClientObjectManager.h
tederis: I think the method should be also marked **const**:

```const auto& GetObjects() const { return m_Objects; }```

File: Client/mods/deathmatch/logic/CClientModel.cpp (hunk 3 above)
Comment (tederis) [reply Client/mods/deathmatch/logic/CClientModel.cpp]: Add these lines to the CMappedArray
```
typename LIST_TYPE::const_iterator begin() const { return m_List.begin(); }
typename LIST_TYPE::const_iterator end() const  { return m_List.end(); }
```

Next, you should make a getter const:
```
const auto& GetObjects() const { return m_objects; }
```
And finally:
```
for (const auto pObject : pObjectManager->GetObjects())
{
    // code
}

```
It should work but you probably want using cbegin and cend instead

File: Client/mods/deathmatch/logic/CClientObjectManager.cpp [hunk 7]
```diff
@@ -10,56 +10,6 @@
 
 #include "StdInc.h"
 
-// Generated by MTA10\utils\gentable\gentable_objmodels.cpp
```
qaisjp, line 13: I think we can delete this file from the repo if we don't use it anywhere else

Client/mods/deathmatch/logic/CClientObjectManager.cpp
```diff
@@ -142,23 +92,18 @@ CClientObject* CClientObjectManager::Get(ElementID ID)
...
-            bIsValid = false;
+            eModelInfoType eType = pModelInfo->GetModelType();
+            if (eType == eModelInfoType::CLUMP || eType == eModelInfoType::ATOMIC || eType == eModelInfoType::WEAPON)
+                return true;
         }
     }
 
-    return bIsValid;
+    return false;
 }
```
qaisjp: Since this entire function has essentially been rewritten, please can you make use of early returns?

(E.g. `if ((x > 2000) || (x < 0)) { return false; }`  etc ... except formatted correctly)

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -516,7 +516,7 @@ bool CModelInfoSA::IsValid()
 {
     if (m_dwModelID >= 20000 && m_dwModelID < MODELINFO_MAX)
         return true;
-    return ppModelInfo[m_dwModelID] != 0;
+    return ppModelInfo[m_dwModelID] != 0 && pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID)->sizeInBlocks;
```
qaisjp: could you split this up onto separate lines? something like this
```suggestion
    if (ppModelInfo[m_dwModelID] == nullptr)
        return false;

    auto sizeInBlocks = pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID)->sizeInBlocks;
    return sizeInBlocks > 0;
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1329,10 +1329,45 @@ void CModelInfoSA::MakePedModel(char* szTexture)
...
+    pTargetModelStreamigInfo->archiveId = pBaseModelStreamigInfo->archiveId;
+    pTargetModelStreamigInfo->offsetInBlocks = pBaseModelStreamigInfo->offsetInBlocks;
+    pTargetModelStreamigInfo->sizeInBlocks = pBaseModelStreamigInfo->sizeInBlocks;
+}
+
 void CModelInfoSA::DeallocateModel(void)
 {
     Remove();
     ppModelInfo[m_dwModelID] = nullptr;
+    CStreamingInfo* pModelStreamigInfo = pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID);
```
qaisjp: `Streamig` -> `Streaming`

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -577,8 +577,10 @@ int CLuaEngineDefs::EngineRequestModel(lua_State* luaVM)
                 eClientModelType eModelType;
                 if (strModelType == "ped")
                 {
-                    eModelType = CCLIENTMODELPED;
+                    eModelType = eClientModelType::PED;
                 }
+                else if (strModelType == "object")
+                    eModelType = eClientModelType::OBJECT;
```
qaisjp: Normally would ask to use a `ReadEnumString` but we can do that in a separate pull request. No changes necessary here.

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -1041,11 +1024,11 @@ void OnMY_CEntity_GetBoundRect(CEntitySAInterface* pEntity)
         if (!pColModel)
         {
             // Crash will occur at offset 00134134
-            CStreamingInfo* pStreamingInfo = GetStreamingInfoFromModelId(usModelId);
+            CStreamingInfo* pStreamingInfo = pGameInterface->GetStreaming()->GetStreamingInfoFromModelId(usModelId);
             SString         strDetails("refs:%d txd:%d RwObj:%08x bOwn:%d bColStr:%d flg:%d off:%d size:%d reqload:%d", pModelInfo->usNumberOfRefs,
```
qaisjp:
```suggestion
            SString         strDetails("refs:%d txd:%d RwObj:%08x bOwn:%d bColStr:%d flg:%d off:%d size:%d loadState:%d", pModelInfo->usNumberOfRefs,
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1329,10 +1329,45 @@ void CModelInfoSA::MakePedModel(char* szTexture)
...
+
+    pModelStreamigInfo->loadState = 0;
+    pModelStreamigInfo->nextInImg = -1;
+    pModelStreamigInfo->nextId = -1;
+    pModelStreamigInfo->prevId = -1;
+    pModelStreamigInfo->archiveId = 0;
+    pModelStreamigInfo->flg= 0;
+    pModelStreamigInfo->offsetInBlocks = 0;
+    pModelStreamigInfo->sizeInBlocks = 0;
```
qaisjp: Maybe you could extract this into `pModelStreamingInfo->Reset()` and then reuse it in `MakeObjectModel` here https://github.com/multitheftauto/mtasa-blue/pull/1684#discussion_r498348484 ?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1329,10 +1329,45 @@ void CModelInfoSA::MakePedModel(char* szTexture)
...
+    CStreamingInfo* pBaseModelStreamigInfo = pGame->GetStreaming()->GetStreamingInfoFromModelId(usBaseID);
+    CStreamingInfo* pTargetModelStreamigInfo = pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID);
+
+    pTargetModelStreamigInfo->loadState = 0;
+    pTargetModelStreamigInfo->nextInImg = -1;
+    pTargetModelStreamigInfo->nextId = -1;
+    pTargetModelStreamigInfo->prevId = -1;
+    pTargetModelStreamigInfo->archiveId = pBaseModelStreamigInfo->archiveId;
+    pTargetModelStreamigInfo->offsetInBlocks = pBaseModelStreamigInfo->offsetInBlocks;
+    pTargetModelStreamigInfo->sizeInBlocks = pBaseModelStreamigInfo->sizeInBlocks;
```
qaisjp: with this https://github.com/multitheftauto/mtasa-blue/pull/1684#discussion_r498348101 maybe this?

```suggestion
    pTargetModelStreamigInfo->Reset();
    pTargetModelStreamigInfo->archiveId = pBaseModelStreamigInfo->archiveId;
    pTargetModelStreamigInfo->offsetInBlocks = pBaseModelStreamigInfo->offsetInBlocks;
    pTargetModelStreamigInfo->sizeInBlocks = pBaseModelStreamigInfo->sizeInBlocks;
```

Client/mods/deathmatch/logic/CClientObjectManager.cpp
TheNormalnij: This file is used by serverside object manager.

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: `pModelStreamingInfo` is a struct. I can write `pGame->GetStreaming()->ResetModelStreamingInfo(m_dwModelID);`

TheNormalnij: `void CModelInfoSA::ResetStreamingInfo(void)` looks good to me

qaisjp: Happy with your approach!

> `pModelStreamingInfo` is a struct.

structs are just classes except members are public by default, so you should be able to add functions like that fine :)

Client/mods/deathmatch/logic/CClientObjectManager.cpp
```diff
@@ -92,15 +92,15 @@ CClientObject* CClientObjectManager::Get(ElementID ID)
...
-        if (pModelInfo && pModelInfo->GetInterface())
-        {
-            eModelInfoType eType = pModelInfo->GetModelType();
-            if (eType == eModelInfoType::CLUMP || eType == eModelInfoType::ATOMIC || eType == eModelInfoType::WEAPON)
-                return true;
-        }
+        eModelInfoType eType = pModelInfo->GetModelType();
+        if (eType == eModelInfoType::CLUMP || eType == eModelInfoType::ATOMIC || eType == eModelInfoType::WEAPON)
+            return true;
     }
```
qaisjp: What about these lines?

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Thx

Client/mods/deathmatch/logic/CClientObjectManager.cpp
TheNormalnij: done

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
TheNormalnij: Seems `ReadEnumString` don't work with `enum class`

Client/game_sa/CStreamingSA.cpp
```diff
@@ -133,4 +133,10 @@ void CStreamingSA::RequestSpecialModel(DWORD model, const char* szTexture, DWORD
...
     }
-}
\ No newline at end of file
+}
+
+CStreamingInfo* CStreamingSA::GetStreamingInfoFromModelId(ushort id)
+{
+    CStreamingInfo* pItemInfo = (CStreamingInfo*)(ARRAY_StreamModelInfo);
+    return &pItemInfo[id];
+}
```
ghost: There's a nice way to do this: 
```
// in CStreamingSA.h
class CStreamingSA
{
    public:
    static CStreamingInfo(&s_aInfoForModel)[26316];
};

// in CStreamingSA.cpp
// no need to use ARRAY_StreamModelInfo preprocessor
CStreamingInfo(&CStreamingSA::ms_aInfoForModel)[26316] = *(CStreamingInfo(*)[26316])0x8E4CC0;
```

Then you can use it just like a normal array:
```
CStreamingInfo* CStreamingSA::GetStreamingInfoFromModelId(ushort id)
{
    return &ms_aInfoForModel[id];
}
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1329,10 +1334,35 @@ void CModelInfoSA::MakePedModel(char* szTexture)
...
+}
+
 void CModelInfoSA::DeallocateModel(void)
 {
     Remove();
     ppModelInfo[m_dwModelID] = nullptr;
+    CStreamingInfo* pModelStreamingInfo = pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID);
+
+    pModelStreamingInfo->Reset();
```
Pirulax:
```cpp
pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID)->Reset();
```

Pirulax (changes requested):

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Looks clean

Client/game_sa/CStreamingSA.h
```diff
@@ -31,4 +29,8 @@ class CStreamingSA : public CStreaming
     void RequestSpecialModel(DWORD model, const char* szTexture, DWORD channel);
 
     CStreamingInfo* GetStreamingInfoFromModelId(ushort id);
+
+private:
+    static CStreamingInfo (&s_aInfoForModel)[26316];
+    CStreamingInfo (&CStreamingSA::ms_aInfoForModel)[26316] = *(CStreamingInfo(*)[26316])0x8E4CC0;
```
tederis: It's better to move this line to the **cpp**. H headers are just for declaration. We don't want to initialize the variable in the every translation unit.

tederis: Line 35 I mean

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -516,7 +516,12 @@ bool CModelInfoSA::IsValid()
 {
     if (m_dwModelID >= 20000 && m_dwModelID < MODELINFO_MAX)
         return true;
-    return ppModelInfo[m_dwModelID] != 0;
+
+    if (ppModelInfo[m_dwModelID] == nullptr)
```
Pirulax:
```cpp
if (!ppModelInfo[m_dwModelID])
```

Pirulax (changes requested):

PlatinMTA:
I have been able to find some bugs:

1) The client will crash if a resource containing an object with a requested ID stops.

```lua
-- some_resource/client.lua

function start()
	local id = engineRequestModel("object")
	print(id)
	createObject(id, 0, 0, 5)
end
addEventHandler("onClientResourceStart", resourceRoot, start)
```

If you /stop some_resource then the game will crash. Destroying it on "onClientResourceStop" event doesn't fix it either. If the object is destroyed before the resource stops, the game will not crash.

~~I think~~ The client also crashes when disconnecting and a custom object is still created. Destroying the object before disconnecting does not crash the game.

---

2) If we reeplace the model with engineReplaceCOL/engineImportTXD/engineReplaceModel, all the objects created in different resources and before this model change was made won't obtain the new model. Futhermore, if we restart the resource that reeplaces the model, all the objects that had the new model will go back to looking like 1337.

This are the model files I used to test: [archivos.zip](https://github.com/multitheftauto/mtasa-blue/files/5362353/archivos.zip)

```lua
-- some_skin_replacer/client.lua

-- This will reeplace the model just fine
function start()
	local col, txd, dff;
	local id = engineRequestModel("object")

	col = engineLoadCOL("archivos/locker.col")
	engineReplaceCOL(col, id)
	txd = engineLoadTXD("archivos/locker.txd")
	engineImportTXD(txd, id)
	dff = engineLoadDFF("archivos/locker.dff")
	engineReplaceModel(dff, id)

	obj = createObject(id, 0, 0, 5)
end
addEventHandler("onClientResourceStart", resourceRoot, start)

function destroyObj()
	destroyElement(obj)
end
addCommandHandler("destroyobj", destroyObj, false, false)
```

Run ``/crun createObject(3, 0, 0, 3)``, use /destroyobj (or else you will crash) and then restart the script

<details>

  <summary>This is what will happen</summary>


</details>

---

3) I don't know this is in intended, but the first ID the function gives is 3, not 321 as the pastebin suggested. Maybe this is what is making all this crashes occur.

---

A thing I should point out is that I wasn't testing on the debug build, i tested this with the nightly preset. I point this out just in case you can't reproduce it.

I hope this isn't to hard to fix, I'll really like to use this feature as soon as possible. Thank you for your work.

TheNormalnij:
Thanks you for help, @PlatinMTA !
1. Crash is fixed.
2. I can't reproduce.
3. It's a list with all default object ID's, engineRequestModel will select another ID's.

Inder00 (approved):

TheNormalnij:
Added test resource

Inder00 (commented):
It looks like good

Client/sdk/game/CStreaming.h
```diff
@@ -11,12 +11,37 @@
...
+struct CStreamingInfo
+{
+    WORD  prevId;
+    WORD  nextId;
+    WORD  nextInImg;
+    BYTE flg;
+    BYTE archiveId;
+    DWORD offsetInBlocks;
+    DWORD sizeInBlocks;
+    DWORD loadState;
```
qaisjp: I think we discourage using `WORD` `BYTE` `DWORD` etc now, and prefer proper types etc

Client/sdk/game/CStreaming.h
```diff
@@ -11,12 +11,37 @@
...
+
 class CStreaming
 {
 public:
     virtual void RequestModel(DWORD dwModelID, DWORD dwFlags) = 0;
     virtual void LoadAllRequestedModels(BOOL bOnlyPriorityModels = 0, const char* szTag = NULL) = 0;
     virtual BOOL HasModelLoaded(DWORD dwModelID) = 0;
     virtual void RequestSpecialModel(DWORD model, const char* szTexture, DWORD channel) = 0;
+    virtual CStreamingInfo* GetStreamingInfoFromModelId(unsigned short id) = 0;
     virtual void ReinitStreaming() = 0;
```
qaisjp: If you run this through the formatter it will look like this:

```
virtual void            Blah0() = 0;
virtual CStreamingInfo* Blah1() = 0;
virtual void            Blah2() = 0;
```

So you should either do this:

```suggestion
    virtual void RequestModel(DWORD dwModelID, DWORD dwFlags) = 0;
    virtual void LoadAllRequestedModels(BOOL bOnlyPriorityModels = 0, const char* szTag = NULL) = 0;
    virtual BOOL HasModelLoaded(DWORD dwModelID) = 0;
    virtual void RequestSpecialModel(DWORD model, const char* szTexture, DWORD channel) = 0;

    virtual CStreamingInfo* GetStreamingInfoFromModelId(unsigned short id) = 0;

    virtual void ReinitStreaming() = 0;
```

Or:


```suggestion
    virtual void            RequestModel(DWORD dwModelID, DWORD dwFlags) = 0;
    virtual void            LoadAllRequestedModels(BOOL bOnlyPriorityModels = 0, const char* szTag = NULL) = 0;
    virtual BOOL            HasModelLoaded(DWORD dwModelID) = 0;
    virtual void            RequestSpecialModel(DWORD model, const char* szTexture, DWORD channel) = 0;
    virtual CStreamingInfo* GetStreamingInfoFromModelId(unsigned short id) = 0;
    virtual void            ReinitStreaming() = 0;
```

Client/sdk/game/CModelInfo.h
```diff
@@ -28,6 +28,17 @@ class CBoundingBox
     float   fRadius;
 };
 
+enum class eModelInfoType : unsigned char
+{
+    ATOMIC = 1,
+    TIME = 3,
```
qaisjp, line 34: Huh interesting, does gta skip over` = 2`?

Client/sdk/game/CStreaming.h
TheNormalnij: Fixed

Client/sdk/game/CModelInfo.h
TheNormalnij: GTA SA don't use model info with enum 2

TheNormalnij: Or this is special type for parts of destroyed car

TheNormalnij:
Fixed formatting

Client/sdk/game/CStreaming.h
Pirulax: Perhaps uint16_t, etc.. should be used? Since, remember, `unsigned int` may be 1 byte on a compiler, since its not standardized. The only thing thats guaranteed is that sizeof(int) >= sizeof(char)

TheNormalnij: It's dependents from processor architecture. For 16 bits `int` will have size 2 bytes, for 32 bits - 4 bytes. MTA client compiled for 32 bit architecture always.

Pirulax: Well, `int` was a bad example, but the only one applicable here.
A better example would be `long` which, depending on the memory model used in the compiler, is either 32 or 64 bits. On MSVC, which uses LLP64, its 32, but iirc on GCC it's 64, because its not using LLP64.

TheNormalnij: Hmm... Changed to uint8, uint16, uint32

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -53,20 +63,45 @@ bool CClientModel::Deallocate(void)
...
+                    }
+                    pPed->SetModel(0);
+
+                    CLuaArguments Arguments;
+                    Arguments.PushNumber(m_iModelID);
+                    Arguments.PushNumber(0);
+                    pPed->CallEvent("onClientElementModelChange", Arguments, true);
+                }
+            }
+        }
```
ghost, line 87: Shorter methods are very easy to read. You can split this code into a function:
```
void CClientPedManager::triggerElementModelChangeEvent(unsigned short modelId)
{
    for (auto& ped : m_List)
    {
        if (ped->GetModel() == modelId)
        {
            if (ped->IsStreamedIn())
                ped->StreamOutForABit();
            ped->SetModel(0);
            CLuaArguments Arguments;
            Arguments.PushNumber(modelId);
            Arguments.PushNumber(0);
            ped->CallEvent("onClientElementModelChange", Arguments, true);
        }
    }
}
```
```
g_pClientGame->GetManager()->GetPedManager()->triggerElementModelChangeEvent(m_iModelID);
```

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -53,20 +63,45 @@ bool CClientModel::Deallocate(void)
...
+                        pObject->StreamOutForABit();
                     }
-                    (*iter)->SetModel(0);
+                    pObject->SetModel(1337);
+
+                    CLuaArguments Arguments;
+                    Arguments.PushNumber(m_iModelID);
+                    Arguments.PushNumber(1337);
+                    pObject->CallEvent("onClientElementModelChange", Arguments, true);
                 }
```
ghost, line 105: Create a function for this, please.

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -791,6 +791,8 @@ namespace SharedUtil
         // list only
         typename LIST_TYPE ::iterator         begin() { return m_List.begin(); }
         typename LIST_TYPE ::iterator         end() { return m_List.end(); }
+        typename LIST_TYPE ::const_iterator   begin() const { return m_List.begin(); }
+        typename LIST_TYPE ::const_iterator   end() const { return m_List.end(); }
```
ghost, line 795: This is not needed.

Client/mods/deathmatch/logic/CClientObjectManager.h
```diff
@@ -25,6 +25,7 @@ class CClientObjectManager
 
     void DeleteAll();
 
+    const auto&           GetObjects() const { return m_Objects; };
```
ghost:
```suggestion
    const CMappedArray<CClientObject*>& GetObjects() const { return m_Objects; };
```
The return type is `CMappedArray<CClientObject*>&` It doesn't take that much space, I don't think the use of auto keyword is justified here.

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -37,7 +37,17 @@ bool CClientModel::Allocate(void)
...
+        {
+            case eClientModelType::PED:
+                pModelInfo->MakePedModel("PSYCHO");
+                break;
+            case eClientModelType::OBJECT:
+                pModelInfo->MakeObjectModel(1337);
+                break;
+            default:
+                return false;
+        }
```
ghost, line 50:
```suggestion
        pModelInfo->MakeModel(m_eModelType);
```
Split this code into a new function so the method becomes shorter.

TheNormalnij: I think i will add more arguments for MakeObjectModel and MakeVehicleModel and MakeTimedObjectModel later

Pirulax:
@saml1er https://github.com/multitheftauto/mtasa-blue/pull/1684#discussion_r515656592
It is needed, otherwise his `range based for loop` woudn't work (with CMappedArray)

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: This part of code use lambda function in #1701

ghost:
> @saml1er [#1684 (comment)](https://github.com/multitheftauto/mtasa-blue/pull/1684#discussion_r515656592)
> It is needed, otherwise his `range based for loop` woudn't work (with CMappedArray)

~~In my other review comment, I suggested him to move the code in CClientPedManager. In this way, the method will become shorter, readable, and this won't be needed.~~
Hmm, I think I get it now why he added it there. Okay.

Client/mods/deathmatch/logic/CClientModel.cpp
ghost: Oh, I just realized the enum is `eClientModelType`. In this case, you can create the function in CClientModel. 

```
bool CClientModel::MakeModel()
{
    switch (m_eModelType)
    {
        case eClientModelType::PED:
            pModelInfo->MakePedModel("PSYCHO");
            break;
        case eClientModelType::OBJECT:
            pModelInfo->MakeObjectModel(1337);
            break;
        default:
            return false;
    }
    return true;
}
```
> I think i will add more arguments for MakeObjectModel and MakeVehicleModel and MakeTimedObjectModel later

Sure, but please don't forget to split the code.

TheNormalnij: Yes, but when we define API for `engineRequestModel`

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: also from another [PR](https://github.com/multitheftauto/mtasa-blue/pull/1701/commits/efb87d85221f90f78aafc60c6c09d564ec887dca#diff-73eeec935449ae07c6300011119ba1692ac7448a961d7a2498abe8ed142e4343R77)

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1365,6 +1370,17 @@ void CModelInfoSA::SetVoice(const char* szVoiceType, const char* szVoice)
     SetVoice(sVoiceType, sVoiceID);
 }
 
+void CModelInfoSA::CopyStreamingInfoFromModel(ushort usBaseModelID)
+{
+    CStreamingInfo* pBaseModelStreamigInfo = pGame->GetStreaming()->GetStreamingInfoFromModelId(usBaseModelID);
```
qaisjp: "Streamig" typo

ghost (approved):

PR #10, opened 2015-08-23
New server side event: onPlayerTeamChange

Citizen01:
This is my first pull-request and I added the onPlayerTeamChange server side event using the player as source and the handler arguments oldTeam, newTeam.

Btw, if the previous team was null, I'm still sending that value as is. So I don't know if it's an ok way to do.

Just got the idea because of a Scripting thread I replied with an homemade onPlayerTeamChange with a setTimer to check the players team because the guy needed that event.

source: The player who changed team
oldTeam: The previous player's team if any, nil otherwise
newTeam: The new player's team

PR #1687, opened 2020-09-30
New team event onPlayerTeamChange(previous, current, reason)

### LINKED ISSUE #1686: onPlayerTeamJoin/Leave & onTeamJoin/Leave
**Description**
Basically, as I mentioned on the topic title. Events when you join or leave a team.

I'm quite surprised that no one requested these events so far, so basically what I have in mind is:

onPlayerTeamJoin(team), onPlayerTeamLeave(team)
onTeamJoin(player), onTeamLeave(player)

**Additional context**
I think these events are really important because right now there's no way proper to detect if the player joined or left a team. I would like to try to implement them by myself, I still need more opinions about these names, feel free to correct them.

N3xT:
This is my first pull request, I would like to hear some feedback!

Related to #1686

### DIFF
--- a/Server/mods/deathmatch/logic/CElementDeleter.cpp
+++ b/Server/mods/deathmatch/logic/CElementDeleter.cpp
@@ -19,6 +19,12 @@ void CElementDeleter::Delete(class CElement* pElement, bool bUnlink, bool bUpdat
     {
         if (!IsBeingDeleted(pElement))
         {
+            // Check if the element is a team and remove players
+            if (IS_TEAM(pElement))
+            {
+               ((CTeam*)pElement)->RemoveAllPlayers();
+            }
+            
             // Before we do anything, fire the on-destroy event
             CLuaArguments Arguments;
             pElement->CallEvent("onElementDestroy", Arguments);
--- a/Server/mods/deathmatch/logic/CElementDeleter.cpp
+++ b/Server/mods/deathmatch/logic/CElementDeleter.cpp
@@ -18,17 +18,17 @@ void CElementDeleter::Delete(class CElement* pElement, bool bUnlink, bool bUpdat
     if (pElement)
     {
         if (!IsBeingDeleted(pElement))
-        {
+        {            
+            // Before we do anything, fire the on-destroy event
+            CLuaArguments Arguments;
+            pElement->CallEvent("onElementDestroy", Arguments);
+
             // Check if the element is a team and remove players
             if (IS_TEAM(pElement))
             {
                ((CTeam*)pElement)->RemoveAllPlayers();
             }
             
-            // Before we do anything, fire the on-destroy event
-            CLuaArguments Arguments;
-            pElement->CallEvent("onElementDestroy", Arguments);
-
             // Add it to our list
             if (!pElement->IsBeingDeleted())
             {
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1518,6 +1518,8 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerNetworkStatus", "type, ticks", NULL, false);
     m_Events.AddEvent("onPlayerScreenShot", "resource, status, file_data, timestamp, tag", NULL, false);
     m_Events.AddEvent("onPlayerDiscordJoin", "justConnected, secret", NULL, false);
+    m_Events.AddEvent("onPlayerTeamJoin", "team", NULL, false);
+    m_Events.AddEvent("onPlayerTeamLeave", "team", NULL, false);
 
     // Ped events
     m_Events.AddEvent("onPedWasted", "ammo, killer, weapon, bodypart", NULL, false);
@@ -1570,6 +1572,10 @@ void CGame::AddBuiltInEvents()
     // Weapon events
     m_Events.AddEvent("onWeaponFire", "", NULL, false);
     m_Events.AddEvent("onPlayerWeaponFire", "weapon, endX, endY, endZ, hitElement, startX, startY, startZ", NULL, false);
+
+    // Team events
+    m_Events.AddEvent("onTeamJoin", "player", NULL, false);
+    m_Events.AddEvent("onTeamLeave", "player", NULL, false);
 }
 
 void CGame::ProcessTrafficLights(long long llCurrentTime)
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1410,6 +1410,16 @@ void CGame::QuitPlayer(CPlayer& Player, CClient::eQuitReasons Reason, bool bSayI
     // If he had joined
     if (Player.IsJoined())
     {
+        // Check if the player was inside a team and fire the event
+        if (Player.GetTeam())
+        {
+            CLuaArguments Arguments;
+            Arguments.PushElement(Player.GetTeam());
+            Arguments.PushNil();
+            Arguments.PushString("disconnected");
+            Player.CallEvent("onPlayerTeamChange", Arguments);
+        }
+
         // Tell our scripts the player has quit, but only if the scripts got told he joined
         CLuaArguments Arguments;
         Arguments.PushString(szReason);
@@ -1518,8 +1528,7 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerNetworkStatus", "type, ticks", NULL, false);
     m_Events.AddEvent("onPlayerScreenShot", "resource, status, file_data, timestamp, tag", NULL, false);
     m_Events.AddEvent("onPlayerDiscordJoin", "justConnected, secret", NULL, false);
-    m_Events.AddEvent("onPlayerTeamJoin", "team", NULL, false);
-    m_Events.AddEvent("onPlayerTeamLeave", "team", NULL, false);
+    m_Events.AddEvent("onPlayerTeamChange", "previous, current, reason", NULL, false);
 
     // Ped events
     m_Events.AddEvent("onPedWasted", "ammo, killer, weapon, bodypart", NULL, false);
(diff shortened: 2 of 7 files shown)


Citizen01:
OH ! Someone is finally trying to add those events in MTA !
I did try too, back in 2015 and lost motivation. I hate not finishing the stuff I start but that's always what happens with me (at least if I have no deadline/pressure). Have a look at my PR #10 **onPlayerTeamChange**.

So yeah I didn't finish it but I can at least point out the problems and edge cases the staff told me on my PR:
- What if the player disconnects ? (it should trigger your Leave events)
- What if you delete/destroy the team in which a set of players are in ? (it should trigger your Leave events for that set of player)
- Can you make your events so they are cancelable ? (first version doesn't need it so don't do it just because I asked if it could)
- What happens if you call **setElementTeam** inside your events ? (a stack overflow is to be expected but iirc it's okay and it's the responsability of the scripters to not do that)

Also, the main benefit of creating an **onPlayerTeamChange** event instead of a **Join** and **Leave** events is that you can listen to that one event only to track **Join** and **Leave** events. (Like onElementDataChange instead of **onElementDataSet** and **onElementDataUnset**)

N3xT:
@Citizen01 
These are really good points, I'm really glad you pointed them out.

> * What if the player disconnects ? (it should trigger your Leave events)

Makes sense. I looked around but I couldn't figure out where to implement this, so I need help here.

> * What if you delete/destroy the team in which a set of players are in ? (it should trigger your Leave events for that set of player)

Hmm, what I have in mind is to use **onElementDestroy** to get the players when a team is destroyed. Technically they didn't leave the team by their choice, so yeah, I'm not sure if I should fire the event in this case or not.

> * Can you make your events so they are cancelable ? (first version doesn't need it so don't do it just because I asked if it could)

Good point.

> * What happens if you call setElementTeam inside your events ? (a stack overflow is to be expected but iirc it's okay and it's the responsability of the scripters to not do that)

Yeah, I guess the same scenario you mentioned above will occur. But it's the scripter responsibility to prevent that and code wisely.

PlatinMTA:
> * What if you delete/destroy the team in which a set of players are in ? (it should trigger your Leave events for that set of player)

Maybe you should fire the onTeamChange event BUT add an extra argument letting the function know if it was forced to quit the team or not.

N3xT:
Okay, so we would have something like this **onPlayerTeamChange(previous, current, forced)**?

N3xT:
I updated my pull request and tested it, so far so good. 

### onPlayerTeamChange(previous, current, reason)
* source is the player whose team changed
* previous is your previous team, nil otherwise
* current is your current team, nil otherwise
* reason is the reason caused your team to change, this can be **(changed, destroyed, disconnected)**

```
addEventHandler("onResourceStart", resourceRoot, 
    function()
        red = Team("Red", 255, 0, 0)
        blue = Team("Blue", 0, 0, 255)
    end
)

addEventHandler("onPlayerTeamChange", root,
    function(previous, current, reason)
        iprint(source.name, previous, current, reason)
    end
)
```
Waiting for a review.

lex128:
Server-side can be controlled by using hooks. It's better to implement such events client-side.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1410,6 +1410,16 @@ void CGame::QuitPlayer(CPlayer& Player, CClient::eQuitReasons Reason, bool bSayI
     // If he had joined
     if (Player.IsJoined())
     {
+        // Check if the player was inside a team and fire the event
```
Citizen01: I'd name the event in the comment to remove the ambiguity
```suggestion
        // Check if the player was inside a team and fire onPlayerTeamChange event
```

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1410,6 +1410,16 @@ void CGame::QuitPlayer(CPlayer& Player, CClient::eQuitReasons Reason, bool bSayI
...
+        if (Player.GetTeam())
```
Citizen01, line 1414: Use a variable to get and store the team just before your if statement instead of calling GetTeam twice.

Server/mods/deathmatch/logic/CElementDeleter.cpp
```diff
@@ -19,6 +19,12 @@ void CElementDeleter::Delete(class CElement* pElement, bool bUnlink, bool bUpdat
     {
         if (!IsBeingDeleted(pElement))
         {
+            // Check if the element is a team and remove players
+            if (IS_TEAM(pElement))
+            {
+               ((CTeam*)pElement)->RemoveAllPlayers();
+            }
```
Citizen01, line 30: Put it **after** the trigger of `onElementDestroy` to make sure we don't brake scripts that needs the list of the players at that specific timing.

Citizen01 (changes requested):

Server/mods/deathmatch/logic/CPlayer.cpp
```diff
@@ -512,11 +512,17 @@ unsigned char CPlayer::GetAttackBodyPart()
     return m_ucAttackBodyPart;
 }
 
-void CPlayer::SetTeam(CTeam* pTeam, bool bChangeTeam)
+void CPlayer::SetTeam(CTeam* pTeam, bool bChangeTeam, const char* szReason)
```
Citizen01: use an enum instead for the reason

N3xT:
Done, I also added a new change reason:

* source is the player whose team changed
* previous is your previous team, nil otherwise
* current is your current team, nil otherwise
* reason is the reason caused your team to change, this can be (changed, removed, disconnected, destroyed)

You'll get **removed** as a reason if you were removed from a team, and **changed** when you were moved from a team to a different one.

Server/mods/deathmatch/logic/CTeam.h
```diff
@@ -23,6 +23,15 @@ class CTeam : public CElement
...
 
+public:
+    enum eChangeReason
+    {
+        CHANGE_CHANGED,
+        CHANGE_REMOVED,
+        CHANGE_DISCONNECTED,
+        CHANGE_DESTROYED,
+    };
+
```
Pirulax, line 34: Use `enum class`:
```cpp
enum ChangeReason
{
    CHANGED,
    REMOVED,
    DISCONNECTED,
    DESTROYED,
};
```
(You will need to change your code where you use this enum.)

Also, I do suggest you not to use the `e` prefix (hungarian notation bs).

Server/mods/deathmatch/logic/CTeam.cpp
```diff
@@ -109,7 +109,7 @@ void CTeam::RemoveAllPlayers()
     list<CPlayer*>::const_iterator iter = m_Players.begin();
     for (; iter != m_Players.end(); ++iter)
     {
-        (*iter)->SetTeam(NULL, false);
+        (*iter)->SetTeam(NULL, false, CHANGE_DESTROYED);
     }
```
Pirulax, line 113: I suggest to refactor this code to:
```cpp
for (CPlayer* player : m_Players)
    player->SetTeam(NULL, false, CHANGE_DESTROYED);
```

Server/mods/deathmatch/logic/CPlayer.cpp
```diff
@@ -512,11 +512,39 @@ unsigned char CPlayer::GetAttackBodyPart()
...
+        case CTeam::CHANGE_REMOVED:
+            szReason = "removed";
+            break;
+        case CTeam::CHANGE_DISCONNECTED:
+            szReason = "disconnected";
+            break;
+        case CTeam::CHANGE_DESTROYED:
+            szReason = "destroyed";
+            break;
+    }
```
Pirulax, line 540: I'd implement `EnumToString` for `CTeam::ChangeReason` instead, and use that, as its more universal.
Altough, im not sure where you should add the `IMPLEMENT_ENUM` things for it, I'd say just stash it into CTeams.cpp (and h) somewhere. (See CLuaFunctionParseHelpers.cpp for some example code)

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1410,6 +1410,12 @@ void CGame::QuitPlayer(CPlayer& Player, CClient::eQuitReasons Reason, bool bSayI
     // If he had joined
     if (Player.IsJoined())
     {
+        // Check if the player was inside a team so we could fire onPlayerTeamChange event
+        if (Player.GetTeam())
```
Pirulax, line 1414: As suggested, use a variable instead:
```cpp
if (CTeam* pPlayerTeam = Player.GetTeam())
```

Server/mods/deathmatch/logic/CElementDeleter.cpp
```diff
@@ -18,11 +18,17 @@ void CElementDeleter::Delete(class CElement* pElement, bool bUnlink, bool bUpdat
...
-        {
+        {            
             // Before we do anything, fire the on-destroy event
             CLuaArguments Arguments;
             pElement->CallEvent("onElementDestroy", Arguments);
 
+            // Check if the element is a team and remove players
+            if (IS_TEAM(pElement))
+            {
+               ((CTeam*)pElement)->RemoveAllPlayers();
```
Pirulax, line 29: Maybe a `static_cast` would look better(and be more proper) here:
`static_cast<CTeam*>(pElement)->RemoveAllPlayers())`.

Pirulax (changes requested):
Your PR imho is fine, but a little bit messy, which is understandable, as the whole MTA code is messy, but don't follow the bad practices :)
If you have any questions, please feel free to ask me @ dc.

Citizen01:
> You'll get **removed** as a reason if you were removed from a team

Then why not adding **added** for when you get added to a team ?

I don't think it's a good idea anyway, the **changed** is enough to cover:

**no-team -> Team A**
**Team A -> Team B**
and
**Team B -> no-team**

There is no reason for the latter case to get a specific "removed" reason.
If anyone thinks it's better to distinguish all 3 cases with their own reason, I'm okay to listen for pros about it and to change my mind.

Server/mods/deathmatch/logic/CGame.cpp
N3xT: I called ``Player.GetTeam()`` twice on my previous pull request, which I changed now. So, I don't think I should declare a variable for a one time use no?

P.S: Sorry for taking too long, will work on this as soon as possible.

```
// Check if the player was inside a team so we could fire onPlayerTeamChange event
if (Player.GetTeam())
{
    Player.SetTeam(NULL, false, CTeam::CHANGE_DISCONNECTED);
}
```

LosFaul:
instead of reason maybe source resource or something would make more sense

for example you have team management resource A but resource B use setPlayerTeam
so you could cancel the event if the source resource is resource B
or you could whitelist resource B to allow changing a players team...

LosFaul:
maybe it's best to limit it for the initial pull request only to previousTeam and currentTeam
and let something like reason be a topic for an additional PR

Issue #1683, opened 2020-09-28
build r20653 crashing randomly without dump


brzys:
**Describe the bug**
MTA just closes in a random moment while playing or connecting.
Crashes are happening from build r20469, tested by my server staff.

**To reproduce**
Can't reproduce. Happening on my server with +100 players, randomly crashing 10-20 players **at the same moment**

**Screenshots**
https://streamable.com/jwr0y1
https://www.youtube.com/watch?v=_uE_IvEFS0Q
![image](https://user-images.githubusercontent.com/5489020/94474910-a5b62300-01ce-11eb-9c7e-2505306dceb6.png)
*(all players with Timed Out have 20653! No problem for players with 20632.)*

**Version**
Multi Theft Auto v1.5.7-release-20653 

**Additional context**
Would like to help with something more, but what I can do without dumps?
Unfortunately my server has too much resources to debug something by myself.

ecastro98:
Can confirm this. It's also happening in the server I admin.

![f](https://user-images.githubusercontent.com/11150785/94488411-f28eff00-01b8-11eb-9635-cf5adf7bb16f.png)
![20200928_183203](https://user-images.githubusercontent.com/11150785/94488413-f3c02c00-01b8-11eb-9360-932cda6202a0.png)

The images are from today but this issue started days ago.

qaisjp:
https://buildinfo.mtasa.com/?Revision=20649&Branch=

One of these is the culprit: 97dbfc6be01120fb635463c5f60601ecbe5aa195 e30f2ec6121c71408ab5efba3d1a1239ea75253e 27c6cc404a28f2af8b47271f6044486857510742

ecastro98:
@qaisjp How does the revert work? Do we have to wait for the next RC?

Dutchman101:
> How does the revert work? Do we have to wait for the next RC?

Just use the latest nightly, either from [http://nightly.mtasa.com/](http://nightly.mtasa.com/) or by using the "nightly" updater channel in MTA advanced settings.

But since this is a random, non-reproducible issue it's more complex to tell if it will be fixed just by that. So, we will do something to bring this build to more users (e.g with a bugged build) and then you should be able to find out by watching the trends. 

[...]

PR #1693, opened 2020-10-01
Potential fix for #1683

ghost:
The marker colshape is no longer attached to the marker, instead the hit point is updated. At least, this is how it's done in pickup code. It seems to work for pickups, so it shouldn't cause any problem for markers either. I tested it, and the marker colshape hit events are triggered properly.

EDIT: yes, the commit history is messed up. Fixing it.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientMarker.cpp
+++ b/Client/mods/deathmatch/logic/CClientMarker.cpp
@@ -47,10 +47,7 @@ CClientMarker::~CClientMarker()
     // Make sure nothing is still referencing us
     m_pManager->UnreferenceEntity(this);
 
-    // Remove the colshape
-    if (m_pCollision)
-        delete m_pCollision;
-    m_pCollision = nullptr;
+    SAFE_DELETE(m_pCollision)
 
     // Stream out first so the element counter is correct
     StreamOut();
--- a/Client/mods/deathmatch/logic/CClientMarker.cpp
+++ b/Client/mods/deathmatch/logic/CClientMarker.cpp
@@ -39,6 +39,8 @@ CClientMarker::CClientMarker(CClientManager* pManager, ElementID ID, int iMarker
 
 CClientMarker::~CClientMarker()
 {
+    AttachTo(nullptr);
+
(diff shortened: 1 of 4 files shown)


Dutchman101:
Nice.. i hope this will be it.

But do you have any idea why a crash like this would be unhandled (no dump, the client just goes poof)? It might be worth looking into that on the sidelines.

ghost:
> But do you have any idea why a crash like this would be unhandled (no dump, the client just goes poof)? It might be worth looking into that on the sidelines.

This is an old issue. I have a theory: If an exception occurs in the exception handler code, then the dump won't be generated. If could somehow reproduce a crash where there's no dump file, then I might be able to track the bug.

Client/mods/deathmatch/logic/CClientMarker.cpp
```diff
@@ -48,6 +50,7 @@ CClientMarker::~CClientMarker()
     // Remove the colshape
     if (m_pCollision)
         delete m_pCollision;
+    m_pCollision = nullptr;
```
qaisjp:
```suggestion
    SAFE_DELETE(m_pCollision);
```

qaisjp (approved):

Pirulax:
> > But do you have any idea why a crash like this would be unhandled (no dump, the client just goes poof)? It might be worth looking into that on the sidelines.
> 
> This is an old issue. I have a theory: If an exception occurs in the exception handler code, then the dump won't be generated. If could somehow reproduce a crash where there's no dump file, then I might be able to track the bug.

We use google breakpad for crash exception handling, dont we? Maybe we should just update it.

Edit: I've looked it up on google("google breakpad no crashdump") and found [this](https://groups.google.com/a/chromium.org/g/chromium-discuss/c/3J69EcSz1Xo?pli=1). They basically say, if its an OOM(out of mem) issue the crash handler itself may not have enough address space to mmap the things it needs. Knowing how much memory MTA uses on a regular basis(2-3gb), i see this as a possibility. 

Edit2: People there suggest to use `crashpad` instead.

PR #1699, opened 2020-10-07
Add fixed lights and stencil shadows

TFP-dev:
Hi, I have added small function to lower stencil shadows and lights.
mta0
mta5
mta1
mta2
mta3
mta4


Einheit-101:
This is basically what dr_blendshad resource does, but different :D 
Good improvement.

ccw808:
Do the shadows flicker when:
1) Using a high value in setFarClipDistance?
2) In 16 bit colour mode?

Client/game_sa/CSettingsSA.cpp
```diff
@@ -327,7 +328,22 @@ void CSettingsSA::SetVolumetricShadowsSuspended(bool bSuspended)
...
+}
+
+void CSettingsSA::SetLightsAndShadowsImprovementEnabled(bool bEnable)
+{
+    m_bLightsAndShadowsImprovement = bEnable;
+
+    if (bEnable == true) 
+        MemPut<DWORD>(0x00859934, 1018127695);
+    else
+        MemPut<DWORD>(0x00859934, 1031127695);
```
Pirulax, line 344:
```cpp
m_bLightsAndShadowsImprovement = bEnable;
MemPut<float>(0x00859934, bEnable ? 0.0214106123894 : 0.0599999986589);
```
Also I see in IDA this is a float value. Im pretty sure you could just use `MemPut<float>`, or?

Edit: Edited my code to use floats.

Woovie:
setFarClipDistance 100000, 32bit
No perceived flicker
image

I cannot get a 16bit color function to work in Visual Studio, not sure what I need to do for it.

Note that you will see the reflection still looks high up here, I noticed when changing my far clip, it slowly rises. 10000 had barely a perceivable difference. 100000 as you can tell is much more obvious. 1000000 has such a huge offset, the reflection is beyond the curb.

patrikjuvonen:
Thanks for the pull request, we've made further iterations to this and committed at e51c8748eda57308e6a856fb4986a1fd1ac158a3

You were marked as co-author.

Issue #1175, opened 2019-12-04
setModelName/setVehicleName/getModelName


Allerek:
Hi, as you know current actually we are forced to use original models/vehicles names in GTA(we can do the table with ours but thats another thing)

I think it would be cool to add
setVehicleName(_int_ model, _string_ name) returns **true/false** (to use with getVehicleNameFromModel/getVehicleName)
setModelName(_int_ model_id, _string_ name) returns **true/false**
getModelName(_int_ model_id) returns _string_ **model name**

I dont think anyone needs explanation how this should work, the functions names are pretty clear.

qaisjp:
Thanks for your proposal. Why do you think this is necessary / should be added?

Theoretically I guess it makes sense that it should be possible to change this metadata, but I'm worried that this would just be a feature that causes confusion.

Allerek:
> Thanks for your proposal. Why do you think this is necessary / should be added?
> 
> Theoretically I guess it makes sense that it should be possible to change this metadata, but I'm worried that this would just be a feature that causes confusion.

F.E. when you want to add car thats out of the GTA SA(i have Infernus from VC on my servers which is replaced for 'Saddle Shit' and named it 'Infernus SX') so it would be easier when you need to use getVehicleName[...].
2.When you want to create some Map Editors by yourself without huuuge list of items/objects/vehicles etc.

Einheit-101:
Map editor uses predefined model names that are listed in
server\mods\deathmatch\resources\[editor]\editor_gui\client\browser\objects.xml

But yeah, its not a totally bad idea to have custom vehicle names added - for most users this resource should be sufficient tho:
https://community.mtasa.com/index.php?p=resources&s=details&id=12424

Allerek:
> Map editor uses predefined model names that are listed in
> server\mods\deathmatch\resources[editor]\editor_gui\client\browser\objects.xml
> 
> But yeah, its not a totally bad idea to have custom vehicle names added - for most users this resource should be sufficient tho:
> https://community.mtasa.com/index.php?p=resources&s=details&id=12424

This one uses a table with names, i want to make it a hard-coded function

###### admin note: put an extra newline after the quote is done otherwise it will treat it as the same line

Pirulax:
Well, changing the model name may break scripts using getModelFromName(or whatever the name of the func. is), but i think we could do a set/getVehicleCustomName or something similar.

MrDadosz:
[...]

PR #1701, opened 2020-10-08
Allow allocating new vehicles models using engineRequestModel 

TheNormalnij:
New API:
```lua
     interger iModelID = engineRequestModel( "vehicle", integer parentID ) 
```

From vehicle with  `parentID` model new vehicle model will have:
* Model (DFF and TXD)
* Vehicle type.
* Sounds. This will include engine sounds, door sounds, slip sounds,, etc.
* Handling. Specific flying/boat/bike handling values also.
* Anim group.
* Default colors.

Does not work:
* Сhassis
* Turrets
* ZR-350 opening headlights
* Rhino, Hydra, Skimmer, Seasparrow
* Some specific effects like a dust effect from some helicopters, smoke effects from bikes.
* Another hardcoded stuff

### DIFF
--- a/Client/game_sa/CBikeHandlingEntrySA.cpp
+++ b/Client/game_sa/CBikeHandlingEntrySA.cpp
@@ -14,25 +14,15 @@
 CBikeHandlingEntrySA::CBikeHandlingEntrySA()
 {
     // Create a new interface and zero it
-    m_pBikeHandlingSA = new tBikeHandlingDataSA;
-    memset(m_pBikeHandlingSA, 0, sizeof(tBikeHandlingDataSA));
-    m_bDeleteInterface = true;
+    m_pBikeHandlingSA = std::make_shared<tBikeHandlingDataSA>();
+    memset(m_pBikeHandlingSA.get(), 0, sizeof(tBikeHandlingDataSA));
 }
 
 CBikeHandlingEntrySA::CBikeHandlingEntrySA(tBikeHandlingDataSA* pOriginal)
 {
     // Store gta's pointer
     m_pBikeHandlingSA = NULL;
-    m_bDeleteInterface = false;
     memcpy(&m_BikeHandling, pOriginal, sizeof(tBikeHandlingDataSA));
-}
-
-CBikeHandlingEntrySA::~CBikeHandlingEntrySA()
-{
-    if (m_bDeleteInterface)
-    {
-        delete m_pBikeHandlingSA;
-    }
 }
 
 // Apply the handlingdata from another data
@@ -49,8 +39,8 @@ void CBikeHandlingEntrySA::Recalculate()
     if (m_pBikeHandlingSA)
     {
         // Copy our stored field to GTA's
-        memcpy(m_pBikeHandlingSA, &m_BikeHandling, sizeof(m_BikeHandling));
+        memcpy(m_pBikeHandlingSA.get(), &m_BikeHandling, sizeof(m_BikeHandling));
 
-       ((void(_stdcall*)(tBikeHandlingDataSA*))FUNC_HandlingDataMgr_ConvertBikeDataToGameUnits)(m_pBikeHandlingSA);
+       ((void(_stdcall*)(tBikeHandlingDataSA*))FUNC_HandlingDataMgr_ConvertBikeDataToGameUnits)(m_pBikeHandlingSA.get());
     }
 }
--- a/Client/game_sa/CBikeHandlingEntrySA.cpp
+++ b/Client/game_sa/CBikeHandlingEntrySA.cpp
@@ -13,34 +13,24 @@
 
 CBikeHandlingEntrySA::CBikeHandlingEntrySA()
 {
-    // Create a new interface and zero it
-    m_pBikeHandlingSA = std::make_shared<tBikeHandlingDataSA>();
-    memset(m_pBikeHandlingSA.get(), 0, sizeof(tBikeHandlingDataSA));
+    memset(&m_pBikeHandlingSA, 0, sizeof(tBikeHandlingDataSA));
 }
 
 CBikeHandlingEntrySA::CBikeHandlingEntrySA(tBikeHandlingDataSA* pOriginal)
 {
-    // Store gta's pointer
-    m_pBikeHandlingSA = NULL;
-    memcpy(&m_BikeHandling, pOriginal, sizeof(tBikeHandlingDataSA));
+    memset(&m_pBikeHandlingSA, 0, sizeof(tBikeHandlingDataSA));
+    m_BikeHandling = *pOriginal;
 }
 
-// Apply the handlingdata from another data
 void CBikeHandlingEntrySA::Assign(const CBikeHandlingEntry* pData)
 {
-    // Copy the data
     const CBikeHandlingEntrySA* pEntrySA = static_cast<const CBikeHandlingEntrySA*>(pData);
     m_BikeHandling = pEntrySA->m_BikeHandling;
 }
 
 void CBikeHandlingEntrySA::Recalculate()
 {
-    // Real GTA class?
-    if (m_pBikeHandlingSA)
-    {
-        // Copy our stored field to GTA's
-        memcpy(m_pBikeHandlingSA.get(), &m_BikeHandling, sizeof(m_BikeHandling));
-
-       ((void(_stdcall*)(tBikeHandlingDataSA*))FUNC_HandlingDataMgr_ConvertBikeDataToGameUnits)(m_pBikeHandlingSA.get());
-    }
+    auto HandlingDataMgr_ConvertBikeDataToGameUnits = (void(_stdcall*)(tBikeHandlingDataSA*))0x6F5290;
+    m_pBikeHandlingSA = m_BikeHandling;
+    HandlingDataMgr_ConvertBikeDataToGameUnits(&m_pBikeHandlingSA);
 }
--- a/Client/game_sa/CBikeHandlingEntrySA.cpp
+++ b/Client/game_sa/CBikeHandlingEntrySA.cpp
@@ -1,0 +1,56 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        game_sa/CHandlingEntrySA.cpp
+ *  PURPOSE:     Vehicle handling data entry
+ *
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+CBikeHandlingEntrySA::CBikeHandlingEntrySA()
+{
+    // Create a new interface and zero it
+    m_pBikeHandlingSA = new tBikeHandlingDataSA;
+    memset(m_pBikeHandlingSA, 0, sizeof(tBikeHandlingDataSA));
+    m_bDeleteInterface = true;
+}
+
+CBikeHandlingEntrySA::CBikeHandlingEntrySA(tBikeHandlingDataSA* pOriginal)
+{
+    // Store gta's pointer
+    m_pBikeHandlingSA = NULL;
+    m_bDeleteInterface = false;
+    memcpy(&m_BikeHandling, pOriginal, sizeof(tBikeHandlingDataSA));
+}
+
+CBikeHandlingEntrySA::~CBikeHandlingEntrySA()
+{
+    if (m_bDeleteInterface)
+    {
+        delete m_pBikeHandlingSA;
+    }
+}
+
+// Apply the handlingdata from another data
+void CBikeHandlingEntrySA::Assign(const CBikeHandlingEntry* pData)
+{
+    // Copy the data
+    const CBikeHandlingEntrySA* pEntrySA = static_cast<const CBikeHandlingEntrySA*>(pData);
+    m_BikeHandling = pEntrySA->m_BikeHandling;
+}
+
+void CBikeHandlingEntrySA::Recalculate()
+{
+    // Real GTA class?
+    if (m_pBikeHandlingSA)
+    {
+        // Copy our stored field to GTA's
+        memcpy(m_pBikeHandlingSA, &m_BikeHandling, sizeof(m_BikeHandling));
+
+       ((void(_stdcall*)(tBikeHandlingDataSA*))FUNC_HandlingDataMgr_ConvertBikeDataToGameUnits)(m_pBikeHandlingSA);
+    }
+}
--- a/Client/game_sa/CBoatHandlingEntrySA.cpp
+++ b/Client/game_sa/CBoatHandlingEntrySA.cpp
@@ -14,25 +14,15 @@
 CBoatHandlingEntrySA::CBoatHandlingEntrySA()
 {
     // Create a new interface and zero it
-    m_pBoatHandlingSA = new tBoatHandlingDataSA;
-    memset(m_pBoatHandlingSA, 0, sizeof(tBoatHandlingDataSA));
-    m_bDeleteInterface = true;
+    m_pBoatHandlingSA = std::make_shared<tBoatHandlingDataSA>();
+    memset(m_pBoatHandlingSA.get(), 0, sizeof(tBoatHandlingDataSA));
 }
 
 CBoatHandlingEntrySA::CBoatHandlingEntrySA(tBoatHandlingDataSA* pOriginal)
 {
     // Store gta's pointer
-    m_pBoatHandlingSA = new tBoatHandlingDataSA;
-    m_bDeleteInterface = false;
-    memcpy(m_pBoatHandlingSA, pOriginal, sizeof(tBoatHandlingDataSA));
-}
-
-CBoatHandlingEntrySA::~CBoatHandlingEntrySA()
-{
-    if (m_bDeleteInterface)
-    {
-        delete m_pBoatHandlingSA;
-    }
+    m_pBoatHandlingSA = std::make_shared<tBoatHandlingDataSA>();
+    memcpy(m_pBoatHandlingSA.get(), pOriginal, sizeof(tBoatHandlingDataSA));
 }
 
 // Apply the handlingdata from another data
--- a/Client/game_sa/CBoatHandlingEntrySA.cpp
+++ b/Client/game_sa/CBoatHandlingEntrySA.cpp
@@ -13,22 +13,16 @@
 
 CBoatHandlingEntrySA::CBoatHandlingEntrySA()
 {
-    // Create a new interface and zero it
-    m_pBoatHandlingSA = std::make_shared<tBoatHandlingDataSA>();
-    memset(m_pBoatHandlingSA.get(), 0, sizeof(tBoatHandlingDataSA));
+    memset(&m_pBoatHandlingSA, 0, sizeof(tBoatHandlingDataSA));
 }
 
 CBoatHandlingEntrySA::CBoatHandlingEntrySA(tBoatHandlingDataSA* pOriginal)
 {
-    // Store gta's pointer
-    m_pBoatHandlingSA = std::make_shared<tBoatHandlingDataSA>();
-    memcpy(m_pBoatHandlingSA.get(), pOriginal, sizeof(tBoatHandlingDataSA));
+    m_pBoatHandlingSA = *pOriginal;
 }
 
-// Apply the handlingdata from another data
 void CBoatHandlingEntrySA::Assign(const CBoatHandlingEntry* pData)
 {
-    // Copy the data
     const CBoatHandlingEntrySA* pEntrySA = static_cast<const CBoatHandlingEntrySA*>(pData);
     m_pBoatHandlingSA = pEntrySA->m_pBoatHandlingSA;
 }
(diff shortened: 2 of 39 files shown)


Client/game_sa/CBikeHandlingEntrySA.cpp
```diff
@@ -0,0 +1,56 @@
...
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+CBikeHandlingEntrySA::CBikeHandlingEntrySA()
+{
+    // Create a new interface and zero it
+    m_pBikeHandlingSA = new tBikeHandlingDataSA;
```
Pirulax: If possible use smart pointers.

Client/game_sa/CBoatHandlingEntrySA.cpp
```diff
@@ -0,0 +1,44 @@
...
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+CBoatHandlingEntrySA::CBoatHandlingEntrySA()
+{
+    // Create a new interface and zero it
+    m_pBoatHandlingSA = new tBoatHandlingDataSA;
```
Pirulax: Same here, if possible use smart pointers.

Client/game_sa/CFlyingHandlingEntrySA.cpp
```diff
@@ -0,0 +1,44 @@
...
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+CFlyingHandlingEntrySA::CFlyingHandlingEntrySA()
+{
+    // Create a new interface and zero it
+    m_pFlyingHandlingSA = new tFlyingHandlingDataSA;
```
Pirulax: Same here, if possible use smart pointers.

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -53,20 +73,70 @@ bool CClientModel::Deallocate(void)
     // ModelInfo must be valid
     if (pModelInfo->IsValid())
     {
-        if (m_eModelType == CCLIENTMODELPED)
+        if (m_eModelType == eClientModelType::PED)
```
Pirulax: I think a `switch` should be used here.
Also, maybe consider adding a lambda(obv. inside the function, as it will be inlined) for calling the event, instead of copy-pasting the code.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1061,8 +1089,34 @@ void CClientVehicle::SetModelBlocking(unsigned short usModel, unsigned char ucVa
...
+             m_pBoatHandlingEntry->Assign(m_pOriginalBoatHandlingEntry);
+        }
+
+        m_pOriginalBikeHandlingEntry = g_pGame->GetHandlingManager()->GetOriginalBikeHandlingData((eVehicleTypes)usHandlingModelID);
+        if (m_pOriginalBikeHandlingEntry)
+        {
+            if (!m_pBikeHandlingEntry)
+                m_pBikeHandlingEntry = g_pGame->GetHandlingManager()->CreateBikeHandlingData();
+
+            m_pBikeHandlingEntry->Assign(m_pOriginalBikeHandlingEntry);
```
Pirulax, line 1117: I dont understand why the original code has used this, when you have `operator=`.
So this would become `*m_pBikeHandlingEntry = *m_pOriginalBikeHandlingEntry)

Client/mods/deathmatch/logic/CClientVehicleManager.h
```diff
@@ -34,6 +34,7 @@ class CClientVehicleManager
 
     static bool               IsTrainModel(unsigned long ulModel);
     static bool               IsValidModel(unsigned long ulModel);
+    static bool               IsStandartModel(unsigned long ulModel);
```
Pirulax: Typo? Is it supposed to be `IsStandardModel`?

Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.h
```diff
@@ -454,6 +455,11 @@ inline SString GetClassByTypeName(eObjectGroup::BreakMode*)
     return "objectgroup-breakmode";
 }
 
+inline SString GetClassByTypeName(eClientModelType)
+{
+    return "client-model-type";
+}
+
```
Pirulax: This is used for classes only afaik.

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -58,6 +58,56 @@ static void _declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+
+
+static CVehicleSAInterface* pRequestSoundSettingsVehicle = 0;
+tVehicleAudioSettings*      pVehicleAudioSettings = nullptr;
+
+static tVehicleAudioSettings* __fastcall getVehicleSoundSettings()
+{
+    ushort usModel = pRequestSoundSettingsVehicle->m_nModelIndex;
+    // Check if it is a custom model
+    if (usModel < 400 || usModel > 611)
```
Pirulax: IMHO it would be cool if there would be a static method from now on, instead of `magic` numbers.
But we do need opinions on this one. Maybe its just me thinking this way

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -58,6 +58,56 @@ static void _declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+
+static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
+{
+    _asm
+    {
+        pushad
+        mov     pRequestSoundSettingsVehicle, edx
+    }
+
+    pVehicleAudioSettings = getVehicleSoundSettings();
```
Pirulax, line 100: In theory you can initalize local variables in a nested block:
```cpp
{
    auto pVehicleAudioSettings = getVehicleSoundSettings();
    _asm
    {
        popad
        lea    edi, [ebp + 80h]
        mov    esi, pVehicleAudioSettings
        jmp CONTINUE_CAEVehicleAudioEntity__Initialise
    }
}
```

Pirulax (changes requested):

Pirulax:
Very nice work, thanks for your time!
This is a long awaited feature, and I'm really hyped about it.

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: Idk, but switch don't work here

Pirulax: Here you go, the refactored code:
```cpp
bool CClientModel::Deallocate(void)
{
    if (!m_bAllocatedByUs)
        return false;

    CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
    if (!pModelInfo) // ModelInfo must be valid
        return false;

    auto unloadModelsAndCallEvents = [&](auto iterBegin, auto iterEnd, unsigned short usParentID, auto setElementModelLambda) {
        for (auto iter = iterBegin; iter != iterEnd; iter++)
        {
            auto& element = **iter;

            if (element.GetModel() != m_iModelID)
                continue;

            if (element.IsStreamedIn())
                element.StreamOutForABit();

            setElementModelLambda(element);

            CLuaArguments Arguments;
            Arguments.PushNumber(m_iModelID);
            Arguments.PushNumber(usParentID);
            element.CallEvent("onClientElementModelChange", Arguments, true);
        }
    };

    switch (m_eModelType)
    {
    case eClientModelType::PED:
    {
        // If some ped is using this ID, change him to CJ
        CClientPedManager*                       pPedManager = g_pClientGame->GetManager()->GetPedManager();

        unloadModelsAndCallEvents(pPedManager->IterBegin(), pPedManager->IterEnd(), 0, 
            [](auto& element) { element.SetModel(0); });

        break;
    }
    case eClientModelType::OBJECT:
    {
        const auto& objects pObjectManager = g_pClientGame->GetManager()->GetObjectManager()->GetObjects();
        unsigned short usParentID = g_pGame->GetModelInfo(m_iModelID)->GetParentID();

        unloadModelsAndCallEvents(objects->begin(), objects->end(), usParentID, 
            [=](auto& element) { element.SetModel(usParentID); });

        break;
    }
    case eClientModelType::VEHICLE:
    {
        CClientVehicleManager* pVehicleManager = g_pClientGame->GetManager()->GetVehicleManager();
        unsigned short         usParentID = g_pGame->GetModelInfo(m_iModelID)->GetParentID();

        unloadModelsAndCallEvents(pVehicleManager->IterBegin(), pVehicleManager->IterEnd(), usParentID, 
            [=](auto& element) { element.SetModelBlocking(usParentID, 255, 255); });

        break;
    }

    // Restore DFF/TXD
    g_pClientGame->GetManager()->GetDFFManager()->RestoreModel(m_iModelID);

    // Restore COL (for non ped models)
    if (m_eModelType != eClientModelType::PED)
        g_pClientGame->GetManager()->GetColModelManager()->RestoreModel(m_iModelID);

    pModelInfo->DeallocateModel();

    this->SetParentResource(nullptr);

    return true;
}
```
Edit: Hopefully fixed the code.

TheNormalnij: This code don't work for peds

TheNormalnij: Thanks you!

TheNormalnij:
I see some potential bugs here.
I think it's easier to make review when  #1684 will be merged.

ghost:
If somebody writes code where CBikeSAInterface members are accessed, then the game will crash. This is because the size of CVehicleSAInterface is wrong. Here's a screenshot from CVehicleSAInterface: <details>
  <summary>screenshot of CVehicleSAInterface</summary>

image
</details>

You can see the red line. Anything that comes after offset `1440` is from the CTrainSAInterface class which the old mta devs did not create. So, if you're going to add CBikeSAInterface class, you'll need to remove the train class members from CVehicleSAInterface and create CTrainSAInterface class. This is how the CTrainSAInterface class should look like: https://github.com/saml1er/gta-reversed/blob/master/source/game_sa/Entity/Vehicle/CTrain.h

If you want to do it in this PR, then that's completely fine by me.

Client/mods/deathmatch/logic/CClientVehicleManager.h
TheNormalnij: Yeah

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
TheNormalnij: It not allowed in naked functions

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
TheNormalnij: There is no way to don't use `magic` numbers.

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
qaisjp: @pirulax we no longer allow asm mixed with C++ in the same function

Client/mods/deathmatch/logic/CClientVehicle.cpp
TheNormalnij: Because `*m_pBikeHandlingEntry = *m_pOriginalBikeHandlingEntry)` will copy `m_bDeleteInterface` field. I'm not sure that it will works without `CBikeHandlingEntrySA` declaration in `sdk/game_sa`

Client/game_sa/CBikeHandlingEntrySA.cpp
TheNormalnij: Smart pointers are cool, but i don't see a reason to use it there

Pirulax: Afaik in new code its recommended(or maybe required? not sure) to use them.

TheNormalnij:
Ready for review

Inder00 (approved):

Server/mods/deathmatch/logic/CVehicle.h
```diff
@@ -322,8 +322,9 @@ class CVehicle : public CElement
     void SpawnAt(const CVector& vecPosition, const CVector& vecRotation);
     void Respawn();
 
-    void            GenerateHandlingData();
-    CHandlingEntry* GetHandlingData() { return m_pHandlingEntry; }
+    void                  GenerateHandlingData();
+    CHandlingEntry*       GetHandlingData() { return m_pHandlingEntry; };
+    //CFlyingHandlingEntry* GetFlyingHandlingData() { return m_pFlyingHandlingEntry; };
```
ghost: Maybe remove this commented line?

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
ghost:
```suggestion
    if (usModel < VT_LANDSTAL || usModel >= VT_MAX)
```

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -58,6 +58,56 @@ static void _declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+static CVehicleSAInterface* pRequestSoundSettingsVehicle = 0;
+tVehicleAudioSettings*      pVehicleAudioSettings = nullptr;
+
+static tVehicleAudioSettings* __fastcall getVehicleSoundSettings()
+{
+    ushort usModel = pRequestSoundSettingsVehicle->m_nModelIndex;
+    // Check if it is a custom model
+    if (usModel < 400 || usModel > 611)
+        usModel = pGameInterface->GetModelInfo(usModel)->GetParentID();
+    return (tVehicleAudioSettings*)0x860AF0 + (usModel - 400);
```
ghost:
```suggestion
    return (tVehicleAudioSettings*)0x860AF0 + (usModel - VT_LANDSTAL);
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -585,7 +585,29 @@ int CLuaEngineDefs::EngineRequestModel(lua_State* luaVM)
...
+                        {
+                            case eClientModelType::PED:
+                                usParentID = 7;
+                                break;
+                            case eClientModelType::OBJECT:
+                                usParentID = 1337;
+                                break;
+                            case eClientModelType::VEHICLE:
+                                usParentID = 400;
+                                break;
```
ghost:
```suggestion
                        switch (eModelType)
                        {
                            case eClientModelType::PED:
                                usParentID = 7; // male01
                                break;
                            case eClientModelType::OBJECT:
                                usParentID = 1337; // BinNt07_LA (trash can)
                                break;
                            case eClientModelType::VEHICLE:
                                usParentID = VT_LANDSTAL;
                                break;
```
Please format this code using the formatter to fix the code comment indentation.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4311,6 +4398,51 @@ CHandlingEntry* CClientVehicle::GetHandlingData()
...
+{
+    if (m_pVehicle)
+    {
+        return m_pVehicle->GetFlyingHandlingData();
+    }
+    else if (m_pFlyingHandlingEntry)
+    {
+        return m_pFlyingHandlingEntry;
+    }
+    return NULL;
```
ghost:
```suggestion
    return nullptr;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4311,6 +4398,51 @@ CHandlingEntry* CClientVehicle::GetHandlingData()
...
+
+    if (m_pVehicle)
+    {
+        return dynamic_cast<CBoat*>(m_pVehicle)->GetBoatHandlingData();
+    }
+    else if (m_pBoatHandlingEntry)
+    {
+        return m_pBoatHandlingEntry;
+    }
+    return NULL;
```
ghost:
```suggestion
        return reinterpret_cast<CBoat*>(m_pVehicle)->GetBoatHandlingData();
    }
    else if (m_pBoatHandlingEntry)
    {
        return m_pBoatHandlingEntry;
    }
    return nullptr;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4311,6 +4398,51 @@ CHandlingEntry* CClientVehicle::GetHandlingData()
...
+
+    if (m_pVehicle)
+    {
+        return dynamic_cast<CBike*>(m_pVehicle)->GetBikeHandlingData();
+    }
+    else if (m_pBikeHandlingEntry)
+    {
+        return m_pBikeHandlingEntry;
+    }
+    return NULL;
```
ghost:
```suggestion
        return reinterpret_cast<CBike*>(m_pVehicle)->GetBikeHandlingData();
    }
    else if (m_pBikeHandlingEntry)
    {
        return m_pBikeHandlingEntry;
    }
    return nullptr;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4311,6 +4398,51 @@ CHandlingEntry* CClientVehicle::GetHandlingData()
...
+    {
+        return m_pBoatHandlingEntry;
+    }
+    return NULL;
+}
+
+CBikeHandlingEntry* CClientVehicle::GetBikeHandlingData()
+{
+    if (m_eVehicleType != CLIENTVEHICLE_BIKE && m_eVehicleType != CLIENTVEHICLE_BMX)
+        return NULL;
```
ghost:
```suggestion
    if (m_eVehicleType != CLIENTVEHICLE_BIKE && m_eVehicleType != CLIENTVEHICLE_BMX)
        return nullptr;
```

ghost (commented):
Things that I highlighted:
1. Use `nullptr` instead of `NULL`
2. Avoid `dynamic_cast`. First try to use `static_cast`, if not then use `reinterpret_cast`. dynamic_cast is slow even though it is considered a good practice to use it. 
3. Always prefer smart pointers over raw pointers. They are very neat and avoids memory leaks. First try to use `std::unique_ptr`, and if you need a shared resource, then go for `std::shared_ptr`. You can see check the IFP implementation, it uses smart pointers.

Client/mods/deathmatch/logic/CClientVehicle.cpp
TheNormalnij: NULL -> nullptr in  CClientVehicle.cpp

TheNormalnij:
@saml1er it's right in bf851c2?

ghost:
> 
> 
> @saml1er it's right in [bf851c2](https://github.com/multitheftauto/mtasa-blue/commit/bf851c23e551b7ee5db57e32f3b7d4d25b20306d)?

Yes, looks good, thanks.

TheNormalnij:
Fixed

ghost (approved):

Client/game_sa/CHandlingManagerSA.cpp
```diff
@@ -192,6 +242,54 @@ const CHandlingEntry* CHandlingManagerSA::GetOriginalHandlingData(eVehicleTypes
...
+}
+
+const CBikeHandlingEntry* CHandlingManagerSA::GetOriginalBikeHandlingData(eVehicleTypes eModel)
+{
+    // Within range?
+    if (eModel >= 400 && eModel < VT_MAX)
+    {
+        // Get our Handling ID
+        eHandlingTypes eHandling = GetHandlingID(eModel);
+        if (eHandling >= HT_BIKE || eHandling <= HT_FREEWAY)
```
I3D99, line 286: I consistently get game crashes (offset 0x0000A6B6) on the latest nightly (r20716) (does not occur on a debug build) whenever a car spawns (specifically tested with a banshee but it happens with other cars too, but not bikes, helis, planes, etc)
The crash is a read access violation on line 28 of CBikeHandlingEntrySA.cpp on `CBikeHandlingEntrySA::Assign` called from `CClientVehicle::CClientVehicle` constructor, only if `CHandlingManagerSA::GetOriginalBikeHandlingData` returns a valid pointer which should only happen with handling IDs between `HT_BIKE` and `HT_FREEWAY` but this line wrongly catches pretty much all handling IDs there are.
```suggestion
        if (eHandling >= HT_BIKE && eHandling <= HT_FREEWAY)
```

qaisjp: Nice catch!

Allerek:
Needs to be adjusted for:
https://wiki.multitheftauto.com/wiki/GetVehicleName
https://wiki.multitheftauto.com/wiki/GetVehicleNameFromModel
https://wiki.multitheftauto.com/wiki/GetVehicleNameFromModel
https://github.com/multitheftauto/mtasa-blue/issues/1175

PR #1704, opened 2020-10-10
Fix engineGetModelTextures memory leak

theSarrum:
Stopping a resource that used engineGetModelTextures does not release the memory used by textures. This pull request solves this problem by setting a parent for the texture.


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1126,11 +1126,24 @@ int CLuaEngineDefs::EngineGetModelTextures(lua_State* luaVM)
...
         return luaL_error(luaVM, argStream.GetFullErrorMessage());
 
+    CLuaMain*  pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+    CResource* pParentResource = pLuaMain ? pLuaMain->GetResource() : NULL;
+
+    if (!pParentResource)
+    {
+        lua_pushnil(luaVM);
+        return 1;
+    }
```
qaisjp: Is `pParentResource` ever actually nil?

ghost: I don't think so. This check is not needed. @theSarrum Please remove this check, we can merge this PR for 1.5.8.

ghost (commented):
Good catch, thanks.

qaisjp (approved):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1126,11 +1126,18 @@ int CLuaEngineDefs::EngineGetModelTextures(lua_State* luaVM)
     if (argStream.HasErrors())
         return luaL_error(luaVM, argStream.GetFullErrorMessage());
 
+    CLuaMain*  pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+    CResource* pParentResource = pLuaMain ? pLuaMain->GetResource() : NULL;
```
qaisjp:
```suggestion
    CResource* pParentResource = pLuaMain->GetResource();
```

qaisjp (commented):
left a small suggestion but this lgtm

StrixG (approved):

PR #1711, opened 2020-10-12
Add intersectsSegmentTriangle method to Vector3

Pirulax:
I was lazy to switch branches lol

This PR exposes the `intersectsSegmentTriangle` method that can be found in the `CVector` class. (its used in testLineAgainstWater)
The performance seems way better(300%, although my algorithm is more optimized, as it has everything possible cached) than the Lua implementation (with vectors. Without it probably would be way better).

Hope this gets merged fast, as the algorithm itself it tested already.

Note:  Haven't looked into the algorithm itself, but im pretty sure its optimized. The bottleneck is our Lua API probably. (Maybe the old parser would be a slightly bit faster)


Woovie (approved):

Pirulax:
@sbx320 Could you help me with this? The non msvc builds

qaisjp (changes requested):
1. Non-windows will need fixing
2. Place them underneath "dot", there's no need to start a brand new section (in this case it starts with create/static, then special operations, then getters, then setters, then variables)

botder:
@Pirulax what's the status on this?

Pirulax:
Well should I make it work?

botder:
Sure, why not.

patrikjuvonen:
Hey @Pirulax, what's the status of this pull request, can we expect it for the upcoming release?

Pirulax:
This function can be called without an instance of a vector. Eg.:
```lua
Vector3.intersectsSegmentTriangle(0, 0, 0, 0, 5, 5, 5, 0, 0, 0, 5, 0, 0, 0, 5);
```
```lua
Vector3(0, 0, 0):intersectsSegmentTriangle(0, 5, 5, 5, 0, 0, 0, 5, 0, 0, 0, 5);
```
Will yield the same result.

Should it be added as a global function as well at this point?

patrikjuvonen:
> This function can be called without an instance of a vector. Eg.:
> 
> ```lua
> Vector3.intersectsSegmentTriangle(0, 0, 0, 0, 5, 5, 5, 0, 0, 0, 5, 0, 0, 0, 5);
> ```
> 
> ```lua
> Vector3(0, 0, 0):intersectsSegmentTriangle(0, 5, 5, 5, 0, 0, 0, 5, 0, 0, 0, 5);
> ```
> 
> Will yield the same result.
> 
> Should it be added as a global function as well at this point?

I'm not sure we have other Vector3-based global functions as such so this would stand out unnecessarily. I think it's fine like this.

Pirulax:
Well `getDistanceBetweenPoints` does the same as `(a - b).Length()`
So thats some kind of a wrapper.

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -11,6 +11,10 @@
 #include <variant>
 #include <array>
 
+#include <CVector2D.h>
+#include <CVector.h>
+#include <CVector4D.h>
+
```
patrikjuvonen: Are these additions needed?

Pirulax:
Just wanna make sure it gets merged to 1.5.9

patrikjuvonen (approved):
Thanks, tested lightly.

patrikjuvonen:
Please document this change on the Vector3 Wiki page at https://wiki.multitheftauto.com/wiki/Vector/Vector3

PR #1712, opened 2020-10-14
Fix #1708: Force people to opt back into Default release type

### LINKED ISSUE #1708: Force people to opt back into Default release type when they update to 1.5.8

people went to nightly because of discord rich presence, and are prolly still on nightly unnecessarily

Ideally when they update MTA they get a window saying "We've reset you back to Default, please go change your settings again if you want Nightly updates." or something like that

**Additional context**

ghost:
Sets the version settings back to stable build if the client is on 1.5.8. A new subkey gets created in the registry after reverting:
```
HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Multi Theft Auto: San Andreas All\1.5\Settings\general
```

image

image

### DIFF
--- a/Client/core/CVersionUpdater.cpp
+++ b/Client/core/CVersionUpdater.cpp
@@ -3384,8 +3384,8 @@ int CVersionUpdater::DoPollPost()
 // We do this only once, this is why we need to write to registry.
 void CVersionUpdater::OnMainMenuFullyVisible()
 {
-    const SString requiredValue = "discord-rich-presence";
-    const SString lastSubKey = "version-revert-reason";
+    const std::string requiredValue = "discord-rich-presence";
+    const std::string lastSubKey = "version-revert-reason";
     if (GetApplicationSetting(lastSubKey) != requiredValue)
     {
         SetApplicationSetting(lastSubKey, requiredValue);
@@ -3393,8 +3393,8 @@ void CVersionUpdater::OnMainMenuFullyVisible()
         if (MTASA_VERSION_MAJOR == 1 && MTASA_VERSION_MINOR == 5 && MTASA_VERSION_MAINTENANCE == 8)
         {
             CVARS_SET("update_build_type", 0);
-            SString strMessage = _("We've reset your update preferences back to Default, please go change your settings again if you want Nightly updates.\n");
-            g_pCore->ShowMessageBox(_("VERSION UPDATE INFORMATION"), strMessage, MB_BUTTON_OK | MB_ICON_INFO);
+            std::string message = _("We've reset your update preferences back to Default, please go change your settings again if you want Nightly updates.\n");
+            g_pCore->ShowMessageBox(_("VERSION UPDATE INFORMATION"), message.c_str(), MB_BUTTON_OK | MB_ICON_INFO);
         }
     }
 }
--- a/Client/core/CVersionUpdater.cpp
+++ b/Client/core/CVersionUpdater.cpp
@@ -3384,6 +3384,8 @@ int CVersionUpdater::DoPollPost()
 // We do this only once, this is why we need to write to registry.
 void CVersionUpdater::OnMainMenuFullyVisible()
 {
+    if (MTASA_VERSION_BUILD != 20670)            // 20670 is the build number with PR #1712 merged.
+        return;
     const std::string requiredValue = "discord-rich-presence";
     const std::string lastSubKey = "version-revert-reason";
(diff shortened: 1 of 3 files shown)


Woovie:
`we've reset you` -> `we've reset your update preferences`

jm2c

Client/core/CVersionUpdater.cpp
```diff
@@ -3377,3 +3378,23 @@ int CVersionUpdater::DoPollPost()
...
     return RES_OK;
 }
+
+// Issue: #1708
+// Set the update settings to stable build if the user is on 1.5.8.
+// We do this only once, this is why we need to write to registry.
+void CVersionUpdater::OnMainMenuFullyVisible()
+{
+    const SString requiredValue = "discord-rich-presence";
+    const SString lastSubKey = "version-revert-reason";
```
qaisjp: Nit: maybe we should start thinking about preferring `std::string`?

qaisjp (approved):
This looks good. I left a tiny nit, but it's optional / non-blocking.

When a nightly is downloaded from nightly.mtasa.com is the `build_type` by default `0`? Just want to make sure that if someone downloads a nightly from nightly.mtasa.com for the first time, they don't get this message.

Client/core/CVersionUpdater.cpp
ghost: `SString` inherits from `std::string`. SString has some very useful string operation functions, like formatting the string, converting case, or trimming the string.

> When a nightly is downloaded from nightly.mtasa.com is the build_type by default 0? Just want to make sure that if someone downloads a nightly from nightly.mtasa.com for the first time, they don't get this message.

Hmm, I'll check it.

qaisjp: Indeed, but in this case you aren't using any of those features.

Pirulax: Can't we somehow modify `std::string` instead?

ccw808: Maybe we should always use SString instead of std::string for consistency

Pirulax: My main problem with `SString` is the lack of modern features (like std::string_view ctor, and as return value), also I dont like the inconsistencies in the naming of the functions, it looks really funny..
```cpp
SString str;
str.erase();
str.Split()
``` 
It just is a pain in the ass to maintain imho, and we should find a way to get rid of it.

ghost:
> When a nightly is downloaded from nightly.mtasa.com is the build_type by default 0? Just want to make sure that if someone downloads a nightly from nightly.mtasa.com for the first time, they don't get this message.

The build type is set to `0x9`. It could mean two things: stable or nightly. 

We can still check if it's a nightly build or not, but apparently it seems there's no way to check if the client downloaded the nightly directly from nightly.mtasa.com.

Client/core/CVersionUpdater.cpp
qaisjp: This is just a nit, so feel free to stay with `SString` @saml1er.

There's some contention here and this isn't the place for discussing / debating `SString` vs `std::string(_view)`, so maybe we can create an issue about it and discuss it there.

(Until an issue is created, if you're curious about why I'm suggesting to prefer `std::string(view)` where possible, [you might find this conversation interesting](https://discord.com/channels/278474088903606273/366384007535001612/740682668647252038). In the meantime, I don't think we should discuss it further in this PR.)

ghost: > (Until an issue is created, if you're curious about why I'm suggesting to prefer std::string(view) where possible, you might find this conversation interesting. In the meantime, I don't think we should discuss it further in this PR.)

@qaisjp Thanks, that explains it well. Btw, we should stick with one and only use `SString` if std::string doesn't suffice. I'll definitely use std::string for this PR. I think it would be nice if we can add this to the coding guidelines, so other contributors know what to use.

ghost:
I did a lot of digging and just can't find any way to check if the current build was downloaded from nightly.mtasa.com. I have another idea. We could just add another button to the update information box like `Don't Reset`.

qaisjp:
What if you make it so that it only changes the build_type (and popup) if the `revision == 20669 || revision == 20668`?

qaisjp (approved):
This should work. We just have to make sure we don't increase the nightly autoupdater above 20670 for a while (a week?). We can still merge things once 20670 has been built, though.

PR #1714, opened 2020-10-14
Add interface for model flags

TheNormalnij:
**API:**
```lua
bool engineSetModelFlags( int modelID, int flags [, bool bIdeFlags] )
int engineGetModelFlags( int modelID )
bool engineResetModelFlags( int modelID )
bool engineGetModelFlag( int modelID, string flagName )
bool engineSetModelFlag( int modelID, string flagName, bool state )
```

Flags:
```
is_road
draw_last
addictive
ignore_lighting
no_zbuffer_write
dont_receive_shadows
is_glass_type_1
is_glass_type_2
is_garage_door
is_damagable
is_tree
is_palm
does_not_collide_with_flyer
is_tag
disable_backface_culling
is_breakable_statue
is_crane
```

[Flags documentation](https://gtamods.com/wiki/Item_Definition)

**Important**
`engineGetModelFlags` returns internal GTA flags.
You can transform IDE flags into GTA flags with 3 argument in `engineSetModelFlags`.


Client/game_sa/CModelInfoSA.h
```diff
@@ -36,7 +36,9 @@ class CPedModelInfoSAInterface;
 #define     ARRAY_ModelInfo                 0xA9B0C8
 #define     CLASS_CText                     0xC1B340
 #define     FUNC_CText_Get                  0x6A0050
-#define     FUNC_GetModelFlags              0x4044E0
+#define     FUNC_SetBaseModelFlags          0x5B3AD0
+#define     FUNC_SetAtimicModelFlags        0x5B3B20
```
StrixG:
```suggestion
#define     FUNC_SetAtomicModelFlags        0x5B3B20
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -482,21 +483,67 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+    if (!m_pInterface)
+        return;
+
+    // Save default value if not done yet
+    if (!MapContains(ms_ModelDefaultFlagsMap, m_dwModelID))
+        MapSet(ms_ModelDefaultFlagsMap, m_dwModelID, m_pInterface->usFlags);
+
+    m_pInterface->usFlags = 0xC0;
+    typedef char(__cdecl * Function_SetAtomicModelFlags)(CBaseModelInfoSAInterface * pInterface, unsigned int uiFlags);
+    Function_SetAtomicModelFlags setAtomicModelFlags = (Function_SetAtomicModelFlags)(FUNC_SetAtimicModelFlags);
```
StrixG:
```suggestion
    Function_SetAtomicModelFlags setAtomicModelFlags = (Function_SetAtomicModelFlags)(FUNC_SetAtomicModelFlags);
```

TheNormalnij:
Can we merge this?

Woovie:
I don't feel it's ready for merge. I tested this codebase while @qaisjp watched as well. The API being provided for get/set is not adequate. It's difficult to use, confusing, and generally not user friendly. We don't have a recommendation/solution for this, but something should change to make it easier for people to use.

dmi7ry:
> I don't feel it's ready for merge. I tested this codebase while @qaisjp watched as well. The API being provided for get/set is not adequate. It's difficult to use, confusing, and generally not user friendly. We don't have a recommendation/solution for this, but something should change to make it easier for people to use.

Do you mean bitwise operations? If so then it would be possible to add functions `engineSetModelFlag(modelID, "flag", state)` and `engineGetModelFlag(modelID, "flag")`. Like `engineSetModelFlag(modelID, "isRoad", true)`.

Client/game_sa/CModelInfoSA.h
```diff
@@ -291,6 +302,7 @@ class CModelInfoSA : public CModelInfo
     RpClump*                                                                     m_pCustomClump;
     static std::map<unsigned short, int>                                         ms_RestreamTxdIDMap;
     static std::map<DWORD, float>                                                ms_ModelDefaultLodDistanceMap;
+    static std::map<DWORD, unsigned short>                                       ms_ModelDefaultFlagsMap;
```
Pirulax, line 305: Why not combine these two into one?
Also, I feel like an `std::map` has a bigger memory overhead(3 pointers / key value pair), than just a large fixed size array(which also would be a lot faster, and would be less of a hassle).
What do others think?
Perhaps merge it like this, and let me open a separate PR merging them into an `std::array`?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2017,7 +2024,58 @@ int CLuaEngineDefs::EngineRestoreObjectGroupPhysicalProperties(lua_State* luaVM)
...
+            if (bIdeFlags)
+                pModelInfo->SetIdeFlags(uiFlags);
+            else
+                pModelInfo->SetFlags(uiFlags);
+
+            return true;
+        }
+    }
+
+    throw std::invalid_argument("Expected a valid model ID in range [0-19999] at argument 1");
```
Pirulax:
```cpp
if (uiModelID >= 20000) 
    throw std::invalid_argument("Expected a valid model ID in range [0-19999] at argument 1");
``` 
would be better imho. But im just nitpicky.

Pirulax (approved):

TheNormalnij:
[Table ](https://docs.google.com/spreadsheets/d/1sDE8lI1LiuwdzqzRblIeq-N0gp_oPJjP68juVtD0y5U/edit?usp=sharing) with all model flags

Woovie:
> [Table ](https://docs.google.com/spreadsheets/d/1sDE8lI1LiuwdzqzRblIeq-N0gp_oPJjP68juVtD0y5U/edit?usp=sharing) with all model flags

Private sheet

TheNormalnij:
Fixed

TheNormalnij:
Ready

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -19,12 +19,50 @@ CBaseModelInfoSAInterface** ppModelInfo = (CBaseModelInfoSAInterface**)ARRAY_Mod
 
 std::map<unsigned short, int>                                         CModelInfoSA::ms_RestreamTxdIDMap;
 std::map<DWORD, float>                                                CModelInfoSA::ms_ModelDefaultLodDistanceMap;
+std::map<DWORD, unsigned short>                                       CModelInfoSA::ms_ModelDefaultFlagsMap;
```
Pirulax, line 31: `std::map` has a huge overhead for such small values, so having them all in a single unified map would actually reduce both CPU and RAM usage.
For this PR its okay, but Im going to make a separate PR to unify these maps.

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -19,12 +19,50 @@ CBaseModelInfoSAInterface** ppModelInfo = (CBaseModelInfoSAInterface**)ARRAY_Mod
...
+        char bFlag20 : 1;
+        char bIsTag : 1;
+        char bDisableBackfaceCilling : 1;
+        char bIsBreakableStatue : 1;
+        char bFlag24 : 1;
+
+        char cPad : 8;
+    };
+    unsigned int uiFlags;
+};
```
Pirulax: FYI from the [cppreference](https://en.cppreference.com/w/cpp/language/union):
> It's undefined behavior to read from the member of the union that wasn't most recently written. Many compilers implement, as a non-standard language extension, the ability to read inactive members of a union.

I know we are doing it this way since forever, but still, relying on UB is a very bad idea.
Altough, I have no idea what would be the correct way to do this.

Sorry for being nitpicky.

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,21 +521,278 @@ BOOL CModelInfoSA::DoIsLoaded()
...
 {
-    DWORD dwFunc = FUNC_GetModelFlags;
-    DWORD ModelID = m_dwModelID;
-    BYTE  bFlags = 0;
-    _asm
+    m_pInterface = ppModelInfo[m_dwModelID];
+    if (m_pInterface)
+        return m_pInterface->usFlags;
+    return 0;
+}
```
Pirulax: Is this function supposed to be called with non-initialized models?
Because If not, then we shouldn't be returning `0` here.

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,21 +521,278 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+
+unsigned short CModelInfoSA::GetOriginalFlags()
+{
+    if (MapContains(ms_ModelDefaultFlagsMap, m_dwModelID))
+        return MapGet(ms_ModelDefaultFlagsMap, m_dwModelID);
+
+    m_pInterface = ppModelInfo[m_dwModelID];
+    if (!m_pInterface)
+        return 0;
```
Pirulax: Same

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,21 +521,278 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+
+            if (ideFlags.bIsBreakableStatue)
+                m_pInterface->eSpecialModelType = eModelSpecialType::BREAKABLE_STATUE;
+
+            if (ideFlags.bFlag24)
+                m_pInterface->eSpecialModelType = eModelSpecialType::UNKNOW_1;
+
+            // Added
+            if (ideFlags.bFlag20)
+                m_pInterface->eSpecialModelType = eModelSpecialType::CRANE;
```
Pirulax: It seems to me that you should use `else if` here, because (it seems like to me) `m_pInterface->eSpecialModelType` should only be set once?
If so, please use `else if`.

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,21 +521,278 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+            return false;
+    }
+}
+
+void CModelInfoSA::SetModelSpecialType(eModelSpecialType eType, bool bState)
+{
+    if (bState)
+        m_pInterface->eSpecialModelType = eType;
+    else if (m_pInterface->eSpecialModelType == eType)
+        m_pInterface->eSpecialModelType = eModelSpecialType::NONE;
```
Pirulax, line 718: I think this is more straightforward: 
```cpp
if (bState)
        m_pInterface->eSpecialModelType = eType;
    else 
          if (m_pInterface->eSpecialModelType == eType)
                m_pInterface->eSpecialModelType = eModelSpecialType::NONE;
```

Client/sdk/game/CModelInfo.h
```diff
@@ -28,6 +28,27 @@ class CBoundingBox
...
+    IS_DAMAGABLE,
+    IS_TREE,
+    IS_PALM,
+    DOES_NOT_COLLIDE_WITH_FLYER,
+    IS_TAG,
+    DISABLE_BACKFACE_CULLING,
+    IS_BREAKABLE_STATUE,
+    IS_CRANE,
+};
+
```
Pirulax: Perhaps make this an `enum` and use it as a bit mask?
Or hack it like this, so nobody can access it like `IS_TAG`:
```cpp
namespace eModelIdeFlag
{
enum
{
    IS_ROAD,
    DRAW_LAST,
    ADDICTIVE,
    IGNORE_LIGHTING,            // Used with animated objects
    NO_ZBUFFER_WRITE,
    DONT_RECEIVE_SHADOWS,
    IS_GLASS_TYPE_1,
    IS_GLASS_TYPE_2,
    IS_GARAGE_DOOR,
    IS_DAMAGABLE,
    IS_TREE,
    IS_PALM,
    DOES_NOT_COLLIDE_WITH_FLYER,
    IS_TAG,
    DISABLE_BACKFACE_CULLING,
    IS_BREAKABLE_STATUE,
    IS_CRANE,
};
};
```

Client/game_sa/CModelInfoSA.h
```diff
@@ -137,6 +139,20 @@ class CVehicleModelInfo_SA_VTBL : public CClumpModelInfo_SA_VTBL
...
+enum class eModelSpecialType : unsigned char
+{
+    NONE = 0,
+    TREE = 1,
+    PALM = 2,
+    GLASS_1 = 4,
+    GLASS_2 = 5,
+    TAG = 6,
+    GARAGE_DOOR = 7,
+    CRANE = 9, // don't present in IDE but used in gta
```
Pirulax: Grammar:
> Not present in IDE but used in GTA

Client/game_sa/CModelInfoSA.h
```diff
@@ -36,7 +36,9 @@ class CPedModelInfoSAInterface;
 #define     ARRAY_ModelInfo                 0xA9B0C8
 #define     CLASS_CText                     0xC1B340
 #define     FUNC_CText_Get                  0x6A0050
-#define     FUNC_GetModelFlags              0x4044E0
+#define     FUNC_SetBaseModelFlags          0x5B3AD0
+#define     FUNC_SetAtomicModelFlags        0x5B3B20
+#define     FUNC_GetModelStreamFlags        0x4044E0
```
Pirulax: IIRC we've stopped using defines for function addresses.
@saml1er am I correct?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,21 +521,278 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+//    _asm
+//    {
+//        push    ModelID
+//        call    dwFunc
+//        add     esp, 4
+//        mov     bFlags, al
+//    }
+//    return bFlags;
+//}
+
```
Pirulax: Left in by accident?

Pirulax (dismissed):

Client/sdk/game/CModelInfo.h
TheNormalnij: Yeah,  i have think about bit mask but i can't use it for some flags ( IS_DAMAGABLE, IS_TREE, IS_GLASS_TYPE_2, ... )

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: This is almost original GTA SA behavior. Game reads flags in the same way

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Old code for GetFlags function but with wrong behavior

Client/game_sa/CModelInfoSA.h
TheNormalnij: Hmm... Unused

Client/game_sa/CModelInfoSA.cpp
Pirulax: I dont know.. Its really weird, because `m_pInterface->eSpecialModelType` will always be overwritten by the last `true if`, thus a reversed `else if` ends up doing the same.
So you can just do a reversed `else if`:
```cpp

m_pInterface->bDontCollideWithFlyer = ideFlags.bDontCollideWithFlyer;

if (ideFlags.bFlag20)
    m_pInterface->eSpecialModelType = eModelSpecialType::CRANE;
else if (ideFlags.bFlag24)
    m_pInterface->eSpecialModelType = eModelSpecialType::UNKNOW_1;
else if (ideFlags.bIsBreakableStatue)
    m_pInterface->eSpecialModelType = eModelSpecialType::BREAKABLE_STATUE;
else if (ideFlags.bIsTag)
    m_pInterface->eSpecialModelType = eModelSpecialType::TAG;
else if (ideFlags.bIsPalm)
    m_pInterface->eSpecialModelType = eModelSpecialType::PALM;
else if (ideFlags.bIsTree)
    m_pInterface->eSpecialModelType = eModelSpecialType::TREE;
else if (ideFlags.bIsGarageDoor)
    m_pInterface->eSpecialModelType = eModelSpecialType::GARAGE_DOOR;
else if (ideFlags.bIsGlassType2)
    m_pInterface->eSpecialModelType = eModelSpecialType::GLASS_2;
else if (ideFlags.bIsGlassType1)
    m_pInterface->eSpecialModelType = eModelSpecialType::GLASS_1;
```

TheNormalnij: if i will use `else if` `m_pInterface->eSpecialModelType` will be overwritten with first `if`

TheNormalnij: Sounds good, thx

Pirulax: You're welcome!

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: What this function can return?

Pirulax: Is this function allowed to be called even if the interface for the model is `null`?
Because if not, you should just:
```cpp
unsigned short CModelInfoSA::GetFlags()
{
   m_pInterface = ppModelInfo[m_dwModelID];
   return m_pInterface->usFlags;
}
```
This will crash if `m_pInterface` is `null`, which is okay if this function shoudn't be called if `m_pInterface` is `null`.

Client/game_sa/CModelInfoSA.cpp
qaisjp: I don't see a problem with the code that is currently there

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -57,10 +57,14 @@ class CLuaEngineDefs : public CLuaDefs
     LUA_DECLARE(EngineSetObjectGroupPhysicalProperty)
     LUA_DECLARE(EngineGetObjectGroupPhysicalProperty)
     LUA_DECLARE(EngineRestoreObjectGroupPhysicalProperties)
-    static bool CLuaEngineDefs::EngineRestreamWorld(lua_State* const luaVM);
+    static uint CLuaEngineDefs::EngineGetModelFlags(uint uiModelID);
+    static bool CLuaEngineDefs::EngineSetModelFlags(uint uiModelID, uint uiFlags, std::optional<bool> bIdeFlags);
+    static bool CLuaEngineDefs::EngineGetModelFlag(uint uiModelID, eModelIdeFlag eFlag);
+    static bool CLuaEngineDefs::EngineSetModelFlag(uint uiModelID, eModelIdeFlag eFlag, bool state);
+    static bool CLuaEngineDefs::EngineResetModelFlags(uint uiModelID);
+    static bool CLuaEngineDefs::EngineRestreamWorld();
```
qaisjp: You don't need to have `CLuaEngineDefs::` here

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -19,12 +19,50 @@ CBaseModelInfoSAInterface** ppModelInfo = (CBaseModelInfoSAInterface**)ARRAY_Mod
...
 std::unordered_map<DWORD, std::pair<float, float>>                    CModelInfoSA::ms_VehicleModelDefaultWheelSizes;
 
+union tIdeFlags
+{
+    struct
+    {
+        char bIsRoad : 1;
+        char bFlag2 : 1;
+        char bDrawLast : 1;
+        char bAddictive : 1;
```
StrixG: Typo
```suggestion
        char bAdditive : 1;
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -19,12 +19,50 @@ CBaseModelInfoSAInterface** ppModelInfo = (CBaseModelInfoSAInterface**)ARRAY_Mod
...
+        char bIsTree : 1;
+        char bIsPalm : 1;
+        char bDontCollideWithFlyer : 1;
+
+        char bFlag17 : 1;
+        char bFlag18 : 1;
+        char bFlag19 : 1;
+        char bFlag20 : 1;
+        char bIsTag : 1;
+        char bDisableBackfaceCilling : 1;
```
StrixG:
```suggestion
        char bDisableBackfaceCulling : 1;
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,19 +521,252 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+    {
+        case eModelIdeFlag::IS_ROAD:
+            m_pInterface->bIsRoad = bState;
+            break;
+        case eModelIdeFlag::DRAW_LAST:
+            m_pInterface->bAlphaTransparency = bState;
+            if (bState)
+                m_pInterface->bAdditiveRender = true;
+            break;
+        case eModelIdeFlag::ADDICTIVE:
```
StrixG:
```suggestion
        case eModelIdeFlag::ADDITIVE:
```

Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
```diff
@@ -666,6 +666,26 @@ ADD_ENUM(eClientModelType::OBJECT, "object")
 ADD_ENUM(eClientModelType::VEHICLE, "vehicle")
 IMPLEMENT_ENUM_CLASS_END("client-model-type")
 
+IMPLEMENT_ENUM_CLASS_BEGIN(eModelIdeFlag)
+ADD_ENUM(eModelIdeFlag::IS_ROAD, "is_road")
+ADD_ENUM(eModelIdeFlag::DRAW_LAST, "draw_last")
+ADD_ENUM(eModelIdeFlag::ADDICTIVE, "addictive")
```
StrixG:
```suggestion
ADD_ENUM(eModelIdeFlag::ADDITIVE, "additive")
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -483,19 +521,252 @@ BOOL CModelInfoSA::DoIsLoaded()
...
+    if (!m_pInterface)
+        return false;
+
+    switch (eIdeFlag)
+    {
+        case eModelIdeFlag::IS_ROAD:
+            return m_pInterface->bIsRoad;
+        case eModelIdeFlag::DRAW_LAST:
+           return m_pInterface->bAlphaTransparency;
+        case eModelIdeFlag::ADDICTIVE:
```
StrixG:
```suggestion
        case eModelIdeFlag::ADDITIVE:
```

Client/sdk/game/CModelInfo.h
```diff
@@ -28,6 +28,27 @@ class CBoundingBox
     float   fRadius;
 };
 
+enum class eModelIdeFlag
+{
+    IS_ROAD,
+    DRAW_LAST,
+    ADDICTIVE,
```
StrixG:
```suggestion
    ADDITIVE,
```

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: i think current state is more readable

TheNormalnij:
conflicts solved, can be merged

lopezloo:
Do all these flags actually work? `is_crane` for example. What does it do?

TheNormalnij:
I saw almost all flags in IDA, but im not sure that all of them have effect in MTA. My "must have" list:
```
draw_last -- fix world models culling
addictive -- fix render issues
ignore_lighting
no_zbuffer_write
dont_receive_shadows
is_palm
```

Shuubaru:
i'm curious, will this fix vehicle/peds shadows over non-world objects created with **createObject** function?

Allerek:
Is there any docs about what these flags do? We can guess some for sure like
`is_palm` is probably about bending the object while wind is strong etc.

lopezloo:
> i'm curious, will this fix vehicle/peds shadows over non-world objects created with **createObject** function?

It won't. MTA created objects are treated by game as small dynamic objects like lampposts etc. and that's why they do not reflect shadows.

Lpsd:
As briefly mentioned on Discord - I think the API needs simplifying and the usage being more "consistent".

We should have only 2 functions for setting and getting flag(s). Each function you can either pass a single flag ID, or a table of flag IDs (ints).

Also, both functions should accept/return the flag IDs (ints). This is the consistency part, as currently you have to work with both ints as well as string names, depending on if you're setting one, or multiple flags.

We can have a table of the flag IDs and their names on the wiki.

This also gets rid of the need for that `bIdeFlags` arg.

Ideally, the functions would be:

```lua
table engineGetModelFlags( int modelID, int flagID / table flagIDs )
bool engineSetModelFlags( int modelID, int flagID / table flagIDs, bool state )
```

TheNormalnij:
As i remember, GTA SA translates flags from IDE files in internal format, and we can't translate GTA format in IDE format.
I think map developers know better ide flags and will be discouraged with internal gta system.
Also people will use precalculated flags from ide files or MTA editors, and this is main usage, i think.

So, user friendly API should handle well-known information in this situation.

As Pirulax said once "No magic numbers in API". We have some unfriendly examples like `get/setVehicleWheelsSates`. I always check wiki about these functions. In my opinion `flagID` will be fast, but harder to use.

Current API is perfect for 2 types of operations:
1) You need set flags for a new object. You use `engineSetModelFlags` with your precalculated value or you get this value from `bitOr` with enabled flags from IDE or GTA.
2) You need tweak objects. If you need change only one flag, or disable/enable some flags for all models, `engineSetModelFlag` will help you.

Zangomangu:
> As i remember, GTA SA translates flags from IDE files in internal format, and we can't translate GTA format in IDE format.

If we can't translate it back, I don't think we should expose it. Most people would set flags in the IDE format, because that's the documented format.
So you set a value, but you get a different value. That causes confusion. Is there any way we can just return the IDE formatted flags? GTA shouldn't be changing the model flags, so only this function would be altering them?


As pointed out above, current scripting functions take alternative parameters instead of having several functions. So we should keep it to one set/get/reset IMO.

I also agree we shouldn't come up with magic numbers, so what about the following.
Scripters not familiar with bitwise operations can use the string variant to set/get individual flags.

`int flags` for the IDE flags.
`string flagName` for a specific flag.

 ```lua
 int/bool engineGetModelFlags( int modelID [, string flagName ] )
 ```
Returns IDE flags, or state of a specific flag if flagName is specified.
Function should error if parsed an invalid modelID to avoid ambiguity.

 ```lua
 bool engineSetModelFlags( int modelID, int flags / string flagName [, bool state ] )
 ```
Sets IDE flags, or a specific flag if flagName and state is specified

gta191977649:
> > As i remember, GTA SA translates flags from IDE files in internal format, and we can't translate GTA format in IDE format.
> 
> If we can't translate it back, I don't think we should expose it. Most people would set flags in the IDE format, because that's the documented format. So you set a value, but you get a different value. That causes confusion. Is there any way we can just return the IDE formatted flags? GTA shouldn't be changing the model flags, so only this function would be altering them?
> 
> As pointed out above, current scripting functions take alternative parameters instead of having several functions. So we should keep it to one set/get/reset IMO.
> 
> I also agree we shouldn't come up with magic numbers, so what about the following. Scripters not familiar with bitwise operations can use the string variant to set/get individual flags.
> 
> `int flags` for the IDE flags. `string flagName` for a specific flag.
> 
> ```lua
> int/bool engineGetModelFlags( int modelID [, string flagName ] )
> ```
> 
> Returns IDE flags, or state of a specific flag if flagName is specified. Function should error if parsed an invalid modelID to avoid ambiguity.
> 
> ```lua
> bool engineSetModelFlags( int modelID, int flags / string flagName [, bool state ] )
> ```
> 
> Sets IDE flags, or a specific flag if flagName and state is specified

I agreed, most people will use the flags id from corresponding idea file, this feature will benefit especially someone who are trying to do a map total convertion in MTA.

patrikjuvonen:

patrikjuvonen (dismissed):
This is great, thank you very much.

patrikjuvonen (dismissed):

PR #250, opened 2018-07-22
Custom train tracks

qaisjp:
This PR closes #388

PR #1715, opened 2020-10-15
Custom train tracks: initial server awareness

qaisjp:
This **fully backwards compatible** pull request contains initial support for the train track element and the ability to change between default train tracks. This is the first in a series of pull requests that intends to pick apart the `feature/custom-train-tracks` branch in #250, so that it can be confidently reviewed and merged without worrying about bugs too much.

You can get default track elements using the below documentation. This documentation is currently only accurate server-side. Client-side it will just return the same number you provide, as clients do not yet have a concept of track elements.

```lua
-- OOP: TrainTrack.getDefaultTrack(trackID)
train-track getDefaultTrack(number trackID)

-- trackID: the ID of the default track, one of `0`, `1`, `2`, `3`.
```

And then set (or get) the track of a train using the following functions:

```lua
-- OOP: v:setTrack(t)
bool setTrainTrack(vehicle v, train-track t)

-- OOP: v:getTrack(v)
train-track getTrainTrack(vehicle v)
```

I have tested sync with one and two players, and track and position _syncing_ seems to work great. The only client changes this PR brings is exposing existing client to the Lua API (a partial revert of 89b5620 / #1713).

The client-side `setTrainTrack` function is not protected, and can run on server trains. This is consistent with the behaviour of `setTrainPosition`, `setTrainDirection` and `setTrainSpeed`.

Finally, since this code is based on the original `feature/custom-train-tracks` branch:

Co-authored-by: Jusonex <jusonex96@gmail.com>
Co-authored-by: Cazomino05 <Cazomino05@gmail.com>
Co-authored-by: "sebasdevelopment@gmx.com" <sebasdevelopment@gmx.com>


ghost (approved):
I'm very happy to see this. Nice work, mate.

PR #1716, opened 2020-10-15
Save server row index instead of finding it

ecastro98:
Saving/Caching the row index of each server in server list allow us to avoid making unnecesary for-loop (for each server) to find if the row already exists. For example on the first load, where the first action is just to add (not update nor remove), avoiding this will increase the performance a lot and it won't get so much time to load.

### DIFF
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -833,7 +833,7 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
         m_pServerPlayerList[Type]->Clear();
     }
 
-    bool didUpdateRowIndexes = false;
+    bool didUpdateRowIndices = false;
 
     // Loop the server list
     for (CServerListIterator it = pList->IteratorBegin(); it != pList->IteratorEnd(); it++)
@@ -850,10 +850,10 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
         // Add/update/remove the item to the list
         if (pServer->revisionInList[Type] != pServer->uiRevision || bClearServerList)
         {
-            if (!didUpdateRowIndexes)
+            if (!didUpdateRowIndices)
             {
                 UpdateRowIndexMembers(Type);
-                didUpdateRowIndexes = true;
+                didUpdateRowIndices = true;
             }
 
             pServer->revisionInList[Type] = pServer->uiRevision;
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -821,7 +821,8 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
     // Get the appropriate server list
     CServerList* pList = GetServerList(Type);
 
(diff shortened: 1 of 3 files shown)


Woovie:
Thank you so much for this PR. I was trying to figure out where the caching opportunity was here on my own but my unfamiliarity with c++ really hindered me.

Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -832,18 +833,29 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
         m_pServerPlayerList[Type]->Clear();
     }
 
+    bool didUpdateRowIndexes = false;
```
qaisjp:
```suggestion
    bool didUpdateRowIndices = false;
```

Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -832,18 +833,29 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
...
+            pServer->iRowIndex = -1;
+
         // Find info from server cache for favourites and recent
         if (Type == ServerBrowserType::FAVOURITES || Type == ServerBrowserType::RECENTLY_PLAYED)
             GetServerCache()->GetServerCachedInfo(pServer);
 
         // Add/update/remove the item to the list
         if (pServer->revisionInList[Type] != pServer->uiRevision || bClearServerList)
         {
+            if (!didUpdateRowIndexes)
```
qaisjp:
```suggestion
            if (!didUpdateRowIndices)
```

Client/core/ServerBrowser/CServerBrowser.cpp
```diff
@@ -832,18 +833,29 @@ void CServerBrowser::UpdateServerList(ServerBrowserType Type, bool bClearServerL
...
+            {
+                UpdateRowIndexMembers(Type);
+                didUpdateRowIndexes = true;
```
qaisjp:
```suggestion
                didUpdateRowIndices = true;
```

Client/core/ServerBrowser/CServerList.h
```diff
@@ -141,6 +141,8 @@ class CServerListItem
         bSerials = false;
         bPassworded = false;
         bKeepFlag = false;
+        iRowIndex = -1;
```
qaisjp, line 144: Does this need to be here (and not just in the declaration as `int iRowIndex = 1;`) because `CServerListItem` instances are pooled/reused?

qaisjp (approved):
This is so gooood! It feels like a 100x browser load speedup! Amazing.

I left a tiny nit, but other than that, it looks great :)

Woovie (changes requested):
Once the changes requested by qaisjp are done, we will merge.

Client/core/ServerBrowser/CServerList.h
ecastro98: I am not familiar with C ++ either, and so I don't know the difference. I wrote it like this based on the existing pattern. Should I add the default value to the property instead?

qaisjp: Lets leave it like this for consistency. We can change it later.

ecastro98: Cool

Woovie (approved):

PR #1720, opened 2020-10-16
Quaternion extension

tederis:
Quaternion class extension. SSE2 included.

This PR is intended to extend an existing OOP by the new Quaternion class. Now this is only a draft so any suggestion would be welcome.

Work in progress.

### DIFF
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaMatrixDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaMatrixDefs.cpp
@@ -32,12 +32,14 @@ void CLuaMatrixDefs::AddClass(lua_State* luaVM)
 
     lua_classfunction(luaVM, "getPosition", GetPosition);
     lua_classfunction(luaVM, "getRotation", GetRotation);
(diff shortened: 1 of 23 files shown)


Pirulax:
Why not just have 2 sections of code, one for `SSE` constructors / operators`, and one for the standard way? Eg.:
```cpp
#ifdef SSE_ENABLED
// all SSE constructors / ops
#else
// all standard way ctors / ops
#endif
```

Shared/mods/deathmatch/logic/luadefs/CLuaMatrixDefs.cpp
```diff
@@ -91,7 +95,7 @@ int CLuaMatrixDefs::Create(lua_State* luaVM)
             argStream.ReadVector3D(vecRotation);
             ConvertDegreesToRadiansNoWrap(vecRotation);
             matrix = CMatrix(vecPosition, vecRotation);
-        }
+        }
```
AlexTMjugador, line 98: I think you've introduced some unwanted whitespace here 🙂

tederis: Oh, my bad

PR #1726, opened 2020-10-17
Fix setCameraTarget minimum required client version

Woovie:
Caused by merging #1398

PR #1727, opened 2020-10-17
Fix setCameraTarget minimum required server version

Woovie:
#1398 and #1726 related issue. This one is server.


Woovie:
I force merged this as it is time sensitive and I could not get a reviewer in time.

PR #1737, opened 2020-10-18
Use enum classes for BitStream

qaisjp:
This pull request allows for querying BitStream compatibility by doing (for example) `bitStream.Can(eBitStreamVersion::FakeLagCommand)` instead of getting the number and doing comparisons. You just need to add to the enum list.

I propose that all future code uses this system as much as possible, instead of doing `GetVersion`.

There is [a second pull request](https://github.com/qaisjp/mtasa-blue/compare/qaisjp/bitstream-enum-class...qaisjp:bitstream-apply-enum-classes) ready that applies the use of this feature to every change since the release of 1.5. I'll submit that pull request if/once this one merges. 

**Note**: this pull request also increments the BitStream version. This is harmless, right?

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -1067,7 +1067,7 @@ void CClientGame::DoPulses()
             m_bWaitingForLocalConnect = false;
 
             // Assume local server has the same bitstream version
-            g_pNet->SetServerBitStreamVersion(MTA_DM_BITSTREAM_VERSION);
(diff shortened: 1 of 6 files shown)


ghost (approved):
> Note: this pull request also increments the BitStream version. This is harmless, right?

I checked it. It should be okay.

Dutchman101:
I advise no one to use the latest nightly, as this change seems to have broken connecting to servers:

Untitled
@ccw808 was notified (we don't really need it on the tracker) - as it probably means he needs to update something in netc according to these changes.

PR #1741, opened 2020-10-19
Revert "Allow to use peds and vehicles as a camera target (#1398)"

qaisjp:
Causes weird camera focusing desync issues that look like this

https://cdn.discordapp.com/attachments/588804725802270773/767816132030758982/2020-10-19_11-26-18.mp4

and 

https://cdn.discordapp.com/attachments/588804725802270773/767849210694336555/MTA__San_Andreas_2020-10-19_21-37-52.mp4

### DIFF
--- a/Client/mods/deathmatch/logic/CNetAPI.cpp
+++ b/Client/mods/deathmatch/logic/CNetAPI.cpp
@@ -1981,11 +1981,11 @@ void CNetAPI::WriteCameraSync(NetBitStreamInterface& BitStream)
     {
         // Write our target
         ElementID      ID = INVALID_ELEMENT_ID;
-        CClientEntity* pTarget = pCamera->GetTargetEntity();
-        if (!pTarget)
-            pTarget = g_pClientGame->GetLocalPlayer();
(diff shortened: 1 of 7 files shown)


Woovie (approved):

StrixG (approved):

qaisjp:
I am too trigger happy. This doesn't fix the issue

qaisjp:
@TheNormalnij please can you create a PR that reverts 84a30b8 and fixes net compatibility issues? When the server doesn't have your commit, but the client does have your commit, there are desync problems. 

Comments of 64f200fe8fd304aa3f01 say:

> This appears to possibly be causing some issues with spectating. Not 100% confirmed yet, but using the build which includes this vs the previous build, we are now getting issues with sync.
>
> [Before](https://cdn.discordapp.com/attachments/588804725802270773/767816132030758982/2020-10-19_11-26-18.mp4), [After](https://cdn.discordapp.com/attachments/588804725802270773/767817698016362556/2020-10-19_11-32-46.mp4)
>
> \- Woovie

PR #1742, opened 2020-10-20
Fix #341: Make exhaust fumes vehicle dependent

### LINKED ISSUE #341: setVehicleModelExhaustFumesPosition for vehicle-specific
Currently it's model-specific, which interferes with tuning-features on certain servers. Vehicle-specific enables more extreme tuning options for players and modellers on servers.

Fumes position should be vehicle-specific, so each player has their own exhaust fume position applied.

Additionally, an option for one fume exhaust for cars with two exhaust fumes would be appreciated (currently available in setVehicleHandling).

Mantis issue: https://bugs.mtasa.com/view.php?id=9778
Related GitHub PR (relation, **to keep in mind**): #390

ghost:
Four new functions:
```
bool setVehicleDummyPosition(vehicle theVehicle, string dummy, float x, float y, float z)
float x, float y, float z getVehicleDummyPosition(vehicle theVehicle, string dummy)
bool setVehicleDummyVisible(vehicle theVehicle, string dummy, bool visible)
bool getVehicleDummyVisible(vehicle theVehicle, string dummy)
```
Properties:
```
exhaust1
exhaust2
```
Now you can set exhaust fumes position per vehicle rather than per model, and make them visible/invisible.


Client/sdk/game/Common.h
```diff
@@ -1518,22 +1518,32 @@ enum eVehicleComponent
...
+        EXHAUST_SECONDARY,
+        WING_AIRTRAIL,
+        VEH_GUN,
+        VEH_UNKNOWNDUMMY,
+        TOTAL_INTERNAL = 15,
+
+        // custom dummies for MTA, not part of GTA
+        EXHAUST_LEFT = 15,
+        EXHAUST_RIGHT = 16,
+        TOTAL = 17
```
qaisjp:
```suggestion
        TOTAL_INTERNAL,

        // custom dummies for MTA, not part of GTA
        EXHAUST_LEFT = TOTAL_INTERNAL,
        EXHAUST_RIGHT,
        TOTAL,
```
Does this work?

Client/sdk/game/CAutomobile.h
```diff
@@ -55,8 +57,9 @@ class CAutomobile : public virtual CVehicle
     virtual void       SetRandomDamage(bool bUnknown) = 0;
     virtual void       SetTaxiLight(bool bState) = 0;
     virtual void       SetTotalDamage(bool bUnknown) = 0;
-    virtual void       SetTowLink(CVehicle* pVehicle, bool bRemoveAdd) = 0;
+    //virtual void       SetTowLink(CVehicle* pVehicle, bool bRemoveAdd) = 0;
```
qaisjp: I am confused as to why this is commented out, and someone in the future will look at this and feel the same way. Either remove the line or add a code comment explaining why it's commented.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -223,8 +223,9 @@ class CStaticFunctionDefinitions
     static bool            IsVehicleWindowOpen(CClientVehicle& Vehicle, uchar ucWindow);
     static bool            SetVehicleModelExhaustFumesPosition(unsigned short usModel, CVector& vecPosition);
     static bool            GetVehicleModelExhaustFumesPosition(unsigned short usModel, CVector& vecPosition);
-    static bool            SetVehicleModelDummyPosition(unsigned short usModel, eVehicleDummies eDummy, CVector& vecPosition);
-    static bool            GetVehicleModelDummyPosition(unsigned short usModel, eVehicleDummies eDummy, CVector& vecPosition);
+    static bool            SetVehicleModelDummyPosition(unsigned short usModel, eVehicleDummy::e eDummy, CVector& vecPosition);
+    static bool            GetVehicleModelDummyPosition(unsigned short usModel, eVehicleDummy::e eDummy, CVector& vecPosition);
+    static bool            GetVehicleModelDummyPosition(CClientVehicle* pVehicle, unsigned short usModel, eVehicleDummy::e eDummy, CVector& vecPosition);
```
qaisjp: Can we change this so that "overload selection" only happens in CLuaDefs, and not in CStaticFunctionDefinitions?

Basically, do whatever is required to make this possible:

```suggestion
    static bool            GetVehicleModelDummyPosition(unsigned short usModel, eVehicleDummy::e eDummy, CVector& vecPosition);
    static bool            GetVehicleModelDummyPosition(CClientVehicle* pVehicle, eVehicleDummy::e eDummy, CVector& vecPosition);
```

That way, in the distant future, if we ever migrate this to the new parser, they can just do `ArgumentParserWarn<OverloadA, OverloadB>` instead of figuring out what's going on here.

If I'm not making sense, I can elaborate.

Client/sdk/game/Common.h
ghost: I just added two new functions for vehicle-specific dummies:
```
setVehicleDummyPosition
getVehicleDummyPosition
```
`setVehicleModelDummyPosition` is a model-specific function, so I thought it would be awkward to have vehicle-specific dummies in there. I have also added a new enum for it.

Client/sdk/game/CAutomobile.h
ghost: I removed the comment in the latest commit. CAutomobile class was never used in the client project, so this function was never implemenented in the derived class. After I added support for CAutomobile, it was throwing compile-time errors. I was going to implement it but forgot, removed it for now as it's not used.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
ghost: I think I understand what you mean. I'll try to do it.

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -11,40 +11,106 @@
...
-            pGame->GetWorld()->Add((CEntitySA *)this);
-    */
-    // create our mirror classes
+    m_automobileInterface = static_cast<CAutomobileSAInterface*>(this->GetInterface());
     for (int i = 0; i < MAX_DOORS; i++)
-        this->door[i] = new CDoorSA(&((CAutomobileSAInterface*)this->GetInterface())->m_doors[i]);
-    /*}
-     */
+    {
+        this->door[i] = new CDoorSA(&m_automobileInterface->m_doors[i]);
```
sbx320: Might be good to have the doors direct members of CAutomobileSA, instead of an array of heap-allocated pointers

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -595,3 +661,24 @@ CDoor* CAutomobileSA::GetDoor(eDoors doorID)
...
+    if (!visible)
+    {
+        auto nitroFxSystem = m_automobileInterface->m_exhaustNitroFxSystem[id];
+        if (nitroFxSystem)
+        {
+            nitroFxSystem->Kill();
+            CFxManagerSA::g_fxMan.DestroyFxSystem(nitroFxSystem);
+            m_automobileInterface->m_exhaustNitroFxSystem[id] = nullptr;
+        }
+    }
```
sbx320, line 684: The branch `SetNitroFxSystemVisible(id, true)` seems to be missing here

ghost:
Added two more functions: 
```
setVehicleDummyVisible(vehicle theVehicle, string dummy, bool visible)
bool visible getVehicleDummyVisible(vehicle theVehicle, string dummy)
```
Currently, it only supports two properties `exhaust1` and `exhaust2`. You can now even hide exhaust fumes if you like.

Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
```diff
@@ -564,21 +564,27 @@ ADD_ENUM(D3DPT_TRIANGLESTRIP, "trianglestrip")
...
-ADD_ENUM(SEAT_REAR, "seat_rear")
-ADD_ENUM(EXHAUST, "exhaust")
-ADD_ENUM(ENGINE, "engine")
-ADD_ENUM(GAS_CAP, "gas_cap")
-ADD_ENUM(TRAILER_ATTACH, "trailer_attach")
-ADD_ENUM(HAND_REST, "hand_rest")
-ADD_ENUM(EXHAUST_SECONDARY, "exhaust_second")
-ADD_ENUM(WING_AIRTRAIL, "wing_airtrail")
-ADD_ENUM(VEH_GUN, "veh_gun")
+IMPLEMENT_ENUM_BEGIN(eVehicleModelDummy::e)
```
sbx320, line 567: `::e`?

Client/game_sa/CAutomobileSA.cpp
ghost: It's not needed. `CAutomobileSAInterface::DoNitroEffect` will recreate the fx system if it's not there when vehicle->IsDummyVisible returns true. I'll add a code comment here.

Client/game_sa/CAutomobileSA.cpp
ghost: Good Idea. Done.

Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
ghost: `eVehicleModelDummy` is a namespace and `e` is the actual enum. 

```
namespace eVehicleModelDummy
{
    enum e
    {
        LIGHT_FRONT_MAIN = 0,
        LIGHT_REAR_MAIN,
        ...
    };
}
```
This is an alternative to enum class. If I use enum class here, then I can't use the enum member as array index. This will throw compile-time errors `vecDummies[eVehicleModelDummy::EXHAUST]`.

But with `eVehicleModelDummy::e`, I can do `vecDummies[eVehicleModelDummy::e::EXHAUST]`, and the enum members are not in the global scope. Eeh, it's definitely not the best solution, but I can't think of anything else :P

sbx320: What about making the inner enum a regular `enum` rather than `enum class`?
```
namespace eVehicleModelDummy
{
    enum Enum
    {
        LIGHT_FRONT_MAIN = 0,
        LIGHT_REAR_MAIN,
    };
}
```
This gives us proper scoping into `eVehicleModelDummy`, but still allows using it as an array index. Only downside of this would be that we need to specify `eVehicleModelDummy::Enum` as a type in some places. 

*edit* nvm, that's what you're already doing. 👍

Pirulax: Perhaps just cast when using it as an array index?
eg.:
```cpp
enum class eVehicleModelDummy
{
        LIGHT_FRONT_MAIN = 0,
        LIGHT_REAR_MAIN,
};
// Somewhere in the code...
array[(size_t)eVehicleModelDummy::LIGHT_REAR_MAIN)]
```
You could cast to `int` as well, it doesn't really matter. (In the the compiler will use `size_t` probably..)

`enum class` is strongly typed, thats why it doesnt implicitly cast to a number, while casting explicitly works fine.

ghost: I think that defeats the purpose of using enum class if we're going to use casts. More importantly, I'd like to use static_cast rather than raw casts, but then it just makes the code look complicated.

qaisjp:
This pull request is too big I don't think it's practically reviewable

PR #620, opened 2018-09-30
Add support for driving and being a passenger in client-side vehicles, client-side vehicle damage and animated vehicle entering/exiting through new functions for client-side peds/local players

MegadreamsBE:
### Summary
This pull request makes it possible for players to drive client-side vehicles or be a passenger in them (a ped can be scripted to drive the vehicle). Similarly it also makes it possible to damage client-side vehicles.

While Multi Theft Auto always was a mod to allow multiple players to play with each other it has grown into something so much bigger. This PR gives a greater control to server owners (in the end it is up to them to either allow or not allow a player to be desynced from others).

There are enough occasions where you may require a vehicle to be in there (cutscenes for example or being flown from one airport to another in a plane) that other players do not need to see but that, if we want our player to be in there, requires us to create a server-side vehicle instead. This removes that need. Similarly allowing client-side vehicles from taking on damage from weapons and/or driving makes them more useful overall.

Fixes #1352 

## New features
- It is now possible to get into and drive client-side vehicles as a player.
- It is now possible to damage client-side vehicles unless this is disabled through scripting.
- Added various new functions/events for client-side peds listed below.

## New functions [CLIENT-SIDE ONLY]
- **getPedNearestVehicleEntryPoint(ped, checkDriverDoor, checkPassengersDoors)**
  This function returns a valid entry point (number) for the nearest vehicle to the ped/player and the vehicle itself.
- **getPedEnterVehicleTarget(ped)**
  This function returns the vehicle together with the seat and entry point of the vehicle the ped is trying to enter.
- **getVehicleEntryPoints(vehicle)**
  This functions returns all entry points for a given vehicle with the entry point id (starting at 0) as the index and then the coordinates of the entry point.
- **setPedEnterVehicle(ped, vehicle, seat)**
  This function makes a client-side ped enter a client-side vehicle. This function also works on the local player whom can also enter server-side vehicles with it.
- **setPedExitVehicle(ped)**
  This function makes a client-side ped (or the local player) leave their current vehicle.
- **cancelPedEnterVehicle(ped)**
  This function makes a client-side ped (or the local player) stop trying to enter a vehicle.

## New events [CLIENT-SIDE ONLY]
- **onClientPedStartEnterVehicle** - ped is the source.
    This event is triggered when a ped is starting to enter a car. It can be cancelled.
    - vehicle: the vehicle the ped is entering.
    - seat: the seat the ped will be in.
    - door: the door the ped is entering through.
[...]

PR #1748, opened 2020-10-23
Implement entering/exiting/jacking for peds

### LINKED ISSUE #373: Ped vehicles are not jackable
**Describe the bug**
You can't hijack a created ped's vehicle.

**To Reproduce**
```
run v = Vehicle(411, me.position + Vector3(1, 1, 1))
run p = Ped(0, 0, 0, 0)
run p.vehicle = v
```

and then try to jack the vehicle

**Expected behavior**
You should be able to get into the vehicle.

**Additional context**
From https://bugs.mtasa.com/view.php?id=4896

### LINKED ISSUE #532: Ped doesn't exit moving vehicle if killed
**Describe the bug**
1. Killing ped in a stationary vehicle makes it exit with death anim as expected.
1. Killing ped in a moving vehicle makes it remain dead behind the wheel even when the vehicle comes to a stop.
1. When the ped is in this state, the car is jackable and it even appears synced.
1. Any enabled control states (eg. accelerate in this video) are disabled when the ped dies, as expected, but they are re-enabled when you start jacking the car causing the last issue in the video.

**Expected behaviour**
A dead ped should always exit the vehicle when it comes to a stop.

**Screenshots**
Video to show issue: https://streamable.com/a8k9q

**MTA Client/Server:** 
 - Version: 1.5.4

**Additional context**
This happens regardless if killed by killPed or weapon.

From https://bugs.mtasa.com/view.php?id=9707

### LINKED ISSUE #535: setPedControlState doesn't work on the control "enter_exit"
**Describe the bug**
I can't use "setPedControlState" with the parameter "enter_exit". There is no error or warning, the ped is just doing nothing, while controls like "jump", "sprint" or "forwards" work. Ped and vehicle are both created server-side.

**To reproduce**
Unstated.

**Expected behaviour**
Be able to cause ped to enter or exit vehicle.

**Screenshots**

**MTA Client:** 
Unstated.

**MTA Server:** 
Unstated.

**Additional context**

From https://bugs.mtasa.com/view.php?id=5113

### LINKED ISSUE #1278: Fix ped fall off bike
**Describe the bug**
Ped falling off bike is not currently synced.
setPedCanBeKnockedOffBike works fine for players, but for peds they only fall off the bike locally and never actually exit the vehicle.

**To reproduce**
Create a ped on a bike and ram it.

**Expected behaviour**
The effect should be synced for all players and they should exit the vehicle.

**Screenshots**

**Version**

**Additional context**
Probably related to a bigger issue on peds enter / exit of vehicles.

Zangomangu:
PR adds entering/exiting/jacking mechanism for peds with other peds and players.
All code related to enter/exit is refactored from CClientGame into CClientPed.
Enter/exit is a sequence of packets between client and server. For peds, the syncing client handles it for them.

Fixes #535, #373 and #532.
If PR is added I can *fix* #1278.

### Network compatibility

To retain network compatibility, bitstream checks make sure the following is compatible:
- New client, old server.
- Old client, new server.
- New client, new server, old remote client.
- New client, new server, old remote client joins during enter/exit.

Peds wont be able to enter/exit if there is an old client on the server.
If one or more old clients join during a ped enter/exit, a warp in/out RPC packet is sent to them in specific situations so they remain synced.

### Lua API

Originally I was going to expand the enter_exit control to work for peds instead of introducing a function. 
But enter_passenger is not a gta control and with a Lua function we can do more in scripting.

**2 clientside functions added**
`bool setPedEnterVehicle(ped [, vehicle = nil, passenger = false ])`
- Works on local player and peds that are synced by the client.
- Without specifying a vehicle, ped will enter the closest vehicle, the same behavior as pressing enter_exit.
- When specifying a vehicle, ped will enter vehicle if possible.
- When passenger = true, ped will enter as passenger, same behavior as pressing G (MTA enter_passenger).
- When this function returns true, the client will ask server for permission to enter a vehicle. Actually entering happens after the server acknowledges the entry.
- - It can still fail in case the server rejects the entry, fe. due to network compatibility.
- - A task interrupts the entering, or a task is blocking the entering task from starting. 

`bool setPedExitVehicle(ped)`
- Works on local player and peds that are synced by the client.
- When this function returns true, the client will ask server for permission to exit the vehicle. Actually entering happens after the server acknowledges the exit.

Both functions aren't guaranteed to make the ped enter or exit the vehicle. Different game events can cancel the tasks. In scripting, you can use getPedTask to monitor the task system and how far the ped is in his entry / exit.

**2 clientside events added**
- onClientPedVehicleEnter
- - Ped variant of onClientPlayerVehicleEnter
- onClientPedVehicleExit
- - Ped variant of onClientPlayerVehicleExit

**4 clientside events changed**
onClientVehicleStartEnter
onClientVehicleEnter
onClientVehicleStartExit
onClientVehicleExit
- Player parameters can now be a ped

**2 serverside events added**
- onPedVehicleEnter
- - Ped variant of onPlayerVehicleEnter
- onPedVehicleExit
- - Ped variant of onPlayerVehicleExit

**4 serverside events changed**
onVehicleStartEnter
onVehicleEnter
onVehicleStartExit
onVehicleExit
- Player parameters can now be a ped

### Compatibility issues with existing API

Currently when using warpPedIntoVehicle and removePedFromVehicle, the onClientPlayer / onPlayer vehicle events are called also for peds.
Now these functions trigger the ped variants instead.
Existing scripts could rely on this practice, but it's unlikely imo.
We could also keep the current behavior by triggering both events for the ped and then make the change in 1.6.


PlatinMTA:
Good job, hope to see this in prime

MegadreamsBE:
This PR needs coordination with the following PR: https://github.com/multitheftauto/mtasa-blue/pull/620

qaisjp:
> This PR needs coordination with the following PR: #620

What in particular do you think should be coordinated? The Lua API?

MegadreamsBE:
> > This PR needs coordination with the following PR: #620
> 
> What in particular do you think should be coordinated? The Lua API?

Mainly yes but also the code changes both PRs introduce. These changes currently are incompatible so whichever PR comes second (likely mine I guess) needs its code modified. Although that part will have to be tracked in my PR.

For the Lua API I propose the following changes:
setPedEnterVehicle - Allow for a seat number rather than whether or not its a passenger/driver seat.

I'll ensure my PR is compatible with everything else in here.

Question: Does setPedEnterVehicle work on client-side peds to client-side vehicles as well in this PR or is that still exclusive to mine?

Zangomangu:
> For the Lua API I propose the following changes:
setPedEnterVehicle - Allow for a seat number rather than whether or not its a passenger/driver seat.

That is a good idea. But I would prefer if you could still make it find a seat automatically.
So I see two solutions:
- bool/int seat
- - Parse true for automatic passenger seat entry, false (or nil) for driver seat entry, and a seat number for specific seat
- int seat
- - Parse 0 for driver, seat number for specific seat, and -1 (or nil?) for automatic passenger seat?

> Question: Does setPedEnterVehicle work on client-side peds to client-side vehicles as well in this PR or is that still exclusive to mine?

It doesn't right now, if you look at CClientPed L6420 in the PR, I have the check for local entity. I assume most of the EnterVehicle and ExitVehicle method can be used for local peds, since it involves finding a valid vehicle/seat/door and checking for some tasks. So basically right before we allocate bitstream, we could have a check for local ped and initiate the enter/exit.

MegadreamsBE:
> > For the Lua API I propose the following changes:
> > setPedEnterVehicle - Allow for a seat number rather than whether or not its a passenger/driver seat.
> 
> That is a good idea. But I would prefer if you could still make it find a seat automatically.
> So I see two solutions:
> 
> * bool/int seat
> * * Parse true for automatic passenger seat entry, false (or nil) for driver seat entry, and a seat number for specific seat
> * int seat
> * * Parse 0 for driver, seat number for specific seat, and -1 (or nil?) for automatic passenger seat?

I'd go for the second option. I'll integrate that into my stuff as well then.

qaisjp:
> or nil

Is there a use case for being able to enter the closest vehicle via the closest door without caring if it's driver or passenger?

Zangomangu:
> Is there a use case for being able to enter the closest vehicle via the closest door without caring if it's driver or passenger?

No, I don't see how thats useful. It can be handled through scripting, by calling the function twice.

So we can do
0: Enter as driver
1 - max passenger: Enter as passenger in that seat, if it's available.
Nil: Enter as passenger, into nearest available seat.

If we down the line implement ability to jack passenger seats, we can keep the same API, and scripts can check if a seat is occupied beforehand, in case they dont want to jack it.

Zangomangu:
I've looked into making the above change, and it's not very simple to do.
If you want to enter into a specific seat, you need to enter through a specific door. And we don't have this information. We can manually map what doors give access to which seats (or find the data somewhere), eg. bus has 1 door that gives access to many seats, and cars have 2 doors that give access to the same seat. 

We use GTA functions right now for finding a door and seat, but we can only specify whether we want driver or passenger.
Maybe there is a GTA function to find the closest door for a seat? Someone with reversing knowledge will have to look at that.

Either way the change would make this PR even bigger, and it should go in a separate PR imo.
So I'm keeping this PR as it is now

MegadreamsBE:
> I've looked into making the above change, and it's not very simple to do.
> If you want to enter into a specific seat, you need to enter through a specific door. And we don't have this information. We can manually map what doors give access to which seats (or find the data somewhere), eg. bus has 1 door that gives access to many seats, and cars have 2 doors that give access to the same seat.
> 
> We use GTA functions right now for finding a door and seat, but we can only specify whether we want driver or passenger.
> Maybe there is a GTA function to find the closest door for a seat? Someone with reversing knowledge will have to look at that.
> 
> Either way the change would make this PR even bigger, and it should go in a separate PR imo.
> So I'm keeping this PR as it is now

You don't have to actually specify the door though to get the game to move a player to a specific seat. In my PR that is working perfectly fine. The game actually seems to ignore the door setting anyways (which is why I haven't been able to implement that part yet).

qaisjp (approved):

Server/mods/deathmatch/logic/CPed.cpp
```diff
@@ -83,6 +85,23 @@ CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel)
...
+            {
+                m_pJackingVehicle->SetOccupant(NULL, 0);
+                pOccupant->SetOccupiedVehicle(NULL, 0);
+                pOccupant->SetVehicleAction(VEHICLEACTION_NONE);
+            }
+        }
+        if (m_pJackingVehicle->GetJackingPed() == this)
+            m_pJackingVehicle->SetJackingPed(NULL);
+    }
+
```
qaisjp, line 104: Note for secondary reviewer: [moved from CPlayer.cpp](https://github.com/multitheftauto/mtasa-blue/pull/1748/files#diff-77128262cd6c6dde2b422d2058afaf5399e73f9c64eb5974d133296b2935a252L148-L163)

Server/mods/deathmatch/logic/CPed.cpp
```diff
@@ -435,5 +454,66 @@ void CPed::SetSyncer(CPlayer* pPlayer)
...
+        m_pJackingVehicle = NULL;
+        pPrev->SetJackingPed(NULL);
+    }
+
+    // Set new
+    m_pJackingVehicle = pVehicle;
+
+    if (m_pJackingVehicle)
+        m_pJackingVehicle->SetJackingPed(this);
+}
```
qaisjp, line 519: Note for reviewer: [moved from CPlayer.cpp](https://github.com/multitheftauto/mtasa-blue/pull/1748/files#diff-77128262cd6c6dde2b422d2058afaf5399e73f9c64eb5974d133296b2935a252L1110-L1128)

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -259,6 +259,9 @@ class CPed : public CElement
     void     SetSyncable(bool bSynced) { m_bSyncable = bSynced; };
     CPlayer* m_pSyncer;
 
+    CVehicle* GetJackingVehicle() { return m_pJackingVehicle; }
+    void      SetJackingVehicle(CVehicle* pVehicle);
+
     bool IsStealthAiming() { return m_bStealthAiming; }
     void SetStealthAiming(bool bAiming) { m_bStealthAiming = bAiming; }
```
qaisjp, line 266: Note for reviewer: this (and the member `m_pJackingVehicle`) has been [moved from CPlayer.h](https://github.com/multitheftauto/mtasa-blue/pull/1748/files#diff-1140b2ac5bd9ec6021d0363fb4db78c9f5218a8561eb38272b87000adb42f5b3L263-L265)

Client/version.h
```diff
@@ -74,7 +74,7 @@
 #define _NETCODE_VERSION_BRANCH_ID      0x4         // Use 0x1 - 0xF to indicate an incompatible branch is being used (0x0 is reserved, 0x4 is trunk)
 #define _CLIENT_NET_MODULE_VERSION      0x0AB       // (0x000 - 0xfff) Lvl9 wizards only
 #define _NETCODE_VERSION                0x1DA       // (0x000 - 0xfff) Increment when net messages change (pre-release)
-#define MTA_DM_BITSTREAM_VERSION        0x06F       // (0x000 - 0xfff) Increment when net messages change (post-release). (Changing will also require additional backward compatibility code).
+#define MTA_DM_BITSTREAM_VERSION        0x070       // (0x000 - 0xfff) Increment when net messages change (post-release). (Changing will also require additional backward compatibility code).
```
qaisjp: - [x] You'll need to migrate to the new bitstream versioning system

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -91,6 +91,8 @@ void CLuaPedDefs::LoadFunctions()
         {"setPedArmor", ArgumentParser<SetPedArmor>},
         {"givePedWeapon", GivePedWeapon},
         {"isPedReloadingWeapon", IsPedReloadingWeapon},
+        {"setPedEnterVehicle", SetPedEnterVehicle},
+        {"setPedExitVehicle", SetPedExitVehicle},
```
qaisjp: - [x] You need to add OOP definitions

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -737,4 +745,16 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
...
+    unsigned long  m_ulLastVehicleInOutTime;    // Last tick where we sent an enter/exit request
+    bool           m_bIsGettingOutOfVehicle;    // Indicates we are exiting a vehicle
+    bool           m_bIsGettingIntoVehicle;     // Indicates we are entering a vehicle
+    bool           m_bIsJackingVehicle;         // Indicates we are jacking a vehicle
+    bool           m_bIsGettingJacked;          // Indicates we are getting jacked
+    ElementID      m_VehicleInOutID;            // ElementID of vehicle received from server
+    unsigned char  m_ucVehicleInOutSeat;        // Seat ID we are entering/exiting received from server
+    bool           m_bNoNewVehicleTask;         // When set, we are not allowed to initiate a new enter/exit task because we are waiting for server reply
+    ElementID      m_NoNewVehicleTaskReasonID;  // ElementID of the vehicle that we are waiting on
+    CClientPed*    m_pGettingJackedBy;          // The ped that is jacking us
```
qaisjp, line 760: Note for reviewer: [moved from CClientGame.h](https://github.com/multitheftauto/mtasa-blue/pull/1748/files#diff-f6a34fb7f68a6430df09e588e8eee796ec2e26fbe777439d96c74d82cd2ace34L725-L736)

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -561,6 +563,11 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
 
 public:
     void _GetIntoVehicle(CClientVehicle* pVehicle, unsigned int uiSeat, unsigned char ucDoor);
+    // Used to control and sync entering/exiting
+    bool EnterVehicle(CClientVehicle* pVehicle, bool bPassenger);
+    bool ExitVehicle();
+    void ResetVehicleInOut();
+    void UpdateVehicleInOut();
```
qaisjp, line 571: The last two are directly from CClientGame.h. I think `EnterVehicle` and `ExitVehicle` are split out of the original file's `ProcessVehicleInOutKey`.

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -168,6 +168,8 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
     const CEntity* GetGameEntity() const { return m_pPlayerPed; }
 
     bool IsLocalPlayer() { return m_bIsLocalPlayer; }
+    bool IsSyncing() { return m_bIsSyncing; }
+    void SetSyncing(bool bIsSyncing) { m_bIsSyncing = bIsSyncing; };
```
qaisjp: - [x] @Zangomangu there's a couple places where we're essentially using `if (syncing || isLocalPlayer)`. I wonder if we could just set `bIsSyncing = true` for the local player... I'm not sure if this is appropriate though.

Note for reviewer: These are new. And the corresponding member.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -179,6 +179,18 @@ void CClientPed::Init(CClientManager* pManager, unsigned long ulModelID, bool bI
...
+    m_bIsGettingOutOfVehicle = false;
+    m_bIsGettingIntoVehicle = false;
+    m_bIsGettingJacked = false;
+    m_bIsJackingVehicle = false;
+    m_bNoNewVehicleTask = false;
+    m_VehicleInOutID = INVALID_ELEMENT_ID;
+    m_NoNewVehicleTaskReasonID = INVALID_ELEMENT_ID;
+    m_pGettingJackedBy = NULL;
+    m_ucVehicleInOutSeat = 0xFF;
+    m_bIsSyncing = false;
```
qaisjp, line 192: Note for reviewer

All except the last is from CClientGame.cpp.

The last is just a new syncing variable 👍

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -75,14 +75,6 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
...
     m_bTriggeredIngameAndConnected = false;
     m_bGracefulDisconnect = false;
-    m_ulLastVehicleInOutTime = 0;
-    m_bIsGettingOutOfVehicle = false;
-    m_bIsGettingIntoVehicle = false;
-    m_bIsGettingJacked = false;
-    m_bIsJackingVehicle = false;
-    m_VehicleInOutID = INVALID_ELEMENT_ID;
-    m_pGettingJackedBy = NULL;
-    m_ucVehicleInOutSeat = 0xFF;
```
qaisjp, line 85: Note to reviewer: moved to CClientPed.cpp

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -100,9 +92,6 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
     m_lastWeaponSlot = WEAPONSLOT_MAX;            // last stored weapon slot, for weapon slot syncing to server (sets to invalid value)
     ResetAmmoInClip();
 
-    m_bNoNewVehicleTask = false;
-    m_NoNewVehicleTaskReasonID = INVALID_ELEMENT_ID;
-
```
qaisjp, line 105: Note to reviewer: moved to CClientPed.cpp

Server/mods/deathmatch/logic/CVehicle.cpp
```diff
@@ -104,38 +104,35 @@ CVehicle::CVehicle(CVehicleManager* pVehicleManager, CElement* pParent, unsigned
...
-                    pPlayer->SetOccupiedVehicle(NULL, 0);
-                    pPlayer->SetVehicleAction(CPlayer::VEHICLEACTION_NONE);
+                    pPed->SetOccupiedVehicle(NULL, 0);
+                    pPed->SetVehicleAction(CPed::VEHICLEACTION_NONE);
 
                     // Tell everyone he has exited the vehicle
-                    CVehicleInOutPacket Reply(GetID(), ucOccupiedSeat, CGame::VEHICLE_NOTIFY_OUT_RETURN);
-                    Reply.SetSourceElement(pPlayer);
+                    CVehicleInOutPacket Reply(pPed->GetID(), GetID(), ucOccupiedSeat, CGame::VEHICLE_NOTIFY_OUT_RETURN);
                     g_pGame->GetPlayerManager()->BroadcastOnlyJoined(Reply);
```
qaisjp, line 136: Note for reviewer: previously everything in this is just variable renames, except for this line.

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
```diff
@@ -22,9 +23,10 @@ CVehicleInOutPacket::CVehicleInOutPacket()
     m_ucDoor = 0;
 }
 
-CVehicleInOutPacket::CVehicleInOutPacket(ElementID ID, unsigned char ucSeat, unsigned char ucAction)
+CVehicleInOutPacket::CVehicleInOutPacket(ElementID PedID, ElementID VehicleID, unsigned char ucSeat, unsigned char ucAction)
```
qaisjp, line 26: - [ ] Please respond @Zangomangu 

So, code previously followed this pattern:

```cpp
//                        This is the VehicleID!
//                          |
//                          |
//                          v
CVehicleInOutPacket Reply(GetID(), ucOccupiedSeat, CGame::VEHICLE_NOTIFY_OUT_RETURN);	                    
Reply.SetSourceElement(pPlayer);	
```

The new syntax for this overload is this:


```cpp
//                        New PedID argument!
//                         |
//                         |             This is the VehicleID!
//                         |               |
//                         v               v
CVehicleInOutPacket Reply(pPed->GetID(), GetID(), ucOccupiedSeat, CGame::VEHICLE_NOTIFY_OUT_RETURN);
```


So how come we can't use Reply.SetSourceElement with a source element of the ped trying to enter? Is it because the source can only be a player?

It would be really useful if you could add code comments to each of these overloads explaining the difference between them. I'm sure you had to figure these out, and maybe you have it written down somewhere in your notes. It would be great if you could include explanations here.

The original code is terribly undocumented/uncommented considering how integral it is to MTA and how complicated it is in how it all fits together.

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
```diff
@@ -69,10 +73,24 @@ CVehicleInOutPacket::~CVehicleInOutPacket()
...
+        BitStream.Read(m_PedID);
+        if (m_PedID == INVALID_ELEMENT_ID)
+        {
+            return false;
+        }
+    }
+    else
+    {
+        m_PedID = GetSourcePlayer()->GetID();
+    }
```
qaisjp: Extra note to reviewer: pretty much everything is just variable renames, except for this chunk of code, and the other comments in this file.

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
```diff
@@ -120,24 +138,22 @@ bool CVehicleInOutPacket::Read(NetBitStreamInterface& BitStream)
...
 {
-    if (m_pSourceElement && m_ID != INVALID_ELEMENT_ID)
+    if (m_PedID != INVALID_ELEMENT_ID && m_VehicleID != INVALID_ELEMENT_ID)
     {
-        ElementID ID = m_pSourceElement->GetID();
-        BitStream.Write(ID);
-
-        BitStream.Write(m_ID);
+        BitStream.Write(m_PedID);
+        BitStream.Write(m_VehicleID);
```
qaisjp, line 144: > Peds wont be able to enter/exit if there is an old client on the server.

I am still reading the pull request so I hope I don't miss this being enforced elsewhere

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4115,7 +4115,10 @@ bool CStaticFunctionDefinitions::WarpPedIntoVehicle(CPed* pPed, CVehicle* pVehic
                     else
                         PlayerVehicleArguments.PushBoolean(false);
                     // Leave onPlayerVehicleEnter for backwards compatibility
-                    pPed->CallEvent("onPlayerVehicleEnter", PlayerVehicleArguments);
+                    if (IS_PLAYER(pPed))
+                        pPed->CallEvent("onPlayerVehicleEnter", PlayerVehicleArguments);
+                    else
+                        pPed->CallEvent("onPedVehicleEnter", PlayerVehicleArguments);
```
qaisjp, line 4121: - [ ] Do you think onPedVehicleEnter should be called for both players and peds (and the old onPlayerVehicleEnter only for players).

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9802,22 +9805,25 @@ bool CStaticFunctionDefinitions::RemovePedFromVehicle(CClientPed* pPed)
...
         // Call onClientPlayerVehicleExit
         CLuaArguments Arguments;
         Arguments.PushElement(pVehicle);            // vehicle
         Arguments.PushNumber(uiSeat);               // seat
         Arguments.PushBoolean(false);               // jacker
-        pPed->CallEvent("onClientPlayerVehicleExit", Arguments, true);
+        if (IS_PLAYER(pPed))
+            pPed->CallEvent("onClientPlayerVehicleExit", Arguments, true);
+        else
+            pPed->CallEvent("onClientPedVehicleExit", Arguments, true);
```
qaisjp, line 9821: - [ ] Do you think onClientPedVehicleExit should be called for both players and peds (and the old onClientPlayerVehicleExit only for players).

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -6372,3 +6399,623 @@ void CClientPed::UpdateStreamPosition(const CVector& vecInPosition)
...
+    return true;
+}
+
+//////////////////////////////////////////////////////////////////
+//
+// CClientPed::ResetVehicleInOut
+//
+// Resets enter/exit variables
+//
+//////////////////////////////////////////////////////////////////
```
qaisjp:
```suggestion
// ResetVehicleInOut resets enter/exit variables, and is only called for the local player or for peds we're syncing.
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4154,28 +4157,28 @@ bool CStaticFunctionDefinitions::RemovePedFromVehicle(CElement* pElement)
...
-            }
+            CLuaArguments Arguments;
+            Arguments.PushElement(pVehicle);                 // vehicle
+            Arguments.PushNumber(ucOccupiedSeat);            // seat
+            Arguments.PushBoolean(false);                    // jacker
+            Arguments.PushBoolean(true);                     // forcedByScript
+            if (IS_PLAYER(pPed))
+                pPed->CallEvent("onPlayerVehicleExit", Arguments);
+            else
+                pPed->CallEvent("onPedVehicleExit", Arguments);
```
qaisjp, line 4168: - [ ] @Zangomangu same stuff about calling events for peds/players

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4154,28 +4157,28 @@ bool CStaticFunctionDefinitions::RemovePedFromVehicle(CElement* pElement)
         unsigned char ucOccupiedSeat = pPed->GetOccupiedVehicleSeat();
         if (pVehicle)
         {
-            CPlayer* pPlayer = static_cast<CPlayer*>(pElement);
-            if (pPlayer && IS_PLAYER(pElement))
```
qaisjp, line 4158: Note for reviewer: Just an indentation change here. Except for the thing I commented below (about the event stuff), the rest of file has just variable renames.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
qaisjp, line 9821: - [ ] @Zangomangu same stuff about calling events for peds/players

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9762,11 +9762,14 @@ bool CStaticFunctionDefinitions::WarpPedIntoVehicle(CClientPed* pPed, CClientVeh
     CLuaArguments Arguments;
     Arguments.PushElement(pVehicle);            // vehicle
     Arguments.PushNumber(uiSeat);               // seat
-    pPed->CallEvent("onClientPlayerVehicleEnter", Arguments, true);
+    if (IS_PLAYER(pPed))
+        pPed->CallEvent("onClientPlayerVehicleEnter", Arguments, true);
+    else
+        pPed->CallEvent("onClientPedVehicleEnter", Arguments, true);
```
qaisjp, line 9767: - [ ] @Zangomangu same stuff about calling events for peds/players

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -1929,79 +1944,89 @@ void CPacketHandler::Packet_Vehicle_InOut(NetBitStreamInterface& bitStream)
...
+                                    pOutsidePed->CallEvent("onClientPedVehicleExit", Arguments3, true);
 
                                 CLuaArguments Arguments4;
                                 Arguments4.PushElement(pVehicle);            // vehicle
                                 Arguments4.PushNumber(ucSeat);               // seat
-                                pInsidePlayer->CallEvent("onClientPlayerVehicleEnter", Arguments4, true);
+                                if (IS_PLAYER(pInsidePed))
+                                    pInsidePed->CallEvent("onClientPlayerVehicleEnter", Arguments4, true);
+                                else
+                                    pInsidePed->CallEvent("onClientPedVehicleEnter", Arguments4, true);
```
qaisjp, line 2029: - [ ]  @Zangomangu same stuff about calling events for peds/players

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -4087,18 +4110,52 @@ void CPacketHandler::Packet_EntityRemove(NetBitStreamInterface& bitStream)
...
+                    {
+                        for (auto iter = listOfPeds.begin(); iter != listOfPeds.end(); ++iter)
+                        {
+                            // Was this ped jacking us?
+                            CClientPed* pPed = *iter;
+                            if (pPed->m_bIsGettingJacked && pPed->m_pGettingJackedBy == pRemovedPed)
+                            {
+                                pPed->ResetVehicleInOut();
+                                pPed->RemoveFromVehicle(false);
+                                pPed->SetVehicleInOutState(VEHICLE_INOUT_NONE);
```
qaisjp, line 4168: Looks good 👍

Client/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -67,12 +78,17 @@ void CPedSync::DoPulse()
...
 }
 
 void CPedSync::RemovePed(CClientPed* pPed)
 {
     if (!m_List.empty())
         m_List.remove(pPed);
+
+    pPed->SetSyncing(false);
+    // Reset vehicle in/out stuff in case the ped was entering/exiting
+    pPed->ResetVehicleInOut();
```
qaisjp: - [x] @Zangomangu  Eh so I know SetSyncing(false) is only called in one  place (here) but I think the point of having function sinstead of directly accessing members is so that we can do these sort of side effects if we need to. Should `ResetVehicleInOut()` actually just be inside the SetSyncing(false) implementation?

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -6372,3 +6399,623 @@ void CClientPed::UpdateStreamPosition(const CVector& vecInPosition)
...
 }
+
+//////////////////////////////////////////////////////////////////
+//
+// CClientPed::EnterVehicle
+//
+// Asks server for permission to start entering vehicle
+//
+//////////////////////////////////////////////////////////////////
+bool CClientPed::EnterVehicle(CClientVehicle* pVehicle, bool bPassenger)
```
qaisjp, line 6414: - [ ] @Zangomangu  I think there should be a common prelude, like a private `ProcessInitialVehicleInOutKey` which booth calls and does these checks. Also note that for some reason these are actually currently inconsistent as well. 
image

qaisjp (changes requested):
Looks pretty great but there are some small things that need addressing. Plus if you have notes adding detailed code comments wherever possible would be a massive help. A lot of this knowledge is either forgotten or hidden, and getting your brain down as code comments will help future excavators.

Client/version.h
Zangomangu: Done

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Zangomangu: Done

Client/mods/deathmatch/logic/CClientPed.h
Zangomangu: Yes they are split up into two separate methods as this makes sense when we call it from scripting.

Client/mods/deathmatch/logic/CClientPed.h
Zangomangu: I don't think that's appropriate in this case, might cause confusion

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Zangomangu: I would say it shouldn't, you can bind both events to the same function to do that. And this way, you won't need to check if a ped specifically.

As I mentioned in the PR message we do need to think about if we want to trigger the player events for the peds, as this is current behavior (in warpPed / removePed functions). To keep Lua compatibility.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Zangomangu: See above.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Zangomangu: See above

Client/mods/deathmatch/logic/CPedSync.cpp
Zangomangu: Good idea, moved it into SetSyncing

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
Zangomangu: The "packet source element" is not used anywhere from what I can tell? unless net is doing something. It's a thing in CPacket that CVehicleInOutPacket is based on.

So because it is seemingly not used, I removed it so it doesnt cause confusion. It might have made sense when enter/exit only worked for players, but now the player who initiates the packet reply isn't necessarily the ped that the packet is about. Hope that makes sense.

Pirulax: Source player is used for the `source` variable in (among other things), for example, trigger event.
Theres another variable `client` (the socket from which the packet came from).

Client/mods/deathmatch/logic/CClientPed.cpp
Zangomangu: A common prelude would basically be: EnterOrExitVehicle(bool bEnterOrExit, bool bPassenger) to save a few of these initial checks.
It could be changed but is unnecessary imo.

The methods are split up and called EnterVehicle and ExitVehicle because they are now also called from scripting,

I added the check for dead ped. There is no reason to continue on if the ped is dead.

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
Zangomangu: Are we talking about the same thing here?
I mean it is not used in CVehicleInOutPacket (specifically) and later when the packet is broadcasted.

TheNormalnij:
Thanks you for this PR

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
qaisjp: All the other events fire ped events for both peds and players, I think going against that would be confusing. If we fire both for peds, we can just deprecate the player event by hiding it from the wiki.

Zangomangu: Event | Triggers for both players and   peds
-- | --
onClientPedDamage | No
onClientPedHeliKilled | No
onClientPedHitByWaterCannon | No
onClientPedWasted | No
onClientPedWeaponFire | No
onClientPedStep | Yes
onPedWasted | No
onPedWeaponSwitch | No

qaisjp: *groans*

alright, you win :D

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
qaisjp: I was suggesting that the packet source element could be the ped that the packet is about. If the packet source element cannot be a ped (and must be a player), then we can scrap this suggestion.

Is it possible for you to still provide code comments about each of the overloads?

Client/mods/deathmatch/logic/CClientPed.cpp
qaisjp: > I added the check for dead ped.

Do you mean the check for the dead vehicle?

qaisjp: Is there a particular reason checks are done in a different order for exits vs enters?

Zangomangu: The check for dead vehicle is not possible in EnterVehicle because we dont have a valid vehicle yet at that point.
As for the order of the checks, you want me to change that to match?

Server/mods/deathmatch/logic/packets/CVehicleInOutPacket.cpp
Zangomangu: Yeah it has to be a player. I think the confusion arrives from this:

**Receiving**
When we receive, the packet source is the client who sent it.
We used the packet source to find who the packet is about (this was the old way).
If we receive packet from old client, ped ID will be him.
If we receive packet from new client, ped ID will be part of packet. (because it can be a ped now)

**Sending**
When we send a packet, the packet source doesn't matter.
We write the player/ped ID to the packet.
(This is unchanged)

So the server->client packet hasn't changed.
Only the client->server packet.

As for CVehicleInOutPacket, I have added an explanation to each of the parameters in the header file.
I don't think it makes much sense to explain the overloads tbh?

We should rather add an explanation to each of the enums in CGame eg. VEHICLE_REQUEST_IN_CONFIRMED and similarly in CClientGame. These took some time to figure out because it's a bounce that goes 
client( CAN I ENTER? )->server( AIGHT )->client( AIGHT I AM INSIDE )->server( AIGHT YOU ARE INSIDE )->client

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2336,3 +2340,51 @@ int CLuaPedDefs::DetonateSatchels(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+int CLuaPedDefs::SetPedEnterVehicle(lua_State* luaVM)
+{
```
ghost: Please consider using the new argument parser.

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2336,3 +2340,51 @@ int CLuaPedDefs::DetonateSatchels(lua_State* luaVM)
...
+    else
+        return luaL_error(luaVM, argStream.GetFullErrorMessage());
+
+    // Failed
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
+
+int CLuaPedDefs::SetPedExitVehicle(lua_State* luaVM)
+{
```
ghost: \+

ghost (commented):
The code quality is good. What's really left is testing. It might be a bit early to say this, but I think the current nightly is stable enough for us to merge PRs like this one.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -6376,3 +6403,628 @@ void CClientPed::UpdateStreamPosition(const CVector& vecInPosition)
...
+    {
+        return false;
+    }
+
+    // Get our occupied vehicle
+    CClientVehicle* pOccupiedVehicle = GetOccupiedVehicle();
+    if (!pOccupiedVehicle)
+    {
+        return false;
+    }
```
Pirulax, line 6627: Minor detail, but this confuses me.
I think you should use a dassert here to indicate that this shoudn't happen at all, since why would `ExitVehicle` be called if you aren't in a veh in the first place?

Client/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -55,6 +55,17 @@ bool CPedSync::ProcessPacket(unsigned char ucPacketID, NetBitStreamInterface& Bi
...
 {
+    // Got any items?
+    if (m_List.size() > 0)
+    {
+        for (auto iter = m_List.begin(); iter != m_List.end(); ++iter)
+        {
+            CClientPed* pPed = *iter;
+            // Update enter/exit sequence
+            pPed->UpdateVehicleInOut();
+        }
```
Pirulax, line 66:
```cpp
// Update enter/exit sequence
for (CClientPed* pPed : m_List)
        pPed->UpdateVehicleInOut();
```

Client/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -55,6 +55,17 @@ bool CPedSync::ProcessPacket(unsigned char ucPacketID, NetBitStreamInterface& Bi
 
 void CPedSync::DoPulse()
 {
+    // Got any items?
+    if (m_List.size() > 0)
```
Pirulax, line 59: No need to check for this, it just slows the code down :D

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Pirulax:
```cpp
bool CLuaPedDefs::SetPedEnterVehicle(CClientPed* pPed, CClientVehicle* pVehicle, std::optional<bool> bPassenger)
{
// use bPassenger like: bPassenger.value_or(false)
// Also, I see you used a default value for pVehicle, I presume thats a typo
// if not, use std::optional for it here
}
// Also, when adding the function to the header this prototype:
// static bool SetPedEnterVehicle(CClientPed* pPed, CClientVehicle* pVehicle, std::optional<bool> bPassenger)
```

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Pirulax:
```cpp
bool CLuaPedDefs::SetPedExitVehicle(CClientPed* pPed)
{
    return pPed->ExitVehicle();
}
```

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Pirulax: Uh, also, use `ArgumentParser` in the function list.
See `CLuaCryptDefs` for examples.

Pirulax (commented):
I suggest you to run the formatter on `CGame`.
Also, if possible, use more `continue`'s in for loops, because the function in `CGame` is already like 700 lines, with 10x indentation, and its horrible to read, so please, early exists, instead of nesting `if`s

Client/mods/deathmatch/logic/CClientPed.cpp
qaisjp: This function is called from CLuaPedDefs

Pirulax: Yeah, my bad.

qaisjp:
Since most of that code moved from CClientGame to CClientPed is mission critical, I think we don't need to refactor/clean it up, and can clean it up in a separate PR. I'd like to get this merged pretty soon so it can be extensively tested.

Is master stable right now?

Pirulax:
Yeah I meant the code he wrote, as I saw some. (At least git marked it with green, so I presume that wasn't part of the copy paste).

qaisjp:
Unfortunately this PR wasn't split up into atomic commits, so it requires quite a bit of careful thinking to determine which part just has code moved, and which is new code.

You can expand a bunch of the comments as I've done that work and left comments where it's just been copied.

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Zangomangu: Thx for the pointers, I've added it.

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Zangomangu: Added

qaisjp (approved):

ghost (approved):

PR #1753, opened 2020-10-24
Allow to use peds and vehicles as a camera target (with fixes)

TheNormalnij:
This PR fixes desync problem in spectator mode for players.
`setCameraTarget( pVehicle )` still have same issue. You need use a vehicle controller in `setCameraTarget`  to avoid lags.
Related to #1398


qaisjp:
@saml1er and I have found some bugs.

Reproduce using this resource: [setCameraTarget.zip](https://github.com/multitheftauto/mtasa-blue/files/5434029/setCameraTarget.zip)

Bug reproduction steps:

1. Spawn a vehicle (/cv nrg-500)
2. Click "Create and focus client ped"
3. Move camera around, works fine!
4. Click enter to get in a vehicle.
5. Camera freezes (!)`, is no longer focused to the ped, and `getCameraTarget` claims to be focused on the vehicle (but it's focused on nothing, it's frozen)
6. Get out of the vehicle
7. Camera is still frozen from earlier, but `getCameraTarget` still thinks you're focused on the vehicle


Another bug, using same resource

1. Get in a vehicle
2. Get out of the vehicle
3. Click "Focus last vehicle"
4. Get in a second vehicle
5. Player is now focused to the second vehicle even though they should be focused to the first vehicle
6. Get out of the second vehicle
7. Player remains focused to the second vehicle. They should not be focused to the second vehicle.

Another bug, using the same resource

1. Get in a vehicle
2. Get out
3. Click "Focus last vehicle"
3. Set your position very far away (run `me.position = Vector3(0, 0, 0)`)
4. Keep clicking "Focus last vehicle": the weapon indicator/fist keeps flickering, and there are graphical glitches appear resembling things on screen streaming out / showing lods temporarily. Players near the vehicle will fall for a few frames, so perhaps collisions are going away temporarily too. I think this is caused by your addition of `SetFocusToLocalPlayer`, because this doesn't happen to players.

Another bug

1. Focus another player
2. Have that player get in a car
3. Server says you're focused on the player, client thinks you're focused on the car
4. Blip faces North image


Some minor bugs related to the HUD radar, which you can look at in the future:

1. Same as above, blip faces North when you deliberately focus a vehicle. It doesn't face north when you are focusing yourself (as passenger or driver).
2. If your player is far away, and you are focused on a vehicle, the radar will zoom out slowly because your player is falling. Have not checked if this behaviour is new, I haven't checked if this used to happen or checked if it currently happens when focused on another player.

qaisjp:
> oh. Seems, MTA use vehicle extrapolation only for cameraTarget player

You said the above on Discord today.

From my tests, I can confirm that if the old server (i.e. a server without this pull request) is not aware of you focusing a vehicle, the vehicle position goes haywire. A server with this PR lets you focus a vehicle even if its syncer isn't a driver (and is just a passenger), and the vehicle syncs correctly.

TheNormalnij:
I hope my fixes won't produce more bugs

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4790,9 +4790,14 @@ bool CStaticFunctionDefinitions::GetCameraMatrix(CVector& vecPosition, CVector&
...
 {
-    if (!m_pCamera->IsInFixedMode())
-        return m_pCamera->GetTargetEntity();
-    return NULL;
+    if (m_pCamera->IsInFixedMode())
+        return NULL;
+
+    CClientEntity* pPlayer = m_pCamera->GetFocusedPlayer();
+    if (pPlayer)
+        return pPlayer;
```
Pirulax, line 4814:
```cpp
    if (CClientEntity* pPlayer = m_pCamera->GetFocusedPlayer())
        return pPlayer;
```

ghost: It's a good practice to limit the scope for variables, but this is optional. It's like one of those changes where you might want to use braces for a single line after if condition (I don't use braces in that case). In such cases, I prefer to let the contributor decide how they want to approach it.

TheNormalnij: I think first way is easy to read

Lpsd:
What's the status on this? Has anyone tested it since the latest changes? 

Could probably do with having this in nightly for a short while before 1.5.9 to ensure there aren't any additional bugs - as there seemed to be quite a few above under specific circumstances (who knows what other conditional side-effects there may be that simply haven't been tested for yet).

patrikjuvonen:
This should be tested before merging it into master. Let's try to organise a testing session soon.

patrikjuvonen:
We hosted a group testing session yesterday and we found no major issues.

We found that using @qaisjp's test resource, clicking on the create buttons for the first time creates the ped, but does not focus on the ped. When you press again, it focuses. If you restart the resource, the bug happens again. I am not sure why this would happen.

TheNormalnij:
I'll check this bug this weekend

TheNormalnij:
@patrikjuvonen Instantly loading was disabled for peds. I didn't found any reasons for this.  Now it works for me.  But this small change should be tested with small group of players. I can't do that myself.

patrikjuvonen:
> @patrikjuvonen Instantly loading was disabled for peds. I didn't found any reasons for this. Now it works for me. But this small change should be tested with small group of players. I can't do that myself.

The reason appears to be here: https://github.com/multitheftauto/mtasa-blue/commit/2881a99991a58b6c06ec45811675549e557d4d2e

I think let's not change this for now. Now we understand why it happens and problem doesn't come from this PR.

Thanks for checking the issue!

I'll revert the change and merge this now :)

patrikjuvonen (approved):

StrixG (approved):

Issue #1718, opened 2020-10-15
Endless Game Freeze during Gameplay


haron4igg:
Got pretty much reports from players about game getting frozen during gameplay. No crash after, PC is unresponsible, but player stays connected. Only hard-reset helps in this case.

MTA Diags, from affected PCs:
https://pastebin.mtasa.com/770694034
https://pastebin.mtasa.com/802091107
https://pastebin.mtasa.com/684407507
https://pastebin.mtasa.com/475454060

(waiting for crash dumps from these people)

Dutchman101:
I just want to note that during the past year, you have presented us a lot of weird, vague and unreproducable issues (both on here and discord) which turned out to be caused by bad scripting practises/mistakes on your end, or were left not cleared up with you not recognizing how you may be predisposing players to said issues. I am sorry for not stating the latter in relevant places and including explanation, but i didn't feel like it.

What im getting to: it's totally possible for a server to cause client freezes. So what MTA devs can do about this issue might be limited. 

But anyways, without further speculation, let's first have someone look on the thread of the dumps to see what it was processing at the time of freeze. Even in the case of abuse by scripts it can give a hint.

haron4igg:
i agree that in most of cases its my fault. I'm just using MTA API, trying to keep-up with players needs.... and when some minority report issues, i cant just stand away and ignore them. I didn't knew that CEF may be such unstable thing, in such a stable thing like MTA so invested a lot of time adopting it. And now its lead to all those Issues i reported before... i felt very exhausted having so much sleepless nights for last 4 months fighting it, so we got what we got now :( 

but anyway i'm very thankful for all your help and very sorry for bothering.
I definitely should learn how to investigate crashdumps on my own.

haron4igg:
Player who had issue with freeze on loading, who reported this crashdump above, just reported that his HDD had damaged cells. After doing cell-remap and reinstalling windows - his issue is gone.

Will look forward for crashdumps from players who experience issue during gameplay

qaisjp:
Nice find!

ghost:
Btw, I did analyze one of the crash dumps you provided. I'm not sure if I got it form this issue, but it was a recent one from `20668` build. I checked the main thread, it was stuck on a random function in ntdll.dll. I couldn't get anything useful out of it. Once you provide us with a few more dumps, I might be able tell where it crashed.

haron4igg:
[...]

PR #1761, opened 2020-10-26
Fix game freeze & updated BASS libraries

Dutchman101:
BASS.dll fixes:
- Fixed bug in BASS.dll that can cause game freezes when a resource seeks corrupted .ogg files (as in issue #1718)
- Avert low memory crash on offset 0000184e (reported by MTA)
- Fix failed alloc (in Ogg Vorbis decoder) crashes on 0000d013 and 0000d79c (reported by MTA)

bass_fx.dll fixes:
- Fixed a bunch of failed reallocs that were remaining after the last update (to stop memory corruption)
Further hardening against out of mem, it should now result in audio glitches (due to no buffer space) rather than crashes

bass_aac.dll fixes:
- Avert low memory crash on offset 000020dc (reported by MTA - the only change)

NOTE: after this PR gets merged, i advise no one gets the latest nightly until at least the Microsoft Defender false positive on BASS.dll has been resolved (already reported it). They usually take less than a day.


qaisjp:
Do you have download links for the dlls?

Dutchman101:
> Do you have download links for the dlls?

yeah, i forgot.. here:

[http://www.un4seen.com/stuff/bass.zip](http://www.un4seen.com/stuff/bass.zip)
[http://www.un4seen.com/stuff/bass_aac.zip](http://www.un4seen.com/stuff/bass_aac.zip)
[http://www.un4seen.com/stuff/bass_fx.zip](http://www.un4seen.com/stuff/bass_fx.zip)

qaisjp (approved):
LGTM / verified, although if there are Windows Defender problems I would recommend waiting for the dlls to be whitelisted before merging.

Dutchman101:
> LGTM / verified, although if there are Windows Defender problems I would recommend waiting for the dlls to be whitelisted before merging.

I agree

Dutchman101:
It's solved, see below

fp

PR #1768, opened 2020-10-27
Remove unused Lua manager pointer from some classes

### LINKED ISSUE #1760: m_pLuaManager used before its defined
**Describe the bug**
`m_pLuaManager` is used before it's defined @ [CGame.cpp:514](https://github.com/multitheftauto/mtasa-blue/blob/d2a906da5b7a17d280613fd55d9a1362e40601c1/Server/mods/deathmatch/logic/CGame.cpp#L514)

**Expected behaviour**
`m_pLuaManager` should be assigned a valid value before usage

**Additional context**
The pointer isn't used anywhere in the classes its passed into, but if it does it'll cause some nice headache.
Just leaving this issue here, so I dont forget about fixing it (Im lazy to commit my changes on another branch)

Pirulax:
Fixes #1760


Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -33,10 +33,9 @@ struct
     {26, "data/timecyc.dat"},
 };
 
-CMainConfig::CMainConfig(CConsole* pConsole, CLuaManager* pLuaMain) : CXMLConfig(NULL)
+CMainConfig::CMainConfig(CConsole* pConsol) : CXMLConfig(NULL)
```
AlexTMjugador: It seems like you accidentally renamed a parameter here.

Pirulax: My bad, thanks!

qaisjp (approved):
Good stuff

Issue #1765, opened 2020-10-27
Refactor our redundant Lua API functions to use templates


Pirulax:
**Is your feature request related to a problem? Please describe.**
There are 2 places where we have copy pasted code that can be easily "templateized":
- `CLuaArgument` - `PushACL, `PushElement`, etc..
- `LuaCommon` - Ex.: `lua_pushacl`, `lua_pushelement`, etc.. 

**Describe the solution you'd like**
We can use some generic programming magic (aka templates)

**Describe alternatives you've considered**
Leave it like it is

**Additional context**
We can use Visual Studio's `Edit->Refactor` option. It's really convenient, this issue can be solved in an hour (or ideally less)
I don't think following Lua's convention is necessary, as it just adds redundant code. Lua has different function names, because C doesn't have templates, but this ain't C, but C++, so... We might as well make use of it.

PR #1772, opened 2020-10-29
Refactor LuaCommon to use template magic instead of lua_push`type`

Pirulax:
Not sure if this is the best solution tho.
But, using a templated function + overloads doesn't work either, because `CColPoint*` will call the templated version, instead of the `CElement*` (and CElement needs special handling). 

This PR:
- Renames all custom `lua_push<something>` functions to `lua_pushobject` (eg.: `lua_pushelement` -> `lua_pushobject`), this is so we definitely dont follow Lua's convention of naming (after all Lua is written in C, but here in c++ we have templates)
- Moves LuaCommon to the shared project (only needed a little bit of macro magic)

Partially solves #1765 (other part is `CLuaArgument`)


Pirulax:
Needs some rebasing later on.

qaisjp:
is there an issue about this? moving things to shared (which is its own refactor) is fine / something I think we could move towards, but I'm worried that this change is not actually something we want?

Pirulax:
Well, only about refactoring this and CLuaArgument together.

qaisjp:
btw great job with the atomic commits 👍  definitely easier to this review this!

qaisjp:
btw, I'm asking if there's an issue for this because I don't know if converting everything to use templates is actually a good idea. i may be wrong, but i heard it can slow build times? 

I'm hoping a more educated reviewer such as @sbx320 could take a look at whether this change is reasonable or not

sbx320:
> i may be wrong, but i heard it can slow build times?

To some extent yes. However the impact isn't as big as it once was. Certain compilers used to implement templates by just copy & pasting the token stream of a template for each type variant, which led to gigantic ASTs which slowed down complicing due to code size. Those times are over though and MTA already compiles fairly slowly due to a very bad usage of precompiled headers, so I wouldn't expect any relevant differences in compile time. If we want to speed up compile times, we'd need to clean out precompiled headers and avoid including *everything* all the time. 

Not 100% sure if I like this implementation in LuaCommon though. 

> But, using a templated function + overloads doesn't work either, because CColPoint* will call the templated version, instead of the CElement* (and CElement needs special handling).

You can disable templated overloads via `std::enable_if_t`. Also CI failed :(

Pirulax:
Yes I know about `std::enable_if_t`.. Actually, now that I think about it, it should work.
Okay, I'll do that, because the ugly `constexpr` code is my only concern.

qaisjp:
"Change from void* to SArrayId" seems like a distinct workstream. is this something you could do later? if so, it's probably best to try and reduce the scope of changes so that if something goes wrong, we know exactly which change caused it. otherwise we will just have to revert the entire PR and leave it to you to figure out what exactly went wrong

Pirulax:
I think we could just use @sbx320's already existing `lua::Push`. This way we dont duplicate code.
Also, with `lua::Push` we would have a unified way of pushing values onto the stack.

This PR it turning into a shitshow pretty fast as I try to make it work, becuase it keeps throwing me random ass errors every time I try to remove something that doesnt belong to `LuaCommon`, and its really annoying.

Pirulax:
Closing in favor of #2145

PR #1777, opened 2020-10-31
Remove unused CPlayerStats 

Pirulax:


Server/mods/deathmatch/logic/CPlayerStats.h
```diff
@@ -29,10 +28,10 @@ class CPlayerStats
 
     bool                           GetStat(unsigned short usID, float& fValue);
     void                           SetStat(unsigned short usID, float fValue);
-    vector<sStat*>::const_iterator IterBegin() { return m_List.begin(); }
-    vector<sStat*>::const_iterator IterEnd() { return m_List.end(); }
+    std::vector<sStat*>::const_iterator IterBegin() { return m_List.begin(); }
+    std::vector<sStat*>::const_iterator IterEnd() { return m_List.end(); }
     unsigned short                 GetSize() { return static_cast<unsigned short>(m_List.size()); }
```
qaisjp: this is misformatted now. could you move IterBegin and IterEnd underneath GetSize (leave a blank line in between) and then reformat the struct?

qaisjp (dismissed):

Server/mods/deathmatch/logic/CPlayerStats.h
Pirulax: I left it like that, because I thought once you said that I should leave these things up to the formatter. But I might remember it incorrectly.

Pirulax:
wtf how is it broken
oh cmon man xD

Server/mods/deathmatch/logic/CPlayerStats.h
qaisjp: try formatting selection, you'll see it's not in line with the formatter

Pirulax:
Turns out this class was unused.. I wonder if there's more?

Issue #1340, opened 2020-04-02
SD 22 and SD 23 occurences have increased drastically lately


zo3n:
**Describe the bug**
I am not really sure how to describe this other than the title. While administering another server I can notice that SD 22 and SD 23 are shown in like 15 players out of 150. This is not normal.
SD 22 seems to happen the most. I have recieved the error now myself:
```
HTTP server file mismatch (admin) admin_spectator.lua [Got size:0 MD5:00000000000000000000000000000000, wanted MD5:8DB609CB32CA48997CE2CF5BAF9DDB7D]
Download error: HTTP server file mismatch (admin) admin_spectator.lua [Got size:0 MD5:00000000000000000000000000000000, wanted MD5:8DB609CB32CA48997CE2CF5BAF9DDB7D]
```
Last time I encountered HTTP errors like this was in **2015**
Some of my friends have also encountered this MD5 0000 thing within last month.

I won't say with absolute certainty that this is not normal, but I have a good feeling that the frequency of it has increased within last week or month. That is the only reason I'm reporting it; so that it can get some attention, or even better, be solved or explained.

**To reproduce**
I don't know

**Expected behaviour**
To not happen

**Version**
I don't know

**Additional Information**
Restarting the resource affected doesn't solve the issue, and still spits out errors:
```
Unable to delete old file admin\client\gui\admin_spectator.lua
HTTP server file mismatch (admin) admin_spectator.lua [Got size:0 MD5:00000000000000000000000000000000, wanted MD5:8DB609CB32CA48997CE2CF5BAF9DDB7D]
Download error: HTTP server file mismatch (admin) admin_spectator.lua [Got size:0 MD5:00000000000000000000000000000000, wanted MD5:8DB609CB32CA48997CE2CF5BAF9DDB7D]
```

This results in admin panel being bugged without having any other fix other than restarting the server, which can become a problem if it's a production server.

zo3n:
![image](https://user-images.githubusercontent.com/923242/78353574-2d193580-75a2-11ea-8704-5de081361567.png)

I have noticed that when i try to delete affected resource from cache, it doesn't allow me to delete it "because it's in use", **even if i am not connected to any server**

Additionally, when I tried to copy the affected resource (ipb in my newer case) it says that it cannot do that because it is open in **Multi Theft Auto WOW64 Helper**

ccw808:
What happens if you try deleting the individual files in the folder. Do you get the same error for every file?

[...]

PR #1778, opened 2020-10-31
Improve checksum error messages for internal HTTP servers

qaisjp:
**Reviewable commit by commit**

Context is available at https://github.com/multitheftauto/mtasa-blue/issues/1340#issuecomment-719888637

This only works for internal HTTP servers right now, with the possibility to extend it for other servers later, in a future PR.

Try turning off whitespace diffs when reviewing the first commit.

**Before**

image

**After**

image

### DIFF
--- a/Client/mods/deathmatch/logic/CDownloadableResource.cpp
+++ b/Client/mods/deathmatch/logic/CDownloadableResource.cpp
@@ -43,7 +43,7 @@ bool CDownloadableResource::DoesClientAndServerChecksumMatch()
 
 CChecksum CDownloadableResource::GenerateClientChecksum()
 {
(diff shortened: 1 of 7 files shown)


AlexTMjugador:
Referring to https://github.com/multitheftauto/mtasa-blue/issues/1340#issuecomment-719888637:
> This option is only available for Windows -- I don't think Linux fully supports file locking.

Linux and other POSIX-compliant systems do in fact support advisory file locking, via the POSIX [fcntl](https://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html) function. Mandatory locks (Windows-like, enforced for even non-cooperative processes) are usually available too, although with more platform-specific calls.

ghost (approved):

PR #1779, opened 2020-10-31
Allow allocating new timed-object models using engineRequestModel

TheNormalnij:
**API**
```lua
engineRequestModel( "timed-object" )
```

You can use `engineSetModelVisibleTime` and `engineGetModelVisibleTime` functions with a new ID


TheNormalnij:
Ready for review

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1407,6 +1407,24 @@ void CModelInfoSA::MakeObjectModel(ushort usBaseID)
     CopyStreamingInfoFromModel(usBaseID);
 }
 
+void CModelInfoSA::MakeTimedObjectModel(ushort usBaseID)
+{
+    CTimeModelInfoSAInterface* m_pInterface = new CTimeModelInfoSAInterface();
+
```
ghost, line 1494: Does the memory get deallocated?

qaisjp:
Should eClientModelType be a strict subset of element types? I'm thinking maybe it would be good to have a options table { timed = true } instead of adding a new "timed-object"?

Pirulax:
Let's do it this way, then refactor it in another PR?

qaisjp:
No thank you

TheNormalnij:
> Should eClientModelType be a strict subset of element types? I'm thinking maybe it would be good to have a options table { timed = true } instead of adding a new "timed-object"?

I think we should use internal gta sa classes. `engineGetModelType` can be added and this function can use internal gta sa classes only

Client/game_sa/CModelInfoSA.cpp
ghost: Apparently, all current `CModelInfoSA::Make*` functions cause a memory leak. The fix is simple, we just need to call delete in `CModelInfoSA::DeallocateModel`.
```
void CModelInfoSA::DeallocateModel(void)
{
    Remove();
    delete ppModelInfo[m_dwModelID]; // <---------- This line fixes the memory leak problem.
    ppModelInfo[m_dwModelID] = nullptr;
    pGame->GetStreaming()->GetStreamingInfoFromModelId(m_dwModelID)->Reset();
}
```

TheNormalnij: But we need select right class type here. I don't what function `Remove` doing.

TheNormalnij: I will add `switch case` here

ghost: > But we need select right class type here. I don't what function Remove doing.

CModelInfoSA::Remove does not delete the model info. It only deletes the copy of allocated clump. Internally, it calls CStreaming::RemoveModel. You can see the reversed code [here](https://github.com/saml1er/gta-reversed/blob/master/source/game_sa/CStreaming.cpp#L2275). So in short, the fix is simple. Since we are allocating a custom interface, then we must delete it in CModelInfoSA::DeallocateModel.

ghost: > I will add `switch case` here

It's really not needed. It's just a pointer.

TheNormalnij: but it's pointer to `CBaseModelInfo`. Will be `delete` works with children classes?

ghost: > 
> 
> but it's pointer to `CBaseModelInfo`. Will be `delete` works with children classes?

CBaseModelInfoSAInterface is the base class, and it has a virtual destructor, so yes, it will work.

EDIT: Hmm, you might be right. In GTA SA, `CBaseModelInfo` class has a virtual destructor, but in MTA, it's called `CBaseModelInfoSAInterface`, and It doesn't have a "real" virtual destructor, so it won't work. I guess you need to use a switch statement here.

TheNormalnij: Done

botder (approved):

PR #1570, opened 2020-07-20
Bring back Discord (updated discord-rpc)

0x416c69:
I think join system is no longer working (The game won't launch), if you guys can test it I would appreciate it.

I think I will have to RE discord game sdk to find the new method of registering game for join feature. I will implement it under https://github.com/0x416c69/discord-rpc and will update the project here.

Everything else seems to have worked fine. That discord rpc repository is tested by my users (about ~2000-3000 users) and there were no unhandled exceptions from discordrpc.dll.

PR #1784, opened 2020-11-01
Remove discord-sdk leftovers

Deihim007:


Deihim007:
Apparently these leftovers are a part of #1570 which I wasn't aware it exists.

PR #1789, opened 2020-11-03
Refactor CPlayerManager

Pirulax:
It was a very outdated code. 
Also, I can't belive we've used it like this, its horrible.. Heap allocating and copying vectors, and maps for every packet?
This PR fixes it all! At once. (Coudnt really break it into more PRs).

**Features:**
- Added static `BroadcastIf` - Send the packet only to a player for which `predicate` is `true`
- `BroadcastJoinedIf` - A member function that uses the internally stored and optimized `m_JoinedByBitStreamVer` for `BroadcastIf`
- The above mentioned function can be used like as a base for other existing functions, for example: `BroadcastJoinedDimension` (I've refactored all of them to use `BroadcastJoinedIf`)
- The former 2 give a pretty good performance boost. Not tested, but it would make sense, especially on servers where there are a lot of packets sent / frame. (No heap allocs anymore, and sorts, so :D)
- Refactored the code to c++17 standards - Now its even more readable
- Added 2 todos!! Maybe even more!
- Added noexcept and const wherever possible
- Cleaned up old code related to the manager (eg.: Use `GetAllPlayers()` instead of `IterBegin` and `IterEnd`, and use `IterateJoined` where the old code used `IterBegin` and `IterEnd` and checked if the player is joined)
- Moved `CSendList` from `CGame` to its own file, because in the process of refactoring it ended up being quite big
- Added `GetNetServerReliability` and `GetNetServerPriority` to `CPacket`. These functions are wrappers and translator of `GetFlags` (They translate from our enum to raknet enums)

**When reviewing**
- See if the old codes logic matches the new one. Eg.: In commit bdbb722 see if I've missed a check we copy pasting code, or something. (I didnt)

@qaisjp Sorry for some fuckups in the commits, but coudn't properly `select staged` in vsc.
I've tried my best to make the commits atomic.


Server/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -158,33 +158,27 @@ CPlayer* CPedSync::FindPlayerCloseToPed(CPed* pPed, float fMaxDistance)
...
-                    }
-                }
-            }
-        }
-    }
+    CPlayer* pClosestAppliableSyncer = nullptr;
+    m_pPlayerManager->IterateJoined([=, &pClosestAppliableSyncer](CPlayer* pPlayer) {
+        // Make sure he isnt being deleted
+        if (pPlayer->IsBeingDeleted())
+            return;
```
Pirulax, line 165: I added this check, because I presume it was unintentionally left out?

Pirulax:
Me when i press `Build`: https://imgur.com/gallery/b4GHdOO

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1291,21 +1291,19 @@ void CGame::InitialDataStream(CPlayer& Player)
...
-        {
-            EntityPacket.Add(pPlayer);
-        }
+        if (pPlayer == &Player) // Dont add us to the list
+            continue;
+
+        if (pPlayer->IsJoined() && !pPlayer->IsBeingDeleted())
+            PlayerList.AddPlayer(pPlayer);
+
+        EntityPacket.Add(pPlayer);
```
Pirulax, line 1309: Seems like we add the entity here, even if its being deleted?

.gitignore
```diff
@@ -416,3 +416,4 @@ utils/DXFiles/
 !*.dll
 !*.exe
 utils/vswhere.exe
+refactor/
```
Pirulax: Whoever merges this PR, dont forget to remove this.

qaisjp: https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally

Pirulax: thanks

Pirulax:
Tested with 3 players, everything seems to work just fine.

turret001:
> Tested with 3 players, everything seems to work just fine.

yes, I tested with him, it works

stoneage-mta:
i also helped with testing, its working fine

Pirulax:
Yeah I feel like this wont be merged soon.
Any way I could speed up the process?

Pirulax:
A little issue: The bitstream is created even if the packet wont be sent to anyone. While this isnt a problem on a server where the bitstream version spread is small (eg.: there are only 1-2 different bitstream versions), bit if there are a lot this might slower than the old method. (Because we allocate bitstreams that aren't used for anything).
I'm not sure what `a bitstream allocation consists of, maybe it's like an empty `std::vector` which is basically free.

Pirulax:
Bump.

Server/mods/deathmatch/logic/CResource.cpp
```diff
@@ -692,14 +692,20 @@ bool CResource::GetCompatibilityStatus(SString& strOutStatus)
...
+        const auto& allPlayers = g_pGame->GetPlayerManager()->GetAllPlayers();
+        const auto numIncompatiblePlayers = std::count_if(allPlayers.begin(), allPlayers.end(), [&](CPlayer* pPlayer) {
+            if (!pPlayer->IsJoined())
+                return false;
+
+            if (pPlayer->ShouldIgnoreMinClientVersionChecks())
+                return false;
 
-        if (uiNumIncompatiblePlayers > 0)
+            return m_strMinClientRequirement >= pPlayer->GetPlayerVersion();
```
botder:
```suggestion
            return m_strMinClientRequirement > pPlayer->GetPlayerVersion();
```
Original code:
```cpp
m_strMinClientRequirement > (*iter)->GetPlayerVersion()
```

Server/mods/deathmatch/logic/CResource.cpp
```diff
@@ -1112,15 +1118,9 @@ bool CResource::CreateVM(bool bEnableOOP)
...
-
-    for (; iter != g_pGame->GetPlayerManager()->IterEnd(); iter++)
-    {
-        CKeyBinds* pBinds = (*iter)->GetKeyBinds();
-
-        if (pBinds)
+    for (CPlayer* pPlayer : g_pGame->GetPlayerManager()->GetAllPlayers())
+        if (auto pBinds = pPlayer->GetKeyBinds())
             pBinds->RemoveAllKeys(m_pVM);
-    }
```
botder: Please don't remove the brackets for the if-loop and don't use **auto** if the variable type isn't clear.

Server/mods/deathmatch/logic/CSendList.h
```diff
@@ -0,0 +1,94 @@
...
+#include <vector>
+#include <unordered_map>
+#include <type_traits>
+
+class CPlayer;
+
+class CSendList : protected std::unordered_map<unsigned short, std::vector<CPlayer*>>
+{
+public:
+    using BaseCont_t = std::unordered_map<unsigned short, std::vector<CPlayer*>>;
```
botder: Don't use the `_t` suffix for types. It's reserved for future standard changes (even if unlikely in this case).

Server/mods/deathmatch/logic/CSendList.h
```diff
@@ -0,0 +1,94 @@
...
+        if (const auto it = find(bsver); it != end())
+            return &it->second;
+        return nullptr;
+    }
+
+    BaseCont_t::mapped_type* GetPlayerGroupByBitStream(unsigned short bsver)
+    { 
+        // As seen in Scott Mayers book.. Thanks stackoverflow!
+        return const_cast<BaseCont_t::mapped_type*>(const_cast<const CSendList*>(this)->GetPlayerGroupByBitStream(bsver));
+    }
```
botder: You could've copied the 3 lines from the const method and you would avoid the const cast magic, no?

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2782,30 +2782,7 @@ bool CStaticFunctionDefinitions::GetPlayerSourceIP(CPlayer* pPlayer, SString& st
...
-                if (i++ == uiRandom)
-                {
-                    return *iter;
-                }
-            }
-        }
-    }
-
-    return NULL;
+    return m_pPlayerManager->GetRandom();
```
botder, line 2785: Does `CPlayerManager::GetRandom` return players that are `IsBeingDeleted()`?

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
     {
         // Remove from other players near/far lists
-        (*iter)->RemovePlayerFromDistLists(pPlayer);
+        itPlayer->RemovePlayerFromDistLists(pPlayer);
 
-        // Update lowest player version
-        if ((*iter)->IsJoined() && ((*iter)->GetPlayerVersion() < m_strLowestConnectedPlayerVersion || m_strLowestConnectedPlayerVersion.empty()))
-            m_strLowestConnectedPlayerVersion = (*iter)->GetPlayerVersion();
+        // Find lowest player version
+        if (itPlayer->IsJoined())
```
botder: This is confusing, why do you check if the player is joined here?
From **CPlayerManager.h**:
```cpp
CFastHashSet<CPlayer*>                    m_Players; // All joined players
```

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
 {
-    m_Players.remove(pPlayer);
-    MapRemove(m_SocketPlayerMap, pPlayer->GetSocket());
-    assert(!m_Players.Contains(pPlayer));
-    assert(m_Players.size() == m_SocketPlayerMap.size());
+    dassert(m_Players.find(pPlayer) != m_Players.end()); // Make sure he's not in the set already
+    m_Players.erase(pPlayer);
+
+    dassert(MapFind(m_SocketPlayerMap, pPlayer->GetSocket())); // Make sure he was in the socket map as well
+    m_SocketPlayerMap.erase(pPlayer->GetSocket());
```
botder, line 223: Any reason why you replaced the original line?
```cpp
MapRemove(m_SocketPlayerMap, pPlayer->GetSocket());
```

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
-    assert(m_Players.size() == m_SocketPlayerMap.size());
 }
 
 void CPlayerManager::RemoveFromList(CPlayer* pPlayer)
 {
-    m_Players.remove(pPlayer);
-    MapRemove(m_SocketPlayerMap, pPlayer->GetSocket());
-    assert(!m_Players.Contains(pPlayer));
-    assert(m_Players.size() == m_SocketPlayerMap.size());
+    dassert(m_Players.find(pPlayer) != m_Players.end()); // Make sure he's not in the set already
```
botder: Misleading comment

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
+    g_pGame->CalculateMinClientRequirement();
+
+    m_JoinedByBitStreamVer.Insert(pPlayer);
 }
 
 void CPlayerManager::AddToList(CPlayer* pPlayer)
 {
-    for (std::list<CPlayer*>::const_iterator iter = m_Players.begin(); iter != m_Players.end(); iter++)
+    dassert(m_Players.find(pPlayer) == m_Players.end()); // Make sure he's not in the set already
+    m_Players.insert(pPlayer); // Before putting this into dassert, look at its definiton :-)
```
botder: The comment is only confusing for the reader/reviewer.

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
-    // Use the flags to determine how to send it
-    NetServerPacketReliability Reliability;
-    unsigned long              ulFlags = Packet.GetFlags();
-    if (ulFlags & PACKET_RELIABLE)
-    {
-        if (ulFlags & PACKET_SEQUENCED)
-        {
-            Reliability = PACKET_RELIABILITY_RELIABLE_ORDERED;
-        }
+    const auto iter = std::find_if(m_Players.begin(), m_Players.end(), [bCaseSensitive, szNick](CPlayer* pPlayer) {
```
botder: If `m_Players` only contains joined players, doesn't this break `getPlayerFromName` (in onPlayerConnect)?

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -74,327 +67,139 @@ void CPlayerManager::PulseZombieCheck()
...
-            if ((bCaseSensitive && strcmp(szNick, szTemp) == 0) || (!bCaseSensitive && stricmp(szNick, szTemp) == 0))
-            {
-                return *iter;
-            }
-        }
-    }
-
-    // If not, return NULL
-    return NULL;
+    return new CPlayer(this, m_pScriptDebugging, PlayerSocket); // Seems very safe
```
botder:
```
// Seems very safe
```
Please don't add such comments, what is this supposed to mean?

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -10,43 +10,36 @@
...
-    m_pScriptDebugging = NULL;
-    m_ZombieCheckTimer.SetUseModuleTickCount(true);
-}
-
 CPlayerManager::~CPlayerManager()
 {
-    DeleteAll();
+    // Deallocate all players
+    for (CPlayer* pPlayer : m_Players)
+        delete pPlayer;
```
botder: This is dangerous if the player destructor manages to change the underlying container **m_Players** while you're iterating over it. The previous code doesn't have this issue:
```cpp
   while (!m_Players.empty())
        delete *m_Players.begin();
```

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -85,14 +85,14 @@ class CPlayer : public CPed, public CClient
     int                GetClientType() { return CClient::CLIENT_PLAYER; };
     unsigned long long GetTimeSinceConnected() { return m_ConnectedTimer.Get(); }
 
-    const char* GetNick() { return m_strNick; };
-    void        SetNick(const char* szNick);
+    const char* GetNick() { return m_strNick; }; // Todo: Return a const SString& here instead. Also, make this function a const
+    void        SetNick(const char* szNick); // Todo: change this to SString, then move SString into m_strNick. Move ctor needs to be added to SString..
```
botder, line 89: Please remove the TODO comments. IMHO it's very obvious this should be changed in a **CPlayer** refactor.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4385,19 +4347,17 @@ CMtaVersion CGame::CalculateMinClientRequirement()
...
+                    return false;
+
+                CStaticFunctionDefinitions::RedirectPlayer(pPlayer, "", 0, NULL);
+                return true;
+            });
 
-            if (uiNumIncompatiblePlayers > 0)
-                CLogger::LogPrintf(SString("Forced %d player(s) to reconnect so they can update to %s\n", uiNumIncompatiblePlayers, *strKickMin));
+            if (playersRedirectedCount > 0)
+                CLogger::LogPrintf(SString("Forced %d player(s) to reconnect so they can update to %s\n", playersRedirectedCount, *strKickMin));
```
botder: > `%d player(s)`

I assume `playersRedirectedCount` is a `size_t` (should be`unsigned long` on Win32). Then you probably should change that to `%u player(s)`?
Also: Please don't use **auto** if the type isn't obvious.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2441,14 +2427,14 @@ void CGame::RelayNearbyPacket(CPacket& Packet)
...
-                    sendList.push_back(pSendPlayer);
+                    sendList.Insert(pSendPlayer);
                 }
             }
         }
     }
 
     // Relay packet
-    if (!sendList.empty())
+    if (!sendList.Empty())
```
botder, line 2455: Wouldn't it make more sense to add that check to `CPlayerManager::Broadcast`?

Server/mods/deathmatch/logic/CConsoleCommands.cpp
```diff
@@ -1175,29 +1175,15 @@ bool CConsoleCommands::WhoIs(CConsole* pConsole, const char* szArguments, CClien
         // Any player requested?
         if (strcmp(szArguments, "*") == 0)
         {
-            // Iterate the players and echo their IPs and ports if anyone was requested
-            unsigned int                   uiCount = 0;
-            CPlayerManager*                pPlayerManager = pConsole->GetPlayerManager();
-            CPlayer*                       pPlayer;
-            list<CPlayer*>::const_iterator iter = pPlayerManager->IterBegin();
-            for (; iter != pPlayerManager->IterEnd(); iter++)
+            if (pConsole->GetPlayerManager()->HasAnyJoined())
```
botder: Does it still work if the player hasn't joined completely, but is already connected?

botder (dismissed):

Server/mods/deathmatch/logic/CResource.cpp
Pirulax: I dont think the type is important here. But okay, sure.

Server/mods/deathmatch/logic/CSendList.h
Pirulax: This way its easier to maintain IMHO, no copy past needed.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Pirulax: It does, but it shouldn't. Thanks!

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: Good question. I'll need to look at the code. I think the comment is wrong.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: This is cleaner, isn't it?

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: You're right.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: Well, one might say that they can put `insert` into a dassert itself, and just do a `dassert(std::get<bool>(m_Players.insert(pPlayer)));`, while in reality that would remove the `insert` in release mode.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: Sorry. It turns out the comment for `m_Players` has wrong wording. It should say `all connected players`, not `all joined players`. My bad.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: I should've added `TODO: Seems very safe`.
It means that this should be cleaned up, so that the manager manages the lifetime of the CPlayer instances.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: Wow, nice catch. Thanks!

Server/mods/deathmatch/logic/CPlayer.h
Pirulax: That's why I added the todo, so that in can be changed in a separate PR.

Server/mods/deathmatch/logic/CGame.cpp
Pirulax: I should add a cast to `unsigned` instead.

Server/mods/deathmatch/logic/CConsoleCommands.cpp
Pirulax: ~~No. I should add a method to check if there are any players connected.~~

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: The comment was wrong. `m_Players` contains all connected players (a player might be connected, but not joined yet tho)

Server/mods/deathmatch/logic/CPlayerManager.cpp
Pirulax: Turns out its not that nice of a catch.
As per the docs of `google_sparse_set` `erase()` doesn't invalidate iterators. (Because the destructor of CPlayer does indeed call `RemoveFromList`, which modifies `m_Players`, but only by calling `erase()`)

Server/mods/deathmatch/logic/CGame.cpp
Pirulax: Well, there are very few scenarios where the sendList might end up being empty. Even without this `if` the code would probably run just as fast (or faster).

Pirulax:
Requires some testing.
Would be nice if we could gather 10-20 players.
One issue that might occur is that any of the containers (m_Players, m_SocketPlayerMap, m_JoinedByBitStreamVer) gets modified while iterating thru them.
It's very unlikely tho.

Pirulax (commented):
~~Please help me out here.~~
Nvm oof, that didn't work the way I imagined it.

patrikjuvonen:

PR #1796, opened 2020-11-04
Refactor some parts of SharedUtil.Misc.h

Pirulax:
See it for yourself


Pirulax:
Actually, I feel like returning a value from the functions I've added a return value to is a bad idea.
Why?
Because, people were doing this before:
```cpp
dassert(ListContains(list, value));
ListRemoveFirst(list, value);
```
But, now they can do this:
```cpp
dassert(ListRemoveFirst(list, value));
```
Which is okay until they're in debug, but, the mommeny they switch to release it'll fail, because in release `dassert` is just `void(0)`

PR #1798, opened 2020-11-04
Add final where possible - server

Pirulax:
The previous PR, but separated:

Adding the final keyword allows the compiler to "devirtualize" (virtual) calls, aka inline them.
This will probably blow some hacky code up pretty hard, im sure.

### DIFF
--- a/Server/mods/deathmatch/logic/CConsoleClient.h
+++ b/Server/mods/deathmatch/logic/CConsoleClient.h
@@ -16,7 +16,7 @@
 #include "CConsole.h"
 #include "CLogger.h"
 
-class CConsoleClient : public CElement, public CClient
+class CConsoleClient final : public CElement, public CClient
 {
 public:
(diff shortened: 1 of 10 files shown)


ghost (approved):

qaisjp:
> This will probably blow some hacky code up pretty hard, im sure.

If so, I think whoever merges this should be careful to make sure it isn't caught in a huge build, otherwise it'll be hard to bisect nightlies and determine if this caused an issue.

Pirulax:
Im saying this based on my experience with MTA code, theres quite some questionable stuff going on.
I'll download the build, and test it later.

Also.. We could group up the changes here by class type into separate PRs, so if anything goes wrong we're going to have a separate nightly for each PR.

Edit: Al-tough, I dont think It will cause any issues here(on the server), but rather on the client.

ghost (approved):

PR #1799, opened 2020-11-04
Add final where possible - client

Pirulax:
The previous PR, but separated:

Adding the final keyword allows the compiler to "devirtualize" (virtual) calls, aka inline them.
This will probably blow some hacky code up pretty hard, im sure.


ghost (approved):

qaisjp:
> This will probably blow some hacky code up pretty hard, im sure.

If so, I think whoever merges this should be careful to make sure it isn't caught in a huge build, otherwise it'll be hard to bisect nightlies and determine if this caused an issue.

ghost (approved):

botder (approved):

patrikjuvonen (approved):

PR #1801, opened 2020-11-04
Add fmt

Pirulax:
Add [fmt](https://github.com/fmtlib/fmt)

When we switch to c++20 (and `std::format` is implemented) we can remove this dependency. 
I'd like to refactor the code to use `fmt::format` (later `std::format`), instead of ugly `SString` constructors everywhere. 
[Which aren't type safe. Eg: this will crash: `SString("%s", SStringX(Im a string"))`, because `vsprintf` expectes a c-style string..]

Also, eventually I'd to refactor SString's constructor to just forward arguments to `fmt::format`, instead of hand crafting `vsprintf`.

- [ ]  Somehow configure premake so we can use `#include <fmt/format.h>` instead of the ugly  `#include <fmt/include/fmt/format.h>` - Altough, since this dependency is temporary, this isn't a priority really.


qaisjp:
can we just wait for c++20 instead of adding another dependency?

Pirulax:
Well, however you feel like.
I'll leave this here anyways, because I'd like to have a working `::format` for my other PR (which makes use of this one).
I'm having doubts `std::format` will be implemented soon in any of the popular stds tho.

qaisjp (changes requested):
I think we should wait for c++20 instead of adding another dependency.

qaisjp:
"Set cppdialect to C++20 #1803" has been closed in favour of just waiting for C++20 to release. Going to close this too.

Issue #1792, opened 2020-11-03
Add efficient modding interface for new model ID's


TheNormalnij:
### **Is your feature request related to a problem? Please describe.**
- [x] engineReplaceModel stores file in RAM (fixed via IMG loading)
- [x] engineImportTexture stores textures in RAM (fixed via IMG loading)
- [ ] engineReplaceCol doesn't support col's library (.col with few collisions)
- [x] We can't set model flags and fix some rendering issues (object with alpha disable render for objects behind this object) and collision issues (#1702) (Fixed via #1702)
This is a reason why we can't properly load big mods

### **Describe the solution you'd like**
I want to load IMG and stream files from this archive and add some engine functions and change some API.

### **Rename API**
`engineRequestModel` can be renamed into `engineRequestNew`

### **IMG support**
Done in #1677
Main functions `engineAddImage` and `engineImageLinkModel`

### **Add engineGetStreamOffset**
```lua
local  iOffset =  engineGetStreamOffset( "txd"/"dff"/"col"/"Ifp" )
```
Will return:
0 for dff
20000 for txd
25000 for col
25256 for ipl
25575 for ifp

This function can return another values when limit adjuster will be added

### **Add TXD support**
```lua
-- Register new TXD
local txdID = engineRequestNew( "txd", "TXDNAME" ) -- will assigned to existing txd
local iStreamOffset = engineGetStreamOffset( "txd" ) + txdID -- Get stream offest
engineImageLinkModel( pIMG, "file.txd", iStreamOffset ) -- Assign file to new txd

-- Assign txd to existing files
engineSetModelTXDID( 1337, txdID )
engineSetModelTXDID( 1338, txdID )
```

### **Change engineRequestModel API**

```lua
local newModelID = engineRequestNew( "obejct", "MODEL_NAME", "MODEL_TXD_NAME", drawDistance, flags )
engineImageLinkModel( pIMG, "file.dff", newModelID  ) -- DFF has 0 stream offset
```
New will have custom name. By default assigned to existing gta file ( from 1337 model )
If MTA find txd "MODEL_TXD_NAME" this txdID will be used for this model
flags - IDE flags

Same with another types, "vehicle", "timed-object", "animated-object" "weapon-object" we will use IDE syntax

### **Add COL from IMG support**

```lua
local colID = engineRequestNew( "col", "COLNAME(?)" ) --Allocate COL id
[...]

PR #1802, opened 2020-11-05
Allow allocating new TXD's

TheNormalnij:
Part of #1792

### API:
```lua
local txdID = engineRequestTXD( "TXDNAME" )
engineFreeTXD( txdID )
```


qaisjp:
It is typical to use elements for things like this.

You mentioned earlier that you were hesitant to reuse our txd elements here. Why's that?

Is it because engineLoadTXD (which returns a txd element) has actual texture data associated with it, and this one doesn't / is more like an "empty container"?

I don't really understand the semantics.

TheNormalnij:
Because i can't use CClientTXD ("txd" element) class and I'm don't sure that a new class "allocated-txd" is a good idea.

qaisjp:
> Because i can't use CClientTXD ("txd" element) class

Why not?

TheNormalnij:
Because this id doesn't present a file. So i need put `if (m_iPoolID)` for all method's in CClientTXD class.

qaisjp:
Okay. maybe others have ideas.

what can the txd ID be used for anyway>

AlexTMjugador:
I think this function is somewhat like `engineRequestModel`, in the sense that it is just meant to return a TXD ID (slot) that can then be used for a model ID. At least from the scripting side, I can't justify the overhead of promoting the TXD ID to a fully qualified element, as the bare integer is all that is needed to work with it, and it wouldn't have any methods apart from requesting and freeing the TXD slot and associating it to a model ID. On the MTA code side, I think it would be better to follow the example of [CClientModel](https://github.com/multitheftauto/mtasa-blue/blob/master/Client/mods/deathmatch/logic/CClientModel.cpp) to keep things tidy.

qaisjp:
Okay I'm convinced.

I still don't understand what TXD IDs are for. Please can someone ELI5?

TheNormalnij:
@qaisjp i will explain.
GTA use `streamID` for dynamic loaded objects(TXD,DFF,COL,IFP,IPL...) to find them in IMG archives.
```c++
struct CStreamingInfo
{
    uint16 prevId;
    uint16 nextId;
    uint16 nextInImg;
    uint8  flg;
    uint8  archiveId;
    uint32 offsetInBlocks;
    uint32 sizeInBlocks;
    uint32 loadState;
}
CStreamingInfo (&CStreamingSA::ms_aInfoForModel)[26316] = *(CStreamingInfo(*)[26316])0x8E4CC0;
```
Info for objects stored into another arrays
```c++
// Array size is 20000
CBaseModelInfoSAInterface** CModelInfoSAInterface::ms_modelInfoPtrs = (CBaseModelInfoSAInterface**)0xA9B0C8;
```
This is array with pointers to models. Almost every model use same DFF as `modelID` (CJ, broken car parts use different loading).
TXD, COL, IFP stored into `CPool<Type>`
Models use field `txdID` as pointer(position in pool) to TXD and `animID` as pointer to IFP. GTA allocate them while read IDE. 
GTA put CStreamingInfo while read .IMG header by file type and find right slotID in pool by file name.

To find resource in .IMG from id use something like this:
```c++
streamID = POOL_TYPE_OFFSET + id
```
For TXD this:
```c++
streamID = 20000 + txdID
```

So we will use `txdID` for:
* Set another texture dictionary for model.
* Set file in .IMG for new allocated TXD.
* Or change streaming info for existing TXD's (HD mods)

qaisjp:
Thank you very much! This explanation is very insightful.

Client/game_sa/CPoolsSA.cpp
```diff
@@ -1393,14 +1393,17 @@ int CPointerNodeSingleLinkPoolSA::GetNumberOfUsedSpaces()
...
     CTextureDictonarySAInterface* pTxd = (*m_ppTxdPoolInterface)->Allocate();
     if (!pTxd)
         return -1;
 
+    typedef uint(__cdecl * Function_TxdGatHashFromName)(char* strName);
+    Function_TxdGatHashFromName getHash = (Function_TxdGatHashFromName)(0x53CF30);
+
     pTxd->usUsagesCount = 0;
-    pTxd->hash = 0;
+    pTxd->hash = getHash(&strTxdName.at(0));
```
Pirulax: I think this should be `.c_str()` instead?

Client/game_sa/CPoolsSA.h
```diff
@@ -241,7 +241,7 @@ class CPoolsSA : public CPools
     void ResetPedPoolCount() { m_pedPool.ulCount = 0; }
     void InvalidateLocalPlayerClientEntity();
 
-    uint AddTextureDictonarySlot();
+    uint AddTextureDictonarySlot(std::string strTxdName);
```
Pirulax: perhaps use `const std::string&`, because you won't be `std::move`ing the value, so a const ref is fine

Client/sdk/game/CPools.h
```diff
@@ -135,6 +135,6 @@ class CPools
     virtual void ResetPedPoolCount() = 0;
     virtual void InvalidateLocalPlayerClientEntity() = 0;
 
-    virtual uint AddTextureDictonarySlot() = 0;
+    virtual uint AddTextureDictonarySlot(std::string strTxdName) = 0;
```
Pirulax: Same.

Pirulax (changes requested):

Client/game_sa/CPoolsSA.h
TheNormalnij: Thanks

Client/game_sa/CPoolsSA.cpp
TheNormalnij: This is only draft and some API still in progress.

TheNormalnij:
I think i will use [CClientModel class](https://github.com/multitheftauto/mtasa-blue/blob/master/Client/mods/deathmatch/logic/CClientModel.cpp) as manager for new TXD's. Because we don't need managers for every type.

Pirulax:
If a TXD is considered a model in GTA as well, they sure, go ahead.

TheNormalnij:
Ready

Client/game_sa/CPoolsSA.cpp
```diff
@@ -1419,3 +1420,42 @@ int CPointerNodeSingleLinkPoolSA::GetNumberOfUsedSpaces()
 
     return iOut;
 }
+
+unsigned int CPoolsSA::AllocateTextureDictonarySlot(uint uiSlotId, std::string& strTxdName)
```
Pirulax, line 1059: Perhaps use `const char*` instead of `std string`?

Client/game_sa/CPoolsSA.cpp
```diff
@@ -1419,3 +1420,42 @@ int CPointerNodeSingleLinkPoolSA::GetNumberOfUsedSpaces()
...
+
+void CPoolsSA::RemoveTextureDictonarySlot(uint uiTxdId)
+{
+    if (!(*m_ppTxdPoolInterface)->IsContains(uiTxdId))
+        return;
+
+    typedef uint(__cdecl * Function_TxdReleaseSlot)(uint uiTxdId);
+    Function_TxdReleaseSlot freeTxdSlot = (Function_TxdReleaseSlot)(0x731E90);
+
+    freeTxdSlot(uiTxdId);
```
Pirulax: `((Function_TxdReleaseSlot)(0x731E90))(uxTxdId)` is fine imho.

Client/game_sa/CPoolsSA.h
```diff
@@ -72,6 +80,82 @@ class CPoolSAInterface
...
+            if (this->m_byteMap[index].bEmpty)
+            {
+                this->m_nFirstFree = index;
+                return index;
+            }
+            index++;
+        }
+
+        return -1;
+    };
```
Pirulax: Is there any reason to use `this->`?

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -125,4 +136,33 @@ void CClientModel::RestoreEntitiesUsingThisModel()
...
+}
+
+bool CClientModel::AllocateTXD(std::string &strTxdName)
+{
+    uint uiSlotID = g_pGame->GetPools()->AllocateTextureDictonarySlot(m_iModelID - MAX_MODEL_DFF_ID, strTxdName);
+    if (uiSlotID == -1)
+        return false;
+
+    return true;
+}
```
Pirulax: return uiSlotID != -1;

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -125,4 +136,33 @@ void CClientModel::RestoreEntitiesUsingThisModel()
...
+
+bool CClientModel::DeallocateTXD(CModelInfo* pModelInfo)
+{
+    uint uiTextureDictonarySlotID = pModelInfo->GetModel() - MAX_MODEL_DFF_ID;
+
+    for (uint uiModelID = 0; uiModelID < MAX_MODEL_DFF_ID; uiModelID++)
+    {
+        CModelInfo* pModelInfo = g_pGame->GetModelInfo(uiModelID, true);
```
Pirulax, line 188: What if `pModelInfo` is `nullptr`?

Client/mods/deathmatch/logic/CClientModel.h
```diff
@@ -32,13 +33,17 @@ class CClientModel
...
-    void             RestoreEntitiesUsingThisModel();
-    void             SetParentResource(CResource* pResource) { m_pParentResource = pResource; }
-    CResource*       GetParentResource(void) const { return m_pParentResource; }
+    int                             GetModelID(void) const { return m_iModelID; };
+    eClientModelType                GetModelType(void) const { return m_eModelType; };
+    bool                            Allocate(ushort usParentID);
+    bool                            AllocateTXD(std::string &strTxdName);
+    bool                            Deallocate(void);
+    void                            SetParentResource(CResource* pResource) { m_pParentResource = pResource; }
+    CResource*                      GetParentResource(void) const { return m_pParentResource; }
```
Pirulax: Why the extra space?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2033,3 +2033,29 @@ bool CLuaEngineDefs::EngineRestreamWorld(lua_State* const luaVM)
     g_pClientGame->RestreamWorld();
     return true;
 }
+
+uint CLuaEngineDefs::EngineRequestTXD(lua_State* const luaVM, std::string strTxdName)
+{
+    if (strTxdName.size() > 24)
+        throw std::invalid_argument("Expected string size up to 24");
```
Pirulax: I'd suggest to change this to something like:
> TXD name length shoudn't be more than 24 characters.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2033,3 +2033,29 @@ bool CLuaEngineDefs::EngineRestreamWorld(lua_State* const luaVM)
...
+    return iModelID - 20000;
+}
+
+bool CLuaEngineDefs::EngineFreeTXD(uint txdID)
+{
+    std::shared_ptr<CClientModel> pModel = m_pManager->GetModelManager()->FindModelByID(20000 + txdID);
+    if (pModel && pModel->Deallocate())
+        return true;
+
+    return false;
```
Pirulax: `return pModel && pModel->Deallocate();`

Pirulax (dismissed):

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: `pModelInfo` can't be `nullptr` here because second arg for `GetModelInfo` is `true`

Client/game_sa/CPoolsSA.cpp
TheNormalnij: Any reason?

patrikjuvonen:

TheNormalnij:
Fixed errors after merge

Client/game_sa/CPoolsSA.h
```diff
@@ -54,8 +62,76 @@ class CPoolSAInterface
...
+    {
+        m_byteMap[index].bEmpty = true;
+        m_byteMap[index].nId = 0;
+        if (index == m_nFirstFree)
+            --m_nFirstFree;
+    }
+
+    void Delete(uint index)
+    {
+        delete m_pObjects[index];
```
Pirulax: Are you sure about that?
Perhaps you meant to call the destructor?
There's no need to call `delete`.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -78,6 +78,9 @@ class CLuaEngineDefs : public CLuaDefs
     static bool                                            EngineSetModelVisibleTime(std::string strModelId, char cHourOn, char cHourOff);
     static std::variant<bool, CLuaMultiReturn<char, char>> EngineGetModelVisibleTime(std::string strModelId);
 
+    static uint CLuaEngineDefs::EngineRequestTXD(lua_State* const luaVM, std::string strTxdName);
+    static bool CLuaEngineDefs::EngineFreeTXD(uint txdID);
```
Pirulax:
```suggestion
    static uint EngineRequestTXD(lua_State* const luaVM, std::string strTxdName);
    static bool EngineFreeTXD(uint txdID);
```

Client/game_sa/CPoolsSA.h
```diff
@@ -54,8 +62,76 @@ class CPoolSAInterface
...
+        return -1;
+    };
+
+    B* Allocate()
+    {
+        uint index = GetFreeSlot();
+        if (index == -1)
+            return nullptr;
+
+        return AllocateAt(index);
```
Pirulax: Improved version of the original code.
```suggestion
        m_nFirstFree++;          // Continue after the last allocated slot
        const auto sz = m_nSize; // Storing size to avoid reloads from memory - should help out the optimizer
        for (auto i = 0u; i < sz; i++) {
            const auto slot = (m_nFirstFree + i) % sz;
            const auto e    = &m_byteMap[slot];
            if (!e->bEmpty) {
                continue;
            }
            m_nFirstFree = slot;
            e->bEmpty = false;
            e->nId++;
            return &m_pObjects[slot];
        }
        return nullptr;
```

Pirulax (commented):
Address these suggestions and I'll happily merge Sir.

Client/game_sa/CPoolsSA.h
TheNormalnij: I suggest 
```c++
m_nFirstFree = slot;
```
after
```c++
 `e->nId++;` 
```

Client/game_sa/CPoolsSA.cpp
TheNormalnij: std::string& can be replaced with string_view here

Client/game_sa/CPoolsSA.h
TheNormalnij: I suggest
```c++
m_nFirstFree = slot;
```
This helps skip original allocated items

Pirulax: Ah shit I forgot to add that. I've fixed the suggestion.
Thanks!

Client/game_sa/CPoolsSA.cpp
Pirulax: Are you sure?
IIRC a few lines down `strTxdName.resize(24)` is called

Client/game_sa/CPoolsSA.h
TheNormalnij: Hmm... I wrote this code with a little experience 3 years ago.
Now i have no clue why this line doesn't break building.
This doesn't look right at all.

Client/game_sa/CPoolsSA.h
```diff
@@ -17,6 +17,14 @@
 
 #define INVALID_POOL_ARRAY_ID 0xFFFFFFFF
 
+struct CTextureDictonarySAInterface
+{
+    RwTexDictionary* rwTexDictonary;
+    unsigned short   usUsagesCount;
+    unsigned short   usParentIndex;
+    unsigned int     hash;
+};
```
TheNormalnij: Should not be declared here i think

Client/game_sa/CPoolsSA.h
TheNormalnij:
```suggestion
```

Client/game_sa/CPoolsSA.cpp
TheNormalnij: You are right

Xenius97:
I've tested `engineSetModelTXDID` and `engineRequestTXD`, look good to me, but I noticed a possible memory leak. When you restart the resource, the requested ID won't be freed anymore (maybe if I manually call `engineFreeModel`). The same applies to `engineRequestTXD` and `engineFreeTXD`.

https://youtu.be/VoTWDzTKSjs
(Some default skins not working, See #3032)

Pirulax:
Is there a good reason as to why `engineRequestTXD` should exist instead of moving it into `engineRequestModel`?
@TheNormalnij
I personally don't see anything code wise that would prevent us from merging the 2 functions together.

> but I noticed a possible memory leak. 

It's working as intended I believe.

I've looked at the video, and it seems like the texture only gets loaded after `setElementModel` is called once again after linking the img, right?

ghost:
Well, kinda off-topic but setElementModel is big memory leak itself, see #2672

TheNormalnij:
> Is there a good reason as to why `engineRequestTXD` should exist instead of moving it into `engineRequestModel`?

Function name "engineRequestModel" confuse if it can allocate a textures dictionary.
engineRequestTXD has a little bit different allocation logic, so i preferred extract logic in the function.

I would rather have two small functions that a big one by interface segregation principle.

Pirulax:
Okay that's fair enough.

Pirulax (approved):
lgtm

Pirulax:
Don't forget to add the Wiki pages

PR #1803, opened 2020-11-05
Set cppdialect to C++20

Timic3:
This pull request sets cppdialect to C++20.

As far as I've tested, nothing broke - I ran client and server, connected to it and played a bit. Might require a bit more testing though - just to be sure.

Premake binaries directly from [here](https://github.com/premake/premake-core/actions/runs/345836649).


Pirulax:
Why did you close it? :(

Pirulax:
Why did you close it? :(

Edit: Hahahha, I see.. g++ for Linux doesn't have c++20 yet ooof. Time to upgrade guys!

Timic3:
Not working on Linux and MacOS, I will reopen as draft for now I guess.

Pirulax:
Yeah, go ahead.
I presume ccw, or someone will update the CI

qaisjp (changes requested):
I think we should wait for C++20 to release. Will close in 1 week unless someone in the MTA team wants to sponsor this, or if its popular among collaborators.

Pirulax:
I feel like this would be a really neat thing.
Theres a lot of functions that would be very useful:
- [std::span](https://en.cppreference.com/w/cpp/container/span) 
- [std::lerp](https://en.cppreference.com/w/cpp/numeric/lerp)
- constexpr std::string and std::vector
- ranges - although only implemented in GCC 
And some more. I've found myself in situations a lot, when I neeed a function, but it was c++20. Nothing crazy, but wouldve made my code easier to read.

Also, I think we should deal with the issues that this might introduce now, rather than when c++20 is actually released

Pirulax:
Even when c++ comes out we'll need to set cppdialect to c++20, so I dont understand why this got  closed?
(Correct me if Im wrong)

qaisjp:
I don't want to keep open PRs that are currently broken and won't be merged anytime soon.

When it's time, we can reopen it or you can submit a new one.

PR #1804, opened 2020-11-05
Use fmt::format instead of SString constructor for string formatting

Pirulax:
Replace all SString consructors that are used for string formatting to `fmt::format`.
Replace SString ctors that are used for number -> string conversion with std::to_string.

Todo: 
- [ ]  Rewrite SString constructor to use `fmt::format` as well.

I've made a python script (250-300 lines, terrible code), that did all this for me, thankfully.


ccw808:
Might cause an issue with localization

Pirulax:
And how could they be solved?

Edit: Searched it up on google, and found this: https://github.com/fmtlib/fmt/issues/305 . Interestingly the doc doesnt say anything about 'n'

Client/core/CConnectManager.cpp
```diff
@@ -128,7 +128,7 @@ bool CConnectManager::Connect(const char* szHost, unsigned short usPort, const c
     SString strAddress = inet_ntoa(m_Address);
     if (m_usPort && !pNet->StartNetwork(strAddress, m_usPort, CVARS_GET_VALUE<bool>("packet_tag")))
     {
-        SString strBuffer(_("Connecting to %s at port %u failed!"), m_strHost.c_str(), m_usPort);
+        SString strBuffer(_("Connecting to {} at port {} failed!"), m_strHost, m_usPort);
```
ccw808: Strings inside _() are matched and replaced depending on the selected language. e.g. "Connecting to %s at port %u failed!" is replaced with "تعذر الإتصال بـ %s في المنفذ %u!" when Arabic is selected. Changing the content of the original string will make the match fail, and probably flag the string as requiring a new translation at https://translate.mtasa.com/

Pirulax: Yeah, this was done by a script, and I didnt verify it yet. (Forgot to mark the PR as draft).
I'll correct this, thanks!

Pirulax: Yeah, this was done by a script, and I didnt verify it yet. (Forgot to mark the PR as draft).
I'll correct this, thanks!

Edit: Can I edit the `.po` files myself, to make it work? Eg.: Replace `Connecting to %s at port %u failed!`  with `Connecting to {} at port {} failed!` in the `.po` as well.

(Reposted it, as edits dont show up on dc)

qaisjp: > Can I edit the `.po` files myself

The files are generated by one of the scripts here https://github.com/multitheftauto/mtasa-blue/tree/master/utils/src I think build_gettext_catalog.py

qaisjp (changes requested):
As with #1804 I think we should wait for c++20 instead of adding another dependency.

Especially hopping from SString to fmt::format to std::string seems like a lot of unnecessary mass changes.

qaisjp:
"Set cppdialect to C++20 #1803" has been closed in favour of just waiting for C++20 to release. Going to close this too.

PR #1817, opened 2020-11-10
Refactor CBanManager

Pirulax:
I still think our implementation is horrible.
But nothing I can do about it. We can't use a database, because all calls are blocking, so it would be too slow I presume.
What do others think? Because I feel like this XML magic isn't optimal at all.

Currently, all changes to the ban list are written to the disk at the end of the frame.
So, even if you change a ban / frame, it'll rewrite the whole xml, which isn't fast at all.

This PR:
- todo...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -416,3 +416,4 @@ utils/DXFiles/
 !*.dll
 !*.exe
 utils/vswhere.exe
+refactor/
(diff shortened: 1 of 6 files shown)


Pirulax:
Not worth the effort, and possible bugs.

PR #1821, opened 2020-11-10
Optimize CCustomData 

Pirulax:
Changelog:
- Removed redundant `m_SyncedData`
- Changed map type to `CFastHashMap` as we want the absolute maximum performance
- Removed `IterBegin` and `IterSyncedBegin`. Replaced it with `GetAll()` for based range loop
- Changed `const char*` for the name (in function argument list) to `const std::string&` because when searching in the hash map we'll need to use `std::string` anyways (Thats how it works sadly)
- I'll open a separate PR for more changes, because this can be optimized even further

Note:
- This is only the server side. I'll open a separate PR - because I dont want to spam this one with changes, would make it hard to review - which will move all this to `Shared`

### DIFF
--- a/Server/mods/deathmatch/logic/CCustomData.cpp
+++ b/Server/mods/deathmatch/logic/CCustomData.cpp
@@ -11,15 +11,6 @@
 
 #include "StdInc.h"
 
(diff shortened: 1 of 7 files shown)


patrikjuvonen:

PR #1822, opened 2020-11-10
Move optimized CCustomData to Shared 

Pirulax:
As said in #1821 it'll be moved to the shared side
so here it is.


Pirulax:
I'll fix this once #1821 gets merged.

PR #1825, opened 2020-11-11
Remove CRefCountableST

Pirulax:
Removes CRefCountableST
It was used in 1 place only(it being CElementListSnapshot's base class).
This possibly makes `CElementListSnapshot` a tiny bit slower, as `shared_ptr`'s control block is atomic (unlike CRefCountableST, which is ST (single threaded), thus not atomic), but I'd say the difference is really negligible. (a few hundred clock cycles probably)


ccw808:
Would it be possible to measure the performance difference between the two methods?

Pirulax:
Yeah, but not the c++ code directly as the compiler will optimize it(so its very hard to measure it head to head), and synthetic benchmarks are prone to CPU caching, while in our case the CPU has no chance to cache it at all.
What I can do is I can measure from the API side (eg from Lua) if a specific function got any slower. Altough, again, thats pretty synthetic, and the CPU might cache something.


From [here](https://quick-bench.com/q/2nS4gcb9CPlpuVXMUOjMfwDce4o) its seems like its 3x slower, but looking at the assembly makes me wonder whenever its actually measured properly. (aka I think `AddRef` and `Release` actually got optimized away, so the code doesnt do anything), I presume the compiler has realized that incrementing and decrementing the values after each other results in no change.

Even if this code is in fact 3x slower, compared to how big overhead Lua has, its really negligible. (As thats the only place this is used at).

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -1579,25 +1574,22 @@ float CClientEntity::GetDistanceBetweenBoundingSpheres(CClientEntity* pOther)
...
-        m_pChildrenListSnapshot->reserve(m_Children.size());
-        for (CChildListType::const_iterator iter = m_Children.begin(); iter != m_Children.end(); iter++)
-        {
-            m_pChildrenListSnapshot->push_back(*iter);
-        }
+        m_pChildrenListSnapshot->reserve(m_Children.size()); // Potentially faster than using vector's range constructor
+
+        // Use stupid iterator loop, because CFastList::itreator isn't std compatible..
+        for (auto it = m_Children.begin(); it != m_Children.end(); it++)
+            m_pChildrenListSnapshot->push_back(*it);
```
botder:
```suggestion
        m_pChildrenListSnapshot->reserve(m_Children.size());

        for (auto iter = m_Children.begin(); iter != m_Children.end(); iter++)
            m_pChildrenListSnapshot->push_back(*iter);
```

Server/mods/deathmatch/logic/CElement.cpp
```diff
@@ -1499,25 +1494,22 @@ void CElement::UpdateSpatialData()
...
-        m_pChildrenListSnapshot->reserve(m_Children.size());
-        for (CChildListType::const_iterator iter = m_Children.begin(); iter != m_Children.end(); iter++)
-        {
-            m_pChildrenListSnapshot->push_back(*iter);
-        }
+        m_pChildrenListSnapshot->reserve(m_Children.size()); // Potentially faster than using vector's range constructor
+
+        // Use stupid iterator loop, because CFastList::itreator isn't std compatible..
+        for (auto it = m_Children.begin(); it != m_Children.end(); it++)
+            m_pChildrenListSnapshot->push_back(*it);
```
botder:
```suggestion
        m_pChildrenListSnapshot->reserve(m_Children.size());

        for (auto iter = m_Children.begin(); iter != m_Children.end(); iter++)
            m_pChildrenListSnapshot->push_back(*iter);
```

botder (changes requested):

botder (approved):

PR #1826, opened 2020-11-11
Remove CRefCountable from wherever possible

Pirulax:
This PR removes `CRefCountable` where using it doesnt make any sense / can be replaced by `unique_ptr`


Pirulax:
Turns out you can't remove `CRefCountable` because the object needs to be kept alive when the `static` callback is called.

PR #1827, opened 2020-11-12
Refactor the database manager to use shared_ptr for CDBResultData

Pirulax:
This PR removes `CAutoRefedPointer` usage from the database manager, because its useless. The struct is small enough to fit on the stack without any issues.
This is a preparation to remove `CAutoRefedPointer`


Pirulax:
Nah man I wont finish this in the near future
Too many things to test.

PR #1831, opened 2020-11-12
Fix builds broken due to removed cdm resource

qaisjp:
I will rebase merge this so that the commits are separate


qaisjp:
Quick force merge because we need nightlies back so we can test the recent fixes

PR #1832, opened 2020-11-13
Refactor lua_newtable to lua_createtable wherever possible

Pirulax:
As with vectors in c++, you can preallocate memory an amount of items in a table. See [lua_createtable refman](https://pgl.yoyo.org/luai/i/lua_createtable).
Doing this has the same advantage as with vectors: less reallocations occur (especially with array like tables), which can potentially save a lot of CPU for frequently used functions such as `getElementsByType`, where 4-5 allocations can occur with a high player count. (I presume Lua allocates the array with power of 2, so thats at least 8 reallocations if you have > 128 players)

I also have refactored here and there.
I also have changed `unsigned int`, `int`, and any counter variable for lua table index to `lua_Number` which I stumbled upon (The reason being that, lua_pushnumber accepts a lua_Number, so counting in `int` or any non-fp type might impact performance. I wouldn't be surprised if the compiler has noticed this, and just changed the type by itself)

There are some `generous preallocations`, which means that we preallocate some amount of items we think fits, as we focus on CPU usage here, the RAM overhead of allocating a few more entries than needed is very negligible.
Edit: When reserving for the array part, a value is 16 bytes (`sizeof(TValue) == 16` that is). I think its beneficial to just preallocate whatever amount possible. 

What to watch out for when reviewing:
- Accidentally starting `lua_Number` indexing from 0 instead of 1 (Note that I always postincrement, while the old code used preincrement, or `i + 1`)
- Anything silly that doesnt make sense

Every function - except the first 3 files - has a separate commit so it can get cherry-pick merged into master on demand.


Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
```diff
@@ -436,12 +434,14 @@ int CLuaFunctionDefs::GetFunctionsBoundToKey(lua_State* luaVM)
...
-
-            // Add all the bound functions to it
+            // Get all bound functions
             unsigned int                          uiIndex = 0;
             list<CScriptKeyBind*>::const_iterator iter = m_pClientGame->GetScriptKeyBinds()->IterBegin();
+
+            // Create new table
+            lua_createtable(luaVM, m_pClientGame->GetScriptKeyBinds()->CountBinds(), 0);
+
+            // Push values into the table
```
qaisjp: This allocates space for all script binds, even if there are no functions bound to the key. In most cases I think there will only be 0-8 functions bound to a key.

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
```diff
@@ -572,13 +572,9 @@ int CLuaFunctionDefs::GetCommandsBoundToKey(lua_State* luaVM)
...
                     bCheckHitState = true, bHitState = false;
             }
-
-            // Create a new table
-            lua_newtable(luaVM);
-
-            // Add all the bound commands to it
             list<CKeyBind*>::const_iterator iter = g_pCore->GetKeyBinds()->IterBegin();
 
+            lua_createtable(luaVM, 0, g_pCore->GetKeyBinds()->CountAll());
```
qaisjp: Similarly, this allocates spaces for all binds, even if there are no commands bound to the key. In most cases I think there will only be 0-8 commands bound to a key.

Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp
```diff
@@ -409,7 +409,8 @@ int CLuaACLDefs::aclListRights(lua_State* luaVM)
     if (!argStream.HasErrors())
     {
         // Create a table to return into
-        lua_newtable(luaVM);
+        lua_createtable(luaVM, pACL->CountRights(), 0);
```
qaisjp: I think this is too generous.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11710,20 +11710,6 @@ bool CStaticFunctionDefinitions::RemoveBan(CBan* pBan, CPlayer* pResponsible)
...
-    unsigned int                uiIndex = 0;
-
-    for (; iter != m_pBanManager->IterEnd(); iter++)
-    {
-        lua_pushnumber(pLua, ++uiIndex);
-        lua_pushban(pLua, *iter);
-        lua_settable(pLua, -3);
-    }
-    return true;
-}
```
qaisjp, line 11740: I think you forgot to remove the header.

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -154,13 +154,17 @@ void CLuaArguments::PushAsTable(lua_State* luaVM, CFastHashMap<CLuaArguments*, i
...
+
         // using registry to make it fail safe, else we'd have to carry
         // either lua top or current depth variable between calls
         lua_setfield(luaVM, LUA_REGISTRYINDEX, "cache");
     }
-
-    lua_newtable(luaVM);
+    
+    // No lua_createtable here, because we dont know what type the key will be
+    lua_newtable(luaVM);
```
qaisjp, line 167: Lets leave CLuaArguments changes to its own pull request, as this could have wide reaching implications.

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -1180,7 +1180,9 @@ int CLuaAudioDefs::GetSoundMetaTags(lua_State* luaVM)
             }
             else
             {
-                lua_newtable(luaVM);
+                // Preallocate a generous amount
+                lua_createtable(luaVM, 0, 15);
```
qaisjp, line 1204: Could this be 13 instead?

```
                lua_createtable(luaVM, 0, 13);
```

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -1344,29 +1344,19 @@ int CLuaAudioDefs::GetSoundEffects(lua_State* luaVM)
...
-            {
-                lua_pushboolean(luaVM, pPlayerVoice->IsFxEffectEnabled((*iter).second));
-                lua_setfield(luaVM, -2, (*iter).first.c_str());
-            }
-            return 1;
-        }
+            if (pSound)
+                lua_pushboolean(luaVM, pSound->IsFxEffectEnabled(effect));
+            else
+                lua_pushboolean(luaVM, pPlayerVoice->IsFxEffectEnabled(effect));
```
qaisjp: Where is pPlayerVoice set?

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -922,13 +922,11 @@ int CLuaElementDefs::GetElementsWithinColShape(lua_State* luaVM)
...
-            // Create a new table
-            lua_newtable(luaVM);
+            unsigned int uiIndex = 0;
+            auto         iter = pColShape->CollidersBegin();
 
-            // Add all the elements within the shape to it
-            unsigned int                        uiIndex = 0;
-            CFastList<CClientEntity*>::iterator iter = pColShape->CollidersBegin();
-            for (; iter != pColShape->CollidersEnd(); iter++)
+            lua_createtable(luaVM, pColShape->CountColliders(), 0);
```
qaisjp: Can we do something like this:

```suggestion
            // Preallocate space if no filter was provided
            if (szType)
                lua_newtable(luaVM);
            else
                lua_createtable(luaVM, pColShape->CountColliders(), 0);

```

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -3682,20 +3682,19 @@ int CLuaGUIDefs::GUIGetChatboxLayout(lua_State* luaVM)
...
-                        lua_settable(luaVM, -3);
-                        lua_pushnumber(luaVM, 2);
-                        lua_pushnumber(luaVM, iG);
-                        lua_settable(luaVM, -3);
-                        lua_pushnumber(luaVM, 3);
-                        lua_pushnumber(luaVM, iB);
-                        lua_settable(luaVM, -3);
-                        lua_pushnumber(luaVM, 4);
-                        lua_pushnumber(luaVM, iA);
-                        lua_settable(luaVM, -3);
```
qaisjp, line 3699: I think this way is clearer, and that you should only change it to `lua_createtable`.

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -3704,17 +3703,21 @@ int CLuaGUIDefs::GUIGetChatboxLayout(lua_State* luaVM)
...
                         ss.clear();
                         ss.str(strCVarValue);
-                        ss >> fX >> fY;
-                        lua_newtable(luaVM);
-                        lua_pushnumber(luaVM, 1);
-                        lua_pushnumber(luaVM, fX);
-                        lua_settable(luaVM, -3);
-                        lua_pushnumber(luaVM, 2);
-                        lua_pushnumber(luaVM, fY);
-                        lua_settable(luaVM, -3);
```
qaisjp, line 3718: I think this way is clearer, and that you should only change it to `lua_createtable`.

Client/mods/deathmatch/logic/luadefs/CLuaNetworkDefs.cpp
```diff
@@ -124,20 +124,24 @@ int CLuaNetworkDefs::GetRemoteRequests(lua_State* luaVM)
...
     // Grab virtual machine
     if (pResource)
         pLuaMain = pResource->GetVM();
 
-    lua_newtable(luaVM);
-    for (const auto& request : g_pClientGame->GetRemoteCalls()->GetCalls())
+    const auto& calls = g_pClientGame->GetRemoteCalls()->GetCalls();
+
+    // Might reserve a lot more than needed, but it'll get GCd pretty fast, so.. CPU over RAM
+    lua_createtable(luaVM, calls.size(), 0);
```
qaisjp, line 137: I have uncertainty about this change.

Also, it's `...` not `..`.

Client/mods/deathmatch/logic/luadefs/CLuaTimerDefs.cpp
```diff
@@ -155,27 +155,29 @@ int CLuaTimerDefs::GetTimers(lua_State* luaVM)
...
             // Add all the timers with less than ulTime left
             CLuaTimerManager*                     pLuaTimerManager = pLuaMain->GetTimerManager();
             CTickCount                            llCurrentTime = CTickCount::Now();
-            unsigned int                          uiIndex = 0;
             CFastList<CLuaTimer*>::const_iterator iter = pLuaTimerManager->IterBegin();
-            for (; iter != pLuaTimerManager->IterEnd(); iter++)
+
+            // Create a new table
+            // If no time is specified we preallocate CountTimers(), otherwise none
+            lua_createtable(luaVM, dTime ? 0 : pLuaTimerManager->CountTimers(), 0);
```
qaisjp: I think you should also add a `bool bFilterTime = dTime != 0` and use that here and below.

Also, can we split this into an if/else with newtable and createtable? I think it'll make for clearer code.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -610,6 +607,9 @@ int CLuaVehicleDefs::GetVehicleOccupants(lua_State* luaVM)
             return 1;
         }
 
+        // Create a new table
+        lua_createtable(luaVM, 0, ucMaxPassengers); // seats start a 0, so we need a hashtable
```
qaisjp:
```suggestion
        // Seats start at 0, so we need to create a hashtable
        lua_createtable(luaVM, 0, ucMaxPassengers);
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -798,20 +798,18 @@ int CLuaVehicleDefs::GetVehicleUpgrades(lua_State* luaVM)
...
             {
-                if (usSlotStates[ucSlot] != 0)
+                if (usSlotStates[i] != 0)
                 {
-                    lua_pushnumber(luaVM, ++uiIndex);
-                    lua_pushnumber(luaVM, usSlotStates[ucSlot]);
+                    lua_pushnumber(luaVM, (lua_Number)i + 1);
+                    lua_pushnumber(luaVM, usSlotStates[i]);
                     lua_settable(luaVM, -3);
                 }
```
qaisjp: Previously it would create a contiguous array, now it creates a map. ipairs won't work as it previously did, on this function.

I also think `ucSlot` is clearer as it's not just a random index, it's the slot index.

I think this commit is broken and should be reverted.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -2655,7 +2655,7 @@ int CLuaVehicleDefs::GetVehicleHandling(lua_State* luaVM)
     {
         CHandlingEntry* pEntry = pVehicle->GetHandlingData();
 
-        lua_newtable(luaVM);
+        lua_createtable(luaVM, 2, 33);
```
qaisjp: Why is `2` here?

```
        lua_createtable(luaVM, 0, 33);
```

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -150,13 +150,15 @@ void CLuaArguments::PushAsTable(lua_State* luaVM, CFastHashMap<CLuaArguments*, i
...
+        // We reserve a lot of entries here, because this table will be freed as soon as pushing is finished
+        lua_createtable(luaVM, 0, 100);
+
         // using registry to make it fail safe, else we'd have to carry
         // either lua top or current depth variable between calls
         lua_setfield(luaVM, LUA_REGISTRYINDEX, "cache");
     }
 
-    lua_newtable(luaVM);
+    lua_createtable(luaVM, m_Arguments.size(), 0);
```
qaisjp, line 161: Lets leave CLuaArguments changes to its own pull request, as this could have wide reaching implications.

Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
```diff
@@ -190,7 +190,8 @@ int CLuaUtilDefs::Split(lua_State* luaVM)
         char*        szToken = strtok(strText, strDelimiter);
 
         // Create a new table
-        lua_newtable(luaVM);
+        // Reserve a generous amount, because it'll get GCd fast probably
```
qaisjp: This isn't particularly generous.

```suggestion
        // Reserve a small amount
```

Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
```diff
@@ -557,7 +557,7 @@ int CLuaUtilDefs::PregMatch(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        lua_newtable(luaVM);
+        lua_createtable(luaVM, 8, 0); // Preallocate a generous amount
```
qaisjp:
```suggestion
        // Reserve a small amount
        lua_createtable(luaVM, 8, 0);
```

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -656,55 +656,45 @@ int CLuaVehicleDefs::GetVehicleSirenParams(lua_State* luaVM)
 int CLuaVehicleDefs::GetVehicleSirens(lua_State* luaVM)
 {
     CScriptArgReader argStream(luaVM);
-    CVehicle*        pVehicle = NULL;
-    SSirenInfo       tSirenInfo;
+    CVehicle*  pVehicle = NULL;
+    unsigned char    ucSirenID = 0;
```
qaisjp, line 660: Seems to be unused

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -876,6 +873,10 @@ int CLuaVehicleDefs::GetVehicleOccupants(lua_State* luaVM)
             return 1;
         }
 
+        // So, someone decided to start seats from 0 in Lua as well..
```
qaisjp:
```suggestion
        // So, someone decided to start seats from 0 in Lua as well...
```

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -1158,20 +1158,18 @@ int CLuaVehicleDefs::GetVehicleUpgrades(lua_State* luaVM)
...
-            for (; ucSlot < VEHICLE_UPGRADE_SLOTS; ucSlot++)
+            for (size_t i = 0; i < VEHICLE_UPGRADE_SLOTS; i++)
             {
-                if (usSlotStates[ucSlot] != 0)
+                if (usSlotStates[i] != 0)
                 {
-                    lua_pushnumber(luaVM, ++uiIndex);
-                    lua_pushnumber(luaVM, usSlotStates[ucSlot]);
+                    lua_pushnumber(luaVM, (lua_Number)i + 1);
+                    lua_pushnumber(luaVM, usSlotStates[i]);
```
qaisjp, line 1172: Previously it would create a contiguous array, now it creates a map. ipairs won't work as it previously did, on this function.

I also think `ucSlot` is clearer as it's not just a random index, it's the slot index.

I think this commit is broken and should be reverted.

Server/mods/deathmatch/logic/luadefs/CLuaTimerDefs.cpp
```diff
@@ -174,27 +174,29 @@ int CLuaTimerDefs::GetTimers(lua_State* luaVM)
...
             // Add all the timers with less than ulTime left
             CLuaTimerManager*                     pLuaTimerManager = pLuaMain->GetTimerManager();
             CTickCount                            llCurrentTime = CTickCount::Now();
-            unsigned int                          uiIndex = 0;
             CFastList<CLuaTimer*>::const_iterator iter = pLuaTimerManager->IterBegin();
-            for (; iter != pLuaTimerManager->IterEnd(); ++iter)
+
+            // Create a new table
+            // If no time is specified we preallocate CountTimers(), otherwise none
+            lua_createtable(luaVM, dTime ? 0 : pLuaTimerManager->CountTimers(), 0);
```
qaisjp: I think you should also add a `bool bFilterTime = dTime != 0` and use that here and below.

Also, can we split this into an if/else with newtable and createtable? I think it'll make for clearer code.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -1238,35 +1238,33 @@ int CLuaResourceDefs::getResourceACLRequests(lua_State* luaVM)
...
-            lua_newtable(luaVM);                     // new table
-            lua_pushnumber(luaVM, i + 1);            // row index number (starting at 1, not 0)
-            lua_pushvalue(luaVM, -2);                // value
-            lua_settable(luaVM, -4);                 // refer to the top level table
-
-            const SAclRequest& request = Result[i];
-            lua_pushstring(luaVM, "name");
-            lua_pushstring(luaVM, request.rightName.GetFullName());
-            lua_settable(luaVM, -3);
+            lua_createtable(luaVM);         // new table
```
qaisjp: No parameters? How does this compile

Server/mods/deathmatch/logic/luadefs/CLuaNetworkDefs.cpp
```diff
@@ -221,22 +221,25 @@ int CLuaNetworkDefs::GetRemoteRequests(lua_State* luaVM)
...
     if (pResource)
         pLuaMain = pResource->GetVirtualMachine();
 
     if (!argStream.HasErrors())
     {
-        lua_newtable(luaVM);
+        const auto& calls = g_pGame->GetRemoteCalls()->GetCalls();
+
+        // Might reserve a lot more than needed, but it'll get GCd pretty fast, so.. CPU over RAM
+        lua_createtable(luaVM, calls.size(), 0);
```
qaisjp, line 235: I have uncertainty about this change.

Also, it's `...` not `..`.

Server/mods/deathmatch/logic/luadefs/CLuaNetworkDefs.cpp
```diff
@@ -221,22 +221,25 @@ int CLuaNetworkDefs::GetRemoteRequests(lua_State* luaVM)
...
 
-        for (const auto& request : g_pGame->GetRemoteCalls()->GetCalls())
+        lua_Number i = 1;
+        for (const auto& request : calls)
         {
-            if (!pResource || request->GetVM() == pLuaMain)
+            if (!pLuaMain || request->GetVM() == pLuaMain)
```
qaisjp: Why change this? The code behaves the same, except the old code seems clearer to me.

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -989,12 +989,10 @@ int CLuaElementDefs::getElementsWithinColShape(lua_State* luaVM)
...
     if (!argStream.HasErrors())
     {
-        // Create a new table
-        lua_newtable(luaVM);
-
-        // Add all the elements within the shape to it
         unsigned int              uiIndex = 0;
         list<CElement*>::iterator iter = pColShape->CollidersBegin();
+
+        lua_createtable(luaVM, pColShape->CountColliders(), 0);
```
qaisjp: Can we do something like this:

```cpp
            // Preallocate space if no filter was provided
            if (szType)
                lua_newtable(luaVM);
            else
                lua_createtable(luaVM, pColShape->CountColliders(), 0);

```

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1028,7 +1028,7 @@ int CLuaElementDefs::getElementsWithinRange(lua_State* luaVM)
         CElementResult result;
         GetSpatialDatabase()->SphereQuery(result, CSphere{position, radius});
 
-        lua_newtable(luaVM);
+        lua_createtable(luaVM, result.size(), 0);
```
qaisjp: Please only preallocate if a filter is not provided

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1172,7 +1172,7 @@ int CLuaElementDefs::getAttachedElements(lua_State* luaVM)
     if (!argStream.HasErrors())
     {
         // Create a new table
-        lua_newtable(luaVM);
+        lua_createtable(luaVM, 2, 0); // Generous estimate
```
qaisjp, line 1165: I don't think there's any point in this change

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -507,26 +507,27 @@ int CLuaDatabaseDefs::DbFree(lua_State* luaVM)
...
-        lua_newtable(luaVM);
+        lua_pushnumber(luaVM, i++); // Push row index
+
+        lua_createtable(luaVM, 0, Result->nColumns); // Table for columns
         for (int j = 0; j < Result->nColumns; j++)
         {
             const CRegistryResultCell& cell = row[j];
 
-            // Push the column name
-            lua_pushlstring(luaVM, Result->ColNames[j].c_str(), Result->ColNames[j].size());
```
qaisjp, line 522: why this change? comments on their own lines are preferred because clang-format will add annoying padding, i think your new code is not clang formatted.

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -507,26 +507,27 @@ int CLuaDatabaseDefs::DbFree(lua_State* luaVM)
...
+
+    lua_Number i = 1;
+    for (CRegistryResultIterator iter = Result->begin(); iter != Result->end(); ++iter)
     {
         const CRegistryResultRow& row = *iter;
-        lua_pushnumber(luaVM, i + 1);
-        lua_newtable(luaVM);
+        lua_pushnumber(luaVM, i++); // Push row index
+
+        lua_createtable(luaVM, 0, Result->nColumns); // Table for columns
```
qaisjp, line 518: again, comments on own lines are preferred

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -599,10 +599,10 @@ int CLuaDatabaseDefs::DbPoll(lua_State* luaVM)
         {
             // One or more results (from multiple statements)
             lua_newtable(luaVM);
-            for (int i = 0; Result; Result = Result->pNextResult, i++)
+            for (lua_Number i = 1; Result; Result = Result->pNextResult)
             {
-                lua_pushnumber(luaVM, i + 1);
-                lua_newtable(luaVM);
+                lua_pushnumber(luaVM, i++);
```
qaisjp, line 604: Idk, this seems odd to me

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -676,16 +676,16 @@ int CLuaDatabaseDefs::ExecuteSQLQuery(lua_State* luaVM)
...
-                // for ( int i = 0; i < Result.nRows; i++ ) {
-                lua_newtable(luaVM);                     // new table
-                lua_pushnumber(luaVM, i + 1);            // row index number (starting at 1, not 0)
-                lua_pushvalue(luaVM, -2);                // value
-                lua_settable(luaVM, -4);                 // refer to the top level table
+
+                lua_createtable(luaVM, Result->nColumns, 0);  // new table
+                lua_pushnumber(luaVM, i++);                   // row index number
+                lua_pushvalue(luaVM, -2);                     // value
+                lua_settable(luaVM, -4);                      // refer to the top level table
```
qaisjp: I don't think the changes to the `i` are useful here

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -676,16 +676,16 @@ int CLuaDatabaseDefs::ExecuteSQLQuery(lua_State* luaVM)
...
+            for (CRegistryResultIterator iter = Result->begin(); iter != Result->end(); ++iter)
             {
                 const CRegistryResultRow& row = *iter;
-                // for ( int i = 0; i < Result.nRows; i++ ) {
-                lua_newtable(luaVM);                     // new table
-                lua_pushnumber(luaVM, i + 1);            // row index number (starting at 1, not 0)
-                lua_pushvalue(luaVM, -2);                // value
-                lua_settable(luaVM, -4);                 // refer to the top level table
+
+                lua_createtable(luaVM, Result->nColumns, 0);  // new table
```
qaisjp: I am worried that this will most frequently overallocate for very wide tables with many columns, with values that are null, which may or may not be a common occurence for MTA scripters

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -754,16 +754,16 @@ int CLuaDatabaseDefs::ExecuteSQLSelect(lua_State* luaVM)
         CPerfStatSqliteTiming::GetSingleton()->SetCurrentResource(luaVM);
         if (CStaticFunctionDefinitions::ExecuteSQLSelect(strTable, strColumns, strWhere, uiLimit, &Result))
         {
-            lua_newtable(luaVM);
-            int i = 0;
-            for (CRegistryResultIterator iter = Result->begin(); iter != Result->end(); ++iter, ++i)
+            lua_createtable(luaVM, Result->size(), 0);
+            lua_Number i = 1;
```
qaisjp, line 758: Same feedback to share in executeSqlQuery

Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.Server.cpp
```diff
@@ -1133,18 +1133,18 @@ int CLuaFunctionDefs::GetPerformanceStats(lua_State* luaVM)
...
-            lua_pushnumber(luaVM, r + 1);            // row index number (starting at 1, not 0)
-            lua_pushvalue(luaVM, -2);                // value
-            lua_settable(luaVM, -4);                 // refer to the top level table
+            lua_createtable(luaVM, Result.ColumnCount(), 0);    // new table
+            lua_pushnumber(luaVM, (lua_Number)r + 1);           // row index number (starting at 1, not 0)
+            lua_pushvalue(luaVM, -2);                           // value
+            lua_settable(luaVM, -4);                            // refer to the top level table
 
-            for (int c = 0; c < Result.ColumnCount(); c++)
+            for (uint c = 0; c < Result.ColumnCount(); c++)
```
qaisjp: RowCount returns an int. I think it's wrong to assign it to a uint.

qaisjp (changes requested):
This is a first pass.

Please can you explain the purpose of changing things to use `lua_Number`?

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
Pirulax: To be quite honest I dont care about overallocating.
Unless there are thousands of script binds the little difference in RAM usage won't be noticable, but reallocating the table serveral times will be.
Also, usually these tables are short lived (eg.: used once in a pairs() loop), so they'll be deallocated in a matter of seconds by the GC.

Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp
Pirulax: Note that Lua tables usually grow by power of 2. ([Source](https://stackoverflow.com/questions/29928379/how-are-lua-tables-handled-in-memory)).
But anyways, what about this:
```cpp
        if (bAll)
            lua_createtable(luaVM, pACL->CountRights(), 0);
        else
            lua_createtable(luaVM, pACL->CountRights() / 2, 0); // Don't be too generous with prealloc here
```

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
Pirulax: I personally think the benefits outweigh the drawbacks. I very rarely see tables with nulls used in MTA, if at all.
Most people use wide tables with non-null values.
Also, keep in mind, that these tables are rarely kept. Usually they're processed once with a pair/ipairs loop, and then the GC takes care of it, so even if we do allocate more than needed it wont last long.

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
Pirulax: This is the rows table.
It'll be filled by the exact amount of rows even if all columns consist of nulls.

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
Pirulax: I know this part of it pretty well, and it'll be deallocated: https://github.com/multitheftauto/mtasa-blue/blob/a4570a5eaa21408a1651bc6692bee97bd80cab3d/Client/mods/deathmatch/logic/lua/CLuaArguments.cpp#L190-L191
Preallocating a table has no side effects.

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
Pirulax: My argument is the same as earlier/above. It'll be deallocated pretty fast.

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
Pirulax: Let it be 14 :P
I hate odd numbers.
Anyways, the difference is 32 bytes.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Pirulax: Lets do half of the amount. What do you think? I think CPU time is more important than a few hundred bytes of RAM.

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
Pirulax: I disagree. What do others think?

Client/mods/deathmatch/logic/luadefs/CLuaNetworkDefs.cpp
Pirulax: My argument is the same as earlier: These tables are rarely stored. But, what do others think? I personally feel like its worth the tradeoff. 
Perhaps, we could allocate '/ 3` the amount?
It's a question of a few hunderd bytes probably (in worst case).

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Pirulax: It's still a vector. It can't be `uc`, it needs to be at least an `int`, since arithmetics require the integer type to be at least a word. I'll name it `slot`.

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
Pirulax: Seems like my comment has disappeared.
This table is `nil`'d out at the end of the function.
It has no side effects whatsoever, as said, me preallocating is the same as when Lua reallocates by itself (so unused keys are nil'd out)

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Pirulax: Let's do `count / 2`. Usually there are only a few entities inside a colshape anyways.

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Pirulax: Indeed.

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Pirulax: My comment has disappeared, again?!
This is still a vector, what makes you think it's not?

Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
Pirulax: Well, considering a split is usually a few tokens :D

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
Pirulax: `lua_pushnumber` takes a `lua_Number` (which is a double, not an unsigned int)

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -798,20 +799,18 @@ int CLuaVehicleDefs::GetVehicleUpgrades(lua_State* luaVM)
         CVehicleUpgrades* pUpgrades = pVehicle->GetUpgrades();
         if (pUpgrades)
         {
-            // Create a new table
-            lua_newtable(luaVM);
+            // Create a new table - preallocates a generous amount
+            lua_createtable(luaVM, 16, 0);
```
ghost:
```suggestion
            lua_createtable(luaVM, VEHICLE_UPGRADE_SLOTS, 0);
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -876,8 +875,8 @@ int CLuaVehicleDefs::GetVehicleCompatibleUpgrades(lua_State* luaVM)
         CVehicleUpgrades* pUpgrades = pVehicle->GetUpgrades();
         if (pUpgrades)
         {
-            // Create a new table
-            lua_newtable(luaVM);
+            // Create a new table - preallocates a generous amount
+            lua_createtable(luaVM, 30, 0);
```
ghost, line 888: Are you sure 30 is the right number?

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
ghost: I agree with qaisjp, it's better to do this in a separate PR just to be safe.

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
ghost: > 
> 
> Let's do `count / 2`. Usually there are only a few entities inside a colshape anyways.

No, this is not good. You never know. I agree with qaisjp here.

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -174,13 +174,15 @@ void CLuaPedDefs::AddClass(lua_State* luaVM)
 
 int CLuaPedDefs::GetValidPedModels(lua_State* luaVM)
 {
-    int iIndex = 0;
-    lua_newtable(luaVM);
-    for (int i = 0; i <= 312; i++)
+    // Out of these 312, 299 are valid, so reserve that much
+    lua_createtable(luaVM, 299, 0);
```
ghost:
```suggestion
    lua_createtable(luaVM, 314, 0);  
```
Let's do 313 or 314, just to be safe.

Pirulax: No need to. There aren't 314 skins actually (that's the highest skinID tho).

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Pirulax:
```suggestion
        // Preallocate space if no filter was provided 
        if (szType && szType[0])
            lua_newtable(luaVM); 
        else 
            lua_createtable(luaVM, pColShape->CountColliders(), 0);
```

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
```diff
@@ -436,12 +434,14 @@ int CLuaFunctionDefs::GetFunctionsBoundToKey(lua_State* luaVM)
...
-            // Create a new table
-            lua_newtable(luaVM);
-
-            // Add all the bound functions to it
+            // Get all bound functions
             unsigned int                          uiIndex = 0;
             list<CScriptKeyBind*>::const_iterator iter = m_pClientGame->GetScriptKeyBinds()->IterBegin();
+
+            // Create new table
+            lua_createtable(luaVM, m_pClientGame->GetScriptKeyBinds()->CountBinds(), 0);
```
Pirulax:
```suggestion
            lua_newtable(luaVM);
```

Client/core/CKeyBinds.h
```diff
@@ -51,6 +51,7 @@ class CKeyBinds : public CKeyBindsInterface
 
     std::list<CKeyBind*>::const_iterator IterBegin() { return m_pList->begin(); }
     std::list<CKeyBind*>::const_iterator IterEnd() { return m_pList->end(); }
+    size_t                               CountAll() const noexcept { return m_pList->size(); }
```
Pirulax:
```suggestion
```

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Input.cpp
Pirulax:
```suggestion
            lua_newtable(luaVM);
```

Client/sdk/core/CKeyBindsInterface.h
```diff
@@ -156,6 +156,7 @@ class CKeyBindsInterface
 
     virtual std::list<CKeyBind*>::const_iterator IterBegin() = 0;
     virtual std::list<CKeyBind*>::const_iterator IterEnd() = 0;
+    virtual size_t CountAll() const noexcept = 0;
```
Pirulax:
```suggestion
```

patrikjuvonen:

PR #1842, opened 2020-11-16
Fix #1840: crash on disconnect

### LINKED ISSUE #1840: Crash on disconnect
**Describe the bug**
Crash.

**To reproduce**
1. Run this code
```lua
local id = engineRequestModel( "object" )

local object = Object( id, 0, 0, 5 )

addEventHandler( "onClientResourceStop", resourceRoot, function()
	object:destroy()
end )
localPlayer:setPosition( 0, 0, 4 )
```
2. Disconnect

**Version**
r20716 <=

**Additional context**
Only event onClientResourceStop will cause crash

TheNormalnij:
Fixes #1840

### DIFF
--- a/Client/mods/deathmatch/logic/CResource.cpp
+++ b/Client/mods/deathmatch/logic/CResource.cpp
@@ -88,6 +88,10 @@ CResource::CResource(unsigned short usNetID, const char* szResourceName, CClient
 
 CResource::~CResource()
 {
+    // Force delete all objects in cache
+    // https://github.com/multitheftauto/mtasa-blue/issues/1840
+    g_pClientGame->GetElementDeleter()->DoDeleteAll();
+
     // Deallocate all models that this resource allocated earlier
     g_pClientGame->GetManager()->GetModelManager()->DeallocateModelsAllocatedByResource(this);
 


Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -88,6 +88,10 @@ CResource::CResource(unsigned short usNetID, const char* szResourceName, CClient
 
 CResource::~CResource()
 {
+    // Force delete all objects in cache
+    // https://github.com/multitheftauto/mtasa-blue/issues/1840
+    g_pClientGame->GetElementDeleter()->DoDeleteAll();
+
```
ghost, line 94: The entire purpose of having the element deleter is to call `CElementDeleter::DoDeleteAll` in pulses from `CClientGame::DoPulses`. There should be a better way to fix this.

TheNormalnij: Then `stopResource` will crashes.
`CClientGame::DoPulses` contains `m_ElementDeleter.DoDeleteAll();` at 1135

ghost: > Then stopResource will crashes.

Nope, if you do it properly. This code will fix one crash but introduce more crashes since you are forcefully deleting the objects. There's a reason why the entity deletion is delayed, i.e., to avoid dangling pointers in gta code by delaying deletion.

I investigated this crash. The `CBaseModelInfoInterface` instance is deleted before the object is internally deleted. You can use `std::shared_ptr<CClientModel>` to solve this. In `CClientObject::CClientObject`, you can increment the shared_ptr reference count by simply holding a shared_ptr and decrement shared_ptr reference count in `CClientObject::~CClientObject` by setting the shared_ptr varaible to nullptr.

In this way, when all objects are deleted, then the reference count of CClientModel will reach zero and the destructor will be called automatically.

TheNormalnij: `~CClientObject` don't delete object from GTA:SA pool. So `~CClientObject` hide obejct from `CClientModel` and `CClientModel` can't replace this model with safe ID. MTA try delete object from pool, but can't, because `modelInfo` is invalid

ghost:
Closing this in favor of #1846.

PR #1846, opened 2020-11-17
 Fix #1840: engineRequestModel crash on disconnect 

### LINKED ISSUE #1840: Crash on disconnect

ghost:
There is an issue with entity deletion order when using `engineRequestModel`. Every model in GTA SA has its own "model info" struct in `CModelInfo::ms_modelInfoPtrs`. When a resource is being unloaded, the custom model info struct gets deleted while it's still in use by peds, vehicles, or objects, and this causes a crash. This also fixes the memory leak problem with CClientModel. When memory for CClientModel is allocated using engineRequestModel, it is never freed.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientModel.cpp
+++ b/Client/mods/deathmatch/logic/CClientModel.cpp
@@ -12,20 +12,14 @@
 
 CClientModel::CClientModel(CClientManager* pManager, int iModelID, eClientModelType eModelType)
 {
-    // Init
     m_pManager = pManager;
-    m_pModelManager = pManager->GetModelManager();
     m_iModelID = iModelID;
     m_eModelType = eModelType;
-
-    m_pModelManager->Add(this);
 }
 
 CClientModel::~CClientModel(void)
 {
     Deallocate();
-
-    m_pModelManager->Remove(this);
 }
 
 bool CClientModel::Allocate(ushort usParentID)
@@ -67,13 +61,16 @@ bool CClientModel::Deallocate(void)
 {
     if (!m_bAllocatedByUs)
         return false;
-
     CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
-
-    // ModelInfo must be valid
     if (!pModelInfo->IsValid())
         return false;
+    pModelInfo->DeallocateModel();
+    this->SetParentResource(nullptr);
+    return true;
+}
 
+void CClientModel::RestoreEntitiesUsingThisModel()
+{
     auto unloadModelsAndCallEvents = [&](auto iterBegin, auto iterEnd, unsigned short usParentID, auto setElementModelLambda) {
         for (auto iter = iterBegin; iter != iterEnd; iter++)
         {
@@ -106,8 +103,8 @@ bool CClientModel::Deallocate(void)
         }
         case eClientModelType::OBJECT:
         {
-            const auto& objects = &g_pClientGame->GetManager()->GetObjectManager()->GetObjects();
-            unsigned short      usParentID = g_pGame->GetModelInfo(m_iModelID)->GetParentID();
+            const auto&    objects = &g_pClientGame->GetManager()->GetObjectManager()->GetObjects();
+            unsigned short usParentID = g_pGame->GetModelInfo(m_iModelID)->GetParentID();
 
             unloadModelsAndCallEvents(objects->begin(), objects->end(), usParentID, [=](auto& element) { element.SetModel(usParentID); });
 
@@ -128,9 +125,4 @@ bool CClientModel::Deallocate(void)
 
     // Restore DFF/TXD
     g_pClientGame->GetManager()->GetDFFManager()->RestoreModel(m_iModelID);
-
-    pModelInfo->DeallocateModel();
-
-    this->SetParentResource(nullptr);
-    return true;
 }
--- a/Client/mods/deathmatch/logic/CClientModel.cpp
+++ b/Client/mods/deathmatch/logic/CClientModel.cpp
@@ -62,10 +62,10 @@ bool CClientModel::Deallocate(void)
     if (!m_bAllocatedByUs)
         return false;
     CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
-    if (!pModelInfo->IsValid())
+    if (!pModelInfo || !pModelInfo->IsValid())
         return false;
     pModelInfo->DeallocateModel();
-    this->SetParentResource(nullptr);
+    SetParentResource(nullptr);
     return true;
 }
 
--- a/Client/mods/deathmatch/logic/CClientModelManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientModelManager.cpp
@@ -70,7 +70,7 @@ std::shared_ptr<CClientModel> CClientModelManager::FindModelByID(int iModelID)
     return nullptr;
 }
 
-std::vector<std::shared_ptr<CClientModel>> CClientModelManager::GetModelsByType(const eClientModelType type, const unsigned int minModelID)
+std::vector<std::shared_ptr<CClientModel>> CClientModelManager::GetModelsByType(eClientModelType type, const unsigned int minModelID)
 {
     std::vector<std::shared_ptr<CClientModel>> found;
     found.reserve(m_modelCount);
--- a/Client/mods/deathmatch/logic/CClientModelManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientModelManager.cpp
@@ -1,6 +1,6 @@
 /*****************************************************************************
  *
- *  PROJECT:     Multi Theft Auto 
+ *  PROJECT:     Multi Theft Auto
  *               (Shared logic for modifications)
  *  LICENSE:     See LICENSE in the top level directory
  *  FILE:        mods/deathmatch/logic/CClientModelManager.cpp
@@ -10,17 +10,8 @@
 
 #include "StdInc.h"
 
-CClientModelManager::CClientModelManager(CClientManager* pManager)
-{
-    for (ushort i = 0; i < MAX_MODEL_ID; i++)
-    {
-        m_Models[i] = nullptr;
-    }
-}
-
 CClientModelManager::~CClientModelManager(void)
 {
-    // Delete all our models
     RemoveAll();
 }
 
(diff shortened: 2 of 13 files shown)


Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -67,13 +61,16 @@ bool CClientModel::Deallocate(void)
...
     if (!m_bAllocatedByUs)
         return false;
-
     CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
-
-    // ModelInfo must be valid
     if (!pModelInfo->IsValid())
         return false;
+    pModelInfo->DeallocateModel();
+    this->SetParentResource(nullptr);
```
Pirulax: `SetParentResource`

Client/mods/deathmatch/logic/CClientModelManager.cpp
```diff
@@ -77,14 +70,14 @@ CClientModel* CClientModelManager::FindModelByID(int iModelID)
     return nullptr;
 }
 
-std::vector<CClientModel*> CClientModelManager::GetModelsByType(const eClientModelType type, const unsigned int minModelID)
+std::vector<std::shared_ptr<CClientModel>> CClientModelManager::GetModelsByType(eClientModelType type, const unsigned int minModelID)
```
Pirulax, line 73: Nitpicky, but I'd say using a `ModeList_t` would be a better option here.
Or just move the return type to the rhs.

Client/mods/deathmatch/logic/CClientModelManager.cpp
```diff
@@ -77,14 +70,14 @@ CClientModel* CClientModelManager::FindModelByID(int iModelID)
...
+    std::vector<std::shared_ptr<CClientModel>> found;
     found.reserve(m_modelCount);
 
     for (int i = minModelID; i < MAX_MODEL_ID; i++)
     {
-        CClientModel* model = m_Models[i];
+        std::shared_ptr<CClientModel> model = m_Models[i];
         if (model && model->GetModelType() == type)
         {
             found.push_back(model);
```
Pirulax: I think this loop can be refactored to be a range based for loop.
I didnt do it for whatever reason back then.
If you decide to keep it this way, make sure `std::shared_ptr<CClientModel> model` is a `const std::shared_ptr<CClientModel>&` or even better `const auto& model` here, because theres no need (yet) to copy the shared ptr (as it's not as cheap as copying a pointer).

Client/mods/deathmatch/logic/CClientModelManager.h
```diff
@@ -23,24 +24,23 @@ class CClientModelManager
     friend class CClientModel;
 
 public:
-    CClientModelManager(class CClientManager* pManager);
+    CClientModelManager(class CClientManager* pManager) {}
```
Pirulax: `= default`

Client/mods/deathmatch/logic/CClientModelManager.h
```diff
@@ -23,24 +24,23 @@ class CClientModelManager
...
+    std::shared_ptr<CClientModel> FindModelByID(int iModelID);
 
+    std::vector<std::shared_ptr<CClientModel>> GetModelsByType(eClientModelType type, const unsigned int minModelID = 0);
 
     void DeallocateModelsAllocatedByResource(CResource* pResource);
 
 private:
-    CClientModel* m_Models[MAX_MODEL_ID];
-    unsigned int m_modelCount = 0;
+    std::shared_ptr<CClientModel> m_Models[MAX_MODEL_ID];
```
Pirulax, line 44: Perhaps we could use `std::array` instead?
All of it's elements are default initalized, so we woudn't need to null it out in the ctor.

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -737,4 +738,5 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
     bool              m_bTaskToBeRestoredOnAnimEnd;
     eTaskType         m_eTaskTypeToBeRestoredOnAnimEnd;
     bool              m_bWarpInToVehicleRequired = false;
+    std::shared_ptr<CClientModel> m_clientModel;
```
Pirulax, line 741: Indentation.

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -28,7 +28,7 @@ int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
 
     // Gather our custom skin model IDs allocated with engineRequestModel
     // (there might be some < 313 as well, and since we don't want duplicates, we start at 313, others are already included by the loop above)
-    for (const CClientModel* model : m_pManager->GetModelManager()->GetModelsByType(eClientModelType::PED, 313)) 
+    for (auto model : m_pManager->GetModelManager()->GetModelsByType(eClientModelType::PED, 313))
```
Pirulax: `const auto&`, again, no need to copy the shared ptr, as its locked already by the vector.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -582,10 +582,10 @@ int CLuaEngineDefs::EngineRequestModel(lua_State* luaVM)
             {
                 int iModelID = m_pManager->GetModelManager()->GetFirstFreeModelID();
                 if (iModelID != INVALID_MODEL_ID) {
-                    CClientModel* pModel = m_pManager->GetModelManager()->FindModelByID(iModelID);
+                    std::shared_ptr<CClientModel> pModel = m_pManager->GetModelManager()->FindModelByID(iModelID);
                     if (pModel == nullptr)
-                        pModel = new CClientModel(m_pManager, iModelID, eModelType);
-
+                        pModel = std::make_shared<CClientModel>(m_pManager, iModelID, eModelType);
+                    m_pManager->GetModelManager()->Add(pModel);
```
Pirulax, line 588: Let's move this to `CClientModelManager` instead, this doesnt belong to the API code.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -631,9 +631,9 @@ int CLuaEngineDefs::EngineFreeModel(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        CClientModel* pModel = m_pManager->GetModelManager()->FindModelByID(iModelID);
-
-        if (pModel && pModel->Deallocate())
+        auto modelManager = m_pManager->GetModelManager();
+        std::shared_ptr<CClientModel> pModel = modelManager->FindModelByID(iModelID);
+        if (pModel && modelManager->Remove(pModel))
```
Pirulax, line 636: I'd argue that adding a function to remove a model by ID would be a better solution.

Pirulax (commented):
If you want I can commit the suggestions.

Dutchman101:
We should try to merge this ASAP, because together with PR #1847 it will make nightlies stable again.

Pirulax:
In that case merge it asap, we can do the suggested changes later.

Dutchman101:
> In that case merge it asap, we can do the suggested changes later.

It can easily wait until saml1er checked out the review, he doesn't take that long. I meant we should ideally merge it within a few days

Client/mods/deathmatch/logic/CClientModelManager.cpp
```diff
@@ -10,45 +10,38 @@
...
-bool CClientModelManager::Remove(CClientModel* pModel)
+bool CClientModelManager::Remove(const std::shared_ptr<CClientModel>& pModel)
 {
-    if (pModel && m_Models[pModel->GetModelID()] != nullptr)
+    int modelId = pModel ? pModel->GetModelID() : 0;
+    if (pModel && m_Models[modelId] != nullptr)
     {
-        m_Models[pModel->GetModelID()] = nullptr;
+        m_Models[modelId]->RestoreEntitiesUsingThisModel();
+        m_Models[modelId] = nullptr;
```
qaisjp: Something looks weird here. Should this just return false if pModel is null?

qaisjp (commented):
mostly looks good. pirulax has some good comments.

Client/mods/deathmatch/logic/CClientModelManager.cpp
Pirulax: Ah yes, indeed

ghost: We shouldn't be verifying a reference in the first place. I'll remove the check.

Client/mods/deathmatch/logic/CClientModelManager.h
ghost: An array of shared_ptr is initialized to nullptr by default. The standard guarantees it.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
ghost: `CClientModelManager::Remove` will use the model id to free the reference. It really doesn't matter if we get the model id outside the function or inside the function.

ghost:
Some of the suggestions by Pirulax requires me to refactor the old code. This PR is more of a crash fix rather than a refactor. I think it would be better to do this in a separate PR (which I probably won't do).

About `shared_ptr<T> const&` there are discussions online against using const& for shared_ptr mainly due to safety. Some suggest that it affects performance. The performance penalty is really not much of a problem, it depends on the situation. Effect C++ book states `Item 20: Prefer pass-by-reference-to-const to pass-by-value` There's a [good article](https://medium.com/@vgasparyan1995/pass-by-value-vs-pass-by-reference-to-const-c-f8944171e3ce) that suggests shared_ptr is an exception to this rule because it can be unsafe in some situations, and it's easy to make a mistake.

qaisjp (approved):
I agree, and that makes sense. This looks good. :shipit:

Client/mods/deathmatch/logic/CClientModelManager.h
Pirulax: Ah, I thought we still have a `memset` for the array. My bad.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
Pirulax: Yes, but `Remove` could do this mess, instead of us.
(Also, its a little slower because of the shared ptr copy, but its negligible)

Pirulax:
Yeah, but in the case of the `vector` loop the `const&` still stands, as the vector won't be deallocated (thus the `shared_ptr` references released) until the loop is done.
Anyways, this is only a fix, not a refactor, so you're right in the end.

PR #1848, opened 2020-11-18
Make `logIn` password argument optional

StrixG:
It would be useful for custom auto-login implementations.


qaisjp (approved):

Pirulax:
Perhaps just use `std::optional<std::string_view>` instead of `const char* szPassword, bool bCheckPassword`?
Or `std::optional<const char*>` if you really need to stick to using `c-style strings`.

Anyways, if this gets merged I'll refactor it 🦘

ghost (approved):

PR #1850, opened 2020-11-18
Move Packet_ functions out from CGame.cpp

Pirulax:
This moves out all `Packet_` functions from the Server's CGame.cpp into a sub-directory called `packet_process`.
This way it's "only" 2500 lines, instead of 5000.


qaisjp (changes requested):
the new files are missing newlines at the end.

Pirulax:
Forgot the run the fromatter. My bad.

qaisjp:
also, filenames should probably be `.hpp`

Pirulax:
Though the same, but all our files are `.h`, even tho `cpp` uses `.hpp`.
I can do that anyways.

Pirulax:
Well that format didnt do much

Pirulax:
For some reason when I try to invoke `clang-format -style=E:/git-repos/mtasa-blue/` (the path being where .clang-format is) it just hangs. It uses 0% CPU, and nothing happens.
Even without giving the `style` option it hangs.
Any ideas?

Pirulax:
@qaisjp I coudn't run clang on the whole folder for any reason.
Would it be possible to merge this as it is, and let it be formatted in the future? (that is, when we run clang on the repo)

qaisjp:
try clicking this https://github.com/multitheftauto/mtasa-blue/blob/master/utils/win-apply-clang-format.bat

(and only pushing changes to files that you modified instead of the entire repo)

Pirulax:
I might as well just add them manually.

PR #1855, opened 2020-11-21
 Refactor LogIn function 

Pirulax:
Reopening as the previous branch was here, instead on my fork. Sorry about that. Need to get used to commit access

Use std::optional for password in CAccountManager as well.
Remove dead code CStaticFnDefs::LogIn

### DIFF
--- a/Server/mods/deathmatch/logic/CAccountManager.cpp
+++ b/Server/mods/deathmatch/logic/CAccountManager.cpp
@@ -481,7 +481,7 @@ void CAccountManager::RemoveAll()
     DeletePointersAndClearList(m_List);
 }
 
-bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, const char* szAccountName, const char* szPassword, bool bCheckPassword)
+bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, const char* szAccountName, std::optional<SString> password)
 {
     // Is he already logged in?
     if (pClient->IsRegistered())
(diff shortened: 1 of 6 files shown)


StrixG (approved):
Thank you! Feel free to merge.

ghost (approved):

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -530,7 +530,8 @@ bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, const char*
             pEchoClient->SendEcho(SString("login: Account for '%s' is already in use", szAccountName).c_str());
         return false;
     }
-    if (bCheckPassword && (!IsValidPassword(szPassword) || !pAccount->IsPassword(szPassword)))
```
qaisjp, line 533: is the `IsValidPassword` not required anymore?

qaisjp: it looks like this function is checked before passwords are used https://github.com/multitheftauto/mtasa-blue/commit/bec0300e5fd456eeed775c773af7d99372de52cf

Pirulax: Have I deleted it?

qaisjp: yes

Pirulax: My bad... Must have overlooked something.

qaisjp: test 123

Pirulax:
```cpp
bool CAccountManager::IsValidPassword(const SString& strPassword)
{
    if (strPassword.length() < MIN_PASSWORD_LENGTH)
        return false;
    return true;
}
```
`IsPassword` will check if the password is empty, so this check was redundant.
Should I add it back anyways?

qaisjp: Probably a good idea

PR #1865, opened 2020-11-23
Refactor exports client side

Pirulax:
I had opened a PR (#1517) for the same thing, but I have a lot of unrelated changes there, so this is a cleaned up version of it, for easier review, and faster merge.

It basically uses a hash set instead of a list to check whenever a function is registered as exported. (It does the same as the old code)


Pirulax:
Tested, it works.
Ready for merge.

Lpsd (approved):
lgtm 👍

PR #1870, opened 2020-11-24
Make 3118 and 3553 usable

### LINKED ISSUE #1869: Two objects cause crash
**Describe the bug**
Crash

**To reproduce**
```lua
Object( 3118, 0, 0, 5 )
Object( 3553, 0, 0, 5 )
```

Crash offset 0x59F8B4

**Version**
Client from r20716

TheNormalnij:
Fixes #1869

This PR add collision for 3118 and 3553. Model 3118 is broken itself, but still usable with custom DFF and COL.

### DIFF
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -494,6 +494,7 @@ void CGameSA::Initialize()
     // Initialize garages
     m_pGarages->Initialize();
     SetupSpecialCharacters();
+    SetupBrokenModels();
     m_pRenderWare->Initialize();
 
     // *Sebas* Hide the GTA:SA Main menu.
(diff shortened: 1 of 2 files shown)


Client/game_sa/CGameSA.cpp
```diff
@@ -814,6 +815,12 @@ void CGameSA::SetupSpecialCharacters()
     */
 }
 
+void CGameSA::SetupBrokenModels()
+{
+    ModelInfo[3118].GetInterface()->pColModel = ModelInfo[3059].GetInterface()->pColModel;
+    ModelInfo[3553].GetInterface()->pColModel = ModelInfo[3554].GetInterface()->pColModel;
```
ghost, line 821: Is the col model always loaded?

TheNormalnij: Col is preloaded for all models

ghost (approved):

Allerek:
We should add 19831 too.
Barbeque model from "Misappropriation" mission.
Dev prineside says its SAMP but it was in game itself.

image

TheNormalnij:
Can you provide more info about this model? This model is not registered in GTA SA

TheNormalnij:
I think this is cutscene model from `cutscene.img`. You can use this model via `engineRequestModel`. You need manually find dff and txd there and add right col

PlatinMTA:
> We should add 19831 too.
> Barbeque model from "Misappropriation" mission.
> Dev prineside says its SAMP but it was in game itself.
> 
> image

A lot of SAMPs custom models are from the base game itself, if the ID is more than 19000 then probably is just added in the IMG of SAMP.

PR #1873, opened 2020-11-25
Revert "Add final to all public CPacket inheriting classes (#1790)"

ghost:
There's a new packet crash in nightly. Reverting this commit might fix it.

image

PR #1882, opened 2020-11-29
Revert "setWaterLevel: add option to change water level outside world…

Pirulax:
According to our research over at [#development](https://discordapp.com/channels/278474088903606273/366384007535001612/782640347288174594), and thanks to [Dutchman's help](https://pastebin.com/ssyaeH8z) the issue seems to be caused by this commit. (Since it modifies the MapInfo packet, and even after reverting bd895707df2204393e2c0586c63a46065a7d2150 it still exists I presume this is the commit causing the issue).

As of my knowledge the client bitstream version reflects the one the stream was written with.
Eg.: Server is at 0x10, client at 0x20, the bitstream (that is, `bitStream.getVersion()` on the client) will at most be 0x10.
If not, then [this line](https://github.com/multitheftauto/mtasa-blue/commit/483c4d0b84f6791f67e73f9a4e45d69bcdc5eaae#diff-ca2b02caf256208d2211a21b062c3202b7334801dffa4106bbf052ab3c85a12fR2241-R2248) causes the issue, since a `bit` might be `true` at the point when `ReadBit` is called, which in turn reads 4 bytes (the float), and hence the whole packet is corrupted.

### DIFF
--- a/Shared/sdk/net/bitstream.h
+++ b/Shared/sdk/net/bitstream.h
@@ -444,10 +444,6 @@ enum class eBitStreamVersion : unsigned short
     // 1.5.8 RELEASED - 2020-10-11
     //
 
-    // setWaterLevel: add bIncludeWorldSeaLevel and bIncludeOutsideWorldLevel
-    // 2020-11-03 0x70
-    SetWaterLevel_ChangeOutsideWorldLevel,
-
     // This allows us to automatically increment the BitStreamVersion when things are added to this enum.
     // Make sure you only add things above this comment.
     Next,
--- a/Client/game_sa/CWaterManagerSA.cpp
+++ b/Client/game_sa/CWaterManagerSA.cpp
@@ -742,25 +742,17 @@ bool CWaterManagerSA::SetPositionWaterLevel(const CVector& vecPosition, float fL
     return SetPolyWaterLevel(pPoly, fLevel, pChangeSource);
 }
 
-bool CWaterManagerSA::SetWorldWaterLevel(float fLevel, void* pChangeSource, bool bIncludeWorldNonSeaLevel, bool bIncludeWorldSeaLevel, bool bIncludeOutsideWorldLevel)
+bool CWaterManagerSA::SetWorldWaterLevel(float fLevel, void* pChangeSource, bool bIncludeWorldNonSeaLevel)
 {
     assert(m_bInitializedVertices);
     CVector vecVertexPos;
-
-    if (bIncludeWorldSeaLevel || bIncludeWorldNonSeaLevel)
+    for (DWORD i = 0; i < NUM_DefWaterVertices; i++)
     {
-        for (DWORD i = 0; i < NUM_DefWaterVertices; i++)
-        {
-            m_Vertices[i].GetPosition(vecVertexPos);
-            if ((bIncludeWorldNonSeaLevel && m_Vertices[i].IsWorldNonSeaLevel()) || (bIncludeWorldSeaLevel && !m_Vertices[i].IsWorldNonSeaLevel())) 
-                vecVertexPos.fZ = fLevel;
-            m_Vertices[i].SetPosition(vecVertexPos, pChangeSource);
-        }        
+        m_Vertices[i].GetPosition(vecVertexPos);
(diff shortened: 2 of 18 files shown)


Pirulax:
Uh, I'm not sure what to do about the bitstream version tho.... Older servers might still write that portion(that is, the bit + float), which would cause newer clients(that is, which have this PR) to crash

TheNormalnij:
> then [this line](https://github.com/multitheftauto/mtasa-blue/commit/483c4d0b84f6791f67e73f9a4e45d69bcdc5eaae#diff-ca2b02caf256208d2211a21b062c3202b7334801dffa4106bbf052ab3c85a12fR2241-R2248) causes

Can you provide better link?

Pirulax:
Yes, https://github.com/multitheftauto/mtasa-blue/blob/94c6aba300460199690b66a38b6db36f4fe63a20/Client/mods/deathmatch/logic/CPacketHandler.cpp#L2517

It is the same crash as in #1873

Pirulax:
> > then [this line](https://github.com/multitheftauto/mtasa-blue/commit/483c4d0b84f6791f67e73f9a4e45d69bcdc5eaae#diff-ca2b02caf256208d2211a21b062c3202b7334801dffa4106bbf052ab3c85a12fR2241-R2248) causes
> 
> Can you provide better link?

Ah, my bad, I see..
This line:
```cpp
if (bitStream.Can(eBitStreamVersion::SetWaterLevel_ChangeOutsideWorldLevel))
```
Just search for it, I can't give a link to it, for whatever reason. It's in `CPacketHandler::Packet_MapInfo(NetBitStreamInterface& bitStream)`

TheNormalnij:
Why `bitStream.Can` doesn't work there? Server and client use different bitstream version?

TheNormalnij:
How to repro related issue?

Pirulax:
> How to repro related issue?

I have no clue. It seems like connecting to a server running an older version (that is, one that doesnt have this commit) will make the client (which has this commit) crash.

That's the part I dont understand either(from the description):
> As of my knowledge the client bitstream version reflects the one the stream was written with.
Eg.: Server is at 0x10, client at 0x20, the bitstream (that is, bitStream.getVersion() on the client) will at most be 0x10.
If not, then this line causes the issue, since a bit might be true at the point when ReadBit is called, which in turn reads 4 bytes (the float), and hence the whole packet is corrupted.

ghost:
Yep, connecting is enough to get a crash (i was connecting to server running r20693).

TheNormalnij:
But i can connect to older MTA version, So `bitStream.Can` works good here. Maybe `r20693` server is broken itself?

TheNormalnij:
r20693 server and client are unavailable to download. Can we confirm this bug for another versions?

Pirulax:
They're available for download.
Just go [here](https://nightly.mtasa.com), and click on:
- [x] show older files

TheNormalnij:
20694 and 20689 only

Pirulax:
Ah, thats because such nighlites dont exist.

Pirulax:
Interesting.. [here](https://buildinfo.mtasa.com/?Revision=20693), see the `Use enum classes for BitStream (#1737)`.

TheNormalnij:
check next version  `Make sure BitStream comparisons still work correctly (#1752)`

Shared/sdk/net/bitstream.h
```diff
@@ -446,12 +446,12 @@ enum class eBitStreamVersion : unsigned short
 
     // setWaterLevel: add bIncludeWorldSeaLevel and bIncludeOutsideWorldLevel
     // 2020-11-03 0x70
-    SetWaterLevel_ChangeOutsideWorldLevel,
+    // SetWaterLevel_ChangeOutsideWorldLevel, - Reverted commit
```
TheNormalnij, line 449: I think we can't do this, because servers with `SetWaterLevel_ChangeOutsideWorldLevel` bitstream version will send `SetWaterLevel_ChangeOutsideWorldLevel`  packets for clients with `PedEnterExit` version.

Pirulax:
Turns out the issue is only present in r20693, it was later fixed in r20694.

PR #1890, opened 2020-12-02
Fix #1836: setObjectScale onClientPedsProcessed event

### LINKED ISSUE #1836: setObjectScale doesn't work in the onClientPedsProcessed event
**Describe the bug**
if we create an object on the server or client and reduce it, the object will be the standard size in the event. 

**To reproduce**
run the code
```lua
addEventHandler ( "onClientPedsProcessed", root,
    function ( )
        local obj = getElementData( localPlayer, "obj" )
        if obj then
            setElementPosition( obj, x,y,z+1 )
            setObjectScale( obj, 0.5 )
        end
    end
)
addEventHandler( "onClientResourceStart", resourceRoot,
function ( startedRes )
    local obj = createObject( 1946, 0,0,0 )
    setElementData( localPlayer, "obj", obj)
end)
```

**Expected behaviour**
the object should shrink

**Screenshots**
onClientRender
https://imgur.com/YKvy302
onClientPedsProcessed
https://imgur.com/SCGjniT

**Version**
1.5.8-20724

**Additional context**
I think the object should decrease, since the event was made so that there was no lag. I mean, if it is used in the bone_attach resource, it will be of no use
https://www.youtube.com/watch?v=U95ls5931lE

ghost:
~~This PR adds an extra parameter `forceUpdate` to setObjectScale.~~
~~When this new parameter is set to true, setObjectScale should be called in every frame in onClientPedsProcessed event.~~

UPDATE: setObject will now work and doesn't need to be called every frame.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
@@ -414,9 +414,9 @@ int CLuaObjectDefs::StopObject(lua_State* luaVM)
 int CLuaObjectDefs::SetObjectScale(lua_State* luaVM)
 {
     //  bool setObjectScale ( object theObject, float scale )
-    CClientEntity*   pEntity;
-    CVector          vecScale;
-    bool             forceUpdate = false;
+    CClientEntity* pEntity;
+    CVector        vecScale;
+
     CScriptArgReader argStream(luaVM);
     argStream.ReadUserData(pEntity);
 
@@ -436,10 +436,9 @@ int CLuaObjectDefs::SetObjectScale(lua_State* luaVM)
         argStream.ReadNumber(vecScale.fY, vecScale.fX);
         argStream.ReadNumber(vecScale.fZ, vecScale.fX);
     }
-    argStream.ReadBool(forceUpdate, false);
     if (!argStream.HasErrors())
     {
-        if (CStaticFunctionDefinitions::SetObjectScale(*pEntity, vecScale, forceUpdate))
+        if (CStaticFunctionDefinitions::SetObjectScale(*pEntity, vecScale))
         {
             lua_pushboolean(luaVM, true);
             return 1;
--- a/Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
@@ -414,9 +414,9 @@ int CLuaObjectDefs::StopObject(lua_State* luaVM)
 int CLuaObjectDefs::SetObjectScale(lua_State* luaVM)
 {
     //  bool setObjectScale ( object theObject, float scale )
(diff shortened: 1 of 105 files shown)


qaisjp:
This feels like a hack

Pirulax:
My exact feelings.

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -436,9 +436,10 @@ int CLuaObjectDefs::SetObjectScale(lua_State* luaVM)
         argStream.ReadNumber(vecScale.fY, vecScale.fX);
         argStream.ReadNumber(vecScale.fZ, vecScale.fX);
     }
+    argStream.ReadBool(forceUpdate, false);
```
TheNormalnij: I think there should be a way to get variable `eventName` from current LuaVM. This will remove additional argument from API

Pirulax: You can, its a global.
```cpp
lua_getglobal(luaVM, "eventName");
std::string_view eventName = lua_tostring(luaVM, -1);
// Not sure if it needs to be popped: lua_pop(luaVM);
```

ghost:
> 
> 
> This feels like a hack

Which part of the code needs to be changed? The extra parameter or the fact the function needs to be called per frame in onClientPedsProcessed? I'd love to fix both of these things, but it's hard to do it without hacking it. The entities are pre-rendered (matrices are updated) before onClientPedsProcessed is called, and this is why setObjectScale doesn't work in onClientPedsProcessed event.

 I can try to change the rendering order and make entities pre-render after onClientPedsProcessed. In this way, we won't need to modify setObjectScale, and it doesn't need to be called every frame in onClientPedsProcessed, but then the frame delay bug for bones will be back and that would defeat the purpose of having this new event in the first place.

qaisjp:
The extra parameter on the Lua API feels like a hack. I don't have insight into the other things.

ghost:
> 
> 
> The extra parameter on the Lua API feels like a hack. I don't have insight into the other things.

I see. I think I just got an idea. I'll fix this.

ghost:
I removed the extra parameter from setObjectScale, and it doesn't need to be called every frame now. MTA will pre-render objects after `onClientPedsProcessed` event is triggered.

StrixG:
The scale of attached objects is always 1 after this change.
https://youtu.be/XXYxbcS2NH0

image

ghost:
> 
> 
> The scale of attached objects is always 1 after this change.
> https://youtu.be/XXYxbcS2NH0

I see. Is this a new bug?

StrixG:
> > The scale of attached objects is always 1 after this change.
> > [youtu.be/XXYxbcS2NH0](https://youtu.be/XXYxbcS2NH0)
> 
> I see. Is this a new bug?

I can't reproduce it on 20751 and erlier versions.

PR #1891, opened 2020-12-02
Add getTowHitch/BarPosition

Pirulax:
This PR adds 2 functions that can be pretty useful. And since they're already in the game, why not expose them, right? :D
[towthingtest.zip](https://github.com/multitheftauto/mtasa-blue/files/5629160/towthingtest.zip)
Teleport to `2042.39624, 1449.91211, 10.67180` (use `setpos 2042.39624, 1449.91211, 10.67180`)
The green line is the tow hitch position, red one is the tow bar.
Only applicable vehicles have their tow bar accurately calculated(using the `misc_a` component), other's have it behind the vehicle, in the air.
Also, while regular vehicles obviously dont have a properly calculated `tow hitch` position (its in the engine deck) I left it in, since I think it might be useful for someone.
Here's how it looks like:
image

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicle.cpp
+++ b/Client/mods/deathmatch/logic/CClientVehicle.cpp
@@ -4634,6 +4634,27 @@ void CClientVehicle::RemoveVehicleSirens()
     m_tSirenBeaconInfo.m_ucSirenCount = 0;
 }
 
+std::optional<CVector> CClientVehicle::GetTowHitchPos()
+{
+    if (!m_pVehicle)
+        return std::nullopt;
+
+    CVector out;
+    m_pVehicle->GetTowHitchPos(&out);
+    return out;
+}
+
+std::optional<CVector> CClientVehicle::GetTowBarPos(CClientVehicle* pAttachTo, bool bIgnoreModelType)
+{
+    CVehicle* pAttachToGameVeh = pAttachTo->GetGameVehicle();
+    if (!m_pVehicle || !pAttachToGameVeh)
+        return std::nullopt;
+
+    CVector out;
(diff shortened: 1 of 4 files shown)


Pirulax:
I've tested it with a few vehicles, and it seems to work just fine. That is, it doesnt crash.
Also, the `getTowBarPosition` has a required second argument, which is the trailer:
Heres a little cpp code:
```cpp
  if ( mi == MODEL_PETRO
    || mi == MODEL_RDTRAIN
    || mi == MODEL_LINERUN
    || mi == MODEL_ARTICT3
    || mi == MODEL_UTILITY && attachTo && attachTo->physical.m_Entity.m_nModelIndex == MODEL_UTILTR1
    || (mi == MODEL_BAGGAGE || mi == MODEL_TUG || mi == MODEL_BAGBOXA || mi == MODEL_BAGBOXB)
    && attachTo
    && ((v5 = attachTo->physical.m_Entity.m_nModelIndex, v5 == MODEL_BAGBOXA)
     || v5 == MODEL_BAGBOXB
     || v5 == MODEL_TUGSTAIR) )
  {
    if ( automobile->m_aCarNodes[CAR_MISC_A] )
    {
      v6 = RwFrameGetLTM(automobile->m_aCarNodes[CAR_MISC_A]);
      v7 = v6->pos.x;
      v6 = (RwMatrixTag *)((char *)v6 + 48);
      outPos->x = v7;
      outPos->y = v6->right.y;
      outPos->z = v6->right.z;
      return 1;
    }
  }
```
Here `mi` is the model of the towing vehicle.
Perhaps we could just reverse the function, and add it to `CClientVehicle`? (So we could remove this `if` check)

botder:
Did you forget to add `GetTowHitchPos` and `GetTowBarPos` to game_sa?

Issue #1217, opened 2020-01-21
Add "ipb" to default startup resources


Dutchman101:
Would it be a good idea to add [ipb](https://github.com/multitheftauto/mtasa-resources/tree/master/%5Badmin%5D/ipb) (ingame performance browser) to default startup resources in mtaserver.conf?

I think that sending high usage warnings to all online admins when a script begins to mess up (or starts getting abused by someone to cause load) can be really helpful, and I can't think of any cons. Since we always aim to provide ideal configuration..

Dutchman101:
Should we also use it for local and map editor configs? I think that can also be helpful to identify problems during the creation process (e.g scripting, or loading maps that demand too much so they can notice and resolve it on the spot)

qaisjp:
Good idea (I doubt non-scripters use the local server anyway)

CrosRoad95:
what about `map performance browser` focused on map, rendering ect? not really scripts?

Dutchman101:
qaisjp Today at 5:38 PM
@Dutchman101 go for the commit

Done, fixed in 2bebc75

PR #1903, opened 2020-12-05
Remove some startup resources

CrosRoad95:
In case someone is lazy :)


qaisjp:
Why?

patrikjuvonen:
Won't fix.

jlillis:
I think we should take a second look at this - some of these startup resources aren't in common use with most players and don't need to be running by default. I would add ipb and performancebrowser to that list.

Dutchman101:
> I would add ipb and performancebrowser to that list

Bad idea.. ipb recently got added to auto startup list, and such things only get added if others think it's a good idea and if there are reasons for it.

That was issue #1217 (see discussion about ipb being added)
Performance browser is for similar reasons probably also a keeper, to warn and alert server admins that regularly use webadmin about bad scripts and lock ups.

PR #1911, opened 2020-12-08
Convert CStreaming to use fn pointers instead of asm

Pirulax:


ghost (approved):

fastman92:
I've taken a look into this.
This pull request puts new bugs to the MTA.
__fastcall calling convention used, where __cdecl should have been used.

By the way, please change  CStreamingSA::GetStreamingInfoFromModelId(ushort id) into this:

CStreamingSA::GetStreamingInfoFromModelId(uint32 id)

ghost:
> __fastcall calling convention used, where __cdecl should have been used.

Yes, it should be __cdecl.

Pirulax:
In my IDB `CStreaming::RequestSpecialModel` uses `__fastcall`.
I won't change `GetStreamingInfoFromModelId` to use `uint`, because it requires quite a few changes which aren't related to this PR.

forkerer:
IDA guesses the calling convention, it often assumes wrong one.

PR #1914, opened 2020-12-11
Fix a few new parser issues

Pirulax:
This PR fixes some issues with the new parser not being able to push some comibination of types, because of the definition order of `Push`.

Also, it fixes some stupid case where `std::vector<CElement*>` was pushed as an array of bools (For whatever reason `Push<bool>` was preferred over the actual pointer `Push`)

### DIFF
--- a/Shared/mods/deathmatch/logic/lua/LuaBasic.h
+++ b/Shared/mods/deathmatch/logic/lua/LuaBasic.h
@@ -127,6 +127,32 @@ namespace lua
         return 1;
     }
 
+    /*****************************************************************\
+    * The functions below may depend on each other, so they need to be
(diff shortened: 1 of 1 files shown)


Pirulax:
Oopsie.
Edit: Fixed.

ghost (approved):

sbx320 (approved):
Not sure I like the solution as discussed on #development, but I don't have a better one either. Guess this works for the time being

CrosRoad95:
i fixed my issues where i was trying to return map of vector of variant ect. previously it does not work

Pirulax:
Yeah I dont like the solution either, but I dont have a better one sadly.

PR #1915, opened 2020-12-11
getElementsWithinRange with interior and dimension check

Pirulax:
Add interior and dimension check to `getElementsWithinRange`.
Also, a small optimization, so now it uses the type hash, instead of the type name, hopefully its faster. (Perhaps I should do that in a separate PR?)

[within-range-test.zip](https://github.com/multitheftauto/mtasa-blue/files/5681286/within-range-test.zip)

Credits: Jayceon and Hexon for giving me the idea.

Note: Contains commits of #1914, because the parser pushed bools onto the stack instead of the elements.


Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -953,43 +954,30 @@ int CLuaElementDefs::GetElementsWithinColShape(lua_State* luaVM)
...
-    argStream.ReadString(elementType, "");
-
-    if (!argStream.HasErrors())
-    {
-        // Query the spatial database
-        CClientEntityResult result;
-        GetClientSpatialDatabase()->SphereQuery(result, CSphere{position, radius});
+    // Remove elements that do not match the criterias
+    result.erase(std::remove_if(result.begin(), result.end(), [&](CClientEntity* pEntity) {
+        if (typeHash && typeHash != pEntity->GetTypeHash())
```
sbx320: This is a (minor) behavioral change when `type = ""`. Previously it matched any type, now it doesn't match any type. Might be good to keep it in case someone misunderstood default arguments.

sbx320 (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Pirulax: Yeah I really should have not added this change, because its unrelated. Perhaps I should change it back, and create a separate PR?

Dutchman101:
Recently, when i asked @Jusonex why the Z argument isn't used (like the [wiki page](https://wiki.multitheftauto.com/wiki/GetElementsWithinRange) contains a note: `Z argument isn't in use currently, but make your scripts like it is for future compatibility reasons` to see if he had any plans for adding it anyways, he said that he wasn't aware of not adding support for it while creating this function (84c680d) and that someone must have filled that in based on discovering the issue of Z argument not working. So, even though the wiki note implies so, it's not by design at all.

Jusonex said he'll look into it soon.. but now that you're working on this function, how about fixing it as well @Pirulax? Or should we create a new issue about the Z argument not working?

Actually reading back on the wiki edit history, it's you (Pirulax) that added the note about Z argument not working.. do you recall a specific #development discussion about the reason why? Because if Jusonex isn't aware and you are still aware, then maybe he just forgot the reason.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
sbx320: Changing it is fine, just needs another check for `!typeHash.empty()` to keep the old behavior

sbx320:
The Z argument isn't used, as `CClientSpatialDatabaseImpl::SphereQuery` doesn't support Z for some reason. Instead it does:
```
    // Make a box from the sphere
    CBox box(sphere.vecPosition, fabsf(sphere.fRadius));
    // Make everything 2D for now
    box.vecMin.fZ = SPATIAL_2D_Z;
    box.vecMax.fZ = SPATIAL_2D_Z;
```
Thus a sphere query is more like a cylinder query. I'm not sure why this was limited, but I'd be careful about changing it, since the spatial database is used in quite a few places. I'd keep that to a separate PR.

Pirulax:
Well its because our entire RTree implementation is using 2 dimensions only :D
https://github.com/multitheftauto/mtasa-blue/blob/6efde5d2ad68404012697488af165720d2a4ee7d/Server/mods/deathmatch/logic/CSpatialDatabase.cpp#L15

`RTree` class template:
https://github.com/multitheftauto/mtasa-blue/blob/6efde5d2ad68404012697488af165720d2a4ee7d/Shared/sdk/RTree.h#L66
No clue why it's 2D only. Altough it being 2D isn't a problem, we don't need to modify it (RTree) at all, we could just add an extra check in the lambda to filter out based on distance, eg.:
```cpp
(entity->GetPosition() - pos).LengthSq() < radiusSq
```

sbx320 (approved):

Pirulax:
@sbx320 should I add support for the Z argument? It would be an additional check inside the `std::remove_if`.

Pirulax:
Waiting for #1951 to be merged.

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1013,43 +1013,30 @@ int CLuaElementDefs::getElementsWithinColShape(lua_State* luaVM)
...
-    argStream.ReadNumber(radius);
-    argStream.ReadString(elementType, "");
-
-    if (!argStream.HasErrors())
-    {
-        // Query the spatial database
-        CElementResult result;
-        GetSpatialDatabase()->SphereQuery(result, CSphere{position, radius});
+    // Remove elements that do not match the criterias
+    result.erase(std::remove_if(result.begin(), result.end(), [&](CElement* pElement) {
```
botder: This should only run if `typeHash != 0 || interior.has_value() || dimension.has_value()` is true

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -953,43 +954,30 @@ int CLuaElementDefs::GetElementsWithinColShape(lua_State* luaVM)
...
-    argStream.ReadNumber(radius);
-    argStream.ReadString(elementType, "");
-
-    if (!argStream.HasErrors())
-    {
-        // Query the spatial database
-        CClientEntityResult result;
-        GetClientSpatialDatabase()->SphereQuery(result, CSphere{position, radius});
+    // Remove elements that do not match the criterias
+    result.erase(std::remove_if(result.begin(), result.end(), [&](CClientEntity* pEntity) {
```
botder: This should only run if `typeHash != 0 || interior.has_value() || dimension.has_value()` is true

botder (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
Pirulax: Sure, I'll do that.

botder (approved):

PR #1920, opened 2020-12-15
Disable train track functions

qaisjp:
Custom train tracks is incomplete and requires further iteration before being a full MTA release.

Since I'm no longer working on MTA and cannot continue iterating on this train tracks work, this PR disables some functions that were turned on in https://github.com/multitheftauto/mtasa-blue/pull/1715.

If you want to continue the work, feel free to refer to https://github.com/multitheftauto/mtasa-blue/pull/250

That branch is kinda broken (crashes or freezes when you get in a train, last I checked), but most of the code works. I was aiming to fix that freezing problem by slowly iterating: submitting working PRs to master and adding small features bit by bit.

So if you want to continue the train tracks work, you'll probably have to rewrite most of it (or manually copy chunks bit by bit) from the huge train tracks branch. I used a `git worktree` to keep two branches checked out in the same git repository (instead of cloning twice).

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -57,7 +57,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"isTrainDerailable", IsTrainDerailable},
         {"getTrainDirection", GetTrainDirection},
         {"getTrainSpeed", GetTrainSpeed},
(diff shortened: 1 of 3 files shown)


Pirulax:
lgtm if it doesn't break scripts.

Pirulax (approved):
lgtm if it doesnt break existing scripts.

qaisjp:
This function is part of an incomplete API, and as it's not part of any release, it had no compatibility guarantees. In that sense, it does not break any production scripts.

Please merge at your convenience.

PR #1924, opened 2020-12-16
Change IDs (file, model and TXD) from 16-bit to 32-bit

fastman92:
This changes file IDs, model IDs and TXD IDs inside the MTA code from 16-bit to 32-bit.
Prepares the MTA for the future to handle the game, which may be heavily modified.

Net protocol (BitStream related) is still kept unchanged. 16-bit IDs in the net protocol still stay 16-bit for the time being.
Altering the net functionality would require more planning and that's something to leave for the future.

Everything done properly and accurately with an understanding of existing code.

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -2067,7 +2067,7 @@ CModelCacheManager* CCore::GetModelCacheManager()
     return m_pModelCacheManager;
 }
 
-void CCore::AddModelToPersistentCache(ushort usModelId)
+void CCore::AddModelToPersistentCache(uint32 usModelId)
 {
     return GetModelCacheManager()->AddModelToPersistentCache(usModelId);
 }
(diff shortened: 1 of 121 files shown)


Pirulax:
I haven't looked at the changes yet, but I do see there are 121 files changed.
Unless you've added casts when calling `bitStream.Write/Read` it's will be broken. Eg.: `bitStream.Write(ID)` should be changed to `bitStream.Write((unsigned short)ID)`.
Have you tested it? Does the client and server work?

fastman92:
> I haven't looked at the changes yet, but I do see there are 121 files changed.
> Unless you've added casts when calling `bitStream.Write/Read` it's will be broken. Eg.: `bitStream.Write(ID)` should be changed to `bitStream.Write((unsigned short)ID)`.
> Have you tested it? Does the client and server work?

Yes. I have thought about it and altered the code wherever necessary. 

As it has been already written - everything was done properly and accurately with an understanding of existing code. Net functionality is not an exception and this problem had to be considered.

I haven't done the test of playing the game on server with others.

Pirulax:
yeah we should probably try that as well. Hit me up if you have time on the weekend, and we can try it.

botder:
Please stick to the standard types with a `_t` suffix and DO NOT mix these with non-standard types like `uint32`.
See https://en.cppreference.com/w/cpp/header/cstdint

PR #1932, opened 2020-12-20
std::monostate support in new parser

CrosRoad95:
it let you use std::monostate as an equivalent of no value. Useful for std::optional<std::variant<>> combo.

```cpp
bool CYourDefs::MonoTest(std::variant<std::monostate, float, bool> var)
{
    return var.index() > 0;
}
```
with definition like above, following syntax will valid:
```lua
monoTest()
monoTest(12.3)
monoTest(true)
```
returns true if value is given

https://en.cppreference.com/w/cpp/utility/variant/monostate

### DIFF
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -87,6 +87,8 @@ struct CLuaFunctionParserBase
             return GetClassTypeName((T)0);
         else if constexpr (std::is_same_v<T, dummy_type>)
             return "";
+        else if constexpr (std::is_same_v<T, std::monostate>)
+            return "";
     }
 
     // Reads the parameter type (& value in some cases) at a given index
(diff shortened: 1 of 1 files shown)


Pirulax:
`optional<variant<>>` should work, what's the advantage of doing it like this? There is an implicit bool operator in `optional` (Which is a synonym for `has_value()`). That is:
```cpp
bool CYourDefs::MonoTest(std::optional<std::variant<float, bool>> var)
{
    return (bool)var;
}
```

CrosRoad95:
https://github.com/multitheftauto/mtasa-blue/blob/8e725707bec9c23a2e461df671273ad18cd56065/Client/mods/deathmatch/logic/luadefs/CLuaPhysicsDefs.cpp#L1285-L1348
example case where monostate was useful, 1 function, 3 overlapping syntaxes

Pirulax:
We can add monostate.
Your solution is very bad tho. Overloads should work.

CrosRoad95:
i tried overloads before and they did not work properly

Pirulax:
So `std::monostate` should be like `LUA_TNONE`?

CrosRoad95:
if none means no value, then yes

sbx320 (approved):

Issue #1916, opened 2020-12-13
setVehicleHandling client-side function to work on vehicles created server-side


Grafu:
**Is your feature request related to a problem? Please describe.**

Please allow using client-side setVehicleHandling on remote vehicles. Server developers like me could then experiment with synchronization improvements like temporary removing mass (between collision moment and collision sync) from remote vehicles. Now the hit vehicle controller stays like a wall until collision synchronizes. And that can take a while if the remote player has high ping or packet loss.

**Describe the solution you'd like**

Remove this check 
```
if (!pVehicle->IsLocalEntity())
    return false;
```
from
```
bool CStaticFunctionDefinitions::SetVehicleHandling(CClientVehicle* pVehicle, eHandlingProperty eProperty, unsigned int uiValue)
{
    assert(pVehicle);

    if (!pVehicle->IsLocalEntity())
        return false;

    CHandlingEntry* pEntry = pVehicle->GetHandlingData();

    if (pEntry)
    {
        if (SetEntryHandling(pEntry, eProperty, uiValue))
        {
            pVehicle->ApplyHandling();
            return true;
        }
    }
    return false;
}
```
and overloads

**Describe alternatives you've considered**

Using setElementCollidableWith would make the collisions lose visual effects like car changing velocity and position on collision.
Using setVehicleHandling serverside changes the physics for remote vehicle controlling player and that is something I am trying to avoid.

**Additional context**

*What does it look like hitting a vehicle with lower mass? Can you push it around more easily?*

If you set remote vehicle mass to 1 using server-side setVehicleHandling and that vehicle is controlled by a high ping player, then pushing his vehicle is easy. Otherwise, if mass of the remote vehicle isn't lowered, your car crashes into that remote vehicle like into a wall.

*if you push it away it's gonna warp around the place when sync packets correct it*

synced packets will warp the vehicle back, but remote vehicle after sync packet won't push the local vehicle since it has no mass (very low mass = 1)
so it will get pushed out once again
and this way local vehicle doesn't lose velocity when hitting the lagging remote vehicle

I think it is worth leaving freedom for server developers to do some experimenting and maybe implementing things based on using setVehicleHandling client-side to set remote vehicle handling.
If it is something weird, server admins will just remove the script
If you set remote vehicle masses serverside to 1 and laggy player drives that vehicle, you will be able to push it easily even if push packets and sync packets lag due to player high ping

LosFaul:
biggest problems are values like engineAcceleration
[...]

PR #1935, opened 2020-12-21
Implement #1916: Make client-side setVehicleHandling to work on vehicles created server-side

Grafu:


Dutchman101:
This PR probably needs to be closed as per the below comment:


"I agree with you but this is too much effort for me. So I give up at this point. Good luck"

_Originally posted by @Grafu in https://github.com/multitheftauto/mtasa-blue/discussions/1916#discussioncomment-230443_

.. in response to a discussion about the desyncs this would bring

Will close if there's no other feedback (or more opinions regarding the 'desync' discussion) soon

LosFaul:
i think we should handle it diplomaticly

i could open a new issue based on this PR after it would be merged
which would be about extending setVehicleHandling for localPlayer controlled vehicle with sync option

what you guys think?

botder:
Adding sync to that would require a server trip to avoid desync and at that point you could've used the serverside function in the first place. I am against this pull request unless somebody has a better idea.

Fernando-A-Rocha:
> Adding sync to that would require a server trip to avoid desync and at that point you could've used the serverside function in the first place. I am against this pull request unless somebody has a better idea.

As I mentioned on the #custom-model-ids channel, intentional desync is not a problem, it is up to the developer and you can easily make "sync" by having all clients set the same handling on X vehicle. This usecase is necessary for engineRequestModel vehicles to work properly with custom handling as their model is changed (setElementModel) clientside and this already causes desync.

Thank you for the additional considerations.

Fernando-A-Rocha:
> Adding sync to that would require a server trip to avoid desync and at that point you could've used the serverside function in the first place. I am against this pull request unless somebody has a better idea.

@botder 
Please re open this PR and consider merging it. There is no need for sync here, for the same reason as there is no sync in setElementModel when it changes a server-created element's model on a client.

PlatinMTA:
>
> @botder 
> Please re open this PR and consider merging it. There is no need for sync here, for the same reason as there is no sync in setElementModel when it changes a server-created element's model on a client.
> 

The current solution is pretty bad (ask the server to set the handling after the client loads the model for example).

Having this function work like rocha asks would help quite a lot.

Xenius97:
This could be a security issue, i don't recommend to merge it to master. With lua injection you can troll servers by modifying vehicle handling (eg. on race servers).

botder:
With Lua injection you could also give your own vehicle more velocity/acceleration before the merge.

Issue #1849, opened 2020-11-18
IsValidModel should be changed


Allerek:
**Version**
nightly 1.5.8-20716 onwards

**Additional context**
#IMPORANT: I'm C++ Amateur, what im saying can be a complete CRAP!#

For now IsValidModel in *Client\mods\deathmatch\logic\CVehicleNames.cpp* returns true only if we check for standard GTA:SA models.
*ulModel >= 400 && ulModel <= 611;* 

We should change it to use 

getModelInfo(ulModel)->isVehicle();

in order to return real values when using vehicles from engineRequestModel.

This will allow us to add functions from #1175

Allerek:
^^Not a bug if thinking more, but can cause problems in future versions

Pirulax:
Seems reasonable.

TheNormalnij:
We need good vehicle manager with dynamic id support first

Pirulax:
I think all these `CClientVehicle::IsValidModel` and such functions should be moved to `CClientModelManager`, so for example.: `IsVehicleModel` would be an alias for `GetModelType() == eClientModelType::VEHICLE`, same for peds, and objects.

TheNormalnij:
`CModelInfoSA` is valid model manager. `CClientModelManager` is only manager for allocated ID's

Pirulax:
Yes I know, but I'd like to use CCLientModelManger as we'll have to support custom model names as well.

FernandoMTA:
I think this should be prioritized more. Vehicle add-ons are a game-changing feature.

Fernando-A-Rocha:
@BCG2000 sorry for the tag but do you think this still needs to be resolved?

BCG2000:
This can only be fixed when https://github.com/multitheftauto/mtasa-blue/issues/2251 is implemented, as revamping functions like this one is part of implementing it. Until then, modifying these functions individually should be avoided, since that is only working around the root issue instead of directly addressing it.

Fernando-A-Rocha:
> This can only be fixed when https://github.com/multitheftauto/mtasa-blue/issues/2251 is implemented, as revamping functions like this one is part of implementing it. Until then, modifying these functions individually should be avoided, since that is only working around the root issue instead of directly addressing it.

Just as I thought! I have the same opinion. I think we have enough temporary solutions for new vehicle models to work. We'll get the best of it once server side allocation is a thing.

Issue #1835, opened 2020-11-13
Allow allocating new ID's on server


TheNormalnij:
### API
```lua
int model = allocateModel( string type, ... config )
```
This function will allocate new modelID from the end of modelIInfo array( from 20000 to 0  )
Will return false if modelIInfo array is full

```lua
bool success = allocateModelAt( int modelID, string type, ... config )
```
This function will provide manual ID control.

```lua
bool success = freeModelAt( int modelID )
```
Will destroy all objects used by this ID and deallocate modelID 
### How to it will works
When resource with new ID's will be started, server will send packed with new registered ID's. If client has conflicts with serverside ID, client should reconnect to this server.
When client try connect to the server. It should get packed with all new allocated ID's before game and resources started and allocate new ID's

Pirulax:
I don't think we should use numeric IDs on the server. How about names instead, like:
```cpp
bool success = allocateModel(string name,  string type, ... config )
```
We'd add support for passing in model names(of server allocated IDs) to existing functions. Maybe even add support to client-sided `engineRequestModel` as well? Eg.:
```cpp
bool success = engineRequestModel(string name, ....)
```
In the case of a name clash:
- Client tried to allocate a server id: fail so we prevent desyncs
- Server tried to allocate a client id: server succeeds

Why?
- No ID clashes possible, because we're aren't returning ID's, so no need to kick anyone unless they're out of free IDs
- In the future if the client's free ID limit gets increased the server will be limited to around 5000 unique IDs
- Future proofing
- numeric model ID's are kinda unpredictable. How would one write a system which allows only certain vehicle models to do something without sending packets back and forth with the server? Managing client-side tables (sent from the server) is still a pain in the ass that would need to be maintained (by the scripter)

TheNormalnij:
> * No ID clashes possible, because we're aren't returning ID's, so no need to kick anyone unless they're out of free IDs

Model names clashes possible

> * Server tried to allocate a client id: server succeeds

Server will silently break something at client, bad idea, i think

> * In the future if the client's free ID limit gets increased the server will be limited to around 5000 unique IDs

Impossible, because server owner manages his own resources
> * Future proofing

But you need replace all current API and sync system for new system. This is too big refactor. My solution is simple.


> * numeric model ID's are kinda unpredictable

It's GTA SA design

Pirulax:
> It's GTA SA design

 GTA wasn't designed to have a whole Lua API.

[...]

Issue #1591, opened 2020-08-03
Implement Lua API version


Lpsd:
As briefly discussed in #1459, we should implement a Lua API version which is set via `meta.xml`.

We always want to ensure backwards-compatibility when making changes to code, however this may prevent us from:

1. Keeping code clean
2. Making "beneficial" changes (that would break backwards-compatibility)

An example of this is with CEGUI, where functionality in newer versions may differ from the version we're currently using. By default, we'll ensure everything looks/feels and functions the same as before, however this may prevent new features/changes in CEGUI being used (or at least implemented cleanly). 

The server/resource should decide if they care about backwards-compatibility over improved functionality.

Another example may be when "extended" function alternatives have been suggested (#638) - instead of adding a separate function we'd simply allow them to set the API version specifically for this function, e.g:

```xml
<api name="setElementData" version="123" />
```

I haven't really thought about how exactly this will be implemented or configured via meta.xml, so input here would be great.

qaisjp:
Have you thought about having versioning apply for all functions, rather than individual functions?

I'm thinking it would get a bit complicated for both developers and scripters to have to keep track of individual function versioning. It would also be easier to handle documentation-wise, as the user has one version number to remember (and we can show documentation that is valid for a single API, rather than asking the user to figure it out per-function).

I'm also thinking, as a version number, we can just use the MTA version number - something similar (or identical) to whatever value `min_mta_version` uses.

Lpsd:
I agree - that could get messy. I'll try my best to explain how I'm imagining the implementation.

For any proposed changes that could potentially break existing scripts (or change an existing feature massively), we'd place it under the current "new" API version (for instance, v2).

Let's say features X, Y and Z are changed - they'd be put into API v2.

Further down the line, if for some reason feature X is changed massively again, we have the option to increment the API version and, in this case, place it into API v3.

To couple that with your `min_mta_version` idea, you could then have something like this:

```xml
<api version="2" /> <!-- This applies the latest API v2 -->
<api version="2-20616" /> <!-- Only apply API v2 up until version 20616 -->
```

Does that make sense, or am I talking shite due to lack of knowledge on MTA versioning?

random thought: if you chose API v3, would it cascade changes from lower versions?

TheNormalnij:
This will duplicate min_client_version functionality.

Lpsd:
[...]

PR #1936, opened 2020-12-21
Ability to create server-side vehicles with custom ids

pentaflops:
Add new argument **specificID**
```lua
int engineRequestModel ( str elementType [, int parentID, int specificID ] )
```

Before creating a vehicle with a custom id, you need to request a model (server-side):
```lua
int requestModel ( int specificID, str elementType [, int parentID ] )
```
Returns false if the model has already been requested

Added new packet to sync and requesting new ID on the client custom models.
If a vehicle with a conflicting id is created on the server, the client will receive a error protocol 39.

Increased bitstream version and added backward compatibility (need add minclientversion to use requestModel)

_If the community will support these changes, then I plan to add support for skins and objects_


Pirulax:
We have an issue about this #1701.
Have you read it?

PlatinMTA:
> We have an issue about this #1701.
> Have you read it?

#1701 is already merged and it works, but only for clientside peds, which in reality won't be of much use. I guess this PR is trying to solve that.

Pirulax:
Sorry, its #1835

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -3176,22 +3176,20 @@ void CPacketHandler::Packet_EntityAdd(NetBitStreamInterface& bitStream)
...
+                    if (!CClientVehicleManager::IsStandardModel(usModel))
                     {
-                        RaiseEntityAddError(39);
-                        return;
+                        bitStream.Read(usModelOriginal);
+                        if (!CClientVehicleManager::IsValidModel(usModel) && !g_pClientGame->m_pManager->GetModelManager()->RequestModel(usModel, usModelOriginal, eClientModelType::VEHICLE, g_pClientGame->m_pManager))
+                        {
+                            RaiseEntityAddError(39);
+                            return;
+                        }
```
Pirulax: You can't simply change the value from a `char` to a `short`, this will cause packet misalignment if the server is running an older version.

Server/mods/deathmatch/logic/CModelManager.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#include "StdInc.h"
+
+bool CModelManager::RequestModel(ushort modelID, ushort parentID, eModelType modelType)
+{
+    if (IsValidModel(modelID))
+        return false;
+
+    MapSet(m_ModelsParent, modelID, parentID);
+    MapSet(m_ModelsType, modelID, modelType);
```
Pirulax: use operator[]:
```cpp
m_ModelsParent[modelID] = parentID;
m_ModelsType[modelID] = modelType;
```
Altough, I'd recommend creating a structure like `ModelInfo`:
```cpp
struct ModelInfo 
{
unsinged short parentID;
eModelType type;
};
```
And storing it it in the map instead.
Or perhaps since there's a fixed number of models (20000), you could create a static array.

Server/mods/deathmatch/logic/CModelManager.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    ++m_modelCount;
```
Pirulax: you can use the map's size.

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -1433,3 +1436,45 @@ int CLuaWorldDefs::getOcclusionsEnabled(lua_State* luaVM)
...
+
+        lua_pushboolean(luaVM, g_pGame->GetModelManager()->RequestModel(usModelID, usParentID, modelType));
+        return 1;
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
```
Pirulax: Use the new parser please.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -433,7 +433,20 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
                     // fit into a char?  Why doesn't someone document this?
                     //
                     // --slush
-                    BitStream.Write(static_cast<unsigned char>(pVehicle->GetModel() - 400));
+
+                    unsigned short usModel = pVehicle->GetModel();
+                    BitStream.Write(usModel);
+
```
Pirulax: Again, this breaks network compatibility.

Pirulax (changes requested):

Pirulax:
I dont feel like this is a good solution. 
And since it breaks backwards compatibility, we might as well just implement the solution described by me in #1835

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
TheNormalnij: Use `BitStream.Can` here

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.h
```diff
@@ -92,4 +92,7 @@ class CLuaWorldDefs : public CLuaDefs
     LUA_DECLARE(RestoreWorldModel);
     LUA_DECLARE(RestoreAllWorldModels);
     LUA_DECLARE(resetMoonSize);
-};
\ No newline at end of file
+
+    // Custom models
+    LUA_DECLARE(RequestModel);
```
TheNormalnij: I think we don't need this comment

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -1433,3 +1436,45 @@ int CLuaWorldDefs::getOcclusionsEnabled(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+int CLuaWorldDefs::RequestModel(lua_State* luaVM)
```
TheNormalnij: Can you add minclientversion check?

TheNormalnij:
> I dont feel like this is a good solution.

It's good simply solution

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -580,37 +580,38 @@ int CLuaEngineDefs::EngineRequestModel(lua_State* luaVM)
...
+                        break;
+                    default:
+                        break;
                     }
-                    pModel->Allocate(usParentID);
-                    pModel->SetParentResource(pResource);
+                }
+
+                int iModelID = INVALID_MODEL_ID;
+                if (argStream.NextIsNumber())
```
TheNormalnij, line 605: I think this argument should be first. So we can put model settings as optional args later

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
TheNormalnij: I think new parser don't support potential optional args for model settings per model type

TheNormalnij:
Please, test it with old client and new server

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Pirulax: `std::optional`?

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
Pirulax: The problem is that, this way only the clients having this feature will see the actual models, older clients wont. We must break backwards compatibility, but that can't be done now. We should wait for 1.6, where network compatibility will be broken anyways.

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
TheNormalnij: `std::optional` and `std::variant` works bad with numeric types (`std::variant<float,int>`) and old parser will be readable here i think

Pirulax: What do you mean by "works bad"?

Pirulax:
> You can't simply change the value from a char to a short, this will cause packet misalignment if the server is running an older version.

This PR must break backwards compatibility, and since it does, we might as well just do it as described in #1835

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
TheNormalnij: `std::variant<float,int>` will accept any number value as float and `std::variant<int, float>` as int (Lua 1.3333 -> C++ 1)

Pirulax: Well, yes, of course it won't chose the more *fitting* type, but why would you need such a feature anyways? You either accept an int or a float, not both. IMHO.

Allerek:
When doing so, we should allow to set custom vehicle-name when requesting ID for it. https://github.com/multitheftauto/mtasa-blue/issues/1849

If name not specified, take parent id name.

pentaflops:
> Please, test it with old client and new server

Since PR change format of reading bitstream, there is no need for this. This PR can only be used in 1.6 version

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
pentaflops: Using it as the first argument may break backward compatibility, otherwise the implementation of such support will look bad. Actually adding this argument is unnecessary for this RP

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
pentaflops: How can i do this?

Allerek:
I see we have some minor checks like:
HasPoliceRadio
DoCheckHasLandingGear
HasTaxiLight
IsTrainModel
HasSirens
etc.
But do we have
-ZR-350 lights
-Packer
-Andromeda back door
(i dont think we have more 'misc' options)
?

pentaflops:
@Pirulax check review pls

sbx320:
I don't think requesting a specific id is a good solution either as it will lead to collisions between resources.

Pirulax:
Thanks for the PR very much. I'm sorry but I (we?) have to decline it.
We've discussed how it should be implemented, and came to the conclusion that using script entities instead of ID's is a better choice (to avoid people hardcoding stuff).
I really have a bias against using integer's as model ID's as it gets really messy, especially with dynamic IDs.
See #custom-model-ids channel on Dev discord.

TheNormalnij:
Why you want avoid hardcoding stuff? Numeric IDs are easy, backward compatible, good for databases.
` script entities instead of ID's` sounds like a creepy abstraction layer.

And yeah this stuff is for manual management

pentaflops:
_From discord:_
My implementation allows you to work with custom vehicles as with a standar vehicles, **without need to adapt many current scripts for this** (only scripts current use engineRequestModel so that there are no id conflicts). And this is **important**.

Obviously, the problem of collisions should be **solved by developer** who will add new vehicles. You are now proposing to complicate something that does not require complication. Adding non-static identification **don't solve any problem**, it only complicates.

AlexTMjugador:
The problem I see with using numeric IDs as-is is that they lure people into thinking they are the simplest solution that keeps backwards compatibility, but on practical terms, when actually requesting custom model IDs in the client and the server, complexity scalates quickly, because 1) each server has to reinvent the wheel to manage this problem, 2) depend on a community resource that does this for them or 3) be ignorant of this problem and wonder why things break.

One may think that, if model IDs could only be requested on the server, then these discrepancies would be solved, but requesting a model for every client just because a single client wants to show a vehicle preview in a car shop, for example, wastes a fair amount of bandwidth. Besides, we would break backwards compatibility anyway, because model requests are clientside right now.

It also makes sense to promote element models to their own entity, as with these changes they would have associated logic and a lifetime that's not necessarily the same as the lifetime of the elements that use them, to name some reasons. Insisting on representing entities with primitive data types is known as the _primitive obsession_ antipattern and usually leads to suboptimal solutions in the long (or even short) run. Also, MTA is reasonably expected to keep serverside things synchronized in a transparent and relatively effortless manner, and this approach breaks this expectation.

For reference and clarity purposes, I made this UML class diagram showing an hypothetical OOP Lua API that I think it'd be nice to implement (with some ideas from @sbx320 and @Pirulax):

<details>
<summary>Custom model IDs API UML class diagram</summary>

Custom model IDs API UML class diagram

PlantUML code of the above diagram:

```puml
@startuml "Custom model IDs API"
title Custom model IDs API

class Element {
	+setModel(ElementModel model) : void
	+getModel() : ElementModel
	-- Rest of methods --
	...
}

class ElementModel {
	{static} +fromGameModelId(int id) : Model
	{static} +getAllModels() : Model [1..*]
	{static} +request(ElementModelType type, ElementModel parent) : CustomModel
	-int internalId
}
note left: This API is shared.\n\ninternalId is only known by MTA C++ code.\n\nWhen a client requests a new model,\nits internalId is the next free one for\nuse with GTA.\n\nWhen a server requests a new model,\nits internalId may be a sequential number,\nand clients should request a new model\n(maybe with a different internalId in each\nclient). If a client runs out of internalIds\nduring this operation, it may 1) be kicked\nor 2) the request may return false\n(preferred).\n\nClients keep a server internal ID -> actual\ninternal ID mapping, so they know how to\ntranslate server IDs in packets and so on\nappropriately to keep sync.\n\nServer databases may store models as a\n(type, ID) pair, where "type" distinguishes\nbetween GTA and custom models, and ID\ncan be an index for a Lua table entry that\ncontains data about that model, like a\nfactory method that calls\nElementModel.request to initialize the\nmodel for the first time.\n\nWhen a ElementModel element gets\ndestroyed, the elements that use it are\nreset to the parent element model or\nanother suitable default. This can happen\nwhen the resource that requested them\nstops or destroyElement() is used.

ElementModel "1" --> "0..*" ElementModel : children
Element <|.. ElementModel

enum ElementModelType {
	PED
	VEHICLE
	OBJECT
}

ElementModel ..> ElementModelType
@enduml
```

Edit: a `getAll` method for ElementModel would not be needed because it'd be yet another element type, and we can use `getElementsByType` already.
</details>

If we agree that this is a good API, then we can implement these features with less risk of them being rejected or problematic in the long term. Of course it'd depend on things like #1591, but I think it's a good way to go.

Issue #1782, opened 2020-10-31
Revision some of server side security policy


Disinterpreter:
2020 is close to be end. We have some of cool technologies for prevent and defend vulnerabilities of the server like Docker, linux jails, LXC and etc.

So I just want to suggest to add many of features which approve working with OS. Like `os.* io.* dofile require`... Lua methods.

Maybe I can't know deeper things of your security. But there are many server owners which want to allow to "talk" their servers with OS. Besides I had an experience in RAGE MP which server-side scripting so simple and it gives to scripters unique functions. (Like all DB drivers, fast connections to analytics like Grafana and a lot of great things! and we needn't to integrate LuaRocks it will be work.)

Yes, I know it is long way, and can't be solve fast. But it is really useful thing for real big servers. 

If someone agree with me I suggest to start with adding in `meta.xml` options like _unsafe_

Of course the bad side of this it is insecurity of newbie players. And that's all I suppose.

P.S (offtop) I don't designed it by the template because it was moved from another repo.

Pirulax:
Perhaps adding it with explicit config in meta.xml would be cool. 
Using ACL would slow it down a lot.

Disinterpreter:
> 
> 
> Perhaps adding it with explicit config in meta.xml would be cool.
> We could use ACL, but only to check if we should load the libs (os, etc..). Having a translation layer between the actual lua function and our would slow things down a lot.

Yes, I know. Thanks for your opinion and ideas! This problem not only about technical side. It's more about understanding modern development and scripting. I mean in the past it looked normal, MTA defend hosters and newbies. But today we have a lot of databases, lua frameworks (like MoonScript), interesting libraries, they so many and creating every day. And I think, today this server security prevent to merge new features to the server. Yes I know, I can write the module. But if module can do anything with OS, we really need to block some functions directly?

So I hope we will see many scripters which agree my idea.

qaisjp:
The Lua package system allows you to set "allowed C modules" in the server config (which allows you to load Lua C modules in server resource). We could use the same system to allow some of the default Lua packages.

qaisjp:
We also don't have to wait for Lua modules to be complete, we just need to make sure we're super happy with the config option we go for, so we reuse the same one when (if ever) we land Lua modules

Disinterpreter:
> 
> 
[...]

PR #1937, opened 2020-12-21
New server function getServerEnvironmentValue

pentaflops:
A useful function for storing passwords and other settings if using docker images
```lua
str getServerEnvironmentValue( str envName )
```


AlexTMjugador:
I like the idea of this PR. Environment variables are a nice way to pass parameters to processes in an easy to automate way, and without requiring command line parameters. I don't know why someone downvoted this, as you can pass a hash of sensitive data, like you would do in a database, and already use modules to retrieve this information.

Anyway, I'd use the new argument parser for this function. I'd also make sure that it is restricted in the ACL by default, so that no random resource can get environment variables unless the admin explicitly allows it, for privacy concerns. Finally, I'd rename the function to `getServerEnvironmentVariable`, but this is up to discussion.

Pirulax:
I'd rename it to `getServerEnvironmentVariable` (in case we decide to add some kind of support to the client in the future)

Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
```diff
@@ -111,6 +111,7 @@ class CLuaFunctionDefs
     LUA_DECLARE(SetServerPassword);
     LUA_DECLARE(GetServerConfigSetting);
     LUA_DECLARE(SetServerConfigSetting);
+    LUA_DECLARE(GetServerEnvironmentValue);
```
Pirulax, line 114:
```cpp
static std::string_view CLuaFunctionDefs::GetServerEnvironmentValue(std::string name) ;
```

Server/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -233,6 +233,7 @@ void CLuaManager::LoadCFunctions()
         {"getServerPassword", CLuaFunctionDefs::GetServerPassword},
         {"setServerPassword", CLuaFunctionDefs::SetServerPassword},
         {"getServerConfigSetting", CLuaFunctionDefs::GetServerConfigSetting},
+        {"getServerEnvironmentValue", CLuaFunctionDefs::GetServerEnvironmentValue},
```
Pirulax, line 236:
```cpp
 {"getServerEnvironmentValue", ArgumentParser<GetServerEnvironmentValue>}
```

Also, for this to work, you need to move it to `CLuaUtilDefs`.
The new parser doesn't work in `CLuaManager`.

Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.Server.cpp
```diff
@@ -686,6 +686,25 @@ int CLuaFunctionDefs::SetServerConfigSetting(lua_State* luaVM)
...
+        lua_pushstring(luaVM, getenv(strName));
+        return 1;
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
+
```
Pirulax, line 707:
```cpp
std::string_view CLuaFunctionDefs::GetServerEnvironmentValue(std::string name) 
{
    return getenv(name.c_str());
|
```
One thing im not sure about: What if `getenv` returns null? Does Lua handle it correctly?

Pirulax (changes requested):

Pirulax:
Also, there is a native Lua function called[ `os.getenv`.](https://www.gammon.com.au/scripts/doc.php?lua=os.getenv) We could enable it on the server, instead of adding a non-native function. It calls the same function we do, since Lua is pure ANSI C.

prnxdev:
What is difference between this nad `get` and `set` functions? I once wrote a script that was connecting to MySQL database and I had different auth credentials for local and production so I used those functions to manage those data. Problem is that `get` function read data correctly but It adds `*` to every key (as I remember).

In other way there is also `fileWrite` and `fileOpen` functions. I don't know how exactly this function works but if it reads data from `.env` file I think it's not so important to have one in MTA API.

patrikjuvonen:
What are the security concerns regarding getenv and the scope of variables one could have access to?

Disinterpreter:
> 
> 
> What are the security concerns regarding getenv and the scope of variables one could have access to?

It is only server side feature and I don't see the problems with accessing to the scope in this feature. If we think about unnecessary security of serverside we risky to lost all connections with OS as it was in: https://github.com/multitheftauto/mtasa-blue/discussions/1782

I can find thousand programs which work with ENV and it's okay.

patrikjuvonen:
I believe we could try implementing the ACL system from the Lua modules branch and then allow getenv through the native Lua os function as suggested by @Pirulax

AlexTMjugador:
> What is difference between this nad `get` and `set` functions? I once wrote a script that was connecting to MySQL database and I had different auth credentials for local and production so I used those functions to manage those data. Problem is that `get` function read data correctly but It adds `*` to every key (as I remember).
>
> In other way there is also `fileWrite` and `fileOpen` functions. I don't know how exactly this function works but if it reads data from `.env` file I think it's not so important to have one in MTA API.

Those functions solve a different problem: allowing the server administrator to set up configuration values for the resources. Environment variables are not meant to store resource configuration; instead, they represent more generic data about the OS environment that may be useful to processes as a whole.

Of course storing the environment variables in files via external programs and reading them back in the MTA: SA Lua VM is a workaround, but that implies writes to a physical block device, which contributes to reducing the lifespan of SSDs (unless they are done in a RAM filesystem, but using such filesystems in Windows is awkward). And, honestly, this function is simple enough to be merged quickly and successfully, so it's not like that workaround is even needed.

> Also, there is a native Lua function called os.getenv. We could enable it on the server, instead of adding a non-native function. It calls the same function we do, since Lua is pure ANSI C.

Good catch! I was concerned about how we can ACL native Lua functions, though, but @patrikjuvonen proposed a solution for that, so it's perfect.

> It is only server side feature and I don't see the problems with accessing to the scope in this feature. If we think about unnecessary security of serverside we risky to lost all connections with OS as it was in: #1782
>
> I can find thousand programs which work with ENV and it's okay.

As @sbx320 said in the discussion you linked to, backdoored and compiled server side resources are a concern. Allowing these functions by default might expose the server to be fingerprinted in undesirable ways that may not be easily noticed by the server owner, and, if the server owner improperly uses environment variables to store secrets like deployment keys, it may even hand down a potential attacker with enough information to compromise the system. Denying this function by default is a way of making the server owner aware of this, and if some random compiled DirectX GUI resource insists in needing access to functions like `os.getenv`, then it's easy to conclude that something is wrong with it.

Disinterpreter:
>  Denying this function by default is a way of making the server owner aware of this, and if some random compiled DirectX GUI resource insists in needing access to functions like `os.getenv`, then it's easy to conclude that something is wrong with it.

I have no complaints of some alerts or ACL lists. It is good, really. But I hate when superfluous security on the server **restrict me**. I think `os.getenv` must be in MTA server. If you want to mark this function as dangerous and implement it with alerts of security - let it be so, this feature is useful.

LosFaul:
in general I think unsafe Lua functions should be enabled/disabeled via meta.xml **for server side**
(default is always disabled)
to have it per resource rather than for whole server

botder:
Enable `os.getenv` but provide a custom C function to check if the executing resource has the ACL permission to do so and default to `false` if not given.

Gallardo994:
> Those functions solve a different problem: allowing the server administrator to set up configuration values for the resources. Environment variables are not meant to store resource configuration; instead, they represent more generic data about the OS environment that may be useful to processes as a whole.

Well, that's unless you're running an immutable Docker image and want to use different configurations for each instance, without any need to pass volumes or store any local data (totally stateless MTA server instance). This is totally normal for containers to have ENVs as configurations. 

But yeah ACL would be required for other usecases, e.g. if you use third-party resources which you may not have sources of, or if ENVs have some sensitive data required only for a selected resource (read: passwords)

patrikjuvonen:
@pentaflops Are you willing to work on the os.getenv + ACL implementation or should this PR be closed and a new issue be created?

PR #1944, opened 2020-12-24
Store bindable key states to fix getKeyState

### LINKED ISSUE #1940: getKeyState not working for gamepad buttons
**Describe the bug**

getKeyState does not work for gamepad buttons joy1, joy2, joy3, ...
it always returns false

**To reproduce**

start runcode
use debugscript 3
crun setTimer(function() for i = 1, 10 do print("joy"..i, getKeyState("joy"..i)) end end, 5000, 1)
if you press gamepad buttons or not they always return false

**Expected behaviour**

getKeyState should work for gamepad buttons too

**Additional context**

gamepad buttons joy1, joy2, joy3, ...
should be listed in wiki under key names too
https://wiki.multitheftauto.com/wiki/Key_names

botder:
Fixes issue #1940 **getKeyState not working for gamepad buttons**

Key state is not stored if any other GUI element or similiar has focus (like main menu, chatbox).

This pull request also resets key state on window-focus loss (ALT-TAB or minimized) - for example, if you hold **mouse1** then you will also receive the **onClientKey** event for key release on ALT-TAB.
This *might* break clientside resource scripts.

### DIFF
--- a/Client/core/CKeyBinds.cpp
+++ b/Client/core/CKeyBinds.cpp
@@ -255,8 +255,24 @@ const SDefaultCommandBind g_dcbDefaultCommands[] = {{"g", true, "enter_passenger
 
                                                     {"", false, NULL, NULL}};
 
+static bool bindableKeyStates[std::size(g_bkKeys)];
+
 // HACK: our current shift key states
 bool bPreLeftShift = false, bPreRightShift = false;
+
+enum eBindableKeys
+{
+    BK_MOUSE_WHEEL_UP = 5,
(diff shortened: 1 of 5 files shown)


thisdp:
new issue: getKeyState("mouse1") keeps false when gui-edit is focused

thisdp:
And will this be normal that getKeyState("mouse1") always return false when gui-edit is focused?

botder:
> And will this be normal that getKeyState("mouse1") always return false when gui-edit is focused?

I am going to investigate this issue

PR #171, opened 2017-11-19
Transfer box customization

CrosRoad95:
This event allow to overwrite currently transfer box into your custom.
Event: `onTransferBoxProgressChange` arguments: int downloaded, int total ( both in bytes, total == 1024 == 1KB )
functions:
1. `bool setTransferBoxEnabled( bool enabled )` show/hide default transfer box, by default: true
2. `bool isTransferBoxEnabled( )` return true if default transfer box is enabled, false otherwise.

example script that show current progress of downloading on chat, default transferbox doesn't appear
```lua
setTransferBoxEnabled(false)
outputChatBox(inspect({"isTransferBoxEnabled",isTransferBoxEnabled()})) -- false
addEventHandler ( "onTransferBoxProgressChange", localPlayer, function( downloaded, total )
  outputChatBox(inspect({"downloading",downloaded,total}))
end )
```
Then when you restart resource and new resources start downloading, script will show:
```
restart: Resource restarting...
Press 'o' to open your AC panel
{ "downloading", 0, 30124 }
{ "downloading", 7915, 30124 }
{ "downloading", 8502, 30124 }
{ "downloading", 17474, 30124 }
{ "downloading", 20157, 30124 }
{ "downloading", 28707, 30124 }
{ "downloading", 30124, 30124 }
```
tested on acpanel above.

Edit.
Added option to give/take server option to control transfer box

image

PR #1955, opened 2020-12-29
Add transfer box customization

botder:
# Transfer box customization
This work is based on the work by @CrosRoad95 in pull request #171 

grafik
You are able to hide and show the transfer box on clients.
You can hide the transfer box from server-side before the client starts resources to not show the download window at all.
You are supposed to use a resource with higher download priority to show a custom transfer box.
You will only receive events for the resource download transfer box.
The map download transfer box does not emit these events and **can not** be hidden with the scripting functions.
Players can force the download window to always appear regardless of the script provided setting through settings.

## New server functions
These functions are global. That means they apply for every player.
```lua
bool setTransferBoxVisible(bool visible)
bool isTransferBoxVisible() -- returns false only if the server wants the transfer box hidden for players
```

## New client functions
```lua
bool setTransferBoxVisible(bool visible)
bool isTransferBoxVisible() -- returns true if transfer box should be visible (if a download is running), it does not represent the value set by setTransferBoxVisible
bool isTransferBoxAlwaysVisible() -- returns true if the user enabled the main menu setting to always display it (cvar)
```

## New client events
```lua
onClientResourceFileDownload(resource-pointer resource, string fileName, number fileSize, string state)
-- state can be "queued" or "finished" or "failed"

onClientTransferBoxProgressChange(number downloadedBytes, number downloadTotalBytes)
onClientTransferBoxVisibilityChange(bool isVisible)
```

### DIFF
--- a/Client/core/CSettings.cpp
+++ b/Client/core/CSettings.cpp
@@ -381,6 +381,11 @@ void CSettings::CreateGUI()
     m_pCheckBoxAllowExternalSounds->GetPosition(vecTemp, false);
     m_pCheckBoxAllowExternalSounds->AutoSize(NULL, 20.0f);
 
+    m_pCheckBoxAlwaysShowTransferBox = reinterpret_cast<CGUICheckBox*>(pManager->CreateCheckBox(pTabMultiplayer, _("Always show download window"), false));
+    m_pCheckBoxAlwaysShowTransferBox->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 20.0f));
+    m_pCheckBoxAlwaysShowTransferBox->GetPosition(vecTemp, false);
+    m_pCheckBoxAlwaysShowTransferBox->AutoSize(NULL, 20.0f);
+
     m_pCheckBoxCustomizedSAFiles = reinterpret_cast<CGUICheckBox*>(pManager->CreateCheckBox(pTabMultiplayer, _("Use customized GTA:SA files"), true));
     m_pCheckBoxCustomizedSAFiles->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 20.0f));
     m_pCheckBoxCustomizedSAFiles->GetPosition(vecTemp, false);
@@ -1219,6 +1224,7 @@ void CSettings::CreateGUI()
     m_pCheckBoxVolumetricShadows->SetClickHandler(GUI_CALLBACK(&CSettings::OnVolumetricShadowsClick, this));
     m_pCheckBoxAllowScreenUpload->SetClickHandler(GUI_CALLBACK(&CSettings::OnAllowScreenUploadClick, this));
     m_pCheckBoxAllowExternalSounds->SetClickHandler(GUI_CALLBACK(&CSettings::OnAllowExternalSoundsClick, this));
+    m_pCheckBoxAlwaysShowTransferBox->SetClickHandler(GUI_CALLBACK(&CSettings::OnAlwaysShowTransferBoxClick, this));
     m_pCheckBoxCustomizedSAFiles->SetClickHandler(GUI_CALLBACK(&CSettings::OnCustomizedSAFilesClick, this));
     m_pCheckBoxWindowed->SetClickHandler(GUI_CALLBACK(&CSettings::OnWindowedClick, this));
     m_pCheckBoxDPIAware->SetClickHandler(GUI_CALLBACK(&CSettings::OnDPIAwareClick, this));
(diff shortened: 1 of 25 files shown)


Lpsd (approved):
Good stuff. I've created a [basic resource](https://github.com/multitheftauto/mtasa-blue/files/5758633/transferbox.zip) to test (and show off) these features. 

It disables the default transfer box and replaces it with a basic dx progress bar, as well as the file name and progress percentage displayed above it.

lgtm :shipit:

patrikjuvonen (approved):
Did light testing and works well. Looks good!

Client/core/CSettings.cpp
```diff
@@ -381,6 +381,11 @@ void CSettings::CreateGUI()
     m_pCheckBoxAllowExternalSounds->GetPosition(vecTemp, false);
     m_pCheckBoxAllowExternalSounds->AutoSize(NULL, 20.0f);
 
+    m_pCheckBoxAlwaysShowTransferBox = reinterpret_cast<CGUICheckBox*>(pManager->CreateCheckBox(pTabMultiplayer, _("Always show download window"), false));
+    m_pCheckBoxAlwaysShowTransferBox->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 20.0f));
+    m_pCheckBoxAlwaysShowTransferBox->GetPosition(vecTemp, false);
+    m_pCheckBoxAlwaysShowTransferBox->AutoSize(NULL, 20.0f);
```
sbx320: `nullptr`

Client/core/CSettings.cpp
```diff
@@ -381,6 +381,11 @@ void CSettings::CreateGUI()
     m_pCheckBoxAllowExternalSounds->GetPosition(vecTemp, false);
     m_pCheckBoxAllowExternalSounds->AutoSize(NULL, 20.0f);
 
+    m_pCheckBoxAlwaysShowTransferBox = reinterpret_cast<CGUICheckBox*>(pManager->CreateCheckBox(pTabMultiplayer, _("Always show download window"), false));
```
sbx320, line 384: Not sure if the description is clear enough for most players

Client/mods/deathmatch/logic/CResourceFileDownloadManager.cpp
```diff
@@ -137,9 +148,16 @@ void CResourceFileDownloadManager::DoPulse()
     }
 
     // Update progress box
-    GetTransferBox()->SetInfo(uiDownloadSizeTotal);
+    GetTransferBox()->SetDownloadProgress(uiDownloadSizeTotal);
     GetTransferBox()->DoPulse();
 
+    // Call Lua event 'onClientTransferBoxProgressChange'
+    CLuaArguments arguments;
+    arguments.PushNumber(uiDownloadSizeTotal);
```
sbx320: Might be worth renaming `uiDownloadSizeTotal` to `uiDownloadedSizeTotal` to make this a bit more clear

Client/mods/deathmatch/logic/CTransferBox.h
```diff
@@ -1,59 +1,73 @@
...
- *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *  Multi Theft Auto is available from https://multitheftauto.com/
  *
  *****************************************************************************/
 
 #pragma once
 
-#define TRANSFERBOX_FRAMES  10
-#define TRANSFERBOX_DELAY   50
+// #include "CClientCommon.h"
```
sbx320: Why are those commented out? Might as well remove them if not used

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -0,0 +1,33 @@
...
+    constexpr static const std::pair<const char*, lua_CFunction> functions[]{
+        {"setTransferBoxVisible", ArgumentParser<SetTransferBoxVisible>},
+        {"isTransferBoxVisible", ArgumentParser<IsTransferBoxVisible>},
+    };
+
+    for (const auto& [name, func] : functions)
+        CLuaCFunctions::AddFunction(name, func);
+}
+
+bool CLuaGenericDefs::SetTransferBoxVisible(bool visible)
```
sbx320: Do we need these extra functions? We could just add the `CStaticFunctionDefinitions` instead.

sbx320 (changes requested):

Client/core/CSettings.cpp
botder: I assumed that even less players will understand what a "transfer box" is. I am open for suggestions for this setting.

botder:
Individual resource file download progress can't be displayed because the network module returns `false` in a `GetDownloadStatus` lookup through the HTTP server interface.

botder:
The remaining suggested-change about the settings window naming: `Always show download window` is the last thing preventing the merge.

Change it? And if yes, to what?

PR #1956, opened 2020-12-29
Move version.h and build_overrides.h under Shared folder

patrikjuvonen:
# Only ccw can merge this

Work done by @qaisjp, moved build_overrides.h under Shared as well as suggested by @ccw808.

As they have minimal differences this makes maintaining easier to have the contents shared between client and server.

Only @ccw808 can merge because this requires build server modifications to support the new build_overrides.h filepath.


Shared/sdk/version.h
```diff
@@ -1,10 +1,13 @@
...
-//
-// Server version file
+// Shared/version.h
 //
 #pragma once
 
+//
+// To compile a client:
+//      1. set MTASA_VERSION_TYPE to VERSION_TYPE_CUSTOM
+//      2. Use netc.dll from the latest unstable build (nightly.mtasa.com)
```
Dutchman101, line 9: This comment looks wrong.. the netc.dll made for custom builds (from install-data.bat), which is located at [https://mirror.mtasa.com/bdata/netc.dll](https://mirror.mtasa.com/bdata/netc.dll), should be used.

Dutchman101 (changes requested):

Shared/sdk/version.h
patrikjuvonen: That's the current comment on master (https://github.com/multitheftauto/mtasa-blue/blob/master/Client/version.h#L10)

What would be an accurate fix?

```suggestion
//      2. Run install-data for netc.dll
```

patrikjuvonen: @Dutchman101 Any comment on this?

patrikjuvonen: @Dutchman101 Any comment on this?

Dutchman101: To correct the existing comment(s) that are incorrect.. either in this PR or a new one. Like just refer to the existing build instructions or install-data without specific mention of netc. I would suggest to just refer to the build instructions readme.

PR #1964, opened 2020-12-30
Multithreading

### LINKED ISSUE #918: Add multi-threading support to Lua

While Lua is fast we are still bound to limitations that can make it hard to do some expensive calculations for things like path finding or dealing with running a SCM virtual machine (Sphene). It would be much nicer if this code could be moved into a separate thread and therefore prevent freezes.

Lua itself doesn't support multi-threading but is has been shown [here](https://love2d.org/wiki/love.thread) that there is definitely a way to do so.

I'd like for new functions to be exposed that allow for a thread to be created and for data to be transmitted between them. In line with the example of LÖVE I propose the following functions to be added at least:

- createThread(filePath) ( **OOP:** Thread(filePath) )
Creates a new thread and returns the newly created thread.
_Parameters:_ 
  - file: Path to a file to be ran in the thread or a string containing Lua code.
- killThread(thread) ( **OOP:** thread:kill() )
Terminates a thread forcefully, ending its execution.
_Parameters:_ 
  - thread: The thread to be terminated.
- startThread(thread) ( **OOP:** thread:start() )
Starts the thread.
_Parameters:_ 
  - thread: The thread to be started.
- isThreadRunning(thread) ( **OOP:** thread:isRunning() )
Returns true if the thread is running and false if not.
_Parameters:_ 
  - thread: The thread to be checked.
- sendThreadMessage(thread, value) ( **OOP:** thread:sendMessage(value) )
Sends a message to the thread.
_Parameters:_ 
  - thread: The thread to send the message to.
  - value: The value to send to the thread.
- receiveThreadMessage() ( **OOP:** Thread.receiveMessage() )
Receives a message from the message queue for the thread in which this is executed. Although this could also be a event (perhaps both?). Returns the thread that has sends the message and the value of that message.
- getThreadMessageCount() ( **OOP:** Thread.getMessageCount() )
Returns the amount of messages in the message queue for the thread in which this is executed.

Please note that this is an initial thought and that this API should be heavily discussed and adjusted where possible.

Using coroutines but you can only run one at once so it can slow down execution even more if improperly used. Coroutines work better for things like task scheduling.

CrosRoad95:
Option to execute custom time consuming tasks without blocking main thread.

current functions list, its syntax and purpose:

- [x] `createThread(string code)` -- Creates a new thread and start executing code which in that thread
- [x] `getThreadState(thread)` -- Return current state of thread
- [ ] `threadCall(thread, functionName, table of arguments)` -- calling function inside thread.
- [ ] `killThread(thread, terminate)` -- sends signal to kill thread, wait 1 second ( if second parametr is false ) and then terminates a thread forcefully, ending its execution.
- [ ] `getThreadFailureReason(thread)` -- In case of error thrown inside
- [x] `threadGetResult(thread)` -- Returns value thread returned at the end ( if something have been returned and code reached end )
- [ ] `getThreads()` -- return all threads
- [ ] `setThreadShutdownHandler(thread, callback)` -- when "exit" get called inside thread

Function availiable only in thread:
- [ ] `sleep(ms)` -- 
- [ ] `exit(...)` -- stopping thread, send data to shutdown handler
- [ ] `sendMessage(...)` -- sends message that you can catch using `setThreadCustomHandler` in main thread

Maybe:
1. `bool isMainThread()` -- return true if called in main thread, otherwise false.
2. `setThreadOutputHandler(thread, callback)` -- to set where are outputs from print, outputChatBox ect. will redirected,
3. `setThreadErrorHandler(thread, callback)` -- to set where to output errors
4. `setThreadCustomHandler(thread, callback)` -- used for communication from thread, to main thread ( i don't have idea how to name it )
5.  `threadLoadstring(thread, code)` -- to load extra code into thread ( if while creating you explicit allow it in options )
6.  `getThreadsLimits()` -- Clientside, returns information about user settings, how many threads can be created ect.
7.  `getThreadtimings(thread)` -- https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreadtimes
8.  `startThread(thread)` -- Starts the thread.
9. `pauseThread(thread)` -- Pause the thread. -just calling "sleep" function all time until "start" get called

Maybe in future, option to create thread-safe variable, name based on std::atomic
1. `atomic createAtomic( ... )`
2. `values getAtomicValue( atomic )`
3. `bool setAtomicValue( ... )`

Other:
1. "threadCall" wil have option to get return of that function, but due problems with argument parser right now you can't do that.
2. while creating new thread, you should have options to a bit more customize it. I'm consider that useful functions would be to set what will loaded by default, what functions, if possible, option to change "speed" of that thread by adding sleep(1) every n instructions,
3. Clientside threads are a bit dangerous eg. malicious code that will run infinity loop to consume all cpu time. Client should have option to somehow limit, slowdown custom threads
4. Only thread-safe functions should be availiable inside 
5. Errors, how to handle them? i wrote proposal function above
6. Should we limit threads serverside, add mtaserver.conf option to limit how many of them can running at the same time? or you can create as many threads as you want?
7.  ACL, by default all functions should be disallowed. When thread get created, it should inherit permissions from resource that created this thread

Current example:
```lua
code = [[
function mineBitcoins()
    for i=1,500000000 do

    end
    print("1 bitcoined nined")
end
function add(a,b)
    print("add",a,b)
    return a+b
end
print("hello, i'm thread")
return 1,2,3
]]

    thread = createThread(code)
    local state = ""
    i = 0
    function update()
        i = i + 1
        local newState = getThreadState(thread)
        if(state ~= newState)then
            state = newState
            print("state =", state)
        end
        if(state == "loaded")then
            print("result",getThreadResult(thread))
            threadCall(thread, "add", {i, 10})
            threadCall(thread, "mineBitcoins")
        end
    end
    update()
    setTimer(update,0,0)
end)
```
image


Einheit-101:
This is cool.

prnxdev:
OOP support pls

Deihim007:
Is it possible to change `createThread(string code)` to `createThread(function)` ?

CrosRoad95:
> Is it possible to change `createThread(string code)` to `createThread(function)` ?

yes

gta191977649:
Is it available on serverside or clientside?

Lpsd:
Both, refer to the code

PR #1978, opened 2021-01-03
Fix #741: Add new CServerRPCControlPacket to toggle RPCs from server

### LINKED ISSUE #741: Disable some server events triggered from clientside
performance

make client side function that can disable events like `onElementClicked`, `onPlayerClick`, `onPlayerTarget`, `onPlayerContact` , client just stop triggering this events.

/

**Additional context**
If you don't use this events, why clients must send them to server?

patrikjuvonen:
## Summary
- Allows server to tell client to stop sending RPCs (currently only `CURSOR_EVENT` is toggleable, as other RPCs are required for sync)
- Basically if your server does not have onPlayerClick or onElementClicked events, your server will not be sent unnecessary RPC calls for these events (reduces packet count)
- This sort of optimization probably mostly benefits large servers with lots of UI interactions
- Moved RPC enums under shared SDK because RPC enums were the same on client and server
- Added a new `CServerRPCControlPacket` packet, which is sent from the server when necessary: on join, as well as when a previously nonexistent onPlayerClick/onElementClicked event is added, or the only remaining event is removed (so no unnecessary packets are sent if you add multiple of these events)
- Closes #741 once merged

## Requesting for feedback
- Naming correctness/is it clear enough?
- Is code good or not, what can be improved and am I missing something? Is there some new compiler magic I could use to make less code?


Pirulax:
> Is there some new compiler magic I could use to make less code?

Probably not compiler magic, but rather `std` magic.

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -824,6 +837,8 @@ class CClientGame
     SString                 m_strACInfo;
     std::map<uint, uint>    m_SentMessageIds;
 
+    std::vector<eServerRPCFunctions> m_disabledServerRPCFunctions;
+
```
Pirulax: This seems pretty harsh to me.
`eServerRPCFunctions` has only 10 or so enums, and even if it has 1000, I'd just create a static array of bools for it, and there you go.
So:
```cpp
#include <array>

std::array<bool, (size_t)eServerRPCFunctions::NUM> m_disabledServerRPCFunctions; // standard guarantees default initalization to false
```
Now, when you want to disable a function just do:
```cpp
m_disabledServerRPCFunctions[(size_t)rpcfn] = true;
```

This is lightyears faster, and compact, and all that.

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -5314,6 +5318,25 @@ void CPacketHandler::Packet_ServerInfoSync(NetBitStreamInterface& bitStream)
...
+        for (unsigned short us = 0; us < usNumFunctions; us++)
+        {
+            SString strServerRPCFunction;
+            bool    bDisabled;
+            if (bitStream.ReadString(strServerRPCFunction) && bitStream.ReadBit(bDisabled))
+            {
+                eServerRPCFunctions eServerRPCFunction;
+                StringToEnum(strServerRPCFunction, eServerRPCFunction);
+                g_pClientGame->SetServerRPCFunctionDisabled(eServerRPCFunction, bDisabled);
+            }
```
Pirulax: RPC packet IDs probably will never be removed, so sending a simple numeric ID should be fine here.

Server/mods/deathmatch/logic/CMapEventManager.cpp
```diff
@@ -341,18 +342,86 @@ bool CMapEventManager::HandleExists(CLuaMain* pLuaMain, const char* szName, cons
...
-    {
-        if (pEvent->IsHigherPriorityThan(iter->second))
+        if (pMapEvent->IsHigherPriorityThan(iter->second))
             break;
-    }
+
     // Do insert
-    m_EventsMap.insert(iter, std::pair<SString, CMapEvent*>(pEvent->GetName(), pEvent));
+    m_EventsMap.insert(iter, std::pair<SString, CMapEvent*>(pMapEvent->GetName(), pMapEvent));
 }
```
Pirulax: Please try to make it not so nested. Do early exists, and stuff.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1245,6 +1245,18 @@ void CGame::JoinPlayer(CPlayer& Player)
     if (Player.CanBitStream(eBitStreamVersion::Discord_InitialImplementation))
         Player.Send(CServerInfoSyncPacket(SERVER_INFO_FLAG_ALL));
 
+    // Control server RPC functions
+    std::map<eServerRPCFunctions, bool> toggleMap;
```
Pirulax: A map of bools? Never.
Just do the `std::array<bool>` trick here.

Server/mods/deathmatch/logic/CMapEventManager.cpp
```diff
@@ -341,18 +342,86 @@ bool CMapEventManager::HandleExists(CLuaMain* pLuaMain, const char* szName, cons
...
+                {
+                    std::map<eServerRPCFunctions, bool> map;
+                    map.insert(std::make_pair(pEvent->eServerRPCFunction, false));
+                    CServerRPCControlPacket Packet(map);
+                    g_pGame->GetPlayerManager()->BroadcastOnlyJoined(Packet);
+                }
+            }
+        }
+    }
+
```
Zangomangu, line 387: These look identical besides the bool in map.insert. Should this be put into a separate "toggle" function instead that takes a boolean? And you can return in there instead of if's

Zangomangu (approved):
Looks good and works as expected, I've tested back and forth on old/new version.

Pirulax:
Forgot to submit my review.. Again...

patrikjuvonen:
Thanks for the reviews. Changes made.

PR #1875, opened 2020-11-27
Update Unifont from 5.1 to 13.0.04

Dutchman101:
This PR updates Unifont from version 5.1 (2008) to 13.0.04 (Nov 2020)

There's no reason not to update, it's tested and works fine, and in fact looks a little better overall. https://github.com/multitheftauto/mtasa-blue/issues/1864#issuecomment-731885264 inspired this PR.

PR #1980, opened 2021-01-04
Update Unifont from 5.1 to 13.0.04

patrikjuvonen:
## Summary
- Font contains various additions, changes and fixes
- Adds a new build action `install_unifont.lua` which is run via `win-create-projects.bat`
- Uses our [unifont repository](https://github.com/multitheftauto/unifont)
- Allows client scripts to use unifont via drawing functions
- Follow up PR for #1875
- Changelogs
  - http://unifoundry.com/unifont/index.html


sbx320:
Wouldn't it be easier to add this to `install_data` rather than to add another buildaction?

patrikjuvonen:
> Wouldn't it be easier to add this to `install_data` rather than to add another buildaction?

I thought about it, but I'm sort of conflicted. Currently install_data mostly affects netc pretty much so adding Unifont stands out as an odd one, but I don't know. I could move it.

Dutchman101 (approved):
Making it available to scripts is pretty nice as well!

utils/buildactions/install_unifont.lua
```diff
@@ -0,0 +1,36 @@
...
+			return
+		end
+
+		-- Download Unifont
+		print("Downloading Unifont...")
+		if not http.download_print_errors(UNIFONT_BASEURL..UNIFONT_TAG.."/"..UNIFONT_FILENAME, archive_path) then
+			return
+		end
+
+		print("Unifont updated.")
```
botder, line 40: Missing SHA256 hash verification for the downloaded unifont file here

botder (changes requested):

utils/buildactions/install_unifont.lua
patrikjuvonen: Thanks, added.

botder (approved):
I have tested the font with a regular GUI label (height: 15) and here are the results:
Font size 9 and aliasing-enabled are the defaults for unifont.

grafik

PR #1981, opened 2021-01-04
Fix #1973: Nested tuples not working in argument parser return values

### LINKED ISSUE #1973: New parser, return `std::vector<std::tuple<...>>` issue
**Describe the bug**
"attempt to index a number value" error gets thrown when trying to use function 

**To reproduce**
make for example std::vector<std::tuple<float, float, floatt>> as a return type, it will compile but will throw errors as above.

**Expected behaviour**
It should just return a table like `{ {1,2,3}, {1,2,3}, ... }`

**Screenshots**
image

**Version**
@CrosRoad95 please add this

**Additional context**
Possible solutions:
1. Make special case, when tuple is in vector - treat tuple as a table
2. Forbid vector of tuples at compile time. Add information to use vector in vector instead.

sbx320:
Rather than directly handling tuples in `Push` for multiple return values, I changed it to act on return values from `CLuaFunctionParser` only. Tuples used for `Push` are now pushed as a simple table. This should also make `lua::Push` more usable for other purposes.

### DIFF
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -642,13 +642,29 @@ struct CLuaFunctionParser<ErrorOnFailure, ReturnOnFailure, Func> : CLuaFunctionP
             }
             else
             {
-                return lua::Push(L, Func(std::forward<Params>(ps)...));
+                return PushResult(L, Func(std::forward<Params>(ps)...));
             }
         }
         else
         {
             return Call(L, ps..., Pop<typename nth_element_impl<sizeof...(Params), Args...>::type>(L, iIndex));
         }
+    }
+
+    // Tuples can be used to return multiple results
+    template <typename... Ts>
+    inline int PushResult(lua_State* L, const std::tuple<Ts...>& tuple)
+    {
+        // Call Push on each element of the tuple
+        std::apply([L](const auto&... value) { (lua::Push(L, value), ...); }, tuple);
+        return sizeof...(Ts);
+    }
+
+    // If `T` is not a tuple, defer to Push to push the value onto the stack
+    template <typename T>
+    inline int PushResult(lua_State* L, const T& value)
+    {
+        return lua::Push(L, value);
     }
 
     inline int operator()(lua_State* L, CScriptDebugging* pScriptDebugging)
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -664,7 +664,8 @@ struct CLuaFunctionParser<ErrorOnFailure, ReturnOnFailure, Func> : CLuaFunctionP
     template <typename T>
     inline int PushResult(lua_State* L, const T& value)
     {
-        return lua::Push(L, value);
+        lua::Push(L, value);
+        return 1;
     }
 
     inline int operator()(lua_State* L, CScriptDebugging* pScriptDebugging)
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -18,6 +18,21 @@ class CLuaArgument;
 #include "lua/CLuaStackChecker.h"
 #include "lua/LuaBasic.h"
 
+// Wrapper around std::tuple to indicate that multiple values should be pushed to
+// the Lua stack
+template<typename... Ts>
+struct LuaMultiReturn
+{
+    // Note: We use a separate template for the constructor arguments
+    // to allow type conversions. For example: return { "hello", 42 };
+    // is a valid statement to construct a LuaMuliReturn<std::string, int>
+    template <typename... Args>
+    LuaMultiReturn(Args... args) : values{args...} {}
+
+    std::tuple<Ts...> values;
+};
+
+
 struct CLuaFunctionParserBase
 {
     // iIndex is passed around by reference
@@ -653,10 +668,10 @@ struct CLuaFunctionParser<ErrorOnFailure, ReturnOnFailure, Func> : CLuaFunctionP
 
     // Tuples can be used to return multiple results
     template <typename... Ts>
-    inline int PushResult(lua_State* L, const std::tuple<Ts...>& tuple)
+    inline int PushResult(lua_State* L, const LuaMultiReturn<Ts...>& result)
     {
         // Call Push on each element of the tuple
-        std::apply([L](const auto&... value) { (lua::Push(L, value), ...); }, tuple);
+        std::apply([L](const auto&... value) { (lua::Push(L, value), ...); }, result.values);
         return sizeof...(Ts);
     }
 
--- a/Shared/mods/deathmatch/logic/lua/LuaBasic.h
+++ b/Shared/mods/deathmatch/logic/lua/LuaBasic.h
@@ -13,7 +13,7 @@
 
 /*
     Basic Lua operations:
-        int Push(L, T value)
+        void Push(L, T value)
         T PopPrimitive(L, std::size_t stackIndex)
 */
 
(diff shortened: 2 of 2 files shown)


Pirulax:
I have already fixed this issue, but havent opened a PR.
What I do is (in case of vector<tuple<>, map<K, tuple<>>, array<tuple<>):
- Check the return value of Push<tuple>, if its more than 1 (that is, multiple values have been pushed onto the stack), pack values on the stack into an array
Altough, my approach isn't 100% consistent, as tuple<optional<>, T> (depending whenever the optional has a value) might become stay as a value, instead of a table with 1 value.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -651,6 +651,22 @@ struct CLuaFunctionParser<ErrorOnFailure, ReturnOnFailure, Func> : CLuaFunctionP
         }
     }
 
+    // Tuples can be used to return multiple results
+    template <typename... Ts>
+    inline int PushResult(lua_State* L, const std::tuple<Ts...>& tuple)
+    {
+        // Call Push on each element of the tuple
+        std::apply([L](const auto&... value) { (lua::Push(L, value), ...); }, tuple);
+        return sizeof...(Ts);
```
Pirulax, line 675: This won't work if there's an `optional` with no value.
Instead you could do:
```cpp
return std::apply([L](const auto&... value) { return (Push(L, value) + ...); }, tuple);
```

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -228,12 +228,25 @@ namespace lua
         return 1;
     }
 
-    // Tuples can be used to return multiple results
+    // Tuples can be used to build fixed-size tables
+    // e.g. `std::tuple<float, int, bool>` will be pushed as { float, int, bool }
```
Pirulax, line 209: I'm quite confused?
Now a tuple can be either a `fixed size table` and a `way to push multiple results`?

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -228,12 +228,25 @@ namespace lua
...
-        return sizeof...(Ts);
+        lua_createtable(L, sizeof...(Ts), 0);
+        
+        std::apply([L](const auto&... values) {
+            int  key = 1;
+            auto PushTable = [](lua_State* L, int& key, const auto& value)
+            {
+                Push(L, key++);
+                Push(L, value);
+                lua_settable(L, -3);
```
Pirulax: `lua_rawsetfield` is more efficient as it doesn't try to call the tag method, also it doesn't need the key on the stack.
```cpp
Push(L, value);
lua_rawsetfield(L, -3, key++);
```

Pirulax:
Also, I'm not sure whenever this fixes the actual underlaying problem when you try to do a `Push<vector<T>>`  where T results in multiple values pushed onto the stack.

Take a look at [my solution](https://github.com/Pirulax/mtasa-blue/tree/fix/luabasic-vector-map).

sbx320:
With this change 

> where T results in multiple values pushed onto the stack.

never occurs. It's either 1 (for most values) or 0 (for an empty optional)

Shared/mods/deathmatch/logic/lua/LuaBasic.h
sbx320: If you use a `std::tuple` as a return value for a scripting function for the arg parser, it'll push multiple values

If you use a `std::tuple` as a parameter to `lua::Push` it'll push a table of values

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: That makes me wonder: Is there even a valid case for empty optionals being pushed? If not we could just drop it entirely and make Push return `void`

Shared/mods/deathmatch/logic/lua/LuaBasic.h
sbx320: Where did you get `lua_rawsetfield` from? That's not a function in Lua 5.1

Shared/mods/deathmatch/logic/lua/LuaBasic.h
sbx320: This is also intended to make `lua::Push` behave more consistently for non-argparser usages. For example we could use it to set up globals.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
Pirulax: Ah, actually I'm sorry. An optional (even if empty) results in a `nil` pushed onto the stack, for whatever reason.

sbx320: Right... the one I saw was `CLuaArgument`. 
```
inline int Push(lua_State* L, const CLuaArgument& arg)
    {
        if (arg.GetType() == LUA_TNONE)
            return 0;
```
Not sure what to do with that...

Shared/mods/deathmatch/logic/lua/LuaBasic.h
Pirulax: Okay, I see. But that's kinda inconsistent, isn't it?
Perhaps we could add a `PushTableValue`, which would make sure if T pushes multiple values onto the stack, it would still become one value (by packing the values into a table, just like I did on my branch)

Shared/mods/deathmatch/logic/lua/LuaBasic.h
Pirulax: I'm sorry. Its [lua_rawseti](https://pgl.yoyo.org/luai/i/lua_rawseti).

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
Pirulax: Ah, yes.. Well, this might break something if it's inside a table, or part of a tuple.
But you really shouldn't be using a tuple in that case anyways..
As of now `CLuaArgument` is only used for element data iirc, so its not a big deal.

Shared/mods/deathmatch/logic/lua/LuaBasic.h
sbx320: Well `lua::Push` would now be consistent (apart from the CLuaArgument edge case above...). 

The arg parser would gain some inconsistency, but we'd otherwise not have a good option to return multiple values.

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
sbx320: It'll likely break all kinds of stuff if you attempt to push a `LUA_TNONE` anyway. Maybe just disallowing that and throwing on any such attempt might be a better solution. Would also clean up `Push` a lot.

sbx320: Added that now: `lua::Push` always pushes exactly one value. Pushing `LUA_TNONE` via  `CLuaArgument` will throw.

Shared/mods/deathmatch/logic/lua/LuaBasic.h
Pirulax: Well, what about disallowing tuples inside tables, and instead use `std::array<std::variant<>>` for fixed size tables?

sbx320: That gets fairly weird quickly, since it obscures the intended return value (which has a static type) behind layers of `variant`. 

We could also add some special multi-return tuple proxy struct, which just acts as a special tuple gadget. That way you could either return a `std::tuple` (and get a table of values) or a `CLuaMultiReturn` (and get multiple values).

sbx320:
Finally got around to finishing this up. Ultimately we now have the following changes:

- `lua::Push` now always pushes exactly one value to the stack and can be used everywhere.
- Returning tuples from a C++ function maps to returning a table to Lua (e.g. `std::tuple<int, std::string>{1, "hello"}` -> `{1, "hello"}`
- If a C++ function needs to return multiple values, `LuaMultiReturn<Ts...>` is used. (e.g. `LuaMultiReturn<int, std::string>{1, "hello"}` -> `1, "hello"`. `lua::Push(LuaMultiReturn<T>`

/cc @Pirulax

Pirulax:
I'm lazy to properly re-review it, hope you considered all things, and it'll work just fine.

Pirulax (approved):

PR #1986, opened 2021-01-05
Fix #521: Network trouble if falling into water while attempting to enter a vehicle

### LINKED ISSUE #521: Network trouble if falling into water while attempting to enter a vehicle
**Describe the bug**
If you fall in water whilst attempting to enter a vehicle, when you reach land again you'll get a short Network Trouble message. This is followed by a warp back to your original position when you first pressed the enter key.

**To Reproduce**
- put a vehicle near a cliff over some water
- attempt to enter it (make sure the auto route will let you fall down in the water)
- swim to the closest land you can stand on

**Expected behaviour**
Not to get network trouble.

**Additional context**
From https://bugs.mtasa.com/view.php?id=7951

Zangomangu:
Fixes #521, abort entering if in water

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -6799,8 +6799,8 @@ void CClientPed::UpdateVehicleInOut()
         else if (m_bIsGettingIntoVehicle)
         {
             // If we aren't working on entering the car (he's either finished or cancelled)
-            // Or we are dead (fix for #908)
-            if (!IsEnteringVehicle() || IsDead())
+            // Or we are dead (fix for #908) or we are in water (fix for #521)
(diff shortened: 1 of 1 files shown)


botder:
I have investigated the issue and the player doesn't actually stop the task to enter the vehicle when fallen into water.
This patch might cause another issue.

Zangomangu:
The problem is GTA doesnt interrupt the entering task when the player is in water, so this fix aborts it manually like the task is cancelled. What issue are you thinking this could cause?

botder:
Nevermind, I missed the task cancellation in `RemoveFromVehicle`

botder (approved):

botder:
Thanks for the pull request! I have tested it briefly and it works as intended.

PR #1989, opened 2021-01-07
Revert "Update Unifont from 5.1 to 13.0.04 (#1980)"

patrikjuvonen:
People are reporting issues with the latest Unifont update (7528ff5f90ba421751c4ed4df2234c4198c29bc4) due to a missing font file via client updater.

I assume build server is not pulling the font file.

unknown

https://discord.com/channels/278474088903606273/675939577281773568/796832551489503283


sbx320:
Notes from Discord
```
[2:25 PM] sbx320: build server doesn't include fonts in autoupdates as far as I can see
[2:31 PM] sbx320: This is actually pretty annoying
[2:32 PM] sbx320: We could add unifont to the autoupdater, but that wastes a lot of traffic as it'd always be part of the update
[2:32 PM] sbx320: But if we don't add it, we can't really update it at all
[2:33 PM] sbx320: (if we only added it into the installer and kept the filename, people would lose new characters)
```
Not quite sure if and how we'd properly upgrade it, so reverting for now to keep nightlies going.

ccw808:
Change the file name to unifont.tff then the updater can do a hash check to optimize download size

PR #1994, opened 2021-01-10
Add proper 3D radius check to getElementsWithinRange

Pirulax:
Since our RTree (aka spatial database) is 2D, and converts spheres into boxes we need to recheck whenever the elements are within the sphere manually.


Pirulax:
Perhaps we should keep the old behavior (that is, keep checking elements within a box, instead of a sphere), and only add Z check?
It might be faster by a very little margin, but also it might offer a better backwards compatibility.
Reply with a 👍 or 👎 .

PlatinMTA:
Thinking of backwards compatiblity, seeing that this function never checked for Z values, would anyone use getElementsWithinRange() to only compare X,Y values and therefore with this update their script would lose backwards compatibility and don't function correctly?

The documentation already warned about this issue but idk. Maybe if we add an extra argument that determinates if we check for Z values or not that should be kinda fixed. It won't be full backwards compatible but it would give scripters a simple fix, it's not that this didn't happen before and I don't think that there is a lot of people that use getElementsWithinRange() to only get elements within X,Y.

What about adding an extra argument to select between spheres and boxes? Maybe thats pushing it too much.

Dutchman101:
More background information: [https://github.com/multitheftauto/mtasa-blue/pull/1915#issuecomment-747254198](https://github.com/multitheftauto/mtasa-blue/pull/1915#issuecomment-747254198) and later comments there

Pirulax:
Look. There's been a note on the wikipedia page since it was added, noting that the `z` argument will be used in the future.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -967,20 +967,22 @@ CClientEntityResult CLuaElementDefs::GetElementsWithinRange(CVector pos, float r
...
-                                            return true;
+            if (dimension.has_value() && dimension != pElement->GetDimension())
+                return true;
 
-                                        return pElement->IsBeingDeleted();
-                                    }),
-                     result.end());
+            // Check if element is within the sphere, because the spatial database is 2D
+            if ((pElement->GetPosition() - pos).LengthSquared() > radiusSq)
+                return true;
```
botder:
```suggestion
            CVector elementPos;
            pElement->GetPosition(elementPos);
            if ((elementPos - pos).LengthSquared() > radiusSq)
                return true;
```

botder (approved):

PlatinMTA:
This should remain a cube, or at least default behaviour should be the one of a cube

PR #1995, opened 2021-01-10
Enforce existing username length limit

patrikjuvonen:
At the moment our database code already limits usernames to 64 characters long, but is not enforced elsewhere in the code, causing potential buffer overflows when handling usernames. To fix this let's limit the usernames to max length.

Note that whether the limit should be changed is out-of-scope for this PR. This PR only wants to get rid of potential issues caused by overlong usernames.

Should be backwards compatible as far as I can see.

### DIFF
--- a/Server/mods/deathmatch/Config.h
+++ b/Server/mods/deathmatch/Config.h
@@ -57,6 +57,10 @@ public:
 
 #define MAX_TEAM_NAME_LENGTH 255
 
+// Min and max number of characters in usernames
+#define MIN_USERNAME_LENGTH 1
+#define MAX_USERNAME_LENGTH 64
+
 // Min number of characters in passwords
 #define MIN_PASSWORD_LENGTH 1
 
--- a/Server/mods/deathmatch/Config.h
+++ b/Server/mods/deathmatch/Config.h
@@ -57,10 +57,6 @@ public:
 
 #define MAX_TEAM_NAME_LENGTH 255
 
-// Min and max number of characters in usernames
-#define MIN_USERNAME_LENGTH 1
-#define MAX_USERNAME_LENGTH 64
-
 // Min number of characters in passwords
 #define MIN_PASSWORD_LENGTH 1
 
--- a/Server/mods/deathmatch/logic/CAccount.cpp
+++ b/Server/mods/deathmatch/logic/CAccount.cpp
@@ -20,7 +20,7 @@ CAccount::CAccount(CAccountManager* pManager, EAccountType accountType, const st
     m_bChanged = false;
     m_pManager = pManager;
     m_AccountType = accountType;
-    m_strName = strName.substr(0, MAX_USERNAME_LENGTH);
+    m_strName = strName.substr(0, CAccountManager::MAX_USERNAME_LENGTH);
     m_iUserID = iUserID;
     m_strIP = strIP;
     m_strSerial = strSerial;
@@ -46,7 +46,7 @@ void CAccount::SetName(const std::string& strName)
 {
     if (m_strName != strName)
     {
-        std::string strNewName = strName.substr(0, MAX_USERNAME_LENGTH);
+        std::string strNewName = strName.substr(0, CAccountManager::MAX_USERNAME_LENGTH);
         m_pManager->ChangingName(this, m_strName, strNewName);
         m_strName = strNewName;
         m_pManager->MarkAsChanged(this);
--- a/Server/mods/deathmatch/logic/CAccount.cpp
+++ b/Server/mods/deathmatch/logic/CAccount.cpp
@@ -20,7 +20,7 @@ CAccount::CAccount(CAccountManager* pManager, EAccountType accountType, const st
     m_bChanged = false;
     m_pManager = pManager;
     m_AccountType = accountType;
-    m_strName = strName;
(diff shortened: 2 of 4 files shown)


Server/mods/deathmatch/Config.h
```diff
@@ -57,6 +57,10 @@ class CVaPassNext
 
 #define MAX_TEAM_NAME_LENGTH 255
 
+// Min and max number of characters in usernames
+#define MIN_USERNAME_LENGTH 1
+#define MAX_USERNAME_LENGTH 64
```
botder: You should move these directly into the account manager class
```cpp
class CAccountManager
{
public:
    static inline constexpr size_t MIN_USERNAME_LENGTH = 1;
    static inline constexpr size_t MAX_USERNAME_LENGTH = 64;
    // ...
};
```

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -46,8 +46,9 @@ void CAccount::SetName(const std::string& strName)
 {
     if (m_strName != strName)
     {
-        m_pManager->ChangingName(this, m_strName, strName);
-        m_strName = strName;
+        std::string strNewName = strName.substr(0, MAX_USERNAME_LENGTH);
+        m_pManager->ChangingName(this, m_strName, strNewName);
+        m_strName = strNewName;
```
botder: You don't use the value anymore, we can move it and avoid a copy (if the compiler doesn't avoid it).
```suggestion
        m_strName = std::move(strNewName);
```

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -46,8 +46,9 @@ void CAccount::SetName(const std::string& strName)
 {
     if (m_strName != strName)
     {
-        m_pManager->ChangingName(this, m_strName, strName);
-        m_strName = strName;
+        std::string strNewName = strName.substr(0, MAX_USERNAME_LENGTH);
```
botder: We could avoid a copy here, if we change `CAccount::SetName` to return a boolean and check if the length is in our predefined boundaries.

Server/mods/deathmatch/logic/CAccountManager.h
```diff
@@ -130,7 +130,7 @@ class CAccountManager
     CAccount* GetAccountFromScriptID(uint uiScriptID);
     SString   GetActiveCaseVariation(const SString& strName);
 
-    bool LogIn(CClient* pClient, CClient* pEchoClient, const char* szAccountName, const char* szPassword);
+    bool LogIn(CClient* pClient, CClient* pEchoClient, std::string strAccountName, const char* szPassword);
```
botder:
```suggestion
    bool LogIn(CClient* pClient, CClient* pEchoClient, const std::string& strAccountName, const char* szPassword);
```

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -481,7 +481,7 @@ void CAccountManager::RemoveAll()
     DeletePointersAndClearList(m_List);
 }
 
-bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, const char* szAccountName, const char* szPassword)
+bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, std::string strAccountName, const char* szPassword)
```
botder:
```suggestion
bool CAccountManager::LogIn(CClient* pClient, CClient* pEchoClient, const std::string& strAccountName, const char* szPassword)
```

botder (changes requested):

Server/mods/deathmatch/logic/CAccount.cpp
Pirulax: std::string_view could be used instead.
```cpp
auto strTrimmedName = std::string_view{strName}.substr(0, MAX_USERNAME_LENGTH);
```

Pirulax: Actually:
```cpp
if (strName.length() > MAX_USERNAME_LENGTH)
    strName.erase(MAX_USERNAME_LENGTH);
```

patrikjuvonen: Let's not do that before 1.6. Breaks scripts that are currently (wrongly) depending on overlong usernames in their code.

patrikjuvonen:
Thanks for the review. Changes should now be addressed.

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -46,8 +46,9 @@ void CAccount::SetName(const std::string& strName)
 {
     if (m_strName != strName)
     {
-        m_pManager->ChangingName(this, m_strName, strName);
-        m_strName = strName;
+        std::string strNewName = strName.substr(0, CAccountManager::MAX_USERNAME_LENGTH);
```
botder: Move this line to the beginning of the function and change the if-condition to:
```cpp
if (m_strName != strNewName)
```

botder (changes requested):

patrikjuvonen:
Thanks, fixed.

botder (approved):

Issue #455, opened 2018-09-11
onVehicleExplode triggers twice [server-side]


ArranTuna:
**Describe the bug**
When you call blowVehicle and a player is in a vehicle, onVehicleExplode triggers twice.

**To Reproduce**
srun addEventHandler ( "onVehicleExplode", root, function ( ) outputChatBox "Triggered" end )
srun player = getRandomPlayer ( )
srun local x, y, z = getElementPosition ( player ); warpPedIntoVehicle ( player, createVehicle ( 429, x, y, z + 1 ) )
srun blowVehicle ( getPedOccupiedVehicle ( player ) )

**Expected behavior**
Not to be triggered twice.

**Screenshots**

**MTA Client (please complete the following information):**
1.3.4 last tested

**MTA Server (please complete the following information):**
1.3.4 last tested

**Additional context**
https://bugs.mtasa.com/view.php?id=7817
https://github.com/multitheftauto/mtasa-blue/pull/213

patrikjuvonen:
Fixed by #1997

PR #213, opened 2018-06-30
0007818: Fix "explode" for blowVehicle, add it for client

emre1702:
Bugtracker:
https://bugs.mtasa.com/view.php?id=7817

Content of PR:
1. onVehicleExplode and onClientVehicleExplode will get called once at explosion  
2. add "explode" argument for clientside blowVehicle

Previously:
1. onVehicleExplode gets called two times if a player is nearby
2. onClientVehicleExplode doesn't get called when using false for explode argument, but onVehicleExplode does

Worked the whole night for it :/
Tested it, but it could be possible that I forgot to test a scenario.

Fixes #455

PR #1997, opened 2021-01-10
Refactor vehicle blowup

botder:
## Summary
- Refactor Lua functions to use new argument parser with warnings
- Remove **explode** parameter for `blowVehicle` (it was never implemented properly and caused onVehicleExplode to trigger twice when a client send the explosion sync for the vehicle blow)
- Fixing or destroying the vehicle in onVehicleExplode avoids broadcasting the blow-up packet

Fixes issue #455 (**onVehicleExplode triggers twice [server-side]**)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicle.h
+++ b/Client/mods/deathmatch/logic/CClientVehicle.h
@@ -198,7 +198,7 @@ public:
     void  SetHealth(float fHealth);
     void  Fix();
     void  Blow(bool bAllowMovement = false);
-    bool  IsVehicleBlown() { return m_bBlown; };
+    bool  IsVehicleBlown() const noexcept { return m_bBlown; };
 
     CVehicleColor& GetColor();
     void           SetColor(const CVehicleColor& color);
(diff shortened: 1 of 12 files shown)


patrikjuvonen:
There is also #213 that attempts to fix this same issue.

AlexTMjugador:
Is there any reason this PR was merged instead of #213? That other PR seems to be more discussed (although more complex) and also does not remove the `blowVehicle` parameter, which may be useful for some scenarios.

botder:
The second parameter was broken by design and it did only one thing very well: let the server-side event get triggered twice.
Even the parameter's name "explode" is rather confusing if you consider the name of the function being called is "blowVehicle".

About the other pull request, I wanted a simple solution to the existing problem without any further feature bloat, and my solution even avoids sending the packets to every client. Also, the pull request is at least one year old and I didn't want to spend time writing back and forth with the pull request author with reviews inbetween, because I could implement it myself in less time.

And if there is a need for a "silent" blowVehicle functionality, then we could always push that information as argument to the event handler function and still trigger the event.

PR #2011, opened 2021-01-15
Update copyright year(s)

fresholia:
Screenshot_1

Updated copyright year validity in many places, including photo

### DIFF
--- a/Shared/data/MTA San Andreas/MTA/locale/ar/client.po
+++ b/Shared/data/MTA San Andreas/MTA/locale/ar/client.po
@@ -3351,8 +3351,8 @@ msgstr ""
 #~ "تصحيح: بناء الجملة هو : [<كلمة المرور> <الأسم>] <المنفذ> <الخادم> connect'"
 
 #, fuzzy
-#~ msgid "Copyright (C) 2003 - 2018 Multi Theft Auto"
-#~ msgstr "حقوق التأليف والنشر © 2003 - 2014 Multi Theft Auto"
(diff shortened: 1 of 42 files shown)


patrikjuvonen:
At least the .po file changes should be reverted as they are commented out and do not affect the mod.

As for header files perhaps we could overall revamp that string to not be year dependent perhaps. Also there are Rockstar North copyright strings that have an old year.

Let's think about this.

patrikjuvonen (changes requested):
At least the .po file changes should be reverted as they are commented out and do not affect the mod.

As for header files perhaps we could overall revamp that string to not be year dependent perhaps. Also there are Rockstar North copyright strings that have an old year.

Let's think about this.

patrikjuvonen:
Additionally I asked ccw to update the copyright year on build server and that is now done, so the exe should have right year.

fresholia:
Instead of replacing all of the header files in bulk, the person who last updated that file can write the current MTA version & year.

Disinterpreter:
https://www.quora.com/If-you-update-a-site-does-the-copyright-year-change
https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

Is it really need?

patrikjuvonen:
Yeah general rule is to update latest year when the file is updated, which is not the case here.

patrikjuvonen:
Thanks for the pull request. This seems to not be necessary. Let's try to keep an eye on those years if code changes within a header file in the future though.

PR #2014, opened 2021-01-17
Fix #817 (clearChatBox() returns false serverside if the player is root)

### LINKED ISSUE #817: clearChatBox() returns false serverside if the player is root
**Describe the bug**
When using clearChatBox() or clearChatBox(root) it returns false even when the chatBox is cleared for every user. This only happens serverside.

**To reproduce**
/srun clearChatBox()

It will return false

**Expected behaviour**
It should return true.

**Version**
Client: v1.5.6-release-16183
Server: v1.5.6-release-16362

Lpsd:
Resolves #817 by adding a simple check if the passed element is equal to the root element.

It's also noted in #817 that other functions have the same issue. Do we want to submit separate PRs for each function affected by this, or bundle it all into this PR?

### DIFF
--- a/Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.Server.cpp
+++ b/Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.Server.cpp
@@ -166,7 +166,7 @@ int CLuaFunctionDefs::ClearChatBox(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        if (CStaticFunctionDefinitions::ClearChatBox(pElement))
+        if (CStaticFunctionDefinitions::ClearChatBox(pElement) || pElement == m_pRootElement)
         {
             lua_pushboolean(luaVM, true);
             return 1;


Lpsd:
also, should we be upgrading any luadefs we touch to the new arg parser? if so i'll do that too

patrikjuvonen:
> It's also noted in #817 that other functions have the same issue. Do we want to submit separate PRs for each function affected by this, or bundle it all into this PR?

> also, should we be upgrading any luadefs we touch to the new arg parser? if so i'll do that too

Let's keep PRs atomic and separate concerns into new PRs!

patrikjuvonen (commented):
I'm thinking if this should be moved into `CStaticFunctionDefinitions::ClearChatBox` somehow since the function is ultimately responsible for it and will cause `true` down the line indirectly...

Maybe we should improve the recursive part to return bool if it affected indirectly, then return true?

patrikjuvonen:
Let's prefer #2110 over this for now, because of my aforementioned concerns.

PR #2018, opened 2021-01-17
Improve shadows by default

patrikjuvonen:
This lowers shadows and lights more towards ground level by default in MTA.

#1699 worked on this, but it edited a variable which is referenced by 20 different places unrelated to shadows and lights which we want to lower, hence it's safer to go through the addresses we want to edit, and apply the new float to them only.

I agree with @botder in that we don't really need to have a setting for this because it's a general improvement and not really something you want to toggle really.

A future PR could make this float editable in Lua maybe if it's useful for any scripters.

- Ground work done by @TFP-dev in #1699 
- Lots of help and list of addresses by @botder - thank you very much for teaching and explaining

I tweaked the float a little bit and found 0.013f is a safe number and doesn't cause any flickering.

- No flickering in 16 bit color mode and/or with near/far clip distance

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -327,6 +327,25 @@ bool        bLocalStatsStatic = true;
 extern bool bWeaponFire;
 float       fDuckingHealthThreshold;
 
+static const std::array<uint32_t, 16> shadowAddr{
+    0x6FAD5D,            // CRegisteredCorona::Update
+    0x7041DB,            // CPostEffects::Fog
+    0x7085A9,            // CShadows::RenderStaticShadows
+    0x709B2F,            // CShadows::CastShadowEntityXY
+    0x709B8E,            // CShadows::CastShadowEntityXY
+    0x709BC7,            // CShadows::CastShadowEntityXY
+    0x709BF6,            // CShadows::CastShadowEntityXY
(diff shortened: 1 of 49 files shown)


botder:
You should add it to the **multiplayer_sa** project, because **game_sa** is meant to be a game interface library and **multiplayer_sa** is there for code modifications.

patrikjuvonen:
> You should add it to the **multiplayer_sa** project, because **game_sa** is meant to be a game interface library and **multiplayer_sa** is there for code modifications.

Thanks, moved the code under multiplayer_sa!

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1498,6 +1498,29 @@ void CMultiplayerSA::InitHooks()
     MemPut((void*)0x6943AD, &fDuckingHealthThreshold);
     fDuckingHealthThreshold = 0;
 
+    // Lower the GTA shadows offset closer to ground/floor level
+    m_fShadowsOffset = 0.013f;            // GTA default = 0.06f
+    std::array<DWORD, 16> shadowAddr{
```
botder: Please move the array out of the function and mark it as static. Also, replace type `DWORD` with `uint32_t`.

patrikjuvonen: Thanks! This has been addressed.

botder (approved):

PR #2022, opened 2021-01-18
Update detours library to version 4.0.1

botder:
https://github.com/microsoft/Detours/releases/tag/v4.0.1

Tested and compiles fine on both Windows and Linux for me.

@ccw808 is there something to be aware of in regards to our build server and network libraries?


ccw808:
Update to detours 3.0 was attempted dfd8f6b4 and subsequently reverted 4d15d024 with no explanations. What do you think of the helper functions DetourRemove and DetourFunction from that commit?
I'll test this with the network library soon

botder:
I don't know the reason why it was reverted back then. The helper functions could be useful in many cases though. AddUtf8FileHooks and RemoveUtf8FileHooks should be done in a single transaction without these helper functions.

botder:
It's looking much better now, with the SharedUtil.Detours.h header.

botder:
@ccw808 please merge it if you don't encounter any issues on your side.

botder:
Any progress on this, @ccw808? I would like to use the updated library for the next changes.

PR #2023, opened 2021-01-18
Add onClientObjectMoveStart, onClientObjectMoveStop, isObjectMoving

### LINKED ISSUE #874: isObjectMoving(), onObjectMoved
We can move an object using `moveObject()` function, but we cannot know whether the object has finished moving. 

Add a new function `isObjectMoving()` which returns `true` if object moving or `false` otherwise. Add a new event `onObjectMoved`.

In some cases I can store target coords and then compare the current position with them, but it won't work with some types of easing (elastic, outback, bounce, etc).

**Additional context**
no

fresholia:
Fixes #874
Added 2 events 1 functions:
```lua
bool isObjectMoving(object)
event onClientObjectMoveStart
event onClientObjectMoveStop
```

Code on lua:
```lua
local obj = createObject(5239, -2417.22339, -606.70374, 132.56250)
moveObject(obj, 3000, -2417.22339, -606.70374, 137.56250)

addEventHandler("onClientObjectMoveStart", obj,
    function()

    end
)

addEventHandler("onClientObjectMoveStop", obj,
    function()
        outputChatBox("done")
    end
)
setTimer(
    function()
        outputChatBox(tostring(isObjectMoving(obj)))
    end,
500, 0)
```

### DIFF
--- a/Client/mods/deathmatch/logic/CDeathmatchObject.cpp
+++ b/Client/mods/deathmatch/logic/CDeathmatchObject.cpp
@@ -100,7 +100,6 @@ void CDeathmatchObject::UpdateMovement()
 
     SetOrientation(positionRotation.m_vecPosition, positionRotation.m_vecRotation);
 
-
     if (!bStillRunning)
     {
         _StopMovement(false);            // We don't unregister ourselves here since CDeathmatchObject::UpdateMovement is called from an iteration in
--- a/Client/mods/deathmatch/logic/CDeathmatchObject.cpp
+++ b/Client/mods/deathmatch/logic/CDeathmatchObject.cpp
@@ -98,19 +98,22 @@ void CDeathmatchObject::UpdateMovement()
 
     SetOrientation(positionRotation.m_vecPosition, positionRotation.m_vecRotation);
 
-    CLuaArguments Arguments;
-    CVector       currentPosition;
-    this->GetPosition(currentPosition);
-    Arguments.PushNumber(currentPosition.fX);
-    Arguments.PushNumber(currentPosition.fY);
-    Arguments.PushNumber(currentPosition.fZ);
-    this->CallEvent("onClientObjectMove", Arguments, true);
 
     if (!bStillRunning)
     {
         _StopMovement(false);            // We don't unregister ourselves here since CDeathmatchObject::UpdateMovement is called from an iteration in
                                          // CMovingObjectsManager::DoPulse
         // and we are automatically removed from the list after CDeathmatchObject::UpdateMovement if we are finished
+    }
+    else
+    {
+        CLuaArguments Arguments;
+        CVector       currentPosition;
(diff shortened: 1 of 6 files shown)


Client/mods/deathmatch/logic/CDeathmatchObject.cpp
```diff
@@ -96,6 +98,14 @@ void CDeathmatchObject::UpdateMovement()
 
     SetOrientation(positionRotation.m_vecPosition, positionRotation.m_vecRotation);
 
+    CLuaArguments Arguments;
+    CVector       currentPosition;
+    this->GetPosition(currentPosition);
+    Arguments.PushNumber(currentPosition.fX);
+    Arguments.PushNumber(currentPosition.fY);
+    Arguments.PushNumber(currentPosition.fZ);
+    this->CallEvent("onClientObjectMove", Arguments, true);
```
botder: This is going to crash the client if somebody decides to use `stopObject` or `moveObject` inside that event handler.

Lpsd (changes requested):
Small thing, single-line ifs like this:
```cpp
if (true)
    foo.DoSomething();
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -3865,14 +3865,17 @@ bool CStaticFunctionDefinitions::IsObjectBreakable(CClientObject& Object, bool&
...
+
     if (IS_OBJECT(&Entity))
-    {
         bIsMoving = Object.IsMoving();
-    }
-    return true;
+    else
+        bIsMoving = false;
+
+    return bIsMoving;
```
Lpsd: Remove this and just initialize `bIsMoving` as false.

Lpsd: Didn't mean to include the `return bIsMoving;` line

Lpsd (commented):
Looks good, left you one thing to deal with.

Client/mods/deathmatch/logic/CDeathmatchObject.cpp
```diff
@@ -96,6 +100,7 @@ void CDeathmatchObject::UpdateMovement()
 
     SetOrientation(positionRotation.m_vecPosition, positionRotation.m_vecRotation);
 
+
```
Lpsd: Unnecessary additional newline

Lpsd (changes requested):
Good work, just fix that extra newline and we're good.

Lpsd (approved):

patrikjuvonen:
@cleopatradev Please create new Wiki pages for these events and the function, thank you.
https://wiki.multitheftauto.com/wiki/Main_Page

PR #2027, opened 2021-01-18
Add matchingDimension param for onPickupLeave

### LINKED ISSUE #846: onPickupLeave+setDimension
onPickupLeave do not detect the dimension change,idk how about interior.

fresholia:
Fixes #846

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1487,7 +1487,7 @@ void CGame::AddBuiltInEvents()
 
     // Pickup events
     m_Events.AddEvent("onPickupHit", "player", NULL, false);
-    m_Events.AddEvent("onPickupLeave", "player", NULL, false);
+    m_Events.AddEvent("onPickupLeave", "player, matchingDimension", NULL, false);
     m_Events.AddEvent("onPickupUse", "player", NULL, false);
     m_Events.AddEvent("onPickupSpawn", "", NULL, false);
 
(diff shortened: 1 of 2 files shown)


botder:
This breaks backwards compatibility and we should only merge it for 1.6

sbx320:
Even for 1.6 this seems like a very significant compat break, since you'd suddenly be able to trigger pickups from other dimensions. To me this seems more of a thing for the API versioning suggestion.

Allerek:
> This breaks backwards compatibility and we should only merge it for 1.6
(Writing this from script perspective, wasnt looking into code)
It shouldnt break compatibility as older script doesnt have that argument in event, so they if they havent had it, it will doesnt make big difference(if parameter order is still same, and MTA Code doesnt brake that compatibility by itself)

Allerek:
> This breaks backwards compatibility and we should only merge it for 1.6

(Writing this from script perspective, wasnt looking into code)
It shouldnt break compatibility as older script doesnt have that argument in event, so they if they havent had it, it will doesnt make big difference(if parameter order is still same, and MTA Code doesnt brake that compatibility by itself)

sbx320:
@Allerek Old resources wouldn't expect pickups to trigger in other dimensions. For example consider a weapon pickup in dimension 1 which gives a player a deagle when leaving the pickup. 
**Old Behavior**:
Players in dimension 0 cannot reach the pickup (onPickupLeave)
**New behavior**:
Players in dimension 0 can reach the pickup. onPickupLeave gets triggered and since I did not check for the dimension, the player gets the weapon.

PR #2032, opened 2021-01-20
Allow multiple ignored elements to be passed to processLineOfSight

### LINKED ISSUE #1549: processLineOfSight/isLineOfSightClear should allow to ignore multiple elements
Currently, i'm facing an scenario where i need to ignore multiple elements using processLineOfSight, however function allows to ignore only one element.

Allow function to accept table of elements aswell.

Calling processLineOfSight/isLineOfSightClear more than once.

**Additional context**
-

Lpsd:
This allows multiple ignore elements to be passed to processLineOfSight, via table of elements.

It is backwards compatible. If you pass a single element, or nil as `ignoredElements` it has the same behaviour as before.

```lua
processLineOfSight ( float startX, float startY, float startZ, 
                       float endX, float endY, float endZ, 
                       [ bool checkBuildings = true, 
                       bool checkVehicles = true, 
                       bool checkPlayers = true, 
                       bool checkObjects = true, 
                       bool checkDummies = true, 
                       bool seeThroughStuff = false, 
                       bool ignoreSomeObjectsForCamera = false, 
                       bool shootThroughStuff = false, 
                       table ignoredElements = nil,
                       bool includeWorldModelInformation = false,
                       bool bIncludeCarTyres ] )
```

Partially resolves #1549 

I will submit a separate PR for isLineOfSightClear


0xHexadecimal:
Wouldn't it be better to take a table of elements?

```lua
elements = {}
processLineOfSight(...,unpack(elements),...) -- can't do this
```

I suppose you could the put last arguments as the last values in the table.

Lpsd:
I'll add the ability to pass a table, as well as vararg.

Having it at the end would be weird, because we have to keep the original argument order to some extent.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6243,20 +6243,28 @@ bool CStaticFunctionDefinitions::SetTime(unsigned char ucHour, unsigned char ucM
...
+    std::vector<CClientEntity*> vecDidIgnoreElements;
+
+    for (CClientEntity* pIgnoredElement : vecIgnoredElements)
+    {
+        if (CStaticFunctionDefinitions::GetElementCollisionsEnabled(*pIgnoredElement))
+        {
+            CStaticFunctionDefinitions::SetElementCollisionsEnabled(*pIgnoredElement, false);
+            vecDidIgnoreElements.push_back(pIgnoredElement);
+        }
+    }
```
Pirulax: You shouldn't be creating a new vector here, it's slow.
I'd suggest you to do this instead:
```cpp
vecIgnoredElements.erase(std::remove_if(vecIgnoredElements.begin(), vecIgnoredElements.end(), [](const CClientEntity* pEntity) {
    // Remove entities that already have their colision disabled. 
    // This prevents us from reenabling them.
    if (CStaticFunctionDefinitions::GetElementCollisionsEnabled(*pIgnoredElement))
        return true;
    
    // Otherwise disable collision and keep it in the array
    CStaticFunctionDefinitions::SetElementCollisionsEnabled(*pIgnoredElement, false);
    
    return false;       
), vecIgnoredElements.end();
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -562,7 +562,7 @@ class CStaticFunctionDefinitions
     // World functions
     static bool GetTime(unsigned char& ucHour, unsigned char& ucMin);
     static bool ProcessLineOfSight(const CVector& vecStart, const CVector& vecEnd, bool& bCollision, CColPoint** pColPoint, CClientEntity** pColEntity,
-                                   const SLineOfSightFlags& flags = SLineOfSightFlags(), CEntity* pIgnoredEntity = NULL,
+                                   const SLineOfSightFlags& flags = SLineOfSightFlags(), std::vector<CClientEntity*> vecIgnoredElements = std::vector<CClientEntity*>(),
```
Pirulax: `= {}`

Pirulax (dismissed):

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -242,18 +252,38 @@ int CLuaWorldDefs::ProcessLineOfSight(lua_State* luaVM)
     argStream.ReadBool(flags.bSeeThroughStuff, false);
     argStream.ReadBool(flags.bIgnoreSomeObjectsForCamera, false);
     argStream.ReadBool(flags.bShootThroughStuff, false);
-    argStream.ReadUserData(pIgnoredElement, NULL);
+
+    if (argStream.NextIsTable()) // Is the next value a table? Read it as a user data table (will error if table contains invalid type)
+    {
+        argStream.ReadUserDataTable(vecIgnoredElements);
+    }
+    else if (argStream.NextIsNil()) // Is the next value nil? We can just skip to reading bIncludeBuildingInfo
```
Pirulax: `NextIsNone()` I'd instead I'd say. Nil is error prone (user thinks he's passing in an element, and is wondering why it's not working)
Do it liket his:
```cpp
if None:
   skip...
else:
    if UserData:

    else if Table:
```

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -242,18 +252,38 @@ int CLuaWorldDefs::ProcessLineOfSight(lua_State* luaVM)
...
+    {
+        CClientEntity* pElement;
+        argStream.ReadUserData(pElement);
+        vecIgnoredElements.push_back(pElement);
+        while(!argStream.NextIsNone() && !argStream.NextIsBool()) // Keep trying to read each successive argument as userdata until the next argument has not been supplied, or it is a bool (bIncludeBuildingInfo).
+        {
+            CClientEntity* pElement;
+            argStream.ReadUserData(pElement);
+            vecIgnoredElements.push_back(pElement);
+        }
```
Pirulax: I think you could `reserve(4096)` here. It might sound a lot, but this vector lives a very short life, and reserving a lot saves a lot of CPU.

Pirulax:
Other than these comments you're fine imho.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Lpsd: I need to check for `NextIsNil` to keep the default behaviour. 

If passing nil, it would would skip to the next arg.

If `NextIsNone`, then it doesn't need to skip because nothing has been passed, so there also can't be another argument after that.

Honestly, I did like an hour or two of testing conditions on this and I'm certain my version is the most consistent and keeps the previous behaviour. If you can provide me with the exact logic you think would work better, I would be happy to change it. (the one you have provided has issues already)

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Lpsd: I will look into this, thanks.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Pirulax: There can be an argument after a none. If you pass `_` it act's as none, not nil. 
Eg.: `dxDrawText("text", 0, 0, _, _, ...)`

Lpsd: I'm sure I tested that and it was nil, I'll try it again later. If it does work how you suggest then I'll just change the `else if (argStream.NextIsNil())` to `else if (argStream.NextIsNone())` - that should be fine.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Pirulax: Any update?

Pirulax:
Time to merge this.
@Lpsd address the comments, and I'll merge it.
I hope you've tested it (But there's really nothing not to work, so)

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Pirulax: I think a table should be used here instead. Realistically people will have to pack it into a table anyways, and then they have to call unpack, so might as well accept a table instead.
I really have a bad mouth taste coming from `dxDrawPrimitive`.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Lpsd: I have removed vararg for this, it's kind of unnecessary. It'll accept either single element (current behaviour) or table of elements.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Lpsd: Not using vararg anymore, ReadUserDataTable reads into vector so this is resolved

Lpsd:
Vararg removed.

Test resource: [los.zip](https://github.com/multitheftauto/mtasa-blue/files/9126630/los.zip)
 
Spawns 50 bin objects in a row. Use command `/los <amountToIgnore>`

e.g `/los 50` should return nil for processLineOfSight in debug output, as you're ignoring all elements
`/los 30` will output `bin_31`, as that's the next object to hit after 30
`/los 0` will output `bin_1`
etc

This resource also tests backwards compatibility by passing `nil` as `ignoredElements` if 0 is supplied to the command. It shouldn't raise a warning / error (just like original behaviour).

patrikjuvonen (approved):

Einheit-101:
Is there anything left to do here or why is it not merged?

patrikjuvonen:
> Is there anything left to do here or why is it not merged? 

Nothing left to do here, not merged yet because feature freeze is on (https://discord.com/channels/801330706252038164/801330706252038167/1059094962853130250). Will be merged for 1.6.

PR #2041, opened 2021-01-23
Fix #1939: argument parser not handling variants properly

### LINKED ISSUE #1658: New parser doesn't support objects
**Describe the bug**
New parser doesn't supports objects produced by eg. setTimer. You also can't refactor any function that use resource

**To reproduce**
try refactor setTimer to use new parser, you will get simillar error:
image

**Expected behaviour**
It should be allowed to use 

**Screenshots**
above

**Version**
newstest

**Additional context**

```cpp

void lua_pushtimer(lua_State* luaVM, CLuaTimer* pTimer)
{
    const char* szClass = NULL;
    CLuaMain*   pLuaMain = g_pClientGame->GetLuaManager()->GetVirtualMachine(luaVM);
    if (pLuaMain->IsOOPEnabled())
        szClass = CLuaClassDefs::GetTimerClass(pTimer);

    lua_pushobject(luaVM, szClass, (void*)reinterpret_cast<unsigned int*>(pTimer->GetScriptID()));
}
```
^ missing in new parser, and rest of, resource and few other

### LINKED ISSUE #1939: New parser, issues with variant of classes inherited from same base class
**Describe the bug**
You can use only first variant, second will ignored and i'll throw error

**To reproduce**
pseudocode:
```cpp
{"feed", ArgumentParser<Feed>}

class Animal {}

class Fox : public Animal {}
class Cat : public Animal {}

void CAnimalDefs::Feed(std::variant<Fox, Cat> animal) {};
```

```lua
feed(fox) -- fine
feed(cat) -- expected "animal-fox"
```

**Expected behaviour**
Both should be work

**Screenshots**

**Version**
newstests

**Additional context**
```cpp
void CAnimalDefs::Feed(Animal animal) {};
```
you can use base class, then use dynamic_cast to cast it to fox or cat depends what user provided

sbx320:
See #1939 for investigation details. It appears we'll just have to retrieve the variant from the luaVM and check it's type. 

Also includes a minor refactor to is_Nspecialization to make it more generic.


sbx320:
Note: failing linux ci is due to a gcc bug. I'll see if we can upgrade that

sbx320:
Blocked by #2045 (and likely a build server update)

Pirulax:
Opened a [PR](https://github.com/sbx320/mtasa-blue/pull/1/) to fix the compilation issues.
Edit: Accidentally merged my branch.... Anyways. It should wort just fine, no breaking changes.
Edit2: Actually I presume the "Failing linux CI due to GCC bug" was referring to this. Oh well, I fixed it already :D

PR #2042, opened 2021-01-23
Fix lua::Push not working on non-element entities

sbx320:
`lua_pushelement` only pushes elements. `lua_pushuserdata` is the correct way to push arbitrary script entities. 

With C++20 we should use a `LuaPushable` concept.

### DIFF
--- a/Shared/mods/deathmatch/logic/lua/LuaBasic.h
+++ b/Shared/mods/deathmatch/logic/lua/LuaBasic.h
@@ -112,7 +112,7 @@ namespace lua
     template <typename T>
     std::enable_if_t<std::is_class_v<T>> Push(lua_State* L, T* val)
     {
-        lua_pushelement(L, val);
+        lua_pushuserdata(L, val);
     }
 
     template <typename T>
--- a/Shared/mods/deathmatch/logic/lua/LuaBasic.h
+++ b/Shared/mods/deathmatch/logic/lua/LuaBasic.h
@@ -102,24 +102,24 @@ namespace lua
 
     // Script entities
     #ifdef MTA_CLIENT
-    void Push(lua_State* L, const CClientEntity* value) { lua_pushelement(L, value); }
+    inline void Push(lua_State* L, const CClientEntity* value) {lua_pushelement(L, const_cast<CClientEntity*>(value)); }
     #else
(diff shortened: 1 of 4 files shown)


Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -123,7 +123,7 @@ namespace lua
     template <typename T>
     std::enable_if_t<std::is_class_v<T>, int> Push(lua_State* L, T* val)
     {
-        lua_pushelement(L, val);
+        lua_pushuserdata(L, val);
```
Pirulax: pushuserdata is way slower.
```cpp
if constexpr (std::is_base_of_v<CElement, T>)
    lua_pushelement(L, val);
else
    lua_pushuserdata(L, val);
```

sbx320: Yeah I thought about that too. But since I'll redo all of that with concepts anyway, I just skipped over it for now and kept the fix small.

sbx320: Actually I'll change this up... `lua::Push` won't change too much with concepts

Shared/mods/deathmatch/logic/lua/LuaBasic.h
```diff
@@ -100,6 +100,28 @@ namespace lua
...
+    #endif
+    inline void Push(lua_State* L, const CResource* value) { lua_pushresource(L, const_cast<CResource*>(value)); }
+    inline void Push(lua_State* L, const CXMLNode* value) { lua_pushxmlnode(L, const_cast<CXMLNode*>(value)); }
+    inline void Push(lua_State* L, const CLuaTimer* value) {lua_pushtimer(L, const_cast<CLuaTimer*>(value)); }
+    inline void Push(lua_State* L, const CLuaVector2D* value) { lua_pushvector(L, *value); }
+    inline void Push(lua_State* L, const CLuaVector3D* value) { lua_pushvector(L, *value); }
+    inline void Push(lua_State* L, const CLuaVector4D* value) { lua_pushvector(L, *value); }
+    inline void Push(lua_State* L, const CLuaMatrix* value) { lua_pushmatrix(L, *value); }
+
+
```
Pirulax, line 124: Nice

Pirulax (approved):

PR #2047, opened 2021-01-25
Refactor shared/sdk headers to include required headers

sbx320:
This PR adjusts all headers (`.h`) in Shared/sdk to include headers required by them. Previously many of the headers relied on getting included in the right order. Long term this will allow us to increase compilation speed by reducing PCH impact. 

To accomplish this I've compiled each header individually, except `CScriptArgReader` and `net/SyncStructures` (those two rely on specific parts of server/client deathmatch) by simply creating a new `.cpp` file which only includes that header. I then added (1) standard C++ headers and (2) other sdk headers until the source file compiled. I've also added `#pragma once` to headers where it was missing.

*Notes*

- I've moved some very common typedefs (e.g. `uint`) to a new header `SharedUtil.TypeDefs.h` to avoid including everything via `SharedUtil.h`.
- Minimal other code changes (mainly to replace transitive includes of macros) to avoid breaking anything at runtime. Might still break `net`, but we'll figure that out. 
- I've also moved `bochs/bochs_internal/crc32.{cpp,h}` to `bochs/bochs_internal/bochs_crc32.{cpp,h}` as I ran into name conflicts with zlibs `crc32.h`.


Client/core/CAdditionalVertexStreamManager.cpp
```diff
@@ -329,6 +329,7 @@ bool CAdditionalVertexStreamManager::UpdateAdditionalStreamContent(SCurrentState
         {
             // Validate
             CVector& Normal = NormalList[i];
+            static constexpr float FLOAT_EPSILON = 0.0001f;
             if (Normal.Normalize() < FLOAT_EPSILON)
```
botder, line 333:
```suggestion
            if (Normal.Normalize() < std::numeric_limits<float>::epsilon())
```

Shared/sdk/CVector.h
```diff
@@ -11,20 +11,17 @@
 
 #pragma once
 
-#ifdef WIN32
-#include <windows.h>
-#endif
-
 #include <cmath>
```
botder, line 14:
```suggestion
#include <cmath>
#include <limits>
```

Shared/sdk/CVector.h
```diff
@@ -11,20 +11,17 @@
...
-#define FLOAT_EPSILON 0.0001f
 #include "CVector4D.h"
 
 /**
  * CVector Structure used to store a 3D vertex.
  */
 class CVector
 {
+private:
+    static constexpr float FLOAT_EPSILON = 0.0001f;
```
botder, line 24:
```suggestion
private:
    static constexpr float FLOAT_EPSILON = std::numeric_limits<float>::epsilon();
```

Shared/sdk/SharedUtil.Defines.h
```diff
@@ -191,3 +191,4 @@
                 : __DATE__ [2] == 't' ? 9 \
                 : __DATE__ [2] == 'v' ? 10 : 11)
 #define BUILD_DAY ((__DATE__ [4]==' ' ? 0 : __DATE__[4]-'0')*10+(__DATE__[5]-'0'))
+static constexpr float FLOAT_EPSILON = 0.0001f;
```
botder, line 194:
```suggestion
static constexpr float FLOAT_EPSILON = std::numeric_limits<float>::epsilon();
```

Shared/sdk/SharedUtil.SysInfo.h
```diff
@@ -8,9 +8,12 @@
  *  Multi Theft Auto is available from http://www.multitheftauto.com/
  *
  *****************************************************************************/
+#pragma once
 
 #ifdef WIN32
-
+#include <vector>
+#include "SString.h"
+#include <Windows.h>
```
botder:
```suggestion
#include <windows.h>
```
For cross-compilation

botder (changes requested):
SharedUtil.IntTypes.h seems unfinished without the regular file comment header, replacing the types with <cstdint>/<cstddef> MIGHT break stuff if not done carefully.

Client/core/CAdditionalVertexStreamManager.cpp
sbx320: `std::numeric_limits<float>::epsilon()` isn't ideal, since it's only defined to 

> the difference between 1.0 and the next value representable by the floating-point type T

which is commonly too small for a lot of floating point calculations.

Shared/sdk/SharedUtil.SysInfo.h
sbx320: The file is called `Windows.h` (uppercase `W`) in the winsdk, so I'm not sure why we'd need to lowercase it for cross compiling.

botder: It only matters if you compile on a system with case-sensitive paths and the Windows header is actually lowercase (like for MinGW)

sbx320: Interesting, haven't run into that yet

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -9,10 +9,15 @@
...
 
+#include "SharedUtil.Misc.h"
+#include "SharedUtil.Time.h"
+#include <map>
 #include "UTF8.h"
 #include "UTF8Detect.hpp"
+#include "CDuplicateLineFilter.h"
 #ifdef WIN32
     #include <ctime>
+    #include "Windows.h"
```
botder:
```suggestion
    #include <windows.h>
```

Shared/sdk/SharedUtil.Logging.hpp
```diff
@@ -9,6 +9,16 @@
...
  *****************************************************************************/
 
+#include "SharedUtil.Logging.h"
+#include <set>
+#include "SharedUtil.Misc.h"
+#include "SharedUtil.File.h"
+#include "SharedUtil.Time.h"
+
+#ifdef _WIN32
+    #include <Windows.h>
```
botder:
```suggestion
    #include <windows.h>
```

Shared/sdk/SharedUtil.Time.hpp
```diff
@@ -8,9 +8,15 @@
...
-
+#include "SharedUtil.Time.h"
+#include "SharedUtil.Logging.h"
 #ifndef WIN32
     #include "sys/time.h"
+#else
+    #ifndef NOMINMAX
+    #define NOMINMAX
+    #endif
+    #include <Windows.h>
```
botder:
```suggestion
    #include <windows.h>
```

botder (changes requested):
The float epsilon stuff is fine, I marked my comments as resolved. My requested changes are trivial and the pull request looks good.

botder (approved):

PR #2049, opened 2021-01-25
Fix arg parser returning tuples rather than values from some functions

sbx320:
In #1981 I missed the fact that we already had some functions returning tuples for multiple return values. Those now incorrectly return tuples. This PR reverts to the correct functionality and also adds a `std::variant` overload so that `std::variant<bool, CLuaMultiReturn<int, int>>` works.

Affected functions (thanks @Zangomangu):
```
getElementBoneRotation
getElementBonePosition
engineGetModelVisibleTime
getVehicleDummyPosition
getVehicleModelDummyDefaultPosition
getColPolygonHeight
```

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
@@ -11,7 +11,7 @@
 
 #pragma once
 #include "CLuaDefs.h"
-#include "CLuaMultiReturn.h"
+#include <lua/CLuaMultiReturn.h>
 
 class CLuaEngineDefs : public CLuaDefs
 {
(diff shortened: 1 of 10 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -11,6 +11,7 @@
 
 #pragma once
 #include "CLuaDefs.h"
+#include "CLuaMultiReturn.h"
```
Zangomangu: Needs to be `lua/CLuaMultiReturn.h` to compile

Zangomangu (approved):
Works, also ran through all ArgumentParser functions to make sure we didnt miss one.

PR #2050, opened 2021-01-26
Fix #2048: Player is not able to leave Vehicle, which is about to explode with 0 health 

### LINKED ISSUE #2048: Player is not able to leave Vehicle, which is about to explode with 0 health
When vehicle is on fire with 0 health, player is not able to leave car anymore.
Issue appeared in: #1748

Zangomangu:
The check for dead vehicle was added in https://github.com/multitheftauto/mtasa-blue/commit/94c6aba300460199690b66a38b6db36f4fe63a20 but is unnecessary when exiting.


botder (approved):
I hope this doesn't cause any desync :)

PR #2053, opened 2021-01-28
Add new "raw" argument fix of #2052

### LINKED ISSUE #2052: Add raw variable for HTTP Web Interface
MTA WEB HTTP add variable `raw` for accepting data

Form-data but it is bad for example for JSONRPC

**Additional context**
https://wiki.multitheftauto.com/wiki/Resource_Web_Access
We have some variables, so we need add new:
```
    table requestHeaders: This is a table containing all the headers that were requested with the page. You can set returned headers using httpSetResponseHeader.
    table form: This is a table containing all the form data submitted to the page using HTTP POST combined with any variables passed in the querystring with HTTP GET.
    table cookies: This is a table of all the cookies. You can modify cookies using httpSetResponseCookie.
    string hostname: This is a string containing the IP address or hostname that requested the page.
    string url: This is the URL of the page.
    account user: This is the account of the current user.
```
image

Disinterpreter:
Fixes #2052
New "raw" argument for manual handling requests

### DIFF
--- a/Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
+++ b/Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
@@ -75,6 +75,47 @@ ResponseCode CResourceHTMLItem::Request(HttpRequest* ipoHttpRequest, HttpRespons
             headers.PushString((*iter).second.c_str());
         }
 
+        std::string sMethod;
+        switch (ipoHttpRequest->nRequestMethod)
+        {
+            case REQUESTMETHOD_GET:
+                sMethod = "GET";
+                break;
+            case REQUESTMETHOD_OPTIONS:
+                sMethod = "OPTIONS";
+                break;
+            case REQUESTMETHOD_HEAD:
+                sMethod = "HEAD";
+                break;
+            case REQUESTMETHOD_POST:
+                sMethod = "POST";
+                break;
+            case REQUESTMETHOD_PUT:
+                sMethod = "PUT";
+                break;
+            case REQUESTMETHOD_DELETE:
+                sMethod = "DELETE";
+                break;
+            case REQUESTMETHOD_TRACE:
+                sMethod = "TRACE";
+                break;
+            case REQUESTMETHOD_CONNECT:
+                sMethod = "CONNECT";
+                break;
+            case REQUESTMETHOD_LAST:
+                sMethod = "LAST";
+                break;
+            case REQUESTMETHOD_UNKNOWN:
+                sMethod = "UNKNOWN";
+                break;
+            case REQUESTMETHOD_INVALID:
+                sMethod = "INVALID";
+                break;
+            default:
+                sMethod = "INVALID";
+                break;
+        }
+
         m_currentResponse = ipoHttpResponse;
         CLuaArguments querystring(formData);
         CLuaArguments args;
@@ -86,6 +127,7 @@ ResponseCode CResourceHTMLItem::Request(HttpRequest* ipoHttpRequest, HttpRespons
         args.PushTable(&querystring);                                     // querystring
         args.PushAccount(account);
         args.PushString(ipoHttpRequest->sBody);                           // raw
+        args.PushString(sMethod);                                         // method
 
         // g_pGame->Lock(); // get the mutex (blocking)
         args.CallGlobal(m_pVM, "renderPage");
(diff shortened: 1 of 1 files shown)


Citizen01:
hmmm what about `body` instead of `raw` ?
I don't think raw is meaningful enough as to what it is (could be the raw http request that contains the headers, cookies, body ...)
I like that WebServer is getting more features like providing the body content. The other thing I really wished it had at some point was to allow other HTTP methods (not just GET and POST) like DELETE and PUT (at least) and why not PATCH, HEAD, OPTIONS ? (I know it's out of the scope, I just wanted to let it out :p)

Disinterpreter:
> 
> 
> hmmm what about `body` instead of `raw` ?
> I don't think raw is meaningful enough as to what it is (could be the raw http request that contains the headers, cookies, body ...)
> I like that WebServer is getting more features like providing the body content. The other thing I really wished it had at some point was to allow other HTTP methods (not just GET and POST) like DELETE and PUT (at least) and why not PATCH, HEAD, OPTIONS ? (I know it's out of the scope, I just wanted to let it out :p)

I can rename it, if other people agree with you. There's no problem.
DELETE/PUT/ETC... It does with ~raw~body. I get it. But as a problem we know nothing about type of request.

Disinterpreter:
I implemented new argument known as **method** it return the name of type of request, like GET/POST/PUT...

you can try it with
```
<*
iprint(requestHeaders)
iprint(method)
iprint(raw)
httpSetResponseHeader ( "content-type", "application/json")
httpWrite("ok")
*>
```

Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
```diff
@@ -75,6 +75,47 @@ ResponseCode CResourceHTMLItem::Request(HttpRequest* ipoHttpRequest, HttpRespons
...
+                break;
+            case REQUESTMETHOD_UNKNOWN:
+                sMethod = "UNKNOWN";
+                break;
+            case REQUESTMETHOD_INVALID:
+                sMethod = "INVALID";
+                break;
+            default:
+                sMethod = "INVALID";
+                break;
```
Citizen01:
```suggestion
            case REQUESTMETHOD_INVALID:
            default:
                sMethod = "INVALID";
                break;
```

Citizen01 (changes requested):
And I still want it to be named `body`. Here are my arguments:

1. You are exposing the `sBody` attribute of the HttpRequest why not naming it `body` ?
2. `sMethod` is exposed as `method`
3. `headers` list is exposed as `headers`
4. `formData` as `form`, `cookies` as `cookies`, then why `sBody` as `raw` ?
5. I think the `raw` only came from postman ?

Citizen01:
I don't think there is a poll system so let's do it like this:

If you want the body to be exposed as `raw`,  react with ❤️ 
If you want the body to be exposed as `body`, react with 🚀

Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
```diff
@@ -75,6 +75,45 @@ ResponseCode CResourceHTMLItem::Request(HttpRequest* ipoHttpRequest, HttpRespons
             headers.PushString((*iter).second.c_str());
         }
 
+        std::string sMethod;
```
botder:
```suggestion
        const char* sMethod = "INVALID";
```
Is the heap allocation inside `std::string` really necessary, if you only assign strings from static storage?
With the suggestion you can also simplify the `default` case blow.
`CLuaArguments::PushString` should support pointer to const char.

botder:
> If you want the body to be exposed as `body`, react with 🚀

What about `requestBody`? That would be analog to `requestHeaders`.

Disinterpreter:
> 
> 
> > If you want the body to be exposed as `body`, react with 🚀
> 
> What about `requestBody`? That would be analog to `requestHeaders`.

sounds fine for me, and I think it same for @Citizen01 so I did it. And added code suggestions of @botder

botder (approved):

Citizen01 (approved):
Tried to brake it sending weird postman requests but everthing I tried worked as expected so it's okay for me too.
Thanks for this PR, this will help me getting back into the API project I had and will simplify the whole thing too ! 👍 🚀

PR #2058, opened 2021-01-31
Add on(Client)ElementInteriorChange

### LINKED ISSUE #2056: Add onElementInteriorChange
We have onElementDimensionChange since r20724, but **onElementInteriorChange is still missing**.

Add **onElementInteriorChange** and **onClientElementInteriorChange**, which will be triggered when element's interior changed by setElementInterior

gptrk0:
Closes #2056

### DIFF
--- a/Server/mods/deathmatch/logic/CElement.h
+++ b/Server/mods/deathmatch/logic/CElement.h
@@ -219,7 +219,7 @@ public:
     void RemoveOriginSourceUser(class CPed* pPed) { m_OriginSourceUsers.remove(pPed); }
 
     unsigned char GetInterior() { return m_ucInterior; }
-    void          SetInterior(unsigned char ucInterior) { m_ucInterior = ucInterior; }
+    virtual  void SetInterior(unsigned char ucInterior);
 
     bool IsDoubleSided() { return m_bDoubleSided; }
     void SetDoubleSided(bool bDoubleSided) { m_bDoubleSided = bDoubleSided; }
--- a/Server/mods/deathmatch/logic/CElement.h
+++ b/Server/mods/deathmatch/logic/CElement.h
@@ -219,7 +219,7 @@ public:
     void RemoveOriginSourceUser(class CPed* pPed) { m_OriginSourceUsers.remove(pPed); }
 
     unsigned char GetInterior() { return m_ucInterior; }
-    virtual  void SetInterior(unsigned char ucInterior);
+    void          SetInterior(unsigned char ucInterior);
 
     bool IsDoubleSided() { return m_bDoubleSided; }
     void SetDoubleSided(bool bDoubleSided) { m_bDoubleSided = bDoubleSided; }
--- a/Server/mods/deathmatch/logic/CElement.h
+++ b/Server/mods/deathmatch/logic/CElement.h
@@ -219,7 +219,7 @@ public:
     void RemoveOriginSourceUser(class CPed* pPed) { m_OriginSourceUsers.remove(pPed); }
 
     unsigned char GetInterior() { return m_ucInterior; }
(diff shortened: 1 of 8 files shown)


Server/mods/deathmatch/logic/CElement.h
```diff
@@ -219,7 +219,7 @@ class CElement
     void RemoveOriginSourceUser(class CPed* pPed) { m_OriginSourceUsers.remove(pPed); }
 
     unsigned char GetInterior() { return m_ucInterior; }
-    void          SetInterior(unsigned char ucInterior) { m_ucInterior = ucInterior; }
+    virtual  void SetInterior(unsigned char ucInterior);
```
Pirulax: Why tho? I see no need to make it a virtual method.

Pirulax:
Also, why didnt you make it cancellable? 
It's totally possible to do so.

Citizen01:
> Also, why didnt you make it cancellable?
> It's totally possible to do so.

on(Client)ElementDimensionChanged is not cancelable either and got merged:
https://github.com/multitheftauto/mtasa-blue/commit/7742f34b1fe9b12d3f75d8637e1d73d8fab7d997

Also this event will be triggered even if calling it with the same dimension the entity is already in whereas on(Client)ElementDimensionChanged does not:
```cpp
void CClientEntity::SetDimension(unsigned short usDimension)
{
    if (m_usDimension == usDimension)
        return;

    unsigned int usOldDimension = m_usDimension;
    m_usDimension = usDimension;

    CLuaArguments Arguments;
    Arguments.PushNumber(usOldDimension);
    Arguments.PushNumber(usDimension);
    CallEvent("onClientElementDimensionChange", Arguments, true);
}
```
Both events are meant to be released in 1.5.8 and should act the same way.

If you want it to be cancelable then fine, do it so, but then let's create an issue for the `on(Client)ElementDimensionChanged` event to be cancelable too.

Question: What should happen if the client event is canceled when the interior is changed from the server side ?

Patrick2562:
> Also this event will be triggered even if calling it with the same dimension the entity is already in whereas on(Client)ElementDimensionChanged does not:Yeah, I'm gonna fix it.

gptrk0:
> Also this event will be triggered even if calling it with the same dimension the entity is already in whereas on(Client)ElementDimensionChanged does not:

Yeah, I'm gonna fix it.

> If you want it to be cancelable then fine, do it so, but then let's create an issue for the on(Client)ElementDimensionChanged event to be cancelable too.

We can do it in a different PR, me or someone else.

> Question: What should happen if the client event is canceled when the interior is changed from the server side ?

Dimension changes, doesn't cancel it.

Pirulax:
Actually, let's not add `cancelEvent()`. The problem we run into is that we can't have a consistent behavior (with for example onClientElementDataChange).

gptrk0:
> Actually, let's not add `cancelEvent()`. The problem we run into is that we can't have a consistent behavior (with for example onClientElementDataChange).

I think I reverted everything back to the first commit.

Pirulax (approved):
LGTM. Test it, if it works I'll merge it.

gptrk0:
@Citizen01 Maybe can you test it if you'll have some time please?

PlatinMTA:
I just tested these changes, and they work. 
[test.zip](https://github.com/multitheftauto/mtasa-blue/files/6150724/test.zip)

I tried:

Using root and an element as a source ✔
Serverside with one object changing interiors every 0,5s ✔
Clientside with one object changing interiors every 0,5s + the trigger from the serverside object ✔
Triggering the event from clientside due to an serverside element ✔

It does seem to work.

Pirulax:
Okay, @Patrick2562 you can go ahead, and merge it IMO.
We'll make the cancellable later on.

PR #2068, opened 2021-02-03
Remove unused hwbrk library

botder:
That hardware breakpoint library is unused, since the code using it, has been already removed.
There is no need to keep this library, if we have no use for it.
It only increases the complexity and maintainability of the project.

# History
Added in commits 6127a699ef3d1aafdab4c5cc6101db69551eab0e and 018b09de6d828239eaaee20b3945082a5f3ffaf5
Relevant code removed in commit f7702d2c60e36af7ff7d1896d33ed1d733afc000


patrikjuvonen (approved):
I guess net doesn't use this either?

ccw808:
> I guess net doesn't use this either?

It does not use it

PR #2075, opened 2021-02-09
Password File Loaded/Read/Create

nerod2310:
Add new features to mta to support zipped password files. This will be a good option for securing models, graphics, etc.
The MTA normally reads resources packed e.g. in zip.

It's a loose proposition and I don't know if it's a good suggestion.

Make functions for reading packed data and loading them into the game. This has to be designed so that it does not burden and protects the content of the files. By downloading such a zip, the client cannot get to it and decode it. Only mta can upload this to the server if a password is provided.

Or, make your own system of password settings for files so that they cannot be modified. I know you can use raw data on models, but what about graphics and other types of files.


Lpsd:
What are the benefits of adding this, rather than simply encrypting file contents?

Also, you need to create a fork of this repo, and apply your changes to a new branch on that fork, then create the pull request using that branch (otherwise you end up with a commit history like this)...

Lpsd:
Have you actually created this functionality, or are you just requesting it? 

If you're simply requesting this, please create an issue instead of a pull request.

ghost:
@nerod2310 `I know you can use raw data on models, but what about graphics and other types of files.` You can do it as well with graphics and sounds.

patrikjuvonen:
Invalid pull request. Please use the Issues tab to create new issues.

PR #2082, opened 2021-02-11
Add nghttp2 1.43.0 and enable HTTP/2 for curl

patrikjuvonen:
## Summary
- Add HTTP/2 support for features that make use of the embedded curl library in MTA, including the MTA backend calls, [fetchRemote](https://wiki.multitheftauto.com/wiki/FetchRemote) and [callRemote](https://wiki.multitheftauto.com/wiki/CallRemote) functions
- Add [nghttp2](https://github.com/nghttp2/nghttp2) as shared library
- Enable NGHTTP2 in curl

## What is HTTP/2
A brief introduction copied from [Introduction to HTTP 2](https://medium.com/@shaved786/introduction-to-http-2-8aa8c565d008) by Mohammad Shaved:
> HTTP/2 will make our applications faster, simpler, and more robust by allowing us to undo many of the HTTP/1.1 workarounds previously done within our applications and address these concerns within the transport layer itself.
> 
> The primary goals for HTTP/2 are to reduce latency by enabling full request and response multiplexing, minimise protocol overhead via efficient compression of HTTP header fields, and add support for request prioritisation and server push.
> 
> HTTP/2 does not modify the application semantics of HTTP in any way. All the core concepts, such as HTTP methods, status codes, URIs, and header fields, remain in place which means all existing applications can be delivered without any modification.

## What's in it for us?
~~Should speed up any HTTPS calls that MTA makes to backend services, or any scripts make to HTTPS endpoints using callRemote or fetchRemote. Curl will fall back to 1.1 if HTTP/2 cannot be negotiated, so there is no loss here.~~

Turns out HTTP/2 is 10-20ms slower than HTTP/1.1 with how we're making use of it right now. Probably not a good idea to switch to it by default. Also, we're not really making use of any of HTTP/2 features by upgrade alone.

## References
- https://curl.se/docs/http2.html
- https://www.nginx.com/resources/glossary/http2/
- https://www.nginx.com/blog/http2-module-nginx/amp/
- https://developers.google.com/web/fundamentals/performance/http2
- https://kinsta.com/learn/what-is-http2/
- https://medium.com/@shaved786/introduction-to-http-2-8aa8c565d008
- https://en.wikipedia.org/wiki/HTTP/2

## Todos
- [x] Check if net code needs changes (https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html)
- [x] Enable http2 for MTA websites in nginx configs (https://www.nginx.com/blog/http2-module-nginx/amp/#config)
- [x] Add linux and macos config.h
- [x] Test server browser
- [ ] Test updater
- [x] Test callRemote
- [x] Test fetchRemote
- [ ] Test everything else :trollface: 
- [ ] Disable HTTP/2 by default


patrikjuvonen:
Closing this since there doesn't appear to be any significant benefits at this time and turns out it's 10-20 ms slower than HTTP/1.1.

PR #2083, opened 2021-02-12
Fix object position return in processLineOfSight

vitorg808:
worldModelPositionX,Y,Z is returning 0, 0, 0 for objects


Client/game_sa/CWorldSA.cpp
```diff
@@ -354,6 +354,13 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
                     pBuildingResult->usLODModelID = 0;
 
                 pBuildingResult->pInterface = targetEntity;
+                if (targetEntity->Placeable.matrix)
```
botder, line 357: You could use `Placeable.m_transform.m_translate` if there is no `Placeable.matrix`

botder (approved):

PR #2084, opened 2021-02-13
Improve player sync when exiting vehicle

### LINKED ISSUE #409: Desync whilst exiting vehicles at high speed
**Describe the bug**
Exit from vehicle at high speed causes desync.

Tested whilst both are holding ak-47s. Tried shooting+walking, just walking, and just standing (after jumping out).

Tested with @saml1er:
```
Saml1er: happens with walking for 1-2 seconds
Saml1er: then goes back to normal
Saml1er: for aiming, it's longer
qaisjp: ok, one last try, without walking
Saml1er: happens with standing as well
```

**To Reproduce**
1. Follow video http://youtu.be/ASn8AGmGEI4

 - 1.5.5 r14280

 - 1.5.5 r14273

**Additional context**
Reproduced for GitHub.
From https://bugs.mtasa.com/view.php?id=7438

Zangomangu:
Fixes #409

Before pr: https://streamable.com/1hqzd8
Currently players are unsynced while exiting a vehicle (ie. rolling out), when they finish exiting the position is corrected by player puresync. It provides an unfair advantage to the player jumping out. Latency and vehicle speed makes it worse.

After pr: https://streamable.com/rrwul7
Players are now synced while exiting.
The fix makes the exiting player send player puresync when he is exiting and actually outside the car. This involves a few changes on the server to allow this.
The driver wont be sending vehicle puresync now, so instead he starts sending unoccupied vehicle sync.

The changes aren't backwards compatible, so should be planned for a major release.


Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -514,21 +514,17 @@ void CClientPed::SetPosition(const CVector& vecPosition, bool bResetInterpolatio
         // Don't set the actual position if we're in a vehicle
         if (!GetRealOccupiedVehicle())
         {
-            // Set it only if we're not in a vehicle or not working on getting in/out
-            if (!m_pOccupiedVehicle || GetVehicleInOutState() != VEHICLE_INOUT_GETTING_OUT)
```
Zangomangu, line 541: This check was here, because when vehicle position is updated from puresync it also updates passengers, so it would have warped him to the roof when exiting. Now we receive player puresync when exiting, we need to allow it.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -368,9 +368,11 @@ void CClientVehicle::SetPosition(const CVector& vecPosition, bool bResetInterpol
     }
 
     // If we have any occupants, update their positions
+    // Make sure we dont update their position if they are getting out and have physically left the car
     for (int i = 0; i <= NUMELMS(m_pPassengers); i++)
         if (CClientPed* pOccupant = GetOccupant(i))
-            pOccupant->SetPosition(vecPosition);
+            if (pOccupant->GetVehicleInOutState() != VEHICLE_INOUT_GETTING_OUT || pOccupant->GetRealOccupiedVehicle())
+                pOccupant->SetPosition(vecPosition);
```
Zangomangu, line 383: When vehicle position is updated, we dont want to update position for peds who are exiting, because we removed the check in `CClientPed::SetPosition` (see above comment).

Client/mods/deathmatch/logic/CUnoccupiedVehicleSync.cpp
```diff
@@ -250,6 +260,10 @@ void CUnoccupiedVehicleSync::UpdateStates()
             g_pNet->DeallocateNetBitStream(pBitStream);
         }
     }
+
+    // Remove our vehicle again
+    if (pVehicle)
+        m_List.remove(pVehicle);
 }
```
Zangomangu, line 268: The reason I put this here here (vs. sending it from `CNetAPI::DoPulse`), is that it must send the packets at the correct rate as remote clients expect them to come in at 400 ms or whatever the sync rate is and apply interpolation based on that.
Maybe there's a better way to achieve this.

patrikjuvonen (approved):
Requires more of a beta test, so didn't do much testing myself, looks good, thanks.

PR #2085, opened 2021-02-13
Update mbedtls to 2.25.0

botder:
Source: https://github.com/ARMmbed/mbedtls/releases/tag/v2.25.0
Build enabled for Windows because I require the library for mongoose.


botder:
I had to remove the entry for `Shared/sdk` from `includedirs` in the root premake5.lua, because mbedtls would've otherwise included the wrong header.

sbx320 (approved):

PR #2102, opened 2021-02-27
(Re-)Add second parameter 'withExplosion' for blowVehicle

botder:
This pull request reimplements the second parameter for blowVehicle and adds that parameter to the clientside function:

```lua
bool blowVehicle(vehicle [, withExplosion = true ])
```

This pull request also changes bitstreams for explosion synchronisation, entity add for vehicles and the vehicle blow remote procedure call (rpc) to support _vehicle blow state_. Changes to the bitstreams are backwards compatible. Servers, which don't support the new bitstream, will signalise a blown vehicle always with zero (or below) health - the newer client will blow these cars on stream-in. I have tested the synchronisation briefly with @CrosRoad95.

Vehicle blow state has been added to properly synchronise the blown state of vehicles. A vehicle may be
- intact or 
- blown but waiting for explosion sync or 
- completely blown (explosion may or may not have happened).

This pull request most likely also fixes the desync for remotely synchronised vehicles repeatedly blowing up over and over again. A vehicle would blow up, then be re-created by CClientVehicle::SetHealth most likely from pure sync only to blow up again on the next streamed-in pulse ad infinitum.

A blown vehicle will always remain at zero health. A vehicle can only be restored with fixVehicle.

I have prepared a document with the packet flow and changes for vehicle blow state here:
https://github.com/botder/mtasa-documents/blob/d47ee3b9d1808570d4b890ff506bb6c81ea76518/vehicle-blow-state.txt

### DIFF
--- a/Client/mods/deathmatch/logic/CClientExplosionManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientExplosionManager.cpp
@@ -49,7 +49,7 @@ bool CClientExplosionManager::Hook_ExplosionCreation(CEntity* pGameExplodingEnti
     if (!pResponsibleGameEntity)
         return false;
 
-    CClientEntity* const pResponsible = pPools->GetClientEntity(reinterpret_cast<DWORD*>(pResponsibleGameEntity->GetInterface()));;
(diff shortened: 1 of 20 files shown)


qaisjp:
Excellent document.

PR #2110, opened 2021-03-07
Use new argument parser for generic server functions

### LINKED ISSUE #817: clearChatBox() returns false serverside if the player is root

botder:
Fixes #817 and includes #2014


Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+    {
+        return result;
+    }
+
+    return false;
+}
+
+bool CLuaGenericDefs::SetServerPassword(std::optional<std::string> rawPassword)
+{
+    std::string password = (rawPassword.has_value() ? std::move(rawPassword.value()) : std::string());
```
Pirulax, line 67:
```cpp
if (!rawPassword)
    rawPassword  = std::string{};
```

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+        throw std::invalid_argument("password must be shorter than 32 chars and just contain visible characters");
+
+    return true;
+}
+
+std::variant<std::nullptr_t, std::string_view> CLuaGenericDefs::GetServerPassword()
+{
+    if (g_pGame->GetConfig()->HasPassword())
+    {
+        return std::string_view{g_pGame->GetConfig()->GetPassword()};
```
Pirulax, line 76: `return g_pGame->GetConfig()->GetPassword()` should work implicitly.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+}
+
+unsigned int CLuaGenericDefs::GetServerHttpPort()
+{
+    return g_pGame->GetConfig()->GetHTTPPort();
+}
+
+std::string_view CLuaGenericDefs::GetServerName()
+{
+    return std::string_view{g_pGame->GetConfig()->GetServerName()};
```
Pirulax: `return g_pGame->GetConfig()->GetServerName()`

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+{
+    unsigned char colorR = 231, colorG = 217, colorB = 176;
+    bool          colorCoded = rawColorCoded.value_or(false);
+
+    if (rawColorR.has_value() && rawColorG.has_value() && rawColorB.has_value())
+    {
+        colorR = rawColorR.value();
+        colorG = rawColorG.value();
+        colorB = rawColorB.value();
+    }
```
Pirulax, line 150: This isn't the old behavior.
```cpp
argStream.ReadNumber(ucRed, 231);
argStream.ReadNumber(ucGreen, 217);
argStream.ReadNumber(ucBlue, 176);
```
This is:
```cpp
unsigned char colorR = rawColorR.value_or(231);
unsigned char colorG = rawColorG.value_or(217);
unsigned char colorB = rawColorB.value_or(176);
```

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+    std::vector<CPlayer*> sendList;
+
+    if (!rawVisibleTo.has_value() || std::holds_alternative<CElement*>(rawVisibleTo.value()))
+    {
+        CElement* element = (rawVisibleTo.has_value() ? std::get<CElement*>(rawVisibleTo.value()) : m_pRootElement);
+
+        if (IS_TEAM(element))
+        {
+            auto team = static_cast<CTeam*>(element);
+            sendList.resize(team->CountPlayers());
```
Pirulax, line 164: `.reserve`

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+            sendList.assign(team->PlayersBegin(), team->PlayersEnd());
```
Pirulax, line 165: `.insert(sendList.begin(), team->PlayersBegin(), team->PlayersEnd())`

Pirulax:
```cpp
sendList.reserve(team->CountPlayers());
sendList.insert(sendList.begin(), team->PlayersBegin(), team->PlayersEnd());
```
Your code default initializes with `resize`, we dont need that.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+        else
+        {
+            CLuaMain* luaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+            CStaticFunctionDefinitions::OutputChatBox(message.c_str(), element, colorR, colorG, colorB, colorCoded, luaMain);
+            return true;
+        }
+    }
+    else
+    {
+        sendList = std::move(std::get<std::vector<CPlayer*>>(rawVisibleTo.value()));
```
Pirulax, line 173: Perhaps add a `throw` here if `sendList.empty()`?

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+    return true;
+}
+
+bool CLuaGenericDefs::OutputConsole(std::string message, std::optional<CElement*> rawElement)
+{
+    CStaticFunctionDefinitions::OutputConsole(message.c_str(), rawElement.value_or(m_pRootElement));
+    return true;
+}
+
+bool CLuaGenericDefs::OutputChatBox(lua_State* luaVM, std::string message, std::optional<std::variant<CElement*, std::vector<CPlayer*>>> rawVisibleTo,
```
Pirulax, line 138: Let's use `std::string_view` here. Lua's string are always null terminated. What do you think?

Pirulax:
Thanks tho, nice refactor.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder:
```suggestion
    return g_pGame->GetConfig()->GetServerName();
```
Converted to a suggestion, so I can apply it directly.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder: I would rather keep it explicit, because the return type is actually a variant in this case.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder: I might be wrong, but string-convertible values won't work with `std::string_view` - needs testing.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder: It actually is the old behavior, because the colors are only used, if all three are available:
https://github.com/multitheftauto/mtasa-blue/blob/7b79541e8d9750ca0156e0f4ac4752e9ae4831e7/Server/mods/deathmatch/logic/lua/CLuaFunctionDefs.Server.cpp#L68-L75

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder:
```suggestion
            sendList.reserve(team->CountPlayers());
            sendList.insert(sendList.begin(), team->PlayersBegin(), team->PlayersEnd());
```
Converted to suggestion block :)

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
botder: We could, but ReadUserDataTable (old behavior) wouldn't show a warning in such case.

patrikjuvonen (approved):

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
Pirulax: It'll work no problem.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
Pirulax: What do you mean?
Tbh, not worth the effort, keep it like this. The performance penality is really minimal.

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
```diff
@@ -26,3 +38,153 @@ bool CLuaGenericDefs::IsTransferBoxVisible()
...
+    return false;
+}
+
+bool CLuaGenericDefs::SetServerPassword(std::optional<std::string> rawPassword)
+{
+    std::string password = (rawPassword.has_value() ? std::move(rawPassword.value()) : std::string());
+
+    if (!CStaticFunctionDefinitions::SetServerPassword(password, true))
+        throw std::invalid_argument("password must be shorter than 32 chars and just contain visible characters");
+
```
Pirulax:
```suggestion
if (!CStaticFunctionDefinitions::SetServerPassword(rawPassword.value_or(""), true))
    throw std::invalid_argument("password must be shorter than 32 chars and just contain visible characters");
```

Server/mods/deathmatch/logic/luadefs/CLuaGenericDefs.cpp
Pirulax: My bad then, sorry.

PR #2123, opened 2021-03-17
Add keybind queue to improve performance

patrikjuvonen:
## Summary
- Add a queue for keybinds and call once per frame to improve performance


Pirulax (approved):
If it works then lgtm.
Just keep in mind if it somehow manages to grow to absurd sizes (like 100s of million) then the allocated vector might be very large, perhaps add a size check, and shrink to fit if too large.

Issue #2122, opened 2021-03-17
Add spawnFlyingComponent for setVehiclePanelState


FileEX:
**Describe the solution you'd like**
Add spawnFlyingComponent argument for bumpers in setVehiclePanelState function

**Describe alternatives you've considered**

**Additional context**

Fernando-A-Rocha:
i think this is not needed due to #3592

FileEX:
> i think this is not needed due to [#3592](https://github.com/multitheftauto/mtasa-blue/pull/3592)

I replied in PR #3572

Issue #2121, opened 2021-03-15
Implement spawnFlyingComponent in a function, like makeVehicleComponentFly()


PlatinMTA:
**Is your feature request related to a problem? Please describe.**
This whole issues is arround the assumption that this function is indeed a hardcoded GTA San Andreas function and not a made one by the creators of plugin-sdk.

Altougth a vehicle can have a lot of components, there is not actual way to make most of them fly, apart from the doors, trunk and hood.

As far as I know, SpawnFlyingComponent (the GTASA function) can be used with everything, including the tires. This could be really usefull for making realistic scripted vehicle damage for example.

**Describe the solution you'd like**
Add a function like makeVehicleComponentFly(element vehicle, component?) that uses SpawnFlyingComponent.

**Describe alternatives you've considered**
None.

**Additional context**
I saw this function being used in https://github.com/DK22Pac/plugin-sdk/blob/b38ab7b07f74ac86f4283095f0386617d12fba73/plugin_sa/game_sa/CAutomobile.cpp, it looks cool and does not seem like it's really complicated to introduce. This obv. does not need to be synced.

You could make things like this: https://youtu.be/B5relHf21lo?t=71

Most of the physhics can be already calculated in game, but in order to attach the component you need to do some trickery. It's possible to recreate that plugin (the one on the video) without actually having this function, but it would be harded to do so, it probably would require more computing power and it will probably look uglier.

PR #2128, opened 2021-03-18
Add spawnFlyingComponent argument to setVehicleWheelStates()

PlatinMTA:
This pull request is related to #2121. I could work in #2122 as well but I don't promise anything.

Now ``setVehicleWheelStates`` has a new boolean argument called ``spawnFlyingComponent`` just like ``setVehicleDoorState``. If it is set to true and a wheel is changed to ``2`` then it will make the wheel visually detach from the vehicle like in this video: https://youtu.be/1zpKGbvpbnY

```lua
bool setVehicleWheelStates ( vehicle theVehicle, int frontLeft, [ int rearLeft = -1, int frontRight = -1, int rearRight = -1, bool spawnFlyingComponent = false ])
```

I would really check twice this pull before mergin it. It is my first time pulling to MTA and I could be breaking the Coding Guidelines or doing something that is not the smartest. 

Thank you!


Pirulax:
In the video I saw you had to interact with the vehicle to make it recalculate the physics. Is there a way to do that automatically?

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -433,13 +441,14 @@ void CVehicleRPCs::SetVehicleFuelTankExplodable(CClientEntity* pSource, NetBitSt
 void CVehicleRPCs::SetVehicleWheelStates(CClientEntity* pSource, NetBitStreamInterface& bitStream)
 {
     unsigned char ucWheelStates[MAX_WHEELS];
-    if (bitStream.Read((char*)ucWheelStates, MAX_WHEELS))
+    bool          spawnFlyingComponent = false;
+    if (bitStream.Read((char*)ucWheelStates, MAX_WHEELS) && (!bitStream.Can(eBitStreamVersion::setVehicleWheelStates_SpawnFlyingComponent) || bitStream.ReadBit(spawnFlyingComponent)))
```
Pirulax:
```cpp
if (!bitStream.Read((char*)ucWheelStates, MAX_WHEELS))
    return;

if (bitStream.Can(eBitStreamVersion::setVehicleWheelStates_SpawnFlyingComponent))
    if (!bitStream.ReadBit(spawnFlyingComponent))
        return;
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6577,6 +6577,7 @@ bool CStaticFunctionDefinitions::SetVehicleWheelStates(CElement* pElement, int i
 
                 CBitStream BitStream;
                 BitStream.pBitStream->Write((const char*)&pVehicle->m_ucWheelStates[0], MAX_WHEELS);
+                BitStream.pBitStream->WriteBit(spawnFlyingComponent);
```
Pirulax, line 6659: I think you need a `BitStream.Can` here as well.

Pirulax (commented):
This PR would've been easier to review if you set a default value for `spawnFlyingComponent`.
Next time please try to keep code changes to minimal.
But thanks for the PR! I dont see any glaring issues.

PlatinMTA:
> In the video I saw you had to interact with the vehicle to make it recalculate the physics. Is there a way to do that automatically?

I don't actually know. This has been default behaviour since forever, but it's inconsistent, sometimes it just falls down when other times it just mantains like that until you interact with the vehicle (even if you deleted all 4 wheels). When driving it does not happen tho.

I'll see what you reviewed in a moment, thanks.

EDIT:

> This PR would've been easier to review if you set a default value for `spawnFlyingComponent`.

What should I do next time? I didn't understand what are you referring to. Just in case the default value is ``false`` so it doesn't break backwards compatibility.

Also, shouldn't I edit the ``version.h`` file because I edited ``bistream.h``?

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
PlatinMTA: Done

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
PlatinMTA: setVehicleDoorState() does not use it so I guess not. The only way that it'll read the bit is if it's on the same version and that only depends on the server.

Proxy-99:
why this pr is not merged until now

FileEX:
> why this pr is not merged until now

I think this PR can be closed because #3592  is doing the job

Fernando-A-Rocha:
indeed this can be closed as #3592 has already been merged

PR #2129, opened 2021-03-18
Use SFixedString in CPlayer & Add string_view support to SFixedString

Pirulax:
Use SFixedString instead of `std::string`, because we are limiting the max length anyways.
I'm not quite sure whenever this is a good idea from a maintainability standpoint tho.

Briefly tested, works fine. Needs a unit test, which I'll do later. In the meantime feel free to review.

### DIFF
--- a/Server/mods/deathmatch/logic/CPlayer.cpp
+++ b/Server/mods/deathmatch/logic/CPlayer.cpp
@@ -237,14 +237,14 @@ bool CPlayer::IsSubscribed(CElement* pElement, const std::string& strName) const
 
 const char* CPlayer::GetSourceIP()
 {
-    if (m_strIP.empty())
+    if (m_IP.Empty())
     {
-        char           szIP[22];
+        dassert(m_IP.GetMaxLength() >= 22);
+
         unsigned short usPort;
-        g_pNetServer->GetPlayerIP(m_PlayerSocket, szIP, &usPort);
-        m_strIP = szIP;
+        g_pNetServer->GetPlayerIP(m_PlayerSocket, m_IP.Data(), &usPort);
     }
-    return m_strIP;
+    return m_IP;
 }
 
 uint CPlayer::Send(const CPacket& Packet)
--- a/Server/mods/deathmatch/logic/CPlayer.cpp
+++ b/Server/mods/deathmatch/logic/CPlayer.cpp
@@ -73,7 +73,6 @@ CPlayer::CPlayer(CPlayerManager* pPlayerManager, class CScriptDebugging* pScript
 
     m_bCursorShowing = false;
 
-    m_szNametagText = NULL;
     m_ucNametagR = 0xFF;
     m_ucNametagG = 0xFF;
     m_ucNametagB = 0xFF;
@@ -125,13 +124,6 @@ CPlayer::~CPlayer()
 
     // Delete the player text manager
     delete m_pPlayerTextManager;
-
-    // Destroy our nick
-    if (m_szNametagText)
-    {
-        delete[] m_szNametagText;
-        m_szNametagText = NULL;
-    }
 
     SetTeam(NULL, true);
 
@@ -542,16 +534,9 @@ void CPlayer::Reset()
 
 void CPlayer::SetNametagText(const char* szText)
 {
-    if (m_szNametagText)
-    {
-        delete[] m_szNametagText;
-        m_szNametagText = NULL;
-    }
+    m_nametagText.Clear();
     if (szText)
-    {
-        m_szNametagText = new char[strlen(szText) + 1];
-        strcpy(m_szNametagText, szText);
-    }
+        m_nametagText = szText;
 }
 
 void CPlayer::GetNametagColor(unsigned char& ucR, unsigned char& ucG, unsigned char& ucB)
--- a/Server/mods/deathmatch/logic/CPlayer.cpp
+++ b/Server/mods/deathmatch/logic/CPlayer.cpp
@@ -176,11 +176,6 @@ void CPlayer::DoPulse()
     }
 }
 
-void CPlayer::SetNick(const char* szNick)
-{
-    m_nick = szNick;
-}
-
 // Ignore min client version checks if is a custom build and both player and server have build number of 0
 bool CPlayer::ShouldIgnoreMinClientVersionChecks()
 {
--- a/Server/mods/deathmatch/logic/CPlayer.cpp
+++ b/Server/mods/deathmatch/logic/CPlayer.cpp
@@ -178,7 +178,7 @@ void CPlayer::DoPulse()
 
 void CPlayer::SetNick(const char* szNick)
 {
-    m_strNick.AssignLeft(szNick, MAX_PLAYER_NICK_LENGTH);
+    m_nick = szNick;
 }
 
 // Ignore min client version checks if is a custom build and both player and server have build number of 0
@@ -530,13 +530,6 @@ void CPlayer::Reset()
 
     m_bNametagShowing = true;
     m_usModel = 0;
-}
-
-void CPlayer::SetNametagText(const char* szText)
-{
-    m_nametagText.Clear();
-    if (szText)
-        m_nametagText = szText;
 }
 
 void CPlayer::GetNametagColor(unsigned char& ucR, unsigned char& ucG, unsigned char& ucB)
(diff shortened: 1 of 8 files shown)


Server/mods/deathmatch/logic/CPlayer.cpp
```diff
@@ -237,14 +229,14 @@ bool CPlayer::IsSubscribed(CElement* pElement, const std::string& strName) const
 
 const char* CPlayer::GetSourceIP()
 {
-    if (m_strIP.empty())
+    if (m_IP.Empty())
```
sbx320, line 227: I'd invert the logic here to reduce indentation

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -203,8 +203,8 @@ class CPlayer final : public CPed, public CClient
     bool IsCursorShowing() { return m_bCursorShowing; }
     void SetCursorShowing(bool bCursorShowing) { m_bCursorShowing = bCursorShowing; }
 
-    char* GetNametagText() { return m_szNametagText; }
-    void  SetNametagText(const char* szText);
+    std::string_view GetNametagText() { return m_nametagText; }
```
sbx320, line 206: We need to be very careful with things like this, as `std::string_view` may not be null-terminated. If other code did something like `if(strlen(pPlayer->GetNametagText()) > 42)`, it could fail.

Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp
```diff
@@ -131,20 +131,20 @@ bool CPlayerListPacket::Write(NetBitStreamInterface& BitStream) const
...
 
         // Nametag stuff
-        unsigned char ucNametagTextLength = 0;
-        char*         szNametagText = pPlayer->GetNametagText();
-        if (szNametagText)
-            ucNametagTextLength = static_cast<unsigned char>(strlen(szNametagText));
+        unsigned char    ucNametagTextLength = 0u;
+        std::string_view nametagText = pPlayer->GetNametagText();
+
+        ucNametagTextLength = (unsigned char)nametagText.length();
```
sbx320: `unsigned char ucNametagTextLength = static_cast<unsigned char>(nametagText.length());`

Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp
```diff
@@ -131,20 +131,20 @@ bool CPlayerListPacket::Write(NetBitStreamInterface& BitStream) const
...
 
         if (!BitStream.Can(eBitStreamVersion::UnicodeNametags))
```
sbx320, line 137: Is this even a 1.5 version? Otherwise we could drop this

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -660,26 +660,49 @@ namespace SharedUtil
...
 
+        constexpr SFixedString& operator=(const char* szOther)
+        {
+            Assign(szOther, MAX_LENGTH + 1);
+            return *this;
+        }
+#ifdef __cpp_lib_string_view
+        constexpr SFixedString& operator=(std::string_view other)
+        {
+            Assign(other.data(), other.length() + 1);
```
sbx320: `other.data() + other.length()+1` may not point to valid memory, for example in an empty string view.

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -660,26 +660,49 @@ namespace SharedUtil
...
+#ifdef __cpp_lib_string_view
         // Out
-        operator const char*() const { return szData; }
+        constexpr operator std::string_view() const { return { szData }; }
+#endif
+        constexpr operator const char*() const { return szData; }
+        constexpr char*           Data() { return &szData[0]; }
+
+        constexpr size_t GetMaxLength() const { return MAX_LENGTH; }
+        size_t GetLength() const { return strlen(szData); }
```
sbx320, line 748: This does not work if the string does not contain a null terminator

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -660,26 +660,49 @@ namespace SharedUtil
...
 
         // Shake it all about
         void Encrypt();
+        constexpr bool Empty() { return szData[0] == 0; }
+        constexpr void Clear() const { szData[0] = 0; }
```
sbx320: Empty should be const, but Clear is not const.

sbx320 (changes requested):
I think we might want to just replace it with `std::array<char, N>` instead. Mixing string_view with const char* is just waiting for crashes to happen due to missing null terminators.

Server/mods/deathmatch/logic/CPlayer.h
Pirulax: Well, there's no implicit conversion to `const char*`. Also I had to refactor a few functions using this method, but thats it, nothing else uses it.

Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp
Pirulax: The unicode stuff?

Shared/sdk/SharedUtil.Misc.h
Pirulax: Yeah if the string view isn't null terminated, you're right.

Shared/sdk/SharedUtil.Misc.h
Pirulax: I made sure it does.

Shared/sdk/SharedUtil.Misc.h
Pirulax: Not sure how that compiled.

Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp
sbx320: yup

Server/mods/deathmatch/logic/CPlayer.cpp
Pirulax: Less code this way.

Pirulax:
I think using `SFixedString` is better than a raw `char` array.

Pirulax:
Tbh we really should just use boost's static_string: It's a header only lib, and when used with c++17 it doesn't require any other boost header.

Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp
Pirulax:
```cpp
//
// 1.5.3 RELEASED - 2016-10-20
//

// Fix player nametag unicode characters missing on player join
// 2016-12-09 0x067 2e582453b476c1183bd9fae5363a7cffdb531834
 ```

File: Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp (hunk 4 above)
Comment (Pirulax) [reply Server/mods/deathmatch/logic/packets/CPlayerListPacket.cpp:137]:
```cpp
//
// 1.5.3 RELEASED - 2016-10-20
//

// Fix player nametag unicode characters missing on player join
// 2016-12-09 0x067 2e582453b476c1183bd9fae5363a7cffdb531834
 ```
 It's pretty recent actually.

File: Server/mods/deathmatch/logic/CPlayer.h (hunk 2 above)
Comment (Pirulax) [reply Server/mods/deathmatch/logic/CPlayer.h:206]: Resolved?

File: Shared/sdk/SharedUtil.Misc.h [hunk 8]
```diff
@@ -660,26 +660,97 @@ namespace SharedUtil
...
+        }
+
+        // Copies at most `m` characters from the array of characters pointed by `other`.
+        // Use this to assign strings of size possibly larger than MAX_LENGTH.
+        // Throws if `m > MAX_LENGTH`.
+        constexpr SFixedString& AssignAtMost(const char* other, size_t m = MAX_LENGTH)
+        {
+            if (m > MAX_LENGTH)
+                throw std::length_error{"m > MAX_LENGTH"};
+            Assign(other, strlen(other)); // Might be slow with if other is very long.
```
sbx320: This isn't working, as `m` isn't used.

sbx320 (changes requested):

Shared/sdk/SharedUtil.Misc.h
Pirulax: Should I feel embarrassed? Thanks.

Pirulax:
Possible issue: Unicode characters: The code originally supported nametag lengths up to 64 characters, which means it accepted 64 unicode characters as well, which isn't the case now.
Let's forget about this PR, I'll open one for the `SFixedString` stuff tho.

PR #2130, opened 2021-03-18
Add support for class methods in new argument parser

Pirulax:
Thanks to @sbx320 for giving me a working code which I adapted.

The code is a little bit hacky. There might be a few issues:
- the `Call` method could be improved. Not sure whenever this `const` removing stuff is the right way.
- The error messages are horrible. If there's no `Push` method for the `R`(return) type of the method, then you'll get a 10 line long error message completely unrelated to the actual error.

I have tested it, and it seems to work just fine.
Merge it with care, and the above issues in mind.

As for the usage:
```cpp
ArgumentParser<&Class::Method>
```


Pirulax:
@sbx320 could you please drop a review?

sbx320 (approved):

PR #2131, opened 2021-03-18
Add support for Pop<const char*> in new parser

Pirulax:
Add support for const char* as argument.


sbx320:
When would we use this over `std::string_view`/´std::string`?

Pirulax:
Well, rarely, but there are `CStaticFuncDef` functions  / class `methods` that take `const char*`.

sbx320:
Those should be changed then ;). 

`PopPrimitive<const char*>` seems like a bad idea in general, since it does not carry length information.

Pirulax:
That would have to be a general refactor of code to use std::string_view / string where possible.
But I'm not quite sure how that would work out, as game_sa is c++14.

PR #2132, opened 2021-03-18
Refactor CLuaVehicleDefs (server side) to use new parser only

Pirulax:
Refactor `CLuaVehicleDefs` server side to use new parser only. (Except `createVehicle`)
Also, moved all functions to be static in the source file, instead of being part of the class, this way I didn't need to `include` a lot of header files (optional, variant, vector, etc..)
Feel free to review. 
Unit test of some kind needed.


patrikjuvonen:
Builds are failing https://github.com/multitheftauto/mtasa-blue/pull/2132/checks?check_run_id=2142932339

Issue #1067, opened 2019-08-23
GTA sounds cause huge FPS drops but don't in SP


ArranTuna:
**Is your feature request related to a problem? Please describe.**
I just discovered that GTA sounds cause a huge FPS drop, thanks to being able to disable them with setWorldSoundEnabled. When I have peds shooting at me, I'd always get low FPS and I always thought it was something like the bullet effects but in fact doing:

for i=0, 44 do setWorldSoundEnabled ( i, false, true ) end

Completely eliminates the FPS drop and the FPS drop is not trivial. It was from ~100 to ~60 with 2 peds shooting at me with M4's. Doing resetWorldSounds() would make the FPS drop again.

**Describe the solution you'd like**
I've not idea if this is possible but is there some way that the sound system not cause such large FPS drops? And it's not just from shooting, '/showsound' is showing me a constant spam of about 10 sounds per second from players in the area but are so distant that none of the sounds are actually audio-able even at high volume and by disabling sounds my FPS would go from ~80 to ~100 so one solution maybe if any sounds where the volume would be so low the player can't hear them, that they'd be aborted which would at least help quite a bit.

**Describe alternatives you've considered**
I have tried cancelling sounds with:

addEventHandler("onClientWorldSound", root, function() cancelEvent() end)

To see if that would improve FPS, but it has no effect. So it must be some code before we get to onClientWorldSound that eats FPS. Could it be the reading from disk? If it is, well then the FPS hit would be way worse for other players because I'm using an SSD. Maybe it needs what ccw did for the CJ clothes, a cache of most commonly played sounds.

Because the FPS drop is so bad I will be adding settings so that players have certain spammy sounds disabled when their FPS drops below a certain range that they can adjust. I'm also going to find out what sounds are created most (I'll add the code to an example in onClientWorldSound) because there might be just a few sounds that are played so much they destroy FPS, like foot steps.

**Additional context**
I was just thinking, @saml1er found that the biggest FPS problem of all was something like "DX draw call state changes" but I'm guessing that test was done with no players running and driving around because must a debug build (and therefore no players making noises) be used for that analysis?

Einheit-101:
Oh, another thing that we may add to the "eats fps" list.

1. setElementData
̶2̶̶.̶̶ ̶̶p̶̶e̶̶d̶̶ ̶̶c̶̶o̶̶l̶̶l̶̶i̶̶s̶̶i̶̶o̶̶n̶̶s̶
3. Too many draw calls
[...]

PR #2133, opened 2021-03-18
Better performance profiling

CrosRoad95:
This idea was in my head for a long time, people uses "rescpu", popular resource to measure performance, but plain numer "10%" next to resource name does not tell the core of the cpu usage. Flag `-g` does not help too much in figuring out what cause server/client to degradate performance. Performance stats can tell you that some part of mta cause high cpu usage, but don't why, what is the excact source of the problem.

Almost everyone have a chrome or other chromium based browser which i used as a tool to display these information.
Idea is simple: make function that let you generate profile data to see them in chrome devtools:

( Based on image below ) With this amount of information, i can tell that "setElementData" in timer ( location can be seen in "summary" when i click on parent box ) cause the most cpu usage due "onElementDataChange" event, not by broadcasting.

It is very easy to use by mta developers, all you have to do is to put `CPerformanceRecorder::Sample sample("name");`, and everything is handled by you in background, you just set the name.  There are also "FunctionSample" - to take a sample of mta native function, and "EventSample" for events. Optionaly you can use `SetArg` over `sample` to add custom content with important informations in "summary" tab.

In initial pr i want to add original timings, timers, events and setElementData function.

Example data you can test yourself by drag and dropping .json file onto "performance" tab in chrome devtools
[result.zip](https://github.com/multitheftauto/mtasa-blue/files/6175700/result.zip)

test code i used

Functions:
`bool startRecordPerformance()` - start performance recording
`bool stopRecordPerformance()` - stopping, pausing performance recording
`bool clearPerformanceRecorder()` - clearing samples buffer
`bool isPerformanceRecording()` - returns true when performance is recording
`string getRecordedSamples()` - returns string json kind with recorded samples

Test code ( 2 hours of running code below, ~300 executions did not cause any crash, memory leak. ):
```lua
assert(stopRecordPerformance() == false)
assert(isPerformanceRecording() == false)
assert(startRecordPerformance() == true)
assert(isPerformanceRecording() == true)
assert(startRecordPerformance() == false)

setTimer(function()
    setElementData(root, "asdf", math.random());
end,0,0)
for i=1,10 do
    setTimer(function()
        
    end,0,0)
end

setTimer(function()
    for i=1,5000000 do

    end
end,5000,1)

setTimer(function()
    assert(isPerformanceRecording() == true)
    assert(stopRecordPerformance() == true)
    assert(isPerformanceRecording() == false)
end,10000,1)

setTimer(function() -- should not be visible in profiler
    for i=1,5000000 do

    end
end,12500,1)

setTimer(function()
    assert(startRecordPerformance() == true)
    assert(isPerformanceRecording() == true)
end,15000,1)

setTimer(function()
    stopRecordPerformance()
    local data = getRecordedSamples()
    clearPerformanceRecorder()
    assert(getRecordedSamples() == "[]")
    local f = fileCreate("result.json");
    fileWrite(f, data)
    fileClose(f)
    setTimer(function()
        restartResource(getThisResource())
    end,500,1)
end, 20000, 1)
```
Memory usage:
image

Example generated data:
image

### DIFF
--- a/Server/mods/deathmatch/logic/CPerformanceRecorder.h
+++ b/Server/mods/deathmatch/logic/CPerformanceRecorder.h
@@ -13,20 +13,35 @@ public:
     bool IsDuringPerformanceRecording() const { return m_bRecordPerformance; }
     std::string GetResult();
 
-    
     class Sample
     {
     public:
         Sample(const char* name);
         ~Sample();
 
-        void Set(const char* szKey, const char* value);
+        void SetArg(const char* szKey, const char* value);
+        void SetArg(const char* szKey, int value);
+
+    protected:
+        std::string  m_szName;
+        const char*  m_szCategory = "default";
     private:
-        const char*  m_szName;
-        TIMEUS       m_start;
+        TIMEUS       m_start = 0;
         json_object* m_pObject = nullptr;
+        json_object* m_pObjectArgs = nullptr;
     };
 
+    class FunctionSample : public Sample
+    {
+    public:
+        FunctionSample(const char* name);
+    };
+    
+    class EventSample : public Sample
+    {
+    public:
+        EventSample(const char* name);
+    };
 
 private:
     bool m_bRecordPerformance = false;
--- a/Server/mods/deathmatch/logic/CPerformanceRecorder.h
+++ b/Server/mods/deathmatch/logic/CPerformanceRecorder.h
@@ -1,0 +1,36 @@
+class CPerformanceRecorder
+{
+    friend class Sample;
+
+public:
+    CPerformanceRecorder();
+    ~CPerformanceRecorder();
+
+    void Start();
+    void Stop();
+    void Clear();
+
+    bool IsDuringPerformanceRecording() const { return m_bRecordPerformance; }
+    std::string GetResult();
+
+    
+    class Sample
+    {
+    public:
+        Sample(const char* name);
+        ~Sample();
+
+        void Set(const char* szKey, const char* value);
+    private:
+        const char*  m_szName;
+        TIMEUS       m_start;
+        json_object* m_pObject = nullptr;
+    };
+
+
+private:
+    bool m_bRecordPerformance = false;
+    TIMEUS m_start;
+
+    std::stringstream m_strStream;
+};
--- a/Client/mods/deathmatch/StdInc.h
+++ b/Client/mods/deathmatch/StdInc.h
@@ -150,6 +150,7 @@
 #include "CLatentTransferManager.h"
 #include "CDebugHookManager.h"
 #include "lua/CLuaShared.h"
+#include "profiler/CPerformanceRecorder.h"
 
 // Deathmatch includes
 #include "ClientCommands.h"
--- a/Client/mods/deathmatch/logic/CClientEntity.cpp
+++ b/Client/mods/deathmatch/logic/CClientEntity.cpp
@@ -732,6 +732,7 @@ bool CClientEntity::AddEvent(CLuaMain* pLuaMain, const char* szName, const CLuaF
 
 bool CClientEntity::CallEvent(const char* szName, const CLuaArguments& Arguments, bool bCallOnChildren)
 {
+    CPerformanceRecorder::EventSample sample(szName);
     if (!g_pClientGame->GetDebugHookManager()->OnPreEvent(szName, Arguments, this, NULL))
         return false;
 
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -733,6 +733,7 @@ void CClientGame::DoPulsePreFrame()
 
 void CClientGame::DoPulsePreHUDRender(bool bDidUnminimize, bool bDidRecreateRenderTargets)
 {
+    CPerformanceRecorder::Sample sample("CClientGame::DoPulsePreHUDRender");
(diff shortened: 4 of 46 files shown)


Server/mods/deathmatch/logic/CPerformanceRecorder.h
```diff
@@ -0,0 +1,51 @@
...
+    {
+    public:
+        Sample(const char* name);
+        ~Sample();
+
+        void SetArg(const char* szKey, const char* value);
+        void SetArg(const char* szKey, int value);
+
+    protected:
+        std::string  m_szName;
```
Pirulax: `m_name`

Server/mods/deathmatch/logic/CPerformanceRecorder.h
```diff
@@ -0,0 +1,51 @@
...
+        const char*  m_szCategory = "default";
```
Pirulax: `std::string category`

Server/mods/deathmatch/logic/CPerformanceRecorder.h
```diff
@@ -0,0 +1,51 @@
...
+    private:
+        TIMEUS       m_start = 0;
```
Pirulax: `m_startTime` or something more descriptive

Pirulax (commented):
Nice.
I'd drop using hungarian notation tho, it's useless, makes your code uglier.

Pirulax:
Also, not directly related to this PR, but I think there should be a special debug build of MTA, where all this debug stuff is enabled, because in the long run the profiler itself might cause performance issues.

ArranTuna:
performancebrowser already shows you which functions in a resource use high CPU and what MTA functions (server side only) are using what. I don't understand why anyone would need any more. It would have been time better spent improving performance like the bug that causes GTA SA sounds to use a huge amount of FPS: https://github.com/multitheftauto/mtasa-blue/issues/1067

CrosRoad95:
Take for example this chunk of code:
```
setTimer(function()
    for i=1,1000000 do

    end
end,10000,1)
```
common case when some timer internal logic cause sudden cpu spike,
performance stats even with -d option, doesn't told me that even something is wrong in "xd" resource ( i took a screenshots every time values changed ) :
image

while my approach told me exactly where is the source of cpu spike:
image

ArranTuna:
That's because there is no function name. I just tested it with a function name via runcode and it gave me:

image

```
function test()
    for i=1,10000000 do
		
    end
end
setTimer(test,1000,1)
```

AlexTMjugador:
I think that, although a Chromium-based web browser is indeed pretty universal, depending on one to show this profile data is kind of awkward. I think that a browser-agnostic solution would be better, as after all web standards exist so Chrome doesn't become the new Internet Explorer. For example, I use Firefox as my daily web browser, and having to switch to another web browser just for this is, at least, inconvenient.

I suggest generating this profile information in a format that is compatible with [profiler.firefox.com](https://profiler.firefox.com/), a web application that works on all major web browsers that allows inspecting JSON profile data. I've just uploaded your example profile there, which you can see on [this link](https://share.firefox.dev/317D6Nl), and it kinda works, so I'd say the generated profiles are in the right path at least.

Addressing @ArranTuna comments, performancebrowser is more focused on measuring the CPU usage of the Lua VM, memory usage and network usage, while this PR, as I understand it, instruments time-critical regions of MTA: SA source code to generate profile data, like `CUnoccupiedVehicleSync::DoPulse`. This allows detecting performance problems within MTA itself that are invisible to `getPerformanceStats`, and may provide insight on what exactly causes some scripting functions to be so slow, for instance.

ArranTuna:
I've tested this PR and it's, interesting.

CLuaMain::PCall - is this lua scripts executing?

I did this: crun clearPerformanceRecorder() startRecordPerformance() setTimer(function() stopRecordPerformance() setClipboard(getRecordedSamples()) end, 5000, 1)

On play at a spawn:

image

With CIT scripts running in a busy area:

image

Even though I had FPS set to unlimited there are huge gaps in the chart, is this all the stuff that doesn't have a sample added or is it GTA SA code so can't have any sampling on the gaps?

I then did CIT scripts in a quiet area (I had 300 FPS instead of 100):

image

The miliseconds breakdown doesn't seem much different even though FPS was so much different so is that because GTA SA stuff which can't be sampled being what's causing the difference?

patrikjuvonen:

Also like @Pirulax mentioned performance profiling should be enabled/compiled only for debug builds.

CrosRoad95:
> Merge conflicts must be resolved.
> 
> Also like @Pirulax mentioned performance profiling should be enabled/compiled only for debug builds.

with current knowlage, rather than enable it in debug mode only we might have a wrappers around some logics that measure performance

```cpp
public class LuaFunctions : ILuaFunctions
{
  public setElementData(...) { ... }
}

public class LuaFunctionsWithProfiling : ILuaFunctions
{
private ILuaFunctions luaFunctions;
  public LuaFunctionsWithProfiling(ILuaFunctions luaFunctions)
  {
    
  }
  
  public setElementData(...) { 
    begin("function setElementData"));
    luaFunctions->setElementData(...);
    end("function setElementData"));
  }
}

public someLuaDefs{

  public static ILuaFunctions luaFunctions = nullptr;
  public static setElementData(string key, ...)
   {
luaFunctions ->setElementData(key,...)
  }
}

CCore::SetProfilingEnabled()
{
  someLuaDefs::luaFunctions = new LuaFunctionsWithProfiling (new LuaFunctions ())
}
```

PR #2145, opened 2021-03-22
Move LuaCommon.h's lua_push* functions to lua::

Pirulax:
Moved all our custom userdata push stuff from `LuaCommon` to `lua::`. Eventually all `lua_push` should be replaced with `lua::Push` for code consistency reasons.

I originally didn't want to do the second commit in this PR, but I couldn't get it to compile otherwise (I got a ton of linker errors, probably should've moved `#pragma once` below the `luaBasic` include.)

Only review the first commit. The second commit is `find and replace all` basically.
I've tested it. (Started freeroam resource and played a little).
Doesn't require any in-depth review.


patrikjuvonen:
Stale. Merge conflicts need to be resolved.

PR #2146, opened 2021-03-22
Add lua_getowner for fast access to owner resource

Pirulax:
I dont like modifying vendor libraries, but we 've already modified Lua here and there, so this small addition won't hurt anyone.
This small change speeds up the API by at least 5%.
Ultimately we could refactor all the old code to use this function.

### DIFF
--- a/Server/mods/deathmatch/logic/luadefs/CLuaDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaDefs.cpp
@@ -10,6 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <lua/LuaCommon.h>
 extern uint g_uiNetSentByteCounter;
 
 namespace
@@ -115,8 +116,7 @@ int CLuaDefs::CanUseFunction(lua_CFunction f, lua_State* luaVM)
     }
     
     // Get associated resource
-    auto pLuaMain{ static_cast<CLuaMain*>(lua_getmtasaowner(luaVM)) };
-    auto pResource{ pLuaMain->GetResource() };
+    auto pResource{ lua_getownerresource(luaVM) };
 
     // Update execution time check
     pResource->GetVirtualMachine()->CheckExecutionTime();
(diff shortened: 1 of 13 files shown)


Pirulax:
That's actually funny. I compiled it on my machine and it worked.

botder:
You don't have to edit the Lua vendor code for this. You can add an upvalue for every function closure here:
https://github.com/multitheftauto/mtasa-blue/blob/970648124bea25d2552e5815cfe0f65eb0e61837/Server/mods/deathmatch/logic/lua/CLuaCFunctions.cpp#L111-L121

You can access these upvalues with `lua_upvalueindex`

Documentation for Lua closures:
https://pgl.yoyo.org/luai/i/3.4+C+Closures
https://pgl.yoyo.org/luai/i/lua_pushcclosure

Pirulax:
I know I can add an upvalue, but that would be the same as the current implementation: a hashtable to value mapping. (Well technically I guess more like array to value), but it probably wouldn't be as fast.
Also, this way its accessible from anywhere, even outside a function call.

Server/mods/deathmatch/logic/luadefs/CLuaDefs.cpp
```diff
@@ -10,6 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <lua/LuaCommon.h>
```
botder:
```suggestion
#include "lua/LuaCommon.h"
```
It's a regular header file

Server/mods/deathmatch/logic/luadefs/CLuaDefs.cpp
```diff
@@ -113,11 +114,9 @@ int CLuaDefs::CanUseFunction(lua_CFunction f, lua_State* luaVM)
     {
         return true;
     }
-
+    
     // Get associated resource
-    CResource* pResource = m_pResourceManager->GetResourceFromLuaState(luaVM);
-    if (!pResource)
-        return true;
+    auto pResource{ lua_getownerresource(luaVM) };
```
botder:
```suggestion
    CResource* pResource{ lua_getownerresource(luaVM) };
```
IMHO, `auto` only makes sense if you can see the type in the assignment

Pirulax:
As suggested by @botder I changed it to return a reference to indicate that it's always valid (thus no need to check for validity)

botder (approved):

PR #2150, opened 2021-03-26
Add onPlayerResourceStart event

Lpsd:
This adds a serverside event, `onPlayerResourceStart`, which is triggered when a client resource has loaded. 

With this you don't have to manually notify the server that the client is ready (for example, if you `triggerClientEvent` in `onResourceStart` for a custom event defined in the same resource, you'll receive an error about the event not being added yet, because the client scripts aren't ready).

**source**: player
**parameter**: resource

### Example

serverside
```lua
addEventHandler("onPlayerResourceStart", root, function(resource)
	outputChatBox(getResourceName(resource) .. " started for " .. getPlayerName(source))

	-- Trigger a custom clientside event defined in this resource
	if (getResourceRootElement(resource) == resourceRoot) then
		triggerClientEvent(source, "onClientCustomEvent", source) 
	end
end)
```

clientside
```lua
addEvent("onClientCustomEvent", true)
addEventHandler("onClientCustomEvent", root, function()
	outputChatBox("wow, a custom client event")
end)
```


Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -324,6 +324,20 @@ void CResource::Load()
...
         Arguments.PushResource(this);
         m_pResourceEntity->CallEvent("onClientResourceStart", Arguments, true);
+
+        NetBitStreamInterface* pBitStream = g_pNet->AllocateNetBitStream();
+        if (pBitStream)
+        {
+            if (pBitStream->Can(eBitStreamVersion::OnPlayerResourceStart))
+            {
+                // Write resource net ID
+                pBitStream->Write(this->GetNetID());
```
Pirulax:
```suggestion
                pBitStream->Write(GetNetID());
```

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -3999,6 +4005,19 @@ void CGame::Packet_DiscordJoin(CDiscordJoinPacket& Packet)
...
+void CGame::Packet_PlayerResourceStart(CPlayerResourceStartPacket& Packet)
+{
+    CPlayer* pPlayer = Packet.GetSourcePlayer();
+    if (pPlayer)
+    {
+        CLuaArguments Arguments;
+        CResource*    pResource = Packet.GetResource();
+
+        Arguments.PushResource(Packet.GetResource());
+        pPlayer->CallEvent("onPlayerResourceStart", Arguments, NULL);
```
Pirulax:
```suggestion
        CLuaArguments Arguments;
        Arguments.PushResource(Packet.GetResource());
        pPlayer->CallEvent("onPlayerResourceStart", Arguments, nullptr);
```

Pirulax: Also I think you should check if `GetResource()` a non-null value. It might happen (due to packet loss) the packet is recieved late, and by that point the resource has been deleted server side. Rare, but it could happen.

Shared/sdk/net/bitstream.h
```diff
@@ -461,6 +461,10 @@ enum class eBitStreamVersion : unsigned short
     // 2021-01-16 0x72
     SetColPolygonHeight,
 
+    // Add serverside event "onPlayerResourceStart" #0000
```
Pirulax:
```suggestion
    // Add serverside event "onPlayerResourceStart" #2150
````

**Review by Pirulax — DISMISSED** [review]

File: Server/mods/deathmatch/logic/CGame.cpp (hunk 2 above)
Comment (Lpsd) [reply Server/mods/deathmatch/logic/CGame.cpp]: And if `GetResource()` is null? Just don't call the event?

Comment (Pirulax) [reply Server/mods/deathmatch/logic/CGame.cpp]: Well, calling the event wouldn't make a lot of sense, since the `resource` argument would be nil.

**Comment by patrikjuvonen** [comment]
@Lpsd Hey, what's the status on the suggested changes?

**Comment by Lpsd** [comment]
Updated as per suggestions, tested and working. Marge?

**Review by sbx320 — APPROVED** [review]

File: Server/mods/deathmatch/logic/CGame.cpp [hunk 4]
```diff
@@ -1205,6 +1205,11 @@ bool CGame::ProcessPacket(CPacket& Packet)
             return true;
         }
 
+        case PACKET_ID_PLAYER_RESOURCE_START:
+        {
+            Packet_PlayerResourceStart(static_cast<CPlayerResourceStartPacket&>(Packet));
```
theSarrum, line 1210:
```suggestion
            Packet_PlayerResourceStart(static_cast<CPlayerResourceStartPacket&>(Packet));
            return true;
```

Lpsd: thanks

patrikjuvonen (approved):
Tested using the provided test scripts

PR #2151, opened 2021-03-26
Replace correct locale flag to zh_TW

SumiHeart:
zh_TW flag is using zh_CN flag,
try to make it correct.


Disinterpreter:
Well... You're just deleting the flag. Could you make it correct and upload valid flag for your language pool, please?

xLive:
> 
> 
> Well... You're just deleting the flag. Could you make it correct and upload valid flag for your language pool, please?

image

Disinterpreter:
> 
> 
> > Well... You're just deleting the flag. Could you make it correct and upload valid flag for your language pool, please?
> 
> image

yes, crap, sorry. Why github shows blobs so bad!?

ccw808:
The Chinese flag was originally chosen to avoid complaints from Chinese users who may regard Taiwan as a region of China

SumiHeart:
> The Chinese flag was originally chosen to avoid complaints from Chinese users who may regard Taiwan as a region of China

But in most modern games or website, Taiwan flag have been used as identification (zh_TW)

AlexTMjugador:
> The Chinese flag was originally chosen to avoid complaints from Chinese users who may regard Taiwan as a region of China

That argument can go both ways, depending on who you ask. Also, AFAIK, Taiwan is a _de facto_ sovereign country, and [polls made in 2013](https://web.archive.org/web/20130707182016/http://www.taipeitimes.com/News/taiwan/archives/2013/07/07/2003566539) show that most taiwanese people call themselves "taiwanese" rather than "chinese", so I'd say that current public perception tends to be aligned with this _de facto_ status.

> But in most modern games or website, Taiwan flag have been used as identification (zh_TW)

Anyway, I'm not chinese or taiwanese, so why is this? Is my understanding of the situation accurate? Do you think most taiwanese/chinese MTA players would agree with the change?

SumiHeart:
> > The Chinese flag was originally chosen to avoid complaints from Chinese users who may regard Taiwan as a region of China
> 
> That argument can go both ways, depending on who you ask. Also, AFAIK, Taiwan is a _de facto_ sovereign country, and [polls made in 2013](https://web.archive.org/web/20130707182016/http://www.taipeitimes.com/News/taiwan/archives/2013/07/07/2003566539) show that most taiwanese people call themselves "taiwanese" rather than "chinese", so I'd say that current public perception tends to be aligned with this _de facto_ status.
> 
> > But in most modern games or website, Taiwan flag have been used as identification (zh_TW)
> 
> Anyway, I'm not chinese or taiwanese, so why is this? Is my understanding of the situation accurate? Do you think most taiwanese/chinese MTA players would agree with the change?

Thank you, it's really helpful and important information,what you said here is really touching.

There may be a better way to make both parties acceptable: take down the flags

I understand that Taiwanese flag will cause problems for some developers, but I know that if I host a Taiwanese server today, some players will care about this, so this problem still needs to be solved.

More games language selection interfaces are also without national flags.

ccw808:
Removing the flags is a good idea. Please open an new issue

SumiHeart:
> Removing the flags is a good idea. Please open an new issue

Thank you so so much,I'll open an issue after,thank you again

PR #2153, opened 2021-03-26
Replace correct locale flag to zh_TW

SumiHeart:
zh_TW flag is using zh_CN flag,
try to make it correct.


Disinterpreter:
LGTM

image

SumiHeart:
Thank you for your helping! @Disinterpreter

PR #2164, opened 2021-03-30
Add new events onPlayerWeaponGiven and onPlayerWeapon(s)Taken

### LINKED ISSUE #2157: Add events onPlayerWeaponGiven/onPlayerWeaponTaken

I wish to know when a player has been given or taken a weapon(s) with giveWeapon/takeWeapon/takeAllWeapons functions.

I would like server-sided events such as 'onPlayerWeaponGiven' and 'onPlayerWeaponTaken' which are triggered before 'giveWeapon' or 'takeWeapon' is executed so that I can cancel the event. For both events the source is obviously a **player element** and the handler will supply a **weapon id**. I'm not quite sure how one should handle 'takeAllWeapons' but I assume you could trigger 'onPlayerWeaponTaken' for each weapon in the player's inventory.

The only alternative I could think of is using 'addDebugHook' which according the wiki description "should only be used when debugging scripts as it may degrade script performance" thus it is far from optimal. For most of my usecases I needed to prevent certain weapons from being given whenever a certain condition is met, e.g. when a player is inside a collision shape, etc. I know that for certain bespoke servers it isn't needed but I would like to keep my resources decoupled from eachother (especially freeroam) and I think this event will help with that.

**Additional context**

ricksterhd123:
When this PR is merged it should fix #2157.

I would like second thoughts on whether I should add these events for peds as well, for e.g:
onPedWeaponGiven, onPedWeaponTaken, onPedWeaponsTaken because events like onPlayerWeaponSwitch and onPedWeaponSwitch are separate but have the same function signature. Maybe they aren't needed because scripts which do ped stuff generally have checks before giveWeapon/takeWeapon anyway whereas player scripts are more impromptu.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1582,7 +1582,7 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerWeaponFire", "weapon, endX, endY, endZ, hitElement, startX, startY, startZ", NULL, false);
     m_Events.AddEvent("onPlayerWeaponGiven", "weapon", NULL, false);
     m_Events.AddEvent("onPlayerWeaponTaken", "weapon", NULL, false);
-    m_Events.AddEvent("onPlayerWeaponsTaken", "weapon", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponsTaken", "", NULL, false);
 }
 
 void CGame::ProcessTrafficLights(long long llCurrentTime)
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1580,6 +1580,9 @@ void CGame::AddBuiltInEvents()
     // Weapon events
     m_Events.AddEvent("onWeaponFire", "", NULL, false);
     m_Events.AddEvent("onPlayerWeaponFire", "weapon, endX, endY, endZ, hitElement, startX, startY, startZ", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponGiven", "weapon", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponTaken", "weapon", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponsTaken", "weapon", NULL, false);
 }
 
 void CGame::ProcessTrafficLights(long long llCurrentTime)
--- a/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -4585,6 +4585,16 @@ bool CStaticFunctionDefinitions::GiveWeapon(CElement* pElement, unsigned char uc
             CPed* pPed = static_cast<CPed*>(pElement);
             if (pPed->IsSpawned())
             {
+                bool          bIsPlayer = IS_PLAYER(pPed);
+                if (bIsPlayer)
+                {
+                    // Tell our scripts a weapon was given
+                    CLuaArguments Arguments;
+                    Arguments.PushNumber(ucWeaponID);
+                    if (!pPed->CallEvent("onPlayerWeaponGiven", Arguments))
+                        return false;
+                }
+
                 unsigned char ucCurrentWeapon = pPed->GetWeaponType();
                 if (ucCurrentWeapon != ucWeaponID && bSetAsCurrent)
                 {
@@ -4593,7 +4603,7 @@ bool CStaticFunctionDefinitions::GiveWeapon(CElement* pElement, unsigned char uc
                     Arguments.PushNumber(ucCurrentWeapon);
                     Arguments.PushNumber(ucWeaponID);
                     bool bEventRet;
-                    if (IS_PLAYER(pElement))
+                    if (bIsPlayer)
                         bEventRet = pPed->CallEvent("onPlayerWeaponSwitch", Arguments);
                     else
                         bEventRet = pPed->CallEvent("onPedWeaponSwitch", Arguments);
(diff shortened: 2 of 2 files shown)


ricksterhd123:
> I'm currently working on this to submit a PR
> 
> For giveWeapon:
> onPlayerWeaponGiven, source is a player element, function (weaponID)
> 
> For takeWeapon:
> onPlayerWeaponTaken, source is a player element, function (weaponID)
> 
> For takeAllWeapons
> onPlayerWeaponsTaken, source is a player element, function()

Just to clarify what events are added, this is a comment quoted from the issue

PlatinMTA:
Does takeAllWeapons() trigger the event by the number of weapons taken or only once?

It should only do it once, you could send a table of weapons taken OR a string saying "all". If you are going the table route, it should always send a table, even when its only one weapon. This could come in handy for knowing what function triggered the event.

ricksterhd123:
> 
> 
> Does takeAllWeapons() trigger the event by the number of weapons taken or only once?
> 
> It should only do it once, you could send a table of weapons taken OR a string saying "all". If you are going the table route, it should always send a table, even when its only one weapon. This could come in handy for knowing what function triggered the event.

There's an event called "onPlayerWeaponsTaken", which is triggered before takeAllWeapons() So if you want to know which weapons are taken then you can simply use getPedWeapon and iterate through all weapon slots. There's no need for any params in the event because all weapons are taken, I think it's also cancelable too.

PlatinMTA:
> There's an event called "onPlayerWeaponsTaken", which is triggered before takeAllWeapons() So if you want to know which weapons are taken then you can simply use getPedWeapon and iterate through all weapon slots. There's no need for any params in the event because all weapons are taken, I think it's also cancelable too.

A use case for knowing which weapons are taken in takeAllWeapons() is to log them using outputServerLog() or any other log system you have in place. It could have some utility apart from logging stuff.

ricksterhd123:
> 
> 
> > There's an event called "onPlayerWeaponsTaken", which is triggered before takeAllWeapons() So if you want to know which weapons are taken then you can simply use getPedWeapon and iterate through all weapon slots. There's no need for any params in the event because all weapons are taken, I think it's also cancelable too.
> 
> A use case for knowing which weapons are taken in takeAllWeapons() is to log them using outputServerLog() or any other log system you have in place. It could have some utility apart from logging stuff.

One of the reasons I added this functionality in the first place is that many servers use the default resource 'freeroam'. Whenever you want to add new gamemodes that restrict certain weapons then you will have to use either: exports, element data or hardcode to check if a certain weapon given to the user is restricted in the gamemode that the user is playing which IMO is a sin. I would rather have an event handler in the gamemode than edit unrelated resources causing unnecessary dependencies/coupling.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1580,6 +1580,9 @@ void CGame::AddBuiltInEvents()
     // Weapon events
     m_Events.AddEvent("onWeaponFire", "", NULL, false);
     m_Events.AddEvent("onPlayerWeaponFire", "weapon, endX, endY, endZ, hitElement, startX, startY, startZ", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponGiven", "weapon", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponTaken", "weapon", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponsTaken", "", NULL, false);
```
Inder00, line 1585: All event should be generalized and simple as possible. A „onPlayerWeaponsTaken” should be removed and replaced with general event - „onPlayerWeaponToken” to avoid more spaghetti code.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4657,6 +4667,15 @@ bool CStaticFunctionDefinitions::TakeWeapon(CElement* pElement, unsigned char uc
             // Just because it's the same slot doesn't mean it's the same weapon -_- - Caz
             if (pPed->IsSpawned() && pPed->GetWeapon(ucWeaponSlot) && pPed->GetWeaponType(ucWeaponSlot) == ucWeaponID)
             {
+                if (IS_PLAYER(pPed))
+                {
+                    // Tell our scripts a weapon was taken
+                    CLuaArguments Arguments;
+                    Arguments.PushNumber(ucWeaponID);
```
Inder00, line 4674: Weapon slot and amount of ammo would be nice

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4585,6 +4585,16 @@ bool CStaticFunctionDefinitions::GiveWeapon(CElement* pElement, unsigned char uc
             CPed* pPed = static_cast<CPed*>(pElement);
             if (pPed->IsSpawned())
             {
+                bool          bIsPlayer = IS_PLAYER(pPed);
+                if (bIsPlayer)
+                {
+                    // Tell our scripts a weapon was given
+                    CLuaArguments Arguments;
```
Inder00, line 4592: Weapon slot and amount of ammo would be nice

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1580,6 +1580,9 @@ void CGame::AddBuiltInEvents()
     // Weapon events
     m_Events.AddEvent("onWeaponFire", "", NULL, false);
     m_Events.AddEvent("onPlayerWeaponFire", "weapon, endX, endY, endZ, hitElement, startX, startY, startZ", NULL, false);
+    m_Events.AddEvent("onPlayerWeaponGiven", "weapon", NULL, false);
```
Inder00, line 1583: Generalized event - onPedWeaponGiven

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1580,6 +1580,9 @@ void CGame::AddBuiltInEvents()
...
+    m_Events.AddEvent("onPlayerWeaponTaken", "weapon", NULL, false);
```
Inder00, line 1584: Generalized event - onPedWeaponTaken

Inder00 (changes requested):
Events should be generalized to „ped” type and clientside would be also implemented.

Fernando-A-Rocha:
PR abandoned

botder:
If anyone wants to pick up the work and continue it, feel free to.

Issue #1972, opened 2021-01-02
Unsafe functions


Disinterpreter:
**Describe the solution you'd like**

From: https://github.com/multitheftauto/mtasa-blue/pull/1937#issuecomment-752962847 https://github.com/multitheftauto/mtasa-blue/discussions/1782 we decided to allow unsafe functions only with ACL rights.

I'd like to start implementing new functions from ACL fixes.

In the first step we deny all unsafe functions
When we will be sure that these functions blocked by default, we can start to implement ones.

I suppose we should mode current os.* functions under ACL too.

**Additional context**

Put something like this in Default (check me with methods like "file:close")
```xml
<right name="function.require" access="false"></right>
<right name="function.dofile" access="false"></right>
<right name="function.loadfile" access="false"></right>
<right name="function.loadstring" access="false"></right>

<right name="function.os.execute" access="false"></right>
<right name="function.os.exit" access="false"></right>
<right name="function.os.getenv" access="false"></right>
<right name="function.os.remove" access="false"></right>
<right name="function.os.rename" access="false"></right>
<right name="function.os.setlocale" access="false"></right>
<right name="function.os.tmpname" access="false"></right>

<right name="function.io.close" access="false"></right>
<right name="function.io.flush" access="false"></right>
<right name="function.io.input" access="false"></right>
<right name="function.io.lines" access="false"></right>
<right name="function.io.open" access="false"></right>
<right name="function.io.output" access="false"></right>
<right name="function.io.popen" access="false"></right>
<right name="function.io.read" access="false"></right>
<right name="function.io.stderr" access="false"></right>
<right name="function.io.stdin" access="false"></right>
<right name="function.io.stdout" access="false"></right>
<right name="function.io.tmpfile" access="false"></right>
<right name="function.io.type" access="false"></right>
<right name="function.io.write" access="false"></right>
<right name="function.file:close" access="false"></right>
<right name="function.file:flush" access="false"></right>
<right name="function.file:lines" access="false"></right>
<right name="function.file:read" access="false"></right>
<right name="function.file:seek" access="false"></right>
<right name="function.file:setvbuf" access="false"></right>
<right name="function.file:write" access="false"></right>


<right name="function.package.cpath" access="false"></right>
<right name="function.package.loaded" access="false"></right>
<right name="function.package.loaders" access="false"></right>
<right name="function.package.loadlib" access="false"></right>
<right name="function.package.path" access="false"></right>
[...]

PR #2169, opened 2021-04-01
Add JavaScript as an alternative to Lua

CrosRoad95:
# Admision:
The goal of this project is to __add__ javascript programming language to mta sa, lua going to be untoutched. Everything what will used by both lua and js will turn into generic class to support both or more languages. I want as much as possible avoid duplication of current.

# General idea
1. Use js objects as much as possible, less arguments but packed into "logical units", classes ( eg dxDrawText would accept 3 arguments: `text, rect ( two vector2 ), style object`.
2. Module importing as a way to get access to mta api, other scripts, other resources.
3. Async await over callbacks, only exception would be filter functions
4. Clean up api - eg use Color object for color everywhere.
5. Safer by default, disabled dangerous features such a "eval" function until you explicit allow it, long execution protection. Memory usage limits.
6. Enum objects over raw numbers, strings ( etter for typescript )
7. More branched functions - eg instead of "getElementBonePosition" you're doing "player.getBone(Bone.Root).position", 
8. Don't use hidden variables at all, Event object over variadic list of arguments.

# Typescript
Yes, it will compatible with typescript. If possible i would like to make mta compilation to generate typescript definition as well or generate them in different way using source code.

# Initial pull request
~~This pull request will implement only two classes "crypt" and "utils", related features, promisses, new parser support ( same as lua ) Vector2/3/4 Matrix, Color, classes. Nothing more, just simple __working__ version of javascript.~~
function "print" to print things, "version" to see a current version of v8. Nothing else

# V8
Because of fakt that compiling javascript is not trivial task i decided to use precompiled libraries for each platform with debug symbols as a part of this project along with instructions, configuration i used to compile it.

It is possible to download precompiled from google server: https://storage.googleapis.com/chromium-v8/official/canary/v8-win32-rel-8.9.276.zip but they are shipped with flag `use_custom_libcxx` set to true. It may work but it adds extra layer of complexity to mta project itself ( eg. forces to use https://releases.llvm.org/ ). 

Current version of v8 is: 8.9.255.20 ( week, two ago it was last stable based on https://omahaproxy.appspot.com/ ), same version is used currently by chrome browser ( check your version at: chrome://version/ )

# Project assumptions
To communicate with v8 library i created shared project "MtaV8". Goal of this project is to implement exact same logic for client and server. Client Deathmatch and Deathmatch ( server side ) are using interface to communicate.
Functions and more what can be accessed from js will declared in previously mentioned projects.

Style of programming ( mostly function names, arguments ) will simillar to lua, but i'm taking better examples of solving problems from node.js, chrome browser.

# Test project
In newly created folder "tests" i put test console project `v8tests` copy pasted from <https://github.com/v8/v8/blob/master/samples/shell.cc>

# JIT
Supported and enabled. If function is "hot and stable" it will optimized and recompiled. That functions are almost as fast as c++. There is a feature to trace what function are JITt'ed ( leaved for future update ).

# Performance
Only raw runtime will benefit, loop without usage of mta api will execute much much faster than lua.

# New configuration options in mtaserver.conf ( Moved into future update )
```xml
<javascript>
    <eval>0</eval> <!-- 0 - disabled completly, 1 - allowed when acl explicit allow -->
    <execution_time_limit>2000</execution_time_limit> <!-- How long js can execute at once before get terminated -->
    <thread_pool_size>4</thread_pool_size> <!-- Threads pool size used to resolve promises -->
</javascript>
```

# Future:

### Webassembly
Disabled right now, may be added in future updates :) .

### Multithreading
Yes, but per resource. Topic for future update. But if you use "await passwordHash(...)" it will computed on background thread.
I also did an investigation and each resource / javascript isolate can run parallel.
It is possible to make resource with exported heavy functions such a path finding and use `await someResource.findWay(...)`

### Exports
I did not tested it, but i'm almost 100% sure you can export/import everything ( including functions, objects ect ) across resources. Topic for future update. Client server commucation will support only primitive types.

### Communication with lua
Possible as well, for future update. I though about form where you can import lua files as a regular javascript module ` import { context from "foo.lua"` and use methods, properties over "context" to set/get/call lua functions.

Todo:

- [ ] Precompile binaries for linux and macos

If anyone have a suggestion, question, ask.

Test resource:
[js.zip](https://github.com/multitheftauto/mtasa-blue/files/6253257/js.zip)


Pirulax:
### The problem with parallel resources ...
is that, they will call MTA functions from different threads as well, so they would need to be guarded by a mutex, as MTA doesn't support MT.
Anyways. Thanks for listening on feedback, and trying to make the PR as small as possible.
Builds will fail here most likely, as dlls and stuff is missing.
Also, I presume V8's performance is better with clang's std, but yeah, using it here would be a pain, so let's leave it for a future PR.

### I recommend to create a new branch...
in `mtasa-blue` like `v8js`, since merging incomplete stuff into master doesn't make a lot of sense.
Eventually, we could start distributing this special build (as a custom version, so no support for release mode servers) to make testing easier.

### I think native JS is a refreshment...
for most people, including myself, especially because of typescript. Theres nothing wrong with Lua, it's awesome, but right now, because our past choices, I dont see us updating it / adding JIT anytime soon.
Also JS had a much bigger ecosystem than Lua.

### A few things we have to discuss
#### API layer?
We could continue to use `ArgumentParser` but rewritten for JS and our own hand-rolled class stuff:
- Pros: Could reuse existing C++ side API functions with `ArgumentParser`.
- Cons: We would need to write class handling, and all that for JS, which will take a lot of time, and needs to be maintained, while, on the other hand, there are libs like v8pp out there. 
#### Precompiled scripts?
I'd say we really should ditch using precompiled scripts. It lead to issues with Lua (we're stuck in the past, since Lua versions arent really bytecode compatible).
`cache=false` is good enough for most people probably. 
#### Why disable eval?
As far as I know `eval` (by itself) can't do anything evil (pun intended), just like Lua's `loadstring` is harmless. Just make sure arbitrary bytecode can't be run.
#### About the API design
I agree is a lot of stuff we need to rethink and do differently to Lua. But we should try to keep the API pretty alike, to make people's life easier.

Pirulax:
Also It seems like you change a lot of code.
Could you please add explain what is what?
New classes, files, etc..

CrosRoad95:
multithreading mta api - if resource has enabled multithreading, imports may be changed a little, allow only thread safe functions.
Wrapping classes in not a big deal, its kinda easy because all i'm doing is to set internal field with pointer to function and value, i already did it and working well. It may be even worse to use 3-rd party library for it.
> Precompiled scripts?

i'm wondering if it is possible to make server compile client side code and send that byte code to client to run.

> Builds will fail here most likely, as dlls and stuff is missing.

i'm working on command to copy dlls properly, on windows everyting compiling fine everything fine, only post-build command failed

> Why disable eval?

top 1 command used to destroy someone else property, if you by default can't use that function, you less likely to be attacked using this function. You still can enable it by giving acl right.

Pirulax:
> i'm wondering if it is possible to make server compile client side code and send that byte code to client to run.

Only way we could do this is to send the script to MTA servers right before sending it to the client.
We can't trust the server owner's server compiling anything, and sending it to the client.

> top 1 command used to destroy someone else property, if you by default can't use that function, you less likely to be attacked using this function. You still can enable it by giving acl right.

maybe im missing something, but eval just executes regular JS code, doesn't it?
just like `loadstring` in lua

jushar:
When this was discussed in the past, the decision always has been against introducing a second scripting language. Before you invest too much time into this feature, I'd recommend you to first make sure that this generally has the chance to be merged.

As far as I remember, the reasons were:
* it's pretty slow, because of many context switches between JS and C++ (which are much more expensive in comparison to Lua) --> at least needs some benchmarking to disprove/weaken this theory
* a second language results in a significant community fragmentation and complicates documentation and scripting help
* if V8 is really going to be introduced, the chance should be used to clean up the scripting API and remove all inconsistencies, deprecations and unsafe functions instead of creating a copy of the Lua API
* is it worth all the effort?

CrosRoad95:
> it's pretty slow, because of many context switches between JS and C++ (which are much more expensive in comparison to Lua) --> at least needs some benchmarking to disprove/weaken this theory

few months ago i made benchmark, you can see results here: https://forum.mtasa.com/topic/129052-javascript-as-a-secondary-programming-language/?do=findComment&comment=993033 

and other benchrmark with vector class, and they were 2-3 times faster than lua

jushar:
Your code is not suitable as a benchmark, because it's entirely written in the target scripting language. Therefore, it can drastically improve from using JIT. This is however far away from the real world scenario: In MTA you rarely do expensive calculations in the scripting language. Most of the time, you're just calling functions that are part of the MTA API which results in lots of context switches.

I suggest you add a couple of functions (like dxDrawText) and repeat the benchmark with calling these.

Pirulax:
I'm not sure what's the correct approach in this scenario, but I suppose discussing stuff in the [original discussion thread](https://github.com/multitheftauto/mtasa-blue/discussions/2008) would be better in terms of organization.

websharik:
### MTA required upgrade to JS!

Lua and XML is realy very dead (no sugar, prerip updates, rip community, even no class spec in 2021 🤦‍♂️ "gaming language").

JS and json is so much rapidly developing community (JS is web - web never die, js is very simple, realy cool updates every... month? (imports, promises), [package manager with 1M+ packages](https://npmjs.com) (example - AI, ML (can be used to make smart bots in MTA (but its just a "web language")))).

Problem of this pull is:
1) Trying refactoring MTA (1, 2?, 4, 7, 8). I dont say that is bad, but the main topic of this pull is adding support JS.
2) Try copy of the "Lua API" as JS module ?

**I suggest just start with realy MINIMAL interference of current.**


### About adding support V8
My version without dublicating - bridge ```JS (V8) <=> Bridge (JS Module|C++ gate) <=> C++ (MTA SA)```

Working example as [**NativeScript (JS (V8) <=> NativeScript (JS module|Java gate) <=> Java (MobilePhoneNativeMethods))**](https://docs.nativescript.org/native-api-access.html)

Prototype code of bridge working:
```js
import * as MTA, Player from "MTANative"

MTA.Chat.local.print('Hello world') //local chat print Hello world
let pos = Player.getByNickname('LionG').position //get another player pos
console.log(pos); //debug log pos Vector(0, 0, 14) or array {0, 0, 14}
let marker = new MTA.Elements.Marker(pos) //create marker to pos
marker.color = 'red' //set marker color
setTimeout(async (marker) => {
  await marker.destroy() //wait for marker destroing
  console.log('marker was destroyed'); //then debug
}, 4000, marker) //destroy marker before 4 seconds
```
MTANative is just proxy and retranslate all calls to native metods in C++!

### The most important feature of bridge:
1. One Native documentation - N languages (C++, Python, GO and more easy can be added to support).
2. No fragmentation community and help (only language specs)
3. I think it be so faster becouse nearer native.

Mb some security problem with full acces to native ? Resolve:
1. WhiteList to limit can be accessible objects.
2. BlackList to block must be unaccessible objects.

AlexTMjugador:
@websharik While I get your love for JavaScript and its ecosystem, and in general I feel that JavaScript could be an interesting addition to MTA, I felt there were some things that were, at least, so confusing in that comment that I should respond.

> Lua and XML is realy very dead (no sugar, prerip updates, rip community, even no class spec in 2021 🤦‍♂️ "gaming language").

XML is still the markup language with query and transformation languages that were standarized by the W3C, like XPath, XSLT and XSD. It also has namespaces and other cool features. There are equivalents for those in JSON, but they're far from being as polished as in XML, and most of those are heavily inspired by XML standards. While MTA may not ever need such things, the fact that XML is mature and isn't updated often isn't bad at all. It doesn't need to, and using stable technologies means less effort towards maintaining them.

A similar situation happens with Lua, IMO (although Lua is still actively developed; the latest version was released on March 29, 2021, and the fact that MTA doesn't update Lua since version 5.1 doesn't mean that Lua is dead). Lua also has [package management](https://luarocks.org/), and is a pretty satisfactory language to work with. Sure, LuaRocks is not as big as npm, and Lua has some unconventional quirks, like the "do it yourself" OOP and that indexes start at 1, but it is still a pretty good contender that can get you pretty far. Besides, Lua is used as a scripting language in all sorts of applications that are active and used by millions of people, like Roblox, TeamSpeak, Angry Birds, VLC media player, Wireshark, the LÖVE game engine (used by games like Move or Die)... and they are not making the switch to any other scripting language not only because they're trapped by their past decisions, but also because Lua is good for an embedded language. There were proposals and PRs to get Lua modules working with MTA, which would address many of the concerns of software reuse stated here, providing ways to import libraries and such.

> My version without dublicating - bridge JS (V8) <=> Bridge (JS Module|C++ gate) <=> C++ (MTA SA)

While this is interesting, this is already also what Lua and this PR does, because it's the way to do it. MTA Lua API functions are C functions that call native C++ code to perform its work. In other words, the Lua API that MTA exposes to scripts already is the "proxy" to call native code you're talking about. You can't get much more closer to native code than that, and using a library that is specific to JavaScript to perform this "binding" automatically doesn't solve much problems IMO.

> No fragmentation community and help (only language specs)

I think Jusonex was precisely thinking on the fragmentation that different languages, with different "specifications", would bring to the scripting community. Let's see this with an example: you're a newbie scripter in MTA, and for some reason you chose to learn to script for MTA in JS. Good for you, but when you ask for help on the forums, maybe you get some answers telling you "I'd do that like this in Lua, which is the scripting language I know best". And then you have to translate that to JS, but, as you are a newbie, you are not sure how to do it. You check the wiki with the hope of finding indications along the lines of "this Lua code is equivalent to this JS code" or that both languages offer a similar API, but then find that the documentation in either the Lua or the JS version of the API is worse, and you're left wondering while your JS code will really be equivalent to the Lua code you had until you test it. In the future, the newbie scripter may either adopt the language that they see more dominant, because it'd lead to less cognitive effort translating code between languages, learn both languages, or stick with one of them and contribute to the generation of "scripting language silos". As you can see, the fragmentation that comes with the choice of scripting language is significant, and should be addressed properly before rolling out a feature like this in MTA.

Disinterpreter:
Doesn't care Lua or JS if you can't escape the MTA "VM".

> package manager with 1M+ packages 

Ok.
image

What is the price of bringing this *clever* community?

websharik:
@Disinterpreter, clever community 🤣 "Чья бы корова мычала". [https://luarocks.org/modules/daelvn/isodd](https://luarocks.org/modules/daelvn/isodd)

*clever humanity*

websharik:
@AlexTMjugador, Who want xml, lua, pain - use xml, lua. Who want json, js, sugar - there should be an opportunity do it. JLM.

Lua has very lot of "handmade required" + MTA limitation (example: i cnow about luarocks, but more packages using "require" ("require" disabled in mta and for use package have to again make "handmade" (example - compile to single file)))
Stable is good but latest JS sugar is too stable. Backward compatibility is care of language devs, not game devs. MTA resource developers should easy develop business logic without worrying about such basic things as 'how to make handmade "class"'. Developers comfort - this is what distinguishes modern languages from Lua

Lua + MTA not supported by IDE. Example: How to debug variable in MTA ? - restart resource and looks to debug minichat - its looks as joke.
JS is simple debuging in IDE via remote debug + sugar (break points,  and more...)

C++ not my lang... i dont know how MTA works now. I just see bridge as resolve some problem as nativescript.

My things is Lua best for devs who want thouch history, make handmade "class" and 100 bicycle, stepped on a 100 rake.
For real cool projects and comfort learning with best help by comunity - required modern language. + JS is "web lang", any begginers easy can go to web after JS MTA (Lua beginners after Lua MTA is deadlock) (4600 JS vs 70 LUA vacancy by [hh.ru](hh.ru) now).

LosFaul:
@websharik
you wont be able to debug anything **Element** related either in any IDE in javascript
unless you make your own handcrafted plugins for MTA
and elements pretty much play the largest part in MTA


the only reason for me to support this pr would be because of webassembly
since it allows to write code in pretty much any language that can compile to webassembly
otherwise i would have no interest in this pr at all
i personally have no interest in using javascript at all


also there is one very important aspect in this pr:
**security**
specially client side must be at all time as safe as possible since malicious people could otherwise do bad things on clients

and this is the very big advantage of Lua it's very easy to safely sandbox it
if you can't do that with a language for scripting it will never find a place in MTA

and this is were javascript engines doesn't have the greatest reputation specially with exploits like spectre and meltdown in recent years

Disinterpreter:
> 
> 
> @Disinterpreter, clever community 🤣 "Чья бы корова мычала". https://luarocks.org/modules/daelvn/isodd
> 
> _clever humanity_

This repo hasn't contain serious context, moreover it was created by fun (look at the license) an it has 20 downloads versus 400k


> **security**
> specially client side must be at all time as safe as possible since malicious people could otherwise do bad things on clients

The same must be on the server side as we remember. Or we need to remove some of restrictions for server side lua (that's impossible)

cpuschma:
> The same must be on the server side as we remember. Or we need to remove some of restrictions for server side lua (that's impossible)

I think the administration and security part should not be ignored either. The KISS principle and the low overhead in the configurations makes MTA relatively easy to maintain and manage. The fact that you can specify which ports are to be used and how many slots the server has with just three start arguments is very handy.

If the implementation of JavaScript using V8 should really bring a benefit with its one million packets, then certain restrictions have to be lifted -- which would contradict the principle mentioned above and would have to be implemented cleanly again, to give the providers something to work with. One would have to think about a concept like Deno in my opinion, that certain functions are available only if they were explicitly permitted.

Disinterpreter:
> 
> 
> > The same must be on the server side as we remember. Or we need to remove some of restrictions for server side lua (that's impossible)
> 
> I think the administration and security part should not be ignored either. The KISS principle and the low overhead in the configurations makes MTA relatively easy to maintain and manage. The fact that you can specify which ports are to be used and how many slots the server has with just three start arguments is very handy.
> 
> If the implementation of JavaScript using V8 should really bring a benefit with its one million packets, then certain restrictions have to be lifted -- which would contradict the principle mentioned above and would have to be implemented cleanly again, to give the providers something to work with. One would have to think about a concept like Deno in my opinion, that certain functions are available only if they were explicitly permitted.

Change the MTA Team mind...
Otherwise we won't able to move on.
https://github.com/multitheftauto/mtasa-blue/discussions/1782
and THIS https://github.com/multitheftauto/mtasa-blue/issues/1972

P.S You know, I'm not against to see JS in MTA. I'm just trying to transparent the MTA Team's vision on the problem overall. All of this depends of MTA security policy, if you want to see good JS implementation in MTA, you must solve this in Lua.

websharik:
@Disinterpreter,
> This repo hasn't contain serious context, moreover it was created by fun (look at the license) an it has 20 downloads versus 400k

Most repos of lua hasn't contain serious context. Your not should use every repos from npm in your prod, if your can make better or lighter and you have time to do it (best - send pulll to "bad" repo to make it greate again, its save time for next devs). When  your making  MWC (minimal working product) or don't feel like writing a bike  some as that repos can be usefull + this approach will help to split up the code and remove duplication functional inside project between diff packages, its make project more light. All perfect.
Different with downloads just show diffirent with community sсales.
NPM have selft troubles but there are everywhere.

@LosFaul,
> security
specially client side must be at all time as safe as possible since malicious people could otherwise do bad things on clients

I think we need to see browsers. Any browser using JS and its safe (they limit some modules, add rules to use it (mic, cam, geo, push), and add self scripts (window object, DOM...)).
V8 can be: Node environment (full access), Browser environment (limited)... we can add self MTA Resource environment (can be based of browser environment (limit some modules (example: fs - set root dir as resource root per resource), add scripts (MTA bridge object))). Resourse will be conteinered in V8 MTA Resource environment, and cant do bad things on clients and server.

Spectre and meltdown was fixed and touch not only JS becouse they based on CPU caching (MMU). Your can google about that for more info...

V8 have included "Inspector protocol" for remote debuging. Lua... i hope too. 
MTA required just expose debuging ports on [setDevelopmentMode](https://wiki.multitheftauto.com/wiki/SetDevelopmentMode) enabled + config ports in mta.conf or meta.xml item to parralel debuging 2+ resources + mb whitelist of developers ips to more safe with default 127.0.0.1 (to allow only local debuging).

@xNevo, The KISS stay will be. You dont should know all server parameters to start server, new parameters should be default with safe values.

Disinterpreter:
@websharik 

If we wanna talk about the place of Lua in the world, we can it do, of course, but it's better to move it in Discussions.

> Most repos of lua hasn't contain serious context. 

Just a lie, one of the best examples of using Lua here: https://habr.com/ru/post/321864/
and here https://www.highload.ru/moscow/2018/schedule more than 5 reports about lua in highload, good choice!
Lua still not a legacy language. But the popularity of lua is low, it is true.

Lua has many cool and modern modules for: postgres, kafka, discord, cassandra, even clickhouse and of course Tarantool (which uses Lua too). But anyway as I've told before. We should solve the "secure" problem in serverside before we can discuss about modules.

Anyway, if you want full power of Lua or JS on your serverside, you need to change MTA Teams' mind.

ERAGON007:
> Lua has many cool and modern modules for: postgres, kafka, discord, cassandra, even clickhouse and of course Tarantool (which uses Lua too). But anyway as I've told before. We should solve the "secure" problem in serverside before we can discuss about modules.

btw, you CAN'T just use that libs and this means:

many powerfull/good libs for lua + You're not allowed to use them = there is no libs

websharik:
@ERAGON007,
Ofcourse - i cant use any lib, but not all libs.

Now i use [TypeScriptToLua](https://github.com/TypeScriptToLua/TypeScriptToLua) to complite TS (and js) to lua. And many libs from npm can be converted too and nice working (tested). But its so raw project and has issues and some issues cant be resolved for lua5.1 ex: [Promise tests down](https://github.com/TypeScriptToLua/TypeScriptToLua/issues/1105).

Now i have no desire to make so projects/resources for MTA, becouse current version MTA SA has very lot mistakes/traps for developers.  Any my idea was trapped with "MTA disallows" etc... old lua version (because old luacs will breacks), "limitation" lua (as disable "require"), "limitation" browser (as [local/remote working mods](https://github.com/multitheftauto/mtasa-blue/issues/2407), "security...security...security" security is good but MTA is very secured (pentagon is envy), and more more any... Some big projects partical resolve with custom launchers, i not ready learn c++ to make custom luancher to fix MTA Team mistakes to nice developming resources.

I waiting MTA team reaction for GTA SA Remasted with hope for MTA Remasted with at least latest Lua5.4/JIT and maybe i come back to MTA developing.

ERAGON007:
hmm, that's not a bad idea

btw you know that converting/transpilling that JS libs into Lua code actually can't be faster than a code that is purely written in Lua.

security in MTA is much better than similiar opponents (like SA:MP, FiveM, IV:MP) BUT this security shouldn't be some bunch of limitations for developers (this disables usage of so many libraries)

problems for using ONLY Lua for developing resources is you can't use a debugger so you have to restart the resource you wanna debug and just put some `print` somewhere and get back to console/game screen to see what's happened

and you have to do that for all your needed variables! a real pain

and we can't properly use OO features in Lua

opponents like FiveM do have more than one language to their dev API's  and leaves the developer itself to choose between what you can and want to use. for example you can use class library project from `.NET`, you can use `Lua` projects (they use Lua 5.3), you can also use JS natively too!

but in mta things are different and almost all of them because of security.

every day we see people in mta community suffer from lack of threads/classes/Intellisense from their IDE while developing

my idea is to merge this PR when it's ready for TESTS and just put an option in `mtaserver.conf` for server devs/owners to turn usage of JS on or off (just like turning OOP on or off in resources)

and the rest of bugs/features will be reported by people they use JS on MTA

Dutchman101:
Even though that for various reasons it's already highly unlikely for MTA to ever natively support another scripting language (with natively i mean like, look at what [Slipe - C# scripting in MTA](https://prod.mta-slipe.com/) accomplished without MTA changing things around.. a transpiler or something like Slipe doesn't have to be native, someone can make it for Javascript as well), i also want to mention additional concerns.

MTA invested a lot of AC team manpower into modifying Lua related stuff and integrating anti-cheat closely with Lua implementation in order to defend against various exploits, Lua injection methods and such. However, this was done years ago, over the span of years, by the lead AC developer (ccw) that no longer has enough free time to do that all over for another language. We also currently don't have a talented developer like him to do it instead, such a developer must obviously be part of AC team/MTA team and have affinity and skills for it.

So what im trying to say, the reality nowadays means implementing a new language comes with a huge security penalty. There's probably going to be some bollocks like "Java code injector", for example, and even if we patch the initial methods it will take us years to catch up to the same level of security as we have grown for Lua, over many years.

Not to speak about other obstacles like years of scripting documentation/tutorials, wiki entries and such being dedicated to Lua in MTA. Even more reasons to make adding another scripting language a difficult task to accomplish were written in here, but even more so in the tons of discussions within our community/forum topics about this matter. Lua is also not bad at all for MTA, you can see that MTA barely has any limits to creativity as we've had full game conversions being made by people (the unimaginable) and it really does suffice for flexibility, features and possibility.. the real reason behind suggestions to replace Lua, or add a second language, is that different people have different preferences and experience with other languages, and are thus opinionated and easy to say using Lua for MTA was a mistake.

Just learn lua, master it, then learn to love it.. and you'll get as far in MTA scripting as you want to imagine. But i dont expect MTA to make certain sacrifices, and sadly we can't keep everyone happy.

As a side note to @CrosRoad95, it will cause much less disappointment to first get the discussion/decision and outlook on a suggested feature in order before spending so much time on starting to implement things like this. I mean, for yourself, in case you see this PR as wasted effort.

**Edit:** check out [Typescript for MTA](https://mtasa-typescript.github.io/), it was sent by NanoBob (the author of Slipe, C# in MTA) when asked about doing something similar as he did but then for Javascript. So apparently, someone already did that, but we don't know about the results with it yet, or if it's as feature & documentation-complete as Slipe is.

Toliak:
Thank you for mentioning [MTASA TypeScript project](https://github.com/mtasa-typescript).

### Few words about it

> then for Javascript

MTASA TypeScript does not use JavaScript in MTASA runtime. It is based on language transpiling and [TypeScriptToLua](https://typescripttolua.github.io/). So, the pipeline (for user) is being like:
```
Writing TS ---> Compiling it into Lua ---> Run in MTASA
```

> results with it

There is:
- [boilerplate](https://github.com/mtasa-typescript/resource-boilerplate) -- anyone can use to create resources,
- [types](https://github.com/mtasa-typescript/mtasa-lua-types) -- TypeScript types definitions for MTASA.

> documentation

I have made every effort to make **types** fully-compatible with MTASA Wiki documentation.
Installation of the boilerplate seems to be easy (install NodeJS, run [one command](https://github.com/mtasa-typescript/resource-boilerplate#-create-new-project-initialize-boilerplate) inside `resources` directories).

#### Features

- YML meta file ([schema](https://github.com/mtasa-typescript/mtasa-lua-utils/blob/master/mtasa-meta.schema.json)) ([example 1](https://github.com/mtasa-typescript/mtasa-lua-utils/blob/master/src/tests/resources.spec/importFunctionFromAnotherResource/mtasa-meta.yml), [example 2](https://github.com/mtasa-typescript/mtasa-lua-utils/blob/master/src/tests/resources.spec/multipleSimpleResources/mtasa-meta.yml)),
- Multiple resources support (with ability to export function between resources),
- `debug.traceback` with source mapping (provided by TSTL),
- Meta file linting (cannot use not existing script),
- Script side linting (error if used server-side function in a client-side script, cannot import non-exported function from another resource),
- Types [linting & hinting features](https://github.com/mtasa-typescript/mtasa-lua-types#-features),
- Premade configurations for [VSCode](https://github.com/mtasa-typescript/resource-boilerplate/tree/master/.vscode) and [WebStorm](https://github.com/mtasa-typescript/resource-boilerplate/tree/master/.idea).

### Few caveats about it

- **types** are not updated to the latest MTASA version (they are updated using [MTASA Wiki parser](https://github.com/mtasa-typescript/mtasa-wiki-parser), but I haven't applied it to the new version yet)
- There are no examples (except mini-examples in [e2e tests](https://github.com/mtasa-typescript/mtasa-lua-utils/tree/master/src/tests/resources.spec))
- Leak of documentation for YML Meta file, Script writing

### Conclusion

I'm open to discuss the project in [the project discord](http://discord.gg/67vxSwhRT4) or in [the issues](https://github.com/mtasa-typescript/resource-boilerplate/issues). If you have any questions, ideas or suggestions, you are welcome.

Disinterpreter:
> problems for using ONLY Lua for developing resources is you can't use a debugger so you have to restart the resource you wanna debug and just put some `print` somewhere and get back to console/game screen to see what's happened
> 
> and you have to do that for all your needed variables! a real pain

https://github.com/Jusonex/MTATD
https://wiki.multitheftauto.com/wiki/AddDebugHook

????

ERAGON007:
So the logic is : "Don't touch it if it's working"

using Lua forever itself is a sacrifice, you know it sucks.

MTA teams main goal is to provide a nice security right? that's nice but not if devs get limited

and now imagine i wanna use "PostgreSQL" as an alternative for "MySQL/MariaDB", i must use C++ for it because of these limitations

at least you could've given Lua library a bit more permissions for ONLY server-side. this is not the Lua people worldwide using. we can't get help from global Lua community - can't use libraries and etc...

and the libraries that are usable in mta, you'll need some time to put all of it in a single Lua file and sometimes people give up (including myself)

At least give MTA users some respect and let them choose to turn a feature like using JS `on` or `off`


image
image

this is what happens to MTA features over time

using a language that's not popular enough = lesser people will contribute = no maintenance = project dies over time

just like what happened to MTA's almost dead feature (Modules)

Toliak:
> and now imagine i wanna use "PostgreSQL" as an alternative for "MySQL/MariaDB"
i must use C++ for it because of these limitations

Not only C++. You can implement your own backend and use REST API to communicate 😃

Disinterpreter:
> and now imagine i wanna use "PostgreSQL" as an alternative for "MySQL/MariaDB", i must use C++ for it because of these limitations

What's difference between JS and Lua if you cannot use denied libraries like in Lua?
https://luarocks.org/search?q=postgres
You think you will? Sorry, but policy works with Lua and future JS.
The problem that we can't use some libraries it not in Lua. it is in restrictions.

> using Lua forever itself is a sacrifice, you know it sucks.

Say it to thousand developers of Roblox, please

Abasalt-yar:
> > and now imagine i wanna use "PostgreSQL" as an alternative for "MySQL/MariaDB"
> > i must use C++ for it because of these limitations
> 
> Not only C++. You can implement your own backend and use REST API to communicate 😃

But using it as module is a bit different I guess 🤔, right?

**Edit**: Not only Postgres even mongo can be awesome

ERAGON007:
> Say it to thousand developers of Roblox, please

If we're going to do that, we need to tell Fivem devs too right? btw it's a similiar project to mta

https://docs.fivem.net/docs/support/server-debug/
VS
image

i don't think people like to restart a resource a thousand times to debug it.

even module sdk on github repo is outdated :)

Dutchman101:
> MTA teams main goal is to provide a nice security right? that's nice but not if devs get limited

I see you want to push a certain sentiment, but from my last reply it was already clear that the security aspect is not the reason for a second language not to be implemented.. of which the reasons for it not being the (main) reason, should also be cleared up in my previous post. My reply was actually prompted by yours (you mentioning a lot about "security limiting developers" - the sentiment) to give some clarity about that aspect, but never was it ever among the main obstables towards implementing a second language..  however, having mentioned those concerns as well is good for the sake of completion during this discussion. The main and other real obstacles were given in here, and throughout many other discussions for "add another scripting language to MTA" in the community throughout the last couple years.

No one mentioned anything regarding security/AC being yet another potential obstacle, before the discussion had already practically ended in "Too many problems with going ahead and adding another scripting language". Btw, your sentiment isn't very pleasurable.. please stop acting like MTA is deliberately limiting creativity/possibilities of developers, or specifically doing anything (serving "one of the main goals" as you implied) and taking that for granted as a consequence.

Also, about the rest of your comments from the same post as my quote:

- some pics of individual projects aren't representative for anything, SDK like any project hangs on contributor interest and the need for updates. Be the change you want to see and open PR's.

- the proper way to get started is to first learn Lua (e.g Lua PIL), and the adaptability of that knownledge towards MTA is high. However, most MTA users that want to get into scripting will learn it from the MTA stack through the available tutorials and docs (from the beginning), and your concern about a lack of help isn't correct: for MTA specific stuff there is a lot of scripting support (1-on-1 help from other people) and docs available in MTA community and wiki.

However @ERAGON007, to keep this discussion useful & prevent it from getting derailed into an argument/word againt someone else's word, please talk about the actual and main obstacles towards implementing another scripting language, that negative sentiment you hold by now has gotten its due attention and answers.

Also note that the oversight reply of MTA team in this issue ([here](https://github.com/multitheftauto/mtasa-blue/pull/2169#issuecomment-813077012)) can be useful to understand some of the obstacles.. including to read the one after it, and better also read all other ones given in here and past discussions about adding another scripting language.

Disinterpreter:
> 
> 
> > Say it to thousand developers of Roblox, please
> 
> If we're going to do that, we need to tell Fivem devs too right? btw it's a similiar project to mta
> 
> https://docs.fivem.net/docs/support/server-debug/ VS image
> 
> i don't think people like to restart a resource a thousand times to debug it.
> 
> even module sdk on github repo is outdated :)

As I said before, MTA gives you
https://github.com/Jusonex/MTATD
https://wiki.multitheftauto.com/wiki/AddDebugHook

You try to compare JS with thousand libraries with restriced mta lua. Give me pure lua and I can work with postgres,mongo, etc... not a problem, really. Need you JS without libraries? I guess not, but that's a MTA's policy.

Yes SDK really outdated. MTA don't pay programmers for the refactoring.

TheNormalnij:
> As I said before, MTA gives you
> https://github.com/Jusonex/MTATD
> https://wiki.multitheftauto.com/wiki/AddDebugHook

Im working on MTA debugger, you can find this extension in VS Code.
This debugger really increases my performance during developing.
https://github.com/TheNormalnij/MTATD

ERAGON007 (approved):

Lpsd:
As a note, that last approval by @ERAGON007 was a mistake 😅

Dutchman101:
Today there was a discussion on dev discord, that i think is valuable for putting a few things into perspective, should CrosRoad95's large PR's like this & Bullet Physics (BP) engine turn out to not get merged. I can very well relate to devs that are hestitant, and dont want to, as they say, burn their fingers on it

The discussion: [https://imgur.com/a/seD0hqQ](https://imgur.com/a/seD0hqQ)

PR #2170, opened 2021-04-01
New optional inputBlocked argument for showChat

Pieter-Dewachter:
A lot of people like to hide the chatbox so they can play without any distractions (eg. during a clanwar, tournament, ...). However there is also the muscle memory of typing commands into the chatbox, or quickly asking a referee for a "free" round. Unfortunately when hiding the chatbox, this also means all input to the chatbox is disabled (messages, but also commands etc.).

This PR adds an extra 'inputBlocked' argument to the existing showChat functions (client- and serverside) which allows to hide the chatbox messages but still keep input to the chatbox possible. The default for this argument is set to true so that existing scripts don't break by this change, but new scripts can implement it if they choose to do so. Also the bitstream changes should be backwards-compatbible as far as I'm aware.


Client/core/CChat.h
```diff
@@ -166,8 +166,9 @@ class CChat
     void         SetDxFont(LPD3DXFONT pDXFont);
 
     bool IsVisible() { return m_bVisible; }
-    void SetVisible(bool bVisible);
-    bool IsInputVisible() { return m_bVisible && m_bInputVisible; }
+    void SetVisible(bool bVisible, bool bInputBlocked = true);
+    bool IsInputBlocked() { return m_bInputBlocked; }
```
Lpsd:
```suggestion
    bool IsInputBlocked() const { return m_bInputBlocked; }
```

Client/core/CChat.h
```diff
@@ -166,8 +166,9 @@ class CChat
...
+    bool IsInputVisible() { return !m_bInputBlocked && m_bInputVisible; }
```
Lpsd:
```suggestion
    bool IsInputVisible() const { return !m_bInputBlocked && m_bInputVisible; }
```

Lpsd (changes requested):
Just a couple of things. Looks good though, thanks for the PR!

Pieter-Dewachter:
> Just a couple of things. Looks good though, thanks for the PR!

Good catch, I included your suggestions!

Client/core/CChat.cpp
```diff
@@ -771,17 +776,26 @@ bool CChat::CharacterKeyHandler(CGUIKeyEventArgs KeyboardArgs)
...
+    {
+        m_bInputBlocked = bInputBlocked;
+
+        if (m_bInputBlocked)
+            SetInputVisible(false);
+    }
+    else
+    {
+        m_bInputBlocked = false;
+    }
```
Lpsd: What if you want to show the chat but block input?

If you do `showChat(true, true)`, the chat is shown but you're still able to type. I don't see any reason why we *shouldn't* allow that.

Lpsd:
https://github.com/multitheftauto/mtasa-blue/pull/2170/files#diff-d462bfda71f5c318c95adfb0db198ba8bca363376ba084a902fe506f375f236cR779

Both `SetVisible` and `SetInputVisible` methods are currently kind of weird, especially this bit:

```cpp
    if (m_bInputBlocked)
        bVisible = false;
```

It's almost like it's trying to do the job of the Lua function definition code?

The code should be changed to:

```cpp
void CChat::SetVisible(bool bVisible, bool bInputBlocked)
{
    m_bVisible = bVisible;
    SetInputVisible(bInputBlocked);
}

void CChat::SetInputVisible(bool bVisible)
{
    if (!bVisible)
    {
        ClearInput();
        ResetHistoryChanges();
    }

    m_bInputVisible = bVisible;
}
```

We should then upgrade the Lua definitions for `showChat` on both the client and server to use the new argument parser and make use of `std::optional`, where we can set the `inputBlocked` argument to true by default if the chat is being hidden, and false if it is being shown.

This will probably cause side-effects elsewhere in the code, where it assumes that the default value of the `bInputBlocked` argument to `CChat::SetVisible` is automatically handled, based on whether the chat is visible or not. For example, in the `CChat` constructor:

```cpp
m_pInput->SetVisible(false);
```

Would now mean the chat is hidden, but input is allowed (which is not the correct behaviour). So it would become this instead:

```cpp
m_pInput->SetVisible(false, true);
```

Shouldn't be hard to fix all of those.

Pieter-Dewachter:
I have fixed the merge conflicts, suggestions by @Lpsd were added in a previous commit as well.
Would it be possible to have another review?

Lpsd (approved):

Lpsd:
completely forgot about this. thanks for your PR. marged

~~@Pieter-Dewachter please update https://wiki.multitheftauto.com/wiki/ShowChat when you get a second~~ 
done

patrikjuvonen:
Please document isChatInputBlocked on Wiki: https://wiki.multitheftauto.com/index.php?title=IsChatInputBlocked&action=edit&redlink=1

Pieter-Dewachter:
> Please document isChatInputBlocked on Wiki: https://wiki.multitheftauto.com/index.php?title=IsChatInputBlocked&action=edit&redlink=1

Done!

PR #2174, opened 2021-04-04
Improve server console input handling

CrosRoad95:
This improvment makes server fps does not affect console typing speed. If you make by accident resource that cause server to hang, you can easier type command to restart/stop resource that cause the problem.

Here's how it works in action:
https://www.youtube.com/watch?v=uJ5BSSbjtjc&feature=youtu.be

### DIFF
--- a/Server/core/CServerImpl.h
+++ b/Server/core/CServerImpl.h
@@ -93,8 +93,8 @@ private:
     SString m_strServerPath;
     SString m_strServerModPath;
 
-    bool m_bRequestedQuit;
-    bool m_bRequestedReset;
+    mutable std::atomic<bool> m_bRequestedQuit;
+    bool                      m_bRequestedReset;
 
     wchar_t m_szInputBuffer[255];
     uint    m_uiInputCount;
@@ -107,6 +107,7 @@ private:
     std::vector<std::vector<SString>> m_vecCommandHistory = {{"", ""}};
     uint                              m_uiSelectedCommandHistoryEntry = 0;
 
+    std::thread m_threadConsoleInput;
 #ifdef WIN32
     HANDLE    m_hConsole;
     HANDLE    m_hConsoleInput;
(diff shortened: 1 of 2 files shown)


Pirulax:
Now I wont need to wait 10s to paste in random code? Nice.

CrosRoad95:
> Now I wont need to wait 10s to paste in random code? Nice.

not really, if you hang server by accident, you no longer need to wait eg 2 seconds to type 1 characted what is really frustrating

sbx320 (changes requested):
There is no synchronization in here, so it's just a race condition waiting to happen. 

This is not how you do proper multithreading.

Server/core/CServerImpl.h
```diff
@@ -93,8 +93,8 @@ class CServerImpl : public CServerInterface
     SString m_strServerPath;
     SString m_strServerModPath;
 
-    bool m_bRequestedQuit;
-    bool m_bRequestedReset;
+    mutable std::atomic<bool> m_bRequestedQuit;
```
Pirulax, line 96: its `mutable` by default, isn't it? Also I don't think it needs to be atomic at all, since you wont be writing to it the same time from different threads.

AlexTMjugador: If my notions from concurrency classes and other programming languages translate to this case in C++, atomic is just only useful if several threads need to do several operations on a value that need not have any effect on any other thread until they all finish. So it's only useful for patterns like incrementing counters, for read-update-write operations.

In the code that this PR touches, only the `load` method is used, which is equivalent to a read operation. If `quit` is never changed in a read-update-write fashion somewhere else, `atomic` is indeed useless and introduces unnecessary atomicity. It'd be enough to guarantee _visibility_, i.e. that other threads will indeed see the updated value, and not that they'll use a stale copy that happens to be in the L1 data cache for a CPU core or something.

PR #2180, opened 2021-04-09
Use int instead of size_t as index in CLuaFunctionParser as it use negative indices

Pirulax:
The code uses negative indices here and there, and the compiler was really annoyed about it. (Since it had to cast -1 to size_t).
Briefly tested, but I see no reason for it not to work.


Pirulax:
Apparently signed integers have faster math on CPU level as well. Who would've thought

sbx320 (dismissed):
Looks good

PR #1723, opened 2020-10-16
Add onClientPlayerWeaponSwitch for remote players

Woovie:
This brings in the changes by @correaAlex in #1143. Created on behalf of @majqqq

PR #2181, opened 2021-04-10
Add onClientPedWeaponSwitch

### LINKED ISSUE #1141: onClientPlayerWeaponSwitch should trigger for remote players
onClientPlayerWeaponSwitch doesn't trigger when weapon slot is changed for remote players.

It should trigger also for remote players. If possible.

At the moment, there's no possibility to detect changes for weapon slot, the only possibility is using loop. (for render usage f.e)

**Additional context**
It would be very nice, if we could detect those changes on client-side by event provided by MTA.
I think this is possible to do, since onClientPlayerWeaponFire triggers for local/remote player.

patrikjuvonen:
Copy of #1723 (deleted repository)

WIP


sbx320 (commented):
Wouldn't this potentially cause a lot of network traffic for high player counts?

It might be better to only send it to streamed in players

PR #2192, opened 2021-04-13
Fix #2178: testLineAgainstWater works incorrectly outside of game boundaries

### LINKED ISSUE #2178: testLineAgainstWater works incorrectly outside of game boundaries
**Describe the bug**
 `testLineAgainstWater` checks only for water at z-coordinate 0 outside of the -3000 : 3000 coords range. This behaviour was introduced along with multiple other fixes with this PR: https://github.com/multitheftauto/mtasa-blue/pull/568
 Some time afterwards, possiblity to change ocean water level outside of bounds was added for `setWaterLevel` function, thus rendering current behaviour incorrect.

**To reproduce**
Set ocean water level to something different than 0, use testLineAgainstWater to check for it, game still thinks the water is at z-coord 0

**Expected behaviour**
`testLineAgainstWater` should check for water at correct height

**Version**
current

**Additional context**
Fix is very simple, the offending line is: https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/game_sa/CWaterManagerSA.cpp#L842
The fix is to change current logic to 
`vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, <oceanWaterLevelHere>), &zeroPoint)`
where `<oceanWaterLevelHere>` is currently used z-coord of ocean water

Allerek:
Just added a check for water level outside Game Area, and commened out a check if both points are outside Game Area(why would we check for that lol).

Fixes https://github.com/multitheftauto/mtasa-blue/issues/2178

### DIFF
--- a/Client/game_sa/CWaterManagerSA.cpp
+++ b/Client/game_sa/CWaterManagerSA.cpp
@@ -841,15 +841,15 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
     CVector oceanWaterIntersectPoint;
 
     float waterHeight = *(float*)0x6E873F;
-    
-    //This checks for intersections between Line and Plane(water is treated like a flat surface) outside a Game Area and returns a CVector with its coordinates if any were found.
+
+    // This checks for intersections between Line and Plane(water is treated like a flat surface) outside a Game Area and returns a CVector with its coordinates if any were found.
     if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, waterHeight), &oceanWaterIntersectPoint) && IsPointOutsideOfGameArea(oceanWaterIntersectPoint))
     {
         *vecCollision = oceanWaterIntersectPoint;
         return true;
     }
 
-    //Early out in case of both points being out of map
+    // Early out in case of both points being out of map
     if (IsPointOutsideOfGameArea(vecStart) && IsPointOutsideOfGameArea(vecEnd))
     {
         // Check if both points are on the same side of the map, in case of some mad person
--- a/Client/game_sa/CWaterManagerSA.cpp
+++ b/Client/game_sa/CWaterManagerSA.cpp
@@ -848,7 +848,7 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
         return true;
     }
 
-    //Commented out, why would we check if both points are outside of the game area? That should'nt change a thing. (Issue 2178)
+    //Commented out, why would we check if both points are outside of the game area? That shouldn't change a thing. (Issue 2178)
 
     // Early out in case of both points being out of map
     //if (IsPointOutsideOfGameArea(vecStart) && IsPointOutsideOfGameArea(vecEnd))
--- a/Client/game_sa/CWaterManagerSA.cpp
+++ b/Client/game_sa/CWaterManagerSA.cpp
@@ -841,14 +841,15 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
     CVector oceanWaterIntersectPoint;
 
     float waterHeight = *(float*)0x6E873F;
+    
     //This checks for intersections between Line and Plane(water is treated like a flat surface) outside a Game Area and returns a CVector with its coordinates if any were found.
     if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, waterHeight), &oceanWaterIntersectPoint) && IsPointOutsideOfGameArea(oceanWaterIntersectPoint))
     {
-       *vecCollision = oceanWaterIntersectPoint;
-       return true;
+        *vecCollision = oceanWaterIntersectPoint;
+        return true;
     }
 
-     //Early out in case of both points being out of map
+    //Early out in case of both points being out of map
     if (IsPointOutsideOfGameArea(vecStart) && IsPointOutsideOfGameArea(vecEnd))
     {
         // Check if both points are on the same side of the map, in case of some mad person
@@ -858,7 +859,7 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
             (vecStart.fY < -3000.0f && vecEnd.fY < -3000.0f) ||
             (vecStart.fY > 3000.0f && vecEnd.fY > 3000.0f))
         {
-        return false;
+            return false;
         }
     }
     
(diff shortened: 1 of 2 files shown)


forkerer:
The commented out check is an early out if both ends are on the same side of the world border, thus skipping all the further calculations, i think that the GetZonesIntersecting return zones at the world border even when both points are ourside of the -3000 : 3000 range
Red arrows show check that will be skipped, while the black one will go through
obraz

forkerer:
I don't see a reason to remove that check, and i doubt you tested those cases or checked what happens in the code further down

Allerek:
> I don't see a reason to remove that check, and i doubt you tested those cases or checked what happens in the code further down

Because thats what this PR is all about, if both of the ends will be outside area, it will automatically return false, we dont want that to happen, we want to be able to check if line is hitting the water in any case, and yeah - i tested that and it worked just fine.

Some users might want to get out of Game Area(for some reason, i'm not here to judge), and we should allow that.

forkerer:
And they still can, the code you commented out doesn't change anything about that, it just skips unnecessary calculations later on. No water other than ocean water will ever be collided outside of -3000 : 3000 range because You can't create custom water outside that

forkerer:
Check for the GetZonesIntersecting behaviour when you check line starting from (-3050, -3050, 10) to (3050, -3050, 10) when ocean water level is 0, it will do unnecessary checks around the world border if I'm not mistaken after You commented the code out

Allerek:
> And they still can, the code you commented out doesn't change anything about that, it just skips unnecessary calculations later on. No water other than ocean water will ever be collided outside of -3000 : 3000 range because You can't create custom water outside that

Yeah, but the code returns false, it doesnt even check if the line really collides with  water, as far as i see it only  checks for X and Y.

-Are both points outside of map?
-If yes then check if they're on the same side of the map
-If so, just return false for some reason? < That doesnt make any sense, whats the reason to suggest theres no water in its way what soever?

forkerer:
You literally check for ocean water collision few lines up, no other water will ever exist there, as it can't be created

forkerer:
That's the only water you can encounter outside of game zone
obraz

Allerek:
After thinking it out, you might be right, further code checks for waves as i see? Forgot the ocean water doesnt have waves at all, so that can make sense. Getting commented code back, leaving the higher fix tho.

Allerek:
After all the checks are done, i guess we're ready to merge, as the change itself worked well in my tests.

Client/game_sa/CWaterManagerSA.cpp
```diff
@@ -839,12 +839,16 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
 
     // Check if we're outside of map area.
     CVector zeroPoint;
-    if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, 0), &zeroPoint) && IsPointOutsideOfGameArea(zeroPoint))
+    float   waterHeight;
+    CVector pvecUnknown;
```
Pirulax: Try passing nullptr for `vecUnknown`. It should work (judging from the source code..)

Allerek: I'll try that. Waiting for tests to pass.

Pirulax: I mean the CI will pass, just test it locally, see if everything is as expected.

Allerek:
[watertest.zip](https://github.com/multitheftauto/mtasa-blue/files/6351383/watertest.zip)
Test resource.
image
(first number-current water level, second is endz for testLineAgainstWater)

Allerek:
For some reason it returns true if water level is lower than 0, and we check at Z=0(or lower), i guess its due to
`GetWaterLevel(CVector(3000,3000,1), &waterHeight, true, nullptr);`
Where i check for Z level '1', i guess we could just set it to a very low number, but this aint the best solution i guess. Any ideas? @forkerer

patrikjuvonen:
> For some reason it returns true if water level is lower than 0, and we check at Z=0(or lower), i guess its due to
> `GetWaterLevel(CVector(3000,3000,1), &waterHeight, true, nullptr);`
> Where i check for Z level '1', i guess we could just set it to a very low number, but this aint the best solution i guess. Any ideas? @forkerer

Did you check his suggested fix in the original issue #2178?

Allerek:
> > For some reason it returns true if water level is lower than 0, and we check at Z=0(or lower), i guess its due to
> > `GetWaterLevel(CVector(3000,3000,1), &waterHeight, true, nullptr);`
> > Where i check for Z level '1', i guess we could just set it to a very low number, but this aint the best solution i guess. Any ideas? @forkerer
> 
> Did you check his suggested fix in the original issue #2178?

i got some help, i'll introduce it soon but for now i have some work to do

Client/game_sa/CWaterManagerSA.cpp
```diff
@@ -838,10 +838,14 @@ bool CWaterManagerSA::TestLineAgainstWater(const CVector& vecStart, const CVecto
...
+
+    float waterHeight = *(float*)0x6E873F;
+
+    // This checks for intersections between Line and Plane(water is treated like a flat surface) outside a Game Area and returns a CVector with its coordinates if any were found.
+    if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, waterHeight), &oceanWaterIntersectPoint) && IsPointOutsideOfGameArea(oceanWaterIntersectPoint))
     {
-        *vecCollision = zeroPoint;
+        *vecCollision = oceanWaterIntersectPoint;
         return true;
     }
```
Pirulax, line 850:
```suggestion
    // Check for intersection with ocean outside the game area (takes water level into account)
    // If a hit is detected, and it is outside, we early out, as custom water can't be created outside game boundaries
    {
        CVector intersection{};
        const float waterHeight = *static_cast<float*>(0x6E873F);
        if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, waterHeight), &intersection))
        {
            if (IsPointOutsideOfGameArea(intersection))
            {
                *vecCollision = intersection;
                return true;
            }
        }
    }
```

Pirulax (changes requested):
LGTM

Client/game_sa/CWaterManagerSA.cpp
Pirulax:
```suggestion
    // Check for intersection with ocean outside the game area (takes water level into account)
    // If a hit is detected, and it is outside, we early out, as custom water can't be created outside game boundaries
    {
        CVector intersection{};
        const float waterHeight = *static_cast<float*>(0x6E873F);
        if (vecStart.IntersectsSegmentPlane(rayDir, CVector(0, 0, 1), CVector(0, 0, waterHeight), &intersection))
        {
            if (IsPointOutsideOfGameArea(intersection))
            {
                *vecCollision = intersection;
                return true;
            }
        }
    }
```

Pirulax (approved):
Well I hope you've tested it.

PR #2193, opened 2021-04-15
An option to flip UV on material lines

tederis:
In my experience there is a lot of circumstances when material line used to draw all kind of ropes, tapes and other rectangular textures. And it tends to be accompanied by difficulties related to a line UV orientation. This PR is indended to simplify this task and brings a new optional argument to flip UV orientation in some cases.

[BEFORE] Instead of writing a shader to rotate a texture:
```Lua
local tapeTex = dxCreateTexture( "tape.png")
local shader = dxCreateShader(
    [[
        texture Tex0;

        float3x3 getTextureTransform()
        {
            return float3x3(0, 1, 0, -1, 0, 0, 0, 0, 1);
        }

        technique simple
        {
            pass P0
            {
                Texture[0] = Tex0;
                TextureTransform[0] = getTextureTransform();
                TextureTransformFlags[0] = Count2;
            }
        }
    ]]
)
dxSetShaderValue( shader, "Tex0", tapeTex )

addEventHandler( "onClientRender", root,
    function()
        local TAPE_LEN = 5
        local x0, y0, z0 = 0, 0, 5
        local x1, y1, z1 = TAPE_LEN, 0, 5

        local tw, th = dxGetMaterialSize( tapeTex )
        local usize = TAPE_LEN / ( 1 * ( tw / th ) )

        dxDrawMaterialSectionLine3D( x0, y0, z0, x1, y1, z1, 0, 0, tw, usize * th, shader, 1 )
    end
)
```

[AFTER] Using this PR you can simply:
```Lua
local tapeTex = dxCreateTexture( "tape.png")

addEventHandler( "onClientRender", root,
	function()
	local TAPE_LEN = 5
        local x0, y0, z0 = TAPE_LEN, 0, 5
        local x1, y1, z1 = 0, 0, 5

        local tw, th = dxGetMaterialSize( tapeTex )
        local usize = TAPE_LEN / ( 1 * ( tw / th ) )

        dxDrawMaterialSectionLine3D( x0, y0, z0, x1, y1, z1, 0, 0, usize * tw, th, true, tapeTex, 1 )
    end
)
```

The syntax:
`bool dxDrawMaterialLine3D ( float startX, float startY, float startZ, float endX, float endY, float endZ, [bool flipUV,] element material, int width [, int color = white, [ bool postGUI = false, ] float faceX, float faceY, float faceZ ] )`

`bool dxDrawMaterialSectionLine3D ( float startX, float startY, float startZ, float endX, float endY, float endZ, float u, float v, float usize, float vsize, [bool flipUV,] element material, int width, [int color = white, [ bool postGUI = false, ] float faceX, float faceY, float faceZ ] )`

### DIFF
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -844,7 +844,7 @@ void CGraphics::DrawLine3DQueued(const CVector& vecBegin, const CVector& vecEnd,
 }
 
 void CGraphics::DrawMaterialLine3DQueued(const CVector& vecBegin, const CVector& vecEnd, float fWidth, unsigned long ulColor, CMaterialItem* pMaterial,
-                                         float fU, float fV, float fSizeU, float fSizeV, bool bRelativeUV, bool bUseFaceToward, const CVector& vecFaceToward,
+                                         float fU, float fV, float fSizeU, float fSizeV, bool bRelativeUV, bool bFlipUV, bool bUseFaceToward, const CVector& vecFaceToward,
                                          bool bPostGUI)
 {
     if (g_pCore->IsWindowMinimized())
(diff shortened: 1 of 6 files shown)


Pirulax:
Well if it works 🤷‍♂️
I dont see anything wrong with it. 
Maybe the syntax.. It seems kinda random now. I suggest adding it to the end.

tederis:
> Well if it works 🤷‍♂️
> I dont see anything wrong with it.
> Maybe the syntax.. It seems kinda random now. I suggest adding it to the end.

It's not possible to add a new argument to the end because of the special meaning of faceToward arguments. Because the default value for them is not provided and a scripter will not be able to skip it.

Einheit-101:
You can skip it by passing nil

tederis:
> You can skip it by passing nil

A nil as an explicit argument must be avoided in my opinion. Three nil arguments one after another can be considered as a weakness in the design.

Pirulax:
Haven't looked at the code, but I'm pretty sure the function used (`ReadIfNextIsBool`) is just:
```cpp
if (NextIsIsBool())
    ReadBool();
```
If the value is anything other than of type `bool` it won't be read.
No explicit nil has to be passed.

Pirulax:
Okay, I looked at current docs.
I think it could be put just after `postGUI`.

tederis:
> Okay, I looked at current docs.
> I think it could be put just after `postGUI`.

We cannot put two optional boolean arguments one after another. Because the next construction:
```c++
if (argStream.NextIsBool())
{

}
if (argStream.NextIsBool())
{

}
```

Gives us the four cases:
```
..., ...., ...., ....... // There is no bool arguments - OK
..., [bool,] [bool,] ... // Two bool arguments - OK
..., [bool,] ........... // Will be considered as bPostGUI boolean
..........., [bool,] ... // Will be considered as bPostGUI boolean
```

As you can see the last two cases is always considered as **bPostGUI** by a parser.

tederis:
In the syntax I suggested, the argument **flipUV** is logically coupled with other UV arguments. And in my opinion it's the best place for it.

Pirulax:
Ah, yeah, sorry, didnt notice that.
Well, okay. If it works as expected, then.. LGTM

Pirulax (approved):

PR #2198, opened 2021-04-18
Potential fix of #1613

tederis:
Here is my try to fix the issue #1613. A description in the discussion below.

### DIFF
--- a/Client/game_sa/CEntitySA.h
+++ b/Client/game_sa/CEntitySA.h
@@ -114,6 +114,11 @@ class CSimpleTransformSAInterface            // 16 bytes
 public:
     CVector m_translate;
     FLOAT   m_heading;
+
+    void UpdateMatrix(CMatrix_Padded* outMatrix)
+    {
+        ((void(__thiscall*)(CSimpleTransformSAInterface*, class CMatrix_Padded*))0x54F1B0)(this, outMatrix);
+    }
 };
 
 class CPlaceableSAInterface            // 20 bytes
@@ -121,6 +126,16 @@ class CPlaceableSAInterface            // 20 bytes
 public:
     CSimpleTransformSAInterface m_transform;
     CMatrix_Padded*             matrix;            // This is actually XYZ*, change later
+
+    void AllocateStaticMatrix()
+    {
+        ((void(__thiscall*)(CPlaceableSAInterface*))0x54F4C0)(this);
+    }
+
+    void AllocateMatrix()
+    {
+        ((void(__thiscall*)(CPlaceableSAInterface*))0x54F560)(this);
+    }
 };
 
 class CEntitySAInterface
--- a/Client/game_sa/CEntitySA.h
+++ b/Client/game_sa/CEntitySA.h
@@ -114,28 +114,13 @@ class CSimpleTransformSAInterface            // 16 bytes
 public:
     CVector m_translate;
     FLOAT   m_heading;
-
-    void UpdateMatrix(CMatrix_Padded* outMatrix)
-    {
-        ((void(__thiscall*)(CSimpleTransformSAInterface*, class CMatrix_Padded*))0x54F1B0)(this, outMatrix);
-    }
 };
 
 class CPlaceableSAInterface            // 20 bytes
 {
 public:
     CSimpleTransformSAInterface m_transform;
-    CMatrix_Padded*             matrix;            // This is actually XYZ*, change later
-
-    void AllocateStaticMatrix()
-    {
-        ((void(__thiscall*)(CPlaceableSAInterface*))0x54F4C0)(this);
-    }
-
-    void AllocateMatrix()
-    {
-        ((void(__thiscall*)(CPlaceableSAInterface*))0x54F560)(this);
-    }
+    CMatrix_Padded*             matrix;            // This is actually XYZ*, change later   
 };
 
 class CEntitySAInterface
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -2,6 +2,22 @@
 #include "gamesa_renderware.h"
 #include "CFileLoaderSA.h"
 
+CFileLoaderSA::CFileLoaderSA()
+{
+}
+
+CFileLoaderSA::~CFileLoaderSA()
+{
+
+}
+
+void CFileLoaderSA::StaticSetHooks()
+{
+    HookInstall(0x5371F0, (DWORD)CFileLoader_LoadAtomicFile, 5);
+    HookInstall(0x537150, (DWORD)CFileLoader_SetRelatedModelInfoCB, 5);
+    HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
+}
+
 class CAtomicModelInfo
 {
 public:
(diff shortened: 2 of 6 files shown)


tederis:
**Preliminaries**:
I have not found any infinite loops in the execution flow. I don’t have enough test cases and I may miss something, but it seems like the assumption of infinite loops was a misconception. There are two main problems:

**The problem 1. Custom mods**.
There are variety of custom mods running under MTA. And, as it turns out, some of custom asset exporters(like KAM’s scripts) may create invalid rotation values, quaternions more specifically. These broken unnormalized quaternions are producing invalid matrices (determinant < 0)  which results in flipped and overgrown rectangles(bounding boxes). That eventually leads to a large number of affected world sectors.

**The problem 2. In-game model replacing**.
By fast camera moving the game is trying to load new objects. But at the same time the old objects are still maintaining a presence in a pointers pool. And it’s possible that in some conditions a pool become running out of free entries. The bigger an object(on XY plane boundary) the more world sectors are affected and more pointers needed. And at some point the game will not be able to allocate a new pointer entry.

**The general solution**:
As shown above, the main problem is lack of free pointers in a pool. So we have to increase a number of entries in a pointers pool(PtrNodeSingleLinkPool). And it may be enough, but according to the problem 1 it would be great to fix rotations. So we have to normalize a quaternion also.

### _Again, I wasn't able to test it on all the cases, but it fixes the cases I found._

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -159,3 +144,87 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
...
+    const float fLenSqr = vecQuaternion.LengthSquared();
+    if (std::fabs(fLenSqr - 1.0f) > std::numeric_limits<float>::epsilon() && fLenSqr > 0.0f)
+        vecQuaternion /= std::sqrt(fLenSqr);
+
+    if (!pEntity->Placeable.matrix) {
+        pPlaceable->AllocateMatrix();
+        pPlaceable->m_transform.UpdateMatrix(pEntity->Placeable.matrix);
+    }
+
+    pEntity->Placeable.matrix->SetRotate(vecQuaternion);
```
forkerer: This is inlined
`pPlaceable->GetMatrix()->SetRotate(vecQuaternion)`

see
https://github.com/codenulls/gta-reversed/blob/d5e83be9f188daf57902d0199f6c5019ba6f3e8d/source/game_sa/Entity/CPlaceable.cpp#L194

I've been also wondering if this issue could be related to max limit of 255 Collisions loaded in CColStore, see
https://github.com/codenulls/gta-reversed/blob/d5e83be9f188daf57902d0199f6c5019ba6f3e8d/source/game_sa/Collision/CColStore.cpp#L55

I've reversed most of the collision related code (other than the collision calculations themselves) there, it may be useful in looking for the solution

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -159,3 +144,87 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
...
     return atomic;
 }
+
+bool IsMatrixNeeed(SFileObjectInstance* pInst, CEntitySAInterface* pEntity)
+{
+    const bool bHasRotation = std::fabs(pInst->rotation.fX) > 0.050000001f || std::fabs(pInst->rotation.fY) > 0.050000001f ||
+        ((pInst->interiorID & 0xFF00) >> 8) & 2;
+    if (!bHasRotation) {
+        return false;
+    }
```
Pirulax:
```suggestion
    if (!(((pInst->interiorID & 0xFF00) >> 8) & 2))
        return false;

    if (std::fabs(pInst->rotation.fX) <= 0.05f && std::fabs(pInst->rotation.fY) <= 0.05f)
        return false;
```
Could you please explain the first `if`?

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -159,3 +144,87 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
...
+    pPlaceable->AllocateStaticMatrix();
+
+    // We don't want to modify the original instance. So we have to copy the rotation.
+    CVector4D vecQuaternion(pInst->rotation);
+
+    /*
+        A quaternion is must be normalized. GTA is relying on an internal R* exporter and everything is OK,
+        but custom exporters might not contain the normalization. And we must do it yourself in MTA.
+    */
+    const float fLenSqr = vecQuaternion.LengthSquared();
```
Pirulax: I think `sq` is the right postfix.

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -159,3 +144,87 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
...
+    if (std::fabs(fLenSqr - 1.0f) > std::numeric_limits<float>::epsilon() && fLenSqr > 0.0f)
```
Pirulax:
```suggestion
    if (fLenSqr > 0.0f && std::fabs(fLenSqr - 1.0f) > std::numeric_limits<float>::epsilon())
```

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -159,3 +144,87 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
     }
     return atomic;
 }
+
+bool IsMatrixNeeed(SFileObjectInstance* pInst, CEntitySAInterface* pEntity)
```
Pirulax: This function does more than it's name says.
I think it should be called `MaybeNormalize<matrix?>`

Pirulax (dismissed):

Pirulax:
Okay, so your solution is to normalize the matrix rotation, in case it's not normalized already, right?

Client/game_sa/CFileLoaderSA.cpp
Pirulax: Okay, just looked at the original GTA code at `0x5381E0`.
Just add a comment it's from GTA code.

tederis:
> Okay, so your solution is to normalize the matrix rotation, in case it's not normalized already, right?

As I have already described in detail, the PR fixes two different problems: broken quaternion and the lack of free pointers in a pool.

Pirulax:
Yeah, I read your comment, just wasn't 100% sure, that's why I asked.
Anyways.. In the future we'd like to get rid of naked functions with C++ in them. 
Although not the topic of this PR, but could the rotation be fixed before calling the function?
Just asking for future reference.

LukeSeers:
> **The general solution**:
> As shown above, the main problem is lack of free pointers in a pool. So we have to increase a number of entries in a pointers pool(PtrNodeSingleLinkPool). And it may be enough, but according to the problem 1 it would be great to fix rotations. So we have to normalize a quaternion also.

I was having the `0x00144BC8` error code with my custom map objects. It would roughly crash the game in the centre of the map and I think you might be right in regards to pooling.

I did discover this though, on server-side using  `createObject( )` with the base custom model and then creating a second `createObject( )` and setting the parameter `isLowLOD = true` it would always crash in the centre of the map, but if I was to not create a LOD then it work fine.


Would crash in the centre of the map with:
```
createObject ( model, x,  y,  z, 0, 0, 0 ) -- base model
createObject ( model, x, y, z, 0, 0, 0, true ) -- Lod
```

Would not crash in the centre of the map with:
```
createObject ( model, x,  y,  z, 0, 0, 0 ) -- base model
-- Removed createObject -- -- Lod
```

> By fast camera moving the game is trying to load new objects. But at the same time the old objects are still maintaining a presence in a pointers pool. And it’s possible that in some conditions a pool become running out of free entries. The bigger an object(on XY plane boundary) the more world sectors are affected and more pointers needed. And at some point the game will not be able to allocate a new pointer entry.

I tested this out on my end, but it made no difference for me on how fast or slow the player was moving.

Dutchman101:
@LukeSeers Thanks for the info, it makes sense to somehow be related (or also, but most likely not exclusively, triggered by) to createObject, as when it happened to me, the server on which was using createObject for maps (afaik, just not with any custom object models).. and the pastebin link from the other PR also made mention of createObject.

LukeSeers:
@Dutchman101 Thanks, but if it was not for you posting https://discord.com/channels/278474088903606273/278521065435824128/639320550774407168 I would have not uncovered it myself. The past you help me fix my problem in regards to leading me here 👍 


After trying to tackle this issues even more on my end I got fed up and decided to read the pull request again. This time a bulb went off, 

>The bigger an object(on XY plane boundary) the more world sectors are affected and more pointers needed.

From my testing it was not the custom model(DFF) itself but the collision(COL) so this in mind I relooked into the COL boundary and tried to optimise it in regards to making the X and Y small as possible. In return it works perfectly now.

Pirulax:
~~@tederis was quicker https://github.com/multitheftauto/mtasa-blue/pull/2198~~
Nvm lmao

Client/game_sa/CFileLoaderSA.cpp
tederis: Thanks, that's good to know.

tederis:
> Yeah, I read your comment, just wasn't 100% sure, that's why I asked.
> Anyways.. In the future we'd like to get rid of naked functions with C++ in them.
> Although not the topic of this PR, but could the rotation be fixed before calling the function?
> Just asking for future reference.

I've reconsidered the solution. I think we could get rid of the naked function and move calculations straight after an IPL line parsing. 

Turning to question, it depends on a particular situation. More generally no, it cannot be. But in this particular situation it's not against the rules.

tederis:
> I tested this out on my end, but it made no difference for me on how fast or slow the player was moving.

I did some experiments where a camera traveled at high speeds(about ~1000 units per second) and found a pointers pool overflow. This behavior occurs only with large custom models. I think this may manifest itself including during player teleportation and spawning.

LukeSeers:
> I did some experiments where a camera traveled at high speeds(about ~1000 units per second) and found a pointers pool overflow. This behavior occurs only with large custom models

Hmm, it's funny that you say that. I've been having an issue kinda just like you mentioned. Where if a player travels at high speeds the custom terrain model takes a while to load the COL. I thought that it could be the fact the loading is slow, but what threw me off was sometimes it would load and other times it didn't. From my testing it only seems to happened when traveling at high speeds. 

> This behavior occurs only with large custom models

The interesting part is that my terrain is large custom models too. I did this for many reasons.

LukeSeers:
~I don't know how useful this information will be for anybody else, but I found a solution to my past comment.~

~I discovered that running the function `createObject` on server-side that it has weird loading problems with COLs for clients. As I mentioned before:~
> Where if a player travels at high speeds the custom terrain model takes a while to load the COL. I thought that it could be the fact the loading is slow, but what threw me off was sometimes it would load and other times it didn't. From my testing it only seems to happened when traveling at high speeds.

~I would have assumed that when using the function `createObject` on server-side all it would be doing is something similar to a event system where the client syncs the objects. I wanted to test this so I created my own and this is the results.~ 

~If I was to move the function `createObject` on client side so that I pass on the data(position, rotation etc) to a client with events and let the client create the objects. I no longer have the weird problem of COLs not loading. I don't understand why this is the case since as I mentioned before I would have assumed that when using the function `createObject` on server-side all it would be doing is something similar.~

~Note:
just for reference I moved these from server-side to client-side.~ 
```
createObject ( model, x,  y,  z, 0, 0, 0 ) -- base model
createObject ( model, x, y, z, 0, 0, 0, true ) -- Lod
```

LukeSeers:
Update:
I don't understand how this is the case, but it is. When in the editor play mode, for some odd reason after traveling a X amount the COL doesn't seem to wanna load.

I tested the resource in my gamemode and it loads the terrains in their positions and I have no problem driving around. When I import the resource into the editor, I enter play mode and start to drive around, then the weird COL not loading problem happens.

@tederis When you was testing was you also using the editor?

Dutchman101:
@LukeSeers after reading a bunch more here, im no longer sure if the issue(s) you're describing have much to do with this pools crash.. but more information (even after this point) never hurts.

Dutchman101:
It would be nice to get this PR merged before 1.5.9, as if from monitoring it turns out that it does fix the issue, we would be fixing a major crash - thousands of people are getting it, it's a top 10 GTA crash. 

There were 3200 unique victims that may have repeatedly crashed during the past month, and they were probably unable to join that server without immediately crashing until the server owner did something to relieve it.

Pirulax:
How does patching IPL loading fix the issue?
AFAIK IPL is only for default GTA models.
Am I missing something?

tederis:
> How does patching IPL loading fix the issue?
> AFAIK IPL is only for default GTA models.
> Am I missing something?

As I have already described in detail in the first paragraph, "There are variety of custom mods running under MTA". And the purpose of a quaternion normalization is to protect them from this sort of crash. In addition, it's also a good practice to normalize quaternions before use.

tederis:
It seems like this fix is not in very big demand.

Dutchman101:
> It seems like this fix is not in very big demand.

yes it is, like any shot at fixing this long term issue would be. As with many other things, code reviews just requires patience.

AlexTMjugador:
> How does patching IPL loading fix the issue?
> AFAIK IPL is only for default GTA models.
> Am I missing something?

It seems (but I'm no expert, so I might be wrong) that GTA insists in loading data from some file when a `CObjectSA` instance is created:

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/game_sa/CObjectSA.cpp#L50-L63

I assume that such instances are created when an object is streamed in, which can happen due to the MTA streamer doing its work or a model being replaced, so patching that load method works for custom models too, replaced with `engineReplaceModel`. But I think that @tederis may be able to shed more light on _why_ does this exactly work, and correct me if I'm wrong. From what file is GTA originally supposed to read that lines from? I took a quick look at GTA data files and I didn't see any file with 11 fields separated by spaces.

botder (approved):
Code looks good.

Dutchman101:
@tederis As you commented about it before, did you notice that you're already free to merge this PR? Is there anything you're still working on?

tederis:
> @tederis As you commented about it before, did you notice that you're already free to merge this PR? Is there anything you're still working on?

In terms of fixing problems declared in the my first comment, everything is ready to merging. After it is released I will waiting for the feedback.

Client/game_sa/CEntitySA.h
```diff
@@ -120,7 +120,7 @@ class CPlaceableSAInterface            // 20 bytes
 {
 public:
     CSimpleTransformSAInterface m_transform;
-    CMatrix_Padded*             matrix;            // This is actually XYZ*, change later
+    CMatrix_Padded*             matrix;            // This is actually XYZ*, change later
```
patrikjuvonen, line 123: Nit: Unwanted whitespace added at the end of the line.

patrikjuvonen (approved):
Looks good to me apart from the tiny whitespace change in CEntitySA.h.

Dutchman101:
@tederis According to crash stats, again this didn't succeed to (fully) fix it.. like saml1er's earlier attempt (#1613) it's better to keep it in anyways as it may end up reducing the incidence due to just fixing certain causes behind it

If you're interested to debug it, the attached .zip file contains most recent dumps of all 3 related crashes:
[dumps.zip](https://github.com/multitheftauto/mtasa-blue/files/6719395/dumps.zip)

0x0013368B remains the main crash

Also, see that upon crashing, usage of POINTER_DOUBLE_LINK_POOL is still 100%:
pools

The red value.

On the main crash 0x0013368B (only one with pic provided, as above) 100% used pool is that POINTER_DOUBLE_LINK_POOL and on the 2 related crashes it's ENTRY_INFO_NODE_POOL instead.

I noticed your PR changed SINGLE LINK pool capacity, not DOUBLE LINK or even the pool from the 2 other related crashes. So, maybe you made a mistake there. But anyways with infinite loop somewhere, increasing capacity of more pools probably won't help anything but to increase base RAM usage (it will breach the limit either way).

PR #2199, opened 2021-04-18
Fix #2183: Tweak bind queue logic

### LINKED ISSUE #2183: getPedOccupiedVehicle() and bindKey() combinations were broken after #2123
**Describe the bug**
Before, the bindKey() was called before any other key and wasn't added in a queue. This was changed on #2123 and brought an issue for binds on keys like "G" or "F". 

Because getPedOccupiedVehicle(player) returns true when the player starts entering a vehicle, if you bind a key to "F" and do a simple check (``if not getPedOccupiedVehicle(thePlayer) then``) the if statment will return false, when before #2123 it returned true because of the order of the key presses.

**To reproduce**
[test.zip](https://github.com/multitheftauto/mtasa-blue/files/6289962/test.zip)

```lua
-- serverside
local thePlayer = getRandomPlayer()
spawnPlayer(thePlayer, 0, 0, 3)
local veh = createVehicle(423, 3, 3, 3)

function skereCMD(thePlayer, cmd)
	if not getPedOccupiedVehicle(thePlayer) then
		outputChatBox("SKERE")
	end
end
addCommandHandler("skere", skereCMD, false, false)
bindKey(thePlayer, "F", "down", "skere")
```

How it works after r20844
1) Download and run the resource test in r20844 (clientside)
2) Try entering the vehicle, you will be able but no message chat will appear.
3) If you get far away from the vehicle (in order to not get into it) then a message will appear.

How it worked before r20844
1) Download and run the resource test in r20839 (clientside)
2) Try entering the vehicle, you will be able and a message will appear
3) Same message will appear every time you press the key, no matter if you started to enter a vehicle of not.

**Expected behaviour**
Don't break backwards compatibility, or make a way to adjust it.

For the test: The message should appear always, getting into the vehicle or not (except if you are inside the vehicle, of course).

**Video**
Disclaimer: I was on my custom server, but no resource should have been disturbing the test, and I'm pretty sure this didn't happen before.

https://youtu.be/WwwyG90V-Xw

**Version**
r20844 (clientside)

**Additional context**
I think that the changes brought with #2123 are more important that the issue with the getPedOccupiedVehicle() and the bindKey, but probably this isn't the only logical statment that was changed because of the queue added to the binds.

I don't like the idea of reverting #2123, provided that it actually does help with performance in an important amount. I suggest that getPedOccupiedVehicle() gets an extra argument so we can check if the ped is already into the car or is just entering it (and isPedInVehicle too if possible, pretty pls).

Something like:

```lua
vehicle getPedOccupiedVehicle ( ped thePed, [ bool checkEntering = true ] )
```

checkEntering will determine if the player is considered inside the vehicle while entering (true) or only when inside (false). The only other way to check this is with events.

I brought this issue up mostly because a lot of my users are having problems with some binds which do this check. ``isPedInVehicle`` is not good because it has the same issue as the getPedOccupiedVehicle() one (but it doesn't return a vehicle, which is faster).

patrikjuvonen:
## Summary
- Fixes #2183
- Execute script created binds same frame
- Execute up to 2 binds in same frame

### DIFF
--- a/Client/core/CKeyBinds.cpp
+++ b/Client/core/CKeyBinds.cpp
@@ -483,7 +483,7 @@ bool CKeyBinds::ProcessKeyStroke(const SBindableKey* pKey, bool bState)
                                                 // don't add if its already added to queue
                                                 if (!bAlreadyProcessed)
                                                 {
-                                                    if (processedList.empty())
+                                                    if (pCommandBind->bScriptCreated || processedList.empty())
                                                         Call(pCommandBind);
                                                     else
                                                         m_vecBindQueue.push_back(pCommandBind);
--- a/Client/core/CKeyBinds.cpp
+++ b/Client/core/CKeyBinds.cpp
@@ -2094,9 +2094,12 @@ void CKeyBinds::DoPreFramePulse()
         m_pChatBoxBind = NULL;
     }
 
-    if (!m_vecBindQueue.empty())
+    // Execute two binds from queue
+    for (auto i = 0; i < 2; i++)
     {
         auto it = m_vecBindQueue.begin();
+        if (it == m_vecBindQueue.end())
+            break;
         Call(*it);
         m_vecBindQueue.erase(it);
     }


sbx320 (approved):

Dutchman101:
I noticed something odd - on a server with multiple resources using bindKey, at least 1 key that should have been bound didn't successfully execute the command it was bound to / was bound at all (didn't debug enough to differentiate.. if it happens again i will). I cannot immediately reproduce again after restarting MTA.

The script was using this to bind clients to a serverside function:
```lua
function client_PlayerJoin()
	bindKey(source, "l", "down", doToggleThing)
end
addEventHandler("onPlayerJoin", getRootElement(), client_PlayerJoin)
```
I only suspect it has something to do with it (the chance to get some script's key not bound or triggered successfully) because i haven't ever seen it before using the nightly build with this change. So far unconfirmed, just a hunch, so not adding issue yet. @patrikjuvonen

ArranTuna:
> 
> 
> I noticed something odd - on a server with multiple resources using bindKey, at least 1 key that should have been bound didn't successfully execute the command it was bound to / was bound at all (didn't debug enough to differentiate.. if it happens again i will). I cannot immediately reproduce again after restarting MTA.
> 
> The script was using this to bind clients to a serverside function:
> 
> ```lua
> function client_PlayerJoin()
> 	bindKey(source, "l", "down", doToggleThing)
> end
> addEventHandler("onPlayerJoin", getRootElement(), client_PlayerJoin)
> ```
> 
> I only suspect it has something to do with it (the chance to get some script's key not bound or triggered successfully) because i haven't ever seen it before using the nightly build with this change. So far unconfirmed, just a hunch, so not adding issue yet. @patrikjuvonen

@Dutchman101 this bug you mentioned is affecting CIT since updating the server yesterday. Various scripts that use bindKey on server side are broken. Strangely can't reproduce on local server but easily reproduced on main server:

[CIT]Arran executed command: bindKey(getPlayerFromName("[cit]arran"), "mouse2", "down", function(a) outputChatBox("bind works", a) end)
Right clicking doesn't output bind works.

Weird that client side bind works fine and weird how updating server has introduced this bug despite this change being client side.

Tried reverting client to r20855 and it didn't help.

PR #2209, opened 2021-04-27
Fix #611: Add line spacing to dxDrawText

### LINKED ISSUE #611: dxDrawText changing line spacing
Some fonts have wrong line spacing and would be nice if we could adjust it.

Add option to change distance between lines in dxDrawText

-

**Additional context**
image
example of font that have to big line spacing

Allerek:
This PR will add lineHeight as last argument for dxDrawText, b.w.c. with older versions as 'false' 'nil' or '0' will use old system for line spacing.
This only works for colorcoded dxDrawText, as non-colorcoded uses different function for text rendering.
If someone manages to add this possilibites to non-colorcoded text, that would be great.
image

Test resource:
[drawTest.zip](https://github.com/multitheftauto/mtasa-blue/files/6387064/drawTest.zip)

Fixes https://github.com/multitheftauto/mtasa-blue/issues/611


Client/core/Graphics/CGraphics.cpp
```diff
@@ -1138,7 +1138,8 @@ void CGraphics::DrawStringQueued(float fLeft, float fTop, float fRight, float fB
             fRight = floor(fRight);
             fBottom = floor(fBottom);
         }
-
+        if (fLineHeight == 0.0f)
```
tederis: It's better to use `std::optional` in here I think

tederis: Okay, I see some of modules(game_sa or multiplayer_sa) do not support C++17 features. You can't use std::optional here.

Pirulax:
Yeah, apparently they've changed something in MSVC which caused naked functions to break.

Client/core/Graphics/CGraphics.cpp
Pirulax: Yeah, apparently they've changed something in MSVC C++17 which caused naked functions to break.

Allerek:
> Yeah, apparently they've changed something in MSVC which caused naked functions to break.

Yeah, i've reverted back to float comparison method, just like it was meant in first commit, doing tests and its ready for review.

Client/core/Graphics/CGraphics.cpp
Pirulax: Btw I think he could use negative values to indicate something is wrong (as you obviously can't have negative spacing).

Allerek:
image
Works just fine.

Pirulax:
Definitely a nice feature, but imho people rarely do text break by themselves. I personally very rarely used any newlines by hand in dxDrawText.
One day hopefully we'll just replace DxDrawTextW with our own stuff, that way both this spacing stuff and colorcoded wordbreak would work.

Allerek:
I think we should merge it for now, and then think about new drawing functions.

theSarrum:
Personally, I think this is a bad practice. People will start using the `colorCoded` argument to be able to change the line spacing, but this will have a great impact on performance (more CPU load).

Allerek:
> Personally, I think this is a bad practice. People will start using the `colorCoded` argument to be able to change the line spacing, but this will have a great impact on performance (more CPU load).

Thats their decision, we just allow them to change the spacing.

Client/core/Graphics/CGraphics.cpp
Allerek: > Btw I think he could use negative values to indicate something is wrong (as you obviously can't have negative spacing).

Well, you can(last example in screenshot below, where the last line became the first)

patrikjuvonen:
Thanks for your work with this!

Unfortunately you seem to have deleted the original repository you created this pull request with.

Please move your changes under a feature branch in your repository and recreate this pull request so we can test it properly.

PR #2215, opened 2021-04-30
Add set/getChatboxCharacterLimit and increase max chat message length to 255

### LINKED ISSUE #2090: Add setChatBoxCharacterLimit (increase default chatbox character limit)
**Describe the feature request**
Add a new scripting function (or configuration to opt-in) that expands the character limit of default chatbox

For more information, refer to this popular suggestion: [https://forum.mtasa.com/topic/58855-chat-length-max-allowed-symbols/](https://forum.mtasa.com/topic/58855-chat-length-max-allowed-symbols/) which is the reason for adding this to the tracker

Allow servers to increase and control the chatbox character limit if they desire so

The alternative is a custom chatbox. But it should be trivial for us to support it natively, in the default chatbox.
Especially for such a trivial change, a lot of scripters don't want to hop to writing a custom chatbox as the complications (and adaptations needed in their other resources) outweigh the feature that they gain. It is a popular suggestion for a reason, imagine the ease of using a scripting function or other toggle (with configurable character limit) for this.

Lpsd:
This PR resolves #2090 by adding set/getChatboxCharacterLimit functions to the **client-side only.**

It also increases the max chat message length to 255 (not including player name or "(TEAM)" tags), over the previous maximum (and default) of 96. The default value will still stay at 96.

___

### setChatboxCharacterLimit
```lua
bool setChatboxCharacterLimit(int charLimit)
```
Passing -1 will reset the character limit  back to 96.
If you pass anything other than an integer, or outside the range of 0-255 (excluding -1), it will return false.
      
___

### getChatboxCharacterLimit
```lua
int getChatboxCharacterLimit()
```
Pretty self-explanatory.

___

When a player disconnects from a server, their character limit is set back to the default (96).


Lpsd:
On hold, waiting for my review / merge of #2170 

Will need to address conflicts in this PR after that.

Client/sdk/core/CCoreInterface.h
```diff
@@ -180,6 +176,11 @@ class CCoreInterface
     virtual void        OnGameTimerUpdate() = 0;
 
     virtual CDiscordManagerInterface* GetDiscordManager() = 0;
+
+    virtual bool SetChatboxCharacterLimit(int charLimit) = 0;
+    virtual void ResetChatboxCharacterLimit() = 0;
+    virtual int  GetChatboxCharacterLimit() = 0;
+    virtual int  GetChatboxMaxCharacterLimit() = 0;
```
Pirulax: For further reference: Netcode (presumeably) uses CCoreInterface. If the order is changed it obviously crashes. This shouldn't be an issue in production nightlies though.

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -281,6 +287,7 @@ void CLuaGUIDefs::AddGuiElementClass(lua_State* luaVM)
     lua_classvariable(luaVM, "text", "guiSetText", "guiGetText");
     lua_classvariable(luaVM, "size", "guiSetSize", "guiGetSize");
     lua_classvariable(luaVM, "position", "guiSetPosition", "guiGetPosition");
+    //lua_classvariable(luaVM, "chatboxCharacterLimit", "setChatboxCharacterLimit", "getChatboxCharacterLimit");
```
patrikjuvonen: Why is this commented out?

Client/mods/deathmatch/logic/Config.h
```diff
@@ -52,17 +52,17 @@ class CVaPassNext
 #define va_pass(valist) CVaPassNext(valist).svapassnext
 /*** va_pass() (passing of ... variable length arguments ***/
 
-// Min and max number of characters in chat messages
+// Min and max number of characters in chat messages (the message itself, not including player nick)
```
patrikjuvonen, line 55: I wish there was a way to merge these client and server configs together in a shared config file, but maybe out of scope for this PR.

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
Lpsd: no good reason that I can think of, probably just forgot to uncomment

Client/mods/deathmatch/logic/Config.h
Lpsd: Yeah it also needs to be made clearer what exactly the difference between max chat length, and max **chat echo** length is (among other things). 

I've cleared it up a little bit here hopefully, including fixing an oversight in `Packet_ChatEcho` where `MAX_OUTPUTCHATBOX_LENGTH` was being used instead of `MAX_CHATECHO_LENGTH`.

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
botder:
```suggestion
    lua_classvariable(luaVM, "chatboxCharacterLimit", "setChatboxCharacterLimit", "getChatboxCharacterLimit");
```

Client/core/CChat.h
```diff
@@ -203,6 +202,11 @@ class CChat
     void SetChatFont(eChatFont Font);
     void OnModLoad();
 
+    void      SetCharacterLimit(int charLimit);
+    int       GetCharacterLimit() const { return m_iCharacterLimit; }
+    const int GetDefaultCharacterLimit() const { return m_iDefaultCharacterLimit; }
+    const int GetMaxCharacterLimit() const { return m_iMaxCharacterLimit; }
```
botder:
```suggestion
    int GetDefaultCharacterLimit() const { return m_iDefaultCharacterLimit; }
    int GetMaxCharacterLimit() const { return m_iMaxCharacterLimit; }
```

Client/core/CChat.h
```diff
@@ -292,4 +296,9 @@ class CChat
     CTickCount m_lastRenderTargetCreationFail;
 
     bool m_bNickCompletion;
+
+    int m_iCharacterLimit;
+    const int m_iDefaultCharacterLimit = 96;
+    const int m_iMaxCharacterLimit = 255;
+    const int m_iMaxInputLines = 5;
```
botder: It would make more sense to use `static inline constexpr` constants instead.

botder (dismissed):

patrikjuvonen (approved):
Tested working, looks good.

PR #2218, opened 2021-05-04
Fix serverside key binds not working since #1978

### LINKED ISSUE #2217: r20854 broke server side key binds
**Describe the bug**
Players using r20854 and above can no longer trigger server side key binds.

**To reproduce**
Can't seem to reproduce this on local server (even tried local server being same as CIT's but different to client) but on CIT various scripts are all failing to work properly with it. For example doing '/afk' to come back from not being AFK there are key binds on the move keys in order to make you no longer AFK, these cause players to be stuck in AFK mode until reconnecting.

r20854 doesn't work.
r20852 works fine:
srun bindKey(getPlayerFromName("[cit]arran"), "mouse2", "down", function(a) outputChatBox("bind works", a) end)
Right clicking makes bind works.

This commit by @patrikjuvonen must be the cause as I see mentions of key bind edits in the commit:
https://github.com/multitheftauto/mtasa-blue/commit/146bd15a87b3391cdedce0cbb3b0e261b1ea457d

**Expected behaviour**
To not be broken.

**Screenshots**

**Version**
Multi Theft Auto v1.5.8-release-20854
MTA:SA Server v1.5.8-release-20863

**Additional context**

Lpsd:
Resolves #2217 

Keybind processing on the client relies on the CURSOR_EVENT RPC which was disabled by default after #1978

(for more info read the code around here, which is calling CClientGame::ProcessMessageForCursorEvents)
https://github.com/multitheftauto/mtasa-blue/blob/17a2f84d733da99db8aa6414c09ee1bdeebe5893/Client/core/CMessageLoopHook.cpp#L443

This PR fixes that, by leaving the RPC disabled by default, but checking everytime a bind is added or removed, whether the RPC needs enabling (or disabling again).

### DIFF
--- a/Server/mods/deathmatch/logic/CKeyBinds.cpp
+++ b/Server/mods/deathmatch/logic/CKeyBinds.cpp
@@ -654,7 +654,7 @@ void CKeyBinds::CheckAndUpdatePlayerRPC()
     }
     else if (m_bIsRPCEnabled && listSize == 0)
     {
-        disabledServerRPCFunctions[eServerRPCFunctions::CURSOR_EVENT] = false;
+        disabledServerRPCFunctions[eServerRPCFunctions::CURSOR_EVENT] = true;
         m_pPlayer->Send(CServerRPCControlPacket(disabledServerRPCFunctions));
         m_bIsRPCEnabled = false;
     }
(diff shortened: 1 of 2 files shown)


Lpsd:
#1978 was reverted in c952054 and we have lightly discussed better solutions to this on Discord.

PR #2235, opened 2021-05-27
Add support aes128 encryption to encodeString/decodeString

onikyannn:
So because of tea encryption is still broken (https://github.com/multitheftauto/mtasa-blue/issues/1443) i made alternative solution using aes from cryptopp lib.

Test resource attached.
[test_resource.zip](https://github.com/multitheftauto/mtasa-blue/files/6557309/test_resource.zip)

To test:
1. Start resource
2. Execute command "test_sync" (syncronous encription)
2. Execute command "test_async" (asyncronous encription)
2. Execute command "test_file" (syncronous encription with writing/reading encrypted data to file)


Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -294,6 +294,57 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
                 }
                 return 1;
             }
+            case StringEncryptFunction::AES128:
+            {
+                SString& key = options["key"];
+                SString& iv = options["iv"];
```
Pirulax: I dont think `iv` should be a required field. Actually I'm not even sure if taking user input is a good idea. Rather we should return the hashed data + iv, and let the user store it.
`iv` is [initialization vector](https://www.cryptopp.com/wiki/Initialization_Vector).

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +30,56 @@ namespace SharedUtil
...
+
+    inline SString Aes128encode(const SString& sData, const SString& sKey, const SString& sIv)
+    {
+		SString result;
+
+		CryptoPP::byte iv[ CryptoPP::AES::BLOCKSIZE ];
+		CryptoPP::byte key[ CryptoPP::AES::DEFAULT_KEYLENGTH ];
+
+		memcpy(iv, sIv.c_str(), CryptoPP::AES::BLOCKSIZE );
+		memcpy(key, sKey.c_str(), CryptoPP::AES::DEFAULT_KEYLENGTH);
```
Pirulax: I dont think you have to use this array. In theory passing in the string + its length should be fine.

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +30,56 @@ namespace SharedUtil
...
+		CryptoPP::CBC_Mode_ExternalCipher::Encryption cbcEncryption( aesEncryption, iv );
+
+		try
+		{
+			CryptoPP::StreamTransformationFilter stfEncryptor(cbcEncryption, new CryptoPP::StringSink( result ) );
+			stfEncryptor.Put( reinterpret_cast<const unsigned char*>( sData.c_str() ), sData.length() );
+			stfEncryptor.MessageEnd();
+		}
+		catch(const CryptoPP::Exception& exception) {
+			return exception.what();
```
Pirulax: Good luck figuring out an exception was thrown in the caller function.

AlexTMjugador:
Also, is there any reason to choose CBC mode? Other modes, like CTR, [can be more suitable](https://stackoverflow.com/questions/1220751/how-to-choose-an-aes-encryption-mode-cbc-ecb-ctr-ocb-cfb), either in performance (CBC doesn't allow for parallel encryption and requires proper padding; CTR and others can be parallelized and do not require such padding) or other characteristics, while providing equivalent security.

onikyannn:
> Also, is there any reason to choose CBC mode? Other modes, like CTR, [can be more suitable](https://stackoverflow.com/questions/1220751/how-to-choose-an-aes-encryption-mode-cbc-ecb-ctr-ocb-cfb), either in performance (CBC doesn't allow for parallel encryption and requires proper padding; CTR and others can be parallelized and do not require such padding) or other characteristics, while providing equivalent security.

I just use the mode that is most often used. Which mode do you suggest?

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
onikyannn: Maybe you're right, but personally, it would be more convenient for me to generate iv myself. In this case, for example, I can generate and store a key and a iv in advance, in order to then encrypt the files later with the already existing key and iv.

onikyannn: Also we can write example function for generating a random string on the wiki, for the convenience of users.

Shared/sdk/SharedUtil.Crypto.h
onikyannn: It returns an error as string, instead of decrypted file. Maybe it should just return false?

AlexTMjugador:
> I just use the mode that is most often used. Which mode do you suggest?

I'm no cryptography expert, but on paper, according to the information I could find, CTR mode seems like the best for this purpose. I don't know if CryptoPP's implementation is actually faster than CBC, though, but GCM mode (which uses CTR under the hood) is in wide use in TLS 1.2 and other protocols.

onikyannn:
> > I just use the mode that is most often used. Which mode do you suggest?
> 
> I'm no cryptography expert, but on paper, according to the information I could find, CTR mode seems like the best for this purpose. I don't know if CryptoPP's implementation is actually faster than CBC, though, but GCM mode (which uses CTR under the hood) is in wide use in TLS 1.2 and other protocols.

Based on [benchmark](https://www.cryptopp.com/benchmarks.html) CTR is fastest

Shared/sdk/SharedUtil.Crypto.h
onikyannn: Fixed

Shared/sdk/SharedUtil.Crypto.h
onikyannn: Changed to `return ""`

onikyannn:
Algorithm changed to CTR.
Also made some performance test on ~11mb file, and here is results:

aes cbc:
encoding time = 74;
decoding time = 368;

aes ctr:
encoding time = 62;
decoding time = 65;

tea:
encoding time = 427;
decoding time = 416;

onikyannn:
@Pirulax Something else is blocking us to merge this pr?

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
AlexTMjugador: We currently do not provide any CSPRNG via the MTA Lua API, and `math.random` can't be assumed suitable for cryptographic purposes, so properly generating IVs in resources is definitely not easy. Also, IVs should **never** be reused. I agree with @Pirulax suggestion here; letting MTA take care of generating a unique, suitable IV and return it is a good idea.

Shared/sdk/SharedUtil.Crypto.h
AlexTMjugador: I think it'd be desireable that the Lua function returns false in this case, like other MTA Lua functions do.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
AlexTMjugador: Also, what's the use case for knowing the IV in advance? What does that allow you to do that you can't already do after MTA returns the IV it used for encryption?

Pirulax:
I mean.. I like to have unit tests.

Edit: You have provided a test resource, but it isn't suitable, as it's just a test resource, not a unit test.
I'd gladly merge it along with a unit test (after the IV issue is resolved).

Shared/sdk/SharedUtil.Crypto.h
Pirulax: I think you shouldn't catch this exception here, and just let the caller handle it (keep in mind the new parser only handles `std::invalid_argument`)

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
Pirulax: We use CTR mode here, see [wiki](https://www.cryptopp.com/wiki/CTR_Mode).

Generally IV is regarded as public information.
So you just storing it somewhere is perfectly fine.
Also, some cipher modes require unique IVs.

onikyannn:
Done, but have issue with windows-debug build.
To be honest, I don’t know how to fix it, I think I need help.
Locally it can be fixed by adding cryptopp to loader project references.

onikyannn:
Windows-debug build fixed. @Pirulax @AlexTMjugador please review pr as possible.
Thank you.

AlexTMjugador (approved):
For what is worth, functionality-wise, this looks good to me, but let's see if someone else has some feedback about other concerns, like code style.

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,48 @@ namespace SharedUtil
...
+
+        sIv.resize(sizeof(iv));
+        rnd.GenerateBlock((CryptoPP::byte*)sIv.data(), sIv.size());
+        memcpy(iv, sIv.c_str(), sIv.size());
+
+        CTR_Mode<AES>::Encryption aesEncryption;
+        aesEncryption.SetKeyWithIV(key, sizeof(key), iv);
+        StringSource ss(sData, true, new StreamTransformationFilter(aesEncryption, new StringSink(result)));
+
+        return std::pair<SString, SString>(result, sIv);
```
Pirulax:
```suggestion
        return {result, sIv};
```

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,48 @@ namespace SharedUtil
...
+        SString result;
+
+        SecByteBlock key(0x00, AES::DEFAULT_KEYLENGTH);
+        SecByteBlock iv(0x00, AES::BLOCKSIZE);
+
+        memcpy(key, sKey.c_str(), sKey.size());
+        memcpy(iv, sIv.c_str(), sIv.size());
+
+        CTR_Mode<AES>::Decryption aesDecryption;
+        aesDecryption.SetKeyWithIV(key, sizeof(key), iv);
```
Pirulax: Use this, and delete the `SecByteBlock` stuff.
```cpp
        aesDecryption.SetKeyWithIV(sKey.c_str(), sKey.length(), sIv.c_str());
```
Also, do some checks before calling this function, as to be `sIv.length() >= AES::BLOCKSIZE`

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,48 @@ namespace SharedUtil
...
+    {
+        using namespace CryptoPP;
+
+        AutoSeededRandomPool rnd;
+
+        SString result;
+        SString sIv;
+
+        SecByteBlock key(0x00, AES::DEFAULT_KEYLENGTH);
+        SecByteBlock iv(0x00, AES::BLOCKSIZE);
```
Pirulax: Why do you even use these? Just pass in the data directly from the `string`s.

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,48 @@ namespace SharedUtil
...
+
+        memcpy(key, sKey.c_str(), sKey.size());
+
+        sIv.resize(sizeof(iv));
+        rnd.GenerateBlock((CryptoPP::byte*)sIv.data(), sIv.size());
+        memcpy(iv, sIv.c_str(), sIv.size());
+
+        CTR_Mode<AES>::Encryption aesEncryption;
+        aesEncryption.SetKeyWithIV(key, sizeof(key), iv);
```
Pirulax: To pass in data directly, use this:
```cpp
aesEncryption.SetKeyWithIV(sIv.data(), sIv.length(), sIv.length()); 
```

Pirulax (dismissed):

Shared/sdk/SharedUtil.Crypto.h
onikyannn: Fixed

Shared/sdk/SharedUtil.Crypto.h
onikyannn: Fixed

Shared/sdk/SharedUtil.Crypto.h
onikyannn: Fixed

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,38 @@ namespace SharedUtil
...
+        StringSource ss(sData, true, new StreamTransformationFilter(aesEncryption, new StringSink(result)));
+
+        return {result, sIv};
+    }
+
+    inline SString Aes128decode(const SString& sData, const SString& sKey, const SString& sIv)
+    {
+        using namespace CryptoPP;
+        using CryptoPP::byte;
+
```
Pirulax, line 59:
```suggestion

        sIv.resize(AES::BLOCKSIZE);
```

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -28,4 +31,38 @@ namespace SharedUtil
...
+        rnd.GenerateBlock((byte*)sIv.data(), sIv.size());
+
+        CTR_Mode<AES>::Encryption aesEncryption;
+        aesEncryption.SetKeyWithIV((byte*)sKey.data(), sKey.size(), (byte*)sIv.data());
+        StringSource ss(sData, true, new StreamTransformationFilter(aesEncryption, new StringSink(result)));
+
+        return {result, sIv};
+    }
+
+    inline SString Aes128decode(const SString& sData, const SString& sKey, const SString& sIv)
```
Pirulax:
```suggestion
    inline SString Aes128decode(const SString& sData, const SString& sKey, SString sIv)
```

Shared/sdk/SharedUtil.Crypto.h
Pirulax: Our new policy is to hard error on usage mistakes, isnt it? 
Although I'm not sure what throws here and what. If it throws for memory allocation errors as well, that should be handled differently (and most likely it does). 
I think the key is automatically padded by the algorithm, so that cant be a usage mistake. 
Maybe we should throw invalid_argument if the `iv` (in AES decode) isn't of the desired length (BLOCKSIZE), instead of resizing it.

onikyannn: I was getting an exception in the first revision when CBC mode was used. Exception was thrown when an incorrect decryption key was passed. Now we use CTR mode, and this problem no longer exists. I tested different scenarios and didn't get an exception. I can assume that it could be removed. Or leave it just in case.

onikyannn: > I think the key is automatically padded by the algorithm, so that cant be a usage mistake

its not

onikyannn: @Pirulax what else do you consider necessary to fix?

onikyannn: Any updates?

onikyannn:
Anybody? Can we merge this pr, please?

Lpsd (approved):
Looks good, all previously requested changes have been made and I can't see anything else.

thanks for the PR!

Issue #1833, opened 2020-11-13
Client-side: changing a 4 doors vehicle model to a 2 doors model while being a passenger in 3rd or 4th seats will get you a network trouble


Shuubaru:
**Describe the bug**
If you enter as a passenger (3rd or 4th seats) in a 4 door vehicle, and then change the model to a 2 door vehicle you will get a network trouble. Changing it back to a 4 door model wont fix it (server or client side), so you will have to reconnect.

**To reproduce**

Steps to reproduce:
1. create a 4 door vehicle `/cv sentinel`
2. enter as a passenger using 3rd or 4th door
3. change model to 2 door vehicle, like an infernus: `crun setElementModel(getCameraTarget(), 411)`
4. you will get frozen with network trouble, forced to reconnect.

**Expected behaviour**
You should be kicked out of the car, and this behavior be synced to avoid any sync problem, which server side setElementModel already does.

**Screenshots**

In this video I show I could reproduce it after using custom models from engineRequestModel function, but this can also be reproduced with default ids, as I explained before:

[https://streamable.com/v889k4](url)

**Version**
Multi Theft Auto v1.5.8-release-20723
both server and client.

**Additional context**
Since custom ids are already a thing in MTA, and they can only be used with client sided setElementModel, I believe this can be (probably) a problem in the future.

Pirulax:
Interesting. 
~~It wasn't a problem when `engineReplaceModel` warped you out of the vehicle.~~
My bad. This was done using `setElementModel`.
Edit: This is kinda unsolvable, as this is done client side. And while the server think's youre in a vehicle, your client disagrees, because you've changed the model. The way to fix this it to not set the model on the client side, but on the server.

Allerek:
> Interesting.
> ~It wasn't a problem when `engineReplaceModel` warped you out of the vehicle.~
> My bad. This was done using `setElementModel`.
> Edit: This is kinda unsolvable, as this is done client side. And while the server think's youre in a vehicle, your client disagrees, because you've changed the model. The way to fix this it to not set the model on the client side, but on the server.

We can check for amounts of doors before replacing model, then if your seat is higher than amount of doors you should be kicked from vehicle with triggering the proper event.

Pirulax:
I remember fixing this issue, but didn't commit it for whatever reason..
Until then, wrap them out of the vehicle yourself.

Fernando-A-Rocha:
Issue can be closed AFAIK

PR #2247, opened 2021-05-30
Fix #2166 and #2167: Use parent model ID of non-standard vehicle model

### LINKED ISSUE #2166: Cannot get in as passenger (G) inside vehicles made using engineRequestModel
**Describe the bug**
Pressing G to get inside add-on vehicles created with the feature of engineRequestModel does not work

**My add-on method**
-- on client side i request new IDs and use them to replace as addon vehicles
-- i create the vehicle normally on server side and set it's elementdata for example "model", "ferrari"
-- back on client side by that previous element data trigger, i fetch the newly generated vehicle ID by that model string "ferrari" and change the vehicle model

-- all in all, everything works nicely so far, the added vehicles work fine, the only issue i faced is not being able to get in them as passenger, 

-- i did a test, while keeping my code EXACTLY as it is, and only disabling the line that changes vehicle model (setElementModel)
solves the issue, so to help the mta dev team fix this, i can HINT towards that mta thinks that server side created vehicle is now a CLIENT side created vehicle due to the model being replaced

-- TEST number 2 : instead of replacing the elementModel with one of those newly generated IDs i replace it with id 400 (landstalker) and that doesn't cause any Bugs or issues, so i assume the newly vehicle model ids generated by engineRequestModel aren't being recognized as proper vehicles !

Please paste reproduction resources (zip files) directly into this section.
-- i unfortunately didn't make any reproduction files

**Expected behaviour**
-- when i press G on server side created vehicles that have their model changed client side to a newly generated ID (using engineRequestModel) to get in as passenger normally

**Version**
- Client v1.5.8-release-20839
- Server v1.5.8-release-20802

- Windows 10

### LINKED ISSUE #2167: getVehicleType does not work on vehicles made using engineRequestModel
**Describe the bug**
i suspect this bug is related to my previous issue #2166  
but to describe it simply, 

vehicles made using newly generated ids by engineRequestModel don't return the proper vehicle type

**Expected behaviour**
it should return for example "Automobile" or returns the type of it's inherited / cloned vehicle id

**Screenshots**
image 1 : showing the banshee working just fine
banshee

image 2 : showing it return EMPTY "" when trying with added vehicle (not replaced) via #2166 
s15

**Code i used to check the values**
addEventHandler("onClientVehicleStartEnter", root, function( p )
	iprint( "Vehicle Type "..source.vehicleType )
end)

**Version**
Please provide the following information:
- server/client version v1.5.8-release-20839

BCG2000:
Make CClientVehicleManager::GetMaxPassengerCount() and CVehicleNames::GetVehicleTypeName() use parent model ID if used on a vehicle added with engineRequestModel().

- I think making functions like these inherit from the parent model is a good default. It shouldn't cause issues if functions to change this are added in the future.

Example of fix working:
unknown

Test resource:
[erm_fix1test.zip](https://github.com/multitheftauto/mtasa-blue/files/6566948/erm_fix1test.zip)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicleManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientVehicleManager.cpp
@@ -470,6 +470,10 @@ eClientVehicleType CClientVehicleManager::GetVehicleType(unsigned long ulModel)
 
 unsigned char CClientVehicleManager::GetMaxPassengerCount(unsigned long ulModel)
 {
+    // Use parent model ID for non-standard vehicle model IDs.
+    if ((ulModel < 400 || ulModel > 611) && IsValidModel(ulModel))
+        ulModel = g_pGame->GetModelInfo(ulModel)->GetParentID();
+
     // Valid model?
     if (IsStandardModel(ulModel))
     {
--- a/Client/mods/deathmatch/logic/CVehicleNames.cpp
+++ b/Client/mods/deathmatch/logic/CVehicleNames.cpp
@@ -287,6 +287,10 @@ unsigned int CVehicleNames::GetVehicleModel(const char* szName)
 
 const char* CVehicleNames::GetVehicleTypeName(unsigned long ulModel)
 {
+    // Use parent model ID for non-standard vehicle model IDs.
+    if ((ulModel < 400 || ulModel > 611) && CClientVehicleManager::IsValidModel(ulModel))
+        ulModel = g_pGame->GetModelInfo(ulModel)->GetParentID();
+
     // Check whether the model is valid
     if ((IsValidModel(ulModel) || IsModelTrailer(ulModel)) && ((ulModel - 400) < NUMELMS(ucVehicleTypes)))
     {


Allerek:
I don't think its that important for now to introduce, as we're discussing a new system for `custom models id` on discord.

BCG2000:
I did look at the discussion on Discord now and it doesn't seem like a solution will be decided any soon. I think being able to at least enter added vehicles as passenger and get their type correctly is helpful for those who are experimenting and building with this feature already, even if it is not fully finished yet. It lets them use added cars and bikes without drawbacks, even if the model sync has to be done manually.

BCG2000:
I added a test resource to the OP to make this easier to verify now (should have done it in the beginning).

SDabdlmounaim96:
> I added a test resource to the OP to make this easier to verify now (should have done it in the beginning).

i hope like you said happens, am currently heavily relying on added vehicles even as an experimental features, in hopes of it eventually becoming more stable by the time am near finishing my project !

BCG2000:
Apologies, I tried to rename my branch to have a more appropriate name, didn't know it would delete it on the PR and close it... I'm new to this whole GitHub thing, I had the account for years but never really used it until now.

PlatinMTA:
It would be nice to actually have this merged soon. A few months since this pr and we still cant get a full use of the new allocation for vehicles, only for objects and peds. At least the handling works.

Xenius97:
When comes this fix? :)

Lpsd:
> When comes this fix? :)

When someone decides to test, review and any required changes are made - could be today, tomorrow, next month or never (aka, no specified date/time).

Addlibs (approved):
Looks good to me, tested with two players, no issues regarding passenger support, except #1833 is also present when switching between and to and from custom model IDs if seat becomes non-existent due to the change.

Pirulax:
Well, okay, #1833 is irrelevant.
I'll just merge this, and see what happens.
This also fixes a network trouble, so thats why I want to merge it asap.

PR #2250, opened 2021-06-01
Refactor of CPlayerManager::IsValidPlayerModel() based on issue #2244

### LINKED ISSUE #2244: Change GetValidPedModels to use a static array
Transferred from #2242 - the original "issue" was incorrect (getValidPedModels does not return 273 as claimed), however the MTA code still made me sigh.

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp#L175-L181

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Server/mods/deathmatch/logic/CPlayerManager.cpp#L323-L329

As suggested by @Pirulax, we should probably use a static array instead.

delinx:
Fixes #2244
Hey, you meant this way of using static array? Let me know if this way of doing it works for you aesthetically, otherwise it behaves exactly the same as original IsValidPlayerModel based on the test I run.

### DIFF
--- a/Server/mods/deathmatch/logic/CPlayerManager.cpp
+++ b/Server/mods/deathmatch/logic/CPlayerManager.cpp
@@ -320,12 +320,15 @@ void CPlayerManager::Broadcast(const CPacket& Packet, const std::multimap<ushort
     DoBroadcast(Packet, groupMap);
 }
 
+const bool CPlayerManager::ValidPlayerModels[313] = {1,1,1,0,0,0,0,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1};
+
 bool CPlayerManager::IsValidPlayerModel(unsigned short usPlayerModel)
 {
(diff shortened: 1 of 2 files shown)


Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -320,12 +320,15 @@ void CPlayerManager::Broadcast(const CPacket& Packet, const std::multimap<ushort
...
-    return (usPlayerModel == 0 || usPlayerModel == 1 || usPlayerModel == 2 || usPlayerModel == 7 ||
-            (usPlayerModel >= 9 && usPlayerModel != 208 && usPlayerModel != 149 && usPlayerModel != 119 && usPlayerModel != 86 && usPlayerModel != 74 &&
-             usPlayerModel != 65 && usPlayerModel != 42 && usPlayerModel <= 272) ||
-            (usPlayerModel >= 274 && usPlayerModel <= 288) || (usPlayerModel >= 290 && usPlayerModel <= 312));
+    if (usPlayerModel < 0 || usPlayerModel > 312)
+    {
+        return false;
+    }
+    return (ValidPlayerModels[usPlayerModel]);
 }
```
Pirulax:
```suggestion
bool CPlayerManager::IsValidPlayerModel(unsigned short model)
{
    if (model > 312)
        return false; // TODO: On client side maybe check if a model was allocated with engineRequestModel and it is a ped
        
    switch (model)
    {
        case 3:
        case 4:
        case 5:
        case 6:
        case 8:
        case 42:
        case 65:
        case 74:
        case 86:
        case 119:
        case 149:
        case 208:
        case 273:
        case 289:
            return false;
        default:
            return true;
    }
}
```

I think this solution is the cleanest. 
We could add a function that returns an array (on server side a static one (eg.: `std::array<unsigned short, 300>`), on client we'd ideally also include the engineRequestModel ones as well).

Server/mods/deathmatch/logic/CPlayerManager.h
```diff
@@ -66,6 +66,8 @@ class CPlayerManager
     void AddToList(CPlayer* pPlayer);
     void RemoveFromList(CPlayer* pPlayer);
 
+    const static bool ValidPlayerModels[];
+
```
Pirulax:
```suggestion
```
Merge this together with the above suggestion.

Pirulax (changes requested):

Pirulax:
I'm not sure if we should fix the client side now as well (as it doesnt include `engineRequestModel` models yet..) 
Also thank you very much for the PR, refactors are always appriciated.
Also very happy to see new contributors. 

Also, nearly forgot. I think you have to add this on the client side as well.

Server/mods/deathmatch/logic/CPlayerManager.cpp
Lpsd: I also prefer this for readability

Lpsd:
> Also, nearly forgot. I think you have to add this on the client side as well.

The clientside version of `getValidPedModels` uses `CClientPlayerManager::IsValidModel()`, here:

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/mods/deathmatch/logic/CClientPlayerManager.cpp#L179-L187

Seems fine. The clientside is also already doing a check for models allocated with `engineRequestModel`, inside the `GetValidPedModels` method:

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp#L29-L36

We should try to also include a check for models allocated with `engineRequestModel` on the serverside too, but that's for another PR imo.

LGTM :shipit:

Pirulax:
Server-side doesnt yet know about client side models, so.. Theres that.

Pirulax (approved):
gut

Pirulax:
Thank you for the PR.

PR #2253, opened 2021-06-05
Add a UI button to delete servers from the "Recent" tab

stoneage-mta:
This PR adds a button to the server browser's interface that allows us to remove servers from the `Recent` tab (this feature was added in PR [#1381](https://github.com/multitheftauto/mtasa-blue/pull/1381) but was only possible to use it by pressing the Backspace button and also we didn't have any info or hint showing that it was possible.

Click [here](https://streamable.com/lvyv4p) to see a video showing how it works.

_Please let me know if you find a better icon._


ghost:
If possible you might add that selection doesn't reset, so player won't need to select server each time button is pressed - this would be helpful when removing few or more servers.

stoneage-mta:
> If possible you might add that selection doesn't reset, so player won't need to select server each time button is pressed - this would be helpful when removing few or more servers.

I just did it, but in my opinion, it made it very easy to miss-click and delete an unwanted server, so i think its better as it is now.

Pirulax:
Maybe add a confirmation window?

Lpsd:
A confirmation window is kinda annoying, especially if you're trying to clear quite a few from the list.

This might be better to wait for the CEGUI update, where we have tooltips. For example, we could require a double click on the delete button, and when you hover over the button (or maybe after you attempt to click it once) there's a tooltip saying "double click to confirm removal"

Then we could also keep the row selected after deleting, as @srslyyyy suggested, since it's unlikely you're going to accidentally double click again.

An idea for another feature which ties in with this would be a "date last played" column, you might want to sort by that and clear servers you haven't played on for a while. That's another discussion though :)

patrikjuvonen:
This looks good enough for now in my opinion! I'll do a proper review and testing today.

> This might be better to wait for the CEGUI update, where we have tooltips. For example, we could require a double click on the delete button, and when you hover over the button (or maybe after you attempt to click it once) there's a tooltip saying "double click to confirm removal"

The CEGUI update is many miles away, we shouldn't hold back too much on updating the menus just slightly where possible.

patrikjuvonen (approved):
Looks good and tested working!

I made a small change to the initial alpha value for the icon, since it was lit up before the server list had loaded.

Thanks and congratulations on your first pull request! 🎉

PR #2264, opened 2021-06-17
Increase streaming memory setting limit

Pirulax:
New limits:
```
// Streaming memory range based on system installed memory:
//
//     System RAM MB     min     max
//           512     =   64      96
//          1024    =   96     128
//          2048    =  128     256
//          4096    =  128     384
//          6146    =  128     512
```

~~Untested, but probably works as expected.`~~ Tested


Dutchman101:
Much more about this was said in the MTA development discord (e.g #streaming-memory channel), including that by now Pirulax tested it and it seemed to work properly.

I also don't see a problem with this, as long the existing older code that this "dynamically decide how much VRAM is a good idea" will work properly with these new limits (as in: does it not need a refactor or further modernisation rather than simply adding higher limits) - you'll probably need a final debug build test where you load heavy mods and stretch streaming memory usage all the way up to its new limits.

The main concern for previous calls to raise streaming memory limits, is for users with on-board GPU (where as the demand by a game grows it will draw more and more shared memory from available system RAM, leading to a risk for out-of-mem crashes). 

This code looks like the thing I proposed elsewhere last week, to get around that concern: to let MTA dynamically assign a safe amount of streaming memory, based on the available amount of system RAM (to accept the chance player has on-board GPU). Against my expectations it already existed in the shape of this old code, which by the way got added in 810bf9c and when it got introduced initially it supported 512MB streaming memory for users with 2GB RAM, and that later got nerfed down to 256MB streaming memory.

I guess the reason for that was reconsideration (just being more careful), but also obviously back then (in 2012) the average PC had different specs and therefore I think that the updated table (per the changes in this PR), which currently looks like this:

```
// Streaming memory range based on system installed memory:
//
//     System RAM MB     min     max
//           512     =   64      96
//          1024     =   96     128
//          2048     =  128     256
```

And in its updated form:
```
// Streaming memory range based on system installed memory:
//
//     System RAM MB     min     max
//           512     =   64      96
//          1024    =   96     128
//          2048    =  128     256
//          4096    =  128     512
//          6146    =  128     768
//          8192    =  128     1024
```

Looks way more appropiate for this age, today's PC specs.

@Pirulax you just forgot to update the table as well ([here: CCore.cpp @ 1912](https://github.com/multitheftauto/mtasa-blue/blob/402d559124460933a5a97fa68c0e39b071236e67/Client/core/CCore.cpp#L1912) along with the changes this PR makes

Pirulax:
Yeah, the table, thanks!

The 1 GiB @ 8 GiB RAM might be too much, because heres a thing: MTA likes keeping models in memory (see [CModelCacheManager](github.com/multitheftauto/mtasa-blue/blob/master/Client/core/CModelCacheManager.cpp)).
Also, GTA doesn't seem to unstream models unless the memory limit is reached, which basically means that.. The memory available is likely to be used all up, which by itself woudln't be an issue if GTA was x64, but it isn't which means we have a finite amount of address space.

As of now these limits are prefectly fine (for the vanilla game), and an increase might lead to crashes (as people are likely to max it out, because they think, since they have lots of RAM, they can max it out, but thats not the full picture) as it might push over the gap on some servers.

So, I propse to lower the maximums a little:
```
// Streaming memory range based on system installed memory:
//
//     System RAM MB     min     max
//          512     =   64      96
//          1024    =   96     128
//          2048    =  128     256
//          4096    =  128     384
//          6146    =  128     512
```

I created this PR with #1677 in mind, as thats where the higher stream memory is basically a must, but for that we should add a function (as discussed earlier), because allowing people to max this setting out might cause more issues than good.

Dutchman101:
> The 1 GiB @ 8 GiB RAM might be too much

Maybe you should reduce the new maximum to 512MB or 768MB streaming memory. I would go for the latter.

Or maybe first go for 512MB, see the metrics & issue reports and if everything turns out fine (no increase in crashes) you can finally raise it to 768MB and see again.

Well, i just saw your edit:

> So, I propose to lower the maximums a little:
> 
> ```
> // Streaming memory range based on system installed memory:
> //
> //     System RAM MB     min     max
> //           512     =   64      96
> //          1024    =   96     128
> //          2048    =  128     256
> //          4096    =  128     384
> //          6146    =  128     512
> ```

Looks good as well i guess

PlatinMTA:
Could you guys add another metric? 16GB RAM -> up to 1GB on streaming memory

This shouldn't affect 16GB users and would be appreciated.

BTW thank you for this pull!

Pirulax:
Heck, I could add that even with 8 GB, but there are diminishing returns after 800 megs. 1024 streaming memory is 4 times the amount vanilla GTA uses. 

If GTA was 64 bits I'd be glad to increase the limit to a gig, but wasting 1 / 4 of total available memory to streaming isn't worth it.

Dutchman101:
We should keep adding support for 768MB streaming memory in mind, if it later turns out no one had issues with the maximum of 512MB that this PR (after last changes) raises it to.

Pirulax:
I dont see how the vanilla game would benefit from 768, but sure, worst case scenario we revert the pr.

Pirulax:
I think we should just go ahead and merge it.
Just remember that users will have to manually increase the slider, as the setting stores absolute values.. So nightlies wont be too effective in catching stuff possible issues..

patrikjuvonen (approved):
Looks good to me.

Dutchman101:
> Just remember that users will have to manually increase the slider, as the setting stores absolute values.. So nightlies wont be too effective in catching stuff possible issues..

Do you mean people with existing installations (and thus config) being bound to 128MB unless they change it after the update? In that case, i guess it will use the dynamically calculated amount of streaming memory for new installations (following the table of new limits)?

If it's like that, it can be worth to clear the values with the update, for all users, once

PR #2265, opened 2021-06-18
Add TotalPhysicalMemory to dxGetStatus

Pirulax:
Allow scripters to get access info about available total physical memory. There's a very slight difference between the actual physicals memory and this (since there's some memory reserved for the hardware (exact value can be seen in task manager))

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
@@ -9,9 +9,13 @@
  *
  *****************************************************************************/
 
+
 #include "StdInc.h"
 #include "CLuaDefs.h"
 #include "lua/CLuaFunctionParser.h"
+
+#include <SharedUtil.SysInfo.h>
+#include <SharedUtil.SysInfo.hpp>
 
 #define MIN_CLIENT_REQ_DXSETRENDERTARGET_CALL_RESTRICTIONS "1.3.0-9.04431"
 extern bool g_bAllowAspectRatioAdjustment;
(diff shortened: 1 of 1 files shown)


Dutchman101:
If PR #2267 gets accepted, it should probably supersede this (there'd be no need for also getting RAM in `dxGetStatus`).

But if for some reason that PR can't pass, this PR (adding it to `dxGetStatus`) will be the minimum feature required to implement PR #1677 (Allow load custom IMG containers).

Pirulax:
Im going to leave this one here, just in case. But the other one should be more than enough.

Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
```diff
@@ -1705,6 +1709,11 @@ int CLuaDrawingDefs::DxGetStatus(lua_State* luaVM)
         lua_pushstring(luaVM, "SettingHighDetailPeds");
         lua_pushboolean(luaVM, dxStatus.settings.bHighDetailPeds);
         lua_settable(luaVM, -3);
+
+        lua_pushstring(luaVM, "TotalPhysicalMemory");
+        lua_pushnumber(luaVM, (lua_Number)(SharedUtil::GetWMITotalPhysicalMemory() / 1024 / 1024));
```
botder:
```suggestion
        lua_pushnumber(luaVM, static_cast<lua_Number>(SharedUtil::GetWMITotalPhysicalMemory()) / 1024.0 / 1024.0);
```
`SharedUtil::GetWMITotalPhysicalMemory()` returns a `long long` integer value, you must promote it to a double (lua_Number) before division.

Pirulax: [C++ standard](https://en.cppreference.com/w/cpp/language/operator_arithmetic) says it will be promoted automatically.
But, I'll do it anyways, as it might generate warnings.

botder (approved):

Issue #1563, opened 2020-07-17
Allow clients to get process ram usage and left amount of ram with functions


Shuubaru:
**Is your feature request related to a problem? Please describe.**
In servers with a really big amount of mods (vehicles, objects, etc) it's really common to crash due to the excessive amount of ram usage. From my tests, when GTA SA ram usage is over 3Gbs it becomes really unstable and tends to crash when attempt to load a file with engine() functions. I have also seen that if you run out of ram you will get a warn in the debug and engine functions will fail to execute, but you cannot check if the function failed since if you include it into another function, the error will stop the execution of that function.

**Describe the solution you'd like**
There are two features I'd like to see in MTA:
1.- Alert in debug if your PC **is about** to run out of ram, or if GTA SA ram usage is **over 2Gbs** (or 2.5).
2.- Allow clients to get left ram and ram usage, with functions like getProcessRamUsage() and getFreeRamAmount() (just an example name) , so, scripters will be able to tell players to disable some mods if, for example, there is a mod loader, or simply prevent from loading a mod if player have no enough ram, or if he is using more than certain amount of ram already.

**Describe alternatives you've considered**
Making a table with the size of every mod you have in the server would be a good option. But with this you can't calculate the amount of ram the client will have on, for example, 1 hour later.

**Additional context**
- This is probably a solution for servers that are using mod loaders, we can prevent crashes by only preventing players to enable more mods that they can handle. Currently it's possible to get the amount of free VRAM and even VRAM usage with [dxGetStatus](https://wiki.multitheftauto.com/wiki/DxGetStatus) , so, a similar option with RAM will be awesome.

Pirulax:
That would be some kind of WinAPI call probably. The problem with "free" RAM is that, you might have 500 mb of it, but it might be fragmented, so you can't even load a 7mb file, since there isnt enough contiguous memory

PR #2267, opened 2021-06-18
Add getSystemInfo function

Pirulax:
This function returns basic information about the client's system in a table.
Example:
```lua
{
    CPU = {
        MaxClockSpeed = 3200,
        NaNumberOfCores = 6,
        Name = "AMD Ryzen 5 1600 Six-Core Processor",
        NumberOfLogicalProcessors = 12
    },
    GPU = {
        Name = "Radeon RX 570 Series",
        RAM = 419328
    },
    TotalPhysicalMemory = 16727152 -- in bytes. I actually think it should be in MB (as the VRAM size)
}
```

Note: First call to this function is veeeeeery slow, as it has to query WMI. 
Subsequent calls are fast, as static variables are used to cache the data.

To clarify, this PR adds:
- CPU info (table under CPU key)
- `TotalPhysicalMemory`
- GPU info is just here for convenience (already available from dxGetStatus)

### DIFF
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
@@ -725,11 +725,12 @@ int CLuaUtilDefs::tocolor(lua_State* luaVM)
 std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>>
 CLuaUtilDefs::GetSystemInfo()
 {
-    static std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>> outValue;
+    using KVTable = std::unordered_map<std::string, std::variant<std::string, lua_Number>>;
+    static std::unordered_map<std::string, std::variant<KVTable, lua_Number>> outValue;
     if (outValue.empty()) {
         auto info = SharedUtil::GetWMISystemInfo();
         {
-            outValue.emplace("CPU", std::unordered_map<std::string, std::variant<std::string, lua_Number>>{
+            outValue.emplace("CPU", KVTable{
                 {"MaxClockSpeed", (lua_Number)info.CPU.MaxClockSpeed},
                 {"Name", info.CPU.Name},
                 {"NaNumberOfCores", (lua_Number)info.CPU.NumberOfCores},
@@ -737,6 +738,15 @@ CLuaUtilDefs::GetSystemInfo()
             });
         }
         outValue["TotalPhysicalMemory"] = (lua_Number)info.TotalPhysicalMemory;
+
+        {
+            BasicGPUInfo gpu;
+            g_pCore->GetGraphics()->GetRenderItemManager()->GetBasicGPUInfo(gpu);
+            outValue.emplace("GPU", KVTable{
+                { "RAM", (lua_Number)gpu.memoryKB},
+                { "Name", gpu.name },
+            });
+        }
     }
     return outValue;
 }
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
@@ -15,6 +15,42 @@
 #endif
 #include <lua/CLuaFunctionParser.h>
 
+#ifdef MTA_CLIENT
+namespace mta
+{
+const auto& GetSystemInfo()
+{
+    using DataSubTable = std::unordered_map<std::string, std::variant<std::string, lua_Number>>;
+    using MainTable = std::unordered_map<std::string, std::variant<DataSubTable, lua_Number>>;
+
+    static MainTable outValue{};
+    if (outValue.empty()) {
+        {
+            auto info = SharedUtil::GetWMISystemInfo();
+            outValue.emplace("CPU", DataSubTable{
+                {"MaxClockSpeed", (lua_Number)info.CPU.MaxClockSpeed},
+                {"Name", info.CPU.Name},
+                {"NaNumberOfCores", (lua_Number)info.CPU.NumberOfCores},
+                {"NumberOfLogicalProcessors", (lua_Number)info.CPU.NumberOfLogicalProcessors},
+            });
+            outValue["TotalPhysicalMemory"] = (lua_Number)info.TotalPhysicalMemory;
+        }
+
+        {
+            BasicGPUInfo gpu;
+            g_pCore->GetGraphics()->GetRenderItemManager()->GetBasicGPUInfo(gpu);
+            outValue.emplace("GPU", DataSubTable{
+                { "RAM", (lua_Number)gpu.memoryKB},
+                { "Name", gpu.name },
+            });
+        }
+
+    }
+    return outValue;
+}
+};
+#endif
+
 void CLuaUtilDefs::LoadFunctions()
 {
     constexpr static const std::pair<const char*, lua_CFunction> functions[]{
@@ -53,7 +89,7 @@ void CLuaUtilDefs::LoadFunctions()
         {"tocolor", tocolor},
 
 #ifdef MTA_CLIENT
-        {"getSystemInfo", ArgumentParser<GetSystemInfo>}
+        {"getSystemInfo", ArgumentParser<mta::GetSystemInfo>}
 #endif
     };
 
@@ -720,34 +756,3 @@ int CLuaUtilDefs::tocolor(lua_State* luaVM)
     lua_pushnumber(luaVM, static_cast<lua_Number>(ulColor));
     return 1;
 }
-
-#ifdef MTA_CLIENT
-std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>>
-CLuaUtilDefs::GetSystemInfo()
-{
-    using KVTable = std::unordered_map<std::string, std::variant<std::string, lua_Number>>;
-    static std::unordered_map<std::string, std::variant<KVTable, lua_Number>> outValue;
-    if (outValue.empty()) {
-        auto info = SharedUtil::GetWMISystemInfo();
-        {
-            outValue.emplace("CPU", KVTable{
-                {"MaxClockSpeed", (lua_Number)info.CPU.MaxClockSpeed},
-                {"Name", info.CPU.Name},
-                {"NaNumberOfCores", (lua_Number)info.CPU.NumberOfCores},
-                {"NumberOfLogicalProcessors", (lua_Number)info.CPU.NumberOfLogicalProcessors},
-            });
-        }
-        outValue["TotalPhysicalMemory"] = (lua_Number)info.TotalPhysicalMemory;
-
-        {
-            BasicGPUInfo gpu;
-            g_pCore->GetGraphics()->GetRenderItemManager()->GetBasicGPUInfo(gpu);
-            outValue.emplace("GPU", KVTable{
-                { "RAM", (lua_Number)gpu.memoryKB},
-                { "Name", gpu.name },
-            });
-        }
-    }
-    return outValue;
-}
-#endif
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
@@ -9,6 +9,11 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#ifdef MTA_CLIENT
+#include <SharedUtil.SysInfo.h>
+#include <SharedUtil.SysInfo.hpp>
+#endif
+#include <lua/CLuaFunctionParser.h>
 
 void CLuaUtilDefs::LoadFunctions()
 {
@@ -46,6 +51,10 @@ void CLuaUtilDefs::LoadFunctions()
         // Utility functions
         {"gettok", GetTok},
         {"tocolor", tocolor},
+
+#ifdef MTA_CLIENT
+        {"getSystemInfo", ArgumentParser<GetSystemInfo>}
+#endif
     };
 
     // Add functions
@@ -711,3 +720,24 @@ int CLuaUtilDefs::tocolor(lua_State* luaVM)
     lua_pushnumber(luaVM, static_cast<lua_Number>(ulColor));
     return 1;
 }
+
+#ifdef MTA_CLIENT
+std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>>
+CLuaUtilDefs::GetSystemInfo()
+{
+    static std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>> outValue;
+    if (outValue.empty()) {
+        auto info = SharedUtil::GetWMISystemInfo();
+        {
+            outValue.emplace("CPU", std::unordered_map<std::string, std::variant<std::string, lua_Number>>{
+                {"MaxClockSpeed", (lua_Number)info.CPU.MaxClockSpeed},
+                {"Name", info.CPU.Name},
+                {"NaNumberOfCores", (lua_Number)info.CPU.NumberOfCores},
+                {"NumberOfLogicalProcessors", (lua_Number)info.CPU.NumberOfLogicalProcessors},
+            });
+        }
+        outValue["TotalPhysicalMemory"] = (lua_Number)info.TotalPhysicalMemory;
+    }
+    return outValue;
+}
+#endif
--- a/Shared/sdk/SharedUtil.SysInfo.h
+++ b/Shared/sdk/SharedUtil.SysInfo.h
@@ -29,7 +29,7 @@ namespace SharedUtil
     struct WMISystemInfo
     {
         WMIProcessorInfo CPU = {};
-        uint64_t TotalPhysicalMemory = 0; 
+        uint64_t TotalPhysicalMemory = 0; // In KiB
     };
 
     struct SQueryWMIResult : public std::vector<std::vector<SString> >
--- a/Shared/sdk/SharedUtil.SysInfo.h
+++ b/Shared/sdk/SharedUtil.SysInfo.h
@@ -18,6 +18,20 @@
 
 namespace SharedUtil
 {
+    struct WMIProcessorInfo
+    {
+        uint32_t MaxClockSpeed = 0;
+        std::string Name = {};
+        uint32_t NumberOfCores = 0;
+        uint32_t NumberOfLogicalProcessors = 0;
+    };
+
+    struct WMISystemInfo
+    {
+        WMIProcessorInfo CPU = {};
+        uint64_t TotalPhysicalMemory = 0; 
+    };
+
     struct SQueryWMIResult : public std::vector<std::vector<SString> >
     {
     };
@@ -40,6 +54,7 @@ namespace SharedUtil
     SString      GetWMIOSVersion();
     unsigned int GetWMIVideoAdapterMemorySize(const unsigned long ulVen, const unsigned long ulDev);
     long long    GetWMITotalPhysicalMemory();
+    WMISystemInfo GetWMISystemInfo();
     void         GetWMIAntiVirusStatus(std::vector<SString>& outEnabledList, std::vector<SString>& outDisabledList);
     void         GetInstalledHotFixList(std::vector<SString>& outInstalledList);
     bool         IsHotFixInstalled(const SString& strHotFixId);
--- a/Client/core/Graphics/CRenderItemManager.cpp
+++ b/Client/core/Graphics/CRenderItemManager.cpp
@@ -711,6 +711,19 @@ void CRenderItemManager::UpdateMemoryUsage()
 
 ////////////////////////////////////////////////////////////////
 //
+// CRenderItemManager::GetBasicGPUInfo
+//
+//
+//
+////////////////////////////////////////////////////////////////
+void CRenderItemManager::GetBasicGPUInfo(BasicGPUInfo& outInfo)
+{
+    outInfo.name = m_strVideoCardName;
+    outInfo.memoryKB = m_iVideoCardMemoryKBTotal;
+}
+
+////////////////////////////////////////////////////////////////
+//
 // CRenderItemManager::GetDxStatus
 //
 //
--- a/Client/core/Graphics/CRenderItemManager.h
+++ b/Client/core/Graphics/CRenderItemManager.h
@@ -56,6 +56,7 @@ public:
     virtual void               FlushNonAARenderTarget();
     virtual void HandleStretchRect(IDirect3DSurface9* pSourceSurface, CONST RECT* pSourceRect, IDirect3DSurface9* pDestSurface, CONST RECT* pDestRect,
                                    int Filter);
+    virtual void               GetBasicGPUInfo(BasicGPUInfo& outInfo);
 
     // CRenderItemManager
     void NotifyContructRenderItem(CRenderItem* pItem);
--- a/Client/multiplayer_sa/CMultiplayerSA_Rendering.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Rendering.cpp
@@ -362,7 +362,7 @@ static void _declspec(naked) HOOK_CTimer_Update()
     {
         popad
         mov     ecx,dword ptr ds:[0B7CB28h]
-        jmp     CONTINUE_CTimer_UpdateEsz
+        jmp     CONTINUE_CTimer_Update
     }
 }
 
--- a/Client/multiplayer_sa/CMultiplayerSA_Rendering.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Rendering.cpp
@@ -362,7 +362,7 @@ static void _declspec(naked) HOOK_CTimer_Update()
     {
         popad
         mov     ecx,dword ptr ds:[0B7CB28h]
-        jmp     CONTINUE_CTimer_Update
+        jmp     CONTINUE_CTimer_UpdateEsz
     }
 }
 
--- a/Client/sdk/core/CRenderItemManagerInterface.h
+++ b/Client/sdk/core/CRenderItemManagerInterface.h
@@ -81,6 +81,12 @@ enum eDxTestMode
     DX_TEST_MODE_NO_SHADER,
 };
 
+struct BasicGPUInfo
+{
+    std::string name;
+    uint32_t    memoryKB;
+};
+
 struct SDxStatus
 {
     eDxTestMode testMode;
@@ -179,6 +185,7 @@ public:
     virtual void           FlushNonAARenderTarget() = 0;
     virtual void           HandleStretchRect(IDirect3DSurface9* pSourceSurface, CONST RECT* pSourceRect, IDirect3DSurface9* pDestSurface, CONST RECT* pDestRect,
                                              int Filter) = 0;
+    virtual void           GetBasicGPUInfo(BasicGPUInfo& outInfo) = 0;
 };
 
 ////////////////////////////////////////////////////////////////
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.h
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.h
@@ -52,4 +52,6 @@ public:
     // Utility functions
     LUA_DECLARE(GetTok);
     LUA_DECLARE(tocolor);
+
+    static std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>> GetSystemInfo();
 };
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.h
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.h
@@ -52,6 +52,4 @@ public:
     // Utility functions
     LUA_DECLARE(GetTok);
     LUA_DECLARE(tocolor);
-
-    static std::unordered_map<std::string, std::variant<std::unordered_map<std::string, std::variant<std::string, lua_Number>>, lua_Number>> GetSystemInfo();
 };
--- a/Shared/sdk/SharedUtil.SysInfo.hpp
+++ b/Shared/sdk/SharedUtil.SysInfo.hpp
@@ -319,6 +319,37 @@ long long SharedUtil::GetWMITotalPhysicalMemory()
     return llResult;
 }
 
+SharedUtil::WMISystemInfo SharedUtil::GetWMISystemInfo()
+{
+    // Cache result, because these wont change while the PC is running
+    static WMISystemInfo info;
+
+    static bool queried = false;
+    if (queried)
+        return info;
+    // Only query stuff it havent already (otherwise TotalPhysicalMemory is non zero)
+
+    info.TotalPhysicalMemory = GetWMITotalPhysicalMemory() / 1024 / 1024;
+
+    SQueryWMIResult result;
+    QueryWMI(result, "Win32_Processor", "MaxClockSpeed,Name,NumberOfCores,NumberOfLogicalProcessors");
+
+    if (result.empty()) 
+        return {}; // Something had went wrong
+
+    const auto& cpuInfoResultRow = result[0];
+
+    auto& cpu = info.CPU;
+    cpu.MaxClockSpeed = (uint32_t)atoll(cpuInfoResultRow[0]);
+    cpu.Name = cpuInfoResultRow[1];
+    cpu.NumberOfCores = (uint32_t)atoll(cpuInfoResultRow[2]);
+    cpu.NumberOfLogicalProcessors = (uint32_t)atoll(cpuInfoResultRow[3]);
+
+    queried = true;
+
+    return info;
+}
+
 /////////////////////////////////////////////////////////////////////
 //
 // GetWMIVideoAdapterMemorySize
--- a/Shared/sdk/SharedUtil.SysInfo.hpp
+++ b/Shared/sdk/SharedUtil.SysInfo.hpp
@@ -341,7 +341,7 @@ SharedUtil::WMISystemInfo SharedUtil::GetWMISystemInfo()
 
     auto& cpu = info.CPU;
     cpu.MaxClockSpeed = (uint32_t)atoll(cpuInfoResultRow[0]);
-    cpu.Name = cpuInfoResultRow[1];
+    cpu.Name = cpuInfoResultRow[1].TrimEnd(" "); // Remove whitespace from the end
     cpu.NumberOfCores = (uint32_t)atoll(cpuInfoResultRow[2]);
     cpu.NumberOfLogicalProcessors = (uint32_t)atoll(cpuInfoResultRow[3]);
 
--- a/Shared/sdk/SharedUtil.SysInfo.hpp
+++ b/Shared/sdk/SharedUtil.SysInfo.hpp
@@ -329,7 +329,7 @@ SharedUtil::WMISystemInfo SharedUtil::GetWMISystemInfo()
         return info;
     // Only query stuff it havent already (otherwise TotalPhysicalMemory is non zero)
 
-    info.TotalPhysicalMemory = GetWMITotalPhysicalMemory() / 1024 / 1024;
+    info.TotalPhysicalMemory = GetWMITotalPhysicalMemory() / 1024;
 
     SQueryWMIResult result;
     QueryWMI(result, "Win32_Processor", "MaxClockSpeed,Name,NumberOfCores,NumberOfLogicalProcessors");


Pirulax:
The LuaUtils code is a little bit ugly, but the compiler didnt let me use `const auto&`.
Edit: Fixed. The code is cleaner now.

Dutchman101:
Nice, this will allow scripters to make scripts smarter (estimate how high-end player's PC is, and decide if they will load heavy mods and scripts or limited / more optimized lightweight versions)

Also, in my eyes this function is fine to return basic GPU specs (model name & video mem), with a note on its wiki page saying `For more than just basic GPU information, use dxGetStatus`. Because people would expect a function called `getSystemInfo` returns info about the most important hardware. So i wouln't consider it as much of a duplicate functionality.

botder:
Why exactly is this needed and how would that be used to "optimize" scripts?

AlexTMjugador:
I think it's main use would be to compute some kind of "performance score" for the client PC, and deciding things like what models to load or what code to run based on that score, as @Dutchman101 said. Obviously, you can't trust things like clock speed or available memory alone (fun fact: I have an old single-core Pentium D from around 2005 lying around that goes up to 3.4 GHz, and obviously nobody would claim that is a good CPU nowadays, despite its high clocks), but if you mix them using some kind of model (i.e. "if this CPU has more than 4 cores, is an Intel Core or Ryzen and has at least 3 GHz of base clock speed then it should be good") you could get pretty interesting conclusions.

The alternatives I can see that are possible right now are:

- Benchmark the client PC on the first connection, which provides accurate data, but this information may become stale if the hardware is changed. Also, the benchmark can't be too hard, because lower end PCs may crash when running it due to lack of memory, for example, so it may have trouble telling apart really high end PCs from those in lower ranges. Not to mention that executing a benchmark is not necessarily so quick and energy efficient.
- Introduce settings that, while configured by default to target low end PCs, allow players with beefier PCs to enjoy enhanced visuals and things like that. While this is a nice concept, some players can't really be bothered to understand and tweak the settings, and it can be argued that the first impression a server gives, in terms of the visual quality it offers by default, can make some difference.

All in all, while I concede that this function is not exactly the most privacy-friendly and easy to use thing in the world (I'm scared that some scripters may hardcode a list of hardware names and, in general, use it in a sloppy way and call it a day), I can definitely see some interesting use cases for it. Although maybe MTA can already provide scripters with a "performance score", similarly to what the Linux kernel does when computing the "bogoMIPS" (but hopefully better)?

starfleet001:
This function is a paramount for heavily modified servers. Knowing basic information about player's hardware will allow to greatly improve user experience and possibly even reduce amount of crashes.

Currently I use relatively ugly approaches to handle low-end PCs. The main one would be to wait for "memory overflow" error (if lucky player wasn't crashed before it showed) and then unload models, switch to low-performance mode (optimized for low end PCs) and load again. Knowing information about player's CPU and physical memory will allow to select best suitable loading mode in advance which is way better. It would also allow to create different loading profiles for different category PCs (so not only generalized "low end" / "good enough specs" but more categorized).

Other approaches were tried over the years too. Like selecting "graphics settings" before loading. This one is nearly useless because most players don't tend to read detailed information or benefits of different modes (or even understand what mode will be the best suitable for them) and just click "high" or something similar. And then in some cases it lead to crashes (either due low amount of memory or something else). Fast benchmarking approach was tried too but in some weird cases it froze the client so it was removed basically after a week of public testing. Also benchmark results were not really reliable. I still keep manual override options for advanced players tho (so they could override server selected models and textures loading options).

Another benefits that I see here (besides creating different models loading profiles):
- Smart scripts. Different optimization strategies could be used based on information about hardware. It might be related to complex computations as well as simple things like reducing amount of repeated heavy calls (so lower amount for low-end PC, higher for better PC).
- Providing suggestions based on hardware (CPU / OS tweaks). Currently I use similar approach based on the name of GPU and GPU memory (suggesting optimal or best settings for NVIDIA / AMD user). This is mostly for advanced users but it's proven to be quite useful in providing good user experience. Specially for cases with outdated drivers (there are users with dedicated GPU that is not used by MTA:SA in some cases and re-installing or updating drivers usually fixes it, sometimes GPU settings tweaks are required).

I think it also makes sense to add information about currently available amount of physical memory or / and current memory usage of gta_sa.exe process. It might slow down this function tho (I'm not sure). But it could be useful to know for sure in advance how many models can be replaced without crashing the client or producing overflow errors.

I don't really see any downside here. I wouldn't be too worried about some scripters using it for lazy scripting techniques. There are many ways to create sloppy things right now and this function wouldn't make a difference in this regard IMO.

Anyway, my apologies for this long comment but it's quite an interesting topic and really useful function. It would be nice to see it reviewed, tested and merged soon since it could provide many benefits right now even before IMG related features are integrated.

Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
```diff
@@ -9,6 +9,47 @@
...
+            });
+            outValue["TotalPhysicalMemory"] = (lua_Number)info.TotalPhysicalMemory;
+        }
+
+        {
+            BasicGPUInfo gpu;
+            g_pCore->GetGraphics()->GetRenderItemManager()->GetBasicGPUInfo(gpu);
+            outValue.emplace("GPU", DataSubTable{
+                { "RAM", (lua_Number)gpu.memoryKB},
+                { "Name", gpu.name },
```
patrikjuvonen, line 44: Inconsistent use of whitespace.
```suggestion
                {"RAM", (lua_Number)gpu.memoryKB},
                {"Name", gpu.name},
```

Shared/sdk/SharedUtil.SysInfo.h
```diff
@@ -18,6 +18,20 @@
 
 namespace SharedUtil
 {
+    struct WMIProcessorInfo
+    {
+        uint32_t MaxClockSpeed = 0;
+        std::string Name = {};
+        uint32_t NumberOfCores = 0;
+        uint32_t NumberOfLogicalProcessors = 0;
```
patrikjuvonen, line 26: Indentation is inconsistent.
```suggestion
        uint32_t    MaxClockSpeed = 0;
        std::string Name = {};
        uint32_t    NumberOfCores = 0;
        uint32_t    NumberOfLogicalProcessors = 0;
```

Shared/sdk/SharedUtil.SysInfo.h
```diff
@@ -18,6 +18,20 @@
...
+    };
+
+    struct WMISystemInfo
+    {
+        WMIProcessorInfo CPU = {};
+        uint64_t TotalPhysicalMemory = 0; // In KiB
```
patrikjuvonen, line 32: Indentation is inconsistent.
```suggestion
        WMIProcessorInfo CPU = {};
        uint64_t         TotalPhysicalMemory = 0; // In KiB
```

Shared/sdk/SharedUtil.SysInfo.h
```diff
@@ -40,6 +54,7 @@ namespace SharedUtil
     SString      GetWMIOSVersion();
     unsigned int GetWMIVideoAdapterMemorySize(const unsigned long ulVen, const unsigned long ulDev);
     long long    GetWMITotalPhysicalMemory();
+    WMISystemInfo GetWMISystemInfo();
```
patrikjuvonen, line 57: Indentation is inconsistent.

Shared/mods/deathmatch/logic/luadefs/CLuaUtilDefs.cpp
```diff
@@ -46,6 +87,10 @@ void CLuaUtilDefs::LoadFunctions()
         // Utility functions
         {"gettok", GetTok},
         {"tocolor", tocolor},
+
+#ifdef MTA_CLIENT
+        {"getSystemInfo", ArgumentParser<mta::GetSystemInfo>}
```
patrikjuvonen, line 92: Missing trailing comma.
```suggestion
        {"getSystemInfo", ArgumentParser<mta::GetSystemInfo>},
```

patrikjuvonen (changes requested):
I stand against this feature until it can provide a way to disable sharing the information to scripts via settings menu.

AlexTMjugador:
> It would be nice to see it reviewed, tested and merged soon since it could provide many benefits right now even before IMG related features are integrated.

Just for the record, in case I didn't state my thoughts clearly on my previous comment, judging by the downvotes and the overall "I would love to see this added and here's why" tone of this comment: I think that this function can be a good addition to MTA. I was just pointing out some considerations that I think are worth thinking about before this is merged in a final release.

starfleet001:
> I stand against this feature until it can provide a way to disable sharing the information to scripts via settings menu.

I probably never gonna understand why some **game** developers so against basic and necessary functionality **for a game**.. Specially when we already had GPU information for years.

Just wanna point out something. Disabling this kind of necessary functionality in the settings would create additional issues and potentially not user friendly behavior. In my case i would probably need to drastically change models loading flow in order to ask the user to share hardware information prior loading. Or simply kick a player (which is a really really bad thing but many servers that will require hardware information will probably do it this way).

Similar to how some servers handle "Allow screen upload" checkbox right now.. I remember in the beginning it wasn't a problem with "screen upload". Just removing some functionality from users with disabled option was enough. But later "render targets" became a part of necessary functionality and i had to force all users to enable this option so they wouldn't lose gameplay features. In this case of course it's not hard to do it without kicks using banner (unlike hardware information which is mostly required prior regular models loading flow).

It's surprising why one of the greatest multiplayers created lacks necessary scripting functionality. Starting from detailed hardware information and ending with client settings (like for example "Audio" volumes). Something is holding us back :(

patrikjuvonen:
> > I stand against this feature until it can provide a way to disable sharing the information to scripts via settings menu.
> 
> I probably never gonna understand why some **game** developers so against basic and necessary functionality **for a game**.. Specially when we already had GPU information for years.
> 
> Just wanna point out something. Disabling this kind of necessary functionality in the settings would create additional issues and potentially not user friendly behavior. In my case i would probably need to drastically change models loading flow in order to ask the user to share hardware information prior loading. Or simply kick a player (which is a really really bad thing but many servers that will require hardware information will probably do it this way).
> 
> Similar to how some servers handle "Allow screen upload" checkbox right now.. I remember in the beginning it wasn't a problem with "screen upload". Just removing some functionality from users with disabled option was enough. But later "render targets" became a part of necessary functionality and i had to force all users to enable this option so they wouldn't lose gameplay features. In this case of course it's not hard to do it without kicks using banner (unlike hardware information which is mostly required prior regular models loading flow).
> 
> It's surprising why one of the greatest multiplayers created lacks necessary scripting functionality. Starting from detailed hardware information and ending with client settings (like for example "Audio" volumes). Something is holding us back :(

Privacy.

starfleet001:
> Privacy.

I understand this point. Thank you. Personally I hardly see any privacy issues with knowing client's hardware. After all, it's a game (multiplayer). And the game is hardware dependent.
Well, I hope that we will get this kind of functionally at least in some way. But in my opinion it would be better without option to disable it (for reasons I mentioned in my previous comment).

AlexTMjugador:
> Personally I hardly see any privacy issues with knowing client's hardware. After all, it's a game (multiplayer). And the game is hardware dependent.

While it can be argued that the device information this function provides is pretty basic, device fingerprinting is an actual concern that has already seen some bad uses. [Wikipedia](https://en.wikipedia.org/wiki/Device_fingerprint) has a good article about the subject. I definitely wouldn't want to introduce a function that may help such activities without even thinking about it.

Besides, making it an option means that you can handle the setting being disabled, as you discussed previously, as you seem fit for your server. It's possible to do the check in a separate resource that loads before the bulk of your resources, so you can take the corresponding action before any model gets loaded, and handle any synchronization via events. Personally, I wouldn't mind such an option being enabled by default, much like CEF is not entirely disabled by default, so I don't think that user-friendliness should be impacted that much for those who seek that "plug and play" experience.

starfleet001:
> Besides, making it an option means that you can handle the setting being disabled, as you discussed previously, as you seem fit for your server. It's possible to do the check in a separate resource that loads before the bulk of your resources, so you can take the corresponding action before any model gets loaded.

Of course it can be handled. But like I mentioned, it will lead to unnecessary steps and complications. Typically users don't like complications (proven in many areas of life and specially in online games).
Also I have seen many projects that decide to kick player instead of explaining at least in some good form what is required from the user. Obviously it's a fault of server developer that such not user friendly way was chosen. But it's avoidable (if there is no option).

Besides, making it optional will not solve device fingerprinting in any way. And for advanced users hardware information spoofing isn't something too complicated. So arguably speaking making it an option is counter productive (if we are talking about potential bad usage).
We already have many things in this game that can be used in a bad manner. I think we should consider amount of benefits in relation to amount of potential bad practices (or something like this). And in this case I believe benefits outweighs.

AlexTMjugador:
> Besides, making it optional will not solve device fingerprinting in any way. And for advanced users hardware information spoofing isn't something too complicated. So arguably speaking making it an option is counter productive (if we are talking about potential bad usage).

Trivially, if a script can't get fingerprintable information at all, every device fingerprinting concern is solved, because device fingerprinting can't happen in the first place. While I agree that spoofing this information solves most concerns related with fingerprinting, I don't think we should make advanced users lives harder just because they are already advanced, the same way we should not make not-so-advanced or privacy-conscious users lives harder by having to explicitly enable such an option.

> We already have many things in this game that can be used in a bad manner. I think we should consider amount of benefits in relation to amount of potential bad practices (or something like this). And in this case I believe benefits outweighs.

This is a fair point. I usually advocate for using each new feature as an opportunity to apply experience or new knowledge that would arguably result in a better software design overall, though, because I fear technical debt creeping up my back quite a bit.

starfleet001:
> Trivially, if a script can't get fingerprintable information at all, every device fingerprinting concern is solved, because device fingerprinting can't happen in the first place.

You are right. Though I was talking about developers not letting users in unless hardware information is shared. So if some server decides to use hardware information with the sole purpose of device fingerprinting then it would be a way to go. This is why I mentioned that option is "counter productive" and I believe there is no perfect solution for this particular problem (if we even consider this as a problem). For example I don't consider this as a problem in 2021 because it's already possible to use different forms of device or in-game behavior fingerprinting without this function (and it can be quite effective).

Shuubaru:
> This function is a paramount for heavily modified servers. Knowing basic information about player's hardware will allow to greatly improve user experience and possibly even reduce amount of crashes.

Unless we can know the amount of ram the game process is using, forget about preventing people from "out of memory" crashes.

ArranTuna:
> 
> 
> I stand against this feature until it can provide a way to disable sharing the information to scripts via settings menu.

Servers can see your IP address and you're worried about them seeing the name of your CPU..?

LosFaul:
to make one thing clear:
hardware information is **NOT** personal information

Pirulax:
Isn't the `serial` the perfect candidate for device fingerprinting? 
This won't change the situation at all. Devices can be fingerprinted by their serial already.
This basic sysinfo won't make it any worse. 
And you can't even spook it (unlike this one, which probably is fairly easy).

patrikjuvonen:
I don't want my GPU information to be read by scripts.

Pirulax:
That's already done by `dxGetStatus`. It isn't a new feature.
This PR just adds CPU info + RAM size.

Dutchman101:
I find this entire privacy concern heavily exaggerated

Your main concern would be device fingerprinting right? Well, the best way for server owners to fingerprint a player is already built into MTA: serials. It can't get worse than that, so having a scripting function that returns CPU and GPU name doesn't do much after that point.

Of course, we can add a client option similar to others (e.g "Allow external sounds" and "Allow screenshots"), but that would be wasted developer effort on this PR as it brings no additional privacy value, and may only complicate a server scripter's job of educating players that they need to allow getting PC specs (either to let the game run properly, or as some servers would do - barr the player unless they allow it, and make a tooltip or something instructing them to enable it. It's all wasted effort, as players like to fiddle with random options without a particular purpose of allowing/disallowing something)

patrikjuvonen:
> That's already done by `dxGetStatus`. It isn't a new feature.

Yes, well that doesn't change my stance on the issue itself. I also don't want my other system information such as CPU information to be read by scripts. I think these privacy improvements should be made to other functions such as dxGetStatus as well.

Serials are a MTA thing, not a global thing.

My main concern was never device fingerprinting (that was someone else's concern). My concern is letting malicious actors have access to data which devices I am using and what are the capabilities that can be exploited in another context.

patrikjuvonen:
> Of course, we can add a client option similar to others (e.g "Allow external sounds" and "Allow screenshots"), but that would be wasted developer effort as it brings no additional privacy value, and may only complicate a server scripter's job of educating players that they need to allow getting PC specs (either to let the game run properly, or as some servers would do - barr the player unless they allow it, and make a tooltip or something instructing them to enable it. It's all wasted effort, as players like to fiddle with random options without a particular purpose of allowing/disallowing something)

An option to toggle this feature literally adds privacy value. This option does not take a lot of effort for a developer to add.

patrikjuvonen:
Whenever you are exposing data of a client you are giving up privacy, which I think is wrong and should be up to the client to decide.

gta191977649:
well, i guess probably the gtaonline also does collect these information too.
I think if the data doesn't contain anything other than hardware itself, should be fine.

Lpsd:
At this point if privacy is a concern when entering a server you might as well nuke MTA from your PC. 

If we're going to start taking a heavy stance on privacy in MTA it needs to be a concerted, consistent effort across the board - otherwise your call for the buck to stop here is just irrelevant. For example, do you think so many people would run MTADiag if they knew what data was shared? Are there even terms and conditions to accept before you run that thing?

I do see valid points on both sides though - why would there be any harm in at least letting the client choose what information he allows the server to see? I see no problem with a checkbox setting "Share my device data" - but again, we need to make that's consistent with all other related hardware-data pulling functions, otherwise it's pointless just changing it here.

Dutchman101:
> My main concern was never device fingerprinting (that was someone else's concern). My concern is letting malicious actors have access to data which devices I am using and what are the capabilities that can be exploited in another context.

I already had some scenario's in mind before you said this, but they have in common that they are extremely far-sought. So i stand with my opinion of heavy exaggeration of the risks.

The ball is probably back at @Pirulax, to see if he wants to include a main menu setting for it. Or else, as per @Lpsd's comment directly above, if it would be inappropiate to make this PR the 'victim' of a change that needs to happen in general.

patrikjuvonen:
I would rather start the privacy enhancement now than never.

Lpsd:
How far down the rabbit hole are we expected to go, though? Do we allow users to mask their IP down the line due to fear of geo-location and tracking?

Dutchman101:
We should probably create a separate issue if we want to discuss this in a general sense. Then if needed we can make that issue a blocker for merging this PR.

patrikjuvonen:
> How far down the rabbit hole are we expected to go, though? Do we allow users to mask their IP down the line due to fear of geo-location and tracking?

For people mindful of their privacy tend to use third-party services such as proxies/VPNs which are allowed in MTA. Hardware spoofing however requires more tooling and I think some of these tools may be banned by MTA in effort to prevent malicious activity (someone with access to the AC can correct me on this specifically).

AlexTMjugador:
> Your main concern would be device fingerprinting right? Well, the best way for server owners to fingerprint a player is already built into MTA: serials. It can't get worse than that, so having a scripting function that returns CPU and GPU name doesn't do much after that point.

This is one of my concerns, although I also share @patrikjuvonen's views on an option being more privacy-friendly, too, even if this is not personal identifiable information. And I personally trust MTA much more at doing a better job at generating a serial for identification purposes using proper hashing and anonymization techniques than any random server I could ever join, which will also be more studied and subject to proper quality standards.

Dutchman101:
> And I personally trust MTA much more at doing a better job at generating a serial for identification purposes using proper hashing and anonymization techniques than any random server I could ever join, which will also be more studied and subject to proper quality standards.

Note that having just basic system information (in this case, the maximum you could get is CPU and GPU name, amount of RAM installed) is far from enough for a bad server owner to create a reliable "unique identifier" system. The combinations are too common. Even if they would narrow it down with alghoritms (for example: regionally by IP range), there would still be way too many users of similar gaming rigs/PC's.

If any new feature were to provide details about more system components (or more detailed per component), then i would no longer call it an exaggeration.

patrikjuvonen:
> Just note that having just basic system information (in this case, the maximum you could get is CPU and GPU name, amount of RAM installed) is far from enough for a bad server owner to create a reliable "unique identifier" system. The combinations are too common.

I'll repeat myself from my previous post:

> My main concern was never device fingerprinting (that was someone else's concern). My concern is letting malicious actors have access to data which devices I am using and what are the capabilities that can be exploited in another context.

Dutchman101:
My opinion changed a little - it doesn't seem to be a big deal if we just add a settings toggle, so the user always has a last say in data being shared with servers. We can satisfy both sides.

If server scripters are really afraid of not meeting their targets of optimization per individual player (or spending additional effort on "forcing" them), they can just send a box to refusers that states: "If you don't allow this in settings, you won't get a good experience because our system will only load LQ textures for everything"

If a player wants to miss out, let them, and use fallback techniques as scripter.

patrikjuvonen:
> If server scripters are really afraid of not meeting their targets of optimization per individual player (or spending additional effort on "forcing" them), they can just send a box to refusers that states: "If you don't allow this in settings, you won't get a good experience because our system will only load LQ textures for everything"
> 
> If a player wants to miss out, let them, and use fallback techniques as scripter.

Eh, sort of off-topic, but I think you are just describing bad UX design. Generally games allow you to download high quality assets separately, and change your quality settings in the graphics settings menu - just like many scripts on MTA do already. Some high profile games use automatic optimization and often times they are not very accurate, but help initial loading times. Nvidia also provides automatic presets for some games. How many actually depend on these automatic settings, I don't know.

The benefit of this feature is primarily for automatic optimization for first time users, or other optimization techniques paired with the graphics settings provided by the user. I don't think the idea itself is bad - just needs to be up to the client to decide whether they prefer a tailored experience out of the box or not.

For example, if a script explicitly wants to ban users who refuse to give up their system information, so be it. I just don't think that's logical.

starfleet001:
> If server scripters are really afraid of not meeting their targets of optimization per individual player (or spending additional effort on "forcing" them), they can just send a box to refusers that states: "If you don't allow this in settings, you won't get a good experience because our system will only load LQ textures for everything"

I previously pointed out that in many ways it's counter-productive to add checkbox like this (like we currently have with "screen upload" and latest "external sounds"). But since we have misguided (in my opinion) privacy protection urge over non personal data, I believe it will benefit everyone if we get this function at least with this checkbox in settings. Hopefully it will not break existing scripts if you plan to affect _dxGetStats()_ with these changes. In my opinion it makes sense for this new function to return false if checkbox is enabled and don't change any existing functions.

Just please don't make it as sloppy as "external sounds" checkbox. In my case some players accidently turn it off and then wondering why external radio stations in game stopped working. It happens more often than you might think. Currently there is no option for scripts to check if user is blocking external sounds or not because related events are called only at the moment of user clicking this checkbox.

botder:
This is pretty off-topic, but what if we add a slider or combobox to settings (and let the MTA client or netc determine the default value), which acts as peformance/quality selection for the user. And the client and server can only access that specific value (let's say 1-3? like Performance / Balanced / Quality)

Similiar to this page in NVIDIA control panel (without the preview):
image

patrikjuvonen:
> This is pretty off-topic, but what if we add a slider or combobox to settings (and let the MTA client or netc determine the default value), which acts as peformance/quality selection for the user. And the client and server can only access that specific value (let's say 1-3? like Performance / Balanced / Quality)

This idea seems sensible, and can provide a more consistent result across the board when joining different servers since it is not up to the scripts to decide whether a CPU or GPU is capable or not, but we can make that determination ourselves in core code. As you said though, it is sort of off-topic, and to my understanding does not cater all of these people's use-cases.

starfleet001:
> This is pretty off-topic, but what if we add a slider or combobox to settings (and let the MTA client or netc determine the default value), which acts as peformance/quality selection for the user. And the client and server can only access that specific value (let's say 1-3? like Performance / Balanced / Quality)

If mtasa client determines this value then the accuracy will not be good for most use cases (it was already discussed earlier). It's not only because benchmarks are generally unreliable but also because all servers are different and only the server developer can know what will be the best option for specific set of CPU / GPU, etc.. Plus if there is no detailed information about what CPU is used then this feature can't be used to make scripts smarter and more efficient based on processing power of the client.

And more to this. If user can easily override "preferred quality" then for example I wouldn't use this feature at all because it will be unreliable and easily abused (just like right now other MTA settings are abused by some players with good PC to gain advantage on competitive servers). "Never trust a client, not even your own" (c)

botder:
> because it will be unreliable and easily abused

It's a user-supplied performance/quality suggestion to the scripting-side. It's up to the scripter to decide if it can be abused or not; and you are not forced to use that user setting at all to prevent players from having a competetive advantage of any sort.

> only the server developer can know what will be the best option for specific set of CPU / GPU

The problem with this approach is that every server scripter/owner will have to analyze performance metrics for bazillion of hardware combinations, if you keep adding queryable system information. Instead I would rather prefer the user-supplied preference and an automatically (by MTA client or netc) determined value and then make my decision on that.

> more efficient based on processing power of the client.

You could always measure per-frame time and then make conclusions on that metric - if you have to throttle or if you can increase the workload.

Dutchman101:
With that implementation, limitations would exist for sure (e.g scripter is aware in which way specifically something they want to do will tax the user's system - some performance heavy functions are known to affect only CPU, only GPU, or only video memory - etc) so my opinion is that the scripter should have freedom on whether to be very specific and finetune everything, or use a more generalized "hardware score".

I propose to keep all sides happy
- Implement getSystemInfo _with_ the user in charge ("Allow sharing system information")
- Implement performance score/selection, as a separate feature (new PR)

Scripters can opt between the systems, depending on their needs and how specific they want to be

If we can't implement both, my opinion is to go with a restricted version of getSystemInfo. Being more detailed, scripters will have more possibilities and be more specific, and as long the user is in charge no one has any problems with it.

Pirulax:
Should it be opt-in, or opt-out?
I'd suggest opt-out, as who wants to disable it, will specifically look for it. Just like browsers allowing stuff like this by default.
(Also this way we save a lot of time for players who dont care)

Pirulax:
Replying to botder:
I think in general it's pretty easy to match find out what is good for a given system.
RAM is quite straight forward.
CPU is a little bit more complicated, but still easy with some regex magic..

Check CPU is Intel / AMD. Lets suppose its AMD:
- Check for Ryzen. If Ryzen, check which gen.
- All AMD CPUs other than Ryzen are crap, so no need to check it

Intel:
- Check which Core I it is (3/5/7/9)
- Check gen

Calculate some kind of score. I'd imagine one can google some info about IPC (instructions per clock), and use that...

starfleet001:
> I propose to keep all sides happy
> 
> * Implement getSystemInfo _with_ the user in charge ("Allow sharing system information")
> * Implement performance score/selection, as a separate feature (new PR)
> 
> Scripters can opt between the systems, depending on their needs and how specific they want to be
> 
> If we can't implement both, my opinion is to go with a restricted version of getSystemInfo. Being more detailed, scripters will have more possibilities and be more specific, and as long the user is in charge no one has any problems with it.

Sounds like a good solution. I believe it makes sense to finalize work on getSystemInfo first and merge it because it's basically ready (except for checkbox). Then someone who is interested could implement this new proposal. It's more complicated in general so probably would take some time.

> you are not forced to use that user setting at all to prevent players from having a competetive advantage of any sort

Of course. This is why I mentioned that I personally wouldn't use something like this for performance optimizations or to provide better gaming experience due the nature of high competitive gamemode (which is already full of workarounds due some players abusing existing settings to gain advantage or use bugs).

> The problem with this approach is that every server scripter/owner will have to analyze performance metrics for bazillion of hardware combinations

It's actually not totally true. I mean, you could do it this way but like you said it would be a nightmare.
Getting general specs of CPU is enough to make the assessment of maximum possible performance (this function fully covers it in current implementation). This is what is needed to make scripts smarter in general - to know possible capabilities of CPU. And with the name of CPU it's possible to make exceptions and tune systems as it goes (edge cases). For example I already tune scripts based on the name of GPU and memory information (graphics related features, functions). It's quite effective and allows to provide better gaming experience for players with mid or low range PC.
Also it makes communication with players who experience issues or crashes more easier (from my experience a lot of players don't even know how to get system specs when asked and sometimes struggle with following step-by-step guides...).

> Should it be opt-in, or opt-out?

I think that it should be enabled by default - meaning that by default getSystemInfo returns information. And if user removed checkbox then this function should return false (this way we don't need to add another function to check for the state of checkbox). Otherwise this function is kinda useless in my opinion.

samr46:
It would be nice to add ProcessUsedMemory to this function. From CMemStats: https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/CMemStats.cpp#L461-L463

Or maybe as separate function since we now have engineStreamingGetUsedMemory() which is a part of process memory. I'm not sure what approach would be the most suitable.

We know that client gets unstable and crashes at ~3.3 GB RAM usage. When player disconnects from a server total RAM usage is usually higher than it was before the first connection (memory leaks or some cache is not fully cleared, I'm not sure but it was like this since forever). And it gets higher and higher when player plays on different servers and switches between without restarting MTA.
With this feature it will be possible to prevent some amount of memory related crashes (like checking current usage before replacing lots of models). Should be easy to implement as well.

Shuubaru:
> It would be nice to add ProcessUsedMemory to this function. From CMemStats:
> 
> https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/CMemStats.cpp#L461-L463
> 
> Or maybe as separate function since we now have engineStreamingGetUsedMemory() which is a part of process memory. I'm not sure what approach would be the most suitable.
> 
> We know that client gets unstable and crashes at ~3.3 GB RAM usage. When player disconnects from a server total RAM usage is usually higher than it was before the first connection (memory leaks or some cache is not fully cleared, I'm not sure but it was like this since forever). And it gets higher and higher when player plays on different servers and switches between without restarting MTA.
> With this feature it will be possible to prevent some amount of memory related crashes (like checking current usage before replacing lots of models). Should be easy to implement as well.

I already made a similar proposal like a year ago [https://github.com/multitheftauto/mtasa-blue/issues/1563](https://github.com/multitheftauto/mtasa-blue/issues/1563) and it has been already mencioned a couple of time here, but I see no real interest on implementing it whatsoever. I really hope devs understand the importance of knowing the client process ram usage.

Dutchman101:
> Should it be opt-in, or opt-out?

Let's stay on par with similar options (such as "Allow screen upload" and "Allow external sounds") - enabled by default, and the ability to opt-out. All that matters is that the user has the final say, if they wanted (majority wouldn't) and this will prevent masses of indifferent users from limiting the possibilities of a scripter or creating additional work in asking them to change it.

> And with the name of CPU it's possible to make exceptions and tune systems as it goes (edge cases)

yeah, and working from something like CPU or GPU model name, you can also make a rough estimate of the PC's overall grade (like you'd find a high-end Ryzen CPU only in gaming rigs).. you really don't need all sorts of computations to make script decisions by markers of high-end enough systems for whatever you want to do.

Pirulax:
Sadly calculating the actual amount of memory used is very hard, if not impossible.
There might be a gig of free memory in theory, while in reality, it would crash the moment you try to allocate even 100 mb.
So yeah... Keep this in mind.

jjustns:
Any hope for this ever being added? It would be really useful for optimization.

Pirulax:
Sure, if anyone has the time to work on it.
I sadly dont.

LosFaul:
> ```lua
> {
>     CPU = {
>         MaxClockSpeed = 3200,
>         NaNumberOfCores = 6,
>         Name = "AMD Ryzen 5 1600 Six-Core Processor",
>         NumberOfLogicalProcessors = 12
>     },
>     GPU = {
>         Name = "Radeon RX 570 Series",
>         RAM = 419328
>     },
>     TotalPhysicalMemory = 16727152 -- in bytes. I actually think it should be in MB (as the VRAM size)
> }
> ```

I think better would be:

```lua
{
     CPU = {
          Name = "AMD Ryzen 5 1600",
          PhysicalCores = 6,
          LogicalCores = 12,
          BaseClockSpeed = 3200,
     },
     HostMemory = 4294967296, -- max 4 GB (32 Bit client)
     GPU = {
            Name = "Radeon RX 570 Series"
            Memory = 4294967296,
     }
}
```

Dutchman101:
Anyone that wants to, feel free to adopt this PR

ghost:
I will create a new PR soon. 😄

Dutchman101:
> I will create a new PR soon. 😄

@Nico8340
Are you still planning for it?

ghost:
> @Nico8340 Are you still planning for it?

Yes

ghost:
Bump?

Fernando-A-Rocha:
PR abandoned ?

ghost:
I just saw that there is such a PR with some funny comments on it. I don't get it why some people dragged here the privacy concerns.
If I install a software I give rights to it to read my full detailed system information by default. This is basic.
If I run and use that software I know that everything I do in it (INSIDE) will be logged, this is also basic. Doesn't matter if I don't like it.
If the software wants to protect itself from malicious users, I also accept that it needs some power against those. If it means that I need to give kernel access, then it means that. For example, if another software would like to change some things in the main software's exe while running.

For me, privacy concers came into my mind, when a software does much more than these. When it logs my activities/actions OUTSIDE of the software. When it logs my browsing history for e.g. When it logs my file architecture alongside its content. When it listens to microphone/webcam, etc.
We know that MTA doesn't do these.

Note: There are known aggressive kernel ACs like Vanguard which even collects some of these mentioned sensitive data and there are huge concerns about this process.

If I connect to a server, I know that everyting I do on it, will be logged. YES, even private messages. Server owners, server administrators must have the full power to fight against their malicious users. If you don't want your server owners to read your pms, then use other software for communitacion. But again, if you use discord for example, or messenger, then Facebook and Discord will read your messages. There is no such things like 'no one reads it only me and my partner'.

In my opinion this PR is okay, but lacks of more detailed data.
For example motherboard manufacturer, model, number etc.
RAM manufacturer, model, HDD/SSD manufacturers etc. 
Server owners would still need to make their own fingerprint system which uses these data. With a CPU name only, there couldn't be created a good fingerprint system which was the main concern for some reason. 
I would even allow a getProcessList() function to be made to get the running process of the client's PC to see if they use something suspicious (if the program doesn't hide itself)

Keep in mind that even this: https://wiki.multitheftauto.com/wiki/Fingerprint
exists, but I don't know how this can generate unique hash for every PC. (reading some comment on DC, people says its not unique everytime which I expected to be. Without the power to read system data there is no such thing as fingerprint)


MTA serial system can no longer be trusted. Serials can easily be spoofed into anything. What can we do to prevent cheaters from coming back? Ban by nick? They choose another. Ban by serial? They spoof it. Ban by IP? They use VPN. Implement VPN-Detect API usage? There are VPN IPs which won't be detected by any of the apis on the internet even if you pay for one.


Topic about map/mod optimisaiton: This PR would sufficient their needs as well, however I maybe read something which I misundrstood, but you can't avoid some players to not to download some files if their PC is lowend. On connect every running resource will be automatically downloaded. You can disable certain things to be loaded for e.g textures which would be the benefit of this PR. (maybe I misread something there are huge ammount of comments anyway)

I just would like to read the answers and reason why knowing such hw infos of a user would break user privacy.
For example I use now i5-9300h processor. What do you can do now? Even if I gave my IP, the worst thing could happen is a DDoS attack, as geolocating hardly works well. Especially with cgnat ips.

TracerDS:
Not everyone shares your opinion.
Its very controversial what an app should and should not do. Mta was always focused on privacy.
I wouldnt like someone reading my private messages. Whatsapp has e2e encryption iirc. That makes it somewhat safe that no one apart from me and the user I message will see that message.
I also hope no one backdoored discord to spy on us.

That being said it all depends what will we introduce to mta. Collecting machine info only? Okay, fair.
Collecting **basic** system info such as Windows version, architecture? Fair enough.
Collecting something more? Absolutely not.

ghost:
> Collecting machine info only? Okay, fair. Collecting **basic** system info such as Windows version, architecture? Fair enough. Collecting something more? Absolutely not.

I understand that there are different opinions, thats not the problem. What I don't understand is having a CPU name how can be malicious to a user if a server knows what CPU the player uses?

I comepletely agree with all these you wrote (in quoted) This is what I wrote as well in more detail.
Only difference between us is the private messages. I don't want to argue on it as it would be hugely OFF topic, but I do understand all concerns about it.

botder:
I still don't really see a convincing reason why a script/server/client needs this sort of information. For process memory there is `getProcessMemoryStats`. And please don't tell me you're going to build a huge CPU database/list just to determine the strength of the computer - we could let the user set a *performance* switch in the settings. When it comes to privacy, less information is always better. When it comes to fingerprinting/anti-cheat, returning fake information from Lua functions is child's play - the information is worthless.

starfleet001:
> When it comes to fingerprinting/anti-cheat, returning fake information from Lua functions is child's play - the information is worthless.

There are ways to make it harder for client to fake this information. Implement getSystemInfo on server side. Send packet with client specs to the server on connection as part of other client data. Sure, it will not stop last gen cheats from sending fake packet. But it's usually harder to do than hooking and spoofing Lua data on client side. And it is better than relying on client side function.

The truth is this. The current state of MTA AC leaves servers no choice but to fingerprint cheaters and try to make ban evasion as hard as possible. Accounts system with 2FA don't stop the abuse and don't even slow down cheaters anymore. Getting hardware information is also important to make better security for normal accounts (users) and prevent potential abuse.

> we could let the user set a _performance_ switch in the settings

So instead of good performance profiling you suggest to add new setting that could be abused by any user (in a lot of cases servers use more optimized but less accurate approach for different things on low end computers and it's ok as long as the number of low end computers isn't too high but if any user with good specs can fool the server by using this simple setting then it will potentially create more competitive advantage issues than any current solution that is based on client specs).

PlatinMTA:
>> we could let the user set a performance switch in the settings
> So instead of good performance profiling you suggest to add new setting that could be abused by any user

I think that adding a server graphics option in the configuration could be quite useful, but yeah the profiling would be good to catch cheaters and ban evasors (serials are simply too easy to spoof, when serial2)

botder:
> There are ways to make it harder for client to fake this information. Implement getSystemInfo on server side. Send packet with client specs to the server on connection as part of other client data. Sure, it will not stop last gen cheats from sending fake packet. But it's usually harder to do than hooking and spoofing Lua data on client side. And it is better than relying on client side function.

Since the code here is going to be public, it will only be a matter of minutes/days until cheaters will submit fake data.

> The truth is this. The current state of MTA AC leaves servers no choice but to fingerprint cheaters and try to make ban evasion as hard as possible. Accounts system with 2FA don't stop the abuse and don't even slow down cheaters anymore. Getting hardware information is also important to make better security for normal accounts (users) and prevent potential abuse.

Have you tried blocking IP ranges and access from non-domestic IP ranges?

> > we could let the user set a _performance_ switch in the settings
> 
> So instead of good performance profiling you suggest to add new setting that could be abused by any user (in a lot of cases servers use more optimized but less accurate approach for different things on low end computers and it's ok as long as the number of low end computers isn't too high but if any user with good specs can fool the server by using this simple setting then it will potentially create more competitive advantage issues than any current solution that is based on client specs).

What do you mean fool the server? If the user is dumb enough to request higher fidelity graphics and their game runs at 5 FPS, then it's a user problem. And if we're talking about actually good players, then they will want to play on the lowest setting anyway, to increase average FPS, smoothness and distractions by details. You're doing something wrong if players gain a clear advantage on lowest settings.

starfleet001:
> Since the code here is going to be public, it will only be a matter of minutes/days until cheaters will submit fake data.

You are in the position to make it harder to bypass. It would be more beneficial for everyone if things like this would be handled by netc (so creating interface with hardware information).

> Have you tried blocking IP ranges and access from non-domestic IP ranges?

Of course. It helps to keep amateur cheaters away. But we have to deal with targeted attacks from time to time. These developers usually use compromised domestic routers for proxy. And you'd be surprised but in a lot of cases they don't bother properly spoofing hardware data to evade restrictions. So better profiting in this case would definitely help.

Also there are legitimate use cases for VPN. Blocking it completely reduces the number of potential players. Every day more countries add more restrictions for its citizens (we are slowly moving into the dark ages of the internet worldwide). So the use of VPN is becoming new norm for good players.

> You're doing something wrong if players gain a clear advantage on lowest settings.

I was just talking about uselessness of any "preferred quality" setting. Because it wouldn't reflect what user hardware is actually capable of.

PlatinMTA:
> I was just talking about uselessness of any "preferred quality" setting. Because it wouldn't reflect what user hardware is actually capable of.

Really useful, I don't understand the argument being made here, they are just graphical settings dude I would be pissed off if the developer chose my graphical settings for me. I want to be able to set my own graphical settings, some players prefer more FPS, some players prefer more visual fidelity (not that much to get from renderwave in 2025 but you get the point).

In my case as a server developer, I have a big ass config panel in which you can change most things, from controls to graphical settings. It would be pretty nice to set the default values depending on their preferred style of gameplay rather than hardware information (this can be already be done, but a global solution would just up the QOL for every user... eventually server owners will take this into account, hopefully. still a really easy thing to integrate), then again with either approach the user should be able to choose what things should be ON or OFF. If something should be always ON, then dont make it togglable, simple as

starfleet001:
> Really useful, I don't understand the argument being made here

Then I'd suggest to create new issue specifically for this feature. It's not related to this PR anyway. I'm not arguing that it could be useful for someone. The thing is... This "graphics" setting already exists (fx quality, draw distance, etc.). I set my custom gamemode settings to low, medium or high by default based on user selected setting from the game.
This is what I tried to explain. It's redundant. But I have nothing against it and it doesn't make sense to argue about it.

"Preferred quality" setting doesn't solve the need for hardware profiling (to know the exact capabilities of user hardware). There are other use cases for this besides fingerprinting. It was already discussed earlier in this topic.

If someone here is worried so bad about function that would return not personally identifiable hardware information then just add new checkbox to disable it (enabled by default). MTA already uses this approach for "screen upload" functionality. This is the most optimal solution that would satisfy everyone in my opinion.

PlatinMTA:
> "Preferred quality" setting doesn't solve the need for hardware profiling

Totally agree, I also agree that this should be handled in a way that scriptkiddies have a hard time circumventing (in netc for example)

ghost:
Well as I see there are two types of opinion for this PR, one for the performance part and the other one for a new fingerprinting opportunity.
As I already wrote a little while ago, to make a good fingerprint based on this systeminfo func we would need way more details than this PR would give us.

--
We have tried to block IP ranges as well as blocking a whole country (because 90% of the players from that country were all cheaters)

They started to use vpn, we started to use a detector api for that but now they found vpn services which won't get detected by any of the best vpn detector apis. 

We also faced a problem which occured among good vpn users. They used vpn to connect to our server because the vpn service gave them better routing so they had -60ms. Now they can't use that. 

They can spoof their serial into anything, not only generating new ones but setting it to other player's serial or set it to an insult word (recently a PR changed this so they have to spoof their serial into a real looking one)

So from our viewpoint a new fingerprinting system would be highly appreciated. 
We already protected our gamemodes so they can't cause much harm but they can do pesky things. Currently the only solution is to manually ban them when they cheat.

botder:
@derxgbb The solution to this would be to update the serial system, from my point of view.

ghost:
> @derxgbb The solution to this would be to update the serial system, from my point of view.

If that's possible to update that system in a way where malicious users will not be able to spoof/fake their serials then its okay.

PR #2268, opened 2021-06-18
Add engineStreamingFreeUpMemory

Pirulax:
This function is a wrapper around [CStreaming::MakeSpaceFor](https://github.com/codenulls/gta-reversed/blob/02532949f2001c20e53037bd339d1ee22f86d427/source/game_sa/CStreaming.cpp#L2161-L2170)

The downside is that it isn't accurate, at all. The function works more like "try to free up at least this amount of bytes". It usually ends up freeing everything that isn't currently needed (which, depending on the scenery might be 1-200 megs).
**EDIT: Turned out I was calling the wrong function. Re-test before merge!**

This function will be useful for scripters when #1677 is introduced (along with a function to get the current streaming mem. usage)


Client/game_sa/CStreamingSA.cpp
```diff
@@ -152,3 +152,8 @@ void CStreamingSA::ReinitStreaming()
 
     reinitStreaming();
 }
+
+void CStreamingSA::MakeSpaceFor(std::uint32_t memoryToCleanInBytes)
+{
+    ((int(__cdecl*)(std::uint32_t))0x40E560)(memoryToCleanInBytes);
```
botder:
```suggestion
    using FunctionSignature = void(__cdecl*)(std::uint32_t);
    auto Function = reinterpret_cast<FunctionSignature>(0x40E120);
    Function(memoryToCleanInBytes);
```
`0x40E560` is the address of `static void __cdecl CStreaming::ReInit(void)`

Pirulax: Thank you very much, very good catch.
It seemed to work, although in a weird manner, now I know why
```suggestion
    ((int(__cdecl*)(std::uint32_t))0x40E120)(memoryToCleanInBytes);
```

Client/game_sa/CStreamingSA.cpp
```diff
@@ -152,3 +152,8 @@ void CStreamingSA::ReinitStreaming()
 
     reinitStreaming();
 }
+
+void CStreamingSA::MakeSpaceFor(std::uint32_t memoryToCleanInBytes)
+{
+    ((int(__cdecl*)(std::uint32_t))0x40E120)(memoryToCleanInBytes);
```
botder:
```suggestion
    (reinterpret_cast<void(__cdecl*)(std::uint32_t)>(0x40E120))(memoryToCleanInBytes);
```
That function doesn't return anything.

botder (approved):

PR #2282, opened 2021-08-04
Add extra checks in CXMLImpl::ParseString to avoid crash when invalid XML data supplied

Lpsd:
As above, doing something like `xmlLoadString("<>asd")` would crash the client or server.


patrikjuvonen (approved):
Looks good!

PR #2287, opened 2021-08-04
Correct flags name and order of CBaseModelInfoSAInterface

thisdp:
Corrent some flags order and name of CBaseModelInfoSAInterface.

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1292,6 +1292,7 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
         // If no collision model has been set before, store the original in case we want to restore it
         if (!m_pOriginalColModelInterface)
             m_pOriginalColModelInterface = m_pInterface->pColModel;
+            m_pInterface->bAlphaTransparency = false;
 
         // Apply some low-level hacks
         pColModelInterface->level = 0xA9;
@@ -1299,6 +1300,8 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
         // Call SetColModel
         DWORD dwFunc = FUNC_SetColModel;
         DWORD ModelID = m_dwModelID;
+        // Don't force the isLod argument to prevent collisionless problem
+        bool  isLOD = m_pInterface->bIsLod;
         _asm
         {
             mov     ecx, ModelID
@@ -1308,7 +1311,7 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
             mov     ecx, dword ptr[eax + ecx*4]
             pop     eax
 
-            push    1
+            push    isLOD
             push    pColModelInterface
             call    dwFunc
         }
@@ -1357,6 +1360,8 @@ void CModelInfoSA::RestoreColModel()
             DWORD dwFunc = FUNC_SetColModel;
             DWORD dwOriginalColModelInterface = (DWORD)m_pOriginalColModelInterface;
             DWORD ModelID = m_dwModelID;
+            // Don't force the isLod argument to prevent collisionless problem
+            bool  isLOD = m_pInterface->bIsLod;
             _asm
             {
                 mov     ecx, ModelID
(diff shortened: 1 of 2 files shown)


Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1292,13 +1292,16 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
         // If no collision model has been set before, store the original in case we want to restore it
         if (!m_pOriginalColModelInterface)
             m_pOriginalColModelInterface = m_pInterface->pColModel;
+            m_pInterface->bAlphaTransparency = false;
```
Lpsd: This statement is not inside the condition above. I assume you want it to be based on the indentation?

A bracketless condition will only execute the first statement directly after it. If you have multiple statements, you need brackets.

```suggestion
        if (!m_pOriginalColModelInterface)
        {
            m_pOriginalColModelInterface = m_pInterface->pColModel;
            m_pInterface->bAlphaTransparency = false;
        }
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1475,7 +1480,7 @@ void CModelInfoSA::MakeObjectModel(ushort usBaseID)
     MemCpyFast(m_pInterface, pBaseObjectInfo, sizeof(CBaseModelInfoSAInterface));
     m_pInterface->usNumberOfRefs = 0;
     m_pInterface->pRwObject = nullptr;
-    m_pInterface->usUnknown = 65535;
+    m_pInterface->uObjectInfoIndex = 65535;
```
Lpsd, line 1480: Is this necessary to fixing the collision bug?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1493,7 +1498,7 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
     m_pInterface->usNumberOfRefs = 0;
     m_pInterface->pRwObject = nullptr;
     m_pInterface->pVisualInfo = nullptr;
-    m_pInterface->usUnknown = 65535;
+    m_pInterface->uObjectInfoIndex = 65535;
```
Lpsd, line 1498: As above

Client/game_sa/CModelInfoSA.h
```diff
@@ -151,34 +151,53 @@ class CBaseModelInfoSAInterface
...
+            unsigned char bHasComplexHierarchy : 1;             // CClumpModelInfo::SetHasComplexHierarchy(unsigned int)
+            unsigned char bUnknown1 : 1;
+            unsigned char bSwaysInWind : 1;                     // Confirmed: Sways in wind
+            unsigned char bCollisionWasStreamedWithModel : 1;   // CClumpModelInfo::SetCollisionWasStreamedWithModel(unsigned int)
+            unsigned char bOwnsCollisionModel : 1;
+            unsigned char bUnknown2 : 1;
+            unsigned char bTagDisabled : 1;
+        };
+    };
+
```
Lpsd, line 196: Again, is any of this inside the scope of this PR? This should be submitted separately if it's not essential to fixing the collision bug.

Client/game_sa/CModelInfoSA.cpp
Lpsd: Also just noticed, is this relevant to fixing the bug either?

Client/game_sa/CModelInfoSA.cpp
thisdp: correcting flag name

thisdp: > As above

The whole flags order of CBaseModelInfoSAInterface is at a mess 😂

Lpsd: but is this necessary to fix the collision glitch? if not, it needs to be in a separate PR

Lpsd:
Submit a separate PR for the flags - just keep the code here necessary to fixing the collision glitch (or vice-versa)

thisdp:
> Submit a separate PR for the flags - just keep the code here necessary to fixing the collision glitch (or vice-versa)

we need merge this first, and then the collision glitch fixing

PR #2289, opened 2021-08-04
Fix all objects in a specific ide will lose their collisions after restoring collisions

### LINKED ISSUE #927: "Random collisionless objects" still exist
**Describe the bug**
Months ago MTA team fixed a bug that caused random objects to lose collisions.
Most users reported that this bug has been fixed once and for all.
Shortly after that fix i noticed some collisionless objects again but they were extremely rare and i thought i am crazy, but now i can confirm that this bug still exists, it is just much more rare.

A reconnect does not fix it.

crun for i, obj in pairs(getElementsByType("object")) do setElementCollisionsEnabled(obj, true) end
--> This does not fix it either.

**Only restarting MTA fixes this**

**To reproduce**
No idea. Sometimes a random object loses its collision. I have that issue in Map editor (with custom objects) and on my server.

**Screenshots**
image

**Version  1.5.6-release-16718**

thisdp:
The 2nd argument of setColModel, "isLod", is set wrongly. In our case, what we need is just the original value, but the code forces this argument to "true".

I can't reproduce the collisionless issue after with this.
Need more tests to find whether this will cause other issues.

Resolves #927

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1299,6 +1299,8 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
         // Call SetColModel
         DWORD dwFunc = FUNC_SetColModel;
         DWORD ModelID = m_dwModelID;
+        // Don't force the isLod argument to prevent collisionless problem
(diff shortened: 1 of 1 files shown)


botder:
Superseded by pull request #2296

PR #2292, opened 2021-08-04
Fix info command displaying two statuses of a running resource

theSarrum:
The `info` command displays 2 statuses (running and stopping) if the resource is running.
This PR fixes that.


patrikjuvonen (approved):
Nice find!

patrikjuvonen:
Thanks for the pull request!

PR #2296, opened 2021-08-06
Resolve remaining collisionless objects (Fixes #927)

### LINKED ISSUE #927: "Random collisionless objects" still exist

botder:
Supersedes pull request #2289 (**Fix all objects in a specific ide will lose their collisions after restoring collisions**).

- Added hooks for object creation and destruction to increase and decrease the collision slot of the original collision model, respectively.
- Improved the interface structures for GTA:SA and added compile-time `sizeof` verification.
- Added CColStore interface
- Simplified collision replacement code in CModelInfoSA
- Fixed memory leak in `CClientColModel::~CClientColModel()` by destroying the internal `CColModel` through the intended virtual method.
- Fixed a case where collision model data was unavailable (read: object was collisionless) until the camera/player moved far away and back.


Client/game_sa/CColModelSA.h
```diff
@@ -11,77 +11,137 @@
...
+    union
+    {
+        std::uint8_t flags;
 
-typedef struct
+        struct
+        {
+            std::uint8_t hasSpheresBoxesTriangles : 1;
+            std::uint8_t isSingleAllocationCollisionData : 1;
+            std::uint8_t flag0x04 : 1;
```
Pirulax:
```suggestion
            std::uint8_t isActive: 1;
```

Client/game_sa/CColModelSA.h
```diff
@@ -11,77 +11,137 @@
...
-    CColTriangleSA*      pColTriangles;
-    CColTrianglePlaneSA* pColTrianglePlanes;
-} CColDataSA;
-
-class CColModelSAInterface
+    std::uint16_t        numSpheres;
+    std::uint16_t        numBoxes;
+    std::uint16_t        numTriangles;
+    std::uint8_t         numSuspensionLines;
+    std::uint8_t         flags;
```
Pirulax:
```suggestion
    uint8_t   usesDisks : 1;
    uint8_t   notEmpty : 1;
    uint8_t   hasShadowInfo : 1;
    uint8_t   hasFaceGroups : 1;
    uint8_t   hasShadow : 1;
```

Client/game_sa/CColModelSA.h
```diff
@@ -11,77 +11,137 @@
...
+    CColSphereSA*        spheres;
+    CColBoxSA*           boxes;
+    CColLineSA*          suspensionLines;
```
Pirulax: In theory its like this:
```suggestion
    union {
        CColLine* lines;
        CColDisk* disks;
    };
```

Client/game_sa/CColStoreSA.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+void CColStoreSA::BoundingBoxesPostProcess()
+{
+    using Signature = void(__cdecl*)();
+    const auto function = reinterpret_cast<Signature>(0x410EC0);
+    function();
+}
+
+int CColStoreSA::AddColSlot(const char* name)
```
Pirulax, line 35: Original name is `allocateCollisionFile`.
Just making sure this is the address you want it to be.

Client/game_sa/CGameSA.cpp
```diff
@@ -265,6 +267,7 @@ CGameSA::~CGameSA()
     delete reinterpret_cast<CAEAudioHardwareSA*>(m_pAEAudioHardware);
     delete reinterpret_cast<CAudioContainerSA*>(m_pAudioContainer);
     delete reinterpret_cast<CPointLightsSA*>(m_pPointLights);
+    delete static_cast<CColStoreSA*>(m_collisionStore);
```
Pirulax, line 270: Why even is this cast needed?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -26,6 +26,28 @@ std::unordered_map<DWORD, unsigned short>                             CModelInfo
 std::unordered_map<DWORD, std::pair<float, float>>                    CModelInfoSA::ms_VehicleModelDefaultWheelSizes;
 
 static constexpr uintptr_t vftable_CVehicleModelInfo = 0x85C5C8u;
+static constexpr uintptr_t RESOURCE_ID_COL = 25000;
```
Pirulax:
```suggestion
static constexpr size_t RESOURCE_ID_COL = 25000;
```

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -26,6 +26,28 @@ std::unordered_map<DWORD, unsigned short>                             CModelInfo
...
+    auto function = reinterpret_cast<Signature>(0x5B2C20);
+    function(idx, colModel);
+}
+
+static void CStreaming_RemoveModel(int idx)
+{
+    using Signature = void(__cdecl*)(int);
+    auto function = reinterpret_cast<Signature>(0x4089A0);
+    function(idx);
+}
```
Pirulax: I'd suggest to move this to `CStreaming`

Pirulax (approved):
Although as per [discussion](https://github.com/multitheftauto/mtasa-blue/discussions/1816) we're leaving HN behind (although not officially), I'd still recommend it's usage in the GameSA project, as it already heavily uses it.
Also in this discussion it was agreed on (from what I see) to keep the `m_` prefix. So please do that.
Other than that I see no glaring issues with the code.
I haven't tested it myself, or looked to deeply into it. (But I hope you did)

I'll approve it for the sake of approving it (as a nightly will definitely show issues if theres any).

Thanks for your time 😁

Client/game_sa/CGameSA.cpp
botder: CColStore class is not using a virtual destructor currently and the member variable is stored as `CColStore*` to keep the code a little bit simpler. ~~But I am going to make the destructor virtual.~~ Not worth it.

Client/game_sa/CColStoreSA.cpp
botder: And where does it say that? Even the Android IDB says AddColSlot.

Pirulax: In my IDB.
But no issue, I just wanted to warn you, just in case.

PR #2305, opened 2021-08-13
Fix onClientElementInteriorChange is triggered even if the element already in the same interior

### LINKED ISSUE #2280: onClientElementInteriorChange is triggered even if the element has not changed its interior
**Describe the bug**
As the title says, the `onClientElementInteriorChange` event is triggered even if the element has not changed its interior.

**To reproduce**
Steps to reproduce:
1. Type `/debugscript 3`
2. Start runcode
3. `crun setElementPosition(me, 0, 0, 3)`
4. `crun addEventHandler("onClientElementInteriorChange", root, function(oldInterior, newInterior) iprint(source, oldInterior, newInterior) end)`
5. `crun vehicle = createVehicle(411, 1, 0, 3)`
5.1. Nothing in the debug console
5.2. `crun setElementInterior(vehicle, 0)`
5.3. Output: `elem:vehicle[Infernus]14ED1BC8    0    0` - the old and the new interior are the same
6. `crun createPed(22, 1, 0, 3)`
6.1. Output: `elem:ped[22]2DE24EB0    0    0` - the debug console shows that the element has changed its interior, although we just created it

**Expected behaviour**
The event should not be triggered.

**Version**
Client: Multi Theft Auto v1.5.8-release-20887

gptrk0:
Issue #2280

### DIFF
--- a/Client/mods/deathmatch/logic/CClientEntity.cpp
+++ b/Client/mods/deathmatch/logic/CClientEntity.cpp
@@ -1298,6 +1298,9 @@ unsigned char CClientEntity::GetInterior()
 
 void CClientEntity::SetInterior(unsigned char ucInterior)
 {
+    if (m_ucInterior == ucInterior)
+        return;
+
     CEntity* pEntity = GetGameEntity();
     if (pEntity)
     {


StrixG (approved):
:shipit:

Lpsd:
Thanks for the fix

Dutchman101:
Reverted in 959dde2 - issue #2309 contains details of a bug in this change, that needs to be fixed before eventually re-adding it.

cc @Patrick2562

PR #2315, opened 2021-08-21
Add getPlayerMapOpacity function

theSarrum:
This PR adds a client side function that allows to get the opacity of the player's map (F11), which the player can change in the settings.

**Syntax**
`int getPlayerMapOpacity()`
<ins>OOP:</ins>
`Player.getMapOpacity()`

Returns a value from 0 to 255, where 0 is fully transparent and 255 is fully opaque.


botder:
> Returns a value from 0 to 255, where 0 is fully transparent and 255 is fully opaque.

You are returning the value of a CVar, which could be any value on the signed 32-bit integer value range.

AlexTMjugador:
The value of this CVar is clamped to the 0-255 interval here at `CClientVariables::ValidateValues`, which is called once during initialization:

https://github.com/multitheftauto/mtasa-blue/blob/180942be846f080a6228f6efacc19e27ca819188/Client/core/CClientVariables.cpp#L258

It's also clamped to that range in the settings:

https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/CSettings.cpp#L1561-L1566

Finally, the value of the CVar is used directly in this code, which ends up casting it to an `unsigned char`:

https://github.com/multitheftauto/mtasa-blue/blob/8b3244c9f35ad82a7073d6e21918ca9ff67e5456/Client/mods/deathmatch/logic/CRadarMap.cpp#L236-L242

So I'd say that in practice the value of this CVar can't be outside of the intended 0-255 range, but I still think it would be nice to make this more explicit in this code.

botder:
If you type `mapalpha=1024` into the F8 console, then it won't get clamped.

image

theSarrum:
Thanks for the tip. I made changes and corrected the code.

Client/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -623,7 +623,8 @@ int CLuaPlayerDefs::GetPlayerMapBoundingBox(lua_State* luaVM)
     return 1;
 }
 
-int CLuaPlayerDefs::GetPlayerMapOpacity()
+unsigned char CLuaPlayerDefs::GetPlayerMapOpacity()
 {
-    return g_pCore->GetCVars()->GetValue<int>("mapalpha");
+    int iMapOpacity = g_pCore->GetCVars()->GetValue<int>("mapalpha");
+    return Clamp(0, iMapOpacity, 255);
```
botder:
```suggestion
    return static_cast<unsigned char>(Clamp(0, iMapOpacity, 255));
```

patrikjuvonen (approved):
Thanks, seems to work. Useful for keeping custom maps synced with the MTA setting with the lack of some getConfig function at this time.

Issue #2262, opened 2021-06-17
Hard-hitting crash (game_sa.dll @ 0x00011d7a)


Dutchman101:
**Describe the bug**
There is a Top 10 crash (game_sa.dll, offset 0x00011d7a) with a count of 11,000 unique victims.

Pre-note: Before describing the issue, below you will find regularly updated offsets. I hope that listed offsets will allow crash victim server owners to find this issue & discover what's going on, triggered by bad model(s) on their server.

**Update (18th Dec 2021)**: The crash offset is 0x000121ea with the new main release (r21048). I got reminded of this issue (as i commented again at the test PR #2319) as this crash is still the first non-memory abuse (server optimization issues) one from the top down. After a week on the current build, it's now at rank #11 as you can see at the [new top crash stats](https://i.imgur.com/gZH1ca0.png) and to give a better start to server owners, here's a list of top servers for this crash to occur: [server crash list](https://i.imgur.com/er76xvn.png)

The offset obviously shifts with new MTA revisions, but then generally starts with 00011d or 000121 (last 2 digits shifting), although as reported by a server owner with affected players (see clarified later in this issue) players probably won't see that anyways, as instead of a crash dialog they are said to have MTA suddenly close and face a window like "MTA was unable to replace a model [id x]" instead of crash dialog.

<details><summary>Dumptrace 1:</summary>

```
*******************************************************************************
*                                                                             *
*                        Exception Analysis                                   *
*                                                                             *
*******************************************************************************

CONTEXT:  (.ecxr)
eax=00000000 ebx=0028f1b8 ecx=11c0af28 edx=00000000 esi=0028f1b8 edi=1b96a2e8
eip=71e51d7a esp=0028f100 ebp=0028f128 iopl=0         nv up ei pl zr na pe nc
cs=0023  ss=002b  ds=002b  es=002b  fs=0053  gs=002b             efl=00000246
game_sa!CFileLoader_SetRelatedModelInfoCB+0xda:
71e51d7a 8be5            mov     esp,ebp
Resetting default scope

EXCEPTION_RECORD:  (.exr -1)
ExceptionAddress: 71e51d7a (game_sa!CFileLoader_SetRelatedModelInfoCB+0x000000da)
   ExceptionCode: c0000409 (Security check failure or stack buffer overrun)
  ExceptionFlags: 00000001
NumberParameters: 1
   Parameter[0]: 00000002
[...]

PR #2319, opened 2021-08-24
Add checks to find out reason of #2262

Pirulax:
Regarding the crash at #2262

I've looked at the models, but I didnt find any frame with name longer than 24 chars.
I'm adding this to see if this is the reason of the crashes.

Ah, also, function signature was wrong for the callback, so I fixed that.

### DIFF
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -43,12 +43,42 @@ public:
     }
 };
 
-static char* GetFrameNodeName(RwFrame* frame) {
-    return ((char* (__cdecl*)(RwFrame*))0x72FB30)(frame);
+static const char* GetFrameNodeName(RwFrame* frame) {
+    return ((const char* (__cdecl*)(RwFrame*))0x72FB30)(frame);
 }
 
-static void GetNameAndDamage(char const* nodeName, char* outName, bool& outDamage) {
-    return ((void(__cdecl*)(char const*, char*, bool&))0x5370A0)(nodeName, outName, outDamage);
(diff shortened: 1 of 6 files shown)


Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -43,12 +43,42 @@ class CDamagableModelInfo
...
+    {
+        outDamage = false;
+        if (EndsWith("_l0") || EndsWith("_L0"))
+            TerminatedCopy(sizeof("_l0") - 1);
+        else {
+            assert(nodeNameLen < BuffSize);
+            strcpy_s(outName, nodeName);
+        }
+    }
 }
```
Pirulax, line 82: This is the POI (point of interest)

Pirulax:
Actually, now that I looked at the code in the IDB, it seems like `bool` is the right type.

Dutchman101:
It seems this was forgotten about, PR's with debug code for temporary logging purposes don't need the same code review attention as features or actual changes would. I feel like we should merge this as soon @Pirulax comments to confirm he's available for the process (that includes intepreting the results of debug logging, and indicating how to fix it). Since i know he's on his regular development activity break..

Dutchman101:
Closing this in favour of [comment #4 on issue 2262](https://github.com/multitheftauto/mtasa-blue/issues/2262#issuecomment-997209318), as the crash cause was already discovered. In case you needed to collect even more info anyways, feel free to reopen

Pirulax:
This also was a fix for the issue though, not just logging. I'll modify the code a little, and reopen the PR.

PR #2323, opened 2021-08-28
Revert "Increase streaming memory setting limit (#2264)"

Pirulax:
As-per discussion in [#streaming-memory](https://discord.com/channels/801330706252038164/854824035329507359/881247673154433094)

It has to be revised, before its put back in.
It causes more issues than it solves, as of now.


patrikjuvonen:
Can we confirm what happens if we revert this? Do the existing values get clamped at the new maximum?

Pirulax:
https://github.com/multitheftauto/mtasa-blue/blob/5f320d7a3280a4340827bb2734d01a20cc57bfc9/Client/core/CSettings.cpp#L1555-L1559

patrikjuvonen:
> https://github.com/multitheftauto/mtasa-blue/blob/5f320d7a3280a4340827bb2734d01a20cc57bfc9/Client/core/CSettings.cpp#L1555-L1559

This function is run only for the settings screen but what about when the configuration is loaded from the disk?

Pirulax:
Just tested it.. It's only reset to 256 if you move the slider, and when you connect to a server.
Otherwise it stays 512 in the config.

patrikjuvonen:
~~Maybe we should add clamp on the setting on load just to make sure we revert user changes to the configs to avoid headache if this change does in fact cause issues.~~

Maybe the reset on connect is enough?

Pirulax:
Should be enough. I remember seeing some code that clamps the value in `DoPulse`

patrikjuvonen (approved):
Tested lightly.

Dutchman101:
> As-per discussion in #streaming-memory

For the record, here is the relevant discussion in its entirety: [https://i.imgur.com/7AKE1S3.png](https://i.imgur.com/7AKE1S3.png) (click to zoom)

So the concern is additional crashes due to base RAM usage increase (streaming memory allocates from RAM, not VRAM) which doesn't float well with the large number of highly unoptimized servers in MTA - we know that the game, due to being based on an 32-bit game from 2005, becomes unstable when SA process usage goes between 2.6GB and 3.4GB, with memory crash risk. No matter how poweful of a PC someone has, the additional 256MB raise can be just the little push needed to bring players that aren't currently crashing on some unoptimized servers, or not so often yet, over the brink towards experiencing memory crashes often. This is why we saw an increase in memory crashes across the board back when it was merged, prompting the revert.

I don't see a way around this side effect given that we're dealing with the reality that many server owners and scripters don't know what they are doing (the 'many unoptimized servers' point). Maybe at most we can turn it into a server opt-in with a function to set a client's streaming memory up to 512 or 768MB, because then the server owner can judge based on their individual case as they may investigate how optimized their server (avg. RAM usage and outcomes on clients) is beforehand, and also notice any adverse effects like a raise in crash complaints, blame it on themselves and undo changing the limit. But at least we'll hand over the power to server owners for unlocking potential, of course there are some projects that aren't too unoptimized for a 256MB raise in RAM usage to mess everything up, we need to provide them the liberty and not block creativity where some mods may need more streaming mem.

^ I see that as the best approach if we ever re-introduce this feature.

PR #2328, opened 2021-08-30
Add LuaYEET 2.0 alongside Lua 5.1

Pirulax:
**I'm just joking. FYI: LuaJIT == LuaYEET**

Add LuaYEET

- [x] Add premake, and yeet src
- [x] Make client build
- [x] Make client run 
- [ ] Make server compile (have to add separate Lua project)
- [ ] Add back / forward port MTA crap (like `nexpectedresults`)
- [ ] Fix issues that arise
- [ ] Test
- [ ] Disable FFI (simple, added here so I wont forget)
- [ ] Disable `require`

**Huge issue: Memory limitation of 2 GB for x86**
Which sadly means, that this has to be an opt-in, as we already have issues with 4gb limitation..
Keep in mind that YEET is API compatible with 5.1, so fortunately we can just switch DLLs run-time (when connecting to the server)
- [ ] Make it able to switch between Lua 5.1 and YEET (Client only)
 
Please discuss anything related over at [the discussion for it](https://github.com/multitheftauto/mtasa-blue/discussions/903).


Disinterpreter:
> Disable FFI (simple, added here so I wont forget)

Eh... It would be great if we solve all our conflicts with potential server "security" and turn it on at the server.

Pirulax:
Yeah I think allowing it on the server is totally possible.

Einheit-101:
Is there anywhere info where i can read something about benefits and problems that come with LuaJit?

Pirulax:
Yes, I think in your case it will be very beneficial. 
Please see the [discussion](https://github.com/multitheftauto/mtasa-blue/discussions/903).
If you would like I can provide you a test client with JIT. (Actually, the artifacts from here should work fine. Server doesnt compile yet, thats why it fails, but client should work.)

Einheit-101:
I could test it with my ww2 Server, it is not as efficient as my new server and also a lot bigger code-wise

ERAGON007:
ah shit, here we go again :(

patrikjuvonen:
This draft pull request was closed because it has been marked stale for 30 days with no activity.

Pirulax:
What I'll do most likely is add LuaJIT for the server.
We can't add it to the client, because of *reasons* (Mostly the memory limit, that can be solved on server by running x64).

ERAGON007:
> What I'll do most likely is add LuaJIT for the server. We can't add it to the client, because of _reasons_ (Mostly the memory limit, that can be solved on server by running x64).

Are there any progress on this matter? what problems did you face while implementing the server side

PR #2333, opened 2021-08-31
Improve addEvent when sharing events over multiple resources

### LINKED ISSUE #2448: triggerEvent/addEvent issue
**Describe the bug**

triggerEvent is not working when a given resource start/stop order is used.

**To reproduce**
1. start test3
2. start test2
3. stop test3
4. start test1

[test.zip](https://github.com/multitheftauto/mtasa-blue/files/7643901/test.zip)

**Expected behaviour**

The event should be triggered (message should appear in debug console)

**Version**

Multi Theft Auto v1.5.9-release-21024.1

Pieter-Dewachter:
Fixes #2448

When adding event handlers for a custom event in multiple resources, you have to use addEvent to make sure the event is already registered when adding the handler. The problem with the existing system happens in a situation where the first resource that used addEvent, is stopped. It is still possible that other resources had handlers for that specific event (and also used addEvent, for that matter), but still the event will never get triggered again. The reason is that the addEvent calls that came after the initial one, immediately returned false and didn't actually do anything.

This PR changes that behaviour, keeping a list of all resources that added a certain event and only removing an event when all resources that used addEvent are stopped. For security reasons, addEvent will return false and not do anything in case the allowRemoteTrigger argument would have been altered between different calls to the function.

Small note: I noticed that the strArguments member of SEvent is never used for anything, perhaps it could be an idea to clean this and remove the unused member?

### DIFF
--- a/Client/mods/deathmatch/logic/CEvents.cpp
+++ b/Client/mods/deathmatch/logic/CEvents.cpp
@@ -28,16 +28,27 @@ bool CEvents::AddEvent(const char* szName, const char* szArguments, CLuaMain* pL
     assert(szName);
     assert(szArguments);
 
-    // If it already exists, return
-    if (Get(szName))
-        return false;
+    // Get the event if it already exists
+    SEvent* pEvent = Get(szName);
 
-    // Create and add the event
-    SEvent* pEvent = new SEvent;
-    pEvent->strName = szName;
-    pEvent->strArguments = szArguments;
-    pEvent->pLuaMain = pLuaMain;
-    pEvent->bAllowRemoteTrigger = bAllowRemoteTrigger;
+    if (pEvent)
+    {
+        // If bAllowRemoteTrigger has been altered, return
+        if (pEvent->bAllowRemoteTrigger != bAllowRemoteTrigger)
+            return false;
+        
+        // Add pLuaMain to the set, std::set guarantees unique elements
+        pEvent->pLuaMainSet.insert(pLuaMain);
+    }
+    else
+    {
+        // Create and add the event
+        pEvent = new SEvent;
+        pEvent->strName = szName;
+        pEvent->strArguments = szArguments;
+        pEvent->pLuaMainSet.insert(pLuaMain);
+        pEvent->bAllowRemoteTrigger = bAllowRemoteTrigger;
+    }
 
     m_EventHashMap[szName] = pEvent;
 
@@ -80,14 +91,21 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
     while (iter != m_EventHashMap.end())
     {
         SEvent* pEvent = (*iter).second;
-        // If they match, delete it null it and set the bool
-        if (pEvent != NULL && pEvent->pLuaMain == pMain)
+        
+        // If they match, remove pMain from the set and check for deletion
+        if (pEvent->pLuaMainSet.find(pMain) != pEvent->pLuaMainSet.end())
         {
-            // Delete the object
-            delete pEvent;
+            pEvent->pLuaMainSet.erase(pMain);
 
-            // Remove from list
-            m_EventHashMap.erase(iter++);
+            // If no pMain is left, delete it null it and set the bool
+            if (pEvent->pLuaMainSet.size() == 0)
+            {
+                // Delete the object
+                delete pEvent;
+
+                // Remove from list
+                m_EventHashMap.erase(iter++);
+            }
         }
         else
             ++iter;
--- a/Client/mods/deathmatch/logic/CEvents.cpp
+++ b/Client/mods/deathmatch/logic/CEvents.cpp
@@ -91,7 +91,7 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
     while (iter != m_EventHashMap.end())
     {
         SEvent* pEvent = (*iter).second;
-        
+
         // If they match, remove pMain from the set and check for deletion
         if (pEvent->pLuaMainSet.find(pMain) != pEvent->pLuaMainSet.end())
         {
@@ -104,11 +104,11 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
                 delete pEvent;
 
                 // Remove from list
-                m_EventHashMap.erase(iter++);
+                m_EventHashMap.erase(iter);
             }
         }
-        else
-            ++iter;
+
+        ++iter;
     }
 }
 
--- a/Client/mods/deathmatch/logic/CEvents.cpp
+++ b/Client/mods/deathmatch/logic/CEvents.cpp
@@ -38,7 +38,7 @@ bool CEvents::AddEvent(const char* szName, const char* szArguments, CLuaMain* pL
             return false;
         
         // Add pLuaMain to the vector, in case it's not already there
-        if (std::find(pEvent->pLuaMainVector.begin(), pEvent->pLuaMainVector.end(), pLuaMain) == pEvent->pLuaMainVector.end())
+        if (!ListContains(pEvent->pLuaMainVector, pLuaMain))
             pEvent->pLuaMainVector.push_back(pLuaMain);
     }
     else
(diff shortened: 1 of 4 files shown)


Client/mods/deathmatch/logic/CEvents.cpp
```diff
@@ -80,17 +91,24 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
...
+        if (pEvent->pLuaMainSet.find(pMain) != pEvent->pLuaMainSet.end())
         {
-            // Delete the object
-            delete pEvent;
+            pEvent->pLuaMainSet.erase(pMain);
 
-            // Remove from list
-            m_EventHashMap.erase(iter++);
+            // If no pMain is left, delete it null it and set the bool
+            if (pEvent->pLuaMainSet.size() == 0)
```
Pirulax:
```suggestion
            if (pEvent->pLuaMainSet.empty())
```

Client/mods/deathmatch/logic/CEvents.h
```diff
@@ -16,10 +16,10 @@
 
 struct SEvent
 {
-    class CLuaMain* pLuaMain;
-    std::string     strName;
-    std::string     strArguments;
-    bool            bAllowRemoteTrigger;
+    std::set<class CLuaMain*> pLuaMainSet;
```
Pirulax: I'd argue that using `unordered_set` would be better in terms of memory fragmentation.
Or even better.. `std vector`. Yeah, might sound stupid, but you'd be surprised what the L1 cache can do.

If you go for the `vector` option, you can use `find_if` to check if the event exists, and then insert.
As for the remove `vec.erase(std::remove(vec.begin(), vec.end(), luaMain), vec.end());` will do the trick.

Server/mods/deathmatch/logic/CEvents.cpp
```diff
@@ -68,17 +79,24 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
...
+        if (pEvent->pLuaMainSet.find(pMain) != pEvent->pLuaMainSet.end())
         {
-            // Delete the object
-            delete pEvent;
+            pEvent->pLuaMainSet.erase(pMain);
 
-            // Remove from list
-            m_EventHashMap.erase(iter++);
+            // If no pMain is left, delete it null it and set the bool
+            if (pEvent->pLuaMainSet.size() == 0)
```
Pirulax:
```suggestion
            if (pEvent->pLuaMainSet.empty())
```

Client/mods/deathmatch/logic/CEvents.cpp
```diff
@@ -28,16 +28,27 @@ bool CEvents::AddEvent(const char* szName, const char* szArguments, CLuaMain* pL
...
-    SEvent* pEvent = new SEvent;
-    pEvent->strName = szName;
-    pEvent->strArguments = szArguments;
-    pEvent->pLuaMain = pLuaMain;
-    pEvent->bAllowRemoteTrigger = bAllowRemoteTrigger;
+    if (pEvent)
+    {
+        // If bAllowRemoteTrigger has been altered, return
+        if (pEvent->bAllowRemoteTrigger != bAllowRemoteTrigger)
+            return false;
```
Pirulax, line 38: This is going to cause backwards compatibility issues with some scripts.

Pieter-Dewachter: Could you elaborate as to why? Current functionality will also return false when calling addEvent() multiple times for the same event. I could be missing something here of course, just curious. ^^

Client/mods/deathmatch/logic/CEvents.h
Pieter-Dewachter: I do agree that std::unordered_set makes more sense, I did not know that the default std::set was ordered (I use Qt containers at my job, QSet is already unordered). Just to verify and make sure we are using the most optimal container, I did some performance tests. See the attached image for some of my results, seems like std::unordered_set is quite a lot faster than std::vector (which is also slower than std::set for that matter). Am I missing something here or is the L1 cache thing harder to measure? See also the attached code for my tests, I replicated as close as possible the use-case relevant to this PR.
unknown
Source code for testing: [https://pastebin.com/j3gYgGYY](https://pastebin.com/j3gYgGYY)

Pirulax: Woah, thats some extensive benchmarking!

Your test set is just way too big. Realistically, at most there will be at most 200 - 300 resources running.
And, realistically, not more than 10 resources will try to register the same event.
With this in mind, I think that `std::vector` would be a better choice, for a couple of reasons:
- We dont care about performance here, this function is very rarely used. Unless it takes seconds to execute, its fine.
- We care about memory fingerprint a lot. `unordered_set` has its quirks, and overhead. [MSVC useses double linked lists with a vector](https://stackoverflow.com/a/37177838/15363969), which is just a hell in terms of memory fingerprint. 

In this case, with such a small set of values, a vector will be more efficient in terms of memory, and possibly even performance (because of prefetching).
Obviously, when the number of items is large, a `vector` won't beat a `set`, let alone `unodered_set` in terms of performance, but will always beat them in terms of memory usage.

Also, just as a side-note: Your benchmarks aren't accurate to a 100% percent. They do reflect some overall picture, but I'd suggest you to use websites like [QuickBench](https://quick-bench.com). Theres a lot of quirks in benchmarking (CPU caching being the most noteable).

Pieter-Dewachter: Thank you for the explanation and the link, always nice to learn something! I have pushed the modification to use std::vector instead.

Pirulax: Thank you very much. Great!

Client/mods/deathmatch/logic/CEvents.cpp
Pirulax: Hm.. Can't remember actually. Probably some brainfart.

Client/mods/deathmatch/logic/CEvents.cpp
```diff
@@ -37,16 +37,17 @@ bool CEvents::AddEvent(const char* szName, const char* szArguments, CLuaMain* pL
         if (pEvent->bAllowRemoteTrigger != bAllowRemoteTrigger)
             return false;
         
-        // Add pLuaMain to the set, std::unordered_set guarantees unique elements
-        pEvent->pLuaMainSet.insert(pLuaMain);
+        // Add pLuaMain to the vector, in case it's not already there
+        if (std::find(pEvent->pLuaMainVector.begin(), pEvent->pLuaMainVector.end(), pLuaMain) == pEvent->pLuaMainVector.end())
+            pEvent->pLuaMainVector.push_back(pLuaMain);
```
Pirulax: Actually, sorry.. Now that I think about it, I think we have this `ListContains` method. Although stupidly named, I think it should work, please take a look at it.

Client/mods/deathmatch/logic/CEvents.cpp
```diff
@@ -92,13 +93,14 @@ void CEvents::RemoveAllEvents(class CLuaMain* pMain)
     {
         SEvent* pEvent = (*iter).second;
 
-        // If they match, remove pMain from the set and check for deletion
-        if (pEvent->pLuaMainSet.find(pMain) != pEvent->pLuaMainSet.end())
+        // If they match, remove pMain from the vector and check for deletion
+        std::vector<class CLuaMain*>::const_iterator pLuaMainIter = std::find(pEvent->pLuaMainVector.begin(), pEvent->pLuaMainVector.end(), pMain);
+        if (pLuaMainIter != pEvent->pLuaMainVector.end())
```
Pirulax: I think this one can be replaced by `ListRemove`

Pieter-Dewachter:
Totally forgot to respond here, I have changed my implementation according to your suggestions.

sbx320 (approved):

Lpsd (approved):
Looks good, tested and working as expected :shipit:

PR #2344, opened 2021-09-05
Fix #2335: Don't allow dead peds to enter vehicles

### LINKED ISSUE #908: Player / vehicle desync after car jacking, if jacker die

### LINKED ISSUE #2335: onVehicleStartExit incorrectly triggers after #1907
I have a simple script which creates a ped, and puts you both in a separate vehicle. If you attempt to leave the vehicle (detected via `onVehicleStartExit`), you'll be killed and your vehicle is destroyed. 

Sometimes, the `onVehicleStartExit` event gets triggered for the ped too, so subsequently they're also killed and vehicle is destroyed. It seems to only occur if you're quite close to the ped, and/or have dealt enough damage to the ped's vehicle recently (unsure on these being the actual criteria for triggering it).

I've uploaded a video below, you can see it's a little inconsistent to trigger. Watch the debug output at the bottom for the event being fired on both the player and ped.

https://streamable.com/5jutas

I've also uploaded a test resource, it's fairly simple. Just (re)start the resource - yourself and the ped will be spawned and placed in vehicles, then go close to the ped / hit the ped's vehicle and attempt to exit your own vehicle, which will kill you (and probably the ped too).

Can reproduce this all the way back to [r20751](https://nightly.mtasa.com/mtasa-1.5.8-rc-20751-20201207.exe), but not [r20749](https://nightly.mtasa.com/mtasa-1.5.8-rc-20749-20201206.exe) or earlier - which is why I believe it's related to #1907 (cc @Zangomangu)

[vehicle.zip](https://github.com/multitheftauto/mtasa-blue/files/7102271/vehicle.zip) <--- test resource

Zangomangu:
#1907 allowed dead peds to send entry/exit packets, because this is necessary to fix #908 
This PR fixes #2335 by not allowing a dead ped to enter a vehicle (serverside).

But why is this clientside check in `CClientPed::EnterVehicle` not preventing a dead ped from sending a packet?

https://github.com/multitheftauto/mtasa-blue/blob/2aa3c38ea24c373e6a7d0050613ea41b7dc14510/Client/mods/deathmatch/logic/CClientPed.cpp#L6456-L6459

Because in `CClientPed::IsDead` when the ped is streamed it queries gta for TASK_SIMPLE_DEAD instead of returning m_bDead, and the task probably isn't initiated yet before the next frame (and it processed the player wasted packet from `killPed` and entity destroy packet from `destroyElement` in the same frame, so when `CClientGame::DoVehicleInKeyCheck()` is pulsed afterwards the player appears to be alive and outside a vehicle for that 1 frame).

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -2834,7 +2834,7 @@ void CGame::Packet_Vehicle_InOut(CVehicleInOutPacket& Packet)
                     }
                 }
 
-                // Check we have a valid ped & he is spawned
+                // Check we have a valid ped
                 if (bValidPed)
                 {
                     // Handle it depending on the action
@@ -2856,6 +2856,13 @@ void CGame::Packet_Vehicle_InOut(CVehicleInOutPacket& Packet)
                                 FAIL_ACTION,
                                 FAIL_TRAILER,
                             } failReason = FAIL_INVALID;
+
+                            // Is he spawned? (Fix for #2335)
+                            if (!pPed->IsSpawned()) {
+                                CVehicleInOutPacket Reply(PedID, VehicleID, 0, VEHICLE_ATTEMPT_FAILED);
+                                pPlayer->Send(Reply);
+                                break;
+                            }
 
                             // Is this vehicle enterable? (not a trailer)
                             unsigned short usVehicleModel = pVehicle->GetModel();


Lpsd:
PR works great, fixes #2335 - although we did stumble upon an issue that has probably been present in MTA since forever, but up until now has only been applicable to players - since peds can now be jacked it also applies to them. 

I've linked the Discord discussion below - but essentially there should be a check on how often a player can attempt to get in/out of a vehicle. If you were to spawn the player directly after destroying the vehicle (same circumstances as in #2335), the "timers" for that check will be reset - meaning if you happen to spawn next to a player/ped who is in a vehicle, you _might_ end up jacking them.

There's no obvious fix to that yet without potentially breaking other stuff. It's a bug that only applies under very specific circumstances; which as I mentioned has been present with players since most likely forever, and which can also partly be blamed on bad logic from the server/scripter - so I don't think we need to worry about it applying to peds now too.  #INABIAF

https://discord.com/channels/801330706252038164/801411628600000522/884517276936712193

anyway - this PR fixes the case described in #2335 and _shouldn't_ affect anything else, so we're good here... :shipit:

Lpsd (approved):

patrikjuvonen:
Thanks!

Issue #450, opened 2018-09-11
Created objects don't reflect lights / shadows


ArranTuna:
**Describe the bug**
Created objects (as maps) don't reflect objects light as world objects do, here go some examples:

Vehicle light in a world object:
![image](https://user-images.githubusercontent.com/923242/45847160-0d088500-bd22-11e8-81b3-8843fd0e396e.png)

Vehicle light in a created object:
![image](https://user-images.githubusercontent.com/923242/45847162-10037580-bd22-11e8-9337-e7c3e35dcbb8.png)

Object (that's also created; it proves that the problem is with the object which should reflect the light, not the vehicle/object that produce the light) light in a world object:

![image](https://user-images.githubusercontent.com/923242/45847169-1560c000-bd22-11e8-989c-2b2fe4aaee5e.png)

Object light in a created object:

![image](https://user-images.githubusercontent.com/923242/45847173-185bb080-bd22-11e8-94f6-fe5f09cd954c.png)

**Additional context**
From https://bugs.mtasa.com/view.php?id=6576

qaisjp:
_@ccw808 made the following comment:_

>  As custom objects are movable, they seem to be treated like peds and vehicles. i.e. No projected textures such as headlight beams, blob and square shadows, blood etc.

_Followed by a later comment:_

> 
> Should be possible, I'll look into it soon.
> 
> Problems might be:
> 1) GTA buildings only have z rotation (no x/y axis tilting), so the shadow/light effect may not work correctly for MTA objects which have x/y rotations.
> 2) Performance - Maybe only do effect if fx quality is not low, or add an option in the video tab.
>

CrosRoad95:
i made an investigation and i found that function:
`CShadows::StoreShadowToBeRendered(uchar,RwTexture *,CVector *,float,float,float,float,short,uchar,uchar,uchar,float,bool,float,CRealTimeShadow *,bool)`
offset 0x707390

Last argument is `bDrawOnBuildings` and almost all time is `0` except for vehicles, where is set 1 is you're drive ( but this also not casting shadows ).
i tried change this byte to 1 while calling function
```cpp
MemSet((void*)0x7081A4, 0x1, 1);
MemSet((void*)0x536241, 0x1, 1);
MemSet((void*)0x707EBD, 0x1, 1);
```
but still ped/vehicle shadows not working on object created by createObject ( i tested with rotation x,y,z = 0 )

[...]

PR #2347, opened 2021-09-07
Fix corona rain reflections do not render (#2345)

### LINKED ISSUE #2345: Missing native rain road reflections detail
**Describe the bug**

We all know that Rockstar North does not future-proof their games. GTA is known to have some graphical issues when played on newer PCs - one of these is bugged road reflections. For some reason, they don't show while it's raining anymore unless you use an ancient/AGP video card.

There are several patches managed to fix this issue in single-player being for a long while, one of the most famous patches called **slient patch** done this job pretty well. however, due to the nature of MTA security measurement, the player is not being able to install any custom patch for the game, thus I'd suggest we can include this fix in the native MTA client, enjoy the authentic graphic that the original game offer.

**Screenshots**

1427018381_1426948187_gallery10
05e416d42a5eee5bd41a0285ae33c32ff6dbefcb33f2359bb073b4f29e6a5447

**Version**
MTASA 1.5.8, but I believe this fix was never added to the MTA yet.

lopezloo:
This PR fixes that corona rain reflections doesn't render (#2345).

I don't know how exactly it works but it seems that `zBufferNear` is invalid during corona render process for some reason so I changed offset to use `zBufferFar` instead and it works properly.

Summary:
- fixes corona rain reflections doesn't render (#2345)
- adds "Corona rain reflections" checkbox in video settings (disabled by default)
- custom created coronas have reflections disabled by default
- it doesn't work on custom placed objects (see issue #450)
- it doesn't work on every road (some models aren't flagged properly and lacks `is_road` flag)

New clientside functions:
Global:
`int getCoronaReflectionsEnabled()`
`bool resetCoronaReflectionsEnabled()` - resets to player desired value based on "Corona rain reflections" video option
`bool setCoronaReflectionsEnabled(int enabled)`
enabled:
0 - disabled
1 - enabled
2 - force enabled, even if there is no rain

Per marker element:
`bool isCoronaReflectionEnabled(marker marker)`
`bool setCoronaReflectionEnabled(marker marker, bool enabled)`
(both return false if marker type is not corona)

Also adds `is/setCoronaReflectionEnabled()` OOP Marker methods.

mta-screen_2021-09-16_01-32-32


patrikjuvonen:
Do we have photos of how it looks in vanilla GTASA?

AlexTMjugador:
I can't find a screenshot or video of GTA: SA PC rain corona effects as rendered by an AGP video card that is claimed to not be affected by this game bug, but in PS2 we can definitely see this effect, and it looks identical to what @lopezloo achieves with this patch:

PS2 vs PC graphics effects comparisons

patrikjuvonen:
I think this could be a toggleable graphic option then, as PC version does not ship with the feature enabled (or many other PS2 version graphics).

AlexTMjugador:
I recall hearing from a trustful source (more precisely, [this Vadim M video](https://youtu.be/jY3LcF-EwPo?t=145)) that the PC version does indeed try to render these effects, but they only look right on PCs with "an old AGP video card". I think it would be interesting for someone with "an old AGP video card" to test this and confirm that it does indeed happen, as maybe it can provide some insight about why `zBufferNear` becomes invalid.

PlatinMTA:
If this is merged, it would be cool to let custom coronas have this effect aswell.

lopezloo:
> I think this could be a toggleable graphic option then, as PC version does not ship with the feature enabled (or many other PS2 version graphics).

This feature exists in code but it's bugged. We could add new setting into ~~`setGlitchEnabled`~~ `setWorldSpecialPropertyEnabled` (I need feedback for that) for scripters but I believe this should be fixed by default. Less minor graphic fixes like sun glare or color temperature could actually go into video settings for users IMHO.

By the way, it doesn't work on custom created objects (same issue as with shadows, see #450).

Also, I have disabled these reflections for custom created coronas now since they are often used as pickups etc., not for actual lighting purposes.

In future we could add new function like `setCoronaReflectionEnabled` (per element) (~~but it is out of scope of this PR~~).

I have updated screenshot in first post.

patrikjuvonen:
> This feature exists in code but it's bugged. We could add new setting into `setGlitchEnabled` (I need feedback for that) for scripters but I believe this should be fixed by default. Less minor graphic fixes like sun glare or color temperature could actually go into video settings for users IMHO.

@AlexTMjugador explained his opinion on force enabling this: https://github.com/multitheftauto/mtasa-blue/issues/2345#issuecomment-914182609

Let's add a toggle for this in video settings, and maybe enable the setting by default?

AlexTMjugador:
I think that enabling this fix by default would be nice, as it is a desirable thing in most cases. I didn't recall the existence of the `setWorldSpecialPropertyEnabled` function while writing the comment that was referenced above, but I find it a good place to put this toggle into, as it already controls graphical details like changing the moon size with a sniper rifle, and it feels "natural" to just add another parameter to that function instead.

Allerek:
`setGamePatchEnabled`? Most accurate, better than 'fix'

patrikjuvonen:
Without having tested it yet, I just think force-enabling this can confuse some players who have never seen this effect and don't expect it to be there. So that's why I'd like the video setting, and yeah, a client-side function or addition to existing world function could be nice too.

gta191977649:
> Without having tested it yet, I just think force-enabling this can confuse some players who have never seen this effect and don't expect it to be there. So that's why I'd like the video setting, and yeah, a client-side function or addition to existing world function could be nice too.

I think add to the existing world function would be nice, in this case the server owner would decide whatever if they want this effect in their server, like set heat haze.
Also if some server has the shader that will conflict with this effect, they would had the ability to decide whatever they want it on their server.

lopezloo:
I have added `coronareflections` property to `setWorldSpecialPropertyEnabled`. It is true (which mean fixed) by default. If it is set to false, then calls to `CCorona::RenderReflections` are blocked so even if someone with so claimed AGP card or something would play, s/he still wouldn't be able to see it.

Personally I'm against adding new option into video settings because:
- it's a bugfix (single player one, but still), not new feature
- this effect is very subtle, it is showing only during the rain and is even barely visible at day time
- custom created coronas have this disabled by default
- there is no performance difference

lopezloo:
Thinking about it now, we actually could move it into seperate function like `setCoronaReflectionsEnabled` and instead of doing it enabled/not enabled we could introduce 3 states:
0 - disabled
1 - enabled (default)
2 - force enabled, even if there is no rain

lopezloo:
Reverted adding `coronareflections` property to `setWorldSpecialPropertyEnabled`.
Instead added new clientside function `setCoronaReflectionsEnabled(int iEnabled)`:
0 - disabled
1 - enabled (default)
2 - force enabled, even if there is no rain

How to test:
```
run me.position = Vector3(643.996, -1390.343, 13.51)
run setTime(0, 0) -- set night
run setWeather(8) -- set rain weather
crun setCoronaReflectionsEnabled(0) -- disable reflections
crun setCoronaReflectionsEnabled(1) -- enable reflections
run setWeather(0) -- set clear sky weather
crun setCoronaReflectionsEnabled(2) -- force enable reflections
crun setCoronaReflectionsEnabled(1) -- enable rain reflections
```

LVRENT:
What's the status on this? I hope it's going to be merged in the 1.5.9 as it is a "original" feature and I would love to see it in the next release.

gta191977649:
> What's the status on this? I hope it's going to be merged in the 1.5.9 as it is a "original" feature and I would love to see it in the next release.

It's ready, just need mta official to review.

I'm not sure whatever they would plan merge it into 1.5.9, it's been put into the backlog though.

lopezloo:
Needs to be tested & reviewed.

AlexTMjugador:
I think the Lua API would be more readable if the enable parameter was a string instead of a magic number, like in `setVehicleModelWheelSize`. I'd not expect any detrimental performance difference to be a concern, because this is the kind of function you'd call infrequently.

And, speaking of making the Lua API more elegant, a straightforward expansion would be to add corona functions that allow setting the reflection property per element, respecting the element tree hierarchy (so if `resourceRoot` is passed to such a function, all coronas created by a resource will be affected, for example). I think it'd be nice allow such a function to be executed with a `nil` element, in which case only world coronas will be affected. Would this be possible to do without complicating the PR too much?

lopezloo:
> And, speaking of making the Lua API more elegant, a straightforward expansion would be to add corona functions that allow setting the reflection property per element, respecting the element tree hierarchy (so if `resourceRoot` is passed to such a function, all coronas created by a resource will be affected, for example). I think it'd be nice allow such a function to be executed with a `nil` element, in which case only world coronas will be affected. Would this be possible to do without complicating the PR too much?

Implementing "forced" state per element is not possible without heavy modifications (hooking into render function). Implementing enabled/not enabled should be simple because it's just one flag.

Unsure about string as parameter, for me current approach is just like boolean - 0, 1 and extra 2 for "forced" state. Could be implemented as `off`, `rain`, `always` aswell though.

lopezloo:
> * it doesn't work on every road (seems like some objects aren't flagged properly)

It will be possible to fix that if we merge #1714 (`is_road` flag).

lopezloo:
Added:
`bool isCoronaReflectionEnabled(marker marker)`
`bool setCoronaReflectionEnabled(marker marker, bool enabled)`
(both return false if marker type is not corona)

And new property `reflectionEnabled` to Marker element.

TheNormalnij:
Thanks you for the PR.
I see this feature has some weird cases, when reflections render after barriers or disappear in some point of view. But this is unnecessary for this PR.
I found this effect a little bit poor to be enabled by default. @patrikjuvonen suggestion is good compromise here.

Dutchman101:
> I found this effect a little bit poor to be enabled by default. @patrikjuvonen suggestion is good compromise here.

MTA strives for feature parity with GTA SA as much as possible, and this is pretty much a fix so a GTA feature works in MTA as well. So in my eyes it should be enabled by default, as long there are no indications that it will interfere with custom shaders on servers (maybe we should go with it in nightlies, to collect results for that)

@lopezloo:
> Personally I'm against adding new option into video settings

But i see that most people are in favour of adding the option (so it's probably a PR requirement by now).. i can see uses for it, after so many years without corona rain reflections in MTA, people may not like it due to not being used to it. The end user should have the final say whether or not to use it

AlexTMjugador:
@TheNormalnij
> I see this feature has some weird cases, when reflections render after barriers or disappear in some point of view. But this is unnecessary for this PR.

For better context, would you mind sharing some screenshots or videos that highlight how this effect might look poor? I think that'd be useful to assess the impact enabling this fix by default would have.

Anyway, I think the idea of making it toggleable via a scripting function and a client setting can be a good thing. If the client disables the reflections in the settings for some reason, then scripts shouldn't be able to enable them (the function could return `false` to indicate this). If the client has them enabled, then it's up to the server scripts to decide how they use them with `setCoronaReflectionsEnabled`. I think this would make everyone involved happy.

lopezloo:
I'm looking forward implementing new video option, enabled by default.

> Anyway, I think the idea of making it toggleable via a scripting function and a client setting can be a good thing. If the client disables the reflections in the settings for some reason, then scripts shouldn't be able to enable them (the function could return `false` to indicate this). If the client has them enabled, then it's up to the server scripts to decide how they use them with `setCoronaReflectionsEnabled`. I think this would make everyone involved happy.

IMHO scripts should be able to overwrite this. Just like "Render vehicles always in high detail" option - player is in control until script changes it. It will be possible to check if player has it enabled via `dxGetStatus()`.

starfleet001:
> Just like "Render vehicles always in high detail" option - player is in control until script changes it

This is false. Unfortunately for some reason it was explicitly designed in a way that scripts can't override ped or vehicles LOD distances unless client checked "Render vehicles always in high detail" (or checkbox for peds). This behavior is creating disbalance on many PVP oriented servers because server owners don't have many options but to limit distances to default low values instead of selecting appropriate distances for all players.

In my opinion `setCoronaReflectionsEnabled` (or however you name it) should always override client's setting up until disconnect. After disconnect it should be reverted to client's setting. And we should do the same with ped and vehicle LOD functions.

lopezloo:
> This is false. Unfortunately for some reason it was explicitly designed in a way that scripts can't override ped or vehicles LOD distances unless client checked "Render vehicles always in high detail" (or checkbox for peds). This behavior is creating disbalance on many PVP oriented servers because server owners don't have many options but to limit distances to default low values instead of selecting appropriate distances for all players.

It looks like regression. It was supposed to be overwritable by scripts. Should be reported as seperate issue.

gta191977649:
I'm also against add it to the video settings, I don't think all the server onwer wants to enable this fix or let the player has the ability to toogle it by their own, e.g. in some special gamemode, the player might get some unfair advantage, like days.
So leave it in the script function is a good way and I personally think that is how should it done regarding to this issue.

thisdp:
> I'm also against add it to the video settings, I don't think all the server onwer wants to enable this fix or let the player has the ability to toogle it by their own, e.g. in some special gamemode, the player might get some unfair advantage, like days. So leave it in the script function is a good way and I personally think that is how should it done regarding to this issue.

It is fine to add this to video settings. We can make this feature can be "forced" set by function.

TheNormalnij:
Video demonstration: https://youtu.be/8tAhdsXql2I

This effect works thought objects  ;/

patrikjuvonen:
> Video demonstration: https://youtu.be/8tAhdsXql2I
> 
> This effect works thought objects ;/

Yeah, that's not good.

lopezloo:
- Added corona reflections video option (enabled by default).
- Added `resetCoronaReflectionsEnabled()` - resets state to player desired value.
- Renamed Marker `reflectionEnabled` property to `coronaReflectionEnabled`.

Player is in control until `setCoronaReflectionsEnabled()` is called. Then script is in control until disconnect, `resetCoronaReflectionsEnabled()` or `resetMapInfo()` being called.

`dxGetStatus().SettingCoronaReflections` can be used to retrieve player setting.

lopezloo:
There is not much I can do about the fact that these reflections are going through vehicles. This is how it was coded. It works the same way on PS2. It's simple effect. These reflections simply render after road objects and before other entities are being rendered.

This effect would be trivial to recreate via scripts if there would be an render event which is called after roads being rendered but before other entities. It's still possible though by applying shader to road textures.

thisdp:
need to be merged if everything is fine.

lopezloo:
Marking it as draft for a moment. I will move some code to `game_sa/CCoronasSA.cpp`.

lopezloo:
* Moved some code to `game_sa/CCoronasSA.cpp`.
* Removed Marker `coronaReflectionEnabled` OOP property and instead added `is/setCoronaReflectionEnabled()` OOP methods (I did it because it can return false if marker is not corona. It would be a bit weird setting object property to true and then it would automatically became false instantly).

patrikjuvonen:
> > Video demonstration: https://youtu.be/8tAhdsXql2I
> > 
> > This effect works thought objects ;/
> 
> Yeah, that's not good.

Because of this I can't see this as acceptable. We shouldn't introduce buggy default behavior for players. This should be opt-in instead, or someone with time to fix the underlying GTA/RW bug.

Lpsd:
I think it's fine as long as this isn't introduced as a default, but rather toggleable via `setWorldSpecialPropertyEnabled` or whatever.

lopezloo:
Well, I can make it disabled by default if needed.

patrikjuvonen:
> Well, I can make it disabled by default if needed.

Please do.

lopezloo:
I have disabled these reflections by default now. Players can turn them on via "Corona rain reflections" checkbox in video settings. Scripters can enable them via `setCoronaReflectionsEnabled` function (overrides player choice). Please review.

TheNormalnij:
Works for me

TheNormalnij (approved):

PR #2354, opened 2021-09-14
Fix #2349 (Interiors lack radio)

### LINKED ISSUE #2349: Interiors lack radio
**Describe the bug**
Some interiors in SA has defined audio zones, some of these zones are supposed to play radio. It doesn't work in MTA.

**To reproduce**
Go to Binco interior:
`
run me.position = Vector3(208.942, -102.377, 1005.258) me.interior = 15
`

**Expected behaviour**
Radio in interiors (ex. Binco) should play like it does in single player.

**Video**
http://youtu.be/qonn6DDIK8w

**Version**
Multi Theft Auto v1.5.8-release-20927

**Additional context**
MTA mess with radio somehow, there are NOPs present which disables it and it affects interior radios too.

Additionaly, if player teleports straight from interior which has background sound (but not radio) to interior which has radio defined, the sound from previous interior doesn't stop.

Kinda moved from https://bugs.multitheftauto.com/view.php?id=8565
There was patch for it: #22
From what I remember the issue with this patch was that it was also changing the last vehicle radio station ID, so if player left such interior and entered vehicle, the radio would be turned off by default. It could be fixed by actually playing radio but not saving station ID.

lopezloo:
This PR fixes radio in interiors (audio zones) does not work (#2349).

Since MTA disables GTA handling of radio and reimplements radio station switching, radio in interiors doesn't work. I have created hook into function which is supposed to switch station when player enter or leave audio zone to make it working great again.

It can be disabled by using `setInteriorSoundsEnabled`.


TheNormalnij (approved):

TheNormalnij:
Works for me. Interior radio can be disabled by setRadioChannel function. Code looks pretty enough. I think we can merge this

Dutchman101:
The crash from issue #2621 may have resulted from this PR

PR #2359, opened 2021-09-16
Use vector length method instead of manual distance calculation

NanoBob:
Simplifies distance calculation by using `.Length()` method on the `CVector` class

### DIFF
--- a/Server/mods/deathmatch/logic/CBuildingRemovalManager.cpp
+++ b/Server/mods/deathmatch/logic/CBuildingRemovalManager.cpp
@@ -73,13 +73,7 @@ void CBuildingRemovalManager::RestoreWorldModel(unsigned short usModel, float fR
         if (pFind)
         {
             const CVector& vecRemovalPos = pFind->GetPosition();
-            // Grab distances across each axis
-            float fDistanceX = vecPos.fX - vecRemovalPos.fX;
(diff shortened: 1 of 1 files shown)


Inder00:
That's very useful because its actually using sqrt which consuming a lot of cpu power.

NanoBob:
> That's very useful because its actually using sqrt which consuming a lot of cpu power.

Well not really, since the actual source of the `.Length()` method is
```cpp
inline float Length() const { return sqrt((fX * fX) + (fY * fY) + (fZ * fZ)); }
```

Pirulax:
`sqrt` isn't really all that expensive actually.
But if you really want to avoid it, square the stuff you test against, so you can just use the squared length.

Pirulax:
`sqrt` isn't really all that expensive actually.
But if you really want to avoid it, square the stuff you test against, so you can just use the squared length.

Eg.:
```cpp
float fDistance = (vecPos - vecRemovalPos).LengthSq();

if (fDistance <= std::pow(pFind->GetRadius(), 2.0f) && (cInterior == -1 || pFind->GetInterior() == cInterior))
```

Pirulax (approved):

Pirulax:
Whatever, I see the change wont be made, so I'm merging it.

PR #2368, opened 2021-09-21
Add missing getType method for camera element

TheNormalnij:
Added missing getType method for camera element
```lua
local camera = getCamera()
print( camera.type ) -- >> "camera"
print( camera:getType() ) -- >> "camera"
```

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
@@ -63,7 +63,7 @@ void CLuaCameraDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getClip", "getCameraClip");
     lua_classfunction(luaVM, "getFarClipDistance", "getFarClipDistance");
     lua_classfunction(luaVM, "getNearClipDistance", "getNearClipDistance");
-    lua_classfunction(luaVM, "getType", GetElementType);
+    lua_classfunction(luaVM, "getType", ArgumentParser<GetElementType>);
 
     lua_classfunction(luaVM, "setPosition", OOP_SetCameraPosition);
     lua_classfunction(luaVM, "setRotation", OOP_SetCameraRotation);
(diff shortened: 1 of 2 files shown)


TheNormalnij:
If it's important, I can use a new argument parser here.

patrikjuvonen:
> If it's important, I can use a new argument parser here. 

Please do, all new code must use the new parser when possible.

TheNormalnij:
Used

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -86,6 +87,7 @@ void CLuaCameraDefs::AddClass(lua_State* luaVM)
     lua_classvariable(luaVM, "position", OOP_SetCameraPosition, OOP_GetCameraPosition);
     lua_classvariable(luaVM, "rotation", OOP_SetCameraRotation, OOP_GetCameraRotation);
     lua_classvariable(luaVM, "matrix", NULL, OOP_GetCameraMatrix);
+    lua_classvariable(luaVM, "type", NULL, ArgumentParser<GetElementType>);
```
StrixG:
```suggestion
    lua_classvariable(luaVM, "type", nullptr, ArgumentParser<GetElementType>);
```

StrixG (approved):

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -590,3 +592,8 @@ int CLuaCameraDefs::OOP_SetCameraRotation(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+std::string CLuaCameraDefs::GetElementType()
+{
+    return std::string(m_pManager->GetCamera()->GetTypeName());
```
Pirulax: What does `GetTypeName` return?
Would be nice to avoid creating an extra string if not necessary. 
`std::string_view` could be used instead.

TheNormalnij: Used original SString

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.h
```diff
@@ -47,4 +47,6 @@ class CLuaCameraDefs : public CLuaDefs
     LUA_DECLARE(OOP_SetCameraPosition);
     LUA_DECLARE(OOP_GetCameraRotation);
     LUA_DECLARE(OOP_SetCameraRotation);
+
+    static SString GetElementType();
```
tederis: Can we use a const ref here? Like `static const SString& GetElementType()`?

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -593,7 +593,7 @@ int CLuaCameraDefs::OOP_SetCameraRotation(lua_State* luaVM)
     return 1;
 }
 
-SString CLuaCameraDefs::GetElementType()
+const SString& CLuaCameraDefs::GetElementType()
```
Pirulax, line 596: Honestly, `string_view` could be used

Pirulax:
I think we can merge this, right?

Issue #1632, opened 2020-08-23
Detect when player left MTA Window with Windowed mode


Arci224:
**Is your feature request related to a problem? Please describe.**
I want new function like a `onClientMinimaze`, which detect when player left window of MTA:SA. Because `onClientMinimaze ` function only with Full screen. And new function like a `onClientRestore`

**Describe the solution you'd like**
Add new Event Handler like `onClientLeftMTAWindow` which detect when player exit MTA Window and `onClientEnterMTAWindow` which detect when player enter MTA Window.
 
**Describe alternatives you've considered**
\-

**Additional context**
\-

Unde-R:
> I want new function like a `onClientMinimaze`, which detect when player left window of MTA:SA. Because `onClientMinimaze ` function only with Full screen. And new function like a `onClientRestore`
>
`onClientMinimize` & `onClientRestore` are events not functions.
They already exist but need some improvements.
[onClientRestore](https://wiki.multitheftauto.com/wiki/OnClientRestore)
[onClientMinimize](https://wiki.multitheftauto.com/wiki/OnClientMinimize)

Arci224:
@Unde-R uhhh, yes sorry my bad.

Arci224:
I think on that can be used Closed PR #185 .. (If it is still actual and can be usable.)

Issue #1113, opened 2019-10-05
setCursorPosition returns true even if MTA main window isn't focused in Windows


MrDadosz:
**Describe the bug**
setCursorPosition returns true even if MTA window isn't focused. When you have game in window mode or you have borderless mode, setCursorPosition returns true. 
You can't determine by any function or any event if player is in game. 

**To reproduce**
Watch YouTube: https://youtu.be/zjQcOE3c_V8
Script: https://pastebin.com/HgvuZb6x

Why it's important for me: https://youtu.be/OItkIjldqBM?t=14

**Expected behaviour**
There's should be function for check if MTA isn't focused or setCursorPosition shouldn't return true if MTA isn't focused

**Screenshots**
Look at "to reproduce"

**Version**
Multi Theft Auto v1.5.7-release-20288

CrosRoad95:
i think could be solved in way i proposed almost 2 years ago #185

qaisjp:
I think we'd be happy with you reviving that PR @CrosRoad95

CrosRoad95:
i remember i had a problems with doing it in way proposed by botder, i think he should finish it

Lpsd:
This is now resolved by #2371

PR #185, opened 2018-01-25
WIP: Event onClientMTAFocusChange

CrosRoad95:
This is event similar to `onClientMinimize` but, insted of checking did mta windows was minimized, this events are triggered when you lost focus on mta window.
If someone play on rpg server in window mode, then he never minimize them. This event can help programmer handle when player is afk

Example script clientside
```lua
addEventHandler("onClientMTAFocusChange", root, function (focus)
  outputChatBox(inspect(focus))
end)
```
Print `false` when you lost focus on mta window, when you back to game, will print `true`

PR #2371, opened 2021-09-22
Add isMTAWindowFocused() and onClientMTAFocusChange

### LINKED ISSUE #1006: onClientMinimize don't working on window mode
**Describe the bug**
onClientMinimize event isn't triggered if window mode is enabled.

**To reproduce**
Set window mode to on
```lua
addEventHandler('onClientMinimize', root, function()
      print('minimized')
end)
```

it's only working on fullscreen, without diffrence if game is minimized by ALT + TAB or by click outside the window.

**Expected behaviour**
onClientMinimize should also triggered on window mode

**Screenshots**

**Version**
1.5.6

**Additional context**

Lpsd:
Original PR by @CrosRoad95 #185

Adds 1 new client function
```lua
bool isMTAWindowFocused()
```

Adds 1 new client event, `onClientMTAFocusChange`, example:
```lua
local function handleFocusChange(focused)
    outputChatBox(focused and "gained focus" or "lost focus")
end
addEventHandler("onClientMTAFocusChange", root, handleFocusChange)

local function getWindowFocus()
    iprint(isMTAWindowFocused(), getTickCount())
end
setTimer(getWindowFocus, 1000, 0)
```

Resolves #1006, #1113 and #1632


Client/core/CMessageLoopHook.cpp
```diff
@@ -124,15 +124,34 @@ LRESULT CALLBACK CMessageLoopHook::ProcessMessage(HWND hwnd, UINT uMsg, WPARAM w
...
+                if (pBase)
+                {
+                    WORD state = LOWORD(wParam);
+                    bool focus = (state == WA_CLICKACTIVE) || (state == WA_ACTIVE);
+
+                    pBase->OnWindowFocusChange(focus);
+                }
+            }
+
+            if (LOWORD(wParam) == WA_ACTIVE)
```
Pirulax: maybe switch case here?
would look neater

Pirulax (dismissed):
Overall, at first glance, LGTM.

Client/core/CMessageLoopHook.cpp
```diff
@@ -124,15 +124,34 @@ LRESULT CALLBACK CMessageLoopHook::ProcessMessage(HWND hwnd, UINT uMsg, WPARAM w
...
+            if (pModManager && pModManager->IsLoaded())
+            {
+                CClientBase* pBase = pModManager->GetCurrentMod();
+
+                if (pBase)
+                {
+                    WORD state = LOWORD(wParam);
+                    bool focus = (state == WA_CLICKACTIVE) || (state == WA_ACTIVE);
+
+                    pBase->OnWindowFocusChange(focus);
```
patrikjuvonen: I believe variable naming should be consistent, so let's use Hungarian notation here as well.

Additionally, I believe the `state` variable could be moved up so that you can use it on e.g. line 144 and 148 as well, thus reducing the amount of repeated code for the same thing especially since it doesn't return here.

```suggestion
                    WORD wState = LOWORD(wParam);
                    bool bFocus = (wState == WA_CLICKACTIVE) || (wState == WA_ACTIVE);

                    pBase->OnWindowFocusChange(bFocus);
```

patrikjuvonen (dismissed):

Lpsd:
Change requests have been implemented & tested. Updated test script

Client/mods/deathmatch/CClient.h
```diff
@@ -32,6 +32,8 @@ class CClient : public CClientBase
     bool HandleException(CExceptionInformation* pExceptionInformation);
     void GetPlayerNames(std::vector<SString>& vPlayerNames);
 
+    void OnWindowFocusChange(bool state);
```
botder:
```suggestion
    void OnWindowFocusChange(bool state) override;
```

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -92,6 +92,8 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
     m_lastWeaponSlot = WEAPONSLOT_MAX;            // last stored weapon slot, for weapon slot syncing to server (sets to invalid value)
     ResetAmmoInClip();
 
+    m_bWasFocused = g_pCore->IsFocused();
```
botder:
```suggestion
    m_bFocused = g_pCore->IsFocused();
```

botder (changes requested):

botder (approved):

PR #2372, opened 2021-09-22
Add getProcessMemory function

samr46:
Solves #1563

Adds `getProcessMemory()`

This function returns the current memory usage of **gta_sa.exe** process. Essentially copied from [CMemStats.cpp](https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/CMemStats.cpp#L461-L463) (win api call)
Returned value in KB is equal to the one displayed in memory stats ("showmemstat" command)

It will be useful for big servers that push memory boundaries to the limit (about 3.3 GB) to avoid crashing the client.
It's not a secret that when you disconnect from one server and connect to another the current memory usage increases and increases (i personally saw values up to 1.6 GB usage after disconnect).
So if you add let's say 1 GB of models to this and some other things the usage will get dangerously close to the limit right after the loading.

This function should allow servers to determine if current memory usage is not too high for their resources. And in this case server could use alternative (more optimized) loading mode or notify the user and ask to restart MTA.
Also it could be used to monitor the memory during gameplay and notify the user if it's getting too high (so user could restart MTA before it unexpectedly crashes in a bad moment).

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
@@ -9,6 +9,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <Psapi.h>
 
 int CLuaFunctionDefs::GetValidPedModels(lua_State* luaVM)
 {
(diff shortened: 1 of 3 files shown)


Vadya963:
maybe it's better to add it to dxGetStatus?

samr46:
> maybe it's better to add it to dxGetStatus?

Maybe. I thought about this function in the first place since it has _TotalPhysicalMemory_ now. But didn't feel good to use this win api call in dxGetStatus. I guess we need more feedback about this.

notnorelock:
Add support for returning in bytes, that should be good.

Shuubaru:
what is the progress of this?

Pirulax:
Let's add this to `dxGetStatus` instead.

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -311,6 +312,18 @@ int CLuaFunctionDefs::GetPerformanceStats(lua_State* luaVM)
     return 1;
 }
 
+int CLuaFunctionDefs::GetProcessMemory(lua_State* luaVM)
```
CrosRoad95, line 315: Use new argument parser, don't old technique

botder:
Superseded by commit dd571b4793ac6773c634a1cdc6b28bfa00891127

PR #2026, opened 2021-01-18
Add SVG support (lunasvg)

Lpsd:
I've implemented [lunasvg](https://github.com/sammycage/lunasvg), which gives us the ability to create SVG documents, edit them on the fly, and also load SVG files (from path or raw).

## API (for initial release)

```lua
svg svgCreate(int width, int height [, string pathOrRawData, function callback ])
xmlnode svgGetDocumentXML(svg svgElement)
bool svgSetDocumentXML(svg svgElement, xmlnode svgDocument [, function callback ])
int, int svgGetSize(svg svgElement)
bool svgSetSize(svg svgElement, int width, int height [, function callback ])
```

`callback` is for when SVG has been updated, invoked when texture is available (if update successful).

Example usage
```lua
local square = svgCreate(300, 300)
local squareXML = svgGetDocumentXML(square)

-- Add a rect svg node
local rect = xmlCreateChild(squareXML, "rect")
xmlNodeSetAttribute(rect, "x", "0")
xmlNodeSetAttribute(rect, "y", "0")
xmlNodeSetAttribute(rect, "width", "100%")
xmlNodeSetAttribute(rect, "height", "100%")
xmlNodeSetAttribute(rect, "fill", "red")

-- Apply our changes
svgSetDocumentXML(square, squareXML)

-- Create a square with blank document
local clone = svgCreate(300, 300)

-- Set the fill to blue
xmlNodeSetAttribute(rect, "fill", "blue")

-- Apply xml to blank svg document
svgSetDocumentXML(clone, squareXML)

addEventHandler("onClientRender", root, function()
    dxDrawImage(500, 150, 300, 300, square, 0, 0, 0, tocolor(255, 255, 255), false)
    dxDrawImage(900, 350, 300, 300, clone, 0, 0, 0, tocolor(255, 255, 255), false)
end)
```
```lua
local monkey = svgCreate(500, 500, "monkey.svg")

addEventHandler("onClientRender", root, function()
    dxDrawImage(0, 0, 500, 500, monkey, 0, 0, 0, tocolor(255, 255, 255), false)
end)
```
<details>
<summary>Preview of the first example</summary>

image
</details>
<details>
<summary>Preview of the second example</summary>

image
</details>

The `svg` element is also derived from CClientTexture, meaning it works with pretty much all dx* functions, including `dxGetTexturePixels`, `dxSetTexturePixels`, `dxSetPixelColor`, etc - direct manipulation is therefore possible, e.g:

```lua
[...]

PR #2378, opened 2021-09-26
Fix missing return values (#2026)

Lpsd:
Should be the reason for `CC31, No such mod installed (deathmatch)` since r20979.

Also changed SVGSetSize to use a CVector2D instead of two ints.


Dutchman101:
Thanks, processing it quickly for a quick test - issue only reproduces on builds produced by MTA build server for some reason.

PR #2379, opened 2021-09-26
Remove unnecessary premake options in lunasvg project

Lpsd:
Following up from #2378 - another attempt to fix issue `CC31, No such mod installed (deathmatch)` since r20979.

Issue only occurs when using MTA build server... local builds, and GitHub artifacts (debug or release) work fine.


Dutchman101:
Merging on the same basis as PR #2378

PR #2380, opened 2021-09-26
Revert "Revert "Add SVG support (lunasvg) (#2026)""

patrikjuvonen:
Original PR #2026

Figure out why official nightly build from our build server causes deathmatch mod to not be loaded with this PR applied.

~~Something to keep in mind is the clang-format that was ran before these (see https://github.com/multitheftauto/mtasa-blue/compare/b42e14d838f5ae99450595509ac8119b0f4e5dcd...ec1b27a7a6f2b15b17a211ad2252056e9be0416d for clang-format changes) in case a change from there affects something in the compiler later and is triggered only by the svg changes.~~

Clang format does not cause the issue. Confirmed in r20987.

Not working: r20979, r20980, r20981, r20982, r20987


ccw808:
Running Process Monitor might help find the problem

ccw808:
Running Process Monitor might help find the problem
https://wiki.multitheftauto.com/wiki/Support_-_MTA_Client_and_Process_Monitor

patrikjuvonen:
@sbx320 found that because lunasvg was marked as SharedLib in premake5.lua, it depended on the then built lunasvg.dll which is not shipped by build server automatically. Hence changing from SharedLib to StaticLib should fix the issue. We also don't need the vendor package as separate dll.

PR #2383, opened 2021-09-28
Add dds pixels support for dxGetTexturePixels

samr46:
Adds dds pixels support for [dxGetTexturePixels](https://wiki.multitheftauto.com/wiki/DxGetTexturePixels)

### New Syntax

> string dxGetTexturePixels ( [ int surfaceIndex = 0, ] element texture **[, string pixelsFormat = "plain" [, string textureFormat = "argb"] [, bool mipmaps = true] ]** [, int x = 0, int y = 0, int width = 0, int height = 0 ] )

> **pixelsFormat:**
> plain
> dds

> **textureFormat:** (from [dxCreateTexture](https://wiki.multitheftauto.com/wiki/DxCreateTexture))
> argb
> dxt1
> dxt3
> dxt5

Main benefit of using dds pixels (native format) is speed / performance. Test resource includes benchmarks.
For example converting render target result into new texture is about 3 times faster than using "plain" format. Also properly saved DDS file will be loaded instantly and can be memory efficient depending on selected compression (should be useful for caching render target output).

### Notes:

- Current syntax / behavior is not affected by new changes;
- Added "xrgb" and "rgb" texture formats to enum (these formats are used by non transparent [render targets](https://wiki.multitheftauto.com/wiki/DxCreateRenderTarget));
- Technically default value of `textureFormat` is "unknown" (determined automatically based on texture) and not "argb" ("argb" used in syntax for consistency with [dxCreateTexture](https://wiki.multitheftauto.com/wiki/DxCreateTexture) syntax which also uses "unknown" by default instead of "argb"). It also supports more formats than listed on wiki;
- There is support for "png" and "jpeg" passed as `pixelsFormat` (D3DX function used to save DDS supports these formats as well) but I believe it shouldn't be mentioned on wiki. It's slower than existing form of conversion ("plain" -> "png / jpeg"). Tests are included;
- I didn't make any changes to [dxGetPixelsFormat](https://wiki.multitheftauto.com/wiki/DxGetPixelsFormat) because it already recognizes "dds" pixels (it's not mentioned on wiki for some reason).

Test resource: [d3dpixels.zip](https://github.com/multitheftauto/mtasa-blue/files/7244741/d3dpixels.zip)


Client/core/Graphics/CPixelsManager.cpp
```diff
@@ -467,6 +477,80 @@ bool CPixelsManager::SetSurfacePixels(IDirect3DSurface9* pD3DSurface, const CPix
...
+        return false;
+
+    ID3DXBuffer*                dxBuffer;
+    CAutoReleaseMe<ID3DXBuffer> Thanks1(dxBuffer);
+
+    D3DXIMAGE_FILEFORMAT dxFileFormat = D3DXIFF_DDS;
+    if (pixelsFormat == EPixelsFormat::PNG)
+        dxFileFormat = D3DXIFF_PNG;
+    else if (pixelsFormat == EPixelsFormat::JPEG)
+        dxFileFormat = D3DXIFF_JPG;
```
Pirulax: Maybe use a switch case here?

Pirulax (commented):
I've skimmed thru the code, havent seen anything weird.

Client/core/Graphics/CPixelsManager.cpp
samr46: Agreed. Looks better this way.

botder (approved):

botder:
@samr46 Can you update the wiki page in the near future?

Issue #2277, opened 2021-07-09
engineRequestModel (vehicle) bugs


Xenius97:
**Describe the bug**
1. Can't enter passanger seats
2. MTA can't get vehicle occupants (you get an error)
3. Doors waving if someone enters into custom vehicle
4. Can't get vehicle type (scripting can fix this)

**To reproduce**
Spawn requested model and load mod

**Expected behaviour**
Work as default vehicles without these issues

**Version**
1.5.8 -

**Additional context**
https://i.imgur.com/yQG6NxQ.gif

jjustns:
Another bug that occurs is the tires appearing flattened 

https://user-images.githubusercontent.com/39979049/126101791-722aa04a-ae41-40cc-ab05-d457a3957e71.mp4

SDabdlmounaim96:
related to my other 2 issues :
#2166 
#2127

Xenius97:
Bikes and boats are working.
https://streamable.com/wj5sl3
https://streamable.com/i5r7sg

Planes are bugged too.
https://streamable.com/w9175t
https://streamable.com/l26qix

Heli:
https://streamable.com/ihrqyh

Xenius97:
Door waving and wheel states bug still exiting + random crashes

Version = 1.5.9-release-21019.0.000
Time = Tue Oct 19 11:02:50 2021
Module = C:\Games\MTA San Andreas\mta\game_sa.dll
Code = 0xC0000005
Offset = 0x00043277

EAX=00000000  EBX=00000001  ECX=0000022F  EDX=00000000  ESI=1DA7C4E8
EDI=18280481  EBP=0177FC24  ESP=0177FC18  EIP=5BDC3277  FLG=00010206
CS=0023   DS=002B  SS=002B  ES=002B   FS=0053  GS=002B


https://streamable.com/2k3f95

Fernando-A-Rocha:
@BCG2000 @Pirulax

BCG2000:
Well, these issues never happen while alone. I guess it is something to do with the packets since the server still thinks the custom model is the parent model the car was created as, so the tyres and doors come up as invalid in the client during sync (since the client model doesn't match the one on the server). I want to take a look at it when I can, or someone else will.

That being said, I don't think the crash has to do with my fix that was recently merged, just don't see why it would cause that.

Fernando-A-Rocha:
I can confirm the issues don't happen alone. While driving a car with a requested model you experience no visual bugs.

[...]

PR #2398, opened 2021-10-05
Partial fix (part 3) #2277, fix #2034: Fix sync door opening, fix sync of vehicle parts (doors, lights, panels and wheels) states on custom allocated vehicle model IDs

### LINKED ISSUE #2034: Add-on vehicle behaving weird
After requesting a model and using setElementModel
on every client, player reports issue of seeing vehicle doors constantly moving.
Other issues include: 'Network Trouble' when player tries to get inside vehicle

**To reproduce**
1. Download the reproduction resource
2. Create a vehicle server-side and enter it
3. Use '/request' command
4. See bug
Reproduction resource: 
[model_requesting.zip](https://github.com/multitheftauto/mtasa-blue/files/5851382/model_requesting.zip)

**Expected behaviour**
 Normal vehicle doors movement in all clients with the ability of players getting into the vehicle.

**- server/client version: 1.5.8**

Addlibs:
This pull makes the server treat vehicle IDs outside the range 400-611 (VEHICLE_NONE) as having doors, thus enables sync, which the client expected and was misinterpreting wrong data as door open ratio sync, which had a knock-on effect on door, lights, panels and wheel states sync.

Because to the client the vehicle copies the type of its parent, if it's a doorless type (i.e. boat, bike, bmx, etc.) it wouldn't normally send door sync. This pull forces this data to be sent for all custom vehicle IDs, the sever always reads this data for non-standard models, always sends it off to other clients, and all clients (with this patch) will read this data and apply it only if it makes sense (i.e. the vehicle has doors on the client)

### DIFF
--- a/Client/mods/deathmatch/logic/CNetAPI.cpp
+++ b/Client/mods/deathmatch/logic/CNetAPI.cpp
@@ -1910,7 +1910,7 @@ void CNetAPI::ReadFullVehicleSpecific(CClientVehicle* pVehicle, NetBitStreamInte
     }
 
     // Read door angles.
-    if (CClientVehicleManager::HasDoors(iRemoteModelID))
+    if (CClientVehicleManager::HasDoors(iRemoteModelID) || !CClientVehicleManager::IsStandardModel(iRemoteModelID))
     {
         SDoorOpenRatioSync door;
         for (unsigned char i = 2; i < 6; ++i)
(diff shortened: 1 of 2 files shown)


BCG2000:
I think https://github.com/multitheftauto/mtasa-blue/pull/2421 is a better way of fixing this, because it addresses the root issue (model ID mismatch between client and server).

lopezloo:
Closing this because #2421 was merged.

PR #2399, opened 2021-10-05
Fix #868, #1662 and #1740: Fix sync of light states (damage) and wheel states

### LINKED ISSUE #868: Wheels and lights are desynced when player damage it
**Describe the bug**
When someone damage lights or wheels (shoot at wheels/hit a wall and damage lights), they will be desynced. Player has to stream out his vehicle and stream it in, after that other players will see that lights or tires are broken. I reported this on MantisBT and I can't find link to this bug, but it's still broken.

**To reproduce**
https://youtu.be/JkzLQ5bUgLU

**Expected behaviour**
Damages should be visible after player damage his wheels or lights, not after streamOut and streamIn

**Screenshots**

**Version**
client: Multi Theft Auto v1.5.6-release-16588
server: Multi Theft Auto v1.5.6-release-16609

### LINKED ISSUE #1662: Vehicle wheel states aren't synced by default
- Wheel states (inflated, flat, etc.) are not synced across clients.
e.g. When I shoot someone's vehicle tire and it gets popped, only the driver (could be for passengers too, I'm not sure) of that vehicle sees them flat.

- Wheel states should be synced across clients by default.
i.e. If I (player A) shoot the wheels of player B's car, it should look flat for me and others, and not just for player B.

none

**Additional context**
Since the vehicle damage models are synced (a broken door, would be broken for every player), why shouldn't the wheel states be synced too?

### LINKED ISSUE #1740: Sync vehicle tire states

currently setVehicleWheelStates(getPedOccupiedVehicle(localPlayer), ...) does not sync the wheel states automaticly
while setVehiclePanelState(getPedOccupiedVehicle(localPlayer), ...) does sync automcliy between clients

automaticly sync wheel states

manually syncing with triggerServer/ClientEvent or setElementData

Addlibs:
Fixes #868 
Fixes #1662 
Fixes #1740

Essentially reverts d51aee516f20ae17c8b96af01d220bfd089f982e from 7 years ago (originally implemented 9bb96e5c421aecb339717b804b6c8b19f0b3c156)
This commit caused the phase to always be an even number (last bit always 0), which prevented wheel and light states to sync in the following code https://github.com/multitheftauto/mtasa-blue/blob/ca3b0b778e9695ce7da6391eda14429aa055d1f2/Server/mods/deathmatch/logic/net/CSimVehiclePuresyncPacket.cpp#L401-L416 https://github.com/multitheftauto/mtasa-blue/blob/16769b8d1c94e2b9fe6323dcba46d1305f87a190/Shared/sdk/net/SyncStructures.h#L1594-L1609 as the phase takes the last two bits (bitwise AND with 3, 0b11) and sequences sync packets between them.

A side effect of this pull is that door states and panel states will be synced half as frequently, with light and wheel syncs taking place instead at half those times.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -2331,9 +2331,6 @@ void CGame::Packet_VehiclePuresync(CVehiclePuresyncPacket& Packet)
             // Increment counter to spread out damage info sends
             pVehicle->m_uiDamageInfoSendPhase++;
 
-            // Increment counter to spread out damage info sends
-            pVehicle->m_uiDamageInfoSendPhase++;
-
             CLOCK("VehiclePuresync", "RelayPlayerPuresync");
             // Relay to other players
             RelayPlayerPuresync(Packet);


BCG2000 (approved):
I tested it with a friend here and I can confirm it closes #868, #1662 and #1740. The vehicle damage always syncs now, when before it was a hit or miss.

image

BCG2000 (approved):
EDIT: Ignore this duplicate message.

PR #2401, opened 2021-10-06
Remove forced TADDRESS_CLAMP for VectorGraphic render item

### LINKED ISSUE #2397: SVG and dxDrawImageSection the image is not cropped properly
when I try to crop an image from top to bottom it gets all blurry
[Link image bug](https://ibb.co/Zg1y0hv) 

```
local rawSvgData = [[

<svg xmlns="http://www.w3.org/2000/svg"
aria-label="Apple" role="img"
viewBox="0 0 512 512"><rect
width="512" height="512"
rx="15%"
fill="#555"/><path fill="#f2f2f2" d="M410 334s-10 29-30 59c-5 9-29 43-58 43-21 0-35-15-62-15-33 0-46 15-67 15-11 1-22-5-34-16-77-73-81-181-52-225 18-29 48-47 81-48 26 0 54 17 65 17 8 0 50-20 74-18 33 3 56 15 73 38-49 24-66 117 10 150zM329 56c8 32-27 93-79 90-3-43 34-87 79-90z"/></svg>

]]

local svg

function init()
    -- Create an SVG containing a circle, using the raw XML data above
    svg = svgCreate(212, 212, rawSvgData, function(didLoad)
        if (not didLoad) then -- SVG failed to load, so don't continue
            return 
        end

    end)
end

function render()
    local h = 0
    dxDrawImageSection(50, 150, 212, 212, 0, 100, 212, 212, svg, 0,0,0, tocolor(255,255,255,255))
end
```

-server/client version (1.5.9)

Lpsd:
Resolves #2397 

Removes forced `TADDRESS_CLAMP` (should default to `TADDRESS_WRAP` - the same as **dxCreateTexture**).

It may introduce edge artifacts when rotating the svg directly via dxDrawImage - in this case they can use **dxCreateTexture** like the workaround in #2397 but instead set `textureEdge` to `clamp`.

### DIFF
--- a/Client/core/Graphics/CRenderItem.VectorGraphic.cpp
+++ b/Client/core/Graphics/CRenderItem.VectorGraphic.cpp
@@ -25,8 +25,6 @@ void CVectorGraphicItem::PostConstruct(CRenderItemManager* pManager, uint width,
     m_uiSurfaceSizeX = width;
     m_uiSurfaceSizeY = height;
 
-    m_TextureAddress = TADDRESS_CLAMP;
-
     CreateUnderlyingData();
 }
 


Lpsd:
~~Out of scope for this PR, but maybe we can introduce a function to set the SVG texture edge type directly, so you don't have to duplicate the SVG texture through dxCreateTexture just to change it?~~

duh
```lua
dxSetTextureEdge(svg, "border")
```

Lpsd:
Would someone mind approving this 1 line change?

lopezloo (approved):

PR #2408, opened 2021-10-09
Fix camera logic after warpPedIntoVehicle

### LINKED ISSUE #2402: Issue with camera target after 1.5.9 release
**Describe the bug**
Removing a player from a vehicle then warping them to a very far away vehicle causes camera to be stuck on the initial one instead of the camera being set on the new vehicle.
Please note that this happens randomly, and that the ped gets warped to a different dimension and a custom map in that dimension as well. The ped's dimension is set beforehand, so is the vehicle's.

**Expected behaviour**
Camera moving along with the ped when being warped to the farther vehicle, as it was the case before the 1.5.9 release.

**Screenshots**
image
image

TheNormalnij:
This PR restores camera autofocus on new car after `warpPedIntoVehicle` , if camera was focused on previous car

Fixes #2402

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -1424,6 +1424,7 @@ void CClientPed::WarpIntoVehicle(CClientVehicle* pVehicle, unsigned int uiSeat)
         }
     }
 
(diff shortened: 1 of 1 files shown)


lopezloo (approved):
Looks good to me.

PR #2409, opened 2021-10-09
Add openBrowserURL function

Inder00:
This pull request adding **openBrowserURL** function **client-side only**.
Function opens the url in the client default system browser.

# openBrowserURL
```lua
bool openBrowserURL(string url)
```
Variable `url` must starts with **`http://`** or **`https://`** otherwise will returns false also if failed to open url returns false.

---------------------------------------
This functions can be used for example to open oauth2 url.
Thank you.

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -227,6 +227,16 @@ bool CStaticFunctionDefinitions::DownloadFile(CResource* pResource, const char* 
     return false;
 }
 
+bool CStaticFunctionDefinitions::OpenBrowserURL(const SString& strUrl)
+{
+    if (strUrl.BeginsWithI("http://") || strUrl.BeginsWithI("https://"))
+    {
+        return ShellExecuteNonBlocking("open", strUrl.c_str());
+    }
+
+    return false;
+}
+
 bool CStaticFunctionDefinitions::OutputConsole(const char* szText)
 {
     m_pCore->GetConsole()->Print(szText);
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
@@ -36,6 +36,7 @@ public:
 
     // Misc funcs
     static bool DownloadFile(CResource* pResource, const char* szFile, CResource* pRequestResource, CChecksum checksum = CChecksum());
+    static bool OpenBrowserURL(const SString& strUrl);
 
     // Output funcs
     static bool OutputConsole(const char* szText);
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
@@ -396,6 +396,29 @@ int CLuaFunctionDefs::DownloadFile(lua_State* luaVM)
     return 1;
 }
 
+int CLuaFunctionDefs::OpenBrowserURL(lua_State* luaVM)
+{
+    SString         strURL;
+
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadString(strURL);
+
+    if (!argStream.HasErrors())
+    {
+        if (CStaticFunctionDefinitions::OpenBrowserURL(strURL))
+        {
+            lua_pushboolean(luaVM, true);
+            return 1;
+        }
+    }
+    else
+    {
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+    }
+    lua_pushboolean(luaVM, false);
+    return 1;
+}
+
 int CLuaFunctionDefs::AddDebugHook(lua_State* luaVM)
 {
     //  bool AddDebugHook ( string hookType, function callback[, table allowedNames ] )
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
@@ -43,6 +43,7 @@ public:
 
     // Misc functions
     LUA_DECLARE(DownloadFile);
+    LUA_DECLARE(OpenBrowserURL);
 
     // Output functions
     LUA_DECLARE(OutputConsole);
--- a/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
@@ -197,6 +197,7 @@ void CLuaManager::LoadCFunctions()
         // Util functions
         {"getValidPedModels", CLuaFunctionDefs::GetValidPedModels},
         {"downloadFile", CLuaFunctionDefs::DownloadFile},
+        {"openBrowserURL", CLuaFunctionDefs::OpenBrowserURL},
 
         // Input functions
         {"bindKey", CLuaFunctionDefs::BindKey},


Lpsd:
```lua
local function render()
    openBrowserURL("https://www.google.com")
end
```

easy way to freeze up a client's PC (and crash it)

Dutchman101:
> easy way to freeze up a client's PC (and crash it)

That's true, but there are countless of other existing ways to cause a client freeze/crash/5 FPS with very dumb, or deliberately unoptimized code.

Lpsd:
> > easy way to freeze up a client's PC (and crash it)
> 
> That's true, but there are countless of other existing ways to cause a client freeze/crash/5 FPS with very dumb, or deliberately unoptimized code. If that happens the player just won't return to a certain server.

In probably all of those cases it's limited to the MTA/GTA process, so worst case is MTA will crash (or you'll get network trouble, etc) - and in most of those cases MTA handles the crash "gracefully".

Here, it's increasing memory usage in an external program/browser instead (which could be anything). Your MTA and browser will crash, maybe everything running on your PC too.

Not to mention this almost locks up your PC when it's done, due to the focusing of the browser when a new tab is opened every frame. So even dealing with this once, you'll struggle to close the MTA process - some users may resort to just restarting their PC (if it doesn't crash before then).

Knowingly implementing a feature with this potential for abuse (however likely) feels shitty. Limit usage to once every second or so and ensure the same link isn't opened twice - that's enough.

Also, shouldn't we be enforcing the client domain/URL whitelist here? (`requestBrowserDomains`)

Dutchman101:
> In probably all of those cases it's limited to the MTA/GTA process
> Here, it's increasing memory usage in an external program/browser instead

CEF entirely runs in the game process and its address space (speaking of where all resource usage/load goes - it's not CEFLauncher.exe which is a stub in this context), so memory and CPU will grow from abusing CEF just as it does from other types of Lua script abuse. I don't know where your personal experience with having trouble to close the game comes from, but that means the same can happen to you with anything else non-CEF.

Lpsd:
> > In probably all of those cases it's limited to the MTA/GTA process
> > Here, it's increasing memory usage in an external program/browser instead
> 
> CEF entirely runs in the game process and its address space (speaking of where all resource usage/load goes - it's not CEFLauncher.exe which is a stub in this context), so memory and CPU will grow from abusing CEF just as it does from other types of Lua script abuse. I don't know where your personal experience with having trouble to close the game comes from, but that means the same can happen to you with anything else non-CEF.

This function opens a URL in your default browser, on Windows, not through MTA's CEF implementation.

Dutchman101:
This PR has to be rejected, MTA operates as sandbox when it comes to resources/scripting and this sort of function would compromise security and have other consequences.

For example, it will be easier to exploit browser bugs or send/redirect the player to a page that downloads or deploys malware. Even if we add a prompt whether or not the player wants to allow going somewhere, we can't rely on our users not being naive enough.

In case of some CEF vulnerability, exploiting it inside the game can be way more difficult (due to constraints) than outside of it, especially with the risk that a malicious server owner somehow knows which browser (and version of it) the user has at that moment, and so to what exactly it's vulnerable, then send the user to a page crafted for exploitation. Related to that, there's not as much of a guarantee that system browser versions are kept up-to-date, as it is with MTA's regular CEF updates (and client auto updater)

There's more risks paired with implementing something like this, but i probably mentioned enough of them..
We just have to adhere to some core principles of security-by-design in MTA.

CrosRoad95:
1. mta could limit use of this function to once, per eg 10 seconds ( easy to make )
2. player could disable option to open browser ( enabled by default )
also, server could have "trusted" flag, what opens some functionality like inder mantioned and function `isServerTrusted` function to check if my server was added to trusted by mta team? ( or mta team could assign server X can open website Y, server website only )

PlatinMTA:
> 2. player could disable option to open browser ( enabled by default )
>    also, server could have "trusted" flag, what opens some functionality like inder mantioned and function `isServerTrusted` function to check if my server was added to trusted by mta team? ( or mta team could assign server X can open website Y, server website only )

I would never trust a server, even if its on a trusted list. Everything is on a black box, we can't really know what is happening.

Tho I do like the idea to have an ability to redirect players to a website without using the browser implementation of MTA. It can be useful to link players to social media or stores.

dmi7ry:
Many game engines has a same function. And it may be useful. For an example, if user want make payment, he doesn't know actually is it real payment page in MTA's window or not, maybe it's fraud page. If the page opened in the system browser, user may control it.

For example, MTA might ask user, is it allowed to open a link in the system browser. And option to remember the choice.

Inder00:
> Many game engines has a same function. And it may be useful. For an example, if user want make payment, he doesn't know actually is it real payment page in MTA's window or not, maybe it's fraud page. If the page opened in the system browser, user may control it.
> 
> For example, MTA might ask user, is it allowed to open a link in the system browser. And option to remember the choice.


Create a option to open system browser in settings or restrict list of hosts and ask player to accept them.
Also a usage limitation is needed like 1/10sec as @CrosRoad95 said.

Inder00:
> For example, it will be easier to exploit browser bugs or send/redirect the player to a page that downloads or deploys malware. Even if we add a prompt whether or not the player wants to allow going somewhere, we can't rely on our users not being naive enough.

You can also deploy a malware via mta http. As I said higher we should add a restricted list of hosts and ask player to accept them (That is actually implemented https://wiki.multitheftauto.com/wiki/RequestBrowserDomains) or make a new setting which allows to open urls in system browser.

> In case of some CEF vulnerability, exploiting it inside the game can be way more difficult (due to constraints) than outside of it, especially with the risk that a malicious server owner somehow knows which browser (and version of it) the user has at that moment, and so to what exactly it's vulnerable, then send the user to a page crafted for exploitation. Related to that, there's not as much of a guarantee that system browser versions are kept up-to-date, as it is with MTA's regular CEF updates (and client auto updater)
> 
> There's more risks paired with implementing something like this, but i probably mentioned enough of them.. We just have to adhere to some core principles of security-by-design in MTA.

For example you want login player to 3-rd party service (like google oauth) from CEF you can use https://wiki.multitheftauto.com/wiki/ExecuteBrowserJavascript and interact fully with website otherwise in player's system browser you can't.

Dutchman101:
Final thoughts were shared in development discord, see [https://i.imgur.com/1R8Jny6.png](https://i.imgur.com/1R8Jny6.png) and click for zoom

All later opinions were, if you read detail, already countered by what i wrote in this issue before, but the main difference is that some people think reducing security (as what they, not MTA, consider to be just a little bit) is reasonable for this feature. But MTA will never knowingly take any action that potentially messes with the safety of our users.

This kind of thing would equal adding a new attack vector, that can continue on into lengths you wouldn't think of yet. But among the most logical of points are:

> In case of some CEF vulnerability, exploiting it inside the game can be way more difficult (due to constraints) than outside of it, especially with the risk that a malicious server owner somehow knows which browser (and version of it) the user has at that moment, and so to what exactly it's vulnerable, then send the user to a page crafted for exploitation. Related to that, there's not as much of a guarantee that system browser versions are kept up-to-date, as it is with MTA's regular CEF updates (and client auto updater)

The risk of server owners somehow managing to, outside of MTA, fingerprint & create profiles on users by software characteristics (in order to target attacks made possible by such a new vector - knowing they are using a vulnerable browser or browser version) _is real_. But still like i said, vectors like this can go deeper than you'd expect, MTA has no control over the safety status of an user's browser.. they can call on features that aren't available in MTA's CEF for security/sandboxing reasons, for directly nefarious purposes (initial exploitation of a vulnerability) but also to actually use the system browser for fingerprinting data collection, which in turn will enable yet another type of attack or attack stage. And so it goes on, recursively into the depths of a web of vectors.

Besides that the idea "its fine as long the user gets a prompt to allow/deny opening their browser beforehand" was answerred like this:

> you can't trust masses of users not to be naive
.. or to want something a server claims/puts as "neccesary to play".. literally, kids may want it so badly that they dont consider denying the prompt. 

Nb: of course, server owners or other malicious parties can tell anyone that they need to go to a certain bad page, but having it inside the game, even with a prompt, means the bar is much lower and more people will conveniently go ahead. It is much easier for server scripters to make that thing intuitive and "gameplay blocking" or desired (feature blocking, like VIP, if they don't do it) if that is scriptable in-game. So you can't deny that it means a risk for larger masses than just telling an user to click their browser and go somewhere.

Bottom line is that MTA attachs a lot of value to security-by-design, in order to keep a maximum security level for our users.. we value our sandbox/Lua VM principles as well. We won't compromise a bit, so in our eyes, nothing justifies it.

PR #2016, opened 2021-01-17
Add Base32 & Base64 on encodeString

fresholia:
Fixes #2002
Adding encodeString base32&base64

Example:
```lua
local str = encodeString("base32", "anan", {})
print(decodeString("base32", str, {})) -- output: anan

local str = encodeString("base64", "anan", {})
print(decodeString("base64", str, {})) -- output: anan

```

When the PR is approved, I will update it on the wiki.

PR #2412, opened 2021-10-12
Add RSA support on encodeString

### LINKED ISSUE #2097: RSA, AES support
Add `RSA`, `AES` cryptography algorithms programically support.

Inder00:
Fixes #2097
This pull request adds **RSA Support**.

# generateKeyPair
```lua
string, string generateKeyPair( string algorithm, table options, [ function callback ])
```
Available algorithms: `rsa`
Return strings representing private and public key
# RSA on encode/decodeString
```lua
string encodeString( "rsa", string, { key = publicKey })
string decodeString( "rsa", string, { key = privateKey})
```

Resource for testing: [rsatest.zip](https://github.com/multitheftauto/mtasa-blue/files/7397299/rsatest.zip)
Thank you.


Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,6 +231,52 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
     return 1;
 }
 
+std::variant<bool, CLuaMultiReturn<SString, SString>> CLuaCryptDefs::GenerateRsaKeyPair(lua_State* const luaVM, int size,
+                                                                                        std::optional<CLuaFunctionRef> callback)
+{
+    // size chceck (based on https://www.cryptopp.com/wiki/RSA_Cryptography)
+    if (size <= 0)
```
CrosRoad95: why? you check anyway if size is < 128

Inder00:
Changed enum name because of https://github.com/multitheftauto/mtasa-blue/pull/2016#discussion_r560499522

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,53 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
     return 1;
 }
 
+std::variant<bool, CLuaMultiReturn<SString, SString>> CLuaCryptDefs::GenerateRsaKeyPair(lua_State* const luaVM, int size,
+                                                                                        std::optional<CLuaFunctionRef> callback)
+{
+    // keysize checks (based on https://www.cryptopp.com/wiki/RSA_Cryptography)
+    if (size < 128)
+        throw std::invalid_argument("Size cannot be lower than 128 bits");
```
Pirulax:
```suggestion
        throw std::invalid_argument("Size cannot be less than 128 bits");
```

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,53 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+    if (size > 4096)
+        throw std::invalid_argument("Size cannot be further than 4096 bits");
```
Pirulax:
```suggestion
        throw std::invalid_argument("Size cannot be further than 4096 bits");
```

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -32,6 +33,60 @@ namespace SharedUtil
         return result;
     }
 
+    inline std::pair<std::string, std::string> GenerateRsaKeyPair(const int size)
```
Pirulax: Take `unsigned int`, because `GenerateRandomWithKeySize` expects an unsigned int as well.
```suggestion
    // Returns a pair of private and public keys, both of the specified size.
    inline std::pair<std::string, std::string> GenerateRsaKeyPair(const unsigned int size)
```

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -32,6 +33,60 @@ namespace SharedUtil
...
+        CryptoPP::RSA::PrivateKey rsaPrivateKey(params);
+        CryptoPP::RSA::PublicKey  rsaPublicKey(params);
+
+        CryptoPP::StringSink rawPrivateKeySink(rawPrivateKey);
+        rsaPrivateKey.DEREncode(rawPrivateKeySink);
+
+        CryptoPP::StringSink rawPublicKeySink(rawPublicKey);
+        rsaPublicKey.DEREncode(rawPublicKeySink);
+
+        return std::make_pair(rawPrivateKey, rawPublicKey);
```
Pirulax: If this is raw binary, it should be converted to hex before passing it to Lua.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,53 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+
+                        arguments.Call(pLuaMain, luaFunctionRef);
+                    }
+                });
+        }
+    }
+    else
+    {
+        // Sync
+        return SharedUtil::GenerateRsaKeyPair(size);
```
Pirulax: How slow is this function? Is it noticeable when its ran in sync? If not, it should not have a callback.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,53 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+        CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+        if (pLuaMain)
+        {
+            CLuaShared::GetAsyncTaskScheduler()->PushTask<std::pair<SString, SString>>(
+                [size]
+                {
+                    // Execute time-consuming task
+                    return SharedUtil::GenerateRsaKeyPair(size);
+                },
+                [luaFunctionRef = callback.value()](std::pair<SString, SString> var)
```
Pirulax: `GenerateRsaKeyPair` returns a pair of `std::string` not `SString`.
Also, rename variable to "result" or something more meaningful.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -362,6 +407,74 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+                                }
+                                return result;
+                            },
+                            [luaFunctionRef](const SString result)
+                            {
+                                CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaFunctionRef.GetLuaVM());
+                                if (pLuaMain)
+                                {
+                                    CLuaArguments arguments;
+                                    if (result == nullptr)
```
Pirulax: How the hell would an `SString` be null? 
I assume you wanted to check if it's empty. For that, use `.empty()`

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -362,6 +407,74 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+
+                        lua_pushboolean(luaVM, true);
+                    }
+                }
+                else            // Sync
+                {
+                    std::string result;
+                    try
+                    {
+                        result = SharedUtil::RsaEncode(data, key);
```
Pirulax: I'd suggest to do this directly: (Also, above you've checked if it's (presumably) empty, dont you want to do that here as well? If its necessary of course..):
```cpp
                    lua::Push(SharedUtil::RsaEncode(data, key));
                    return 1;
```

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -516,6 +629,74 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
...
+                                }
+                                return result;
+                            },
+                            [luaFunctionRef](const SString result)
+                            {
+                                CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaFunctionRef.GetLuaVM());
+                                if (pLuaMain)
+                                {
+                                    CLuaArguments arguments;
+                                    if (result == nullptr)
```
Pirulax: Same as above.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -516,6 +629,74 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
...
+
+                        lua_pushboolean(luaVM, true);
+                    }
+                }
+                else            // Sync
+                {
+                    SString result;
+                    try
+                    {
+                        result = SharedUtil::RsaDecode(data, key);
```
Pirulax: Same as above.

Shared/sdk/SharedUtil.Tests.hpp
```diff
@@ -801,6 +801,26 @@ void SharedUtil_Hash_Tests()
         TEST_END
     }
 
+    // RSA (keypair generation, encryption and decryption)
+    {
+        TEST_FUNCTION
+        std::pair<std::string, std::string> keypair = GenerateRsaKeyPair(keysize);
```
Pirulax: `auto [private, public] = GenerateRsaKeyPair(keysize)`

Change the code to use this.

Pirulax (changes requested):

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
Inder00: Key Pair generation can took up to 0.7 sec with 4096-bit size so it should have both of implementation

Pirulax:
I think the returned key should be in `hex` for consistency with other stuff we already have (like md5) What do others think? 

React with emojis.

AlexTMjugador:
I think that representing keys with hexadecimal characters in this PR would go against the precedents set by the `encodeString` and `teaEncode` functions to treat encryption keys as binary strings. Therefore, in my opinion, treating keys as binary strings in this PR minimizes Lua API perplexity.

However, I agree that hash functions should continue to return hexadecimal strings for their results.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -362,6 +412,73 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+                    {
+                        CLuaShared::GetAsyncTaskScheduler()->PushTask<SString>(
+                            [data, key]
+                            {
+                                SString result;
+                                try
+                                {
+                                    result = SharedUtil::RsaEncode(data, key);
+                                }
+                                catch (const CryptoPP::Exception&)
```
Pirulax: would be nice to do something with the exception though...

Pirulax (approved):
LGTM if the test resource passes.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,58 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
     return 1;
 }
 
+std::variant<bool, CLuaMultiReturn<SString, SString>> CLuaCryptDefs::GenerateRsaKeyPair(lua_State* const luaVM, int size,
+                                                                                        std::optional<CLuaFunctionRef> callback)
+{
+    // keysize checks (based on https://www.cryptopp.com/wiki/RSA_Cryptography)
+    if (size < 128)
```
sbx320: These limitations look very odd to me. RSA128 (allowed) is very weak, RSA8192 (disallowed) is somewhat reasonable (although a bit on the paranoid scale...).

I'd say we should either prevent < RSA2048 or just allow everything and let the user decide.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,58 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+        throw std::invalid_argument("Size cannot be less than 128 bits");
+    if (size > 4096)
+        throw std::invalid_argument("Size cannot be further than 4096 bits");
+
+    // Since 2015, NIST recommends a minimum of 2048-bit keys for RSA, an update to the widely-accepted recommendation of a 1024-bit minimum since at least 2002.
+    // See: https://en.wikipedia.org/wiki/Key_size
+    if (size < 2048)
+        m_pScriptDebugging->LogWarning(luaVM, "Key sizes less than 2048-bit isn't recommened. Key Pair might be incesure.");
```
sbx320: If we allow < 2048 keys issuing a warning seems pointlessly annoying here.

Shared/sdk/SharedUtil.Tests.hpp
```diff
@@ -801,6 +801,26 @@ void SharedUtil_Hash_Tests()
         TEST_END
     }
 
+    // RSA (keypair generation, encryption and decryption)
```
sbx320: Note that this test does not actually verify RsaEncode/RsaDecode do anything.

SString RsaEncode(SString data, SString key) { return data; }
SString RsaDecode(SString data, SString key) { return data; }

would also succeed the test. Ideally we would have a proper test which attempts to encrypt/decrypt specific data and compares it to a reference.

sbx320 (dismissed):
Some minor things. 

Also we need some nicer way of handling async callbacks... too much duplicaiton here. I'll have a look for getting a generalized solution for it going. Doesn't matter for this PR though.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
Inder00: Keypair generation for 8192-bit keysize took 43 sec 
```
cryptopp$ rsa_kgen.exe 8192
Elapsed time for 8192 RSA key: 43.08s (43 seconds)
```
Source: https://www.cryptopp.com/wiki/RSA_Cryptography

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -362,6 +405,73 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+                                    {
+                                        m_pScriptDebugging->LogWarning(luaFunctionRef.GetLuaVM(), result.first.c_str());
+                                        arguments.PushBoolean(false);
+                                        arguments.Call(pLuaMain, luaFunctionRef);
+                                    }
+                                    else
+                                    {
+                                        arguments.PushString(result.first);
+                                        arguments.Call(pLuaMain, luaFunctionRef);
+                                    }
```
Pirulax: invert logic please.

Inder00:
@sbx320 review

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,50 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+        CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+        if (pLuaMain)
+        {
+            CLuaShared::GetAsyncTaskScheduler()->PushTask<std::pair<SString, SString>>(
+                [size]
+                {
+                    // Execute time-consuming task
+                    return SharedUtil::GenerateRsaKeyPair(size);
+                },
+                [luaFunctionRef = callback.value()](std::pair<SString, SString> result)
```
tederis: Is it necessary to pass a result by value?

tederis: Changing it to `const std::pair<SString, SString>& result` will cause more unnecessary code?

Inder00:
To the moon 🚀

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,92 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+                            try
+                            {
+                                result = SharedUtil::GenerateRsaKeyPair(size);
+                            }
+                            catch (const CryptoPP::Exception& ex)
+                            {
+                            }
+                            return result;
+                        },
+                        [luaFunctionRef = callback.value()](const std::pair<SString, SString> result)
```
tederis: It should be `const std::pair<SString, SString>& result`

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -362,6 +446,73 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+                                try
+                                {
+                                    return std::make_pair(SharedUtil::RsaEncode(data, key), true);
+                                }
+                                catch (const CryptoPP::Exception& ex)
+                                {
+                                    return std::make_pair(SString(ex.GetWhat()), false);
+                                }
+                            },
+                            [luaFunctionRef](const std::pair<SString, bool> result)
```
tederis: It should be `const std::pair<SString, bool>& result`

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -516,6 +667,73 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
...
+                                try
+                                {
+                                    return std::make_pair(SharedUtil::RsaDecode(data, key), true);
+                                }
+                                catch (const CryptoPP::Exception& ex)
+                                {
+                                    return std::make_pair(SString(ex.GetWhat()), false);
+                                }
+                            },
+                            [luaFunctionRef](const std::pair<SString, bool> result)
```
tederis: It should be `const std::pair<SString, bool>& result`

tederis (changes requested):

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -230,9 +231,92 @@ int CLuaCryptDefs::PasswordVerify(lua_State* luaVM)
...
+                            // Execute time-consuming task
+                            std::pair<SString, SString> result;
+                            try
+                            {
+                                result = SharedUtil::GenerateRsaKeyPair(size);
+                            }
+                            catch (const CryptoPP::Exception& ex)
+                            {
+                            }
+                            return result;
```
Pirulax: No exception handling. 
Could you please do the same thing as before?

Inder00:
To the moon 🚀 (review)

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -516,6 +667,73 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
...
+                            [data, key]
+                            {
+                                try
+                                {
+                                    return std::make_pair(SharedUtil::RsaDecode(data, key), true);
+                                }
+                                catch (const CryptoPP::Exception& ex)
+                                {
+                                    return std::make_pair(SString(ex.GetWhat()), false);
+                                }
```
AlexTMjugador, line 697: This is better exception handling than just ignoring it and returning false. However, I'm not sure if we want to expose scripts to the implementation-defined messages of CryptoPP. According to the [docs](https://www.cryptopp.com/docs/ref/class_exception.html#ac2c2e8a3c1871c532660252cd1894b90), it should be feasible to get an error type and control what string is logged ourselves.

AlexTMjugador:
About the removed tests, I think @sbx320 was not pointing that they were unnecessary and should be removed, but that they were not that useful either because they just tested the entire encryption-decryption cycle, and as such a trivial identity function implementation would also satisfy those tests.

I think that tests are useful for this PR and belong to it, as long as they test encryption and decryption separately: at least one test should test that encryption of a known plaintext with a known key-pair and cipher parameters yields the expected ciphertext, and at least other test should check that decryption of another ciphertext is able to yield back the plaintext.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
Inder00: That was @Pirulax idea to expose exception cause

AlexTMjugador: I agree with exposing the exception cause somehow. But I'm not sure if we want to potentially make backwards compatibility more complicated just because some script decided to rely on some precise warning text. This is a pretty minor and far-fetched concern, though, and I'd be happy with this merged as-is.

Inder00:
> About the removed tests, I think @sbx320 was not pointing that they were unnecessary and should be removed, but that they were not that useful either because they just tested the entire encryption-decryption cycle, and as such a trivial identity function implementation would also satisfy those tests.
> 
> I think that tests are useful for this PR and belong to it, as long as they test encryption and decryption separately: at least one test should test that encryption of a known plaintext with a known key-pair and cipher parameters yields the expected ciphertext, and at least other test should check that decryption of another ciphertext is able to yield back the plaintext.

Test has been applied:
- `Encryption`: encrypts data using random keypair (encrypt, decrypt and then compare) - encrypt cannot be compared as static data because used RSA encryption scheme `OAEP using SHA` outputs different encrypted data each time - have to be decrypted to compare.
- `Decryption`: decrypts encrypted data in string using private key stored in another string and compare decypted data with expected data.

AlexTMjugador (approved):
Design-wise this looks fine to me, but I have not tested it, and other people might have some stylistic feedback.

Inder00:
> Design-wise this looks fine to me, but I have not tested it, and other people might have some stylistic feedback.

Thanks for review, a test resource is included in pr 🚀

Inder00:
Review someone 🔥

TheNormalnij:
Code looks well now.  I think it should be refactored into "strategy" pattern with next added algorithm.

sbx320:
@TheNormalnij not a bug.

Github keeps it as "Changes Requested" regardless of unresolved threads, as there may be changes requested in a top level comment.

PR #2417, opened 2021-10-19
Server modules

CrosRoad95:
Because of modules in mta are almost dead feature ( except ml_socket ), i spent a few days of improving it and adding some new small features.
I changed architecture of mta a little, new folder called "server/modules" contains all extra features, for example lua modules as they are called in code base. This approach let offload main project ( deathmatch ) by removing few headers but added few to communicate between main project and modules.
image
No more waiting few minutes to compile, LuaModules project compiles in seconds what improve productivity.

Inside that project you can see first "module" called "LuaModules" - This project has only one job: manage lua modules, nothing else while main project don't know what are lua modules.

New lua manager, 2.0 has been added right now with only two functions to get `IResource` - a new interface class used to expose functionality to different projects, lua modules.

New command "modules" will tell you current status of modules and list of availiable ones:
image

The main goal is to make module be able to do everything. Right now they let you add custom functions and events, nothing more. 

This pr is mainly a refactor and preparing a base for further updates.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.h
+++ b/Server/mods/deathmatch/logic/CGame.h
@@ -670,5 +670,6 @@ private:
     bool m_DevelopmentModeEnabled;
     bool m_showClientTransferBox = true;
 
+    CModuleInterfaceImpl* m_pModuleInterface;
     std::vector<IModule*> m_vecModules;
 };
--- a/Server/mods/deathmatch/logic/CGame.h
+++ b/Server/mods/deathmatch/logic/CGame.h
@@ -60,6 +60,7 @@ class CGame;
 #include "CBanManager.h"
 
 // Forward declarations
+class CModule;
 class ASE;
 class CAccessControlListManager;
 class CAccountManager;
@@ -462,6 +463,19 @@ public:
     bool IsClientTransferBoxVisible() const { return m_showClientTransferBox; }
     void SetClientTransferBoxVisible(bool visible) { m_showClientTransferBox = visible; }
 
+    template <typename T>
+    T* GetModule()
+    {
+        for (auto const& pModule : m_vecModules)
+        {
+            if (T* p = dynamic_cast<T*>(pModule))
+            {
+                return p;
+            }
+        }
+        return nullptr;
+    }
+
 private:
     void AddBuiltInEvents();
     void RelayPlayerPuresync(class CPacket& Packet);
@@ -655,4 +669,6 @@ private:
 
     bool m_DevelopmentModeEnabled;
     bool m_showClientTransferBox = true;
+
+    std::vector<IModule*> m_vecModules;
 };
--- a/Server/modules/luaModules/include/CLuaModulesModule.h
+++ b/Server/modules/luaModules/include/CLuaModulesModule.h
@@ -5,18 +5,23 @@ class CLuaModuleManager;
 class CLuaModulesModule : public IModule
 {
 public:
-    void                     Load(IModuleInterface* pModuleInterface);
-    void                     UnLoad();
-    void                     DoPulse();
-    void                     ResourceStarted(IResource* pResource);
-    void                     ResourceStopped(IResource* pResource);
-    void                     ResourceStopping(IResource* pResource);
-    int                      LoadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
-    int                      UnloadModule(const char* szShortFileName);
-    int                      ReloadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
-    ILuaModule*              GetModuleByName(const char* szName);
-    std::vector<ILuaModule*> GetModules();
+    CLuaModulesModule(IModuleInterface* pModuleInterface) : m_pModuleInterface(pModuleInterface) {}
+
+    void                      Load();
+    void                      UnLoad();
+    void                      DoPulse();
+    void                      ResourceStarted(IResource* pResource);
+    void                      ResourceStopped(IResource* pResource);
+    void                      ResourceStopping(IResource* pResource);
+    int                       LoadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    int                       UnloadModule(const char* szShortFileName);
+    int                       ReloadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    ILuaModule*               GetModuleByName(const char* szName);
+    std::vector<ILuaModule*>  GetModules();
+    std::vector<SFoundModule> GetAllModules();
+    bool                      IsModuleRunning(std::string moduleName);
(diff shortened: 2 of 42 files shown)


Server/mods/deathmatch/logic/CGame.h
```diff
@@ -655,4 +669,7 @@ class CGame
 
     bool m_DevelopmentModeEnabled;
     bool m_showClientTransferBox = true;
+
+    CModuleInterfaceImpl* m_pModuleInterface;
+    std::vector<IModule*> m_vecModules;
```
tederis, line 674: Isn't it better to use std::unique_ptr<IModule> here instead of a raw ptr?

tederis: I mean `std::vector<std::unique_ptr<IModule>> m_vecModules;`

Server/modules/luaModules/include/CLuaModulesModule.h
```diff
@@ -0,0 +1,27 @@
...
+    void                      ResourceStarted(IResource* pResource);
+    void                      ResourceStopped(IResource* pResource);
+    void                      ResourceStopping(IResource* pResource);
+    int                       LoadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    int                       UnloadModule(const char* szShortFileName);
+    int                       ReloadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    ILuaModule*               GetModuleByName(const char* szName);
+    std::vector<ILuaModule*>  GetModules();
+    std::vector<SFoundModule> GetAllModules();
+    bool                      IsModuleRunning(std::string moduleName);
```
tederis, line 22: It's better to pass a string by reference here. It should be `bool IsModuleRunning(const std::string& moduleName) const;`

Server/modules/luaModules/src/CLuaModuleManager.cpp
```diff
@@ -111,34 +101,33 @@ int CLuaModuleManager::ReloadModule(const char* szShortFileName, const char* szF
...
 int CLuaModuleManager::UnloadModule(const char* szShortFileName)
-{            // 0 = Success, 9 = Can't find module by name
-    list<CLuaModule*>::iterator iter = m_Modules.begin();
-    for (; iter != m_Modules.end(); ++iter)
+{
+    // 0 = Success, 9 = Can't find module by name
+    std::vector<CLuaModule*>::iterator iter = m_vecModules.begin();
+    for (; iter != m_vecModules.end(); ++iter)
     {
         if (strcmp((*iter)->_GetName().c_str(), szShortFileName) == 0)
```
tederis, line 109: Why not just `if ((*iter)->_GetName() == szShortFileName)`?

Server/modules/luaModules/src/CLuaModulesModule.cpp
```diff
@@ -0,0 +1,104 @@
...
+        foundModule.name = file;
+        if (IsModuleRunning(file))
+        {
+            foundModule.state = "running";
+        }
+        else
+        {
+            foundModule.state = "loaded";
+        }
+        foundModules.push_back(foundModule);
```
tederis, line 101: How about if we use a move semantics here? `foundModules.push_back(std::move(foundModule));`

Server/modules/luaModules/src/CLuaModulesModule.cpp
```diff
@@ -0,0 +1,104 @@
...
+            return luaModule;
+        }
+    }
+    return nullptr;
+}
+
+std::vector<ILuaModule*> CLuaModulesModule::GetModules()
+{
+    std::vector<CLuaModule*> luaModules = m_pLuaModuleManager->GetLoadedModules();
+    std::vector<ILuaModule*> modules(luaModules.size());
```
tederis, line 64: You can use std::copy instead of a manual upcasting

Server/mods/deathmatch/logic/CGame.h
Pirulax: Or just use `std::list`

Server/modules/luaModules/include/CLuaModulesModule.h
Pirulax: might as well take a `string_view` instead.

Server/modules/luaModules/src/CLuaModulesModule.cpp
Pirulax: Could just:
```cpp
    const auto& modules = m_pLuaModuleManager->GetLoadedModules();
    return {modules.begin(), modules.end()};
```
 instead of this mess.

Server/modules/luaModules/include/CLuaModulesModule.h
```diff
@@ -0,0 +1,27 @@
...
+    void                      UnLoad();
+    void                      DoPulse();
+    void                      ResourceStarted(IResource* pResource);
+    void                      ResourceStopped(IResource* pResource);
+    void                      ResourceStopping(IResource* pResource);
+    int                       LoadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    int                       UnloadModule(const char* szShortFileName);
+    int                       ReloadModule(const char* szShortFileName, const char* szFileName, bool bLateLoad);
+    ILuaModule*               GetModuleByName(const char* szName);
+    std::vector<ILuaModule*>  GetModules();
```
Pirulax, line 20: Why are you returning a copy here?
I'd suggest returning a `const&`

Server/modules/luaModules/src/CLuaModulesModule.cpp
Pirulax: You really should be using Enums for the state, instead of strings.
Also, move semantics would be nice:
```cpp

// Using this instead of `SFoundModule`. Could also just use an std::pair to be honest...
struct ModuleDescriptor {
    std::string path;
    enum class State {
        LOADED,
        RUNNING,
    } state;
}


std::vector<SString> itemList = SharedUtil::FindFiles(SharedUtil::PathJoin(m_pModuleInterface->GetModulesPath(), "*"), true, false);

std::vector<SFoundModule> foundModules;
foundModules.reserve(itemList.size());

for (auto& file : itemList)
{
    foundModules.emplace_back(
        std::move(file),
        IsModuleRunning(file) ? ModuleDescriptor::State::RUNNING : ModuleDescriptor::State::LOADED
    );
}
```

Tbh, the naming of the function is kinda off... You have `GetModules`, which also returns all modules, from what I see.
Could you please add comments to the code explaining what each function does?
If necessary, please rename the function as well. To me it seems like its more like `FindAllModules` or something.

ERAGON007:
Maybe this PR can be added to `1.9` release milestone?

PR #2418, opened 2021-10-19
Add option to set allowed events from cef

CrosRoad95:
Add option to specify what events should be allowed to call from CEF.
You can not change it after browser get created. This change is to improve security, now one misused "innerHTML" can compromise the whole server and you can do nothing.
If table is empty, every event is allowed, this is by default to provide backward compatiblity.

example code, "hackme" will not triggered:

```lua
local screenWidth, screenHeight = 800, 600

local window = guiCreateWindow( 600, 100, screenWidth, screenHeight, "Web Browser", false )
local browser = guiCreateBrowser( 0, 28, screenWidth, screenHeight, true, false, false, window, { "foo", "bar" } )
local theBrowser = guiGetBrowser( browser )

setDevelopmentMode(true, true)

addEvent("foo", true)
addEventHandler("foo", root, function(message)
    outputChatBox(message)
end)

addEvent("bar", true)
addEventHandler("bar", root, function(message)
    outputChatBox(message)
end)

addEvent("hackme", true)
addEventHandler("hackme", root, function(message)
    outputChatBox(message)
end)

addEventHandler("onClientBrowserCreated", browser,
    function ()
        loadBrowserURL(source, "http://mta/local/example.html")
    end)

addEventHandler( "onClientBrowserDocumentReady", theBrowser, 
	function( )
        executeBrowserJavascript(source, [[
mta.triggerEvent("foo", "foo triggered")
mta.triggerEvent("hackme", "you has been backed")
mta.triggerEvent("foo", "foo triggered again")
mta.triggerEvent("bar", "bar triggered")
]])
	end
)

```

### DIFF
--- a/Client/mods/deathmatch/logic/CClientWebBrowser.cpp
+++ b/Client/mods/deathmatch/logic/CClientWebBrowser.cpp
@@ -227,6 +227,10 @@ void CClientWebBrowser::Events_OnChangeCursor(unsigned char ucCursor)
 
 void CClientWebBrowser::Events_OnTriggerEvent(const SString& strEventName, const std::vector<std::string>& arguments)
 {
+    if (m_allowedEvents.size() > 0)
+        if (m_allowedEvents.find(strEventName) == m_allowedEvents.end())
+            return;
+
     CLuaArguments Arguments;
     for (std::vector<std::string>::const_iterator iter = arguments.begin(); iter != arguments.end(); ++iter)
     {
@@ -327,12 +331,18 @@ bool CClientWebBrowser::ToggleDevTools(bool visible)
     return m_pWebView->ToggleDevTools(visible);
 }
 
+void CClientWebBrowser::SetAllowedEvents(std::vector<SString> vecAllowedEvents)
+{
+    m_allowedEvents = std::set(vecAllowedEvents.begin(), vecAllowedEvents.end());
+}
+
 CClientGUIWebBrowser::CClientGUIWebBrowser(bool isLocal, bool isTransparent, uint width, uint height, CClientManager* pManager, CLuaMain* pLuaMain,
-                                           CGUIElement* pCGUIElement, ElementID ID)
+                                           CGUIElement* pCGUIElement, std::vector<SString> vecAllowedEvents, ElementID ID)
     : CClientGUIElement(pManager, pLuaMain, pCGUIElement, ID)
 {
     m_pManager = pManager;
     m_pBrowser = g_pClientGame->GetManager()->GetRenderElementManager()->CreateWebBrowser(width, height, isLocal, isTransparent);
+    m_pBrowser->SetAllowedEvents(vecAllowedEvents);
 
     if (m_pBrowser)
     {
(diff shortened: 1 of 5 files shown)


Client/mods/deathmatch/logic/CClientWebBrowser.cpp
```diff
@@ -327,12 +331,18 @@ bool CClientWebBrowser::ToggleDevTools(bool visible)
     return m_pWebView->ToggleDevTools(visible);
 }
 
+void CClientWebBrowser::SetAllowedEvents(std::vector<SString> vecAllowedEvents)
```
tederis, line 334: It's better to pass a vector by a reference (const std::vector<SString>& vecAllowedEvents)

Client/mods/deathmatch/logic/CClientWebBrowser.cpp
```diff
@@ -327,12 +331,18 @@ bool CClientWebBrowser::ToggleDevTools(bool visible)
...
+{
+    m_allowedEvents = std::set(vecAllowedEvents.begin(), vecAllowedEvents.end());
+}
+
 CClientGUIWebBrowser::CClientGUIWebBrowser(bool isLocal, bool isTransparent, uint width, uint height, CClientManager* pManager, CLuaMain* pLuaMain,
-                                           CGUIElement* pCGUIElement, ElementID ID)
+                                           CGUIElement* pCGUIElement, std::vector<SString> vecAllowedEvents, ElementID ID)
```
tederis, line 340: The same, pass by a reference

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -441,7 +441,7 @@ class CStaticFunctionDefinitions
     static CClientGUIElement* GUICreateComboBox(CLuaMain& LuaMain, const CVector2D& position, const CVector2D& size, const char* szCaption, bool bRelative,
                                                 CClientGUIElement* pParent);
     static CClientGUIElement* GUICreateBrowser(CLuaMain& LuaMain, const CVector2D& position, const CVector2D& size, bool bIsLocal, bool bIsTransparent,
-                                               bool bRelative, CClientGUIElement* pParent);
+                                               bool bRelative, CClientGUIElement* pParent, std::vector<SString> vecAllowedEvents);
```
tederis, line 444: Similarly, pass by a reference

Client/mods/deathmatch/logic/CClientWebBrowser.cpp
```diff
@@ -227,6 +227,10 @@ void CClientWebBrowser::Events_OnChangeCursor(unsigned char ucCursor)
 
 void CClientWebBrowser::Events_OnTriggerEvent(const SString& strEventName, const std::vector<std::string>& arguments)
 {
+    if (m_allowedEvents.size() > 0)
+        if (m_allowedEvents.find(strEventName) == m_allowedEvents.end())
```
tederis, line 231: I think you can combine two conditionals together

Pirulax:
Passing by value isn't an issue if its moved. 
So yeah, either pass by value, or move it.
I'd say moving it may be a better option.

Client/mods/deathmatch/logic/CClientWebBrowser.cpp
Pirulax: no need to combine, just skip the size check, no need.

Pirulax (commented):
Is this PR WIP?
Because I dont see the part where the event trigger check is actually done.
Also, I'd argue creating a `set` directly, instead of a `vector`, then a `set` is more efficient.
Uh, also, what if I want to `allow`/`disallow` events after creating the browser? Not sure what the use case would be, but I'd expect there to be a function(s) to do that as well.

PR #2421, opened 2021-10-25
Fix door and damage sync of vehicles added with engineRequestModel()

### LINKED ISSUE #2034: Add-on vehicle behaving weird

BCG2000:
This fixes the issue where vehicles added with engineRequestModel() display several bugs when someone else drives them, such as the doors constantly switching between open and closed, as well as the tires always appearing popped (https://github.com/multitheftauto/mtasa-blue/issues/2277#issuecomment-882222297).

It does this by sending the parent model ID instead of the allocated clientside model ID during a vehicle pure sync (to match the one on the server, which is always the parent model).

Test resource is the same one from https://github.com/multitheftauto/mtasa-blue/pull/2247.
Example of fix working (click to open YouTube video):
[Alt text](https://www.youtube.com/watch?v=AfAwN5BzzAw)

Also, https://github.com/multitheftauto/mtasa-blue/pull/2399 has to be merged as well to fix any remaining sync inconsistency.

EDIT: I forgot to add the description in the commit message (so what it is doing is clearer). When squashing I ask it to be this if possible:
Make CNetAPI::WriteVehiclePuresync() use parent model ID when syncing vehicles added with engineRequestModel().

Also this can close https://github.com/multitheftauto/mtasa-blue/issues/2034, but https://github.com/multitheftauto/mtasa-blue/issues/2277 can possibly remain open so people can keep listing bugs (not sure, up to you).


lopezloo (approved):
Looks good to me.

PR #2422, opened 2021-10-27
Add grain effect functions

tederis:
`bool setGrainMultiplier(string type="master,infrared,night,rain,overlay,all", float multiplier)`
This function allows you to modulate the intensity of different applications of the grain effect. For example, you can disable a grain effect for a nightvision camera or make it less strong. All multipliers work independently but you can use the master multiplier to modulate the all values.

`bool setGrainLevel(int level)`
This function allows you to enable an **overlay** grain effect and set the level of that effect. The game will draw it on top of another grain effects.


tederis:
I'm not sure about the function names. So proposals are welcome.

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -578,6 +578,76 @@ CMultiplayerSA::CMultiplayerSA()
     m_dwLastStaticAnimID = eAnimID::ANIM_ID_WALK;
 }
 
+namespace GrainEffect
+{
+
+static BYTE ucGrainEnabled = FALSE;
+static DWORD dwGrainStrength = 1.0f;
```
sbx320: Should be `bool` and `float`. BYTE and DWORD are really just legacy stuff

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -578,6 +578,76 @@ CMultiplayerSA::CMultiplayerSA()
...
+float RainModifier::fMultiplier = 1.0f;
+float OverlayModifier::fMultiplier = 1.0f;
+
+}
+
+template <typename T>
+int _cdecl HOOK_GrainEffect(char a1, char a2)
+{
+    return T::ApplyEffect(a1, a2);
+}
```
sbx320: Is there some special reason for this template? It seems a bit pointless, since you should be able to just do
```
HookInstallCall(0x123456, (DWORD)GrainEffect::OverlayModifier::ApplyEffect);
```
instead

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -578,6 +578,76 @@ CMultiplayerSA::CMultiplayerSA()
     m_dwLastStaticAnimID = eAnimID::ANIM_ID_WALK;
 }
 
+namespace GrainEffect
```
sbx320: I'd suggest also moving this into a separate file. CMultiplayerSA.cpp is long enough as-is

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -755,6 +825,11 @@ void CMultiplayerSA::InitHooks()
     HookInstall(HOOKPOS_CAnimManager_AddAnimationAndSync, (DWORD)HOOK_CAnimManager_AddAnimationAndSync, 10);
     HookInstall(HOOKPOS_CAnimManager_BlendAnimation_Hierarchy, (DWORD)HOOK_CAnimManager_BlendAnimation_Hierarchy, 5);
 
+    HookInstallCall(0x704EE8, (DWORD)HOOK_GrainEffect<GrainEffect::NightModifier>);
```
sbx320: The memory adresses here (and elsewhere) should really be constants somewhere. (similar to the HOOKPOS constants, although I'd put them into the GrainEffect namespace), to make it a bit clearer what we're actually hooking here

sbx320 (dismissed):

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -578,6 +578,76 @@ CMultiplayerSA::CMultiplayerSA()
     m_dwLastStaticAnimID = eAnimID::ANIM_ID_WALK;
 }
 
+namespace GrainEffect
+{
+
+static BYTE ucGrainEnabled = FALSE;
+static DWORD dwGrainStrength = 1.0f;
+
+struct MasterModifier
```
Pirulax: Dont mix float and integer multiply without casting first, because it causes warnings.
Please add the necessary casts.

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -578,6 +578,76 @@ CMultiplayerSA::CMultiplayerSA()
...
+float MasterModifier::fMultiplier = 1.0f;
+float InfraredModifier::fMultiplier = 1.0f;
+float NightModifier::fMultiplier = 1.0f;
+float RainModifier::fMultiplier = 1.0f;
+float OverlayModifier::fMultiplier = 1.0f;
+
+}
+
+template <typename T>
+int _cdecl HOOK_GrainEffect(char a1, char a2)
```
Pirulax: just use `static`, its `__cdecl` by default.
also, perhaps naming these variables correctly would be nice.

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1669,6 +1744,61 @@ void CMultiplayerSA::SetColorFilter(DWORD dwPass0Color, DWORD dwPass1Color)
     }
 }
 
+void CMultiplayerSA::SetGrainMultiplier(eGrainMultiplierType type, float fMultiplier)
```
Pirulax: Now that I think about it, if you really wanted to complicate things, you could make a class called `Modifier`, with a template argument `eGrainMultiplierType`.
For ex, nigh vision would become: `Modifier<eGrainMultiplierType::NIGHT>`.

Just a thought though.

Pirulax (dismissed):
lgtm

Client/multiplayer_sa/CMultiplayerSA.cpp
tederis: bool and BYTE is not the same. A compiler does not guarantee equality sizeof(bool) and sizeof(BYTE). This is not an equitable type conversion. The more correct would be unsigned char instead of BYTE. The floating point value is a misprint, it should be unsigned int actually.

Client/multiplayer_sa/CMultiplayerSA.cpp
tederis: The floating point value is a misprint, it should be unsigned int actually.

tederis:
Issues finally got resolved.

botder (approved):

botder:
@tederis, please add the functions to the wiki.

PR #2432, opened 2021-11-09
Add missing vehicle colors to vehiclecolors.conf

Lord-Henry:
The vehicle **"Street Clean Trailer (611)"** should use the same colors from its truck **"Utility Van (552)"** instead of being always fully black.


TheNormalnij:
`Server/mods/deathmatch/logic/CVehicleDefaultColors.cpp` also contains this data

ghost:
> `Server/mods/deathmatch/logic/CVehicleDefaultColors.cpp` also contains this data

True, it remains for the model 611 if I were to edit and by, I would put it as follows:

I would put the model 611 under 609, since it would be the next most model number

Code line: https://github.com/multitheftauto/mtasa-blue/blob/a59308d0dea5199d23ca22c522d68f93dc3fd610/Server/mods/deathmatch/logic/CVehicleDefaultColors.cpp#L694

```cpp
    "\n"
    "611 56 56\n"
    "611 49 49\n"
    "611 26 124\n"
    "\n"
```

But maybe he didn't put it on because he could make some conflict, I don't know very well.

lopezloo:
Please also add these colors to `Server/mods/deathmatch/logic/CVehicleDefaultColors.cpp`.

lopezloo (dismissed):
Please also add these colors to `Server/mods/deathmatch/logic/CVehicleDefaultColors.cpp`.

patrikjuvonen:
Where is this new color defined in GTA:SA (carcols.dat doesn't have it)
I cannot find where it should be gray. The decal text on the side is white, which would mean the color should be dark/black.
Unless GTA has it gray, we shouldn't change the default behavior.

Secondly, you also need to address lopezloo's review above.

lopezloo:
> Where is this new color defined in GTA:SA (carcols.dat doesn't have it)
> I cannot find where it should be gray. The decal text on the side is white, which would mean the color should be dark/black.
> Unless GTA has it gray, we shouldn't change the default behavior.

This vehicle is unused in single player so that model probably isn't finished. It's black if you spawn it single player as colors aren't defined.

It has longitudinal rectangle which color is controllable via color 2 setting which matches rectangle found on "Utility Van (552)" vehicle so I think it was supposed to be gray and this text color is actually wrong.

mta-screen_2022-12-25_17-14-08

Lord-Henry:
@lopezloo 
The requested changes are done. What about this pull request?

lopezloo:
What is the status of this PR? Tested it and this vehicle still spawns with black color by default.

```lua
run veh = createVehicle(611, 0, 0, 4)
run getVehicleColor(veh)
```

```
Command results: 0 [number], 0 [number], 0 [number], 0 [number]
```

Lord-Henry:
> What is the status of this PR? Tested it and this vehicle still spawns with black color by default.
> 

Open. It was not merged yet.

lopezloo:
> > What is the status of this PR? Tested it and this vehicle still spawns with black color by default.
> 
> Open. It was not merged yet.

Have compiled it and it doesn't work. Spawns with black color.

PR #2435, opened 2021-11-11
Bugfix (onColShape/onElementColShape)

Inder00:
Event needs a was-element-destroyed check.

PoC:
```lua
local vehicleModel = 404
local vehiclePos = Vector3(0, 0, 5)
createVehicle(vehicleModel, vehiclePos)
createColSphere(vehiclePos, 10)
addEventHandler("onColShapeLeave", resourceRoot, function( el, md )
    if not getElementType(el) then outputChatBox("invalid element type. gotcha?", root) return end -- returns a warning
    if getElementType(el) ~= "player" then return end
    if not isPedInVehicle(el) then outputChatBox("you should be in vehicle to reproduce that", root) end
    destroyElement( getPedOccupiedVehicle( el ) )
end)
```

### DIFF
--- a/Client/mods/deathmatch/logic/CClientColManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientColManager.cpp
@@ -146,7 +146,7 @@ void CClientColManager::HandleHitDetectionResult(bool bHit, CClientColShape* pSh
             pEntity->AddCollision(pShape);
 
             // Can we call the event?
-            if (pShape->GetAutoCallEvent())
+            if (pShape->GetAutoCallEvent() && !pEntity->IsBeingDeleted())
             {
                 // Call the event
                 CLuaArguments Arguments;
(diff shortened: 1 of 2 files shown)


TheNormalnij:
I tested this PR a week ago. It works good at my server. MTA client hasn't regressions for me.
But i think this check `!pEntity->IsBeingDeleted()` should called before `HandleHitDetectionResult` in `DoHitDetectionForEntity` or `DoHitDetectionForColShape`

Inder00:
> I tested this PR a week ago. It works good at my server. MTA client hasn't regressions for me. But i think this check `!pEntity->IsBeingDeleted()` should called before `HandleHitDetectionResult` in `DoHitDetectionForEntity` or `DoHitDetectionForColShape`

Anyway it needs `pShape->RemoveCollider(pEntity);` to be executed which is executed in code block above entity has been deleted check.

TheNormalnij (approved):

Pirulax:
should be good, dont think it can possibly break any code.
we'll see.

PR #2449, opened 2021-12-04
Variable set relative to false when it has invisible space

nsxh:
When I use 'gui' functions I don't set the 'relative' part to false, because most of us MTA programmers don't use 'relative' as true then, so it's not having to set variable 'relative' to false all the time.


TheNormalnij:
We can't break backward compatibility with old scripts.

r12750:
> We can't break backward compatibility with old scripts.

If it is changed do not anger the errors, why most MTA programmers are already set the variable 'relative' to 'false'

That's what I understood even at the time of your comment

TheNormalnij:
My bad. I thought that we use `relative = true` by default.
In this case it has some sense

patrikjuvonen (approved):

PR #2451, opened 2021-12-11
Fix CEF crashing (#2446)

### LINKED ISSUE #2446: CEF causing random crash, without a crash-dump
Somewhere around Septeber-October there is a weird crash start appearing. Game just close without a crash report when users were using CEF  UI on my server. Occurance is pretty big ~5% of weekly sessions in CEF UI are not exiting properly (most likely crash).

Crash details from Windows Event Viewer:
```
Faulting application name: gta_sa.exe, version: 0.0.0.0, time stamp: 0x437101ca
Faulting module name: cefweb.dll, version: 1.5.9.21024, time stamp: 0x617e4432
Exception code: 0xc0000409
Fault offset: 0x00092bd5
Faulting process id: 0x682c
Faulting application start time: 0x01d7e4b6b2d3b457
Faulting application path: C:\Games\NS-DayZ\GTA San Andreas\gta_sa.exe
Faulting module path: C:\Games\NS-DayZ\Multi Theft Auto1.5\mta\cefweb.dll
Report Id: 8d30e17f-05b9-4705-a962-09416c1b9bac
Faulting package full name: 
Faulting package-relative application ID: 
```

I'm attaching resource, with pretty blank implementation of webui. which is causing this crash for me.
MTA Ver: v1.5.9-release-21024.1

[webtest.zip](https://github.com/multitheftauto/mtasa-blue/files/7614737/webtest.zip)

Press F2 to open CEF window.
Resource using ajax handlers, to allow JS part of code grab data from LUA using 'fetch() method'.
JS side will trigger refresh of content, based on timeout or when window maximized. 
The smaller timeout - the faster crash will occur.
Additional F2 flood may also speed up a crash. But if you just live it for long time with low timeout, it will anyway crash at some point.

Lpsd:
Should resolve #2446 - due to:

```
Implicit conversion of CefRefPtr<T> or scoped_refptr<T> to T* is gone; use x.get() instead
```
(see https://bitbucket.org/chromiumembedded/cef/issues/3140/update-include-base-for-c-11-14-features)

Haven't tested these changes yet, upgrading to VS 2022 now - will report shortly 😅

### DIFF
--- a/Client/ceflauncher_DLL/CCefApp.h
+++ b/Client/ceflauncher_DLL/CCefApp.h
@@ -116,7 +116,16 @@ public:
             return;
 
         CefRefPtr<CefProcessMessage> message = V8Helpers::SerialiseV8Arguments("TriggerLuaEvent", arguments);
-        frame.get()->GetBrowser()->GetMainFrame()->SendProcessMessage(PID_BROWSER, message);
+
+        auto browser = frame.get()->GetBrowser();
+
+        if (!browser)
+            return;
+
+        auto mainFrame = browser->GetMainFrame();
+
+        if (mainFrame)
+            mainFrame.get()->SendProcessMessage(PID_BROWSER, message);
     }
 
 public:
--- a/Client/ceflauncher_DLL/CCefApp.h
+++ b/Client/ceflauncher_DLL/CCefApp.h
@@ -22,6 +22,8 @@ public:
     // http://magpcss.org/ceforum/apidocs3/projects/(default)/CefRenderProcessHandler.html#OnFocusedNodeChanged(CefRefPtr%3CCefBrowser%3E,CefRefPtr%3CCefFrame%3E,CefRefPtr%3CCefDOMNode%3E)
     virtual void OnFocusedNodeChanged(CefRefPtr<CefBrowser> browser, CefRefPtr<CefFrame> frame, CefRefPtr<CefDOMNode> node) override
     {
+        CefBrowser* pBrowser = browser.get();
+
         if (m_bHasInputFocus)
         {
             if (node)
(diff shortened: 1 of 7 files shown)


Lpsd:
Unfortunately I'm still crashing with the test resource from #2446 - but I think we're on the right path here, probably missing this same fix (`.get()`) elsewhere. I'll have another look shortly - need to see where stacktrace differs with these changes

Lpsd:
^ scratch that, I think my latest commit fixed it - an oversight in my changes. 

Haven't crashed yet, been about 5 minutes with the test resource running. Will leave MTA open for a bit and see what happens.

Edit: no crash after 20 minutes, seems good

Edit 2: some more random crashing, investigating

Lpsd:
alright, there's quite a few more changes needed than I originally thought (see latest commit) - nothing too major though.

again, seems fine for now, haven't crashed yet... needs testing by others

haron4igg:
> ^ scratch that, I think my latest commit fixed it - an oversight in my changes.
> 
> Haven't crashed yet, been about 5 minutes with the test resource running. Will leave MTA open for a bit and see what happens.
> 
> Edit: no crash after 20 minutes, seems good
> 
> Edit 2: some more random crashing, investigating

you can reduce timer delay in index.html aggresively, like 5-10 ms, it will cause crash almost instantly, if there is an issue, but it does no problems for 1.5.8

Lpsd:
@haron4igg with my latest changes, I can set the timeout to 1ms in `index.html`, and I haven't had a crash yet.

haron4igg:
> @haron4igg with my latest changes, I can set the timeout to 1ms in `index.html`, and I haven't had a crash yet.

yeah, i see now there is a lot of places this CEF change affected... have they even thought about consumers introdusing this pointer change... :) 

i'll try to test it locally now as well.

Lpsd:
@haron4igg since my last comment i actually crashed, lmao. it didn't crash instantly like you said it would, took a while, so I guess we're getting somewhere. probably just a few things left to catch and clean up.

Lpsd:
ok, seems everything i've done here is either wrong, or i'm missing something

at first i thought i was clearing up some slow corruption, as the crash time kept getting longer and longer with each "fix" i made - however I haven't been able to stop the crash completely (it still crashes after a long time).

weird thing is, I've just tried reverting everything to test how fast it crashes without these changes, and I cannot reproduce the crash at all now - so I have no clue what's going on.

PR #2452, opened 2021-12-12
Fix CEF crashing (#2446)

Lpsd:
Alright, hopefully the actual fix this time (#2451 was completely the wrong idea) 😅 

not sure why we're calling `m_callback->Continue()` here in the first place?

without these changes, I can reproduce a crash every single time from a **Release** build, setting the interval in `index.html` of the test resource in #2446 to 1ms (in **Debug** mode, the crash hardly ever happens, because of magic or whatever)

with these changes, I'm unable to reproduce the crash (@haron4igg also tested heavily, no crash)


haron4igg:
And, at least, my original CEF-UI with a lot of different HTML and JS features on board works as expected!

botder (approved):

PR #2457, opened 2021-12-16
Fix crash for non-file based CXMLNode (don't use IDs)

### LINKED ISSUE #2456: svgCreate causes random crashes
**Describe the bug**

**To reproduce**

/start svgtest

**Screenshots**

Przechwytywanie

**Version**

Lpsd:
Resolves #2456 - before xmlLoadString and vector graphic stuff, XML was exclusively being used when dealing with files, so this `m_bUsingIDs` stuff was never properly checked, and the underlying implementations have also changed since those conditions were made.

Also added delete on `m_pXMLDocument` in CClientVectorGraphic destructor.


patrikjuvonen (changes requested):
It seems it's still crashing but with a different error this time.

Untitled

PR #2459, opened 2021-12-19
Fix buffer overflow issues in `GetNameAndDamage`

Pirulax:
Potential overflows that were spotted during the investigation of issue #2262. It fixes issue #2262 as well, see [this comment](https://github.com/multitheftauto/mtasa-blue/issues/2262#issuecomment-999223341) for details)


botder (approved):
LGTM, but some comments are distracting

Dutchman101:
I hope issue #2479 can be fixed soon enough, to prevent having to revert this PR. We'll see, but we need to make nightlies stable again.

PR #2460, opened 2021-12-19
Fix #2458 (Always render HUD)

### LINKED ISSUE #2458: Changing GTA San Andreas Singleplayer HUD affects MTA
**Describe the bug**
If you disable part of your HUD in GTASA then it will affect your MTA and you will be unable to display the default HUD.

**To reproduce**
Open your GTA San Andreas (gtasa.exe), go to config and disable your hud.
Open MTA and join a server, you won't be able to see your hud.

Even using ``setPlayerHudComponentVisible`` will not work.

```lua
start runcode
crun isPlayerHudComponentVisible("health")
crun setPlayerHudComponentVisible("health", true)
crun isPlayerHudComponentVisible("health")
```

**Expected behaviour**
You should be able to see your HUD regardless of your GTASA config

**Screenshots**
image

**Version**
All versions (this is an old bug)

**Additional context**
MTA should have an option to disable your hud in the configuration. I understand that maybe MTA wants players to be able to choose whenever your HUD is visible or not, but after years of being a server owner I can't even remember how many times people had asked me why the couldn't see their HUD only to notice that they had disabled it on their GTASA.

Merlin:
Fixes #2458

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1520,6 +1520,9 @@ void CMultiplayerSA::InitHooks()
     for (auto uiAddr : shadowAddr)
         MemPut(uiAddr, &m_fShadowsOffset);
 
+    // Skip check for disabled HUD
+    MemSet((void*)0x58FBC4, 0x90, 9);
+
     InitHooks_CrashFixHacks();
 
     // Init our 1.3 hooks.


TheNormalnij:
Thanks you for this PR!
I see you've disabled this check.
image
Is there any less hackly way to make this value true?

Merlin:
Why? It forces the HUD always on and we have functions to disable individual parts of the hud and scripters expect it to be on by default. We could also nop the CTheScripts::bDisplayHud check but it will always be true anyways because we dont process any opcodes.

lopezloo:
As alternative, we could just set `FrontEndMenuManager.m_bHudOn` to true.
```
MemPutFast<DWORD>((0xBA6748 + 0x21), 1);
```

That would save in game settings though.

We are doing same thing with radar mode:
https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/game_sa/CSettingsSA.cpp#L58

Maybe we could disable saving radar & hud mode to GTA settings somehow.

Pirulax:
> Thanks you for this PR! I see you've disabled this check. image Is there any less hackly way to make this value true?

This seems to be perfectly okay way to do it.
Just gotta add some note somewhere, to let people know about it.

Pirulax:
Doing that would be better perhaps.

lopezloo (approved):
Going to merge this. We could set `FrontEndMenuManager.m_bHudOn` to true instead but that would save in GTA settings.

Thank you.

PR #2463, opened 2021-12-20
Add bUseIDs parameter to CXMLNodeImpl constructor & fix #2456 (svgCreate crashing)

### LINKED ISSUE #2456: svgCreate causes random crashes

Lpsd:
This PR adds a parameter to CXMLNodeImpl constructor (`bUseIDs`) which defaults to true. It's only effective for non file-based xml nodes. 

This new parameter is used within `CXMLImpl::ParseString` (set to false).

I also removed the debug assert from `CXMLNodeImpl::GetID`... why would we need a debug specific assert here, or rather why do we have an assert in the first place? We should always be able to use this method, even if the ID is `INVALID_XML_ID` - other code shouldn't be relying on an assert here, it should just check for `INVALID_XML_ID` and assert itself.

Resolves #2456

### DIFF
--- a/Shared/XML/CXMLImpl.cpp
+++ b/Shared/XML/CXMLImpl.cpp
@@ -53,7 +53,7 @@ std::unique_ptr<SXMLString> CXMLImpl::ParseString(const char* strXmlContent)
 
             if (xmlDocumentRoot)
             {
-                CXMLNodeImpl* xmlBaseNode = new CXMLNodeImpl(nullptr, nullptr, *xmlDocumentRoot);
+                CXMLNodeImpl* xmlBaseNode = new CXMLNodeImpl(nullptr, nullptr, *xmlDocumentRoot, false);
 
                 if (xmlBaseNode && xmlBaseNode->IsValid())
                 {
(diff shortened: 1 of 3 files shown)


botder (approved):

Lpsd:
don't merge yet, found some weird cases (our xml stuff is dodgy) - I'm not around for a couple hours though

Lpsd:
alright, I'm more than certain now that this PR is dumb and unnecessary - I was reading into the spaghetti rabbit hole too deep and neglected the fact this will render any non file-based XML representation on the Lua side invalid.

The real reason for #2456 is likely due to a race condition with XML construction (id) and the async task scheduler (used for SVG creation within `CLuaVectorGraphicDefs::SVGCreate` **only** if a callback function is provided)  - when two XML's are created at basically the same time, sometimes they can generate the same ID from the stack (not always) and therefore the assertion on `CXMLArray::PopUniqueID` occurs (specifically for `m_Elements[ulPhysicalIndex] == NULL`). 

I only figured that out late last night before my head hit the desk - I'll look into it properly later on. I'll close this PR and open another one for the "real" fix 😅

PR #2470, opened 2021-12-21
Update a string with better looking text

turret001:
better wording? removed a hyphen and space from the string so it looks nicer


Dutchman101:
I thought it was a no-brainer to review, but then i realized that the message string is used to identify for translations as you can see [here](https://github.com/multitheftauto/mtasa-blue/search?q=Do+you+want+to+see+some+on-line). So it would need to be modified through the translation framework. My bad, reverted in f386f28

turret001:
oh oops

EDIT: 
I've edited the translation strings' source file to have "Do you want to see some online help about this problem?" string on Crowdin instead of "Do you want to see some on-line help about this problem ?"

shouldn't we change something to make sure they both match?
I also found [this](https://github.com/multitheftauto/mtasa-blue/blob/27e3d7671d8165a1c61bd92a2fa8e473ed9b3404/Shared/sdk/SharedUtil.Misc.hpp#L696).

PR #2472, opened 2021-12-22
Revert "Add RSA support on encodeString (#2412)"

TheNormalnij:
This reverts commit e7e3ba5b337f791203ef977bd083a28226614da7.

Causes regression on client when compiled in debug mode. Didn't notice it this when i was on release.
@Inder00, recheck tests, please.


Dutchman101:
> Causes regression on client when compiled in debug mode

Can you describe it here?

Btw, my personal opinion is that reverting it is low priority and to be delayed as long as possible, in favour of the fixing alternative. Because first of all, it working on release builds is the priority (In the meanwhile, any developer that builds MTA and wants to try using encodeString with the new feature, can revert it locally after taking note) so the chance of a niche case where a developer will have a bug on custom builds, still provides you enough time and room to _fix_ the issue instead of choosing to revert.

I'm pulling on this a bit because we're trying to get the perfect stable build (a few changes and PR's are in line for it), and RSA support would be a nice addition for a widely circulated build soon.. keeping it in master for now serves to get to that point. So as long it really just affects Debug builds, this applies. And yeah, feel free to describe the regression so we can determine if that's expected to be the case.

TheNormalnij:
So we can just disable broken test

Dutchman101:
> So we can just disable broken test

Yes, that sounds better.. still, do you have any details? The tests aren't important for release MTA builds (after the feature got implemented already) right?

Dutchman101:
Replaced by PR #2475

PR #2474, opened 2021-12-22
Fix object stream

### LINKED ISSUE #2461: setElementModel forces an element to be streamed in no matter the distance from localPlayer
**Describe the bug**
As the title says, if you setElementModel on let's say an object that is streamed out, it will become streamed in (`onClientElementStreamedIn` triggered for it).
This also happens when you `engineFreeModel` a certain model ID that an element is using and it is streamed out, MTA will reset this element's model to default (1337 for objects) using `SetElementModel` and so it will be force-streamed in again.

This is unwanted behavior and particularly game-breaking for the `engineFreeModel` case.

**To reproduce**
Resource that will let you reproduce:

**Method 1 (fast)**
```lua
crun o = Object( 1337, 0, 0, 1000 )
crun #getElementsByType( "object", root, true ) --> 0
crun o:setModel( 1336 )
crun #getElementsByType( "object", root, true ) --> 1
```

**Method 2**

- download zip https://github.com/Fernando-A-Rocha/mta-samp-maploader
- download samp objects files and put in sampobj_reloaded/models from https://www.mediafire.com/file/mgqrk0rq7jrgsuc/models.zip/file
- ./start newmodels
- ./start sampobj_reloaded
- ./start samp_maploader
- To test go to blueberry /bb
- Wait for maps to finish loading then TP with /gotomap 1
- Then TP back to /bb
- You will notice the objects free and you should get the [BUG] message in chat because upon doing setElementModel, the element is streamed in again, despite you being in a completely different dimension from the objects.

**Expected behaviour**
Element's model changes when the element is streamed out. It should stay streamed out.

**Version**
(works on any version afaik)

Confirmed by @TheNormalnij 
EDIT: added other reproduction method

TheNormalnij:
Fixes #2461

### DIFF
--- a/Client/mods/deathmatch/logic/CClientObject.cpp
+++ b/Client/mods/deathmatch/logic/CClientObject.cpp
@@ -268,10 +268,14 @@ void CClientObject::SetModel(unsigned short usModel)
         m_pModelInfo = g_pGame->GetModelInfo(usModel);
         UpdateSpatialData();
 
-        // Request the new model so we can recreate when it's done
-        if (m_pModelRequester->Request(usModel, this))
+        // Create the object if we're streamed in
+        if (IsStreamedIn())
         {
-            Create();
+            // Request the new model so we can recreate when it's done
+            if (m_pModelRequester->Request(usModel, this))
+            {
+                Create();
+            }
         }
     }
 }


TheNormalnij:
Works for me. Potentiality fixes streaming related issues.

Inder00 (approved):
Works for me. Code looks good.

Pirulax (approved):

PR #2475, opened 2021-12-22
Fix RSA tests

TheNormalnij:
First test is fixed, RSA decryption test is broken itself, but encryption and decryption still work on client and server.

Broken tests block MTA loading on debug build, so it should be fixed asap.


Inder00:
Fixed in https://github.com/Inder00/mtasa-blue/commit/4df60aeeaccef4d220c413cab036c2c30db4bb07

Inder00:
Fixed in https://github.com/Inder00/mtasa-blue/commit/9af034c88ead17bb9b4d6a8cf2f137575f3472cc

Dutchman101:
> Fixed in [Inder00@9af034c](https://github.com/Inder00/mtasa-blue/commit/9af034c88ead17bb9b4d6a8cf2f137575f3472cc)

Please open a PR to implement the fixes, or add commits here

Lpsd:
This fixes the crashing on both client and server in debug builds.

However @Inder00's commit for the test portion still crashes the client & server, so I'm happy to push this through for now without the test considering it breaks debug builds completely. They can add the test back later.

Lpsd (approved):

PR #2477, opened 2021-12-23
Fix svgCreate crashing (#2456)

### LINKED ISSUE #2456: svgCreate causes random crashes

Lpsd:
Turns out you shouldn't create XML inside async task scheduler (specifically due to `CXMLArray::PopUniqueID` and stack ID stuff, that should only be done in main thread, ty sbx).

This PR generates the XML inside the Lua definition and passes it to the async task scheduler, instead of generating the XML inside the async task.

Resolves #2456


Lpsd:
Any chance of a quick review? svgCreate using callback arg is kinda broken without this

Inder00:
The fix works just fine for PoC with issue #2456, however I also encountered on the official build (before the fix) with another bug in the game shutdown situation.

image

Lpsd:
Yeah, that crash is related - this PR fixes it.

Inder00 (approved):
LGTM

Client/mods/deathmatch/logic/luadefs/CLuaVectorGraphicDefs.cpp
```diff
@@ -56,6 +56,22 @@ bool CLuaVectorGraphicDefs::LoadFromData(lua_State* luaVM, CClientVectorGraphic*
...
+bool CLuaVectorGraphicDefs::LoadFromNode(lua_State* luaVM, CClientVectorGraphic* pVectorGraphic, CXMLNode* pNode)
+{
+    if (!luaVM || !pVectorGraphic)
+        return false;
+
+    if (!pVectorGraphic->SetDocument(pNode))
+    {
+        delete pVectorGraphic;
+
+        m_pScriptDebugging->LogCustom(luaVM, "Unable to load SVG data (check for XML syntax errors)");
```
Pirulax, line 68: image


No, like honestly, you're calling this from a secondary thread, this is a bomb waiting to explode.

Pirulax:
I'm honestly surprised SVG works.
I just had a quick look at the source code, and it's pure luck.
There are so many calls to non thread-safe functions from secondary threads...
There will be more random crashes.
Good example is ` CLuaVectorGraphicDefs::SetSize` which is pushed as an async task in `CLuaVectorGraphicDefs::SVGSetSize`
Calling logging functions from secondary threads isn't good either.

Issue #2473, opened 2021-12-22
MTA crashes when streamed in object with custom model is deleted (0x000C4BB0)


Dezash:
**Describe the bug**
When a streamed in custom object, which is allocated by engineRequestModel is deleted due to the resource being stopped, MTA crashes with offset 0x000C4BB0.

**To reproduce**
crun customModel = engineRequestModel("object")
crun pickup = createPickup(me.position.x, me.position.y, me.position.z, 3, customModel)
/reconnect   or  /stop runcode

**Expected behaviour**
No crash

**Version**
Multi Theft Auto v1.5.9-release-21048

**Additional context**
There is a warning in the wiki that custom pickup models can cause crashes, however, I'm posting this anyway as it may be related to some unhandled logic in CClientModelManager.

Dutchman101:
It crashes in this GTA function:

```
int __thiscall CBaseModelInfo::RemoveRef(CBaseModelInfo *this)
{
  int v1; // eax
  int result; // eax

  v1 = this->m_nTxdIndex;
  --this->m_nRefCount;
  CTxdStore::RemoveRef(v1);
  return result;
}
```
Something not being cleared? cc @TheNormalnij 

Note: crash offset gta_sa.exe @ 0x000C4BB0 was already a known crash before engineRequestModel got added.. so this is a split cause, another trigger for a crash. Maybe it even tells more about the crash overall, like it's related to pickups in all cases. 

In fact, crash popularity (the 'original', so pre-engineRequestModel) appears to have been bumped by img/model loading tweaks from TheNormalnji along the course of 1.5.8 - it started somewhere shortly before r20779, i mean it's not measurable precisely due to crash stats and the popularity jump for this crash offset relying on the most well-circulated builds, but there are several model loading related changes before that revision, not even just by TheNormalnij (Btw, if it instead wasn't some unknown, bad practise at the top servers from the time of measuring: 95% occured on BS DayZ Ultimate, which now, with the introduction on engineRequestModel and its popularisation amongst servers, has expanded and risen to other servers at the top).. and now, another cause to rise it for engineRequestModel usage significantly dilutes the picture and what we can discover purely by crash stats. You can see im writing a lot of information & unknowns, because our investigation on the previously known main crash is simply unclear and surrounded by questions, but i want to mention as much as possible to better understand this.

[...]

PR #2486, opened 2021-12-25
Fix crash when pickup has invalid custom model

TheNormalnij:
Fixes one case from #2473
Tested with commands
```
crun customModel = engineRequestModel("object")
crun pickup = createPickup(me.position.x, me.position.y, me.position.z, 3, customModel)
restart runcode
```


Inder00:
Works for me.

Pirulax (approved):

PR #2494, opened 2021-12-30
Fix chatbox when using recently added inputBlocked argument

Pieter-Dewachter:
There is a "hack by ChrML" in the source code for the chatbox that makes sure it always gets focus when the chatbox is shown, this is needed for the _CharacterKeyHandler_ to get triggered. After some more intensive testing I found out that my PR #2170 partially broke that functionality when using _showChat(false, false)_.

Basically the code responsible for setting the focus was only called as part of getting the draw list for the chatbox, but when only the input of the chatbox is visible it was not called anymore. This new PR fixes the issue by moving the "hack" from _CChat::GetDrawList_ to an earlier point in the _CChat::Draw_ method which is also reached in case only the input is visible.

### DIFF
--- a/Client/core/CChat.cpp
+++ b/Client/core/CChat.cpp
@@ -165,6 +165,15 @@ void CChat::Draw(bool bUseCacheTexture, bool bAllowOutline)
     bool bUsingOutline = m_bTextBlackOutline && bAllowOutline && bUseCacheTexture;
     DrawInputLine(bUsingOutline);
 
+    if (m_bInputVisible)
+    {
+        // ChrML: Hack so chatbox input always works. It might get unfocused..
+        if (!m_pBackground->IsActive())
+        {
+            m_pBackground->Activate();
+        }
+    }
+
     // Are we visible?
     if (!m_bVisible)
         return;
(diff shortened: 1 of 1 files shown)


Inder00:
Can you attach a test resource ?

Pieter-Dewachter:
The problem does not occur 100% of the time, but the following script seems to trigger it reasonably consistent. Type /chat_test in the chatbox (NOT in the F8 console, it does not work that way) and click a few times randomly while the cursor is visible. Once the cursor is hidden again, you should see that you are unable to type in the chatbox. You are able to press 'T' to open the chatbox, but typing anything in it should not work. In case the bug is not triggered immediately, you can try again (for me it triggers about 9 out of 10 tries). Obviously it does not trigger anymore after this PR.

```Lua
addCommandHandler("chat_test", function()
    -- visible false, inputBlocked false
    showChat(false, false)

    -- try to defocus the chatbox after 500ms
    setTimer(function()
        outputDebugString("showing cursor")

        showCursor(true)
        guiSetInputMode("no_binds_when_editing")
    end, 500, 1)

    -- hide the cursor again after 5000ms
    setTimer(function()
        outputDebugString("hiding cursor")
        showCursor(false)
    end, 5000, 1)
end)
```

Inder00 (approved):
Works for me. LGTM

TheNormalnij (approved):
Simple fix. Works for me. I see no additional regressions and regression potential is missing.

PR #2495, opened 2022-01-02
Add new world special property (vehicle sun glare effect)

gta191977649:
### Introduction
This PR adds the vehicle sun glare effect to the  `setWorldSpecialPropertyEnabled`, `isWorldSpecialPropertyEnabled`. The effect is already presented in the SA engine and it's a leftover feature from VC, so it will add some classic aesthetic to the game.

QQ截图20220102234519

### Summary:
- This effect is `disabled` by default, keeping consistent to the original SA graphical favor.
- new propname: `vehiclesunglare` is added into the world function, you can set it on or off like other world properties. this includes `setWorldSpecialPropertyEnabled`, `isWorldSpecialPropertyEnabled`
- the vehicle glare is shown from 6:00 AM to 19:00 PM (excluding the raining or cloudy weather), it's the default behavior implemented by the game.
- Not all the vehicles will work, e.g. bandito, most `planes`, maybe due to they are not present in VC and it is no need for R* to update them, but most vehicles should work fine (maybe we need more investigation to find out).
### Code Example:
- Enable: `setWorldSpecialPropertyEnabled("vehiclesunglare",true)` 
- Disable: `setWorldSpecialPropertyEnabled("vehiclesunglare",false)` 

### Test Resource
[glare.zip](https://github.com/multitheftauto/mtasa-blue/files/7799367/glare.zip)
command `/glare` `on` or `off` to switch this effect.


TheNormalnij:
Thanks you for the PR!
Can you revert c5628c8? I think it's a small mistake.

gta191977649:
@TheNormalnij  allright, should cool now

Client/game_sa/CGameSA.cpp
```diff
@@ -34,16 +34,46 @@ unsigned long* CGameSA::VAR_Framelimiter;
 
 unsigned int OBJECTDYNAMICINFO_MAX = *(uint32_t*)0x59FB4C != 0x90909090 ? *(uint32_t*)0x59FB4C : 160;            // default: 160
 
+static int m_iVehicleSunGlare;
+#define HOOK_CAutomobile_OnVehiclePreRender 0x6ABCFD
```
TheNormalnij: I think it's better to put this in CMultiplayerSA.cpp. We use hooks there. Im not sure

gta191977649: An Example? i don't think i can get the `HOOK_CAutomobile_OnVehiclePreRender`,etc directly from `CMultiplayerSA.cpp to` `CGameSA.cpp`

> I think it's better to put this in CMultiplayerSA.cpp. We use hooks there. Im not sure

TheNormalnij: I see, some hooks are scoped there. Maybe, it's better to move this hooks in CVehicleSA::StaticSetHooks

Client/game_sa/CGameSA.cpp
```diff
@@ -33,17 +33,43 @@ float*         CGameSA::VAR_TimeStep;
 unsigned long* CGameSA::VAR_Framelimiter;
 
 unsigned int OBJECTDYNAMICINFO_MAX = *(uint32_t*)0x59FB4C != 0x90909090 ? *(uint32_t*)0x59FB4C : 160;            // default: 160
+static int m_iVehicleSunGlare;
```
TheNormalnij: Can bool be used?

gta191977649: > Can bool be used?

I've tried that before, but it seems will resulting crashing on the client.

Client/game_sa/CGameSA.cpp
gta191977649: @TheNormalnij Should be cool now, i moved the hooks and corespond code to the CVehicleSA, please check

Client/game_sa/CGameSA.cpp
TheNormalnij: `BOOL` is bool type with int size. Can be used here

Client/game_sa/CVehicleSA.cpp
```diff
@@ -2318,6 +2342,20 @@ void CVehicleSA::OnChangingPosition(const CVector& vecNewPosition)
...
+
+void CVehicleSA::SetVehiclesSunGlareEnable(bool bEnabled)
+{
+    m_iVehicleSunGlare = bEnabled ? 1 : 0;
+}
+bool CVehicleSA::GetVehiclesSunGlareEnable()
+{
+    return m_iVehicleSunGlare == 1 ? true : false;
+}
 namespace
```
TheNormalnij, line 2361: Add spaces between blocks, please.

Client/game_sa/CGameSA.cpp
gta191977649: > `BOOL` is bool type with int size. Can be used here

yeah, done! it's works

TheNormalnij: m_iVehicleSunGlare -> m_bVehicleSunGlare also

gta191977649: > m_iVehicleSunGlare -> m_bVehicleSunGlare also

Roger that

TheNormalnij (approved):

Inder00:
Marge time :shipit:

PR #2496, opened 2022-01-03
HMAC Support

### LINKED ISSUE #2003: Hmac support
Add `hmac` support programically.
Hmac is used by two-factor-authentication (ex. TOTP) to generating time-based (for TOTP) codes.
Algorithms that can be used are already built into the MTA, and they are `sha1`, `sha256`, `sha512`.

Shared functions:
- `string hmac( string value, string key, string algorithm )`

Inder00:
Fixes #2003
This pull request adds **HMAC Support**.

# hmac
```lua
string hash( "hmac", string data, table options { string key, string algorithm } )
```
Available algorithms: `md5`, `sha1`, `sha224`, `sha256`, `sha384`, `sha512`

## Examples
```lua
hash("hmac", "hello world", { key = "abc", algorithm = "md5" })
hash("hmac", "sha384 example", { key = "botder hecker", algorithm = "sha384" })
hash("hmac", "sha512 example", { key = "secret", algorithm = "sha512" })
```

To the moon 🚀


Breaker59:
❤️

Inder00:
I’ll refactor that feature to `encodeString` - https://github.com/multitheftauto/mtasa-blue/issues/2003#issuecomment-761828782

Pieter-Dewachter:
Maybe I am nitpicking here - or misunderstanding, but I feel like _encodeString_ is not the right function for HMAC.
Personally I think it makes more sense to add it as an option to the existing _hash_ function instead.

The current MTA standard seems to be as follows:
- _encodeString_ is used for ciphers (TEA and AES128 currently), which can be decoded if you have the correct key / info.
- _hash_ is used for hashes (unbelievable!), which means that you cannot go back to the input data (no decoding counterpart)

I think a syntax as follows would be quite nice: _hash ( string algorithm, string dataToHash [, table options ] )_
- New algorithms could be: "hmac-md5", "hmac-sha1", "hmac-sha224", "hmac-sha256", "hmac-sha384", "hmac-sha512"
- The new _options_ table is only needed for the HMAC algorithms as follows: _{ key = "secret" }_

Inder00:
> Maybe I am nitpicking here - or misunderstanding, but I feel like _encodeString_ is not the right function for HMAC. Personally I think it makes more sense to add it as an option to the existing _hash_ function instead.
> 
> The current MTA standard seems to be as follows:
> 
> * _encodeString_ is used for ciphers (TEA and AES128 currently), which can be decoded if you have the correct key / info.
> * _hash_ is used for hashes (unbelievable!), which means that you cannot go back to the input data (no decoding counterpart)
> 
> I think a syntax as follows would be quite nice: _hash ( string algorithm, string dataToHash [, table options ] )_
> 
> * New algorithms could be: "hmac-md5", "hmac-sha1", "hmac-sha224", "hmac-sha256", "hmac-sha384", "hmac-sha512"
> * The new _options_ table is only needed for the HMAC algorithms as follows: _{ key = "secret" }_

However, in my opinion, we should work on encode/decodeString as much as possible. This is a good strategy to make it easier to add new features and for scripters to use them.
Big guys should decide that - @sbx320 @botder @Pirulax

AlexTMjugador:
HMAC is defined as successive applications of a hash function to the message, mixing that with a secret key to provide authentication, so the argument that it's a hash function after all makes sense. However, I think that the `encodeString` function is named generically enough so that HMAC fits in there (and in fact so do more traditional hash functions such as MD5 or SHA). From Wikipedia:

> In communications and information processing, code is a system of rules to convert information—such as a letter, word, sound, image, or gesture—into another form, sometimes shortened or secret, for communication through a communication channel or storage in a storage medium.

If `encodeString` is easier to maintain, I don't see a strong reason why not favor that function for new additions.

Edit: when talking about _codes_ in such a generic form it also makes sense to think on _decoding_ them, but functions like HMAC are by design not meant to be decodeable. But then coding theory defines _code_ as just a set, so...

Pirulax:
IMHO we should just be consistent. 
Up to this point `encodeString` has only ciphers added to it. So let's continue that trend, and add only ciphers to it, while `hash` has the hashing functions.

Inder00:
So far I have done 3 versions of the hmac implementation.
Implmenetation under the hash function (most relevant to hmac) has been rewritten under a new parser argument.
Please give me feedback which implementation is the most suitable.

1. hmac function (https://github.com/multitheftauto/mtasa-blue/pull/2496/commits/30aefdb197e0b5f7472ba4d46bb20cb1fcad6492)
```lua
string hmac( string data, string key, string algorithm )
```

2. encodeString function (https://github.com/multitheftauto/mtasa-blue/pull/2496/commits/47f336aff790c7866df2468b1305af6a6fc2212b)
```lua
string encodeString( "hmac", string data, array options { string key, string algorithm }, [ function callback ] )
```

3. hash function (https://github.com/multitheftauto/mtasa-blue/pull/2496/commits/0862a4cf30d3a00c70456a67efe5ca566ae80b81)
```lua
string hash("hmac", string data, array options { string key, string algorithm } )
```

Inder00:
I'll also add that the cryptographic functions (md5, sha1, and derivatives...) on Linux returned the hash in lowercase, while on Windows in uppercase.
Refactoring the hash function can cause problems.

Linux needs to be tested ⏰

Pirulax:
I prefer the `hash` one.

Shared/sdk/SharedUtil.Hash.h
```diff
@@ -36,14 +47,24 @@ enum class StringEncodeFunction
...
 };
 
+enum class HmacAlgorithm
+{
+    MD5,
+    SHA1,
+    SHA224,
+    SHA256,
+    SHA384,
+    SHA512
```
botder:
```suggestion
    SHA512,
```

Shared/sdk/SharedUtil.Hash.h
```diff
@@ -36,14 +47,24 @@ enum class StringEncodeFunction
 {
     TEA,
     AES128,
-    RSA,
+    RSA
```
botder:
```suggestion
    RSA,
```

Shared/sdk/SharedUtil.Hash.h
```diff
@@ -22,11 +22,22 @@ namespace EHashFunction
...
 
+enum class HashFunctionType
+{
+    MD5,
+    SHA1,
+    SHA224,
+    SHA256,
+    SHA384,
+    SHA512,
+    HMAC
```
botder:
```suggestion
    HMAC,
```

Shared/sdk/SharedUtil.Hash.h
```diff
@@ -22,11 +22,22 @@ namespace EHashFunction
         SHA224,
         SHA256,
         SHA384,
-        SHA512,
+        SHA512
```
botder:
```suggestion
        SHA512,
```

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -39,6 +42,32 @@ namespace SharedUtil
...
+        return result;
+    }
+
+    template <class HmacType>
+    inline SString Hmac(const SString& value, const SString& key)
+    {
+        SString mac;
+        SString result;
+
+        CryptoPP::HMAC<HmacType> hmac((CryptoPP::byte*)key.c_str(), key.size());
```
botder:
```suggestion
        CryptoPP::HMAC<HmacType> hmac(reinterpret_cast<CryptoPP::byte*>(key.c_str()), key.size());
```

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -17,7 +17,7 @@ void CLuaCryptDefs::LoadFunctions()
     constexpr static const std::pair<const char*, lua_CFunction> functions[]{
         {"md5", ArgumentParserWarn<false, Md5>},
         {"sha256", ArgumentParserWarn<false, Sha256>},
-        {"hash", ArgumentParserWarn<false, Hash>},
+        {"hash", ArgumentParser<Hash>},
```
botder:
```suggestion
        {"hash", ArgumentParserWarn<false, Hash>},
```

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -49,9 +49,88 @@ std::string CLuaCryptDefs::Sha256(std::string strSourceData)
...
+
+                }
+
+                throw std::invalid_argument("Invalid hmac algorithm");
+        }
+
+        throw std::invalid_argument("Unknown algorithm");
+
+    }
+    catch(std::exception& ex)
```
botder:
```suggestion
    catch (std::exception& ex)
```

botder (changes requested):

botder (dismissed):

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -38,6 +41,32 @@ namespace SharedUtil
...
+        return result;
+    }
+
+    template <class HmacType>
+    inline SString Hmac(const SString& value, const SString& key)
+    {
+        SString mac;
+        SString result;
+
+        CryptoPP::HMAC<HmacType> hmac(reinterpret_cast<CryptoPP::byte*>(key.c_str()), key.size());
```
botder:
```suggestion
        CryptoPP::HMAC<HmacType> hmac(reinterpret_cast<const CryptoPP::byte*>(key.c_str()), key.size());
```

PR #2501, opened 2022-01-05
Bugfix: setElementModel reset handling for vehicles created on the server-side

### LINKED ISSUE #2427: Supporting setVehicleHandling on the client side, for vehicles created on the server side. (Custom vehicles, engineRequestModel)

To create a vehicle with a custom ID (engineRequestModel), you need to create the vehicle on the server side, then change the model on the client side. Unfortunately, changing the model resets the handling of the vehicle, and it is not possible to change the handling through the client side, so handling cannot be set for a custom vehicle.

Adding the ability to edit handling on the client side for vehicles created on the server side.

I used this solution which doesnt work for setVehicleHandling due to the described problem.
```
-- c-side
local v1,v2 = getVehicleVariant(source)
local handling = getVehicleHandling(source)
setElementModel(source,4542)
setVehicleVariant(source,v1,v2) -- return true, and I wish the setVehicleHandling function would work like this.
for i,v in pairs(handling) do
	setVehicleHandling(source,i,v) -- return false, because the vehicle is created on the server side. :(
end
```

Inder00:
Fixes #2427

That commit won't break any backwards compatibility.
Changing the model for vehicles created on the server side resets the handling only if the vehicle is created locally (client - resets client-side vehicle handling, server - resets server-side vehicle handling).
The problem was caused by the client not checking if the vehicle belongs to them (if it is local) and resetting the handling without checking.


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1068,32 +1068,35 @@ void CClientVehicle::SetModelBlocking(unsigned short usModel, unsigned char ucVa
         m_ucMaxPassengers = CClientVehicleManager::GetMaxPassengerCount(usModel);
 
         // Reset handling to fit the vehicle
-        ushort usHandlingModelID = usModel;
-        if (usHandlingModelID < 400 || usHandlingModelID > 611)
-            usHandlingModelID = m_pModelInfo->GetParentID();
+        if (restoreHandling)
```
botder: It looks like the value for `restoreHandling` always depends on `IsLocalEntity()`. Why not just use that here?

Inder00: That's a good point. I'll change that asap :)

botder (approved):

PR #2511, opened 2022-01-13
Add social buttons to main menu

ERAGON007:
ezgif-5-4975120f10

Implements #1547 with it's requested code changes and problems.

didn't work on that PR itself because of some issues with vendor files etc...

*EDIT: Added in-game preview


botder:
Can you provide an ingame screenshot and the source/license for these _socialset_ pictures?

ERAGON007:
not sure about source, but i got them from PR #1547

Inder00:
Discord icon needs to be [updated](https://discord.com/branding) to latest.

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +224,47 @@ CMainMenu::CMainMenu(CGUI* pManager)
...
+    };
+
+    m_socialMediaLinks = {
+        {"website", "https://multitheftauto.com"},
+        {"discord", "https://discord.gg/mtasa"},
+        {"twitter", "https://twitter.com/mtaqa"},
+        {"facebook", "https://www.facebook.com/multitheftauto"},
+        {"youtube", "https://www.youtube.com/user/mtaqa"},
+        {"github", "https://github.com/multitheftauto/mtasa-blue"},
+    };
```
botder: Move this outside the constructor function and replace it with
```cpp
static const std::array<std::pair<std::string, std::string>, 6> socialMediaLinks{{
    {"website", "https://multitheftauto.com"},
    {"discord", "https://discord.gg/mtasa"},
    {"twitter", "https://twitter.com/mtaqa"},
    {"facebook", "https://www.facebook.com/multitheftauto"},
    {"youtube", "https://www.youtube.com/user/mtaqa"},
    {"github", "https://github.com/multitheftauto/mtasa-blue"},
}};
```

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +224,47 @@ CMainMenu::CMainMenu(CGUI* pManager)
...
+        button->SetZOrderingEnabled(false);
+        button->SetMouseEnterHandler(GUI_CALLBACK(&CMainMenu::OnSocialMediaButtonHover, this));
+        button->SetMouseLeaveHandler(GUI_CALLBACK(&CMainMenu::OnSocialMediaButtonUnhover, this));
+        button->SetClickHandler(buttonClickHandler);
+
+        m_socialButtonLinks[button] = m_socialMediaLinks[name];
+
+        m_socialButtons[i++] = button;
+        offsetX += 0.03f;
+    }
```
botder: You should use lambda functions for the event handlers. You avoid the static methods in CMainMenu class and it keeps the handler functions close to the point of creation. Furthermore, you could've stored the URL in the CEGUI element user data field.
```cpp
    size_t socialsIndex = 0;

    for (const auto& nameAndLink : socialMediaLinks)
    {
        CGUIStaticImage* button = pManager->CreateStaticImage(m_pCanvas);
        m_socialsImages[socialsIndex++].reset(button);
        button->LoadFromFile(SString("cgui\\images\\socialset\\%s.png", nameAndLink.first.c_str()).c_str());
        button->SetProperty("InheritsAlpha", "False");
        button->SetAlpha(0.35f);
        // button->SetSize();
        // button->SetPosition();
        button->SetZOrderingEnabled(false);
        button->SetUserData(const_cast<char*>(nameAndLink.second.c_str()));
        button->SetMouseEnterHandler(GUI_CALLBACK(
            [](CGUIElement* button)
            {
                button->SetAlpha(0.08f);
                return true;
            }));
        button->SetMouseLeaveHandler(GUI_CALLBACK(
            [](CGUIElement* button)
            {
                button->SetAlpha(0.35f);
                return true;
            }));
        button->SetClickHandler(GUI_CALLBACK(
            [](CGUIElement* button)
            {
                auto url = static_cast<const char*>(button->GetUserData());
                ShellExecuteNonBlocking("open", url);
                return true;
            }));
    }
```

Client/core/CMainMenu.h
```diff
@@ -93,10 +93,18 @@ class CMainMenu
     bool OnQuitButtonClick(CGUIElement* pElement);
     bool OnNewsButtonClick(CGUIElement* pElement);
 
+    bool OnSocialMediaButtonClick(CGUIElement* pElement);
+    bool OnSocialMediaButtonHover(CGUIElement* pElement);
+    bool OnSocialMediaButtonUnhover(CGUIElement* pElement);
+
```
botder: Remove these if you start using the lambda functions.

Client/core/CMainMenu.h
```diff
@@ -93,10 +93,18 @@ class CMainMenu
...
     void HideServerInfo();
 
     CGUI* m_pManager;
 
+    CGUIStaticImage* m_socialButtons[6] = {nullptr};
+    std::map<CGUIStaticImage*, const char*> m_socialButtonLinks;
+    std::map<const char*, const char*> m_socialMediaLinks;
```
botder: These two std::map are unnecessary and the static images were leaked. Here is the replacement:
```cpp
std::array<std::unique_ptr<CGUIStaticImage>, 6> m_socialsImages{};
```

Client/core/CMainMenu.cpp
```diff
@@ -1010,6 +1054,29 @@ bool CMainMenu::OnNewsButtonClick(CGUIElement* pElement)
...
+    return true;
+}
+
+bool CMainMenu::OnSocialMediaButtonUnhover(CGUIElement* pElement)
+{
+    pElement->SetAlpha(0.35f);
+
+    return true;
+}
+
```
botder: Remove these if you start using the lambda functions.

ERAGON007:
> Discord icon needs to be [updated](https://discord.com/branding) to latest.

I've currently changed the code, icons can be updated later btw

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +224,53 @@ CMainMenu::CMainMenu(CGUI* pManager)
     m_pLatestNews->SetProperty("InheritsAlpha", "False");
     m_pLatestNews->SetVisible(false);
 
+    static const std::array<std::pair<std::string, std::string>, 6> m_socialMediaLinks{{
```
Pirulax: `string_view` or `const char*` pls no `std::string` needed here.
Also, it can be marked `constexpr` (Eg.: `static constexpr`)

Client/core/CMainMenu.h
```diff
@@ -97,6 +97,9 @@ class CMainMenu
 
     CGUI* m_pManager;
 
+    std::array<std::unique_ptr<CGUIStaticImage>, 6> m_socialsImages{};
+    std::map<const char*, const char*> m_socialMediaLinks;
```
Pirulax: Seems like this is unused. (But you've shadowed it by `static const std::array<std::pair<std::string, std::string>, 6> m_socialMediaLinks`)
You can get rid of this.

Pirulax (changes requested):

Pirulax:
Also IIRC there was a discussion about not using static links like this.
It could be routed thru `mtasa.com` or `bit.ly` or something similar (Probably @botder could help with that?)

ERAGON007:
> Also IIRC there was a discussion about not using static links like this. It could be routed thru `mtasa.com` or `bit.ly` or something similar (Probably @botder could help with that?)

This requires someone with access to build and api or smth for it, also social media URLs don't change often maybe we can avoid it and keep using static links?

Client/core/CMainMenu.h
ERAGON007: done

Client/core/CMainMenu.cpp
ERAGON007: done

botder:
We can setup redirection links on the webserver, for example: `https://mtasa.com/goto/<tag>`

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +224,53 @@ CMainMenu::CMainMenu(CGUI* pManager)
     m_pLatestNews->SetProperty("InheritsAlpha", "False");
     m_pLatestNews->SetVisible(false);
 
+    static const std::array<std::pair<const char*, const char*>, 6> m_socialMediaLinks{{
```
botder: Move this out of `CMainMenu::CMainMenu` into file scope and remove the `m_` prefix, please.

patrikjuvonen:
It should be fine to use those direct links. I don't recommend having us proxy them. More for us to maintain. These services are not going to change these urls, and if they are, they will provide a redirect as per good web etiquette.

With that said, make sure to avoid urls that redirect to another url. Always go direct to the source.

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +224,53 @@ CMainMenu::CMainMenu(CGUI* pManager)
...
+        {"website", "https://multitheftauto.com"},
+        {"discord", "https://discord.gg/mtasa"},
```
patrikjuvonen:
```suggestion
        {"discord", "https://discord.com/invite/mtasa"},
```

patrikjuvonen (changes requested):
Use direct links.

Client/core/CMainMenu.cpp
Inder00: What if your invite expires?
The botder idea of forwarding this from the MTA web server is safer in my opinion.

patrikjuvonen: Then we update it. It won't expire out of the blue.

Fernando-A-Rocha: > What if your invite expires?
> The botder idea of forwarding this from the MTA web server is safer in my opinion.

In general it's good practice to keep links in one place so you only have to update once. Afaik there is already [mtasa.com/discord](mtasa.com/discord) so it can be put here, no?

Edit: added hyperlink

patrikjuvonen: I'm part of the team that has to maintain this stuff, it's just not a good thing. There is very little chance these links would ever change, and we have full control over the main menu.

Vampiiree: Wouldn't be better to don't use a custom invite link once we would be relying on the server's boosts and they might expire or people might not want to boost again? Not sure if that ever happened but just in case.

patrikjuvonen: We are not relying on server boosts, Discord has granted us permanent vanity url as per their open source program. This has been the case for a long time now (since February 9 2021).

Fernando-A-Rocha: Very cool then sorry didn't know. Makes sense to use discord.com/invite/mtasa here then imo.

Client/core/CMainMenu.cpp
ERAGON007: fixed

botder (approved):

Client/core/CMainMenu.cpp
```diff
@@ -221,6 +233,44 @@ CMainMenu::CMainMenu(CGUI* pManager)
     m_pLatestNews->SetProperty("InheritsAlpha", "False");
     m_pLatestNews->SetVisible(false);
 
+    float offsetX = 0.01f;
+    size_t i = 0;
```
theSarrum, line 237: Is this line even used?

ERAGON007: `offsetX` is used on lines `249` and `271` for position calculation.
`socialsIndex` is used on line `244`

ERAGON007:
Don't reopen pls, the job was done by @Deihim007 on his own branch/PR. you can merge that if you want.

PR #2523, opened 2022-01-23
Stop async tasks belonging to resource

### LINKED ISSUE #1843: Async task scheduler tasks not removed after the VM starting that has closed
**Describe the bug**

**To reproduce**
Start an async function, for example, `passwordHash`

**Expected behaviour**
The async task should get aborted after the VM that has started it has closed

**Additional context**
Altough theres no task-vm association yet, there should be, as a resource might start a lot of async tasks, and then get restarted(when testing for example). Depending on the amount of tasks it can take quite a significant time to finish them.

I'll assign this to myself, just wanted to document it.

Pirulax:
Fixes #1843


Pirulax:
Now that I think about it.. I might've overcomplicated it.
I really wanted to separate `Async Task Sched` for any resource-related logic, and I used a linked list.. Honestly, just iterating through `AsyncTaskSched::m_Tasks` might've been a better solution.. Anyways..

PR #2535, opened 2022-02-01
Add getColorFilter

thisdp:
We have setColorFilter and resetColorFilter, but getColorFilter doesn't exist.

With the effect of color filter, dxDrawMaterialPrimitive3D doesn't show its real color on screen.

To remove the effect of color filter without change the color filter, I have to know the color of color filter.


Einheit-101:
Cant you disable color filter, draw the material line and then re-enable the color filter?

thisdp:
> 
Will it work?
I remember that dxDraw functions are not instant. Your draw call will be pushed into a queue, and will be rendered together.
I guess this won't work.

tederis:
**GetColorFilter** was not added for a reason. Because **SetColorFilter** actually **_overrides_** the color filter, not just set it. And trying to get a color when it doesn't override the defaults will result in an invalid **GetColorFilter** value. In another words, you cannot get a real filter color. Due to this inconsistency I decided to not include **GetColorFilter** to the **_API_**.

thisdp:
> **GetColorFilter** was not added for a reason. Because **SetColorFilter** actually **_overrides_** the color filter, not just set it. And trying to get a color when it doesn't override the defaults will result in an invalid **GetColorFilter** value. In another words, you cannot get a real filter color. Due to this inconsistency I decided to not include **GetColorFilter** to the **_API_**.

fixed

tederis:
> 

Perfect! But the inconsistency still remains a challenge. Because calling GetColorFilter you actually cannot distinguish GTA colors from overrides ones. The question is: is it a problem or not?

thisdp:
> > 
> 
> Perfect! But the inconsistency still remains a challenge. Because calling GetColorFilter you actually cannot distinguish GTA colors from overrides ones. The question is: is it a problem or not?

Not a big problem, because getColorFilter will get "current" color filter. 

Or we can add an extra argument to tell apart the overrided one from the original one.

Einheit-101:
I just tried it out, disabling the color filter and re-enabling it in the same frame has no effect. I wanted to temporarily disable it in order to draw my dx image without filtering.
Is it impossible to add an argument to dx functions that excludes them from the filtering effect? 
How can i equalize the filtering to restore original color if getColorFilter gets available?

tederis:
> I just tried it out, disabling the color filter and re-enabling it in the same frame has no effect. I wanted to temporarily disable it in order to draw my dx image without filtering. Is it impossible to add an argument to dx functions that excludes them from the filtering effect? How can i equalize the filtering to restore original color if getColorFilter gets available?

The color filter effect takes a place just before the GTA HUD rendering. When you draw an image you don't need to disable filtering since it had already taken place before onClientRender/onClientHUDRender (and even onClientPreRender when we are talking about dxDrawImage) events.

Einheit-101:
But for some reason my dx image actually is affected. Maybe because of the shader.

thisdp:
World 3D dx draws are affected, for example dxDrawMaterialLine3D

thisdp:
> But for some reason my dx image actually is affected. Maybe because of the shader.

I'd like to get the color filter, and let shader fight aginst the color filter.

Einheit-101:
Looks like my dx image is affected because it is drawn onClientHUDRender

thisdp:
> > 
> 
> Perfect! But the inconsistency still remains a challenge. Because calling GetColorFilter you actually cannot distinguish GTA colors from overrides ones. The question is: is it a problem or not?

now we have "isOriginal" argument for this

botder (dismissed):

botder (approved):

botder:
@thisdp Please create a page for getColorFilter on our wiki.

PR #2548, opened 2022-02-14
Add rotor functions for plane

gta191977649:
### Description
This PR adds the ability to get/set the plane's rotor, just similar to helicopters' rotor functions.
### Summary
* combined the plane and heli to get/setVehicleRotorSpeed(), now this function will work on both heli and plane!
* added `getVehicleRotorSpeed(vehicle)` and  `setVehicleRotorSpeed(vehicle,speed)`
* the function has exact identical behavior to the `set/getHelicopterRotorSpeed()`, has the same parameters and return type.
* This function also works on the jet plane, e.g. `at-400`, `shamal`, however, no visual effect on them, but will affect physics just like the heli does.
* the old  `set/getHelicopterRotorSpeed()` is still there and working, I didn't touch that.
### Sample code snippet
[rotor.zip](https://github.com/multitheftauto/mtasa-blue/files/8066541/rotor.zip)

Hope this PR fills the gap of rotor ralated things, now we got both heli & plane working :p


Zangomangu:
Nice work. Looking at the two functions, wouldnt it make sense to merge them?
Eg `get/setVehicleRotorSpeed`

Although that involves deprecating and later removing the heli one which is probably a hassle

gta191977649:
> Nice work. Looking at the two functions, wouldnt it make sense to merge them? Eg `get/setVehicleRotorSpeed`
> 
> Although that involves deprecating and later removing the heli one which is probably a hassle

It's possible, but it will require deprecating the old one, so breaks the backward compatibility.
Another possible way to do that is I can add heli detection into my `get/setPlaneRotorSpeed()` and combine them into one, so in the future MTA can properly deprecate the get/setHelicopterRotor().
idk, what you guys recommend?

AlexTMjugador:
Deprecating functions doesn't break backwards compatibility, as the only change would be that using the deprecated function shows a deprecation warning, and that's non-breaking.

I agree with @Zangomangu that that merging these functions in `get/setVehicleRotorSpeed` is nice, although it feels a bit weird that only two vehicle types will ever be able to have its rotor speed changed with that function. Anyway, there are functions like `getVehicleLandingGearDown` which only work on planes, so I think that the `vehicleRotorSpeed` naming is consistent at least.

bum8hj:
Perhaps this is inconsistent, thus not a good idea, but wouldn't `setAircraftRotorSpeed` be more suitable since (from what I know) this only works for aircrafts?

gta191977649:
> Deprecating functions don't break backwards compatibility, as the only change would be that using the deprecated function shows a deprecation warning, and that's non-breaking.
> 
> I agree with @Zangomangu that merging these functions in `get/setVehicleRotorSpeed` is nice, although it feels a bit weird that only two vehicle types will ever be able to have their rotor speed changed with that function. Anyway, there are functions like `getVehicleLandingGearDown` which only work on planes, so I think that the `vehicleRotorSpeed` naming is consistent at least.

Sounds reasonable, I've update my PR to `get/setVehicleRotorSpeed` approach, now this function works on both plane & heli, the old `set/getHelicopterRotorSpeed()` still remains untouched, for backward compatibility.

gta191977649:
Done, The PR is ready for review.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -52,6 +52,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"getVehicleNameFromModel", GetVehicleNameFromModel},
         {"getVehicleAdjustableProperty", GetVehicleAdjustableProperty},
         {"getHelicopterRotorSpeed", GetHelicopterRotorSpeed},
+        {"getVehicleRotorSpeed", GetVehicleRotorSpeed},
```
AlexTMjugador: New functions should use the new argument parser and not be defined in `CStaticFunctionDefinitions`. More context: https://github.com/multitheftauto/mtasa-blue/pull/1644#discussion_r480343689, https://github.com/multitheftauto/mtasa-blue/pull/1648

AlexTMjugador (changes requested):

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -3316,6 +3333,21 @@ bool CStaticFunctionDefinitions::SetHelicopterRotorSpeed(CClientVehicle& Vehicle
...
+
+    if (Vehicle.GetVehicleType() == CLIENTVEHICLE_HELI)
+    {
+        Vehicle.SetHeliRotorSpeed(fSpeed);
+        return true;
+    }
+    else if (Vehicle.GetVehicleType() == CLIENTVEHICLE_PLANE) {
+        Vehicle.SetPlaneRotorSpeed(fSpeed);
+        return true;
+    }
```
AlexTMjugador: Can this be refactored to a `switch` statement? I think it'd look more elegant that way.

AlexTMjugador:
> Perhaps this is inconsistent, thus not a good idea, but wouldn't `setAircraftRotorSpeed` be more suitable since (from what I know) this only works for aircrafts?

I think you've a point, and aircraft is a good word to choose, but so far it has only been used in the function names of `getAircraftMaxHeight` and `getAircraftMaxVelocity`, which are in the "world functions" category, not the "vehicle functions" category on the wiki. There are lots of `get/setVehicleSomething` that don't work on any vehicle type anyway, and it's conventional that functions that manipulate a vehicle element have the `vehicle` infix. Functions like `get/setHeliBladeCollisionsEnabled` and `get/setHelicopterRotorSpeed` are, coincidentally, the odd ones.

gta191977649:
@AlexTMjugador, Done! 
upgrade to the new parser.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
gta191977649: > Can this be refactored to a `switch` statement? I think it'd look more elegant that way.

Sure, I'll do the refactoring.

AlexTMjugador:
Nice! Is there any reason to keep the `CStaticFunctionDefinitions::GetVehicleRotorSpeed` and `CStaticFunctionDefinitions::SetVehicleRotorSpeed` methods, instead of doing their logic in `CLuaVehicleDefs::GetVehicleRotorSpeed` and `CLuaVehicleDefs::SetVehicleRotorSpeed`, respectively?

gta191977649:
> CStaticFunctionDefinitions

not really, I'm being a bit lazy
Alright, I'm gonna do that do it now.

gta191977649:
> CLuaVehicleDefs

@AlexTMjugador Done!
Removed the implementation from `CStaticFunctionDefinitions` and moved it all to `CLuaVehicleDefs`.
Thanks for taking your time for the review, anything else?

Client/game_sa/CVehicleSA.h
```diff
@@ -636,6 +636,7 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
     bool           GetTyresDontBurst() { return GetVehicleInterface()->m_nVehicleFlags.bTyresDontBurst; };
     unsigned short GetAdjustablePropertyValue() { return *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned long>(m_pInterface) + 2156); };
     float          GetHeliRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124); };
+    float          GetPlaneRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 0x09C4); };
```
AlexTMjugador: I personally like using hexadecimal offset literals here, but the surrounding code uses decimal offset literals. I'm not sure whether we should prefer consistency with the existing code over my style preferences. What do others think about this?

Client/game_sa/CVehicleSA.h
```diff
@@ -660,6 +661,7 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
         *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned int>(m_pInterface) + 2156) = usAdjustableProperty;
     };
     void SetHeliRotorSpeed(float fSpeed) { *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124) = fSpeed; };
+    void SetPlaneRotorSpeed(float fSpeed) { *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 0x09C4) = fSpeed; };
```
AlexTMjugador: Same as above.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.h
```diff
@@ -117,6 +119,8 @@ class CLuaVehicleDefs : public CLuaDefs
     LUA_DECLARE(SetVehicleFrozen);
     LUA_DECLARE(SetVehicleAdjustableProperty);
     LUA_DECLARE(SetHelicopterRotorSpeed);
+    //LUA_DECLARE(SetVehicleRotorSpeed);
+    static bool SetVehicleRotorSpeed(CClientVehicle* pVehicle,float fSpeed);
```
AlexTMjugador: New functions do not need to point out that they use the new argument parser in a comment; they just can use it directly.

Also, I think you might forgot to run the clang formatter, as it'd format some return types with an additional space.

```suggestion
    static bool SetVehicleRotorSpeed(CClientVehicle* pVehicle, float fSpeed);
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.h
```diff
@@ -59,6 +59,8 @@ class CLuaVehicleDefs : public CLuaDefs
     LUA_DECLARE(GetVehicleNameFromModel);
     LUA_DECLARE(GetVehicleAdjustableProperty);
     LUA_DECLARE(GetHelicopterRotorSpeed);
+    //LUA_DECLARE(GetVehicleRotorSpeed); use new parser instead
+    static std::variant<bool,float> GetVehicleRotorSpeed(CClientVehicle* pVehicle);
```
AlexTMjugador: Same as above:

```suggestion
    static std::variant<bool, float> GetVehicleRotorSpeed(CClientVehicle* pVehicle);
```

AlexTMjugador (approved):
Thank you, I think it looks better now :smile: 

There are some more nitpicks I wanted to point out, though, but I'd be fine with this being merged as-is.

Client/game_sa/CVehicleSA.h
gta191977649: personally, I l like to go with hex literals too, I think it's kinda a universal standard to represent something involved with low-level things. but seems the code for heli is using decimal, so there are 2 solution comes up from my mind.
1. Should I change old the heli's address to hex? this will make it constant, I guess the majority of the memory address related code uses hex representation in MTA, since it's a universal agreement.
2. Should I stay with the heli's rotor format, using decimals?

What would you suggest? @AlexTMjugador

AlexTMjugador: I'd suggest 1, but it'd nice to know what other developers think about it.

gta191977649:
Done! Refactor the rotor related code to use `CPlaneSAInterface` instead.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.h
```diff
@@ -160,6 +162,7 @@ class CLuaVehicleDefs : public CLuaDefs
     static bool SetVehicleModelWheelSize(const unsigned short usModel, const eResizableVehicleWheelGroup eWheelGroup, const float fWheelSize);
     static int  GetVehicleWheelFrictionState(CClientVehicle* pVehicle, unsigned char wheel);
 
+
```
AlexTMjugador: This looks like an unintended white line change. If so, please remove it.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -1291,6 +1293,31 @@ int CLuaVehicleDefs::GetHelicopterRotorSpeed(lua_State* luaVM)
     return 1;
 }
 
+std::variant<bool,float> CLuaVehicleDefs::GetVehicleRotorSpeed(CClientVehicle* pVehicle)
+{
+    float fSpeed = 0.0f;
+    // It's a plane or heli?
```
AlexTMjugador: I think it's pretty obvious what the `switch` statement does, so I would remove the comment explaining it.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -2202,6 +2229,30 @@ int CLuaVehicleDefs::SetHelicopterRotorSpeed(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaVehicleDefs::SetVehicleRotorSpeed(CClientVehicle* pVehicle, float fSpeed)
+{
+    // It's a plane or heli?
```
AlexTMjugador: Same as above.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1542,6 +1543,13 @@ float CClientVehicle::GetHeliRotorSpeed()
     return m_fHeliRotorSpeed;
 }
 
+float CClientVehicle::GetPlaneRotorSpeed()
+{
+    if (m_pVehicle && m_eVehicleType == CLIENTVEHICLE_PLANE)
+        return m_pVehicle->GetPlaneRotorSpeed();
+    return m_fPlaneRotorSpeed;
```
AlexTMjugador: I think it looks a bit prettier, and more consistent with surrounding code, if you add a white line before the return:

```suggestion
    if (m_pVehicle && m_eVehicleType == CLIENTVEHICLE_PLANE)
        return m_pVehicle->GetPlaneRotorSpeed();

    return m_fPlaneRotorSpeed;
```

Client/game_sa/CVehicleSA.h
```diff
@@ -715,6 +720,8 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
 
     CVehicleSAInterface* GetVehicleInterface() { return (CVehicleSAInterface*)m_pInterface; }
 
+ 
+
```
AlexTMjugador: These look like unintended white line changes.

Client/game_sa/CVehicleSA.h
```diff
@@ -636,6 +636,8 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
     bool           GetTyresDontBurst() { return GetVehicleInterface()->m_nVehicleFlags.bTyresDontBurst; };
     unsigned short GetAdjustablePropertyValue() { return *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned long>(m_pInterface) + 2156); };
     float          GetHeliRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124); };
+    //float        GetPlaneRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 0x09C4); };
```
AlexTMjugador: I'd remove this comment and put the `0x09C4` offset in a comment on the `m_fPropSpeed` field of the CPlaneSAInterface class.

Client/game_sa/CVehicleSA.h
```diff
@@ -660,6 +662,9 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
         *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned int>(m_pInterface) + 2156) = usAdjustableProperty;
     };
     void SetHeliRotorSpeed(float fSpeed) { *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124) = fSpeed; };
+    //void SetPlaneRotorSpeed(float fSpeed) { *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 0x09C4) = fSpeed; };
```
AlexTMjugador: Same as above.

AlexTMjugador (commented):
I think the PR looks much better now! We got rid of new hardcoded offsets in the code :tada: 

Here there are some nitpicks I think may be worth sharing.

AlexTMjugador:
Two more things I'd like to ask before giving my symbolic approval to this PR:

- Did you run the clang code formatter on the PR? Use this script if not: https://github.com/multitheftauto/mtasa-blue/blob/master/utils/win-apply-clang-format.bat
- Have you tested that things continue to work well in-game after the refactors?

gta191977649:
> Two more things I'd like to ask before giving my symbolic approval to this PR:
> 
> * Did you run the clang code formatter on the PR? Use this script if not: https://github.com/multitheftauto/mtasa-blue/blob/master/utils/win-apply-clang-format.bat
> * Have you tested that things continue to work well in-game after the reactors?

1. for the clang formatter , it pops up an error says `Cannot find clang-format-r325576.exe,` i guess it may not contain in the source code, so where do I obtain it then?
2. for reactors code testing, yes, I've tested it in my local machine, it behaves identically to the code before refactoring.

gta191977649:
@AlexTMjugador alright, i've updated the changes according to the comments.
please have a look :)

AlexTMjugador:
> 1. for the clang formatter , it pops up an error says `Cannot find clang-format-r325576.exe,` i guess it may not contain in the source code, so where do I obtain it then?

The comments in the format script point out where to get the needed executables. For the clang formatter [try out this link](https://prereleases.llvm.org/win-snapshots/clang-format-r325576.exe), and for fnr.exe [this one](http://findandreplace.io/download).

Lpsd:
Just so you know for future, both Visual Studio and Visual Studio Code have support for formatting via the provided `.clang_format` file in the repo - simply use the relevant formatting shortcuts while writing code (if you don't want to use the .bat file).

gta191977649:
> * clang

Alright, I'll run through clang now

AlexTMjugador (dismissed):
Taking @gta191977649's word that these changes were tested to work okay, I think this is ready to merge.

For other reviewers: some files unrelated to the PR purpose itself were modified due to the clang formatter being run repo-wide. See [this commit](https://github.com/multitheftauto/mtasa-blue/pull/2548/commits/215fe2197cfc64efdc21a116846694d8f18c26e9) for the exact clang formatter diff.

gta191977649:
So anyone else wants to review this? 
perhaps it's being hanging up there for too long.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -1291,6 +1293,30 @@ int CLuaVehicleDefs::GetHelicopterRotorSpeed(lua_State* luaVM)
...
 }
 
+std::variant<bool, float> CLuaVehicleDefs::GetVehicleRotorSpeed(CClientVehicle* pVehicle)
+{
+    float fSpeed = 0.0f;
+    switch (pVehicle->GetVehicleType())
+    {
+        case CLIENTVEHICLE_PLANE:
+        {
+            fSpeed = pVehicle->GetPlaneRotorSpeed();
```
tederis: Why not just return it?

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -292,6 +292,9 @@ class CClientVehicle : public CClientStreamElement
     float GetHeliRotorSpeed();
     void  SetHeliRotorSpeed(float fSpeed);
 
+    float GetPlaneRotorSpeed();
```
tederis, line 300: It is highly recommended to make Get methods **const**.

gta191977649: all right, updated

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -1291,6 +1293,26 @@ int CLuaVehicleDefs::GetHelicopterRotorSpeed(lua_State* luaVM)
     return 1;
 }
 
+std::variant<bool, float> CLuaVehicleDefs::GetVehicleRotorSpeed(CClientVehicle* pVehicle)
+{
+    float fSpeed = 0.0f;
```
tederis: Looks like fSpeed is not used

Client/game_sa/CPlaneSA.h
```diff
@@ -27,4 +88,6 @@ class CPlaneSA : public virtual CPlane, public virtual CAutomobileSA
 public:
     CPlaneSA(CPlaneSAInterface* plane);
     CPlaneSA(eVehicleTypes dwModelID, unsigned char ucVariation, unsigned char ucVariation2);
+
+    CPlaneSAInterface* GetPlaneInterface() { return (CPlaneSAInterface*)m_pInterface; };
```
tederis: Don't forget to use a const modifier

Client/sdk/game/CVehicle.h
```diff
@@ -212,6 +212,7 @@ class CVehicle : public virtual CPhysical
     virtual bool           GetTyresDontBurst() = 0;
     virtual unsigned short GetAdjustablePropertyValue() = 0;
     virtual float          GetHeliRotorSpeed() = 0;
+    virtual float          GetPlaneRotorSpeed() = 0;
```
tederis, line 202: It's the most marvellous place to put a const modifier

gta191977649: most marvellous place?
So should I put or not?

gta191977649: @tederis most marvellous place?
So should I put or not?

Lpsd: yes, most the time your getters should be const

lopezloo:
Looks like it doesn't compile?

PlatinMTA:
last two commits don't work afaik

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -3883,11 +3900,15 @@ void CClientVehicle::SetHeadLightColor(const SColor color)
...
+    #define WARN(x) g_pCore->GetConsole()->Printf x
 #else
-    #define INFO(x)    {}
-    #define WARN(x)    {}
+    #define INFO(x) \
+        { \
+        }
+    #define WARN(x) \
+        { \
+        }
```
lopezloo, line 3911: Why it looks like that?

Client/sdk/game/CVehicle.h
```diff
@@ -212,6 +212,7 @@ class CVehicle : public virtual CPhysical
     virtual bool           GetTyresDontBurst() = 0;
     virtual unsigned short GetAdjustablePropertyValue() = 0;
     virtual float          GetHeliRotorSpeed() = 0;
+    virtual float          GetPlaneRotorSpeed() const = 0;
```
lopezloo: You have added `const` here but not in `CVehicleSA.h` and `CVehicleSA.cpp`. Also, it looks like this method can't be constant because it calls `GetVehicleInterface()` which is not constant? I think we can make `GetVehicleInterface()` constant though.

Client/game_sa/CVehicleSA.cpp
```diff
@@ -611,6 +611,18 @@ void CVehicleSA::SetTrainSpeed(float fSpeed)
     pInterface->m_fTrainSpeed = fSpeed;
 }
 
+void CVehicleSA::SetPlaneRotorSpeed(float fSpeed)
+{
+    auto pInterface = static_cast<CPlaneSAInterface*>(GetVehicleInterface());
+    pInterface->m_fPropSpeed = fSpeed;
+}
+
+float CVehicleSA::GetPlaneRotorSpeed()
```
lopezloo:
```suggestion
float CVehicleSA::GetPlaneRotorSpeed() const
```

Client/game_sa/CVehicleSA.h
```diff
@@ -636,6 +636,7 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
     bool           GetTyresDontBurst() { return GetVehicleInterface()->m_nVehicleFlags.bTyresDontBurst; };
     unsigned short GetAdjustablePropertyValue() { return *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned long>(m_pInterface) + 2156); };
     float          GetHeliRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124); };
+    float          GetPlaneRotorSpeed();
```
lopezloo, line 544:
```suggestion
    float          GetPlaneRotorSpeed() const;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
lopezloo:
```suggestion
    #define INFO(x)    {}
    #define WARN(x)    {}
```
Previous code is more readable.

patrikjuvonen:
Build is failing.

gta191977649:
> Build is failing.

@patrikjuvonen fixed all the build problems & updated the PR to align with the latest MTA code, ready for review.

AlexTMjugador (commented):
GitHub keeps telling me that I have a pending review of this PR even though I can't submit an approving review, but for what it's worth, I think this is good to merge!

patrikjuvonen:
I think we should go ahead and deprecate the old helicopter functions. See how to deprecate functions over here: https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/logic/CResourceChecker.Data.h#L444-L517

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1545,6 +1554,14 @@ void CClientVehicle::SetHeliRotorSpeed(float fSpeed)
     m_fHeliRotorSpeed = fSpeed;
 }
 
+void CClientVehicle::SetPlaneRotorSpeed(float fSpeed)
+{
+    if (m_pVehicle && m_eVehicleType == CLIENTVEHICLE_PLANE)
+        m_pVehicle->SetPlaneRotorSpeed(fSpeed);
```
botder, line 1560: You also have to apply `m_fPlaneRotorSpeed` like we do with `m_fHeliRotorSpeed` when the vehicle gets created.

https://github.com/multitheftauto/mtasa-blue/blob/e66cc7a4623f65d46a4b711371197c2ba57aa926/Client/mods/deathmatch/logic/CClientVehicle.cpp#L2551

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -1293,6 +1295,25 @@ int CLuaVehicleDefs::GetHelicopterRotorSpeed(lua_State* luaVM)
...
+    switch (pVehicle->GetVehicleType())
+    {
+        case CLIENTVEHICLE_PLANE:
+        {
+            return pVehicle->GetPlaneRotorSpeed();
+        }
+        case CLIENTVEHICLE_HELI:
+        {
+            return pVehicle->GetHeliRotorSpeed();
+        }
```
botder: Wouldn't it be easier to read this, if you moved this logic to `CClientVehicle`?
For example `bool CClientVehicle::GetRotorSpeed(float&)` and `bool CClientVehicle::SetRotorSpeed(float)`

botder (dismissed):

Client/mods/deathmatch/logic/CClientVehicle.cpp
AlexTMjugador: Indeed, without that method call chances are that the plane rotor speed won't apply properly on stream-in. Good catch!

botder: @gta191977649 Can you do this?

gta191977649: ah yeah, nice catch

gta191977649: > @gta191977649 Can you do this?

Yeah, sorry for the long wait :x

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
gta191977649: updated :)

Client/game_sa/CPlaneSA.h
```diff
@@ -14,16 +14,75 @@
 #include <game/CPlane.h>
 #include "CAutomobileSA.h"
 
+enum ePlaneNodes
```
TheNormalnij:
```suggestion
enum class ePlaneNodes
```

Client/game_sa/CPlaneSA.h
```diff
@@ -14,16 +14,75 @@
...
+    float        m_fLandingGearStatus;
+    int32        m_planeDamageWave;
+    FxSystem_c** m_pGunParticles;
+    uint8        m_nFiringMultiplier;
+    int32        field_9DC;
+    int32        field_9E0;
+    int32        field_9E4;
+    FxSystem_c*  m_apJettrusParticles[4];
+    FxSystem_c*  m_pSmokeParticle;
+    uint32       m_nSmokeTimer;
```
TheNormalnij:
```suggestion
    uint32_t     m_nStartedFlyingTime;
    float        m_fPropSpeed;            // Rotor speed 0x09C4
    float        field_9C8;
    float        m_fLandingGearStatus;
    int32_t      m_planeDamageWave;
    FxSystem_c** m_pGunParticles;
    uint8_t      m_nFiringMultiplier;
    int32_t      field_9DC;
    int32_t      field_9E0;
    int32_t      field_9E4;
    FxSystem_c*  m_apJettrusParticles[4];
    FxSystem_c*  m_pSmokeParticle;
    uint32_t     m_nSmokeTimer;
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -2204,6 +2217,16 @@ int CLuaVehicleDefs::SetHelicopterRotorSpeed(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaVehicleDefs::SetVehicleRotorSpeed(CClientVehicle* pVehicle, float fSpeed)
+{
+    if (pVehicle == nullptr)
+    {
+        return false; 
+    }
```
TheNormalnij: Do you need this check?

Client/game_sa/CPlaneSA.h
gta191977649: updated

Client/game_sa/CPlaneSA.h
```diff
@@ -14,16 +14,75 @@
...
+    PLANE_ELEVATOR_R = 18,
+    PLANE_AILERON_L = 19,
+    PLANE_AILERON_R = 20,
+    PLANE_GEAR_L = 21,
+    PLANE_GEAR_R = 22,
+    PLANE_MISC_A = 23,
+    PLANE_MISC_B = 24,
+    PLANE_NUM_NODES
+};
+
```
TheNormalnij: You don't use this enums

gta191977649: @TheNormalnij just updated =)

TheNormalnij (approved):

Fernando-A-Rocha:
https://wiki.multitheftauto.com/wiki/SetHelicopterRotorSpeed and https://wiki.multitheftauto.com/wiki/GetHelicopterRotorSpeed will be marked as deprecated ??

PR #2552, opened 2022-02-17
Improve setSoundEffectParameters error messages

Pirulax:
Right now, if anything goes wrong in BASS a very weird error message is given:
image

Now, instead it displays a pretty useful one:
image

But there's a catch.. `luaL_error` straight out crashes when I call it with a format string that has strings in it. I tested the feature by replacing `luaL_error` with  `m_pScriptDebugging->LogError` (which btw uses format string, but works fine..)

Now, there's ~~3~~ 4 options:
- Fix `luaL_error`
- Use `sprintf`
- Use `SString` to construct it
- Upgrade to C++20 and use `std::format` :D

Also, now it hard errors if the effect wasn't enabled (currently it only returns false, but we should hard error on usage mistakes) with the following message:
`Effect's parameters can't be set unless it's enabled"`


Pirulax:
`luaL_error` workaround added. I can't be bothered to fix it.
Good enough for now.
Let's get this merged fast, it's a pretty nice QoL enhancement.

Lpsd:
I did already reply on Discord, but your `luaL_error` crash is likely due to using `%i` instead of `%d` for int, see [here](https://www.lua.org/manual/5.3/manual.html#luaL_error) (and [here](https://www.lua.org/manual/5.3/manual.html#lua_pushfstring))

Pirulax:
Well, thank you very much!

Client/mods/deathmatch/logic/CBassAudio.cpp
```diff
@@ -237,6 +237,109 @@ bool CBassAudio::BeginLoadingMedia()
...
+    case BASS_ERROR_DENIED:
+         return "access denied";
+    case BASS_ERROR_UNKNOWN:	
+         return "some other mystery problem";
+
+    // End of possible BASS error codes
+    // If this label is reache that means there's a
+    // new error code that has to be added here
+    default:
+        return "Unknown error - `CBassAudio::ErrorGetMessage` should be updated to return a message for it!";
```
botder:
```suggestion
        return "unknown error (?)";
```

Client/mods/deathmatch/logic/CBassAudio.h
```diff
@@ -94,6 +94,12 @@ class CBassAudio
     float  GetSoundBPM();
     void   SetSoundBPM(float fBPM) { m_fBPM = fBPM; }
 
+    // Bass audio helpers
+    // For now these AREN'T PER SOUND ERRORS, so they always refer to the last modified sound.
+    // One day we could store errors for each sound separtely.
+    static const char* ErrorGetMessage();
+    static int         ErrorGetCode();
```
botder:
```suggestion
    // Retrieves the error message for the most recent BASS function call in the current thread.
    static const char* ErrorGetMessage();

    // Retrieves the error code for the most recent BASS function call in the current thread.
    static int ErrorGetCode();
```

Client/mods/deathmatch/logic/CBassAudio.cpp
```diff
@@ -237,6 +237,109 @@ bool CBassAudio::BeginLoadingMedia()
...
+    case BASS_ERROR_PROTOCOL:	
+         return "unsupported protocol";
+    case BASS_ERROR_DENIED:
+         return "access denied";
+    case BASS_ERROR_UNKNOWN:	
+         return "some other mystery problem";
+
+    // End of possible BASS error codes
+    // If this label is reache that means there's a
+    // new error code that has to be added here
```
botder:
```suggestion
    // BASS may add new error codes in the future, which must be appended to the switch above.
    // Also, add-ons may introduce additional error codes.
```

Client/mods/deathmatch/logic/CBassAudio.cpp
```diff
@@ -237,6 +237,109 @@ bool CBassAudio::BeginLoadingMedia()
...
+    default:
+        return "Unknown error - `CBassAudio::ErrorGetMessage` should be updated to return a message for it!";
+    }
+}
+
+// Just a wrapper
```
botder:
```suggestion
```

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -1401,467 +1402,452 @@ int CLuaAudioDefs::SetSoundEffectParameter(lua_State* luaVM)
     argStream.ReadUserData(pSound);
     argStream.ReadEnumString(eEffectType);
 
-    if (argStream.HasErrors())
+    // Hard error on usage mistake
```
botder:
```suggestion
```

botder (changes requested):

botder (approved):

PR #2554, opened 2022-02-18
Add new onClientCoreCommand event

Pieter-Dewachter:
MTA has a few built-in clientside commands, however there is currently no way for scripts to know if any of those commands were executed. All existing ways of retrieving such info don't work for various reasons, namely: _onPlayerCommand_, _addCommandHandler_ and _onClientConsole_. This pull request adds a new clientside event _onClientCoreCommand_ which gets triggered for those specific commands, with one parameter: _string command_.

FYI: There are a lot of built-in client commands,, some are written here as reference:
- help
- ver
- time
- showhud
- serial
- bind
- unbind
- copygtacontrols
- screenshots
- saveconfig
- cleardebug
- test
- showmemstat
- showframegraph
- fakelag
- shownametags
- showchat
- shownetstat
- showcol
- showsound
- ...


lopezloo:
What is the use case for this? I see point in triggering such event for commands like "showchat", "showhud" or "shownametags" but why for all commands?

Pieter-Dewachter:
I personally would use it for "showchat" and "time", but I figured other people could have a use for the other commands as well.

patrikjuvonen:
I don't see the point in this.

PlatinMTA:
You can't really add command handlers for hardcoded commands. For example if I want to make a command using /ver I can't, I need to name it differently. This should help.

ffsPLASMA:
You would have to blacklist certain commands like login/register as you could simply build a keylogger for them.

Fernando-A-Rocha:
> You would have to blacklist certain commands like login/register as you could simply build a keylogger for them.

https://wiki.multitheftauto.com/wiki/Client_Commands

login/register aren't client commands, and register isn't even an MTA command, it's built into the admin panel resource.

The only client command that pontentially exposes a password is `connect <host> [<port> <nick> <pass>]`.

It would be useful to detect when the player uses /help or /time to display other stuff in the chatbox or in some UI.

Proxy-99:
I think it can be useful with side scripting with those commands to add extra functionality to them

PR #2558, opened 2022-02-20
Fix #2545: isElementInWater in vehicles (client)

### LINKED ISSUE #2545: isElementInWater in vehicles
**Describe the bug**
isElementInWater returns false if ped or player is in vehicle - Returns true only for the vehicle. 
**This bug occurs only on client side**

**To reproduce**
1. Get vehicle
2. Drive into the water 
3. 
```lua
print(isElementInWater(me)) -- false
print(isElementInWater(me.vehicle)) -- true
```

**Expected behaviour**
isElementInWater should returns true also for peds/players which are in vehicle in water

**Screenshots**

**Version**
1.5.9

**Additional context**

matheustkn:
Fixed issue #2545

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -766,8 +766,16 @@ bool CStaticFunctionDefinitions::IsElementInWater(CClientEntity& Entity, bool& b
         case CCLIENTPLAYER:
         {
             CClientPed& Ped = static_cast<CClientPed&>(Entity);
-            bInWater = Ped.IsInWater();
-            break;
+            if (Ped.GetOccupiedVehicle())
+            {
+                bInWater = Ped.GetOccupiedVehicle()->IsInWater();
+                break;
+            }
+            else
+            {
+                bInWater = Ped.IsInWater();
+                break;
+            }
         }
         case CCLIENTVEHICLE:
         {


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -766,8 +766,16 @@ bool CStaticFunctionDefinitions::IsElementInWater(CClientEntity& Entity, bool& b
...
+            if (Ped.GetOccupiedVehicle())
+            {
+                bInWater = Ped.GetOccupiedVehicle()->IsInWater();
+                break;
+            }
+            else
+            {
+                bInWater = Ped.IsInWater();
+                break;
+            }
```
r12750, line 778: Have you tried to put the 'break' down here? What would you put your case there?

androksi: This is the pattern that this repository uses. I don't think that changing it would be a good choice. Clone this repo. to see how things work - it's an important decision to make when it comes to reviewing a PR.

patrikjuvonen (approved):

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Lpsd: not sure if this was the point being made - but it would be cleaner with a single break, infact the whole condition is perfectly readable as two lines:

```cpp
bInWater = Ped.GetOccupiedVehicle() ? Ped.GetOccupiedVehicle()->IsInWater() : Ped.IsInWater();
break;
```

down to preference I guess

PR #2566, opened 2022-03-06
Add getVehicleEntryPoints

### LINKED ISSUE #2565: Add a way to get the entry points to a vehicle

Right now the only ways to make a ped enter a vehicle doesn't allow you to pick a specific seat. The closest you can get is by using setPedEnterVehicle and specifying whether it has to be a driver or passenger seat. However you can still not decide which passenger seat the ped takes.

It would be nice to have a function that returns all entry points to a vehicle (note that entry points and seats are not exactly the same thing, a vehicle can have more entry points than seats). Having these entry points would make it possible to first make the ped walk to  the location of one of these entry points before using the setPedEnterVehicle function, thus allowing you to choose the resulting seat. This solution is also necessary for my Co-Op project (Sphene) to properly implement its task logic.

You could technically manually build a list of offsets of each entry points and use that but it would be a huge amount of work and won't work with custom vehicles.

MegadreamsBE:
Fixes #2565

This PR adds a new function called **getVehicleEntryPoints** which returns a table containing the positions to 4 possible entry points to a vehicle. Note that this does not directly relate to the amount of doors a vehicle has as vehicles with two doors can have multiple entry points to the same door. This usually influences the type of animation that is being used which is most prominent with bikes (stepping onto the bike vs jumping onto it).

This function can be used alongside setPedEnterVehicle to make a ped enter a specific seat by first moving the ped to a entry point retrieved through getVehicleEntryPoints and then using setPedEnterVehicle to make them enter. At this time setPedEnterVehicle gives you no such control.

The non-OOP version returns a table containing 4 tables itself which hold the x, y, z coordinates of each entry point. The OOP version returns a table containing 4 Vector3's instead. When no entry points exist for a vehicle it will return false.

Test resources for both the non-OOP as OOP version of this function are provided on the bottom of this PR. These test resources display the entry points for each vehicle near you as seen in the screenshot below:

proxy_sa_2022-03-06_15-58-21

[entrypoints.zip](https://github.com/multitheftauto/mtasa-blue/files/8193631/entrypoints.zip)
[entrypoints-oop.zip](https://github.com/multitheftauto/mtasa-blue/files/8193632/entrypoints-oop.zip)


Einheit-101:
are these points hardcoded or are they somehow affiliated with the seat or door dummy?

MegadreamsBE:
> are these points hardcoded or are they somehow affiliated with the seat or door dummy?

Looking at the underlying logic of GetPositionToOpenCarDoor which is used by the game to figure out what point to get the ped to before making it enter the vehicle (it also causes the endless going around the vehicle as it tries to line up per the ALIGN task) it does seem to calculate these from some of the vehicle's dummies. Unfortunately I can't say for certain what these exact dummies are and how it derives these entry points from it.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4966,3 +4966,31 @@ void CClientVehicle::ResetWheelScale()
...
+        case 1:
+            saDoorIndex = 8;
+            break;
+        case 2:
+            saDoorIndex = 11;
+            break;
+        case 3:
+            saDoorIndex = 9;
+            break;
+    }
```
Pirulax: Consider using a lambad here, eg.:
```cpp
const auto GetSADoorIndex = []() {
 // Switch case  here, and return directly
};
```

Much neater IMHO.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4148,3 +4150,45 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
 }
+
+std::variant<bool, std::vector<std::array<float, 3>>> CLuaVehicleDefs::GetVehicleEntryPoints(CClientVehicle* vehicle)
+{
+    if (CClientVehicleManager::GetMaxPassengerCount(vehicle->GetModel()) == 255)
+    {
+        return false;
+    }
+
+    std::vector<std::array<float, 3>> entryPoints;
```
Pirulax: If you already know the size of the vector, why use a vector?
```cpp
std::array<std::array<float, 3>, 4>
```
Would be much better.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4148,3 +4150,45 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+
+    for (unsigned int i = 0; i <= 3; i++)
+    {
+        CVector entryPoint;
```
Pirulax: IIRC I added a `ToArray` to `CVector`.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4148,3 +4150,45 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+}
+
+std::variant<bool, std::vector<CVector>> CLuaVehicleDefs::OOP_GetVehicleEntryPoints(CClientVehicle* vehicle)
+{
+    if (CClientVehicleManager::GetMaxPassengerCount(vehicle->GetModel()) == 255)
+    {
+        return false;
+    }
+
+    std::vector<CVector> entryPoints;
```
Pirulax: Same story as above, using `std::array<CVector, 4>` would be preferred.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4148,3 +4150,45 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+std::variant<bool, std::vector<std::array<float, 3>>> CLuaVehicleDefs::GetVehicleEntryPoints(CClientVehicle* vehicle)
+{
+    if (CClientVehicleManager::GetMaxPassengerCount(vehicle->GetModel()) == 255)
+    {
+        return false;
+    }
+
+    std::vector<std::array<float, 3>> entryPoints;
+
+    for (unsigned int i = 0; i <= 3; i++)
```
Pirulax: Use `i < 4` as the condition please, its really confusing.

Pirulax (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: Interestingly enough, that's what I usually do. No idea why I didn't in this case. I did take this piece of code from an earlier PR I had worked on so I past me is to blame for this. Will change it!

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: Doesn't seem like it.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4966,3 +4966,35 @@ void CClientVehicle::ResetWheelScale()
 
     m_bWheelScaleChanged = false;
 }
+
+void CClientVehicle::GetEntryPoint(CVector& entryPoint, int entryPointIndex)
+{
+    const auto GetSADoorIndex = [entryPointIndex]()
```
tederis: Why not capture **entryPointIndex** by reference here?

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -4966,3 +4966,35 @@ void CClientVehicle::ResetWheelScale()
...
+    {
+        switch (entryPointIndex)
+        {
+            case 0:
+                return 10;
+            case 1:
+                return 8;
+                break;
```
tederis: The **break** after the **return** is meaningless.

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -530,6 +530,8 @@ class CClientVehicle : public CClientStreamElement
     bool SetDummyPosition(eVehicleDummies dummy, const CVector& position);
     bool ResetDummyPositions();
 
+    void GetEntryPoint(CVector& entryPoint, int entryPointIndex);
```
tederis: It seems like it would be better to return CVector instead of passing it as an argument.

Fernando-A-Rocha:
Good idea! Is the code ready?

MegadreamsBE:
> Good idea! Is the code ready?

Yes, it's been ready for quite a while. Just waiting for this to be reviewed.

Client/game_sa/CCarEnterExitSA.cpp
```diff
@@ -78,6 +78,19 @@ bool CCarEnterExitSA::GetNearestCarPassengerDoor(CPed* pPed, CVehicle* pVehicle,
...
+{
+    CVehicleSA* vehicleSA = dynamic_cast<CVehicleSA*>(vehicle);
+
+    if (vehicleSA)
+    {
+        CVehicleSAInterface* vehicleInterface = vehicleSA->GetVehicleInterface();
+
+        auto CCarEnterExit_GetPositionToOpenCarDoor = (void(__cdecl*)(CVector&, CVehicleSAInterface*, int))FUNC_GetPositionToOpenCarDoor;
+        CCarEnterExit_GetPositionToOpenCarDoor(position, vehicleInterface, door);
+    }
```
TracerDS: Early return

Client/game_sa/CCarEnterExitSA.h
```diff
@@ -17,12 +17,14 @@
...
 #define FUNC_ComputeTargetDoorToExit            0x64F110
 #define FUNC_IsRoomForPedToLeaveCar             0x6504C0
+#define FUNC_GetPositionToOpenCarDoor           0x64E740
 
 class CCarEnterExitSA : public CCarEnterExit
 {
 public:
     bool GetNearestCarDoor(CPed* pPed, CVehicle* pVehicle, CVector* pVector, int* pDoor);
     bool GetNearestCarPassengerDoor(CPed* pPed, CVehicle* pVehicle, CVector* pVector, int* pDoor, bool bUnknown, bool bUnknown2, bool bCheckIfRoomToGetIn);
+    void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door);
```
TracerDS: Can const qualifier be applied here?

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,35 @@ void CClientVehicle::ResetWheelScale()
...
+            case 2:
+                return 11;
+            case 3:
+                return 9;
+        }
+
+        return 0;
+    };
+
+    unsigned int saDoorIndex = GetSADoorIndex();
```
TracerDS: Why bother creating a lambda? Just assign the value directly

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -543,6 +543,8 @@ class CClientVehicle : public CClientStreamElement
     bool SetDummyPosition(eVehicleDummies dummy, const CVector& position);
     bool ResetDummyPositions();
 
+    CVector GetEntryPoint(int entryPointIndex);
```
TracerDS: const qualifier

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,39 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
 }
+
+std::variant<bool, std::array<std::array<float, 3>, 4>> CLuaVehicleDefs::GetVehicleEntryPoints(CClientVehicle* vehicle)
+{
+    if (CClientVehicleManager::GetMaxPassengerCount(vehicle->GetModel()) == 255)
+    {
+        return false;
+    }
+
+    std::array<std::array<float, 3>, 4> entryPoints;
```
TracerDS, line 4183: CVector instead of float array might be a better fit

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,39 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+
+    for (unsigned int i = 0; i < 4; i++)
+    {
+        CVector entryPoint = vehicle->GetEntryPoint(i);
+
+        entryPoints[i] = {entryPoint.fX, entryPoint.fY, entryPoint.fZ};
+    }
```
TracerDS: Is it possible to use for-range loop instead?

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,39 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+
+    std::array<CVector, 4> entryPoints;
+
+    for (unsigned int i = 0; i < 4; i++)
+    {
+        entryPoints[i] = vehicle->GetEntryPoint(i);
+    }
+
+    return entryPoints;
```
TracerDS: The function is the same as `GetVehicleEntryPoints`.
Link it with that function instead of creating a new one with the same definition

TracerDS (changes requested):

Client/mods/deathmatch/logic/CClientVehicle.cpp
MegadreamsBE: Would be going in circles here. Earlier review from Pirulax in 2022 made me change from directly assigning to using the lambda.

TracerDS: > Would be going in circles here. Earlier review from Pirulax in 2022 made me change from directly assigning to using the lambda.

How so? This would be exactly the same thing
```cpp
std::uint32_t saDoorIndex = 0;

switch (entryPointIndex) {
    case 0:
        saDoorIndex = 10;
        break;
    case 1:
        saDoorIndex = 8;
        break;
    case 2:
        saDoorIndex = 11;
        break;
    case 3:
        saDoorIndex = 9;
        break;
}
```
There is no point in creating a lambda in this case

MegadreamsBE: I agree and will change it back. Just hope nobody comes around to make me change it again lol. This is how I originally had it.

Client/game_sa/CCarEnterExitSA.h
MegadreamsBE: On CVehicle? Not quite, it gets casted to CVehicleSA.

TracerDS: On the function itself

Client/mods/deathmatch/logic/CClientVehicle.cpp
tederis: I prefer using arrays:

```
uint32_t DOOR_LOOOKUP[4] = {
 10, 8, 11, 9
};

uint32_t saDoorIndex = DOOR_LOOOKUP[entryPointIndex];
```

TracerDS: This could be moved out of the function and marked constexpr then.

MegadreamsBE: I decided to stick to your first suggestion Tracer. Perfectly readable and acceptable.

MegadreamsBE:
Hmm, bit of an oversight I need to solve. Didn't consider that changing to CVector as per your request @TracerDS it would also make the non-OOP version return a vector which is not expected. Any suggestions on how to best solve that scenario?

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -240,6 +241,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getWheelScale", "getVehicleWheelScale");
     lua_classfunction(luaVM, "getModelWheelSize", "getVehicleModelWheelSize");
     lua_classfunction(luaVM, "getWheelFrictionState", "getVehicleWheelFrictionState");
+    lua_classfunction(luaVM, "getEntryPoints", ArgumentParser<GetVehicleEntryPoints>);
```
TracerDS: It should be
```cpp
lua_classfunction(luaVM, "getEntryPoints", "getVehicleEntryPoints");
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,21 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+
+    std::array<CVector, 4> entryPoints;
+
+    std::uint32_t i = 0;
+    for (auto& entryPoint : entryPoints)
+    {
+        entryPoint = vehicle->GetEntryPoint(i++);
+    }
+
+    return entryPoints;
```
TracerDS, line 4211: You said it doesnt return an array? If `std::array` is not supported then you could use `std::vector` instead

Client/sdk/game/CCarEnterExit.h
```diff
@@ -23,4 +23,6 @@ class CCarEnterExit
                                             bool bCheckIfRoomToGetIn) = 0;
     virtual int  ComputeTargetDoorToExit(CPed* pPed, CVehicle* pVehicle) = 0;
     virtual bool IsRoomForPedToLeaveCar(CVehicle* pVehicle, int iDoor, CVector* pUnknown = 0) = 0;
+
+    virtual const void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) = 0;
```
TracerDS: If its possible, could you mark this function as const?
```cpp
virtual const void GetPositionToOpenCarDoor(...) const = 0;
```
If the function doesnt throw, mark ot as `noexcept` too, please

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: I more so meant that if I return CVector here as part of the array then that's also exactly what I'm getting in MTA. However, that's not intended to be the case for a non-OOP function. That's why I used to have a separate function for the OOP version. Note that this was written years ago when things were still a bit different.

TracerDS: > I more so meant that if I return CVector here as part of the array then that's also exactly what I'm getting in MTA. However, that's not intended to be the case for a non-OOP function. That's why I used to have a separate function for the OOP version. Note that this was written years ago when things were still a bit different.

so what do you expect the POP function to return?

Client/sdk/game/CCarEnterExit.h
MegadreamsBE: Compiler doesn't like this, adding noexcept though.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: ArgumentParser is required here

TracerDS: > ArgumentParser is required here

its not. Look at the line above:
```cpp
lua_classfunction(luaVM, "getWheelFrictionState", "getVehicleWheelFrictionState");
```
`getVehicleWheelFrictionState` also has an ArgumentParser but for OOP only the function name is provided.

MegadreamsBE: Sure, but it's not possible in my use case as we cannot return vectors on non-OOP functions. Therefore I can't use the same function for both cases.

TracerDS: > Sure, but it's not possible in my use case as we cannot return vectors on non-OOP functions. Therefore I can't use the same function for both cases.

You can return vectors.

MegadreamsBE: > > Sure, but it's not possible in my use case as we cannot return vectors on non-OOP functions. Therefore I can't use the same function for both cases.
> 
> You can return vectors.

Different scenario - this is an array with vectors. I've ran more tests and the argument parser isn't mapping them correctly to an array of 3 floats that you'd expect from a non-OOP function. Modifying the argument parser to support this is out of the scope of this PR.

Client/game_sa/CCarEnterExitSA.h
```diff
@@ -17,12 +17,14 @@
...
 #define FUNC_ComputeTargetDoorToExit            0x64F110
 #define FUNC_IsRoomForPedToLeaveCar             0x6504C0
+#define FUNC_GetPositionToOpenCarDoor           0x64E740
 
 class CCarEnterExitSA : public CCarEnterExit
 {
 public:
     bool GetNearestCarDoor(CPed* pPed, CVehicle* pVehicle, CVector* pVector, int* pDoor);
     bool GetNearestCarPassengerDoor(CPed* pPed, CVehicle* pVehicle, CVector* pVector, int* pDoor, bool bUnknown, bool bUnknown2, bool bCheckIfRoomToGetIn);
+    const void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) noexcept;
```
tederis: Is it really needed to return const void here?

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,35 @@ void CClientVehicle::ResetWheelScale()
 
     m_bWheelScaleChanged = false;
 }
+
+const CVector CClientVehicle::GetEntryPoint(int entryPointIndex)
+{
+    std::uint32_t saDoorIndex = 0;
+
+    switch (entryPointIndex)
```
tederis: I would use a lookup table here. For example: `static const uint32_t lookup[4] = {10, 8, 11, 9}; assert(entryPointIndex < 4); const uint32_t saDoorIndex = lookup[entryPointIndex];`

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,35 @@ void CClientVehicle::ResetWheelScale()
 
     m_bWheelScaleChanged = false;
 }
+
+const CVector CClientVehicle::GetEntryPoint(int entryPointIndex)
```
tederis: Since `entryPointIndex` is non-negative it's better to choose `uint32_t` for it.

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -543,6 +543,8 @@ class CClientVehicle : public CClientStreamElement
     bool SetDummyPosition(eVehicleDummies dummy, const CVector& position);
     bool ResetDummyPositions();
 
+    const CVector GetEntryPoint(int entryPointIndex);
```
tederis: There is practically no sence in returning a const vector by value.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,43 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+{
+    if (CClientVehicleManager::GetMaxPassengerCount(vehicle->GetModel()) == 255)
+    {
+        return false;
+    }
+
+    std::array<CVector, 4> entryPoints;
+    std::uint32_t          i = 0;
+
+    for (auto& entryPoint : entryPoints)
```
tederis, line 4206: `for (uint32_t i{}; auto& entryPoint : entryPoints)` would be neater

Client/sdk/game/CCarEnterExit.h
```diff
@@ -23,4 +23,6 @@ class CCarEnterExit
                                             bool bCheckIfRoomToGetIn) = 0;
     virtual int  ComputeTargetDoorToExit(CPed* pPed, CVehicle* pVehicle) = 0;
     virtual bool IsRoomForPedToLeaveCar(CVehicle* pVehicle, int iDoor, CVector* pUnknown = 0) = 0;
+
+    virtual const void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) noexcept = 0;
```
tederis: `unsigned int door` -> `uint32_t door`

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,43 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+    if (std::holds_alternative<bool>(entryPointVectors))
+    {
+        return false;
+    }
+
+    std::array<std::array<float, 3>, 4> entryPoints;
+    std::array<CVector, 4>              vectorArray = std::get<std::array<CVector, 4>>(entryPointVectors);
+
+    std::uint32_t i = 0;
+    for (auto& entryPoint : entryPoints)
```
tederis, line 4187: `for (uint32_t i{}; auto& entryPoint : entryPoints)` would be neater

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TracerDS: > Different scenario - this is an array with vectors. I've ran more tests and the argument parser isn't mapping them correctly to an array of 3 floats that you'd expect from a non-OOP function. Modifying the argument parser to support this is out of the scope of this PR.

What about vector of vectors? (`std::vector<CVector>`)
And what does being non-oop function have to do with it?

tederis (changes requested):

Client/game_sa/CCarEnterExitSA.h
MegadreamsBE: Came from an earlier code review

Client/mods/deathmatch/logic/CClientVehicle.cpp
TracerDS: > Since `entryPointIndex` is non-negative it's better to choose `uint32_t` for it.

`std::uint32_t` to be exact. If we use C++ then lets use C++ headers (and thus `std` namespace)

Client/sdk/game/CCarEnterExit.h
TracerDS: `std::uint32_t`*

Client/game_sa/CCarEnterExitSA.h
tederis: But there is no sense in returning const void

TracerDS: > Came from an earlier code review

You probably mean my review about the const function? I meant the const as for the modifying function:
```cpp
void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) const noexcept;
```

tederis: Hmm. I think that the previous reviewer suggested to add a const qualifier this way: `void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) const noexcept;`

Client/mods/deathmatch/logic/CClientVehicle.h
tederis: I think that the previous reviewer suggested to add a const qualifier this way: `CVector GetEntryPoint(int entryPointIndex) const;`

MegadreamsBE: Gotcha, I'll make sure to update it

Client/game_sa/CCarEnterExitSA.h
MegadreamsBE: > > Came from an earlier code review
> 
> You probably mean my review about the const function? I meant the const as for the modifying function:
> 
> ```c++
> void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) const noexcept;
> ```

My bad! I'll get it updated.

Client/sdk/game/CCarEnterExit.h
tederis: What's wrong with uint32_t? std::uint32_t and uint32_t are the same.

Client/mods/deathmatch/logic/CClientVehicle.h
MegadreamsBE: Ended up not being happy with adding a const qualifier due to the call to GetGameVehicle inside. So ended up just removing const altogether.

Client/mods/deathmatch/logic/CClientVehicle.cpp
tederis: > > Since `entryPointIndex` is non-negative it's better to choose `uint32_t` for it.
> 
> `std::uint32_t` to be exact. If we use C++ then lets use C++ headers (and thus `std` namespace)

std::uint32_t and uint32_t are the same in all contexts. There is no need to use the cumbersome namespace prefix.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: Seems to require C++20 whereas the projects are defined as C++17. So will have to keep the current structure

Client/sdk/game/CCarEnterExit.h
```diff
@@ -23,4 +23,6 @@ class CCarEnterExit
                                             bool bCheckIfRoomToGetIn) = 0;
     virtual int  ComputeTargetDoorToExit(CPed* pPed, CVehicle* pVehicle) = 0;
     virtual bool IsRoomForPedToLeaveCar(CVehicle* pVehicle, int iDoor, CVector* pUnknown = 0) = 0;
+
+    virtual void GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, unsigned int door) const noexcept = 0;
```
tederis: You have changed `door` to `uint32_t` in `CCarEnterExitSA` but it's still `unsigned int` here.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,21 @@ void CClientVehicle::ResetWheelScale()
...
+
+CVector CClientVehicle::GetEntryPoint(std::uint32_t entryPointIndex)
+{
+    static const uint32_t lookup[4] = {10, 8, 11, 9};
+    assert(entryPointIndex < 4);
+    const uint32_t saDoorIndex = lookup[entryPointIndex];
+
+    CVector      entryPoint;
+
+    if (saDoorIndex != 0)
```
tederis: We're using the lookup table now. So `saDoorIndex` cannot not be zero in any case. This conditional can be removed with no regrets.

Client/game_sa/CCarEnterExitSA.cpp
```diff
@@ -78,6 +78,21 @@ bool CCarEnterExitSA::GetNearestCarPassengerDoor(CPed* pPed, CVehicle* pVehicle,
...
 }
 
+void CCarEnterExitSA::GetPositionToOpenCarDoor(CVector& position, CVehicle* vehicle, std::uint32_t door) const noexcept
+{
+    CVehicleSA* vehicleSA = dynamic_cast<CVehicleSA*>(vehicle);
+
+    if (vehicleSA == nullptr)
+    {
+        return;
+    }
```
TracerDS: `if(!vehicleSA) return;`

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,21 @@ void CClientVehicle::ResetWheelScale()
 
     m_bWheelScaleChanged = false;
 }
+
+CVector CClientVehicle::GetEntryPoint(std::uint32_t entryPointIndex)
+{
+    static const uint32_t lookup[4] = {10, 8, 11, 9};
```
TracerDS, line 5030: `std::uint32_t`

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -240,6 +241,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getWheelScale", "getVehicleWheelScale");
     lua_classfunction(luaVM, "getModelWheelSize", "getVehicleModelWheelSize");
     lua_classfunction(luaVM, "getWheelFrictionState", "getVehicleWheelFrictionState");
+    lua_classfunction(luaVM, "getEntryPoints", ArgumentParser<OOP_GetVehicleEntryPoints>);
```
TracerDS, line 244:
```cpp
lua_classfunction(luaVM, "getEntryPoints", "getVehicleEntryPoints");
```

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4168,3 +4170,43 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
 {
     return CStaticFunctionDefinitions::BlowVehicle(*entity, withExplosion);
 }
+
+std::variant<bool, std::array<std::array<float, 3>, 4>> CLuaVehicleDefs::GetVehicleEntryPoints(CClientVehicle* vehicle)
```
TracerDS, line 4174: Why is the POP different from OOP function? They should be the same

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.h
```diff
@@ -151,6 +151,9 @@ class CLuaVehicleDefs : public CLuaDefs
     static std::variant<bool, CVector>                              OOP_GetVehicleDummyPosition(CClientVehicle* vehicle, eVehicleDummies dummy);
     static bool                                                     ResetVehicleDummyPositions(CClientVehicle* vehicle);
 
+    static std::variant<bool, std::array<std::array<float, 3>, 4>>    GetVehicleEntryPoints(CClientVehicle* vehicle);
+    static std::variant<bool, std::array<CVector, 4>>                 OOP_GetVehicleEntryPoints(CClientVehicle* vehicle);
```
TracerDS, line 155: Whats the point of OOP function?

TracerDS (changes requested):

Client/sdk/game/CCarEnterExit.h
MegadreamsBE: Interesting, I must have accidentally undone it

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: I've tested this. Does not work with an array of vectors. Changing the argument parser to make this work is out of the scope of this PR.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.h
MegadreamsBE: To return the array of vectors, with the non-OOP version returning an array of floats. The argument parser cannot handle this case, therefore separate functions _are_ necessary.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: Not supported by the argument parser, returns OOP related values in non-OOP functions in-game. I've tested this scenario multiple times.

Client/sdk/game/CCarEnterExit.h
TracerDS: > What's wrong with uint32_t? std::uint32_t and uint32_t are the same.

We use C++ so lets stick to C++ headers. Thus, `std::` is mandatory. `cstdint` will include `uint` in the `std` namespace but may or may not include them in the global namespace and vice versa

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TracerDS: > I've tested this. Does not work with an array of vectors. Changing the argument parser to make this work is out of the scope of this PR.

It works fine for me for both methods
image
image

func1 - `std::vector<CVector>`
func2 - `std::vector<std::array<float, 3>>`

In the 2nd image I replaced `classfunction(...,ArgumentParser)` with `classfunction(..., "func1")`


Let me get your pr and check it

TracerDS: Array of vectors works fine for me. Do you have any compilation issues?

MegadreamsBE: No compilation issues, however non-OOP functions aren't supposed to return vectors at all. Those are OOP only.

TracerDS: > No compilation issues, however non-OOP functions aren't supposed to return vectors at all. Those are OOP only.

Whats the difference? OOP function should be the same as POP one except that its not registered in the global namespace but in a class.

Either use CVector or `std::array`. Keep it simple

TracerDS: If `CVector` is too slow then `std::array` seems like an obvious choice here

MegadreamsBE: It changes what is output in Lua, that's the point. OOP functions returns Vector3 objects for positions and non-OOP functions don't

TracerDS: > It changes what is output in Lua, that's the point. OOP functions returns Vector3 objects for positions and non-OOP functions don't

Thats what I mean. Either have them both return CVector or use `std::array`. Splitting them up into 2 dont make much sense especially when you'd want to get individual entry point locations.

MegadreamsBE: The difference is that the OOP function gives us a table with 4 Vector3 objects, one for each entrypoint. The POP version gives us a table that by itself has 4 tables (one for each entrypoint) that each contain 3 floats (the x, y & z positions).

If you can find a way to combine these functions, yet have those intended outputs, let me know. I can't find a way to do it.

Changing the POP version to be CVector will make it return a table with 4 Vector3 objects, just like the OOP version. That's not what a POP function is supposed to do in MTA. Similarly changing the OOP one to std:array with floats will make it so that it's not returning those Vector3 objects, which a OOP function is expected to do.

TracerDS: > That's not what a POP function is supposed to do in MTA

What is POP function supposed to do?

> Similarly changing the OOP one to std:array with floats will make it so that it's not returning those Vector3 objects, which a OOP function is expected to do.

Why is it expected to do so?

Lpsd: @MegadreamsBE is right, the method should be split (as it already is) so it can return the correct value between POP/OOP.

MegadreamsBE: > > That's not what a POP function is supposed to do in MTA
> 
> What is POP function supposed to do?

Vector3 (which CVector will result into thanks to the argument parser) is a OOP-only object. It should never be returned by POP functions. It's also why you don't see it being used outside of OOP functions.

> > Similarly changing the OOP one to std:array with floats will make it so that it's not returning those Vector3 objects, which a OOP function is expected to do.
> 
> Why is it expected to do so?

Because that's how the OOP functions in MTA work. It's an expectation for them to return Vector3 objects that can easily be manipulated using the methods defined here: https://wiki.multitheftauto.com/wiki/Vector/Vector3.

I'm merely keeping my functions consistent with the remainder of the code base.

TracerDS: If all OOP functions return Vector then its fine, nevermind my review then

MegadreamsBE: > If all OOP functions return Vector then its fine, nevermind my review then

Jup, that's consistently the case.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -5023,3 +5024,17 @@ void CClientVehicle::ResetWheelScale()
...
```
TracerDS, line 5030: you use `std::uint32_t` everywhere else so please use it here too

TracerDS (approved):

Client/mods/deathmatch/logic/CClientVehicle.cpp
Dutchman101: This is minor, can be done through a follow up commit

MegadreamsBE: Spoiler: It'll never happen

PR #2575, opened 2022-03-19
Add missing world model names and IDs

Zangomangu:
Adds missing world models to `engineGetModelNameFromID` / `engineGetModelIDFromName`
The new ones are 99.99% LODs


patrikjuvonen:
Could you clarify also how you exported these names and from where and is there a way we can verify these?

Zangomangu:
Used a script to go through all IPL files in data/maps and then output the `inst` ID/name combos, where the ID didnt exist in the MTA table `bigFOTable`

For verifying maybe you can cross reference some of them here https://wiki.multitheftauto.com/wiki/IDE_List to see if the ID and names match (assuming the wiki list is complete)

Merlin:
There are a few entries that dont exist:
```
6020, "vegasxrexhse09",
6021, "vegasxrexhse03",
6022, "vegasxrexhse04",
6023, "vegasxrexhse05",
6024, "vegasxrexhse06",
6025, "vegasxrexhse07",
6026, "vegasxrexhse08",
6027, "vegasxrexhse10",
```
and a bunch more

Scraping them from only .ide and not .ipl would be the better approach. 
Because vegasxref.ipl for example is never loaded and has wrong ids for the models.

Looks like the original list also has a lot of duplicates with alternative names:
```
1733, "BurntWashingMachine",
1773, "BurntStove1",
1777, "BurntStove2",
```

where the actual names are:
```
1733, "CJ_WASHINGMAC",
1773, "CJ_COOKER2",
1777, "CJ_COOKER1",
```

There are also vehicle names added to the list not the actual model name.
For example id 497 is called "Police Maverick" where the actual model name would be "polmav"

Same with the weapon models.
`321, "Regular_Dildo",`

But the original model name is
`321, "gun_dildo1",`

Also contains a few mispellings:
`16562, "cn2_rosm/ot1",`
should be
`16562, "cn2_rosmot1",`

The list also doesn't contains any peds.

If there is interest i have a complete list from all ids and modelnames and we would only need to add all these edge cases for backwards compatibility

Zangomangu:
I used the ipls here out of laziness because I had a merged one laying around.
And specifically only added missing objects, not peds/vehicles/weapons.

All these errors you mention in the original list will have to stay there for backwards comp, no one's gonna touch that

Could you open a PR with the .ide scraped ids/models and we can close this one?

Zangomangu:
Closed in favor of #2592

PR #2499, opened 2022-01-05
Remove leftover Discord implementation

Lpsd:
Fixes #1370

Removes the leftover Discord implementation - aside from CServerInfo / CQueryReceiver stuff introduced in the original PR, which is actually a benefit.

PR #2577, opened 2022-03-22
Add discord-rpc

### LINKED ISSUE #2579: Add Discord RPC
**Add Discord rich response**

please add discord rich response function 

That we can leave one rich response for each player

Capture

This is a sample photo

Deihim007:
Based on #2499 
Prior discussions/implementations #721, #1031, #1570

This PR adds basic discord-rpc features to the client.
(Other features will be added later on a separate PR)

Todo:
- [x] Todolist
- [x] Add discord-rpc
  - [x] Installation script
  - [x] dependencies (rapidjson)
- [x] State:
  - [x] Main Menu
  - [x] In-game
    - [x] Elapsed time

Maybe:
- [ ] Default game-mode implementation
  - [ ] Idle (Custom state)
  - [ ] Loading
  - [ ] Player count ?
- [ ] Custom Application Id
  - [ ] Server-side function
  - [ ] Client settings option to disable it (prevent potential abuse)
  - [ ] Wiki tutorial
    - [ ] How to make an application on discord developer portal
      - [ ] Assets template ?


Client/sdk/core/CDiscordInterface.h
```diff
@@ -0,0 +1,19 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
```
CrosRoad95, line 3: we are no longer mta 1.0 :)

Client/core/CCore.cpp
```diff
@@ -136,12 +136,18 @@ CCore::CCore()
 
     // Create tray icon
     m_pTrayIcon = new CTrayIcon();
+
+    // Create discord rich presence
+    m_pDiscordRichPresence = new CDiscordRichPresence();
```
tederis: Why not use std::unique_ptr instead of a raw ptr?

Client/core/CCore.h
```diff
@@ -101,6 +102,7 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
     CLocalizationInterface*   GetLocalization() { return g_pLocalization; };
     CWebCoreInterface*        GetWebCore();
     CTrayIconInterface*       GetTrayIcon() { return m_pTrayIcon; };
+    CDiscordInterface*        GetDiscord() { return m_pDiscordRichPresence; };
```
tederis: This method definitely should be **const**, because we just return a value.

Client/core/CDiscordRichPresence.cpp
```diff
@@ -0,0 +1,73 @@
...
+    m_uiDiscordAppEnd = 0;
+
+    Discord_Initialize(m_strDiscordAppId, &handlers, 1, nullptr);
+}
+
+CDiscordRichPresence::~CDiscordRichPresence()
+{
+    m_strDiscordAppId = DEFAULT_APP_ID;
+    Discord_Shutdown();
+    DiscordEventHandlers handlers;
```
tederis: It seems like **handlers** is never used in the destructor.

Client/core/CCore.cpp
Deihim007: https://stackoverflow.com/questions/6675651/when-should-i-use-raw-pointers-over-smart-pointers

Client/core/CDiscordRichPresence.cpp
Deihim007: That has been present in all the examples I've seen. I'm not sure about removing it.

Client/sdk/core/CDiscordInterface.h
Deihim007: tell that to other files :)

Client/core/CCore.h
Deihim007: I'm not sure if that's the case here. If so same applies for m_pTrayIcon

Client/core/CCore.cpp
tederis: But that is the case of ownership and we should use a smart pointer here. Isn't that right?

botder: Use a smart pointer here: `std::unique_ptr` to automatically run the destructor and free the memory.

Client/core/CCore.cpp
```diff
@@ -136,12 +136,18 @@ CCore::CCore()
...
 }
 
 CCore::~CCore()
 {
     WriteDebugEvent("CCore::~CCore");
 
+    // Destroy discord rich presence
+    delete m_pDiscordRichPresence;
```
botder: Once the smart pointer is introduced, use `.reset()` here

Client/core/CCore.cpp
```diff
@@ -586,6 +592,24 @@ void CCore::SetConnected(bool bConnected)
...
+        g_pCore->GetDiscord()->SetPresenceDetails("");
+        g_pCore->GetDiscord()->SetPresenceStartTimestamp((long)timer);
+        g_pCore->GetDiscord()->UpdatePresence();
+    } else
+    {
+        g_pCore->GetDiscord()->SetPresenceState("Main Menu");
+        g_pCore->GetDiscord()->SetPresenceDetails("");
+        g_pCore->GetDiscord()->SetPresenceStartTimestamp(0);
+        g_pCore->GetDiscord()->UpdatePresence();
+    }
```
botder:
```suggestion
    CDiscordInterface* discord = g_pCore->GetDiscord();
    discord->SetPresenceState(bConnected ? "In-game" : "Main menu");
    discord->SetPresenceStartTimestamp(0);
    discord->SetPresenceDetails("");

    if (bConnected)
    {
        time_t timer;
        time(&timer);
        discord->SetPresenceStartTimestamp((long)timer);
    }

    discord->UpdatePresence();
```

Client/core/CCore.h
```diff
@@ -39,6 +39,7 @@ class CCore;
 #include <ijsify.h>
 #include <core/CWebCoreInterface.h>
 #include "CTrayIcon.h"
+#include "CDiscordRichPresence.h"
```
botder: Use class forwarding to avoid pulling in the entire header file in here.

Client/core/CCore.h
botder: If you start using the smart pointer and class forwarding here, and it gives you an error about casting: Move the method definition to the cpp file.

Client/core/CDiscordRichPresence.cpp
```diff
@@ -0,0 +1,73 @@
...
+    memset(&handlers, 0, sizeof(handlers));
```
botder: This has no effect, remove it.

Client/core/CDiscordRichPresence.cpp
```diff
@@ -0,0 +1,73 @@
...
+    m_strDiscordAppDetails = "";
+    m_uiDiscordAppStart = 0;
+    m_uiDiscordAppEnd = 0;
+
+    Discord_Initialize(m_strDiscordAppId, &handlers, 1, nullptr);
+}
+
+CDiscordRichPresence::~CDiscordRichPresence()
+{
+    m_strDiscordAppId = DEFAULT_APP_ID;
```
botder: What's the reasoning for this in a destructor? It appears to be useless here.

Client/core/CDiscordRichPresence.h
```diff
@@ -0,0 +1,43 @@
...
+private:
+    const char* m_strDiscordAppId;
+    const char* m_strDiscordAppAsset;
+    const char* m_strDiscordAppAssetSmall;
+    const char* m_strDiscordAppAssetText;
+    const char* m_strDiscordAppAssetSmallText;
+
+    const char*   m_strDiscordAppCurrentId;
+    const char*   m_strDiscordAppState;
+    const char*   m_strDiscordAppDetails;
```
botder: Please don't store pointers to characters if you can not guarantee the memory behind these pointer will remain always valid for the entire lifetime. This is a trap for other developers if they are not aware of this design choice. Use `std::string`.

Client/core/CMainMenu.cpp
```diff
@@ -286,6 +286,10 @@ CMainMenu::CMainMenu(CGUI* pManager)
     // We're not ingame
     SetIsIngame(false);
 
+    // Discord
+    g_pCore->GetDiscord()->SetPresenceState("Main Menu");
+    g_pCore->GetDiscord()->UpdatePresence();
```
botder: Don't you have to reset the `SetPresenceStartTimestamp` to 0 here?

Client/sdk/core/CDiscordInterface.h
```diff
@@ -0,0 +1,20 @@
...
+ *               (Shared logic for modifications)
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        sdk/core/CDiscordInterface.h
+ *  PURPOSE:     Discord interface class
+ *
+ *****************************************************************************/
+
+#pragma once
+
+class CDiscordInterface
```
botder, line 13: Add a virtual destructor for every class that is supposed to be inherited from. There are exceptions to this rule, but the overhead of the virtual dispatch is less than the headaches for any developer down the road.

utils/buildactions/install_discord.lua
```diff
@@ -0,0 +1,250 @@
...
+	-- Move all files from rapidjson*/* to ./
+	os.expanddir_wildcard(RAPID_PATH.."rapidjson*", RAPID_PATH)
+end
+
+newaction {
+	trigger = "install_discord",
+	description = "Downloads and installs discord-rpc",
+
+	execute = function(...)
+		local upgrade = _ARGS[1] == "upgrade"
```
botder: Please rename `upgrade` to `should_upgrade` or `is_upgrade_requested` for readability.

vendor/discord-rpc/premake5.lua
```diff
@@ -0,0 +1,23 @@
...
+	files {
+		"premake5.lua",
+		"discord/src/discord_rpc.cpp",
+		"discord/src/rpc_connection.cpp",
+		"discord/src/serialization.cpp",
+		"discord/src/connection_win.cpp",
+		"discord/src/discord_register_win.cpp",
+	}
+
+	filter "architecture:x64"
```
botder: Exclude for anything that is not x86 (excludes ARM, ARM64 and x64).
```suggestion
	filter "architecture:not x86"
```

botder (changes requested):

Client/core/CDiscordRichPresence.cpp
Deihim007: Done

Client/core/CDiscordRichPresence.cpp
Deihim007: Removed for now, It was meant to reset app id if it was changed in the process. (todo: set app id)

Client/core/CCore.cpp
Lpsd: Now using `shared_ptr`

Fernando-A-Rocha:
I think we can prioritize making customization features for servers to decide what shows on their players' Discord Rich Presence when in-game. Imo, the PR can be merged once this is complete.

Then, allowing servers to integrate custom applications to their servers is something that can be addressed separately.

JrSchmidtt:
@botder  this feature been discontinued?
It is a very interesting feature to increase the player base.

Dutchman101:
We need someone to adopt this PR
Removed draft status, as todo list has been fully checkmarked, so that should have been outdated

Lpsd:
Superseded by #3167

PR #2586, opened 2022-04-02
Add more rope related features

### LINKED ISSUE #2488: Add ropes
In the vanilla game there is a rope system already implemented. It is used with magnets on the cranes and with a Leviathan on one mission: https://youtu.be/hl48M4Asiqo?t=45

It should be possible to add this functionality into MTA, and it should be faster than the lua alternatives you can create.

Add a function to create a rope.

None. I am not sure how difficult would be to implement this, I am not even sure if it's possible.

**Additional context**
- https://youtu.be/RaZIaInVrN4?t=68
- https://youtu.be/hl48M4Asiqo?t=45
- https://bugs.mtasa.com/view.php?id=6591

gta191977649:
This PR attempt to implement some additional roped related feature in CObjects:

- `0796: GET_ROPE_HEIGHT_FOR_OBJECT` - [Object](https://library.sannybuilder.com/#/sa/classes/Object).GetRopeHeight(self: Object, var height: float)
- `0797: SET_ROPE_HEIGHT_FOR_OBJECT` - [Object](https://library.sannybuilder.com/#/sa/classes/Object).SetRopeHeight(self: Object, height: float)

working in progress

### DIFF
--- a/Client/game_sa/CObjectSA.cpp
+++ b/Client/game_sa/CObjectSA.cpp
@@ -407,6 +407,12 @@ float CObjectSA::GetRopeHeight()
     return CRopesSA::GetRope(ropeIndex).m_fSegmentLength;
 }
 
+void CObjectSA::SetRopeHeight(float height)
+{
+    const auto ropeIndex = CRopesSA::FindRope(static_cast<CObjectSAInterface*>(this->GetInterface()));
+    CRopesSA::GetRope(ropeIndex).m_fSegmentLength = height;
+}
+
 CVector* CObjectSA::GetScale()
 {
     return &m_vecScale;
--- a/Client/game_sa/CObjectSA.cpp
+++ b/Client/game_sa/CObjectSA.cpp
@@ -401,6 +401,12 @@ void CObjectSA::SetScale(float fX, float fY, float fZ)
     GetObjectInterface()->fScale = std::max(fX, std::max(fY, fZ));
 }
 
+float CObjectSA::GetRopeHeight()
+{
+    const auto ropeIndex = CRopesSA::FindRope(static_cast<CObjectSAInterface*>(this->GetInterface()));
+    return CRopesSA::GetRope(ropeIndex).m_fSegmentLength;
+}
+
 CVector* CObjectSA::GetScale()
 {
     return &m_vecScale;
--- a/Client/game_sa/CObjectSA.h
+++ b/Client/game_sa/CObjectSA.h
@@ -155,6 +155,9 @@ public:
     CVector* GetScale();
     void     ResetScale();
 
+    float GetRopeHeight();
+    void  SetRopeHeight(float height);
+
 private:
     void CheckForGangTag();
 };
--- a/Client/game_sa/CRopesSA.cpp
+++ b/Client/game_sa/CRopesSA.cpp
@@ -10,9 +10,9 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#define MAX_NUM_ROPES 8
 DWORD dwDurationAddress = 0x558D1El;
-
-CRopesSAInterface (&CRopesSA::ms_aRopes)[8] = *(CRopesSAInterface(*)[8])0xB768B8;
+CRopesSAInterface (&CRopesSA::ms_aRopes)[MAX_NUM_ROPES] = *(CRopesSAInterface(*)[MAX_NUM_ROPES])0xB768B8;
 
 int CRopesSA::CreateRopeForSwatPed(const CVector& vecPosition, DWORD dwDuration)
 {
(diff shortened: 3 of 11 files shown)


lopezloo:
You might be interested:
https://github.com/lopezloo/mtasa-blue/tree/ropes
https://github.com/lopezloo/mtasa-blue/commits/ropes?author=lopezloo
https://github.com/lopezloo/mtasa-blue/blob/ropes/MTA10/game_sa/CRopeSA.h
https://github.com/lopezloo/mtasa-blue/blob/ropes/MTA10/game_sa/CRopeSA.cpp
https://github.com/lopezloo/mtasa-blue/blob/ropes/MTA10/game_sa/CRopesSA.h
https://github.com/lopezloo/mtasa-blue/blob/ropes/MTA10/game_sa/CRopesSA.cpp

I have implemented ropes in this branch. I'm not sure though if there is point in adding this to MTA since it can be implemented in Lua.

gta191977649:
now the `GetRopeHeight` and `SetRopeHeight` should work
the reset is to figure out how does the height exactly works, seems setting it to 1 will cause the magnet to fall off. 
NOTE: test it with object `1382`, other object id seems will crash the client, I haven't made the check in code yet.

gta191977649:
> You might be interested: https://github.com/lopezloo/mtasa-blue/tree/ropes https://github.com/lopezloo/mtasa-blue/commits/ropes?author=lopezloo
> 
> I have implemented ropes in this branch. I'm not sure though if there is point in adding this to MTA since it can be implemented in Lua.



> You might be interested: https://github.com/lopezloo/mtasa-blue/tree/ropes https://github.com/lopezloo/mtasa-blue/commits/ropes?author=lopezloo
> 
> I have implemented ropes in this branch. I'm not sure though if there is point in adding this to MTA since it can be implemented in Lua.

maybe for someone who wants to re-create the crane, like in single-player, I know it can be done in pain Lua, but this is the vanllia sa methods :p

Client/game_sa/CObjectSA.cpp
```diff
@@ -401,6 +401,18 @@ void CObjectSA::SetScale(float fX, float fY, float fZ)
     GetObjectInterface()->fScale = std::max(fX, std::max(fY, fZ));
 }
 
+float CObjectSA::GetRopeHeight()
+{
+    const auto ropeIndex = CRopesSA::FindRope(static_cast<CObjectSAInterface*>(this->GetInterface()));
```
tederis, line 406: What is the practical sense of explicitly use the **this->**?

Client/game_sa/CObjectSA.h
```diff
@@ -155,6 +155,9 @@ class CObjectSA : public virtual CObject, public virtual CPhysicalSA
     CVector* GetScale();
     void     ResetScale();
 
+    float GetRopeHeight();
```
tederis, line 158: The method GetRopeHeight doesn't change anything in the class, right? The similar methods should be marked as **const**.

Client/game_sa/CRopesSA.cpp
```diff
@@ -52,3 +52,20 @@ void CRopesSA::RemoveEntityRope(CEntitySAInterface* pEntity)
...
+{
+    for (auto ropeId = 0; ropeId < MAX_NUM_ROPES; ropeId++)
+    {
+        if (ms_aRopes[ropeId].m_ucRopeType != eRopeType::NONE && ms_aRopes[ropeId].m_pRopeEntity == id)
+            return ropeId;
+    }
+    return -1;
+}
+
+CRopesSAInterface& CRopesSA::GetRope(int32 index)
```
tederis, line 66: What happens if the index becomes negative? All the function's paths must return a value. Otherwise we have an undefined behaviour, that's not a good thing.

Client/sdk/game/CObject.h
```diff
@@ -41,4 +41,6 @@ class CObject : public virtual CPhysical
     virtual void     SetScale(float fX, float fY, float fZ) = 0;
     virtual CVector* GetScale() = 0;
     virtual void     ResetScale() = 0;
+    virtual float GetRopeHeight() = 0;
```
tederis, line 44: Similarly, the method should be const.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -3900,6 +3900,22 @@ bool CStaticFunctionDefinitions::GetObjectElasticity(CClientObject& Object, floa
     return true;
 }
 
+bool CStaticFunctionDefinitions::GetRopeHeight(CClientObject& Object, float& height)
```
tederis, line 3903: The first argument should be a const reference. Just because we want to get a value, but not change anything inside **Object**.

PR #2589, opened 2022-04-02
Remove async tasks in SVG and update callback usage

### LINKED ISSUE #2521: Remove async task usage and callbacks from SVG functions
After some discussion on Discord (due to the points mentioned in #2477) I'll be removing async task usage and callbacks from SVG functions entirely.

D3D9 stuff is not thread safe, XML stuff is not thread safe (in this case, even with #2477 , destructors still matter), and more...

Callbacks are useless for the SVG functions if we aren't doing stuff async.

SVG was introduced in the 1.5.9 release. Obviously, any scripts using the callback args will not work after this has been done (hard error in Lua due to "extra" arg supplied). However, I'm actually fine with making these changes (soon) into nightly, before 1.6 - I don't think many people are using SVG right now - and considering SVG is currently pretty unstable in some situations with callbacks, I don't see an issue with resolving that for the sake of minor bwc concerns.

Lpsd:
This PR removes all async tasks in SVG code, due to the thread-safety mentioned in #2521 (this PR resolves #2521).

All Lua functions arguments remain the same and these changes should be fully backwards compatible.

Each SVG can have a single "update callback" which is called whenever the SVG texture has been updated/changed.

The callback function parameter has changed, it is now the SVG that has been updated (see example below)

Here's an example of the changes in action:

```lua
local rawSvgData =
    [[
    <svg viewBox="0 0 500 500" xmlns="http://www.w3.org/2000/svg">
        <circle cx="250" cy="250" r="250" fill="#0fc0fc" />
    </svg>
]]

local myCircleSvg
local firstUpdate = false

local function renderMyCircle()
    if (myCircleSvg) then
        local width, height = svgGetSize(myCircleSvg)
        dxDrawImage(0, 0, width, height, myCircleSvg, 0, 0, 0, tocolor(255, 255, 255, 255))
    end
end

local function onUpdate(svg)
    iprint("svg updated (onUpdate)", svg, getTickCount())

    if (not firstUpdate) then
        -- Start drawing our SVG
        addEventHandler("onClientRender", root, renderMyCircle)

        -- Update to a new callback
        svgSetUpdateCallback(
            svg,
            function()
                iprint("svg updated (anonymous function)", svg, getTickCount())
            end
        )

        -- Resize svg (which will trigger the new update callback once complete)
        svgSetSize(svg, 250, 250)

        -- Note: both svgSetSize and svgSetDocumentXML have an argument for the update callback - providing one there is the same as calling svgSetUpdateCallback (it will overwrite the current callback)
        -- e.g: svgSetSize(svg, 250, 250, onUpdate), svgSetDocumentXML(svg, svgXML, onUpdate)

        firstUpdate = true
    end
end

local function init()
    myCircleSvg = svgCreate(500, 500, rawSvgData, onUpdate)
    iprint("svg created", myCircleSvg, getTickCount())
end
```

The two new functions which have been added:

```lua
bool svgSetUpdateCallback(svg-element, func)
bool/func svgGetUpdateCallback(svg-element)
```

This PR also clears up some of the existing code (variable naming and such).

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVectorGraphic.h
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphic.h
@@ -43,6 +43,12 @@ public:
 
     bool SetDocument(CXMLNode* node);
 
+    bool                                SetUpdateCallback(CLuaFunctionRef& funcRef);
+    std::variant<CLuaFunctionRef, bool> GetUpdateCallback() const { return m_updateCallbackRef; }
+    bool                                RemoveUpdateCallback();
+
+    void OnUpdate();
+
 private:
     CResource*      m_pResource = nullptr;
     CClientManager* m_pManager;
@@ -52,6 +58,8 @@ private:
     CXMLNode*                          m_pXMLDocument = nullptr;
 
     std::unique_ptr<CClientVectorGraphicDisplay> m_pVectorGraphicDisplay;
+
+    std::variant<CLuaFunctionRef, bool> m_updateCallbackRef = false;
 
     bool m_bIsDestroyed = false;
 };
--- a/Client/mods/deathmatch/logic/CClientVectorGraphic.cpp
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphic.cpp
@@ -60,6 +60,44 @@ bool CClientVectorGraphic::SetDocument(CXMLNode* node)
     return true;
 }
 
+bool CClientVectorGraphic::SetUpdateCallback(CLuaFunctionRef& funcRef)
+{
+    if (!VERIFY_FUNCTION(funcRef))
+        return false;
+
+    m_updateCallbackRef = funcRef;
+    return true;
+}
+
+bool CClientVectorGraphic::RemoveUpdateCallback()
+{
+    if (!std::holds_alternative<CLuaFunctionRef>(m_updateCallbackRef))
+        return false;
+
+    m_updateCallbackRef = false;
+    return true;
+}
+
+void CClientVectorGraphic::OnUpdate()
+{
+    m_pVectorGraphicDisplay->UpdateTexture();
+
+    if (std::holds_alternative<CLuaFunctionRef>(m_updateCallbackRef))
+    {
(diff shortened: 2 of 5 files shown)


Dutchman101:
The amount of instability (as in: crashes) that SVG implementation brought to MTA warrants this to be merged quicker than usual, having less eyes on it with code review will be compensated by test runs and nightly channel roll-out with metrics and stability monitoring.

Just be aware, if you have spare time test the latest nightly when it's built and report any new SVG bugs

Client/mods/deathmatch/logic/CClientVectorGraphic.h
```diff
@@ -53,5 +59,7 @@ class CClientVectorGraphic final : public CClientTexture
 
     std::unique_ptr<CClientVectorGraphicDisplay> m_pVectorGraphicDisplay;
 
+    std::variant<CLuaFunctionRef, bool> m_updateCallbackRef = false;
```
sbx320, line 62: std::optional? You don't seem to be using = true

Lpsd: brain fart, will update

Lpsd:
as mentioned on Discord it appears you can't use `CLuaFunctionRef` with `std::optional` due to:

```
T in optional<T> must meet the Cpp17Destructible requirements (N4828 [optional.optional]/3)
```
 
so I think it's fine like this for now.

also seems like I forgot about adding `svgGetUpdateCallback` to the Lua definitions, but funnily enough I guess I forgot to test that as it doesn't work anyway... scripts can just keep track of that themselves, I can't really see it being used much... I'll submit a PR to clean up the unused code.

Pirulax:
Skip the standard mumblejumble, what's the actual issue? The error?

Lpsd:
> Skip the standard mumblejumble, what's the actual issue? The error?

read the code, test it and then refer to what I wrote - if you want to delve deeper into the actual issue with CLuaFunctionRef and std::optional then feel free, but do that in another PR - it's outside the scope of this one.

PR #2592, opened 2022-04-04
Add missing model ids and names scraped from .ide files

Merlin:
Adds all the missing model ids and names.

ids and names come from .ide files


Client/mods/deathmatch/logic/CModelNames.cpp
```diff
@@ -10017,6 +10017,4421 @@ struct
...
+    9161,  "lodasEroad138",
+    10279, "LODstu69c_sfe",
+    6241,  "LODcanal_floor2",
+    10518, "LOD_sfs055",
+    18511, "LODcunts12",
+    10331, "LODgroundbit_12SFS",
+    3725,  "LODhillhouse06_LA",
+    7255,  "lodasNedge22",
+    8892,  "lodbond03_lvs",
+//  16562, "cn2_rosmot1",
```
Zangomangu: Could you add some info to the ones you commented out eg. "Wrong name retained for BC"

Pirulax:
btw...  Isn't all this info available run-time in `CModelInfo`?

Zangomangu:
No idea
If it is let's retrieve it from there instead?

Merlin:
This list resolves model name -> id and vice versa.

CBaseModelInfo only has a hash of the name not the actual name.
And CModelInfo::GetModelInfo(0x4C5940) hashes the name and goes over CModelInfo::ms_modelInfoPtrs and checks the hash to get you the id/modelinfo.
But there is no way to get the name from the id.

Zangomangu (approved):

Zangomangu:
Can be merged

botder (approved):

PR #2593, opened 2022-04-08
Fix failing build since 8754bbb (CEF update)

Lpsd:
See https://bitbucket.org/chromiumembedded/cef/commits/28c7f040016a


Dutchman101 (approved):
Thanks for fixing 8754bbb, it was discussed on dev discord (#home) and to merge this it was further confirmed that MTA doesn't invoke CEF plugins in any way.

PR #2594, opened 2022-04-09
Fix build failing since 66a16ff (unrar update)

Lpsd:
Starting with updating premake file to ensure the macros are defined properly (as seen in `os.hpp`)

I'm pretty sure that's not going to fix it, but I'll use this PR in order to debug the issue as I don't have a Mac / Linux system to hand with a build environment for MTA.


Lpsd:
seems kinda obvious now that `isnt.cpp` should only be included on Windows platforms. updated premake and trying again

PR #2595, opened 2022-04-09
Add notes about svgGetUpdateCallback

Lpsd:
It seems I forgot to add this function to the Lua definitions in #2589 - but it doesn't work anyway 😅 

Just commenting it out here and adding notes in case anyone else wonders why you can't grab the callback from SVG on Lua side.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVectorGraphicDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVectorGraphicDefs.cpp
@@ -20,7 +20,7 @@ void CLuaVectorGraphicDefs::LoadFunctions()
         {"svgGetSize", ArgumentParser<SVGGetSize>},
         {"svgSetSize", ArgumentParser<SVGSetSize>},
         {"svgSetUpdateCallback", ArgumentParser<SVGSetUpdateCallback>},
-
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaVectorGraphicDefs.cpp
```diff
@@ -257,12 +257,16 @@ bool CLuaVectorGraphicDefs::SVGSetUpdateCallback(CClientVectorGraphic* vectorGra
...
-std::variant<CLuaFunctionRef, bool> CLuaVectorGraphicDefs::SVGGetUpdateCallback(CClientVectorGraphic* vectorGraphic)
-{
-    auto func = vectorGraphic->GetUpdateCallback();
-
-    if (std::holds_alternative<CLuaFunctionRef>(func))
-        return func;
-
-    return false;
-}
+// This doesn't work, there is no conversion from CLuaFunctionRef to the Lua side, so the parser always uses bool.
```
Pirulax, line 260: Addint an overload to `lua::Push` should be possible.
`CLuaFunctionRef` just holds a `lua_ref` to the function.
I can't remember the actual name of the function, but you can find it IIRC in `CLuaArguments::CallGlobal`.

patrikjuvonen:
This pull request shall be the documentation for if anyone wants to add it in the future.
Let's not add unused/commented out code into the already bloated codebase.

Thanks for looking into it.

PR #2620, opened 2022-05-17
Consider only affected players for element data stats

### LINKED ISSUE #2619: setElementData's "subscribe" mode does seems to be sending the data to every player
### Describe the bug

Seeing the ``Event Packet usage`` using ``performancebrowser`` shows that element data that has been set with the ``suscribe`` mode will still try to get synced to every player on the server, even if it does not have any suscriber.

I have noticed this on my own server. I had some elementdata being synced only to a few suscribed players, but on the ``performancebrowser`` they showed as being shared to every player on the server. Because of that, I made a test script:
[testSubscribe.zip](https://github.com/multitheftauto/mtasa-blue/files/8703278/testSubscribe.zip)
 
 The test script will start setting two element datas every second for an object. One key is called ``data-suscribe`` which is trying to be read by the client and shown on screen (which will always be seen as ``false`` because it's not synced). The other key is a random number from 1-1000.
 
 

### Steps to reproduce

1. Start ``testSuscribe`` or make an script that changes an element's data with ``suscribe`` mode
2. Check that the data is indeed not changing clientside
3. Open ``performancebrowser`` and go to ``Event Packet usage``
4. You will see that the server is sending updates on the element's data even tho there aren't any suscribers

unknown (14)

### Version

Server: v1.5.9 r21022 WINDOWS | v1.5.9 r21203 LINUX
Client: v1.5.9 r21210

### Additional context

The packets could be shown as being sent but in reality they aren't really tried to be sent. Eitherway, this should be fixed.

Also, element datas that have been set using ``false`` mode will NOT show on the ``Event Packet usage``, implying that they are, indeed, not synced.

### Relevant log output

_No response_

### Security Policy

tederis:
Fix for the issue #2619. Affected players count values for methods BroadcastOnlyJoined, BroadcastDimensionOnlyJoined & BroadcastOnlySubscribed were also added for the sake of consistency and later use.

### DIFF
--- a/Server/mods/deathmatch/logic/CPlayerManager.cpp
+++ b/Server/mods/deathmatch/logic/CPlayerManager.cpp
@@ -145,7 +145,7 @@ void CPlayerManager::DeleteAll()
         delete *m_Players.begin();
 }
 
-void CPlayerManager::BroadcastOnlyJoined(const CPacket& Packet, CPlayer* pSkip)
+size_t CPlayerManager::BroadcastOnlyJoined(const CPacket& Packet, CPlayer* pSkip)
 {
     // Make a list of players to send this packet to
     CSendList sendList;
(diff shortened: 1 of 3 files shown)


PlatinMTA:
Tested and seems to be working without any issues.

botder (approved):

PR #2634, opened 2022-05-26
Fix onPedDamage event

xLive:
When merged, this PR will register [onPedDamage](https://wiki.multitheftauto.com/wiki/OnPedDamage) event
https://github.com/multitheftauto/mtasa-blue/blob/4dac0e237865e5d13927b42c1c93b69b1eb9a401/Server/mods/deathmatch/logic/CPedSync.cpp#L245-L248
As we can see, the event was always called. But since the event is not registered, you cannot add an event handler unless you manually register it using [addEvent](https://wiki.multitheftauto.com/wiki/AddEvent). This PR will fix it.


patrikjuvonen (approved):
Nice catch! Thank you for the pull request. Lightly tested working.

PR #2637, opened 2022-06-01
Fix building removal

### LINKED ISSUE #2631: 'restoreWorldModel' does not use it's 'fRadius' argument
### Describe the bug

The documentation says that the function needs the ``radius`` argument, even tho in code it isn't used.

### Steps to reproduce

Read ``CBuildingRemovalManager.cpp``.

### Version

All versions

### Additional context

If possible, making this function only restore world models in ``radius`` would be wonderful, since you could hide all models using ``removeWorldModel`` and then you can restore only the models you really need with ``restoreWorldModel``, for example:

```lua
removeWorldModel(1337, 10000, 0, 0, 0)
restoreWorldModel(1337, 100, 1234, -1234, 11.5)
```

### Relevant log output

_No response_

### Security Policy

tederis:
Fixes #2631. When radius passed is zero, it works as before. Otherwise it restores all models that intersect with a sphere of the specified radius. Not backward compatible, but there's no other way to do it.


TheNormalnij (dismissed):

TheNormalnij:
Does not work for me
```
Executing client-side command: removeWorldModel(1337, 10000, 0, 0, 0)
Command results: true [boolean], 9 [number]

Executing client-side command: restoreWorldModel(1337, 5, -2710, 972, 55)
Command results: true [boolean], 9 [number]

Executing client-side command: removeWorldModel(1337, 10000, 0, 0, 0)
Command results: true [boolean], 2 [number]
```

Fernando-A-Rocha:
@tederis ?

PR #2638, opened 2022-06-03
Update user agent for youtube.com/tv

MrSol0:
Youtube has once again changed their platform so it no longer supports video resume for most user agents, however roku user agent seems to still work fine with no problems similar to the old user agent.

### DIFF
--- a/Client/cefweb/CWebView.cpp
+++ b/Client/cefweb/CWebView.cpp
@@ -846,7 +846,7 @@ CefResourceRequestHandler::ReturnValue CWebView::OnBeforeResourceLoad(CefRefPtr<
 
             // Allow YouTube TV to work (#1162)
             if (domain == "www.youtube.com" && UTF16ToMbUTF8(urlParts.path.str) == "/tv")
-                iter->second = iter->second.ToString() + "; SMART-TV; Tizen 4.0";
+                iter->second = iter->second.ToString() + "; Roku 3/7.0";
 
             request->SetHeaderMap(headerMap);
         }


ffsPLASMA:
Wouldn't it be much more beneficial to set the browser user agent to a custom one via lua in conjunction with setBrowserProperty rather than hardcoding it?

patrikjuvonen:
> Wouldn't it be much more beneficial to set the browser user agent to a custom one via lua in conjunction with setBrowserProperty rather than hardcoding it?

It's set by MTA to respect the purpose of user agent header. It is used occasionally by ISPs and services to better understand and route traffic. There isn't any reason to use any other custom user agent than what MTA provides. In this case a part is added which is needed to workaround automatic routing on YouTube. There may be other issues that we don't realize by allowing custom user agents. So for now, no, we're not going to change this behavior.

I'll also need to ask people to stop discussing this in this pull request, because it is off-topic. Please talk over in a new issue or discussion.

Client/cefweb/CWebView.cpp
```diff
@@ -846,7 +846,7 @@ CefResourceRequestHandler::ReturnValue CWebView::OnBeforeResourceLoad(CefRefPtr<
 
             // Allow YouTube TV to work (#1162)
             if (domain == "www.youtube.com" && UTF16ToMbUTF8(urlParts.path.str) == "/tv")
-                iter->second = iter->second.ToString() + "; SMART-TV; Tizen 4.0";
+                iter->second = iter->second.ToString() + "; Roku 3/7.0";
```
theSarrum, line 849: Using "Roku 3/7.0" is not enough, so I tried it out and found that we could add "Wireless" and it works.
```suggestion
                iter->second = iter->second.ToString() + "; Roku 3/7.0, Wireless";
```

lopezloo:
Doesn't work for me. Neither current UA, `Roku 3/7.0` nor `Roku 3/7.0, Wireless`. I'm getting "who's watching?" screen on youtube.com/tv.

The URL I have used:
```
https://www.youtube.com/tv#/watch?mode=transport&v=jofNR_WkoCE
```

lopezloo (changes requested):
Current user agent works for me. Please provide more information about this issue.

MrSol0:
> Current user agent works for me. Please provide more information about this issue.

Youtube has restored their old system, the resume is now back again, however i still think "Roku 3/7.0" is better because it has no login page when you resume or watch videos and even no ads!

lopezloo:
> > Current user agent works for me. Please provide more information about this issue.
> 
> Youtube has restored their old system, the resume is now back again, however i still think "Roku 3/7.0" is better because it has no login page when you resume or watch videos and even no ads!

I still see ads with `Roku 3/7.0` UA. What do you mean by saying "resuming videos" and "login page"?

If I visit video URL directly, ex. this one:
```
https://www.youtube.com/tv#/watch?mode=transport&v=jofNR_WkoCE
```

I'm getting "who's watching?" page on all user agents, which can be hidden by injecting mouse click.

MrSol0:
> https://www.youtube.com/tv#/watch?mode=transport&v=jofNR_WkoCE

i have tested `https://www.youtube.com/tv#/watch?v=jofNR_WkoCE` with user agent as `Roku 3/7.0` in chrome and i don't see the who's watching page even with US & UK vpn's, and i mean by resume video this `https://www.youtube.com/tv#/watch?v=jofNR_WkoCE&resume&t=120` as an example, it tells you who's watching with other user agents but not with Roku

lopezloo:
I'm getting this login page even with `Roku 3/7.0, Wireless` UA. Without `Wireless` I got redirect to desktop version.

MrSol0:
> I'm getting this login page even with `Roku 3/7.0, Wireless` UA. Without `Wireless` I got redirect to desktop version.

It seems like you are using the wrong user agent changer, try use user-agent changer and manager which is a chrome extension. i would also like anyone else to confirm this.

Lpsd:
> > I'm getting this login page even with `Roku 3/7.0, Wireless` UA. Without `Wireless` I got redirect to desktop version.
> 
> It seems like you are using the wrong user agent changer, try use user-agent changer and manager which is a chrome extension. i would also like anyone else to confirm this.

Why are you using a Chrome extension for testing this? How can you be sure the request details are exactly the same as in MTA? If you've already accounted for this can you prove it?

patrikjuvonen:
Testing must be done through the MTA client, not external browser.

Lpsd:
It's also worth noting that YouTube generally rolls out their updates/changes slowly, so different users may have different experiences while performing the same actions. This could be down to browser, device, location, IP, account, and many other variables.

PR #2655, opened 2022-06-30
Fix #2654: Uncap money serverside

### LINKED ISSUE #2654: Unlimit player money serverside

``setPlayerMoney`` clientside is not capped to an arbitrary value like ``99999999``, it lets you use all the values of the ``long`` type, so you could set up to ``2147483647``, which is 21 times larger than the capped version.

On ``bool CStaticFunctionDefinitions::SetPlayerMoney`` in ``Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp`` there is this line of code:

```c
// Is it above 99999999? Limit it to it
if (lMoney > 99999999)
    lMoney = 99999999;
// Is it below -99999999?
else if (lMoney < -99999999)
    lMoney = -99999999;
```

I am unaware of a reason why we should cap the money. You can already pass a value larger than ``99999999`` on ``setPlayerMoney`` serverside, but it will be capped. Also, you can set numbers larger than ``2147483647`` which makes the value revert down to ``-2147483647``.

If I am not incorrect, you can already have more than 100M in cash on singleplayer
image

The vanilla HUD can show up to 999M, altought if you don't use the ``instant`` argument it will take a few minutes for it to show it, since it keeps adding money up.

image
(My real money on this picture is ``2147483647``)

The money value should not be capped serverside. Or, at least, it should be capped to the biggest value the vanilla HUD can show, which is ```999'999.999``` and not ```99'999.999```.

Cap the value clientside and make it consistent.

### Additional context

This can help server-owners that don't use a custom money system and are approaching the 99M value on some of their users accounts.

This is an easy fix and it should not break anything. I do not understand a reason why this is capped, maybe they were trying to save some bandwidth? I don't understand how that would work since the value it's still ``long`` and does not seem to be changed.

If someone knows a problem that could be generated by this, please, let me know.

### Security Policy

PlatinMTA:
This closes #2654. Maybe this PR should also block using a value argument larger than ``2147483647``, but we already had overflowing values with this function.


CrosRoad95:
did  you, anyone found backward compatibility issue?

Lpsd:
The only issue I can think of, is resources which are relying on the server-side variant of this function being capped.

As for backwards compatibility issues that break scripts, of course it's possible (depending on the amount of spaghetti code), but I can't think of any reasonable scenario where this change should break your script. Again, unless you have some spaghetti-fied logic.

It's actually better this way, because currently if you wanted to bypass the server limit, you'd have to set the money on the client - which obviously can't be trusted - so we'll be doing people a favor.

I'm fine with this change

CrosRoad95:
MAYBE! for now this feature should be hidden behind some flag

PlatinMTA:
I can't think of a resource that will break with this change. The only thing that was changed is that now, when you pass values higher than 99M, you will actually set values higher than 99M. You could always overflow the value.*

The only thing I could think scripters are doing is checking that the value they are trying to set is never above 99M, because that would mean they are giving players less money.

I forgot to update ``givePlayerMoney``. Also, fun fact, you could already set up more than 99M serverside for a player, using ``takePlayerMoney`` with negative values:

```lua
run takePlayerMoney(me, -150000000)
run getPlayerMoney(me)
```

``Command results: 150000000 [number]``


*Maybe an anti-cheat system that checks if the money of the player is above 99M. Highly unlikely imo.

Lpsd:
> MAYBE! for now this feature should be hidden behind some flag

I think it's fine, next release is 1.6 and we're being a bit more relaxed on this kind of stuff - anyway this is a really minor change with little to no impact.

patrikjuvonen:
We first need to understand the underlying reasons behind this behavior before we change it. I'll try to ask around.

patrikjuvonen:
Cazomino:
> There's definitely a wrap around bug in SA that causes it to wrap, I don't know where that is but I don't think it was an int32
> In some cases like that we restricted it to something sensible which was far less though
> It might also be a rendering issue in the UI

qaisjp:
> i would check what happens if you remove the limit and test it out

woovie:
> character limit on cash display?

ryden didn't know.

I'll still ask ccw, but I think cazo is probably right.

We need to triple check that changing this behavior doesn't break money: the actual value, the displayed value, the UI. Everything needs to be 100% consistent across the board.

Since the max value in the GTA display is 999.999.999, then that is what the limit should be, as long as everything works absolutely the same. Going beyond the UI limit means a broken UI (we don't want that). And by going beyond what the game actually supports (whatever data type it actually is), can cause game breaking bugs, which is not what we want either.

This change would obviously also have an effect for those who depend on the current limit to be there. How likely is it to cause (major) issues? Probably not very likely. But we need to understand this, that's why deprecations and upgrade notices exist. It all needs to be documented.

As for why the client version doesn't have the limit, no one knows. Could have been a deliberate design decision, networking related, security related, simply overlooked, whatever. Right now though I don't see the point of supporting more than what the UI can display. If servers need larger values, they'll need a custom UI element and custom store for the money value. The alternative is to rewrite the native renderer and wrap any money related functionality with our own system. Neither of which I don't think is all that wanted at the end of the day (just get a lua resource for that).

PlatinMTA:
I'm fine with changing the cap from 99M to 999M. Do we fix the takePlayerMoney issue? Maybe some servers are using that method in order to bypass the cap.

I won't be capping ``takePlayerMoney`` for the moment.

patrikjuvonen (approved):

PR #2659, opened 2022-07-03
Workaround fix for "Climbing over certain objects kills you, when you have high FPS" (#602)

### LINKED ISSUE #602: Climbing over certain objects kills you, when you have high FPS
**Describe the bug**
A single player bug (if you turned off frame limiter) where climbing over certain objects will rapidly drain your health and almost always kill you.

**To reproduce**
```lua
run setFPSLimit(100) me.position = Vector3(2262.86035, 169.6757, 31.4844) me.rotation = Vector3(0, 0, 1.5)
```
Jump, your player should then try to climb over it and die.

**Expected behaviour**
Not dying would be nice but can't be expected since this is a GTA SA high frame dependent bug,

**Screenshots**

**Version**
Both are v1.5.6-release-14414

**Additional context**
Alternative positions:
```lua
run me.position = Vector3(2520.356, -1692.440, 18.47)
```
```lua
run me.position = Vector3(921.3, -1283.7, 18.32)
```

From https://bugs.mtasa.com/view.php?id=7155

lopezloo:
This PR introduces workaround fix for "Climbing over certain objects kills you, when you have high FPS" (#602). It's basically the same fix as [script fix for this issue](https://github.com/multitheftauto/mtasa-blue/issues/602#issuecomment-1167673919) but done in mtasa-blue instead. We basically cancel ped damage in certain case, when:

- there is no attacker
- damage type is fall
- hit zone is torso
- ped has jump task
- ped has climb task

(in script version of this fix there is also check for `TASK_COMPLEX_IN_AIR_AND_LAND` task but it's redudant and makes it actually not working in some cases)

<sub>(damage sound is still being played)</sub>


PlatinMTA:
Crash on server connection

```
Version = 1.5.9-custom.0.000
Time = Mon Jul  4 19:48:44 2022
Module = D:\Rockstar Games\GTA San Andreas\gta_sa.exe
Code = 0xC0000005
Offset = 0x001B5209

EAX=21A54830  EBX=00000000  ECX=00000000  EDX=00000092  ESI=00000C90
EDI=00000000  EBP=008E2CB0  ESP=0177FD08  EIP=005B5209  FLG=00210297
```

sbx320:
Wouldn't it be more advisable to fix the source of the damage, rather than applying a workaround?

lopezloo:
> Wouldn't it be more advisable to fix the source of the damage, rather than applying a workaround?

I don't know what is the cause of this bug. It's looped fall damage happening during climb task. I think it happens because ped both touches the ground (like it would fall off) and starts climbing at near same time. Maybe climb task overrides the fall off task and then game tries to initiate the fall off task again forever? But even if that's the cause - how to exactly patch this in SA?

PlatinMTA:
Doesn't SilentPatch fix this? I'm not completely sure it does.

lopezloo:
> Doesn't SilentPatch fix this? I'm not completely sure it does.

I don't think so. There are no framerate related fixes in SilentPatch AFAIK. And this issue is quite minor by the way.

PlatinMTA:
> Crash on server connection
> 
> ```
> Version = 1.5.9-custom.0.000
> Time = Mon Jul  4 19:48:44 2022
> Module = D:\Rockstar Games\GTA San Andreas\gta_sa.exe
> Code = 0xC0000005
> Offset = 0x001B5209
> 
> EAX=21A54830  EBX=00000000  ECX=00000000  EDX=00000092  ESI=00000C90
> EDI=00000000  EBP=008E2CB0  ESP=0177FD08  EIP=005B5209  FLG=00210297
> CS=0023   DS=002B  SS=002B  ES=002B   FS=0053  GS=002B
> ```

Just noticed, I had the same crash with #directx11 and #2663, so it's probably on my end (always using the artifacts, and the crash comes after joining a server). I don't crash on nightlies or normal versions.

lopezloo:
This bug could be possibly fixed by patching this in SA:
https://github.com/codenulls/gta-reversed/blob/3cfeabd54aa9cefab3b8770aaec730d19b579e55/source/game_sa/Tasks/TaskTypes/CTaskSimpleClimb.cpp#L189
https://github.com/codenulls/gta-reversed/blob/3cfeabd54aa9cefab3b8770aaec730d19b579e55/source/game_sa/Tasks/TaskTypes/CTaskSimpleClimb.cpp#L210

Canceling this damage in MTA seems to be working well though.

lopezloo:
Closing in favor of #2667.

lopezloo:
#2667 doesn't work for one particular position. I'm reopening this.

PlatinMTA:
Can we get this merged? We have uncapped the fps for almost a year now and this issue stills bothers users with high fps.

ghost:
@PlatinMTA before it happens you can use Lua version of it i guess? It doesn't really touch the default damage, unless some circumstances are met. Either the one from here or [example from wiki](https://wiki.multitheftauto.com/wiki/SetFPSLimit#Fix_for_climbing_over_certain_objects)

PlatinMTA:
> @PlatinMTA before it happens you can use Lua version of it i guess? It doesn't really touch the default damage, unless some circumstances are met. Either the one from here or [example from wiki](https://wiki.multitheftauto.com/wiki/SetFPSLimit#Fix_for_climbing_over_certain_objects)

That one works, this one doesn't https://github.com/multitheftauto/mtasa-blue/issues/602#issuecomment-1167673919

Thank you srsly

botder (dismissed):
I wonder how the call stack looks like when you place a breakpoint in that special case. I don't really like this fix, because it ran ~400 times for a single jump and climb in my test.

```
client.dll          CClientGame::ApplyPedDamageFromGame
client.dll          CClientGame::DamageHandler
client.dll          CClientGame::StaticDamageHandler
multiplayer_sa.dll  ProcessDamageEvent
multiplayer_sa.dll  HOOK_CEventDamage__AffectsPed
gta_sa.exe          004ab4c4 CEventGroup::Add
gta_sa.exe          005e2fac CPed::ProcessEntityCollision
gta_sa.exe          0054c962 CPhysical::ProcessCollisionSectorList
gta_sa.exe          0054da89 CPhysical::CheckCollision
gta_sa.exe          0054e4fc CPhysical::ProcessCollision
gta_sa.exe          00568776 CWorld::Process
multiplayer_sa.dll  HOOK_CGame_Process
multiplayer_sa.dll  HOOK_Idle
```

PR #2662, opened 2022-07-07
Add hook to change vehicle damage debris to the rgb vehicle color

### LINKED ISSUE #2661: Vehicle damage spawns wrongly colored debris
### Describe the bug

When hitting a wall with a vehicle it spawns debris with the color of the original vehicle color not the rgb one.

### Steps to reproduce

Drive against a wall

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

Merlin:
Fixes #2661

https://www.youtube.com/watch?v=OfTIp8BPhss

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -293,6 +293,8 @@ const DWORD RETURN_Idle_CWorld_ProcessPedsAfterPreRender = 0x53EA08;
 #define HOOKPOS_CAEAmbienceTrackManager__UpdateAmbienceTrackAndVolume_StartRadio    0x4D7198
 #define HOOKPOS_CAEAmbienceTrackManager__UpdateAmbienceTrackAndVolume_StopRadio     0x4D71E7
 
+#define HOOKPOS_CAutomobile__dmgDrawCarCollidingParticles 0x6A6FF0
+
 CPed*         pContextSwitchedPed = 0;
 CVector       vecCenterOfWorld;
 FLOAT         fFalseHeading;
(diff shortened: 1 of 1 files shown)


Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -7027,3 +7033,39 @@ void _declspec(naked) HOOK_CAEAmbienceTrackManager__UpdateAmbienceTrackAndVolume
...
+
+        RwColor color = {
+            colors[0].R * pVehicleInterface->m_fLighting,
+            colors[0].G * pVehicleInterface->m_fLighting,
+            colors[0].B * pVehicleInterface->m_fLighting,
+            0xFF
+        };
+
+        // Fx_c::AddDebris
+        ((void(__thiscall*)(int, CVector&, RwColor&, float, int))0x49F750)(0xA9AE00, vecPosition, color, 0.06f, count / 100 + 1);
```
lopezloo:
```suggestion
        ((void(__thiscall*)(int, CVector&, RwColor&, float, int))0x49F750)(CLASS_CFx, vecPosition, color, 0.06f, count / 100 + 1);
```
How about using `CLASS_CFx` here?

`Fx_c::AddDebris()` could be also defined.

lopezloo:
We could include `..\game_sa\CFxSA.h` in `StdInc.h` and then use `FUNC_CFx_AddDebris` instead of raw address. Or just define it at the top.

Merlin:
I think its fine for now. I plan to change a bunch of stuff in game_sa including moving all *SAInterface to there own file and add all the functions we call directly as methods. That way we can just include that header and call g_fx.AddDebris without having addresses spread everywhere

lopezloo:
Ok. Otherwise looks good to me.

lopezloo (approved):
Looks good to me. Thank you.

PR #2663, opened 2022-07-10
Fix walking while aiming on higher fps

### LINKED ISSUE #502: Cannot move sideways while aiming when FPS is higher than ~75
**Describe the bug**
This is a well known bug that annoys alot of people.

**To Reproduce**
1. Set FPS limit to any value above 73 (74-100)
2. Get (for example) a deagle
3. Aim and keep aiming, then try to move sideways without moving forward or backward
4. The character will not move or it takes a while until he starts moving

**Expected behavior**
FPS should have no impact on gameplay

**Multi Theft Auto v1.5.6-release-14394**
 
**Additional context**
The original bug report was about 48 fps but problems start happening at 73 fps for me, it gets worse with even more fps.

From https://bugs.mtasa.com/view.php?id=4291

Merlin:
Fixes #502

https://www.youtube.com/watch?v=AtEML33IW0E

### DIFF
--- a/Client/mods/deathmatch/CClient.cpp
+++ b/Client/mods/deathmatch/CClient.cpp
@@ -128,6 +128,7 @@ int CClient::ClientInitialize(const char* szArguments, CCoreInterface* pCore)
     pCore->GetCommands()->Add("debug2", "debug function 2", COMMAND_Debug2);
     pCore->GetCommands()->Add("debug3", "debug function 3", COMMAND_Debug3);
     pCore->GetCommands()->Add("debug4", "debug function 4", COMMAND_Debug4);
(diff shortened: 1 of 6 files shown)


lopezloo:
It seems to be faster at higher FPS now? That time delay seems to be shorter. It's noticeable with MP5 and 40 vs 100 FPS.

Merlin:
It looks like the same speed to me. The time it takes to fully walk is definitly faster. Animations also jitters badly when walking with over 250+ fps. I will take a look at it again in a few days.

Merlin:
https://www.youtube.com/watch?v=3XTLCPWNmLQ

That fixes the faster animations at the begining of walking.

lopezloo (approved):
Very nice[.](https://youtu.be/aZVkW9p-cCU?t=55) Thank you.

PR #2664, opened 2022-07-10
Fix #1905: Refresh rate limited at 60Hz in full screen mode

### LINKED ISSUE #1905: Refresh rate limited at 60Hz in full screen mode
**Describe the bug**
Refresh rate limited at 60Hz in full screen mode.

**To reproduce**
1. Set server FPS to 100.
2. Check refresh rate on monitor that supports more than 60Hz.

**Expected behaviour**
MTA should use available refresh rate and work with adaptive-sync.

**Screenshots**
az9hzG8lolU

**Version**
From https://github.com/multitheftauto/mtasa-blue/commit/ee9a827fe6a34a29bd0262d29b06281cf38ad262

samr46:
Fixes #1905 

Hardcoded places mentioned in [this comment](https://github.com/multitheftauto/mtasa-blue/issues/1905#issuecomment-952182365) by @Dutchman101 are used for failsafe mode (if **CreateDevice** fails) and testing. I don't think we should modify it to use adapter's refresh rate.
> Either way, the places it gets hardcoded:
> 
> 1. https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/DXHook/CProxyDirect3D9.cpp#L421
> 2. https://github.com/multitheftauto/mtasa-blue/blob/c58c3678f263d810f61eb5aecba48e6c4b4dce0c/Client/core/DXHook/CProxyDirect3D9.cpp#L446

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Direct3D.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Direct3D.cpp
@@ -47,6 +47,11 @@ void _cdecl OnPreCreateDevice(IDirect3D9* pDirect3D, UINT Adapter, D3DDEVTYPE De
     ms_BehaviorFlags = *BehaviorFlags;
     ms_pPresentationParameters = pPresentationParameters;
     ms_ppReturnedDeviceInterface = ppReturnedDeviceInterface;
+
+    // Set fullscreen refresh rate to display adapter's value
+    D3DDISPLAYMODE DisplayMode;
+    if (!FAILED(ms_pDirect3D->GetAdapterDisplayMode(Adapter, &DisplayMode)))
+        ms_pPresentationParameters->FullScreen_RefreshRateInHz = DisplayMode.RefreshRate;
 }
 
 // Hook info


lopezloo:
> For windowed mode, the refresh rate must be 0.

https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dpresent-parameters#:~:text=the%20D3DPRESENTFLAG%20constants.-,FullScreen_RefreshRateInHz,-Type%3A%20UINT

Shouldn't we set it to 0 when MTA is running in windowed mode?

samr46:
> Shouldn't we set it to 0 when MTA is running in windowed mode?

It is 0 in windowed mode. It's overwritten later using D3DPRESENT_RATE_DEFAULT.

```
20:58:21 - [DEBUG]     SwapEffect:1  Windowed:1  EnableAutoDepthStencil:1  AutoDepthStencilFormat:75  Flags:0x0
20:58:21 - [DEBUG]     FullScreen_RefreshRateInHz:0  PresentationInterval:0x00000001
```

lopezloo (approved):
Looks good to me. Thank you.

PR #1402, opened 2020-04-22
setWaterLevel: add bIncludeWorldSeaLevel and bIncludeOutsideWorldLevel

TheNormalnij:
This PR makes it possible to change the water level outside the game world.

New syntax:
```lua
bool setWaterLevel ( float level [, bool includeWaterFeatures = true, bool includeWaterElements = true, bool includeWorldSea = true, bool includeOutsideWorld = false ] )
```
Reopened #1342 #889

PR #2665, opened 2022-07-10
setWaterLevel: set water sound level outside of game boundaries

samr46:
Improvement for PR #1402

Fixes water sound level outside of game boundaries (currently it's always 0 even if water level is set to -50).

```
setWaterLevel(-50, true, true, true, true)
```

It modifies hardcoded 0 to water level for ambient water sound check outside of game boundaries:

image


lopezloo (approved):
Looks good to me. Thank you.

PR #2670, opened 2022-07-12
Fix #2510: setVehiclesLODDistance() & setPedsLodDistance() can't override player settings

### LINKED ISSUE #2510: setVehiclesLODDistance() & setPedsLodDistance() can't override player settings
**Describe the bug**
setVehiclesLODDistance() & setPedsLodDistance() values are being clamped to values based on user video settings.

**To reproduce**
Disable "Render vehicles always in high quality" option in video settings.
Stay away from some vehicles to see their low LOD models.
`crun setVehiclesLODDistance(300)`
... no effect

**Expected behaviour**
setVehiclesLODDistance() & setPedsLODDistance() should be able to override player settings (?).

**Version**
v1.5.9-release-21067

**Additional context**
- Originally reported [here](https://github.com/multitheftauto/mtasa-blue/pull/2347#issuecomment-1008150235).
- https://github.com/multitheftauto/mtasa-blue/pull/832/commits/2308f5a940cc9e4ebee7701649ab87e05f8e8ed1

samr46:
Fixes #2510

LOD distances will be restored to player settings on connection or if server calls reset functions.

Now server owners will be able to resolve disbalance issues caused by these checkboxes using appropriate server defined values instead of forcing default small distances for everyone.

### DIFF
--- a/Client/game_sa/CSettingsSA.cpp
+++ b/Client/game_sa/CSettingsSA.cpp
@@ -563,48 +563,35 @@ void CSettingsSA::SetFieldOfViewVehicleMax(float fAngle, bool bFromScript)
 // Vehicles LOD draw distance
 //
 ////////////////////////////////////////////////
-float ms_fClientMaxVehicleLODDistance = DEFAULT_VEHICLE_LOD_DISTANCE;
-float ms_fClientMaxTrainPlaneLODDistance = DEFAULT_VEHICLE_LOD_DISTANCE * TRAIN_LOD_DISTANCE_MULTIPLIER;
 bool  ms_bMaxVehicleLODDistanceFromScript = false;
 
 void CSettingsSA::SetVehiclesLODDistance(float fVehiclesLODDistance, float fTrainsPlanesLODDistance, bool bFromScript)
 {
-    if (!bFromScript)
-    {
-        ms_fClientMaxVehicleLODDistance = fVehiclesLODDistance;
-        ms_fClientMaxTrainPlaneLODDistance = fTrainsPlanesLODDistance;
-    }
-
     ms_fVehicleLODDistance = fVehiclesLODDistance;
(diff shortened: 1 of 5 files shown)


Client/game_sa/CSettingsSA.cpp
```diff
@@ -565,39 +565,24 @@ void CSettingsSA::SetFieldOfViewVehicleMax(float fAngle, bool bFromScript)
...
-        ms_fVehicleLODDistance = Min(fVehiclesLODDistance, ms_fClientMaxVehicleLODDistance);
-        ms_fTrainPlaneLODDistance = Min(fTrainsPlanesLODDistance, ms_fClientMaxTrainPlaneLODDistance);
-    }
+    ms_fVehicleLODDistance = fVehiclesLODDistance;
+    ms_fTrainPlaneLODDistance = fTrainsPlanesLODDistance;
+    ms_bMaxVehicleLODDistanceFromScript = bFromScript;
 }
 
-void CSettingsSA::ResetVehiclesLODDistance(bool bFromScript)
+void CSettingsSA::ResetVehiclesLODDistance(bool bFromScript, bool bFromConfig)
```
lopezloo: Why this method looks so complicated? Couldn't it look more like `CSettingsSA::ResetCoronaReflectionsEnabled()`?

samr46: I can remove this argument and restore the additional function that was used. I didn't like the name of this function.

Or I can simplify the overall approach. I believe this more complicated way (saving distance from config into variable) was chosen by the author for performance reasons. With current implementation this functions requests user settings only on start, connection or if settings were changed during gameplay.

I can make it more like corona reflections implementation with way less code (because this additional variable that author used is kinda useless unless he attempted to reduce amount of calls to config).

lopezloo (approved):
Looks good to me. Thank you.

PR #2674, opened 2022-07-17
Add destructor to CAnimBlendAssocGroupSA

### LINKED ISSUE #2672: setElementModel - memory leaking
### Describe the bug

Calling setElementModel would cause GTA process to claim memory and never free it. It happens for peds, didn't tested other types though.

### Steps to reproduce

```lua
local playerSkins = {11, 9, 7}
local skinsCount = #playerSkins

function changeSkin()
    local randomSkin = math.random(1, skinsCount)
    local newSkin = playerSkins[randomSkin]

    setElementModel(localPlayer, newSkin)
end
setTimer(changeSkin, 1, 0)
```

Run this code and observe GTA process memory usage in task manager.

### Version

Client/server: Multi Theft Auto v1.5.9-release-21261

### Additional context

/

### Relevant log output

_No response_

### Security Policy

Lpsd:
Resolves #2672

Here's a [test resource](https://github.com/multitheftauto/mtasa-blue/files/9128490/model.zip) for both of the reproducible scenarios mentioned in that issue. Simply start the resource and use `/mode skin` or `/mode spawn` to switch between different modes/scenarios.

The memory should no longer increase when spawning peds or changing models.

### DIFF
--- a/Client/game_sa/CAnimBlendAssocGroupSA.cpp
+++ b/Client/game_sa/CAnimBlendAssocGroupSA.cpp
@@ -18,6 +18,11 @@ CAnimBlendAssocGroupSA::CAnimBlendAssocGroupSA(CAnimBlendAssocGroupSAInterface* 
     SetupAnimBlock();
 }
 
+CAnimBlendAssocGroupSA::~CAnimBlendAssocGroupSA()
+{
+    delete m_pAnimBlock;
+}
+
 CAnimBlendAssociationSAInterface* CAnimBlendAssocGroupSA::CopyAnimation(unsigned int AnimID)
 {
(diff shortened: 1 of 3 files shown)


Lpsd:
weird, right as I posted this PR (after testing 5-10 minutes) the memory usage starts increasing. Needs more investigation

Lpsd:
Not sure how I wasn't seeing a memory increase at all when I initially tested this, the destructor wasn't being called due to a missing virtual destructor which I've now added.

This certainly seems to clean up the disparity in amount of CAnimBlockSA objects, but memory usage still increases with this test resource running. Looks like we still have memory leaks elsewhere?

PR #2678, opened 2022-07-22
Fix random foliage on replaced collisions

### LINKED ISSUE #2694: Random foliage on replaced collisions bug
### Describe the bug

When you replace collision on model which originaly have random foliage generated on it and this original collision was not loaded by GTA since you started your game, then random foliage will not generate on replaced collision because MTA currently does not use ```CPlantMgr::SetPlantFriendlyFlagInAtomicMI```. However if you first go to place where this model is and then replace collision - random foliage will generate if appropriate materials are assigned in collision. 

### Scenario 1
1. Re/start MTA.
2. Open Map Editor.
3. Start resource [collision_test_resource.zip](https://github.com/multitheftauto/mtasa-blue/files/9210220/collision_test_resource.zip)
4. Fly there: -2026, 2275, 23.82

Result:
mta-screen_2022-07-28_14-43-29

### Scenario 2
1. Re/start MTA.
2. Open Map Editor.
3. Fly there: -2026, 2275, 23.82
4. Start resource [collision_test_resource.zip](https://github.com/multitheftauto/mtasa-blue/files/9210220/collision_test_resource.zip)
5. Fly back to Grove Street.
6. Fly back to replaced model and collision.

Result:
mta-screen_2022-07-28_14-47-53

### Version

_No response_

### Additional context

PR #2678 fixes this problem.

### Relevant log output

_No response_

### Security Policy

TFP-dev:
### This PR introduces fix for random foliage on custom collisions.

Fixes #2694

### This is how custom model and custom collision looks like on current MTA version:

mta-screen_2022-07-22_15-35-06

**This is how it looks after this fix:**

mta-screen_2022-07-22_15-36-42
[test_resource.zip](https://github.com/multitheftauto/mtasa-blue/files/9169041/test_resource.zip)


patrikjuvonen:
This changes MTA current default behavior and may be unwanted in a number of cases so I think we should introduce some way (client-side Lua) to toggle this flag before merging as such.

Also, does randomfoliage world property affect this?

lopezloo:
> This changes MTA current default behavior and may be unwanted in a number of cases so I think we should introduce some way (client-side Lua) to toggle this flag before merging as such.

There is `engineSetSurfaceProperties` function with disabled `createobjects` and `createplants` flag:
https://github.com/multitheftauto/mtasa-blue/blob/cd1d2088a92b685b3b6294acb388fc1154ddbee4/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp#L611-L615 https://github.com/multitheftauto/mtasa-blue/blob/ebf54d4fcfe9b0b5600c2b632595029ea97f2b3f/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp#L1280-L1288 Maybe it could be fixed.

//edit:
I have created issue about that: #2681.

TFP-dev:
> This changes MTA current default behavior and may be unwanted in a number of cases so I think we should introduce some way (client-side Lua) to toggle this flag before merging as such.

Appropriate materials assigned in the collision are required for this to generate. Most servers use material 0 in their collisions, so I think lua function is unnecessary.

> Also, does randomfoliage world property affect this?

`setWorldSpecialPropertyEnabled` works fine with this

> There is `engineSetSurfaceProperties` function with disabled `createobjects` and `createplants` flag:

But `engineSetSurfaceProperties` works completely different.

patrikjuvonen:
> > This changes MTA current default behavior and may be unwanted in a number of cases so I think we should introduce some way (client-side Lua) to toggle this flag before merging as such.
> 
> Appropriate materials assigned in the collision are required for this to generate. Most servers use material 0 in their collisions, so I think lua function is unnecessary.
> 
> > Also, does randomfoliage world property affect this?
> 
> `setWorldSpecialPropertyEnabled` works fine with this
> 
> > There is `engineSetSurfaceProperties` function with disabled `createobjects` and `createplants` flag:
> 
> But `engineSetSurfaceProperties` works completely different.

Ok thanks for clarification. I think we just need to test this and try the existing property function on this.

Great find all in all.

TFP-dev:
Fixes #2694

patrikjuvonen:
When you unload the resource, the foliage stays there. Probably should be fixed, but can be worked around by warping out of stream distance and warping back (and maybe some other method). engineRestreamWorld did not remove the foliage.

Untitled

patrikjuvonen (approved):
The unloading issue should be addressed somehow in my opinion. But other than that ok.

TFP-dev:
> When you unload the resource, the foliage stays there. Probably should be fixed, but can be worked around by warping out of stream distance and warping back (and maybe some other method). engineRestreamWorld did not remove the foliage.
> 
> Untitled

Foliage also stays when you use removeWorldModel in current MTA version on original models, so this is not an issue with this PR.
obraz

patrikjuvonen:
> Foliage also stays when you use removeWorldModel in current MTA version on original models, so this is not an issue with this PR.

Thanks for clarifying. Maybe we'll create a new issue for that.

PR #2684, opened 2022-07-26
Update CTimer::m_FrameCounter only every 33+ ms

Merlin:
Fixes aircraft and boats lights blinking faster on high fps. Probably fixes more stuff that's done with CTimer::m_FrameCounter.
This is also the base for more frame based fixes I will do.

Right now CTimer::m_FrameCounter is incremented every frame. This PR changes this to be updated only if 33ms passed.
That way we have multiple frames with the same frame counter.

Also removes CHud::RenderHealthBar hook as it's not needed anymore with this fix.

https://www.youtube.com/watch?v=DJv-4vKHB3g


lopezloo (approved):
Looks good to me. Thank you.

lopezloo:
Also fixed "When diving, oxygen is depleted quickly, and afterwards player is killed almost instantly cause health is reduced so fast".

PR #2704, opened 2022-08-06
Fix helicopters taking off faster on high fps

Merlin:


lopezloo (approved):
Looks good to me. Thank you.

PR #2706, opened 2022-08-09
Improve performance of CClientObjectManager::UpdateLimitInfo

Merlin:
This improves the performance of CClientObjectManager::UpdateLimitInfo.
I measured with SharedUtil::GetTimeUs and it consistently took 212 microseconds for these 3 calls.
It is called every frame and also whenever a object is streamed in or out.

Instead of calling these functions I hooked the pool functions that add and remove entries to the pool and now keep track of the amount of used spaces in a separate variable.

Now it only takes 0-1 microseconds.


Pirulax:
I don't think that will work, because sometimes the delete operator is inlined.
Although, you could paralellize it, see https://en.cppreference.com/w/cpp/algorithm/count ExceutionPolicy

Merlin:
I also patched all places that got inlined. There is also some test code left in a comment for whoever reviews this that can be used to check if the values match.

Client/sdk/multiplayer/CMultiplayer.h
```diff
@@ -390,4 +390,8 @@ class CMultiplayer
     virtual eAnimGroup GetLastStaticAnimationGroupID() = 0;
     virtual eAnimID    GetLastStaticAnimationID() = 0;
     virtual DWORD      GetLastAnimArrayAddress() = 0;
+
+    virtual unsigned int EntryInfoNodePool_NoOfUsedSpaces() = 0;
+    virtual unsigned int PtrNodeSingleLinkPool_NoOfUsedSpaces() = 0;
+    virtual unsigned int PtrNodeDoubleLinkPool_NoOfUsedSpaces() = 0;
```
botder:
```suggestion
    virtual unsigned int EntryInfoNodePool_NoOfUsedSpaces() const noexcept = 0;
    virtual unsigned int PtrNodeSingleLinkPool_NoOfUsedSpaces() const noexcept = 0;
    virtual unsigned int PtrNodeDoubleLinkPool_NoOfUsedSpaces() const noexcept = 0;
```

Client/multiplayer_sa/CMultiplayerSA_ObjectStreamerOptimization.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+unsigned int CMultiplayerSA::PtrNodeSingleLinkPool_NoOfUsedSpaces()
+{
+    return CPtrNodeSingleLinkPool_UsedSpaces;
+}
+
+unsigned int CMultiplayerSA::PtrNodeDoubleLinkPool_NoOfUsedSpaces()
+{
+    return CPtrNodeDoubleLinkPool_UsedSpaces;
+}
```
botder:
```suggestion
unsigned int CMultiplayerSA::EntryInfoNodePool_NoOfUsedSpaces() const noexcept
{
    return CEntryInfoNodePool_UsedSpaces;
}

unsigned int CMultiplayerSA::PtrNodeSingleLinkPool_NoOfUsedSpaces() const noexcept
{
    return CPtrNodeSingleLinkPool_UsedSpaces;
}

unsigned int CMultiplayerSA::PtrNodeDoubleLinkPool_NoOfUsedSpaces() const noexcept
{
    return CPtrNodeDoubleLinkPool_UsedSpaces;
}
```

Client/multiplayer_sa/CMultiplayerSA.h
```diff
@@ -296,6 +297,10 @@ class CMultiplayerSA : public CMultiplayer
     eAnimID    GetLastStaticAnimationID() { return m_dwLastStaticAnimID; }
     DWORD      GetLastAnimArrayAddress() { return m_dwLastAnimArrayAddress; }
 
+    unsigned int EntryInfoNodePool_NoOfUsedSpaces();
+    unsigned int PtrNodeSingleLinkPool_NoOfUsedSpaces();
+    unsigned int PtrNodeDoubleLinkPool_NoOfUsedSpaces();
```
botder:
```suggestion
    unsigned int EntryInfoNodePool_NoOfUsedSpaces() const noexcept override;
    unsigned int PtrNodeSingleLinkPool_NoOfUsedSpaces() const noexcept override;
    unsigned int PtrNodeDoubleLinkPool_NoOfUsedSpaces() const noexcept override;
```

botder (changes requested):

botder (approved):

PR #2708, opened 2022-08-12
Add isCapsLockEnabled client function

Lpsd:
Self explanatory, there's no consistent method to determine whether a client has caps lock enabled (or disabled) currently. 

(you can probably find this out via JavaScript by creating a CEF instance, but that seems unnecessary)

```lua
function checkCaps()
    iprint(isCapsLockEnabled(), getTickCount())
end
addCommandHandler("caps", checkCaps)
```

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaClientDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaClientDefs.cpp
@@ -15,13 +15,16 @@
 
 void CLuaClientDefs::LoadFunctions()
 {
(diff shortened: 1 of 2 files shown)


Pirulax:
Perhaps add it to `getKeyState`?

Lpsd:
Hmm, I guess we could add an extra return value to `getKeyState`, nil by default. 

Any keys which can be toggled return true or false based on their toggle state, which means we can also add support for other keys such as scroll lock.

thisdp:
maybe num lock/scroll lock?

botder (approved):

PR #2713, opened 2022-08-15
Clear script memory after it has been loaded

Pirulax:
Clear out cached scripts memory after they're loaded to make it harder to dump scripts.


ward1s:
oh no

Client/mods/deathmatch/logic/lua/CLuaMain.cpp
```diff
@@ -227,6 +227,7 @@ bool CLuaMain::LoadScriptFromBuffer(const char* cpInBuffer, unsigned int uiInSiz
     // If compiled script, make sure correct chunkname is embedded
     CLuaShared::EmbedChunkName(strNiceFilename, &cpBuffer, &uiSize);
 
+    bool ret{};
```
TheNormalnij: Do you need this value?

Lpsd: It's set below on line 284 and used as the return value shortly after.

https://github.com/multitheftauto/mtasa-blue/pull/2713/files#diff-5646d00f776d78bd8a35295ac5f410aa81465a191106fc8fa49a5d6f90952b3cR284

TheNormalnij: Yeah, but just `return true` and `return false` are better

Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -267,14 +267,14 @@ void CResource::Load()
     }
 
     // Load the no cache scripts first
-    for (std::list<SNoClientCacheScript>::iterator iter = m_NoClientCacheScriptList.begin(); iter != m_NoClientCacheScriptList.end(); ++iter)
-    {
+    while (!m_NoClientCacheScriptList.empty()) {
```
TheNormalnij: simple and readable variant 
`for (SNoClientCacheScript item : m_NoClientCacheScriptList)`

Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -267,14 +267,15 @@ void CResource::Load()
...
     // Load the no cache scripts first
-    for (std::list<SNoClientCacheScript>::iterator iter = m_NoClientCacheScriptList.begin(); iter != m_NoClientCacheScriptList.end(); ++iter)
-    {
+    for (auto& item : m_NoClientCacheScriptList) {
         DECLARE_PROFILER_SECTION(OnPreLoadNoClientCacheScript)
-        const SNoClientCacheScript& item = *iter;
+
         GetVM()->LoadScriptFromBuffer(item.buffer.GetData(), item.buffer.GetSize(), item.strFilename);
+        item.buffer.ZeroClear();
+        m_NoClientCacheScriptList.pop_front();
```
botder: You are invalidating every iterator and reference to the first element in the list here. This is not going to work with a range-based for loop.

Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -267,14 +267,15 @@ void CResource::Load()
     }
 
     // Load the no cache scripts first
-    for (std::list<SNoClientCacheScript>::iterator iter = m_NoClientCacheScriptList.begin(); iter != m_NoClientCacheScriptList.end(); ++iter)
-    {
+    for (auto& item : m_NoClientCacheScriptList) {
```
botder: This is subjective, but you should only use auto when the type is visible [on the right of an assignment].

```suggestion
    for (SNoClientCacheScript& item : m_NoClientCacheScriptList) {
```

Client/mods/deathmatch/logic/lua/CLuaMain.cpp
```diff
@@ -242,10 +242,18 @@ bool CLuaMain::LoadScriptFromBuffer(const char* cpInBuffer, unsigned int uiInSiz
...
             strUTFScript = std::string(cpBuffer, uiSize);
+        }
 
         // Run the script
-        if (CLuaMain::LuaLoadBuffer(m_luaVM, bUTF8 ? cpBuffer : strUTFScript.c_str(), uiSize, SString("@%s", *strNiceFilename)))
+        const bool loadFailed = CLuaMain::LuaLoadBuffer(m_luaVM, bUTF8 ? cpBuffer : strUTFScript.c_str(), uiSize, SString("@%s", *strNiceFilename));
+
+        // Clear raw script from memory
+        std::fill(strUTFScript.begin(), strUTFScript.end(), 0);
+        memset((void*)cpBuffer, 0, uiSize);
```
botder:
```suggestion
        std::fill_n(const_cast<char*>(cpBuffer), uiSize, 0);
```

Client/mods/deathmatch/logic/lua/CLuaMain.cpp
```diff
@@ -273,19 +283,22 @@ bool CLuaMain::LoadScriptFromBuffer(const char* cpInBuffer, unsigned int uiInSiz
             // Cleanup any return values
             if (lua_gettop(m_luaVM) > luaSavedTop)
                 lua_settop(m_luaVM, luaSavedTop);
+
             return true;
         }
     }
 
+    memset((void*)cpBuffer, 0, uiSize);
```
botder:
```suggestion
    std::fill_n(const_cast<char*>(cpBuffer), uiSize, 0);
```

botder (changes requested):

Client/mods/deathmatch/logic/CResource.cpp
Pirulax: The type of the item is not important in this context, that's why I use auto.

Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -266,15 +266,19 @@ void CResource::Load()
...
         GetVM()->LoadScriptFromBuffer(item.buffer.GetData(), item.buffer.GetSize(), item.strFilename);
+        item.buffer.ZeroClear();
+
         DECLARE_PROFILER_SECTION(OnPostLoadNoClientCacheScript)
     }
-    m_NoClientCacheScriptList.clear();
+
+    // Load the no cache scripts first
+    for (auto& item : m_NoClientCacheScriptList) {
+    }
```
botder:
```suggestion
```

botder (approved):

boti666:
hah i dumped seemta today

Issue #1365, opened 2020-04-09
Ability to enable/disable custom weapons collisions


Brian9221:
**Describe the bug**
Creating a custom weapon with createWeapon doesn't let you to disable its collisions with setElementCollisionsEnabled or setElementCollidableWith. 

**To reproduce**
crun w = createWeapon("uzi", getElementPosition(localPlayer))
crun setElementCollisionsEnabled(w, false)
crun setElementCollidableWith(w, localPlayer, false) 

both collisions functions return false since they don't support weapon elements

crun setWeaponFlags(w, "disable_model", true)

doesn't remove collisions but only the model

**Expected behaviour**
At least one of the 3 functions above should be useful to disable other elements from colliding with custom made weapons

**Version**
<!--
Please provide the following information:
Client 1.5.7-9 20358.5
Windows 10

Pirulax:
Why is this flagged as 'bug', when its clearly a feature request?

Pirulax:
Okay, so I've tried it, but they dont seem to have collisions by default..
Please provide me a custom model if you have used one on it. It makes sense why `setElementCollisionsEnabled` didnt support it.

Lpsd:
A weapon is not an element. It should not work with any `setElement*` functions unless we change weapons to be elements.

Therefore (if the original issue is even relevant / replicable), weapons should have their own functions to support this (following the `setWeapon*` / `getWeapon*` naming scheme)

qaisjp:
Weapon extends object https://github.com/multitheftauto/mtasa-blue/blob/e11685cab4beb7958ab202261f9c9d9b4ce71e58/Client/mods/deathmatch/logic/CClientWeapon.h

Which extends CClientStreamElement which extends CClientEntity (element)  https://github.com/multitheftauto/mtasa-blue/blob/ff6715d7354d54b821672aa83fecf82457299575/Client/mods/deathmatch/logic/CClientStreamElement.h

So I think weapons are elements

Lpsd:
You are correct, I confused normal weapons with custom weapons created using `createWeapon`

Pirulax:
They're elements, and I did add `setCollisionsEnabled` with 2 lines of code, but after testing it I realized weapons have no collisions by default. 
Thats why I said please provide custom models, so I can see whats happening. Also, I think if u just dont replace the COL, it'll work just fine.

1iminal:
[...]

PR #2716, opened 2022-08-18
Split grenade collision from weapon collision

### LINKED ISSUE #1123: The grenade bug / exploit
**Describe the bug**
The "grenade" exploit/ bug is an issue that has been getting abused more and more recently.
This game breaking issue generally affects the Shooter game-mode (if you don't know what Shooter is, it's a race-based mode that involves vehicles shooting missiles from their bonnet using *createProjectile(vehicle, 19, x, y, z, 1.0, nil)*).
This bug, or rather exploit, results in the game to behave abnormally. After being triggered, shots don't look or feel the same, they often teleport, appear laggy, explode near enemies, do more damage, etc.

**To reproduce**
Steps to reproduce the behaviour generally involve switching between freeroam to race:
**_-REMOVED_**
I removed the reproduction steps because so far, this only helped spread this issue.

**Video proof**
https://youtu.be/_N-JGYyti9I
You can see how the shot magically teleports/explode near another player.

https://youtu.be/Sau5Y9faFS4
Another example showing how the shot explodes near me without hitting me or the ground **(the other player also has the bug active, for us to test it around)**

**Additional context**
Something worth mentioning is if you restart MTA, everything goes back to normal and the bug doesn't work unless you trigger it again. Also this generally involves switching between servers (one to trigger the bug at, one to benefit from it at), so a potential fix could be by somehow resetting whatever this bug changes about the game whenever you leave a server.

Merlin:
This most likely fixes #1123 
Atleast I couldn't identify any other change that happens on first grenade throw.

Currently on the first grenade throw we add a colsphere into the colmodel of the grenade. But all weapon models have the same colmodel set. This PR splits grenades into their own colmodel and also adds the colsphere on game start.

I tested it and it splits the collision. But I can't say if this completely fixes #1123.
This is however still an useful change because multiple people ran into the issue that weapons/projectiles have collision where they should not have any.


lopezloo:
> This most likely fixes #1123

Needs to be confirmed.

> This is however still an useful change because multiple people ran into the issue that weapons/projectiles have collision where they should not have any.

Why not? Personally I use the fact that weapon objects have collisions in my gamemode. You can disable these via `setElementCollisionsEnabled`, which apparently doesn't work on weapon elements but it's different issue (#1365).

Weapon objects doesn't have any collisions after this change?

Merlin:
> Weapon objects doesn't have any collisions after this change?

Weapons by default do not have any collisions. Only after the first grenade throw they have collision.
I think it should be consistent and not change mid game when someone throws a grenade.

lopezloo:
> I think it should be consistent and not change mid game when someone throws a grenade.

I agree. We could initialize that collision then. In my gamemodes I always used to create projectile at the resource start to fix this. I just don't get why do you want to seperate it from other weapon models. I think it would be nice for weapon models to have collisions.

patrikjuvonen:
> I think it would be nice for weapon models to have collisions.

Wouldn't more collisions cause more overhead? If there aren't any actual benefits from collisions there shouldn't be any. We should benchmark any difference since depending on the server there may be thousands of weapons in use at a time.

lopezloo:
> > I think it would be nice for weapon models to have collisions.
> 
> Wouldn't more collisions cause more overhead? If there aren't any actual benefits from collisions there shouldn't be any. We should benchmark any difference since depending on the server there may be thousands of weapons in use at a time.

I don't think it affects weapons carried by players.

1iminal:
Player held weapon don't have collisions, so why should custom weapons have them?
In any case, you can always attach an invisible object with a collision to a custom weapon if you need them, it'd be still a less hacky solution than relying on current blatantly erroneous behavior.

ghost:
Player held weapons/custom created weapons shouldn't have collisions - at least not by default. If it would be added, it should be toggleable via function, and not being enforced.

lopezloo:
> Player held weapon don't have collisions, so why should custom weapons have them?

But that change affect objects too.

We could fix #1365 and then make custom created weapons collisionless by default.

So, in summary, my proposal is:
- initialize that collision
- add custom created weapons support to `setElementCollisionsEnabled` (fix #1365)
- make custom created weapons collisionless by default

Merlin:
I agree that setElementCollisionEnabled should work for custom weapons.
But I don't think making custom weapons collisionless by default is the right choice.
When I replace the collision of a weapon model I expect the collision be there by default for custom weapons.

If anyone needs the grenade collision on weapon objects they can always just add a collision to that weapon model.
It's just a col sphere with center = {0.0, 0.0, 0.0} and radius = 0.1875

lopezloo:
Thinking about it now, making custom created weapons collisionless by default would break backward compatibility because it's actually possible to create weapon with model different than weapon model (it's possible to use `setElementModel` on weapon elements). Also it's possible to replace weapon model as you said.

I don't like the idea of making weapon models collisionless though. Maybe we could initialize small box collison for weapon models. But that would kinda break backward compatibility too.

sbx320 (approved):

botder (approved):

PR #2727, opened 2022-09-01
New "MTA Blue" icon

TFP-dev:
Current MTA icon is old and does not represent "MTA:SA Blue". This PR changes them to be more modern with blue color taken from menu logo.

### Comparison:
obraz
obraz
obraz


patrikjuvonen:
Sorry, but we have no plans of changing the icon. This sort of change would need to be part of an overall brand redesign and GitHub is not the right medium for those discussions.

Thanks though.

PR #2737, opened 2022-09-12
Fix crash when dxCreateTexture higher than 4096 pixels

### LINKED ISSUE #287: Crashes with dxCreateTexture with pixel count+:setPixels()
**Describe the bug**
So, I was doing some random thing, and for some reason I crashed a few times.
I was using dxCreateTexture() with some big pixel counts(12k+ on x and y)+setPixels()
I crashed out every time, but I dont know exactly caused the crash, but it has something to do with the mentioned functions.

**To Reproduce**
Maybe:
1. C a texture(dxCreateTexture(20000, 20000))
2. Set pixels on each axis:
`for x = 0, math.floor(20000/32), 32 do
    for y = 0, math.floor(20000/32), 32 do
        tex:setPixels(<Add pixels here from a 32x32 image>, x, y, 32, 32)
    end
end`
This is how I crashed, or at least, it was like this.

**Expected behavior**
It should create the texture and set its pixels

 - Multi Theft Auto v1.5.5-release-12411
- Running on Wine? Nein.

 - OS: Microsoft Windows [Version 10.0.17134.165]
 - Multi Theft Auto Server v1.5.5-release-12411

Allerek:
Should fix #287 according to
> When it comes to operations with textures MTA is relying on D3D9 API. But the largest texture size is 4096 × 4096 in DX9. Of course you can try to create a bigger texture but it's unsafe and can result in unpredictable consequences.

Test resource
[texture_test.zip](https://github.com/multitheftauto/mtasa-blue/files/9548718/texture_test.zip)

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
@@ -1001,13 +1001,11 @@ int CLuaDrawingDefs::DxCreateTexture(lua_State* luaVM)
         // element dxCreateTexture( int width, int height [, string textureFormat = "argb", string textureEdge = "wrap", string textureType = "2d", int depth ]
         // )
         argStream.ReadNumber(width);
-        //if (width > 4096)
-        //    argStream.SetCustomError("Expected number lower than 4096 at argument 1");
+        if (width > 4096)
+            argStream.SetCustomError("Expected number lower than 4096 at argument 1");
         argStream.ReadNumber(height);
-        /*if (height > 4096)
-            argStream.SetCustomError("Expected number lower than 4096 at argument 2");*/
-
-
+        if (height > 4096)
+            argStream.SetCustomError("Expected number lower than 4096 at argument 2");
         argStream.ReadEnumString(renderFormat, RFORMAT_UNKNOWN);
         argStream.ReadEnumString(textureAddress, TADDRESS_WRAP);
         argStream.ReadEnumString(textureType, TTYPE_TEXTURE);
--- a/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
@@ -1001,7 +1001,13 @@ int CLuaDrawingDefs::DxCreateTexture(lua_State* luaVM)
         // element dxCreateTexture( int width, int height [, string textureFormat = "argb", string textureEdge = "wrap", string textureType = "2d", int depth ]
         // )
         argStream.ReadNumber(width);
+        //if (width > 4096)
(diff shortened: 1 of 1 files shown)


Lpsd:
Assuming you've forgotten to uncomment the code?

Also, the wording should be `Expected number less than, or equal to 4096 at argument 1` since you're comparing as `x > 4096`

Allerek:
> Assuming you've forgotten to uncomment the code?
> 
> Also, the wording should be `Expected number <less/greater> than, or equal to 4096 at argument 1` since you're comparing as `x > 4096`

Hmm, github desktop automatically commented code out, weird.

Lpsd (approved):
LGTM

samr46:
This will break current behavior of dxCreateTexture.

Also as far as I know it crashes (in really rare cases) only if setPixels is used. So if you wanna block anything higher than 4096 (in my opinion a bad idea) then it makes sense to patch setPixels leaving dxCreateTexture unchanged. Or better just don't patch anything and put a giant warning on wiki about dimensions higher than 4096.

Lpsd:
We shouldn't allow textures larger than the API specification regardless

CrosRoad95:
e? people are using bigger textures for radar, does it break backward compatibility?

Allerek:
> This will break current behavior of dxCreateTexture.
> 
> Also as far as I know it crashes (in really rare cases) only if setPixels is used. So if you wanna block anything higher than 4096 (in my opinion a bad idea) then it makes sense to patch setPixels leaving dxCreateTexture unchanged. Or better just don't patch anything and put a giant warning on wiki about dimensions higher than 4096.

If you check test resource and code, you will notice it only blocks it when creating textures by giving pixels, using image that have even 20kx20k is fine.

No idea why, and how it works - but it does.

Lpsd:
possibly. but I feel like we often make decisions based on bad practices in existing scripts and being scared of breaking those. We really should care less if someone's spaghetti script breaks in a major release (1.6) due to beneficial changes (conforming to API specifications) and they are unable to access original source or don't have the knowledge to upgrade it.

Let's consider this for 1.6

You really don't need to be creating textures larger than 4096, and if for some reason you do need textures larger than this you can just use multiple textures...

Allerek:
> e? people are using bigger textures for radar, does it break backward compatibility?

Breaking BWC for crash fixes was done before a lot of times, i think its ok to do this.

Lpsd:
> Or better just don't patch anything and put a giant warning on wiki about dimensions higher than 4096.

This does precisely nothing. See MySQL module

Lpsd:
See discussion on MTA development discord (just search for `4096`, lol) - I don't agree now to clamp this, as the underlying issue is more likely related to ``setPixel`` itself so we're just covering up the issue here.

The texture size limit comes from GPU, not the DX9 specification (100% sure it used to be in the specification by Microsoft, but later removed).

We need to investigate #287 properly

PR #2741, opened 2022-09-15
Train speed fix

### LINKED ISSUE #2732: setTrainSpeed vs getTrainSpeed
### Describe the bug

When you use setTrainSpeed( ) you are able to set it to a positive or negative number.

When using getTrainSpeed( ) it will not return a negative number.
When the speed is set below 0 getTrainSpeed() will only return 0.

To be able to make a train that is reversible and have all cars match the engine speed this function must return the negative value that was set to the train using setTrainSpeed( )

### Steps to reproduce

create train
set speed to a negative value using setTrainSpeed( )
output value using getTrainSpeed( ) to see only a 0.

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

Allerek:
Fixes #2732

Most likely breaks BWC, should be merged to 1.6.


Allerek:
Somehow changes from last PR introduced here, will remove them.
Edit:fixed

Allerek:
Closing due to being incorrect. Might create another PR in future.

PR #2743, opened 2022-09-15
Recalculate frame rate limit on fps_limit cvar change

Merlin:
This will force update fps limit while connected to a server.


botder:
I would prefer to have an update-limit for that to prevent players from abusing this by *lagging* their game on purpose. Feel free to disestablish my concerns here.

PlatinMTA:
> I would prefer to have an update-limit for that to prevent players from abusing this by *lagging* their game on purpose. Feel free to disestablish my concerns here.

There are other similar ways to lag your game, I doubt anybody would exploit this for that purpose, but since it's likely I'm not against the idea

patrikjuvonen:
> > I would prefer to have an update-limit for that to prevent players from abusing this by *lagging* their game on purpose. Feel free to disestablish my concerns here.
> 
> There are other similar ways to lag your game, I doubt anybody would exploit this for that purpose, but since it's likely I'm not against the idea

We encourage preventing exploitation since there is no harm in that and makes exploiters life more miserable.

Merlin:
I don't think calling CCore::RecalculateFrameRateLimit will result in much more CPU work that would be enough to be used as an efficient way to lag your client.

However, I added an update-limit of 500ms. 
That should be enough to stop anyone trying to abuse it but low enough to hopefully not annoy players.

botder (approved):

PR #2745, opened 2022-09-16
Add dynamic ped shadows

### LINKED ISSUE #2736: Dynamic ped shadows

In MTA, peds (and players) are always using static shadows. In single player, it's based on visual quality option.

Enable dynamic shadows for peds and players and add new checkbox in visual settings for that.

_No response_

### Additional context

I requested this feature many years ago at bugs.mtasa.com and it was rejected but to be honest I still don't quite understand why. I think it's possible to implement it.

### Security Policy

lopezloo:
This PR adds dynamic ped shadows feature found originally in single player version of GTA:SA. It can be enabled via new "Dynamic ped shadows" checkbox found in visual settings. This checkbox is enabled only if visual quality option is set to "high" or "very high". It's disabled by default.

In single player, these shadows are enabled for player ped only when quality is set to "high". When quality is set to "very high", these are enabled also for non player peds. I ditched that and this checkbox enables it for all peds. The limit is 16 shadows, after that game starts rendering blob shadows.

These shadows - just like blob shadows - do not render on custom placed objects (see #450).

Value of this option can be retrieved in scripts via `dxGetStatus().SettingDynamicPedShadows`.

In addition, this PR fixes vehicle blob shadows and lights not rendering on some objects when such vehicle is unoccupied (example position: `2463.023, -1713.818, 13.508`). Similar thing was happening to these shadows aswell which is also fixed.

Video: https://youtu.be/7wvIl9jzEE0

### DIFF
--- a/Client/core/Graphics/CRenderItemManager.cpp
+++ b/Client/core/Graphics/CRenderItemManager.cpp
@@ -805,6 +805,11 @@ void CRenderItemManager::GetDxStatus(SDxStatus& outStatus)
         outStatus.settings.bGrassEffect = false;
     }
 
+    if (outStatus.settings.iFXQuality < 2)
+    {
+        outStatus.settings.bDynamicPedShadows = false;
+    }
+
     // Display color depth
     D3DFORMAT BackBufferFormat = g_pDeviceState->CreationState.PresentationParameters.BackBufferFormat;
     if (BackBufferFormat >= D3DFMT_R5G6B5 && BackBufferFormat < D3DFMT_A8R3G3B2)
--- a/Client/core/Graphics/CRenderItemManager.cpp
+++ b/Client/core/Graphics/CRenderItemManager.cpp
@@ -783,6 +783,7 @@ void CRenderItemManager::GetDxStatus(SDxStatus& outStatus)
     outStatus.settings.bHighDetailVehicles = false;
     outStatus.settings.bHighDetailPeds = false;
     outStatus.settings.bCoronaReflections = false;
+    outStatus.settings.bDynamicPedShadows = false;
 
     CVARS_GET("streaming_memory", outStatus.settings.iStreamingMemory);
     CVARS_GET("volumetric_shadows", outStatus.settings.bVolumetricShadows);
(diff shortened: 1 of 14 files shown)


Client/core/Graphics/CRenderItemManager.cpp
```diff
@@ -803,6 +805,11 @@ void CRenderItemManager::GetDxStatus(SDxStatus& outStatus)
         outStatus.settings.bGrassEffect = false;
     }
 
+    if (outStatus.settings.iFXQuality < 2)
+    {
+        outStatus.settings.bDynamicPedShadows = false;
+    }
+
```
botder, line 812: According to your code, a user could run `dynamic_ped_shadows=1` to enable the ped shadows and see them, but in dx-status you would still see it disabled.

botder (changes requested):

Client/core/Graphics/CRenderItemManager.cpp
lopezloo: It's not gonna work if FX quality is lower than "high". GTA does a check somewhere.

botder (approved):

Einheit-101:
This is amazing.

C0DELEVEL:
@lopezloo These shadows are also displayed for peds/players whose alpha is 0.
It is worth correcting this because on some gamemodes it can give a significant advantage.

lopezloo:
> @lopezloo These shadows are also displayed for peds/players whose alpha is 0. It is worth correcting this because on some gamemodes it can give a significant advantage.

It behaves the same way as blob shadows. There is even [nice screenshot on wiki presenting how changing ped alpha looks](https://wiki.multitheftauto.com/wiki/SetElementAlpha) - shadow and held weapon aren't being affected. So I don't think we should hide these if ped alpha is set to 0.

Blob shadows can be hidden via shaders though and these dynamic shadows not so indeed there may be introduced advantage if scripter have hided ped blob shadows and doesn't want shadows displayed for peds.

If that's critical issue then scripters can kick players with enabled ped dynamic shadows currently.

We could introduce new function to control displaying of shadows for all or specific peds/elements. See:
- #2777

PR #2746, opened 2022-09-16
Disable camera collisions for detached vehicle parts and projectiles

### LINKED ISSUE #2434: Disable camera collisions for detached vehicle parts and projectiles
I propose disabling camera collisions for detached vehicle parts and projectiles by default because it's annoying and looks like oversight from R* side.

Make these objects non collidable with camera.

**Additional context**
GTA already disables camera collisions for other small objects like fences so I think this proposal fits very well.

lopezloo:
Disables camera collisions for detached vehicle parts and projectiles. Resolves #2434.


Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -7046,3 +7052,65 @@ void _declspec(naked) HOOK_CAutomobile__dmgDrawCarCollidingParticles()
...
+
+    return true;
+}
+
+void _declspec(naked) HOOK_CCollision__CheckCameraCollisionObjects()
+{
+    _asm
+    {
+        // Restore instructions replaced by hook
+        jz      loc_41AC26
```
sbx320: `jz RETURN_CCollision__CheckCameraCollisionObjects_2`?

sbx320 (dismissed):

Client/multiplayer_sa/CMultiplayerSA.cpp
lopezloo: But it can jump to label only?

lopezloo: I have tweaked it a bit.

patrikjuvonen (approved):

PR #2761, opened 2022-09-21
GameSA cleanup #1

Merlin:
Removes a bunch of unused code
The only real change is a call to CGameSA::DisableRenderer is removed from CGameSA::Reset.
There is no call to DisableRenderer that actually disables it and I also did not found any changes to that address anywhere else.


CrosRoad95:
i'm not sure if it is a good move, i would rather remove all `#include` to unused files, someone in the future may want to use these functions

botder (approved):

PR #2765, opened 2022-09-23
Use MTA screenshot function for photos taken by in-game camera weapon

### LINKED ISSUE #2759: Camera (weapon) image compression

In this case, the problem is that photos taken by camera are highly compressed. Propably due to limited space on PS2 memory cards. 
F.e. one of the below images is Screenshot(F12) and the other one is actual photo.
Screenshot is around ~2MB, while photo is around 200 KB
[Screenshot]
mta-screen_2022-09-20_21-24-25
[Photo]
gallery3

Prevent photo compression, its no longer needed due to high disk capacities now.

Instead of trying to fix GTA:SA way of taking photos, just recreate entire function from scratch using modern libraries and solutions.

### Additional context

_No response_

### Security Policy

lopezloo:
Resolves #2759.

Uses MTA screenshot function for photos taken by in-game camera weapon. By doing that we can have a lot better photo quality. Additionaly, GUI and DX elements aren't being captured. Photos are saved into same directory as before - `C:\Users\ccw808\Documents\GTA San Andreas User Files\Gallery`. File name uses same format as normal MTA screenshots - ex. `mta-screen_2022-09-23_17-04-22.png` which I think is better than previous `GalleryX.jpg`. After every photo taken, message is being printed into console with file path (just like for normal screenshots).

Additionaly it slighty refactors screenshot related code.


patrikjuvonen:
There appears to be an undocumented visual regression: camera zooms out on shot. Can this be fixed?

Single player
https://user-images.githubusercontent.com/22572159/210175976-15760ca7-7130-41f9-a991-241e618c1d7b.mp4

Your patch
https://user-images.githubusercontent.com/22572159/210175978-7a8a348f-5d7c-47b2-9114-79197431afeb.mp4

Does not happen on latest nightly.

lopezloo:
That's funny because it looks like it's not related to my changes at all. I just have built master (e3b489c495e1e5aa19bb22dc52754fadbc87f130) and it seems to be happening too. Could you please try to reproduce it on custom built master? I have set `MTASA_VERSION_TYPE` to `VERSION_TYPE_CUSTOM`, release build.

I have also built it with:
```c++
#define MTASA_VERSION_TYPE          VERSION_TYPE_UNSTABLE
#define MTASA_VERSION_BUILD         21591
```

and seems to be also happening.

patrikjuvonen:
> That's funny because it looks like it's not related to my changes at all. I just have built master ([e3b489c](https://github.com/multitheftauto/mtasa-blue/commit/e3b489c495e1e5aa19bb22dc52754fadbc87f130)) and it seems to be happening too. Could you please try to reproduce it on custom built master? I have set `MTASA_VERSION_TYPE` to `VERSION_TYPE_CUSTOM`, release build.

I didn't compare with master at the time, interesting. This seems to only happen on custom builds (including release), but not on nightlies. I'll try to find a reason for this odd behavior. Thanks for looking into it, sorry for alarming you :D

patrikjuvonen (dismissed):

Pirulax:
@patrikjuvonen So it can be merged?

patrikjuvonen:
> @patrikjuvonen So it can be merged?

In 1.6, yes.

lopezloo:
Converting to draft for a moment. I'm looking forward tweaking one thing.

lopezloo:
I have tweaked this PR to save camera screenshots on main thread now. Normal screenshots (initiated by F12) are saved on seperate thread as before. By saving image on main thread we can avoid creation of glitched pictures like this: https://i.imgur.com/reJbTKD.png (done by very fast clicking aim + fire) (screenshot was saved too late).

Pirulax:
That is hacky. The issue is that the data from `m_pData` is copied only when the other thread starts.
Solution: Always copy that data on the main thread (this way even F12 s.c.'s wont be buggy), and do the rest on a separate thread.


Also, the way data is allocated is totally garbage, so will have to improve that, but that might be out of scope for now.

lopezloo:
Actually it looks like buffer is being overwritten because I have removed `ms_bIsSaving` check from `CScreenShot::CheckForScreenShot()`. Will test it a bit more.

lopezloo:
Looking at code further... Aren't pixels already being copied on main thread currently? In `CRenderTargetItem::ReadPixels()` (called from `CScreenGrabber::GetBackBufferPixels()`? That copy in seperate thread seems to be only an conversion for libpng data format. I moved it into main thread and it still happens. It's very hard to reproduce this glitch by the way. It looks like MTA screen saving code is actually working properly but it's some kind of race condition. GTA saves these screenshots in `CPostEffects::Render()` on next frame. I am currently saving them post GTA frame. I could hook here but I wanted to keep things simple.

I will investigate it more.

lopezloo:
Well... I managed to reproduce this glitch on `v1.5.9-release-21554`: https://user-images.githubusercontent.com/7338099/220678445-1cfae197-d4cf-4c93-8c2e-589c4bc9bff3.jpg (gallery image created by GTA)

lopezloo:
Ok. Since I was able to reproduce this glitch on nightly build too, I'm reverting 34face5f4fd889098b5243431adefb991a7f678b. I haven't tried reproducing it in single player but it doesn't really matter in that case though. This glitch doesn't happen during normal gameplay. You have to try really hard* to produce such glitched picture on both nightly build and this PR. While I was trying to reproduce this glitch I also managed to glitch running task/animation so such bugs happen apparently. I'm not going to do anything with that in this PR.

\* Random-like repro:
1. Take camera.
2. Sprint for a few seconds.
3. Stop sprinting, but still run.
4. During transition from sprinting to running: Stop running and press AIM + FIRE very fast to produce photo. It should flash.

In last commit I also have:
- changed method of hiding GUI for screenshot (now we basically take camera screenshot before MTA draws GUI instead of using ugly conditions to check if UI should be drawn)
- cleaned code a bit
- started clearing `ms_ScreenShotBuffer` after screenshot
- 1s timeout is now seperate for F12 and camera screenshots
- moved `screenshot taken` console message print into save thread; also it won't get printed if something fails now
- removed calls to `CHud::Disable()`, it should be fine without that

PR #2766, opened 2022-09-24
Add nullptr checks in CWorld::FindObjectsKindaCollidingSectorList

Merlin:
Adds nullptr checks to CWorld::FindObjectsKindaCollidingSectorList and logs the model id of the faulty model.

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
@@ -1783,6 +1783,49 @@ static void _declspec(naked) HOOK_CPlaceName__Process()
     }
 }
 
+static void LOG_CWorld__FindObjectsKindaCollidingSectorList(unsigned int modelId)
+{
+    CBaseModelInfoSAInterface* pModelInfo = ((CBaseModelInfoSAInterface**)ARRAY_ModelInfo)[modelId];
+    if (!pModelInfo)
+    {
+        LogEvent(840, "Model info missing", "CWorld__FindObjectsKindaCollidingSectorList", SString("Corrupt model: %d", modelId), 5601);
+        return;
+    }
+
+    if (!pModelInfo->pColModel)
+    {
+        LogEvent(840, "Col model missing", "CWorld__FindObjectsKindaCollidingSectorList", SString("Corrupt col model: %d", modelId), 5601);
+    }
+}
+
+#define HOOKPOS_CWorld__FindObjectsKindaCollidingSectorList 0x56508C
+#define HOOKSIZE_CWorld__FindObjectsKindaCollidingSectorList 0xA
+static const unsigned int RETURN_CWorld__FindObjectsKindaCollidingSectorList = 0x565096;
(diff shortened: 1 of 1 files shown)


Dutchman101:
This is a test fix for 1 of the most popular gta_sa crashes in MTA (offset: 0x00165096), to avert the crash if the server or GTA mods load a custom model with corrupt collision

00165096 (*) = clump.base.m_pColModel->m_Sphere.sphere*.m_fRadius
```
CWorld::FindObjectsKindaCollidingSectorList

corrupt collision model (one or multiple col spheres)

called from RpClumpForAllAtomics```

botder (approved):

Pirulax:
If not there, it'll crash somewhere else.
This is just a workaround, why not fix the root cause?

Dutchman101:
> If not there, it'll crash somewhere else

I don't think so, im pretty sure this is meant to result in not loading the corrupt model or not load its collision and move on? At least that was the request for how this patch should work.

Additional info that may be relevant to understanding: ```FindObjectsKindaCollidingSectorList generates a list of objects that are close to the entity supplied in the 2nd arg. If modelinfo or colmodel is null it will not include these objects in the list. It will just skip these```

As the event is logged, we will be able to see how it works out by keeping track of crash stats and logs of what happens to a client after it has logged this crash being averted. It's a test fix after all, we will find out if it works, or crashes somewhere else, soon enough.

> This is just a workaround, why not fix the root cause?

The root cause is a corrupt custom model.. all we got are crash stats (not a reproduction resource) to work with, averting the crash is easier on stability and MTA support, than making every player and server owner aware that it's their own fault by loading a corrupted model. It was already easy to figure out by a web search on that offset, but yet there were too many of these crashes in practise.

For years, we are dealing with a trend of not having any devs/contributors that are willing to reverse and understand everything about corrupted model crashes, to modify the GTA model loading mechanisms ('fix the root cause' in your words) to be able to handle this type of corruption. So the short term solution at hand to reduce player crashes is averting them, as has been established project practise for these very reasons.

If a model doesn't load as the server owner is adding it, they will know something is wrong with it and look into it again. The event is logged as 'Corrupt model' which we can also advise them if they ask for support about the model or its collision not being loaded, but the main advantage is reducing crashes from popular servers that already use a model corrupted in this way at places, persistently causing their players to crash

Pirulax:
The fix would be to add proper error checking of models when loading them.
There's already some checking done in the netc module, but that's mainly just security, not actual error checking.
The function used to link collision models to model infos is `CFileLoader::LoadCollisionFile`, and probably something else too.
I also managed to crash when `CColStore::LoadAllCollision` loaded models in the incorrect order, so the order is important.

Dutchman101:
@Pirulax
I get your point now. Let's see if this change has the desired effect, as per my explanation about tracking it. Feel free to make a revision patch to do it in the way you have in mind, but for now the main goal is to stop a crash that has been in top 10 for too many years, so if this already works and no one invests time in doing it in another way we just keep this fix.

Pirulax:
I agree with you leaving this in though.
Won't hurt anybody.

Issue #2570, opened 2022-03-12
getPedControlState returns true for zoom_out when zooming in, and vice versa


bum8hj:
**Describe the bug**
Zoom controls are inverted
- When **zooming in** with the sniper, `getPedControlState(localPlayer, "zoom_out")` returns true.
- When **zooming out** with the sniper, `getPedControlState(localPlayer, "zoom_in")` returns true.

**To reproduce**
Run this code and try zooming in and out with your sniper.
```lua
addEventHandler("onClientRender", root, function()
    dxDrawText("zoom_in: " .. (getPedControlState(localPlayer, "zoom_in") and "true" or "-"), 500, 500)
    dxDrawText("zoom_out: " .. (getPedControlState(localPlayer, "zoom_out") and "true" or "-"), 500, 515)
end)
```

**Expected behaviour**
When **zooming in**, `getPedControlState(localPlayer, "zoom_in")` should return true, and vice versa.

**Video**
https://github.com/multitheftauto/mtasa-blue/assets/27828549/3b0d3b09-b270-4fc3-a7d8-6111e9147c25
(mirror: https://i.imgur.com/YTkWL7n.mp4)

**Version**
- **Server & Client** v1.5.9-release-21135
- Windows 11 (if relevant)

**Additional context**
Somewhat off-topic, but I think neither of the zoom controls should return true if we use the scroll wheel when we're not aiming. Like when switching weapons, etc.
**Edit:** probably related to issue #2275

Issue #2275, opened 2021-07-07
toggleControl next_weapon and previous_weapon takes sniper zoom with Q and A!


mihayy5:
**Describe the bug**
Toggling *next_weapon* and *previous_weapon* controls off will also toggle sniper zoom off with Q and A keys!
*Note: Mouse wheel will still zoom in/out
next_weapon = zoom_in **&** previous_weapon = zoom_out*

**To reproduce**
1. toggleControls off.
2. Take sniper.
3. Press Q or A to zoom in/out.
4. Doesn't work.

**Expected behaviour**
Expected to zoom in or out depending on the key.

**Version**
Latest version from the MTA website: 1.5.8
Server: 1.5.8-9.20881.0
Client: v1.5.8-release-20881

androksi:
I can confirm the issue.

Allerek:
Its most likely due to this check
https://github.com/multitheftauto/mtasa-blue/blob/70272f9a91f3217963239df55ca103f8dff4b566/Client/core/CKeyBinds.cpp#L1249
I'll check more about that later today.

ghost:
If i recall correctly disabling jump won't let you zoom as well when holding LSHIFT.

Allerek:
https://github.com/multitheftauto/mtasa-blue/blob/70272f9a91f3217963239df55ca103f8dff4b566/Client/core/CKeyBinds.cpp#L1325

^If we set the breakpoint here, with `if` statement to break only 
"if action == ZOOM_IN"
and try to press Q, it wont break for some reason(breaks on if action == PREVIOUS_WEAPON). It looks like GTA Hardcoded 'NEXT/PREVIOUS Weapon' to ZOOM_IN/OUT. 

Maybe we should make an IF statement with checking if player is using scope/zoom? Share your opinions.

mihayy5:
I'm not used to cpp, but in the line you shared there is this check that checks 

if (action == action) 

But where does actually 'action' get defined ?

Allerek:
> I'm not used to cpp, but in the line you shared there is this check that checks
> 
> if (action == action)
> 
> But where does actually 'action' get defined ?

`pTemp` is temporary pointer to they keyBind that holds the same action as main key that was pressed, so `->action` is its action, and 

`action` is action of original key
https://github.com/multitheftauto/mtasa-blue/blob/70272f9a91f3217963239df55ca103f8dff4b566/Client/core/CKeyBinds.cpp#L1315

[...]

Issue #515, opened 2018-09-20
Can't unbind wheel (zoom in/out)


ArranTuna:
**Describe the bug**
Cannot unbind mouse wheel "zoom".

**To Reproduce**
client code:
```lua
bindKey('aim_weapon', 'both', function(key, state)
    toggleAllControls(state ~= 'down')
    toggleControl('aim_weapon', true)
end)
```

Do this in game:
1. take sniper, push aim button
2. use wheel

You can zoom in and out.

**Expected behavior**
Be unable to zoom in and out.

**Additional context**
Gallardo9944 said:
> Don't know if it's related or not, but if you're not inside a vehicle and your camera target is set to someone with a vehicle, scrolling mouse wheel will simulate look-left and look-right tasks, meanwhile if you're in vehicle and spectate another player in vehicle, mouse wheel will do nothing. And it doesn't look like there's any way to prevent such behavior from within Lua.

From https://bugs.mtasa.com/view.php?id=9069

qaisjp:
At first look at this it seems that this is only a problem with the mouse wheel - if you try to reproduce using the pgup/down keys it won't work.

```lua
bindKey("zoom_in", "both", function(key, state) outputChatBox(key..state) end )
```

The above code will execute for the page keys and won't zoom - this is the intended behaviour.

You don't get any output using the scroll, so that's probably where things are going wrong - MTA isn't catching the wheel zoom there.

This might have something to do with the following snippet of code (core/CKeyBinds.cpp):

![image](https://user-images.githubusercontent.com/923242/45841925-27873200-bd13-11e8-8ce6-e3fa3653a993.png)

Removing those lines will cause the bind to be called. See https://github.com/multitheftauto/mtasa-blue/commit/3f5d2118663bd3daa2359f7b6ba67e5e020086d4

PR #2767, opened 2022-09-25
Zoom In/Out key binding fixes

### LINKED ISSUE #524: Zoom in/out binds are inverted
**Describe the bug**
Well, basically in the 'Bindings' tab in 'Settings' the 'Zoom in' and 'Zoom out' bindings are named incorrectly, they should be switched.

Apparently 'Zoom in' is meant to decrease the field of view, whilst 'Zoom out' is meant to increase the field of view, it is like this in Single Player.

**To Reproduce**
Get sniper rifle, aim it, use zoom stuff.

**Expected behavior**
To zoom the right way.

**Screenshots**

Unstated.

Unstated.

**Additional context**
From https://bugs.mtasa.com/view.php?id=4781

darkdrifter:
Resolves #524

Hello,

This is my first contribution to any open source project, tried to follow the guidelines and coding style but I may have missed some things. Don't hesitate to tell me if I did.

Threes fixes are included in this pull request:
- Fix for the inverted zoom in/out bindings. For some reason, _CControllerState RightShoulder2_ controls Zoom Out, while the _CControllerState LeftShoulder2_ controls the Zoom In, I don't know if it's the game behavior on PS2.
- Fix for the Zoom In/Out behavior being also controlled by the Weapon Previous/Next key binds. This behavior isn't present in the base game so it shouldn't be in MTA.
- ~~Thanks to the fix just above, we can remove a hack that disconnected Zoom In/Out behavior using mouse wheel to try and fix zoom being applied twice. I think it was actually due to the fact that _Weapon Previous/Next_ was applied first (if bound on the scroll wheel) and them _Zoom In/Out_ gets applied and adds another zoom on top of it. With that hack removed, we can now bind mouse scroll wheel for the zoom. On master if you try to bind the scroll wheel only on the zoom in/out, you'll observe that the zoom isn't working.~~ (moved into #2769)

Don't mind the first two commits, it was a first try to fix the issue that I reverted after I dug up further and found what I think is a better fix.


lopezloo:
> I think it was actually due to the fact that Weapon Previous/Next was applied first (if bound on the scroll wheel) and them Zoom In/Out gets applied and adds another zoom on top of it.

I haven't tested your PR yet but I have unbinded all keys (via `unbind all` command) except `aim_weapon` and I'm still being able to zoom in/out sniper using scroll wheel. (tested on `v1.5.9-release-21326`)

darkdrifter:
I tried with `unbind all` on `v1.5.9-release-21331` and I can't zoom using mouse scroll until I bind mouse scrolling to `Weapon Next` and `Weapon Previous`
I'm testing in Map Editor in "Full Test" BTW
EDIT: Tried in a online server, same behavior.

lopezloo:
Looks like it actually depends on single player control settings. I have unbinded it in single player and it doesn't work anymore. It shouldn't work like that.

single_player

darkdrifter:
Yeah indeed when I set the mouse scrolling in single player I get a smooth zoom in/out even with everything unbound in MTA. And if I bind mouse scrolling in MTA I get a jerky "double zoom", probably the thing that the hack tried to fix. 
I'll tried on my build with this PR, I don't have the said "double zoom" if I bind the mouse scroll on solo and MTA, though I still have mouse scrolling when I `unbind all` if I have it set in solo. So it's an improvement but there is still something weird going on.

darkdrifter:
I found the issue with the scrolling that caused it to be affected by single player. It seems that GTA SA uses Direct Input to get the scroll wheel value when using mouse scroll wheel on zoom. It's probably to have smoother zoom, as now that I have hooked the zoom value on direct input, the single player no longer affects MTA but the scrolling is a bit harsh.
As I described in the comment where I placed the hook, a better option could be to instead give this data only when we want it (when zoom is bound to scroll wheel in MTA) to retain the proper smooth zoom, but it would require also successfully forcing the game to use this data even when it is not bound in single player, but I haven't looked closely at this option yet.

lopezloo:
That mouse wheel zoom logic is defined in `CCam::Process_M16_1stPerson()`. It calls `CControllerConfigManager::GetMouseButtonAssociatedWithAction(&ControlsManager, PED_SNIPER_ZOOM_IN)`. The question is - why does it seems to be returning something if it's not defined in MTA binds? We need to investigate it more since it would be nice to have smooth zoom.

lopezloo:
I suggest to remove that mouse wheel zoom fix and create seperate PR for that instead.

Related issues:
- #2275 
- #2570
- #515 

Jump bind also triggers zoom in (it doesn't happen in single player).

bum8hj:
> Jump bind also triggers zoom in (it doesn't happen in single player).

This is slightly off-topic, but I believe jumping also forcefully re-enables `aim_weapon` when it's disabled with `toggleControl`.

darkdrifter:
Hello, I reverted the last commit that I will move into a WIP PR that will look into the whole zoom behavior issues (to the best of my abilities).
I don't know how to quickly reproduce issue #2275 or #2570 to see if it's still present with this PR, but it would be nice to know if it fixes them indirectly.
@lopezloo we stop this PR with only this fix then and integrate it with its improvements even if some flaws are still present ?

lopezloo:
> @lopezloo we stop this PR with only this fix then and integrate it with its improvements even if some flaws are still present ?

I just would like to merge fix for inverted controls first. Please restore that zoom check which MTA used to have (and then remove this condition in your other PR).

> This is slightly off-topic, but I believe jumping also forcefully re-enables aim_weapon when it's disabled with toggleControl.

Unable to reproduce that.

darkdrifter:
> Please restore that zoom check which MTA used to have (and then remove this condition in your other PR).

Done ;)

lopezloo (approved):
Looks good to me. Thank you.

I have slighty edited first post.

PR #2769, opened 2022-09-28
Fix weapon zoom behavior

darkdrifter:
This PR is aimed to fix several issues present in the weapon zoom right now.
Created after discussions and findings in PR #2767 

- Issue #515, which is probably caused by the fact that single player bindings currently affects MTA (if zoom is bound on mouse scroll wheel in solo, it will be usable in MTA, even if not bound or controls are disabled) currently fixed by this hotfix
- Fixing zoom when using mouse scrolling, which is currently different than in single player, due to the fact that MTA doesn't seems to utilize the Direct Input mouse wheel values value like single player to control the zoom


lopezloo:
Zoom seems to be jittery at current state. IMO we just should call `CControllerConfigManager::SetMouseButtonAssociatedWithAction()` and let GTA handle wheel zooming.

darkdrifter:
> Zoom seems to be jittery at current state. IMO we just should call `CControllerConfigManager::SetMouseButtonAssociatedWithAction()` and let GTA handle wheel zooming.

I agree, but the function isn't implemented currently in MTA. Need some help from someone more experimented to add the function as I can't find the function pointer to use for it and then I could take a look to improve the behaviour.
I might need the `CControllerConfigManager::GetMouseButtonAssociatedWithAction` function also.

PR #2773, opened 2022-10-01
Clean up game_sa project round 5

Merlin:
Changes VOID to void and FLOAT to float.
Fixes a out of bounds read and removes some unused code.


Client/game_sa/CRopesSA.cpp
```diff
@@ -37,7 +37,7 @@ void CRopesSA::RemoveEntityRope(CEntitySAInterface* pEntity)
 {
     CRopesSAInterface* pRope = nullptr;
 
-    for (uint i = 0; i <= ROPES_COUNT; i++)
```
botder, line 40: Nice catch!

botder (approved):

PR #2783, opened 2022-10-10
Add getElementLighting()

samr46:
Adds lua interface to `GetLighting()` for peds / players, vehicles and objects.

Test resource: [elight.zip](https://github.com/multitheftauto/mtasa-blue/files/9748117/elight.zip)

It's more efficient and accurate than [processLineOfSight](https://wiki.multitheftauto.com/wiki/ProcessLineOfSight) (for this particular purpose).

Example of what it could be used for:

e-lighting-tests01

There is also `SetLighting()`. It works for peds if called every frame. Not sure if we need it tho. But could be added.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
@@ -1317,22 +1317,24 @@ std::variant<bool, float> CLuaElementDefs::GetElementLighting(CClientEntity* ent
         case CCLIENTPED:
         case CCLIENTPLAYER:
         {
-            CPlayerPed* Ped = static_cast<CClientPed*>(entity)->GetGamePlayer();
-            if (Ped) return Ped->GetLighting();
+            CPlayerPed* ped = static_cast<CClientPed*>(entity)->GetGamePlayer();
+            if (ped) return ped->GetLighting();
             break;
         }
         case CCLIENTVEHICLE:
         {
-            CVehicle* Vehicle = static_cast<CClientVehicle*>(entity)->GetGameVehicle();
-            if (Vehicle) return Vehicle->GetLighting();
+            CVehicle* vehicle = static_cast<CClientVehicle*>(entity)->GetGameVehicle();
+            if (vehicle) return vehicle->GetLighting();
             break;
         }
         case CCLIENTOBJECT:
         {
-            CObject* Object = static_cast<CClientObject*>(entity)->GetGameObject();
-            if (Object) return Object->GetLighting();
+            CObject* object = static_cast<CClientObject*>(entity)->GetGameObject();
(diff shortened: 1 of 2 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1308,6 +1310,33 @@ int CLuaElementDefs::GetElementAlpha(lua_State* luaVM)
...
+            if (Vehicle) return Vehicle->GetLighting();
+            break;
+        }
+        case CCLIENTOBJECT:
+        {
+            CObject* Object = static_cast<CClientObject*>(entity)->GetGameObject();
+            if (Object) return Object->GetLighting();
+            break;
+        }
+    }
```
botder:
```suggestion
    switch (entity->GetType())
    {
        case CCLIENTPED:
        case CCLIENTPLAYER:
        {
            CPlayerPed* ped = static_cast<CClientPed*>(entity)->GetGamePlayer();
            if (ped) return ped->GetLighting();
            break;
        }
        case CCLIENTVEHICLE:
        {
            CVehicle* vehicle = static_cast<CClientVehicle*>(entity)->GetGameVehicle();
            if (vehicle) return vehicle->GetLighting();
            break;
        }
        case CCLIENTOBJECT:
        {
            CObject* object = static_cast<CClientObject*>(entity)->GetGameObject();
            if (object) return object->GetLighting();
            break;
        }
        default:
            break;
    }
```

botder (approved):

PlatinMTA:
> e-lighting-tests01

How can you achieve this without ``setElementLighting``? I think that it should be added.

samr46:
> How can you achieve this without `setElementLighting`? I think that it should be added.

Simple pixel shader. This was the idea (to pass accurate lighting information to shaders).
As I mentioned in original post, existing SetLighting works only for peds. I didn't look deep into how lighting is processed for custom objects. Not sure if it's even worth it.

PlatinMTA:
> Simple pixel shader. This was the idea (to pass accurate lighting information to shaders). As I mentioned in original post, existing SetLighting works only for peds. I didn't look deep into how lighting is processed for custom objects. Not sure if it's even worth it.

I think if it's possible to not use shaders then that's the best solution, but if it only works for peds and it's too much a hazzle to implement then there's no need. I don't think many people will use this anyway. I will probably use this though.

gptrk0:
> > How can you achieve this without `setElementLighting`? I think that it should be added.
> 
> Simple pixel shader. This was the idea (to pass accurate lighting information to shaders). As I mentioned in original post, existing SetLighting works only for peds. I didn't look deep into how lighting is processed for custom objects. Not sure if it's even worth it.

Sharing an example code with the shader on wiki would be very useful for many people.

Issue #2782, opened 2022-10-10
Boats are slower on high FPS 


Brian9221:
### Describe the bug

Boats are significatly slower when you set a high FPS limit on yourself:

Take a Squalo or another fast boat, do

crun setFPSLimit(144) -> boat is slow
crun setFPSLimit(45) -> boat is faster

I tested this on CIT server, we have a speedometer that outputs vehicle speed on screen. To verify this bug you might need a separate script to check speed.

### Steps to reproduce

Take a Squalo or another fast boat, do

crun setFPSLimit(144) -> boat is slow
crun setFPSLimit(45) -> boat is faster

### Version

Client v1.5.9 release 21342 (Win 10 64bit)

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

PlatinMTA:
Also boats slide less distance on high fps. These two things are marked on the framerate fixes project.

lopezloo:
Fixed in 7c8a1ab93d091b06262749b482d6c8142ca69eea

PR #2784, opened 2022-10-11
Fix various high fps issues

Merlin:
Fixes fog moving too fast on high fps.
Fixes glass shards spinning and expanding too fast on high fps.
Fixes vehicles turning slowly in air on high fps.
Fixes vehicles stopping too fast on high fps.
Fixes vehicle cam moving too slowly to the back on high fps.
Fixes boats being slow on high fps. (#2782)

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
@@ -188,11 +188,10 @@ static void _declspec(naked) HOOK_CPhysical__ApplyAirResistance()
     }
 }
 
-static unsigned int RETURN_VehicleRapidStopFix;
 template <unsigned int returnAddress>
 static void _declspec(naked) HOOK_VehicleRapidStopFix()
 {
-    RETURN_VehicleRapidStopFix = returnAddress;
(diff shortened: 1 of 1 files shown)


Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -74,6 +73,134 @@ static void _declspec(naked) HOOK_CHeli__ProcessFlyingCarStuff()
...
+        fstp [esi+0x58]
+        jmp RETURN_CPhysical__ApplyAirResistance
+    }
+}
+
+static unsigned int RETURN_VehicleRapidStopFix;
+template <unsigned int returnAddress>
+static void _declspec(naked) HOOK_VehicleRapidStopFix()
+{
+    RETURN_VehicleRapidStopFix = returnAddress;
```
botder: Wouldn't it make more sense to turn this line into a static variable plus assignment?

Merlin: Changed it. For some reason I thought that you can not have a static variable in a naked function. But apparently you can. I also checked the assembly of release and debug build and it looks fine.

botder (approved):

PR #2787, opened 2022-10-11
Remove vehicle framerate fixes again

Merlin:
Can be reintroduced as a world special property at some point


lopezloo (approved):

lopezloo:
This PR reverted these changes from previously merged PR #2784 (`Fix various high fps issues`):
* `Fixes vehicles turning slowly in air on high fps.`
* `Fixes vehicles stopping too fast on high fps.`

Fernando-A-Rocha:
> Can be reintroduced as a world special property at some point

setGlitchEnabled is more appropriate because they are bug fixes as @lopezloo mentioned above.

q8X:
> setGlitchEnabled is more appropriate because they are bug fixes as @lopezloo mentioned above.

isn't possible to configure kOriginalTimeStep? would be great for DM race players

Issue #2213, opened 2021-04-29
engineReplaceModel is not replacing 3D markers


TFP-dev:
**Describe the bug**
Function engineReplaceModel is not replacing 3D marker model. GTA is caching marker models at address 0xC7C6DC, so replacing 3D marker model takes effect only when placing marker as normal object.

**To reproduce**
1. Use engineReplaceModel with id 1559 or 1316 or 1317.
2. Markers are not replaced if using createMarker, but they are replaced if using createObject.

**Expected behaviour**
Abbility to replace marker models.

**Screenshots**
![obraz](https://user-images.githubusercontent.com/72511827/116620030-8d171280-a941-11eb-8298-975706f7615b.png)

![obraz](https://user-images.githubusercontent.com/72511827/116620053-97391100-a941-11eb-8b2b-20d1e6abf077.png)

**Version**

- server/client v1.5.8-release-20845

**Additional context**

Lpsd:
Making the suggested changes could break backwards compatibility, whereby servers who are replacing those models don't expect the marker itself to change when created via createMarker (there would have to be an optional arg defaulting to false on whether to load custom marker models).

Then there's the question of whether it's even feasible (or worth it) to "fix" this. I have some doubts - and I also believe it would make more sense to solve this using Lua (make the marker invisible and put your custom model in the same position).

Personally I don't agree with changing / adding this, there's already simple enough ways to solve this using Lua.

PR #2794, opened 2022-10-20
Add engineReinitMarkers()

TFP-dev:
This function will reinit 3D markers. This can be useful if you want to replace marker 3D model. Fixes closed issue #2213.

**API:**
`engineReinitMarkers()`

**Test resource:** 
This resource replaces "arrow" marker and adds command `reinitMarkers`.
[marker_test_resource.zip](https://github.com/multitheftauto/mtasa-blue/files/9832216/marker_test_resource.zip)

**Screenshot:**
comp


CrosRoad95:
". This can be useful if you want to replace marker 3D model."
so, make if you replace 3d model of marker -> call "engineReinitMarkers", we don't use "reinitVehicles", "reinitObjects" ect after element will be replaced

TFP-dev:
> ". This can be useful if you want to replace marker 3D model." so, make if you replace 3d model of marker -> call "engineReinitMarkers", we don't use "reinitVehicles", "reinitObjects" ect after element will be replaced

But we use `engineRestreamWorld`. Currently servers can use fact that you can't replace marker model so `engineReinitMarkers` is backwards compatible.

CrosRoad95:
you actually fixing a bug, everyone will happy if suddenly markers will work as intended

TFP-dev:
> you actually fixing a bug, everyone will happy if suddenly markers will work as intended

Check issue #2213

lopezloo:
This issue applies to `cylinder` markers too, right?

sbx320 (commented):
Wouldn't it be more useful to apply this whenever we replac a marker model? That way scripters wouldn't need to work around it.

TFP-dev:
> Wouldn't it be more useful to apply this whenever we replac a marker model? That way scripters wouldn't need to work around it.

Currently you can use fact that you can't replace markers, for example you can replace id 1559 with the building model and use it instead of allocating new model id with `engineRequestModel` and markers will stay intact.

Client/game_sa/C3DMarkersSA.cpp
```diff
@@ -122,3 +122,15 @@ C3DMarker* C3DMarkersSA::FindMarker(DWORD Identifier)
     }
     return NULL;
 }
+
+void C3DMarkersSA::ReinitMarkers()
+{
+    typedef int(__cdecl * Function_ShutdownMarkers)();
```
sbx320: Should be using rather than typedef

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2058,3 +2059,9 @@ bool CLuaEngineDefs::EngineRestreamWorld(lua_State* const luaVM)
     g_pClientGame->RestreamWorld();
     return true;
 }
+
+bool CLuaEngineDefs::EngineReinitMarkers(lua_State* const luaVM)
```
sbx320: No need to return a bool here, if it's always `true`. Also if the lua_State* arg isn't used, it doesn't need to be specified. 

The argparse can handle `void [func]()`.

sbx320 (dismissed):
Some minor comments regarding argparser usage

lopezloo:
> Wouldn't it be more useful to apply this whenever we replac a marker model? That way scripters wouldn't need to work around it.

Maybe that could be implemented and documented as 1.6 backwards incompatible change.

botder:
Superseded by pull request #3059.

PR #2813, opened 2022-11-10
Edit Quick Connect to reconnect by default

### LINKED ISSUE #2469: Remove or Replace Quick Connect button
Quick Connect is only there for someone to accidentally press it and almost no one knows that you can r-click it to connect to the last server you played on.
We should either remove it or replace the functionality(and text) to connect to the last server user played on.

sebastianiszek:
Makes the reconnect function be the default instead of opening the server browser when pressing quick connect. 
Updates  _beta_qc_rightclick_command to be empty by default and only execute if it isn’t empty.
Updates status box to say “Reconnecting” instead of “Connecting” if reconnecting.

Resolves #2469: Remove or Replace Quick Connect button


bum8hj:
• I would prefer if we could still right click to execute a certain command.

I can't think of another suitable default behavior. So, the user would just have have to set it using `_beta_qc_rightclick_command=`


• Wouldn't it be better to change the option's name to '**Reconnect**' rather than Quick Connect?

• What should happen when there is no server to reconnect to? Should it just open the server browser like it currently does?

sebastianiszek:
> • I would prefer if we could still right click to execute a certain command.
> 
> I can't think of another suitable default behavior. So, the user would just have have to set it using `_beta_qc_rightclick_command=`
> 
> • Wouldn't it be better to change the option's name to '**Reconnect**' rather than Quick Connect?
> 
> • What should happen when there is no server to reconnect to? Should it just open the server browser like it currently does?

The default value is "127.0.0.1", so if you didn't join any servers before it will try to connect you to your local sever.

The command could stay with either no action or reconnect by default. Not sure what would be better. Edit: Actually thinking about it, it might be pretty neat if you could select your default server and right click would make it connect to that server or just open favorites tab by default?

I like the name quick connect to be honest, it could say "Connecting to the server you last joined..." after your press it instead of "Connecting to... (ip:port)". It might make it more clear as to what it does

sebastianiszek:
Updated status box to say "Reconnecting to... host:port" instead of "Connecting to... host:port" when reconnecting.
Kept _beta_qc_rightclick_command and made it empty by default. Made it execute only if the command isn't empty.

Fernando-A-Rocha:
Honestly, for what you want to do in this PR, you should get rid of the `_beta_qc_rightclick_command` setting, and simply make **left-clicking** quick connect execute `'reconnect'`, and **right-clicking** open the quick connect menu.

I'll make a new PR for this ;)

Fernando-A-Rocha:
@botder stated:

> I wouldn't mind if we swap the behavior of the button to reconnect on left-click and open the 'Local' tab of the server browser on right-click.

PR #2112, opened 2021-03-08
Move vendor/lua to a submodule

Pirulax:
I made a PR in which Qais suggested me to use create a repo for our Lua vendor, and do modifications there instead.
This PR implements his idea.
It contains all previous commits back to 2008.

TODO: Test if everything works.

PR #1246, opened 2020-02-08
Bullet physics

CrosRoad95:
Adds bullet physics engine to mta.

Plan changed.
Initial pr ( this pr ) will contain:

physicsCreateBoxShape, physicsCreateRigidBody, physicsCreateStaticCollision, physicsGetDebugLines + destroyElement, setElementPosition, setElementRotation support for rigid body and static collision

[tests v1.zip](https://github.com/multitheftauto/mtasa-blue/files/8459389/tests.v1.zip)

More details at mta development discord

<details>
<summary>More info</summary>
<p>
= SHERED =
physics-box-shape physicsCreateBoxShape( Dimensions size ) - creates box, cube shape. Minimum size of each dimension: 0.0001, maximum: 8196
physics-rigid-body physicsCreateRigidBody( physics-shape shape, RigidBodyOptions options ) - spawns a rigid body with given options, when spawned in air, it will immidently starts falling. Shape must be convex ( see below ). In future: eg.: some types of triangle mesh will not accepted, but compound-shape will even technically is not convex
physics-static-collision physicsCreateStaticCollision( physics-shape, Vector3 position, Vector3 rotation ) - spawns static collision, it will not affected by gravity, always stay in one place (like a platform)
table physicsGetDebugLines( Vector3 position, float radius ) - returns table {{fromX, fromY, fromZ, toX, toY, toZ, color}, ...} with collision wireframe at given position and radius. Send to client and loop to see current state of simulation. Color indicate current state of rigid body.

= CLIENT ONLY =
void physicsDrawDebug() - renders all debug lines, use in onClientRender

= Types =
RigidBodyOptions - table with accepted keys: {["mass"] = float, ["localInertia"] = Vector3, ["centerOfMass"] = Vector3, ["position"] = Vector3, ["rotation"] = Vector3}
Dimensions- can be x,y,z, or one number with all dimensions

= Hierarchy, type description:
physics-element - base of any physics element
physics-world-element : physics-element - element that has position and rotation
physics-shape : physics-element - base of all physics shape
physics-convex-shape : physics-shape - base of all convex shape
physics-box-shape : physics-convex-shape - box shape
physics-rigid-body : physics-world-element - shape affected by gravity ( eg: car, player, ball, some gta models )
physics-static-collision : physics-world-element - non moving collision ( eg: terrains, buildings, roads :D )

=== Extras:
physics-rigid-body and physics-static-collisions is always made of single physics-shape
[...]

PR #2815, opened 2022-11-12
Luau

reflectrpteam:
This is an attempt to add a new Lua version along with the default one. In this PR you can find an implementation of an interface allows to switch between _Lua 5.1_ and _Luau_ in runtime. Can be considered as stable but due to the complexity can potentially contain issues that are expected to be revealed after the release.

_lua5.1 and luau libraries are shared now for all build configurations. But Lua_Server and Lua_Client are static. It makes the system flexible on the one hand and resolves the name conflict on the other._

You can enable Luau for a specific resource for **client**, **server** or **both**.
```
<meta>	
	<lua client="luau" server="luau" both="luau" />	
</meta>
```
Due to the flexibility of the interface new versions of Lua can be added in the future.


CrosRoad95:
One thing from my side: if possible, download vendor files from luau repository to do not bloat mta codebase and make future updates easier,
example how you can do this you can find in mta bulletphysics pr :) it contains example for client and server side download of 3-rd party files
https://github.com/multitheftauto/mtasa-blue/pull/1246

reflectrpteam:
> 

Unfortunately, we cannot use the default Luau code in the future. Some adjustments required in the code.

CrosRoad95:
> > 
> 
> Unfortunately, we cannot use the default Luau code in the future. Some adjustments required in the code.

you can download luau and make a folder with patched files you copy onto luau files

This pr just show how much the mta team doesn't give a damn about this project

Xenius97:
Instead of < lua > could be better to set lua version per script.
Just like < script src="server.lua" type="server" lang="luau"/ >
If lang is not set script uses 5.1 as default.

Lpsd:
> Instead of < lua > could be better to set lua version per script.
> Just like < script src="server.lua" type="server" lang="luau"/ >
> If lang is not set script uses 5.1 as default.

This goes against the current implementation whereby each **resource** has its own Lua VM (on client and server)

MegadreamsBE:
> Instead of < lua > could be better to set lua version per script. Just like < script src="server.lua" type="server" lang="luau"/ > If lang is not set script uses 5.1 as default.

Disagree, if you have a need to mix Lua versions you should split it up in multiple resources.

MysterAnonimi:
Thanks

Disinterpreter:
IMHO pros and cons:

Pros:
1. Good syntax and language
2. It is really popular and developed version or Roblox's lua. Maybe we can use their tools in future
3. Knowing developers made this (this user account hide people which we work many years)


Cons:
1. It won't work for our luac.mtasa (but you can cache scripts)
2. Diffrent VMs, like you can't write on Lua and Lualu in one resource (but i think exports for multiple resources are working)

My opinion - lgtm.

PrototypeX:
Давно пора совершенствовать среду разработки в MTA:SA.

MegadreamsBE:
> It's long past time for development to take place in MTA:SA.

MTA is still thriving and will keep on thriving for quite some time. It's most definitely not past time for any development to take place nor is this the place to have such a discussion.

Lpsd:
> It's long past time for development to take place in MTA:SA.

Be the change you want to see in the world. MTA is open source

dmi7ry:
Not finished, but already usable
https://github.com/dmi7ry/Luau-definition-for-MTA

ERAGON007:
Any ETA for merging?

dmi7ry:
> Any ETA for merging?

I don't think it will ever be merged.

CrosRoad95:
As elders of mta said: if will work, it will be merged.

ERAGON007:
> 

Sometimes i don't know why TF is mta open-source if you can't do some improvements to codebase /:

MegadreamsBE:
> > 
> 
> Sometimes i don't know why TF is mta open-source if you can't do some improvements to codebase /:

You can, as long as the improvement makes sense and follows the guidelines. Testing and approval might just take a while. I personally do hope this makes it into the codebase as this will help with some of my projects.

Lpsd:
Rather than complaining, you need to actually contribute if you want to see significant changes in MTA.

Yes, it's an open source project. No, this does not mean you can push things out on a whim without proper discussion or testing.

The details of this PR are rather vague and we don't know the implications of accepting this. Actual testing (and more importantly, evidence of such testing) needs to be shown in order to advance this. You could do this yourself, if you really want to help!

The person(s) who will ultimately decide whether this gets accepted will be much more willing to spend time reviewing, if you provide sufficient details; help them to help you.

CrosRoad95:
> Sometimes i don't know why TF is mta open-source if you can't do some improvements to codebase /:

Mta team can merge everything, but later someone need to care about it

From my own experience i see how people are unlikely to test for example BP, feature does not implement risky features but still require testing that almost noobody ddid

tederis:
The problem lies in the lack of discussion from the maintainers. To date, there was not a single question or a remark about the PR from authorized members. Just terrifying and cold silence. And that is the reason why this PR is still just a draft.

Lpsd:
Do you genuinely believe there is enough information in the initial pull request for a maintainer to _want_ to consider picking this up? Why must the reviewer put in all the work for something you are wanting to implement?

dmi7ry:
Maybe because there is a noticeable community request for support for other versions of Lua, at least?

CrosRoad95:
> Maybe because there is a noticeable community request for support for other versions of Lua, at least?

Result: people are starting reimplement server, create custom forks to get what they want

dmi7ry:
> Result: people are starting reimplement server, create custom forks to get what they want

Yep. Therefore, I doubt that we will contribute anything else from ReflectRP to MTA, but MTA's bugfixes. Because it's just a waste of time.

CrosRoad95:
> > Result: people are starting reimplement server, create custom forks to get what they want
> 
> Yep. Therefore, I doubt that we will contribute anything else from ReflectRP to MTA, but MTA's bugfixes. Because it's just a waste of time.

I understand most of core devs are busy, but really? they didn't give feedback even for pull requests created YEARS ago

tederis:
According to my observations, contribution (PR in particular) in MTA is considered as something that needed exclusively to the contributor. But this manner of posing the question is destructive.

Lpsd:
Just food for thought. But try submitting a PR like this with such lack of documentation or testing in the "real world" and see if it doesn't get closed almost instantly

dmi7ry:
> Just food for thought. But try submitting a PR like this with such lack of documentation or testing in the "real world" and see if it doesn't get closed almost instantly

1. What kind of documentation do you mean?
2. We initially created it as a draft so that anyone can try it or test it.
3. We actively use it on ReflectRP.
4. It's easy to test. Just take any existing project and enable Luau for each resource. It will work without any problems.

Lpsd:
Documentation could cover almost anything, but in this instance I'd be referring to any notable changes (considering the size of the PR, regardless of whether most changes come from vendor) or regressions that may occur due to tests carried out; of which there are currently none.

Unfortunately your final point doesn't really add up considering the following note by the author:

> It can be unstable and tests required.

As I've said, proper documentation and testing would be best. At this point it doesn't really matter who does that, but if you want this in MTA it's certainly beneficial. Reviewers rely upon tests to ensure they're actually testing things properly/consistently compared to others - this is how it works pretty much everywhere.

ERAGON007:
Maybe i can do the testing stuff, can you pls just list them

CrosRoad95:
I want to admit that we don't really have any "test branch" which mta should have. we could use huge playerbase to test features, make in main menu so kind of information to switch to "testing" version of mta if you want to help mta bla bla bla.
There we could put not properly tested features, features that require more tests. Because mta have some servers in its disposals, they could be used to get at least approximite estimation of how much time people had spent testing them, what crashed ect. I know someone might want to help mta, he plays mta but he is not aware of people pr's

PlatinMTA:
> I want to admit that we don't really have any "test branch" which mta should have. we could use huge playerbase to test features, make in main menu so kind of information to switch to "testing" version of mta if you want to help mta bla bla bla.
> There we could put not properly tested features, features that require more tests. Because mta have some servers in its disposals, they could be used to get at least approximite estimation of how much time people had spent testing them, what crashed ect. I know someone might want to help mta, he plays mta but he is not aware of people pr's

Nightly?

Pirulax:
> > > 
> > 
> > 
> > Unfortunately, we cannot use the default Luau code in the future. Some adjustments required in the code.
> 
> you can download luau and make a folder with patched files you copy onto luau files

This.
Making manual patches all the time is not a great idea.
And I'm sure keeping Luau up-to-date is preferred.
Also, what are the changes made? Are they really necessary? 
Maybe we can do some workaround to avoid having to modify it.

> I want to admit that we don't really have any "test branch" which mta should have. we could use huge playerbase to test >? 
features, make in main menu so kind of information to switch to "testing" version of mta if you want to help mta bla bla bla.

I guess there are nightlies, but even for that the PR has to be tested (and be somewhat stable).

-----
#### So anyways, let's get the ball rolling.
I'd like this to be merged, unlike my attempt at LuaJIT (Also, Luau is better maintained, so it's superior anyways).
My idea for testing is as simple as taking some opensource mta mod, running it, and testing if it works.
If it does, then we can make a nightly out of it, and encourage a few servers to start using it (I'm pretty sure there'd be people willing to help, there always are, shoutout to them!).

#### On another note...
Before we do all of that, we need some netcode stuff too for checking if luau scripts are signed, until then compiled scripts are a no-go, executing untrusted bytecode is dangerous (Same reason why currently Lua scripts go thru MTA's site, they get signed too)

dmi7ry:
At the moment there is one inconvenience. The analyzer knows nothing about MTA meta files, so it throws errors on functions or variables declared in other files of the same resource. We plan to add such support a bit later. Most likely in March or April.

dmi7ry:
BTW, about testing. We've been using it for five months now. And during this time there was not a single error or problem with it.

Pirulax:
I'd be more than happy to merge this once Luau is made into a submodule [or something we agree on, because I see we still don't use submodules....], because making changes to something that is regularly updated is just a ***bad*** idea [Why?.... CEGUI].
@botder [or not sure who to tag] any ideas on what should be done here? I guess we could go with the approach of creating a fork of Luau in the mta org, and downloading it directly [as we do with the fonts for example]

tederis:
Submodules can be used, I agree. But a pure Luau is incompatible with this PR. Thus a submodule must be based on a fork that should be regularly updated as well. But this goes beyond a normal PR because requires a close collaboration with the project maintainers that should create a fork of Luau first of all. But considering a small interest to this PR(and many others), it is unlikely that it will be done.

ERAGON007:
A repo should first be created at `multitheftauto/luau`, we need the patched version of Luau there, and then this PR could be merged?

Pirulax:
Okay I can create the luau repo, no problem, but should we add a submodule? @botder

Allerek:
My suggestion - to make our future life easier.

- <script src="x.lua" type="x" lang="lua-5.1" />

- <script src="x.lua" type="x" lang="lua-5.x" />

- <script src="x.lua" type="x" lang="luau-X.X" />

As there are attempts already to update current lua and there were discussions on how to define Lua version.

Lpsd:
Again, Lua version shouldn't be tied to individual scripts but to the resource itself. All scripts within a resource must use the Lua version defined by that resource.

This is how the implementation in MTA currently works (Lua VM per resource) and this PR doesn't change anything regarding that.

MegadreamsBE:
> My suggestion - to make our future life easier.
> 
>     * <script src="x.lua" type="x" lang="lua-5.1" />
> 
>     * <script src="x.lua" type="x" lang="lua-5.x" />
> 
>     * <script src="x.lua" type="x" lang="luau-X.X" />
> 
> 
> As there are attempts already to update current lua and there were discussions on how to define Lua version.

Why would you want to use different versions among different scripts rather than on a resource basis? This complicates the PR, makes your code less readable (as now you have to check the version each script is using) and just screams bad coding practices & design...

ERAGON007:
> My suggestion - to make our future life easier.
> 
> * <script src="x.lua" type="x" lang="lua-5.1" />
> * <script src="x.lua" type="x" lang="lua-5.x" />
> * <script src="x.lua" type="x" lang="luau-X.X" />
> 
> As there are attempts already to update current lua and there were discussions on how to define Lua version.

Bad idea maybe? language version should be set per resource, not per script. And also i assume that global data sharing between scripts in the same resource would troublesome.

ERAGON007:
> Okay I can create the luau repo, no problem, but should we add a submodule? @botder

Maybe you can use a submodule for this PR and any PR that adds some vendor library to mta? this way we can gradually start adopting submodules one at a time

dmi7ry:
> Bad idea maybe? language version should be set per resource, not per script. And also i assume that global data sharing between scripts in the same resource would troublesome.

Of course, it can be only per resource.
And this PR already has proper solution. It would be extended to another lua versions/languages, like:
```
<meta>	
    <lua both="luau"/>	
</meta>
```
```
<meta>	
    <lua client="lua54"/>	
</meta>
```
```
<meta>	
    <lua client="luau" server="lua52"/>	
</meta>
```

Pirulax:
I already did review it, and my stance is still the same.
(And I'm still waiting for @botder's answer)

I think for the sake of making the review easier, I think we should use a submodule for Luau.
If possible, we should use the official git repo directly, but I'm quite certain we'll need some custom changes (I'm planning on reworking the mtasaowner thing I added a while back to be more portable, instead of modifying Lua directly).
Other than that, I'm not sure what changes we've made to Lua.
I'd like to avoid modifying Luau, since it's frequently updated (We don't need another CEGUI)

Also, there's one thing that might not work well with multiple runtimes across different resources, and it's function refs, will have to test that.

Other than that, I think (for the script version) something like `<script_language type="luau">` could work.
Using `<lua>` would limit us to Lua only.
Specifying per-script makes no sense as stated above.
We could even add this tag directly into the meta as an attribute: `<meta script_language="luau">` or something.

Luau would be a great (and easy) addition (eventually we could also add LuaJIT). Luau is especially interesting, due to it's support for static typing (thank god), and a lot of useful features (that regular Lua just lacks).
Adding this def. is a priority for me.

reflectrpteam:
Currently there is only one change in **Luau**: 
`LUA_MINSTACK` was changed from `20` to `50`.

On the other hand **Lua 5.1** has the following changes:
`LUA_INTEGER` ptrdiff_t was changed from ptrdiff_t to int(for the consistency with Luau)
`LUAI_EXTRASPACE` was changed from `0` to `sizeof(size_t)` to keep additional data
`luai_userstateopen(L)` was changed from `((void)L)` to `*((size_t*)L - 0x1) = 0x5989C266`
`luai_userstatethread(L,L1)` was changed from `((void)L)` to `*((size_t*)L1 - 0x1) = *((size_t*)L - 0x1)`

Thus the only problem we have faced is how to patch **Luau** submodule with just one line of the code. 

In general, Luau requires more adjustment because currently it uses more memory than it could. But this can be solved by creating a PR in Luau repository.

tederis:
> Currently there is only one change in **Luau**: `LUA_MINSTACK` was changed from `20` to `50`.
> 
> On the other hand **Lua 5.1** has the following changes: `LUA_INTEGER` ptrdiff_t was changed from ptrdiff_t to int(for the consistency with Luau) `LUAI_EXTRASPACE` was changed from `0` to `sizeof(size_t)` to keep additional data `luai_userstateopen(L)` was changed from `((void)L)` to `*((size_t*)L - 0x1) = 0x5989C266` `luai_userstatethread(L,L1)` was changed from `((void)L)` to `*((size_t*)L1 - 0x1) = *((size_t*)L - 0x1)`
> 
> Thus the only problem we have faced is how to patch **Luau** submodule with just one line of the code.
> 
> In general, Luau requires more adjustment because currently it uses more memory than it could. But this can be solved by creating a PR in Luau repository.

I see that LUA_MINSTACK has the following definition: `#ifndef LUA_MINSTACK #define LUA_MINSTACK 20 #endif` in the code. It means that we can add LUA_MINSTACK as project-wide macro(define it within `premake5.lua`). So, Luau as a submodule is more than viable.

Pirulax:
Ah, okay so I didn't know about `LUAI_EXTRASPACE`.
I can use this to store the `CResource*` of the state, instead of editing the Lua source code.
Thanks.

> In general, Luau requires more adjustment because currently it uses more memory than it could
You mean it uses more than it *should*?

Could you explain the last 2 changes? (`luai_userstateopen` `luai_userstatethread`)
It doesn't make a whole lot of sense to me.

tederis:
> Ah, okay so I didn't know about `LUAI_EXTRASPACE`. I can use this to store the `CResource*` of the state, instead of editing the Lua source code. Thanks.
> 
> > In general, Luau requires more adjustment because currently it uses more memory than it could
> > You mean it uses more than it _should_?
> 
> Could you explain the last 2 changes? (`luai_userstateopen` `luai_userstatethread`) It doesn't make a whole lot of sense to me.

Yes, it uses more than it **should**, right.

**luai_userstateopen** is called at the end of **lua_newstate** so you can write a custom data just after the lua_State creation. In the context of this PR this macros adds a "magic" number _0x5989C266_ before a lua_State object(that's why LUAI_EXTRASPACE was used). _0x5989C266_ is just a number for the unambiguous lua 5.1 recognition. (Luau at its own side is also pushes some "magic" data before a Lua state.).

**luai_userstatethread** is called when a new coroutine is created. In the context of this PR this macros copies a "magic" number to a new coroutine Lua state.

Pirulax:
Since I got no response from Botder, I went ahead and made the [fork](https://github.com/multitheftauto/luau).
When I have time (or you), the premake5.lua, and the changes have to be implemented in that repo, and a submodule should be used here.
For reference, I tried[ moving Lua to be a submodule](https://github.com/multitheftauto/mtasa-blue/pull/2112)  once.
It's not too bad.
I'll shortly make a PR to remove `mtasaowner` from the Lua source code, and use the `LUAI_EXTRASPACE` instead (To make it portable). But I don't seen anything like it in Luau's code tho.

tederis:
> I'll shortly make a PR to remove `mtasaowner` from the Lua source code, and use the `LUAI_EXTRASPACE` instead (To make it portable).

Please, reserve an extra 4 bytes for the Lua selection system(which is added in this PR). You can create a structure:
```
struct
{
    CResource* resource;
    size_t userData;
}
```
(or something like that)
So I will be able to make this PR compatible with the new LUAI_EXTRASPACE space.

tederis:
>  But I don't seen anything like it in Luau's code tho.

This is the reason for the excessive memory usage of Luau. Luau doesn't has LUAI_EXTRASPACE feature so I was forced to implement it on my own by allocating an extra 4 bytes for every memory page. And this must be removed in future by commiting to the Luau project directly.

Pirulax:
> > Ah, okay so I didn't know about `LUAI_EXTRASPACE`. I can use this to store the `CResource*` of the state, instead of editing the Lua source code. Thanks.
> > > In general, Luau requires more adjustment because currently it uses more memory than it could
> > > You mean it uses more than it _should_?
> > 
> > 
> > Could you explain the last 2 changes? (`luai_userstateopen` `luai_userstatethread`) It doesn't make a whole lot of sense to me.
> 
> Yes, it uses more than it **should**, right.
> 
> **luai_userstateopen** is called at the end of **lua_newstate** so you can write a custom data just after the lua_State creation. In the context of this PR this macros adds a "magic" number _0x5989C266_ before a lua_State object(that's why LUAI_EXTRASPACE was used). _0x5989C266_ is just a number for the unambiguous lua 5.1 recognition. (Luau at its own side is also pushes some "magic" data before a Lua state.).
> 
> **luai_userstatethread** is called when a new coroutine is created. In the context of this PR this macros copies a "magic" number to a new coroutine Lua state.

That is the answer to my problem, thanks.
Luau doesn't have `LUAI_EXTRASPACE`, but it uses the `userdata` that can be given to every state in `lua_open`.
So that's what I'm going to do for both Lua 5.1 and Luau.

Also, is this `0x5989C266` value official, or we came up with it?
Would be nice to have it in a macro, like `LUA_51_MAGIC`.
I assume Luau has a macro for it's magic number.

tederis:
> > > Ah, okay so I didn't know about `LUAI_EXTRASPACE`. I can use this to store the `CResource*` of the state, instead of editing the Lua source code. Thanks.
> > > > In general, Luau requires more adjustment because currently it uses more memory than it could
> > > > You mean it uses more than it _should_?
> > > 
> > > 
> > > Could you explain the last 2 changes? (`luai_userstateopen` `luai_userstatethread`) It doesn't make a whole lot of sense to me.
> > 
> > 
> > Yes, it uses more than it **should**, right.
> > **luai_userstateopen** is called at the end of **lua_newstate** so you can write a custom data just after the lua_State creation. In the context of this PR this macros adds a "magic" number _0x5989C266_ before a lua_State object(that's why LUAI_EXTRASPACE was used). _0x5989C266_ is just a number for the unambiguous lua 5.1 recognition. (Luau at its own side is also pushes some "magic" data before a Lua state.).
> > **luai_userstatethread** is called when a new coroutine is created. In the context of this PR this macros copies a "magic" number to a new coroutine Lua state.
> 
> That is the answer to my problem, thanks. Luau doesn't have `LUAI_EXTRASPACE`, but it uses the `userdata` that can be given to every state in `lua_open`. So that's what I'm going to do for both Lua 5.1 and Luau.
> 
> Also, is this `0x5989C266` value official, or we came up with it? Would be nice to have it in a macro, like `LUA_51_MAGIC`. I assume Luau has a macro for it's magic number.

No, 0x5989C266 is the completely custom invention. It just a way to differentiate Luau and Lua 5.1 by having just a pointer to a Lua state.

Pirulax:
Okay, so we ended up deciding that adding an extra `size_t` to the beginning of `lua_State` would be a good solution.
The same will be done for Luau too (since we have a submodule for it), it's a really small change (3 lines in total).

Pirulax:
Update: Adding anything before the header probably won't work. I tried adding LUA_EXTRASPACE to `Luau`, but that isn't possible (easily) right now either (As, unlike in `Lua`, in `Luau` the `lua_State` objects are gc objects, not raw mem allocs, and `luaM_freegco` expects a pointer directly to a gco, not an altered raw pointer.
~~I guess another option would be to add store this info in the beginning of `CommonHeader` somehow.~~ (Nvm, waste of memory)

Pirulax:
Okay, I give up.
I think we have to make a compromise here.
1. We keep Lua 5.1 at a cost of a higher memory (extra 4-8 bytes for every Lua allocation) and CPU (due to the translation layer - vlua) usage.
2. Drop Lua 5.1 support, have only Luau - Luau supports Lua 5.1 scripts, **BUT** not bytecode - Ergo, all compiled scripts would cease to work. This was brought up many times by @sbx320 and me (But for LuaJIT), and I always vouched for people simply re-compiling their scripts - As anyone who obtained those scripts in a "legal" manner, should be able to recompile them without any issue. (Expect that there are a few community scripts that are distro'd like that).
3. Similar to 2. but build a transpiler from Lua 5.1 VM bytecode Luau - Let's be realistic this is way too much effort, nobody's going to do it
4. No Luau - Shit option, I advise against it - Luau is the future, it's just an upgrade in both performance and productiveness.

I'd honestly go with 2. Main reason is performance and a memory usage - Merging something that could drastically increase memory usage could harm servers that are already on the edge of that 4GB limit.
I haven't tested, but with this change I'd anticipate at least a 20-30% increase in Lua VM memory usage. This would also affect cache lines heavily (I really hope TValue's aren't padded with this, if they're 2. isn't even an option imo, as the performance impact is too heavy).

I think higher performance+less memory usage would be totally acceptable for the cost of having to recompile your scripts.
As for servers that don't have the original sources, they can always stick to an older version that has Lua 5.1, not Luau.

MegadreamsBE:
From my standpoint option 2 is the best. As you said, if the scripts were obtained legally it's easy enough to re-compile them, not a big deal. The pros outweigh the cons in this scenario.

hasitotabla:
Same, 2. option is the best. Just the new features (continue statement, compound assignments, etc..) alone could help a lot writing better codes.

This is a no brainer in my opinion, if it’s possible.

sbx320:
> We keep Lua 5.1 at a cost of a higher memory (extra 4-8 bytes for every Lua allocation) and CPU (due to the translation layer - vlua) usage.

Why do we need a translation layer if we ship both Luau and Lua 5.1 and use the respective VM depending on the resource setting? In both cases only one VM would be running and Lua(u)-objects would never hit another VM. 

2 is not a realistic option. There are too many compiled scripts out there that'd just break, where source isn't available because no one ever cared about it. At that point we'd just be creating another MTA fork.

> Merging something that could drastically increase memory usage could harm servers that are already on the edge of that 4GB limit.

Those should move to x64 anyway.

ERAGON007:
> 2 is not a realistic option. There are too many compiled scripts out there that'd just break, where source isn't available because no one ever cared about it. At that point we'd just be creating another MTA fork.

Isn't it fair to say that if someone didn't care about source code of the resource that they use, they wouldn't even probably *try* to update their MTA server binary?

MegadreamsBE:
> 2 is not a realistic option. There are too many compiled scripts out there that'd just break, where source isn't available because no one ever cared about it. At that point we'd just be creating another MTA fork.

A line has to be drawn somewhere for the sake of progression. We just can't keep hindering ourselves to maintain backwards compatibility for outdated scripts that are more likely to be insecure (and with the source lost you can't even verify it) anyways.

We need to force the community to improve and ensure MTA does so right alongside them.

sbx320:
> Isn't it fair to say that if someone didn't care about source code of the resource that they use, they wouldn't even probably try to update their MTA server binary?

Not really. Many people update their servers that have no idea how to do anything in Lua. Also this would break client scripts as well, for example in various race maps. 

> A line has to be drawn somewhere for the sake of progression.

Yes. However the benefit of breaking compatibility also needs to be weighted against the downsides of it. I don't consider Luau to be that significant of an upgrade over standard Lua for it to be worht breaking all existing compiled scripts. 

- Option 1 seems managable in terms of development effort and future maintenance, although I'm still not sure why there would be any relevant runtime overhead if we select the VM on resource load.
- Option 2 isn't happening. 
- Option 3 is unlikely to work (even if we find someone to build a translator) given that Luau doesn't have a compiled format (and going 5.1 bytecode -> Luau source is likely to be very slow). 
- Option 4 seems most likely.

ffsPLASMA:
Regarding compiled scripts (from race maps or whatever), if mta team would provide a fast batch decompiler of luac to get the readable pseudocode back and potentially recompile it for luau, then this could work.

Pirulax:
> Those should move to x64 anyway.

GTA is still 32bit, and will be until we finish reversing it. So x64 won't help.

> Why do we need a translation layer if we ship both Luau and Lua 5.1 and use the respective VM depending on the resource setting? In both cases only one VM would be running and Lua(u)-objects would never hit another VM.

Because we need to know which `lua_State` is which Lua version.
Obviously the 2 `lua_State's` aren't compatible, thus we have to virtualize the calls (and that's what this PR is doing), but that inherently comes with a performance loss.

# New: Option 5
We would increment the version to 1.7, this way everybody gets to keep MTA with Lua 5.1. Server owners with compiled scripts without sources would be stuck on the last 1.6 version, and everybody else would be on 1.7 enjoying Luau (And let's be real, majority of players play on servers that will be able to move to 1.7).  Also, server owners not having sources and not know much about Lua wouldn't have much use anyways of newer versions - They can't take advantage of new features, if they don't know how to script in Lua, no?

tederis:
> Why do we need a translation layer if we ship both Luau and Lua 5.1 and use the respective VM depending on the resource setting? In both cases only one VM would be running and Lua(u)-objects would never hit another VM.

VM selection on resource start is the best(from the perfromance perspective) solution. But how does this can be implemented? if-else statements for every lua function call site? The strategy pattern? It will require a lot of refactoring. And that is the reason why the translation layer system was chosen. It requires almost no modification of the existing MTA codebase, works as fast as possible and can be easily extended.

AlexTMjugador:
> But how does this can be implemented? if-else statements for every lua function call site? The strategy pattern? It will require a lot of refactoring. And that is the reason why the translation layer system was chosen. It requires almost no modification of the existing MTA codebase, works as fast as possible and can be easily extended.

I didn't review this PR deeply enough to know if I'm about to say something impractical, but why can't we just default to Lua 5.1 for resources that don't explicitly opt-in to Luau in the `meta.xml` file?

If we ship both Lua VMs, and only create the appropriate one after parsing the `meta.xml` file, that sounds reasonable enough. If communication between VMs of different versions is necessary, we could look into using userdata instead, perhaps?

tederis:
> > But how does this can be implemented? if-else statements for every lua function call site? The strategy pattern? It will require a lot of refactoring. And that is the reason why the translation layer system was chosen. It requires almost no modification of the existing MTA codebase, works as fast as possible and can be easily extended.
> 
> I didn't review this PR deeply enough to know if I'm about to say something impractical, but why can't we just default to Lua 5.1 for resources that don't explicitly opt-in to Luau in the `meta.xml` file?
> 
> If we ship both Lua VMs, and only create the appropriate one after parsing the `meta.xml` file, that sounds reasonable enough. If communication between VMs of different versions is necessary, we could look into using userdata instead, perhaps?

In this PR the only one VM exists at the same time for each resource. And Lua 5.1 is already default. Without the virtualization we simply won't be able so dispatch the Lua function calls.

AlexTMjugador:
> In this PR the only one VM exists at the same time for each resource. And Lua 5.1 is already default. Without the virtualization we simply won't be able so dispatch the Lua function calls.

Right, I understand this overhead is needed because we want to instance the same, multi-version VM across resources. But what is preventing us from using different VMs for different Lua versions for different resources?

tederis:
It seems that this PR and its implementation details are unclear for most of the people. It adds a custom translation layer (called vlua) that replaces Lua API functions by its own "proxy" versions. So, lets say MTA is calling lua_pushstring, which is actually a vlua function that internally deduces the version of Lua from a lua_State pointer. And then using that information it invokes the corresponding Lua function which can be Lua 5.1 or Luau. Currently the Lua version is deduced by a special "magic" field that's placed before the lua_State object. Simple, fast, scalable.

AlexTMjugador:
Thanks @tederis, that explanation made it much clearer to me! I think it's a fundamentally smart approach.

sbx320:
> But how does this can be implemented? if-else statements for every lua function call site?

Some if-else and a bunch of templates. It's a bit more work, but it doesn't sound too bad.

I would've reimplemented the Lua API as `template<LuaVariant>`. So that `lua_tonumber<Lua51>` calls `lua_tonumber` from 5.1. Most API usages could then also be templated (e.g. CLuaArguments) to catch existing usages. Legacy Script functions would also become templates (new functions via argparser could support this directly), instantiated once for each VM type.

That way we can avoid any runtime CPU overhead and only gain some code size.

TheNormalnij:
> We would increment the version to 1.7, this way everybody gets to keep MTA with Lua 5.1. Server owners with compiled scripts without sources would be stuck on the last 1.6 version, and everybody else would be on 1.7 enjoying Luau (And let's be real, majority of players play on servers that will be able to move to 1.7). Also, server owners not having sources and not know much about Lua wouldn't have much use anyways of newer versions - They can't take advantage of new features, if they don't know how to script in Lua, no?

They can install resources with new features without Lua knowledge.

It's allowed to use EmmyLua with MTA to have all benefits of typed language. Or you can use [TypeScript -> Lua](https://github.com/mtasa-typescript) transaction layer and have significantly better language that Luau. This options request no changes in MTA codebase.

Lua community provides a lot of native modules for Lua. [LuaRock](https://luarocks.org/) is [partially integrated](https://github.com/TheNormalnij/MTA-package) in MTA and provide some multi-threading support for windows.
Also [debugging tools](https://github.com/satoren/LRDB) can be used with vanilla Lua.

There is also a lot of ways to use Lua knowledge. [Defold](https://defold.com/) uses Lua as main language. Project Zomboid, Garrys Mod, [BG3 SE](https://github.com/Norbyte/bg3s) use Lua for mods 

I am not a fan of this changes. EmmyLua + [debugger](https://github.com/TheNormalnij/MTATD) provide the best Lua developing experience for me. Maybe somebody is going to add Luau in MTA. Just please, don't break old resources. Luau has not enough benefits for it.

ERAGON007:
>I am not a fan of this changes. EmmyLua + [debugger](https://github.com/TheNormalnij/MTATD) provide the best Lua developing experience for me. Maybe somebody is going to add Luau in MTA. Just please, don't break old resources. Luau has not enough benefits for it.

With these kind of mindsets, why don't we switch to VLang? it has all the benefits of c :), or even better option, Python!
If no one is willing to adopt new things in tech, the project will inevitably lose it's popularity because no new devs will use it.

Pirulax:
It's just the way MTA works.
We deprecate functions, then keep them for 10 years at which point it's impossible to remove them due to "backwards compatibility".
When there's a new feature nobody wants to merge it, because "it might break old scripts" (Bonus is that those scripts are compiled so they can't even be fixed).
This endless backwards compatibility drags the whole project back. We aren't Microsoft.
And sadly it also fragmentates the community, because so many changes just aren't ever added to this repo, that people just end up making forks where they can freely merge the PRs from this repo (Bullet Physics for example).

> It's allowed to use EmmyLua with MTA to have all benefits of typed language. Or you can use [TypeScript -> Lua](https://github.com/mtasa-typescript) transaction layer and have significantly better language that Luau. This options request no changes in MTA codebase.

Lua is slow. Luau is faster. LuaJIT is the fastest, but security is questionable (The main reason why Luau isn't JIT).
Luau is a drop-in replacement with TYPING, and lots of other features that improve dev. experience out-of-the-box, not to mention that it's faster than Lua in a lot of scenarios, and doesn't *force* you to write unreadable code just to be faster (numeric for loop vs ipairs, etc).

AlexTMjugador:
For what it's worth, I think the approach outlined by @sbx320 seems worth exploring to move this PR closer to being merged, in addition to moving Luau to a submodule whose changes can be easily tracked under version control.

I find that @TheNormalnij has some valid points about Luau's addition not being strictly necessary to MTA, but given that:

- Luau is well-maintained and battle-tested by Roblox, one of the most popular MMOs in the world, for a very similar use case,
- We've discussed workable ideas to address backwards compatibility and runtime overhead concerns,
- It's a lightweight library (a few MiB big judging from the size of the official `luau-analyze.exe` build, which is comparable to the average website size),
- Its addition is mostly a one-time effort (and it could pave the way to implementing ways of reducing the required manpower to maintain vendor components; see below),
- It would serve as an enabler for things everyone wants: more complex client-side scripts (c.f. @MegadreamsBE's Sphene and console emulators, which are unlikely to have performance good enough to be used in production without the aid of a fundamentally faster runtime), and proper vendor dependency management in MTA (Luau can be used as an example of how to get things done when it comes to adding vendor code),

I think realistically there is no downside to this PR, so I would recommend against letting this good effort go to waste. I would indeed advocate for caution in adding big features like Bulletphysics, because those are either complex in themselves or open cans of worms in terms of future complexity, practical usability in their MVP form, expansion, and maintainability, but this kind of "infrastructure" improvements pack, in my opinion, a lot of bang for the buck that's worth the relatively little effort.

Pirulax:
Also, something LuaJIT, and Lua 5.1 doesn't have is native vectors, that cause no GC pressure.
Vectors are something you really don't want to use currently, because of how much they tank the performance (because of the gc pressure).
Luau was literally designed for MTA :D
If it's good enough for Roblox, it's good enough for us.

TheNormalnij:
> We deprecate functions, then keep them for 10 years at which point it's impossible to remove them due to "backwards compatibility".

Sometimes we do very small changes in API. This changes doesn't breaks much and old resources can be adopted. This is ok for me. Lua is a very sensitive part of MTA. And changes in bytecode break a lot of old resources.  There is a lot of script markets, where server owners buy compiled scripts. Some freelancers sell compiled resources only. I think we are going to break too many things.
This is strong enough reason to keep  Lua 5.1 for me.

> Lua is slow. Luau is faster. LuaJIT is the fastest, but security is questionable (The main reason why Luau isn't JIT).

In my experience Lua is fast enough. MTA performance is suffering from bad decisions in API + `LuaVM <-> engine`  serialization.
It's nice to see real benchmarks. Significant performance impact is more like reason to add Luau as second LuaVM. 

<details>
  <summary>Bad API</summary>

   * People uses raw and slow dx functions instead of any library.
   * .. or ever CEF for GUI
   * Objects in MTA are slow. They use dynamic objects pool and additional MTA streamer over it,
   *  All DB functions sucks. dbQuery provide some sort of queue for queries. This decreases maximum performance. 
   * EngineReplaceModel doesn't work well.
   * fileRead is blocking operation. It copies file content between MTA engine and LuaVM. With encryption function and  engineReplace*** function this is huge loss.
   * toJSON is joke
   * Vectors are implemented in wrong way
  
</details>

> Also, something LuaJIT, and Lua 5.1 doesn't have is native vectors, that cause no GC pressure.

#1728 we can change our Lua implementation and do event better for MTA.

Fernando-A-Rocha:
Keep Lua as it is, add Luau (optional) that can be enabled in meta.xml
Everybody is happy, end of story ❗

Rilot06:
> Keep Lua as it is, add Luau (optional) that can be enabled in meta.xml Everybody is happy, end of story ❗

Others have said before that if we keep 5.1, it would need a translation layer, causing more cpu usage and it would use more memory, I don't think it's an option, performance isn't that good currently either, and it would be even worse if we keep 5.1. If we keep worrying about backwards compatibility, the game will never get a decent upgrade, I don't see why the game has to support 15 years old compiled (mostly stolen) scripts. 99.9% of the time if you are a server owner/developer, you wouldn't backup just the compiled scripts, you would backup the sources too, only if you didn't steal it from other servers of course..

TheNormalnij:
I think we are talking about two parallel version without big performance impact. Transaction layer is too hard

Rilot06:
> I think we are talking about two parallel version without big performance impact. Transaction layer is too hard

I don't understand what you are saying, there would be a pretty big performance hit if we keep both languages, Pirulax said it already.

> 1. We keep Lua 5.1 at a cost of a higher memory (extra 4-8 bytes for every Lua allocation) and CPU (due to the translation layer - vlua) usage.

Also the translation layer is literally the chosen method in this pr, it's already working, so I don't get what you are trying to say.

TheNormalnij:
Ah, i forgot this information, sorry

CrosRoad95:
Any update? Eta when it will be merged?

tederis:
As it was previously discussed this PR is ready to use. It's stable and backwards compatible. The version of Luau will be updated soon.

CrosRoad95:
Therefore i propose  to update luau and merge it. it will be gamechanger for creators

ghost:
In fact, the creators waiting for this to merge even I'm waiting

sbx320:
As outlined in previous comments, I don't see this getting merged at all in its current form.

CrosRoad95:
> As outlined in previous comments, I don't see this getting merged at all in its current form.

I have been creating servers in lua for years, i know by heart how hard it can be to maintain hundred of resources, thousend upon thousend of lines code. I also know javascript and typescript where luau give you at least basics of typescript.

Not merging this just proves that mta team take care of low effort thousend of empty servers which won't benefit community at all. While on just top 10 servers by players count play 10+% of all players.

Rilot06:
> > As outlined in previous comments, I don't see this getting merged at all in its current form.
> 
> I have been creating servers in lua for years, i know by heart how hard it can be to maintain hundred of resources, thousend upon thousend of lines code. I also know javascript and typescript where luau give you at least basics of typescript.
> 
> Not merging this just proves that mta team take care of low effort thousend of empty servers which won't benefit community at all. While on just top 10 servers by players count play 10+% of all players.

That's how it has been for years. Get used to it. Barely any good big PRs get merged because of bAcKwArDs CoMpAtIbIlItY, even if it's backwards compatible completely.

CrosRoad95:
> bAcKwArDs CoMpAtIbIlItY

this pr has no bc issues, just managment issues. They should create server by themself to know what i'm talking about

> And sadly it also fragmentates the community,

oh no 😭, it fragments the community, what can you do if people are not dependent on you? how can you enfoce them to do what you want if you have no power to do it. I have to probably continue work on my net reimplementation

Mta can't be like miencraft where you have 999 launchers, 9999 forks, 99999 developers,

Rilot06:
> > bAcKwArDs CoMpAtIbIlItY
> 
> this pr has no bc issues, just managment issues. They should create server by themself to know what i'm talking about

There were some bc concerns if you read back old comments, that's what they said why they won't merge it as it is iirc, but doesn't matter, we'll probably never get this merged

AlexTMjugador:
> Not merging this just proves that mta team take care of low effort thousend of empty servers which won't benefit community at all.

At the risk of uselessly stating the obvious here, nobody said this won't ever be merged. It was just said that this is unlikely to be merged in its current form, which makes sense given that e.g. @sbx320 proposed using templates to reduce CPU overhead and the PR author did not address that feedback, or for that matter any other feedback that suggested any improvement.

I very much want to see Luau in MTA, but submitting a PR, gathering lots of feedback about it, and then letting that feedback rot or be ignored, but nevertheless ask for an incomplete work to be merged anyway because it's useful, is unnecessarily unpolite to reviewers and demoralizing for a volunteer-run project.

Please keep the discussion civil and focused on bringing the PR to a state closer to consensus for merging. Any other discussion simply does not belong to a GitHub repository. Thanks.

tederis:
This PR is the offering of one of MTA-based projects and it's successfully used there in the current state. This implementation is most successful and flexible and there were no plans to change it. So, if there are no people willing to rewrite it completely(as some of the suggestions implies) then this PR can be closed. At least this PR can be used by other projects.

Fernando-A-Rocha:
tederis said this works fine in another big MTA project. We trust tederis, so let's put this to test in nightly branch please. Nothing better than having servers and users put it to good use.

dmi7ry:
> and it's successfully used there in the current state

Almost 2 years of use, actually.

ghost:
image
This version has **luau**?

TracerDS:
> image This version has **luau**?

No

TheNormalnij:
There are two problems now:
* compiled scripts
* backend changes

I can check if i can write a service to migrate old lua bytecode to luau bytecode.
This service can be integrated with MTA infrastructure and deal with encrypted scripts too.

At least lua decompilers exists. This give me a hope

CrosRoad95:
> I can check if i can write a service to migrate old lua bytecode to luau bytecode.

Prune to make mistake, people should recompile scripts again fo luau or leave them on lua.

Fernando-A-Rocha:
Just merge and don't support compiled luau scripts yet.
This can be supported later, it's stalling the project.

TheNormalnij:
> Just merge and don't support compiled luau scripts yet.

Im talking about all lua legacy. Without legacy support we can close a lot of projects.

hasitotabla:
why do yall care so much about stolen compiled scripts? lol

ghost:
> why do yall care so much about stolen compiled scripts? lol

Huge servers that use compiled .lua, hence the concern, which may generate problems for all MTA servers

hasitotabla:
> Huge servers that use compiled .lua, hence the concern, which may generate problems for all MTA servers

Large servers (hopefully) have their own CI/CD, they'd need to force build master and that's it. I don't get it why would this be a problem for anyone, if they have the actual source code...

botder:
> > Huge servers that use compiled .lua, hence the concern, which may generate problems for all MTA servers
> 
> Large servers (hopefully) have their own CI/CD, they'd need to force build master and that's it. I don't get it why would this be a problem for anyone, if they have the actual source code...

Maybe re-read their post before writing? This was never an issue for servers with source-access.

Rilot06:
> > > Huge servers that use compiled .lua, hence the concern, which may generate problems for all MTA servers
> > 
> > 
> > Large servers (hopefully) have their own CI/CD, they'd need to force build master and that's it. I don't get it why would this be a problem for anyone, if they have the actual source code...
> 
> Maybe re-read their post before writing? This was never an issue for servers with source-access.

If you don't have the source code for your own server, you've stolen the scripts for it. There is literally no other explanation why you would only backup the compiled scripts, not the sources.

botder:
> If you don't have the source code for your own server, you've stolen the scripts for it. There is literally no other explanation why you would only backup the compiled scripts, not the sources.

So you're telling me that racing maps with compiled scripts are all stolen, or because mappers protect their scripts with encryption?

ghost:
> > If you don't have the source code for your own server, you've stolen the scripts for it. There is literally no other explanation why you would only backup the compiled scripts, not the sources.
> 
> So you're telling me that racing maps with compiled scripts are all stolen, or because mappers protect their scripts with encryption?

Which is not even the biggest problem, because there are a lot of scripters selling compiled scripts, the owners of the servers paid for them, and they would lose them if backward compatibility cannot be solved

hasitotabla:
> Which is not even the biggest problem, because there are a lot of scripters selling compiled scripts, the owners of the servers paid for them, and they would lose them if backward compatibility cannot be solved

Hmm maybe they should ask the developer to compile the sauce again for them?

Rilot06:
> > If you don't have the source code for your own server, you've stolen the scripts for it. There is literally no other explanation why you would only backup the compiled scripts, not the sources.
> 
> So you're telling me that racing maps with compiled scripts are all stolen, or because mappers protect their scripts with encryption?

If I understand correctly, this PR allows each resource to be used with Lua OR Luau. If "your" script is compiled, you could just use it with the old Lua version, why would you want to recompile it then?

CrosRoad95:
Compiling should also obfucate name and generate source map, so your script failed at position "2354" in function "hR" and you lookup in `myscript.lua.map` what "hR" is, and how position translates to original line and index because i have seen people decompiling lua and they were getting almost 1:1 code with function names - ofc in this case you should not relay on symbol names otherwise your resource can break.

hasitotabla:
Oh wait, if it's possible to select runtime then where's the problem? If the server detects the script is compiled, then convert the meta.xml line to explicitly using lua5.1. Otherwise just load it as luau code...

Rilot06:
Actually not even just on a per-resource basis, you can even use different Lua version for server and client side in the same resource, so yeah, I really don't see where the problem is. Yes, there is some CPU and memory overhead because of the translation layer, but they have been using it for more than a year or so. A nightly test for high player count tests would be really great.

ghost:
> > > If you don't have the source code for your own server, you've stolen the scripts for it. There is literally no other explanation why you would only backup the compiled scripts, not the sources.
> > 
> > 
> > So you're telling me that racing maps with compiled scripts are all stolen, or because mappers protect their scripts with encryption?
> 
> If I understand correctly, this PR allows each resource to be used with Lua OR Luau. If "your" script is compiled, you could just use it with the old Lua version, why would you want to recompile it then?

You had a point, if the script is already compiled you don't need to recompile, since you are using Lua version

kxndrick0:
I just had a quick question, is it possible to detect the language of each script and then run it in that language? If so there is no problem according to me.

dmi7ry:
> If the server detects the script is compiled, then convert the meta.xml line to explicitly using lua5.1. Otherwise just load it as luau code...

Luau has own compiler, of course.

dmi7ry:
> I just had a quick question, is it possible to detect the language of each script and then run it in that language? 

Theoretically it is possible, but it has many pitfalls, when the definition will not work correctly. In general, it is a very unreliable method. The current version is much simpler and more reliable.

> If so there is no problem according to me.

There are no problems here anyway

tederis:
> > Just merge and don't support compiled luau scripts yet.
> 
> Im talking about all lua legacy. Without legacy support we can close a lot of projects.

Like I said, this PR is 100% backward compatible. Luac works as before for Lua 5.1. But it no longer matters...

MegadreamsBE:
@sbx320 What was the reason for the closure? I understand that you probably have your reasons, but it'd be best to share them as there was a lot of support for this PR.

CrosRoad95:
If you @sbx320 don't want to help develop mta, at least don't disturb

I seen you MegadreamsBE gave and removed thumb down emoji ;)

Xenius97:
RIP
MTA will never have alternative programming languages. Someone worked on this 2+ years and closed within a second.

shadylua:
For about 2 years many developers for Luau have collaborated and shared their knowledge with each other, but what I don't understand is why this topic is closed.

Fernando-A-Rocha:
> RIP MTA will never have alternative programming languages. Someone worked on this 2+ years and closed within a second.

Two things I don't understand:

1. Why this PR was closed after 2 years of existence by @sbx320 who did not comment anything
2. Why you guys say that this will never be merged, it's literally ready to be tested in nightly

Rilot06:
> 2. Why you guys say that this will never be merged, it's literally ready to be tested in nightly

Because it was open for 2 years, and all the "MTA team" did was cry about backwards compatibility, even when it wasn't an issue. Big new features almost never get merged into MTA.

Xenius97:
About backward compatibility, I read somewhere, maybe in the comments here, that the version number should be raised to MTA 1.7. This would give time for server developers to update scripts.

dmi7ry:
> About backward compatibility, I read somewhere, maybe in the comments here, that the version number should be raised to MTA 1.7. This would give time for server developers to update scripts.

There is no a reason for it. It's 100% compartible.
Nobody see difference until he decided to switch resource(s) to Luau (then it brings performance, including native compilation support, string interpolation, optional types support, etc).

sbx320:
> @sbx320 What was the reason for the closure? I understand that you probably have your reasons, but it'd be best to share them as there was a lot of support for this PR.

See https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-2395241570

As there are apparently no intentions to address the mentioned issues, this won't be merged.

ghost:
staff team is retarded, they dont want to merge anything that requires them to work more than 5 minutes (they would AT least have to create a secure bytecode loader for luau, just like for lua)

Fernando-A-Rocha:
> > @sbx320 What was the reason for the closure? I understand that you probably have your reasons, but it'd be best to share them as there was a lot of support for this PR.
> 
> See [#2815 (comment)](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-2395241570)
> 
> As there are apparently no intentions to address the mentioned issues, this won't be merged.

Could the mentioned issues (that need to be addressed) be recapitulated? I think this information is lost/unclear after the dozens of comments.

Rilot06:
> > @sbx320 What was the reason for the closure? I understand that you probably have your reasons, but it'd be best to share them as there was a lot of support for this PR.
> 
> See [#2815 (comment)](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-2395241570)
> 
> As there are apparently no intentions to address the mentioned issues, this won't be merged.

"Issues" xdd It's literally an optional language that you don't have to use, there are no issues if you use 5.1...

CArg22:
Checkouted this branch and ran my gamemode, couldn't notice any issues

tederis:
> Checkouted this branch and ran my gamemode, couldn't notice any issues

Although this PR will never be merged it's still viable and can be used in your own forks. Have a fun!

ghost:
I feel very dissatisfied with the lack of appreciation that the MTA team shows in relation to important PRs like this. It's frustrating to see people who don't dedicate themselves to improving the repository, who simply disappear and only reappear when there are great PRs, only to claim that it wouldn't be feasible and close the PR.

CrosRoad95:
I understand they don't have time, but they could at least once every year, every update introduce one bigger feature

ghost:
> I understand they don't have time, but they could at least once every year, every update introduce one bigger feature

True, in time that a "big" PR was not added, this is very frustrating for developers who want more functions or commitment for MTA

Dark-Dragon:
> Could the mentioned issues (that need to be addressed) be recapitulated?

From [these options](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-1872389067) 2 was not an option because luau isn't deemed to be enough of an upgrade to break obfuscated script file compatibility, 3 was not an option because it's likely not going to work even if someone were to put in ridiculous amounts of work to attempt it and 4 would just scrap the pull request. Later an option 5 was presented, which would keep lua 5.1 on mta 1.6 and below and luau from 1.7 onward, however it would divide the community just for the sake of luau.

This leaves us with option 1 which would keep lua 5.1 alongside luau, however the implementation in this pull request introduces a new translation layer which is explained [here](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-1872520657) and requires noticeable overhead in cpu and memory. While the original comment explaining why this was chosen seems to be gone, [quotes](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-1872516102) left enough context to explain that this was a clever trick that made it easier to implement. Other comments seem to suggest a quote on quote proper implementation would have required direct changes to luau directly through a fork or custom code (probably similarly to how lua is currently slightly modified in mta) which would have then required separate maintenance or collaboration from outside. With some of the context being lost I assume the last suggestion to get rid of the overhead was [this comment](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-1872533654) by sbx. There might have been more conversation internally as to why this either wasn't feasible, or not within the scope of the amount of work the PR author is willing to take on, but either way that's where this particular chain of thoughts seems to have ended.

That's it, at least from what I understand.

hasitotabla:
Guys revolutionary idea, support 5.1 for a year or so (so servers have time to do their thing), then make it deprecated and yeet the translation layer out the window. 😎👍

tederis:
> noticeable overhead in cpu and memory

This is not entirely true. The translation layer obviously implies some overheads of course. But they aren't noticeable. The layer is thin and can be considered(for the simplicity) as an additional dereferencing(or indirection). My results: just a few tens of nanoseconds.

Rilot06:
> Guys revolutionary idea, support 5.1 for a year or so (so servers have time to do their thing), then make it deprecated and yeet the translation layer out the window. 😎👍

You can support it for 10-20 years, it still won't be enough for the MTA team, they want ancient compiled scripts to run too forever..

Lpsd:
The summary above by @Dark-Dragon is accurate, for this PR to proceed the following items need to be addressed:

- Luau implementation to be adjusted to reduce/eliminate the unnecessary overheads ([context](https://github.com/multitheftauto/mtasa-blue/pull/2815#issuecomment-1872533654))

Option 1 from previous discussion should be used; meaning Lua 5.1 and Luau are shipped side-by-side with Lua 5.1 being the default (Luau configurable via meta.xml). 

Since Lua 5.1 is the default, existing compiled scripts are not affected. We can add support for Luau script compilation at a later date.

There may be other things to consider, but these are the main items to get this moving.

As the original author(s)/team have stated they don't have the time for these changes (and otherwise may simply not agree on the suggested approach), it would be up to someone else to adjust or re-implement Luau based on that approach.

I'll lock the PR now to avoid any more unnecessary convolution of this thread. These are the facts, use them as you wish.

PR #218, opened 2018-07-04
Add new on(Client)ElementCreated event

CrosRoad95:
Client side:
name: **onClientElementCreated**
1st argument: type: boolean; true if element was created by server, false if created by client
**source** is created element.

Server side:
name: **onElementCreated**
no arguments.
**source** is created element.

working with all types of elements from https://wiki.multitheftauto.com/wiki/Element . If i something missed, please mention.

General test of almost every element type.
[test1.zip](https://github.com/multitheftauto/mtasa-blue/files/2163605/test1.zip)

Some script that collect every created element by type and return them.
[test2.zip](https://github.com/multitheftauto/mtasa-blue/files/2163620/test2.zip)

This PR closes #389

PR #2826, opened 2022-12-05
Add new events onClientVehicleCreate and onVehicleCreate

syncc01:
These events are called when a vehicle is created.

I have seen that a group of people have requested this event in our wonderful MTA and I have decided to add them. 

This is my first pull request, good luck to me (:

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -2528,7 +2528,7 @@ void CClientGame::AddBuiltInEvents()
     m_Events.AddEvent("onClientPedStep", "foot", nullptr, false);
 
     // Vehicle events
-    m_Events.AddEvent("onClientVehicleCreate", "", NULL, false);
+    m_Events.AddEvent("onClientVehicleCreated", "", NULL, false);
     m_Events.AddEvent("onClientVehicleRespawn", "", NULL, false);
     m_Events.AddEvent("onClientVehicleEnter", "player, seat", NULL, false);
     m_Events.AddEvent("onClientVehicleExit", "player, seat", NULL, false);
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -2528,6 +2528,7 @@ void CClientGame::AddBuiltInEvents()
     m_Events.AddEvent("onClientPedStep", "foot", nullptr, false);
 
     // Vehicle events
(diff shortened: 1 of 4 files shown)


CrosRoad95:
https://github.com/multitheftauto/mtasa-blue/pull/218 - read comments,
make "vehicle factory" on your server that going to trigger certain event

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -2528,6 +2528,7 @@ void CClientGame::AddBuiltInEvents()
     m_Events.AddEvent("onClientPedStep", "foot", nullptr, false);
 
     // Vehicle events
+    m_Events.AddEvent("onClientVehicleCreate", "", NULL, false);
```
Lpsd: The event should be called `onClientVehicleCreated`

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2806,11 +2806,14 @@ CClientVehicle* CStaticFunctionDefinitions::CreateVehicle(CResource& Resource, u
         if (ucVariant2 == 255 && ucVariant == 255)
             CClientVehicleManager::GetRandomVariation(usModel, ucVariation, ucVariation2);
 
-        CClientVehicle* pVehicle = new CDeathmatchVehicle(m_pManager, NULL, INVALID_ELEMENT_ID, usModel, ucVariation, ucVariation2);
+        CClientVehicle* pVehicle = new CDeathmatchVehicle(m_pManager, NULL, INVALID_ELEMENT_ID, usModel, ucVariation, ucVariation2);;
```
Lpsd: Why is there an extra semicolon here?

Lpsd (commented):
Left some comments. I think we should consider some of the points raised in #218 - although this event has less scope (vehicles only), the event system is still quite slow. 

Some performance tests would be nice.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
syncc01: oh, it was a mistake

CrosRoad95:
it should be rejected because of the same reason as in #218
add "onElementCreated" or nothing, perhaps mta should have some kind of separete event system for frequently called events

Client/mods/deathmatch/logic/CClientGame.cpp
syncc01: Done

PR #2828, opened 2022-12-11
Clean up game_sa project round 10

Merlin:
Removes VC's ePedModel enum.
Cleans up few classes.
Moves the vehicle creation into CPoolSA and creates the right vehicle class. That way we can start to shift vehicle type specific stuff inside CVehicleSA into the right vehicle classes.


Client/game_sa/CHeliSA.cpp
```diff
@@ -0,0 +1,8 @@
+#include "StdInc.h"
```
botder, line 11:
```suggestion
/*****************************************************************************
 *
 *  PROJECT:     Multi Theft Auto
 *  LICENSE:     See LICENSE in the top level directory
 *  FILE:        game_sa/CHeliSA.cpp
 *
 *  Multi Theft Auto is available from http://www.multitheftauto.com/
 *
 *****************************************************************************/

#include "StdInc.h"
```

Client/game_sa/CAutomobileSA.h
```diff
@@ -163,19 +159,9 @@ static_assert(sizeof(CAutomobileSAInterface) == 0x988, "Invalid size for CAutomo
...
-    CAutomobileSA(eVehicleTypes dwModelID, unsigned char ucVariation, unsigned char ucVariation2);
-    CAutomobileSA(CAutomobileSAInterface* automobile);
-    ~CAutomobileSA();
-
-    bool BurstTyre(DWORD dwTyreID);
-    bool BreakTowLink();
-
-    bool       IsComponentPresent(int iComponentID);
-    CPhysical* SpawnFlyingComponent(int iCarNodeIndex, int iUnknown);
+    CAutomobileSA() = default;
```
botder, line 163: You didn't define a default constructor for the other classes, why did you do it here?

botder (changes requested):

Client/game_sa/CAutomobileSA.h
Merlin: CAutomobileSA will not have an implicitly-defined constructor because we have already one defined which takes a CAutomobileSAInterface*.
We need a default constructor for inherited classes to call because we don't pass the vehicle pointer down.
CHeliSA(CHeliSAInterface*) will automatically call CAutomobileSA(), CVehicleSA(), CPhysicalSA() and CEntitySA().

botder (approved):

PR #2830, opened 2022-12-14
Add a new upscaled/remade "Default New" skin as a replica (128x128 -> 128x258)

Haxardous:
This PR updates the current Default skin with a higher resoultion including remade/upscaled vectors.

("X" button is not part of the skin)*
image
image
image
image
image
image


Haxardous:
The latest commits fixes all the mentioned notes via Development's Discord.

I'll rename the issue to something else, as I have reverted the changes to the Default skin and moved the newer version to a new skin folder "Default New" and set as the new default skin.

image

Dutchman101 (dismissed):
Ideally it should have been done within this PR (as code review entries), but i come to the conclusion that all concerns, feedback and recommendation from Dev discord since the initial PR has been properly addressed in this one. Including that users who still don't feel comfortable with the rescale can switch back from "Default New" to "Default"

Dutchman101:
As it is now, only fresh installs will use "Default New" skin, and existing installations that have "Default" in coreconfig.xml need to manually switch. I think we can merge this now and use a separate PR to sort that out with a "write setting value once" in a similar way as c6f5ae2 did, while it continues to respect if the user switches back to "Default", so it wont write again. For which the registry or a `confirmed-revert` flag of some kind can be used, so it continues to only make the change once

Dutchman101:
Can we just rename it to "Default 2023", idea from botder

Haxardous:
Latest commit covers the latests requests, ready to be merged.

botder (approved):

PR #2832, opened 2022-12-16
SVG related fixes

tederis:
Simularly to PR #2831 this is doing the same: texture lock checks and removes ColorFill call that not only is incompatible with textures created in MANAGED pool(according to MSDN), but can lead to undefined behaviour when operating on locked texture.


Lpsd (approved):
tested, lgtm :shipit:

PR #2833, opened 2022-12-16
Gif player

MobinYengejehi:
hi
i made a gif player which load and play gifs as fast as possible
you can create a gif like:
```lua
local gif = gifCreate(file_path or gif_buffer) -- "logo.gif" for example

local size = Vector2(dxGetMaterialSize(gif))

addEventHandler("onClientRender",getRootElement(),function()
    if isElement(gif) then
        dxDrawImage(0,0,size,gif)
    end
end)
```
cuz it is a texture element you can use it, inside shaders or inside all dx related functions


patrikjuvonen:
Thanks for the pull request! Cool idea!

I'm wondering if there is an open source Gif loader library out there that we could use instead of implementing the intricacies ourselves, thinking from a future maintainability and security perspective.

tederis:
Great. But I think you should use the new Lua parser.

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+    IDirect3DSurface9* surface = pGif->m_pD3DRenderTargetSurface;
+    if (!surface){
+        return;
+    }
+    IDirect3DDevice9* device = pGif->m_pDevice;
+    // Lock Surface
+    D3DLOCKED_RECT LockedRect;
+    surface->LockRect(&LockedRect, nullptr, 0);
+    device->ColorFill(surface, nullptr, D3DCOLOR_RGBA(0, 0, 0, 0));
```
tederis, line 109: You cannot use ColorFill on MANAGED textures. You should also check if a texture is actually locked. See PR #2832.

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
```diff
@@ -0,0 +1,95 @@
...
+    }
+    IDirect3DSurface9* surface = pGifItem->m_pD3DRenderTargetSurface;
+    if (!surface){
+        return;
+    }
+    IDirect3DDevice9* device = pGifItem->m_pDevice;
+    // Lock Surface
+    D3DLOCKED_RECT LockedRect;
+    ZeroMemory(&LockedRect, sizeof(D3DLOCKED_RECT));
+    surface->LockRect(&LockedRect, nullptr, 0);
```
tederis: You should always check if a texture is actually locked. See PR #2832.

MobinYengejehi:
> Thanks for the pull request! Cool idea!
> 
> I'm wondering if there is an open source Gif loader library out there that we could use instead of implementing the intricacies ourselves, thinking from a future maintainability and security perspective.

yeah we can
i just used https://github.com/hidefromkgb/gif_load and made it a little better because it needed some changes. you can go and see the real code. it is from @hidefromkgb
i thought that we must write code like this
this is my first time trying to contribute mta-blue

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
```diff
@@ -0,0 +1,95 @@
...
+        return;
+    }
+    IDirect3DSurface9* surface = pGifItem->m_pD3DRenderTargetSurface;
+    if (!surface){
+        return;
+    }
+    IDirect3DDevice9* device = pGifItem->m_pDevice;
+    // Lock Surface
+    D3DLOCKED_RECT LockedRect;
+    ZeroMemory(&LockedRect, sizeof(D3DLOCKED_RECT));
```
tederis: ZeroMemory is redundant here

Client/mods/deathmatch/logic/CClientRenderElementManager.h
```diff
@@ -47,6 +49,7 @@ class CClientRenderElementManager
     uint GetScreenSourceCount() { return m_uiStatsScreenSourceCount; }
```
tederis, line 49: All these getters must be const

Client/sdk/core/CRenderItemManagerInterface.h
```diff
@@ -161,6 +162,7 @@ class CRenderItemManagerInterface
     virtual CScreenSourceItem*  CreateScreenSource(uint uiSizeX, uint uiSizeY) = 0;
     virtual CWebBrowserItem*    CreateWebBrowser(uint uiSizeX, uint uiSizeY) = 0;
     virtual CVectorGraphicItem* CreateVectorGraphic(uint uiSizeX, uint uiSizeY) = 0;
+    virtual CGifItem*           CreateGif(uint uiSizeX,uint uiSizeY) = 0;
```
tederis: Missing space here

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.cpp
```diff
@@ -0,0 +1,359 @@
...
+            return 1;
+        }
+        uint width = (uint)loader.GetWidth();
+        uint height = (uint)loader.GetHeight();
+        if (width < 1 || height < 1) {
+            m_pScriptDebugging->LogError(luaVM, "gif must be 1x1 at least");
+            lua_pushboolean(luaVM,false);
+            return 1;
+        }
+        CClientGif* gif = g_pClientGame->GetManager()->GetRenderElementManager()->CreateGif(width,height);
```
tederis: Missing space here

Client/mods/deathmatch/logic/CClientGif.cpp
```diff
@@ -0,0 +1,356 @@
...
+    m_pManager = pManager;
+    stride = GetRenderItem()->m_uiSizeX * 4;
+    m_pGifDisplay = std::make_unique<CClientGifDisplay>(m_pManager->GetDisplayManager(), this);
+}
+
+CClientGif::~CClientGif(){
+    Unlink();
+}
+
+void CClientGif::Register(std::vector<unsigned char*> frms, std::vector<int> dls) {
```
tederis: It's better to pass vectors by a reference or use a move semantics

tederis: Move would be better here

Client/mods/deathmatch/logic/CClientGif.cpp
```diff
@@ -0,0 +1,356 @@
...
+        showing = 1;
+    }else{
+        showing++;
+    }
+}
+
+void CClientGif::Unlink(){
+    for (unsigned char* frame : frames) {
+        if (frame) {
+            delete[] frame;
```
tederis: Looks dangerous. Would not it be better to use smart pointers like unique_ptr?

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,129 @@
...
+    CClientGif(CClientManager*, ElementID, CGifItem*);
+    ~CClientGif();
+    virtual void                 Unlink() override;
+    void                         Register(std::vector<unsigned char*>,std::vector<int>);
+    CResource*                   GetResource() { return m_pResource; }
+    void                         SetResource(CResource* resource) { m_pResource = resource; }
+    eClientEntityType            GetType() const { return CCLIENTGIF; }
+    CGifItem*                    GetRenderItem() const { return static_cast<CGifItem*>(m_pRenderItem); }
+    CClientGifDisplay*           GetDisplay() const { return m_pGifDisplay.get(); }
+    std::vector<unsigned char*>& GetFrames() { return frames; }
```
tederis: I think it should be const

tederis: All getters below are too

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,129 @@
...
+    CClientManager*                            m_pManager;
+    std::unique_ptr<CClientGifDisplay>         m_pGifDisplay;
+    SString                                    format = "Undefined";
+    std::vector<unsigned char*>                frames;
+    std::vector<int>                           defaultDelays;
+    std::vector<int>                           delays;
+    uint                                       frameCount = 0;
+    uint                                       stride;
+    uint                                       showing = 1;
+    double                                     tick = (double)GetTickCount64_();
```
tederis: Should be moved to the constructor

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
MobinYengejehi: at this stuff i just copied what lopsi wrote inside CClientVertualGraphicDisplay.cpp
because i didn't know what these are at first
actually i think ClearTexture is useless but cuz lopsi used it i did too
if there is no problem i can delete it too

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: do you mean i should use their address ? for example `CClientGif::Register(std::vector<unsigned char*>& frms)`

tederis: I mean you should use std::move() to pass these vectors. It should be 
`void CClientGif::Register(std::vector<unsigned char*>&& frms, std::vector<int>&& dls)`

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: how should i use smart pointers for char*? 
like `unique_ptr<unsigend char[]>` ?
if yes then how should i set its size when it is allocating the memory?
i checked it but i didn't see any problems

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
MobinYengejehi: fixed this one!

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
tederis: It looks like even in SVG code ClearTexture is never called. But I'm not sure.

Client/mods/deathmatch/logic/CClientGif.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
tederis: Yes, you can use `unique_ptr<unsigend char[]>`. But there is a better solution: use `std::vector<unsigned char>`. It will automatically release all the stuff inside.

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
tederis: When working with raw ptrs it's always dangerous.

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.cpp
MobinYengejehi: i couldn't understand what you mean from 'Missing space'

tederis: The space between **width** and **height**

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
```diff
@@ -0,0 +1,357 @@
...
+    UpdateTick();
+    m_pGifDisplay = std::make_unique<CClientGifDisplay>(m_pManager->GetDisplayManager(), this);
+}
+
+CClientGif::~CClientGif(){
+    Unlink();
+}
+
+void CClientGif::Register(std::vector<unsigned char*>&& frms, std::vector<int>&& dls) {
+    frames = frms;
```
tederis: In order to use the whole power of move semantics you also have to move arrays here: `frames = std::move(frms)` and `delays = std::move(dls)`. But since dls is just moved, defaultDelays should be `defaultDelays = delays` now;

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,129 @@
...
+        uint8_t r;
+        uint8_t g;
+        uint8_t b;
+    }* pallete;
+    SString        format = "Undefined";
+    uint8_t*       buffer = nullptr;
+    uint8_t*       address = nullptr;
+    void*          palleteDecoder = nullptr;
+    void*          palleteDecoderPrevious = nullptr;
+    long           size = 0;
```
tederis: Also all class members must have prefix m_ to explicitly identify them.

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
```diff
@@ -0,0 +1,357 @@
...
+}
+
+CClientGif::~CClientGif(){
+    Unlink();
+}
+
+void CClientGif::Register(std::vector<unsigned char*>&& frms, std::vector<int>&& dls) {
+    frames = std::move(frms);
+    delays = std::move(dls);
+    defaultDelays = dls;
```
tederis: Be aware that `dls` is now empty and thus `defaultDelays` is also empty. You should copy `defaultDelays` from `delays`.

MobinYengejehi: ah yeah your right i just forgot that

Client/mods/deathmatch/logic/CClientGif.h
MobinYengejehi: because those were private i defined them like that?
is changing them important?

tederis: This is the code style: https://github.com/multitheftauto/mtasa-blue/wiki/Style-Guide

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.cpp
MobinYengejehi: aha fixed

Client/sdk/core/CRenderItemManagerInterface.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientRenderElementManager.h
MobinYengejehi: because i didn't write these codes i perfer not to touch them
if you can fix it yourself

Client/mods/deathmatch/logic/CClientGif.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: fixed this problem too but it works a little bit slower now
for example it could load a 500x240 gif with 15 frames in 23ms but now it loads it in 78ms

MobinYengejehi:
is there any other problems that i should solve?

CrosRoad95:
In my opinion your solution is over engineered, instead i propose to make "createGif" returns a regular texture ( texture atlas ) + number of frames.

MobinYengejehi:
it does. it will return a texture element that is gif and if you want to get frames you can use `gifGetFrameCount(element gif)`
and also it has a player that you can play or stop or you can set its speed too
for example:
```lua
local gif = gifCreate("img.gif")
local width,height = dxGetMaterialSize(gif)

-- you can adjust the delay of each frame
local newDelay = 5 -- ms : 5*(10^-3) s
-- 5 ms can make gif really fast!

local frameCount = gifGetFrameCount(gif)
for frame = 1,frameCount do -- for all frames
    gifSetFrameDelay(gif,frame,newDelay)
end

--

gifPlay(gif) -- play the gif

addEventHandler("onClientRender",root,function()
    if isElement(gif) then
        dxDrwaImage(0,0,width,height,gif)
    end
end)
```
it s really not that hard!

CrosRoad95:
in my opinion these all functions are useless, it is trivial to calculate what frame should be displayed and calculating uv coors is not a rocket science

MobinYengejehi:
but i think they can be useful !
for example imagine you are writing a phone script and you have an application like *whatsapp
for making that you have 2 ways:
1- using browser -- which is hard to load on weak computers (that in some countries most of the people who are playing MTA have)
2- using dx functions
if you choose the dx interface you need some functions that will allow you to control your gifs
and many other ideas that scripters can have when they want to use dx functions instead of browser

MobinYengejehi:
but your opinion is respected and if you want we can remove them

Disinterpreter:
IMHO, half useless function. There aren't any questions about implementation for this. But, sorry, all world is trying to move out from GIF to APNG, WEBP, and Lottie of course, modern things for today. Even huge GIF hostings are storing GIFs in MP4 format.

MobinYengejehi:
Ok

MobinYengejehi:
well i removed useless functions.
the functions that have been removed are:
```lua
gifGetTick(element [gif (dx material)]); -> int, boolean
gifGetFormat(element [gif (dx material)]); -> string, boolean
gifGetFrameDelay(element [gif (dx material)], int [frame index]); -> int, boolean
gifSetFrameDelay(element [gif (dx material)], int [frame index], int [delay]); -> boolean
gifGetShowingFrame(element [gif (dx material)]); -> int
gifGetFrameDefaultDelay(element [gif (dx material)], int [frame index]); -> int
gifGetFrameCount(element [gif (dx material)]); -> int
```
and the remaining functions are:
```lua
gifCreate(string [path or rawdata]); -> gif element (dx material) -- this function creates a gif element
gifPlay(element [gif (dx material)]); -> boolean -- this function plays the gif if it is stopped
gifStop(element [gif (dx material)]); -> boolean -- this function stops the gif if it is playing
gifNavigateToThumbnail(element [gif (dx material)]); -> boolean -- this function shows the first frame
isGifPlaying(element [gif (dx material)]); -> boolean -- this function checks if gif is playing or not
```

MobinYengejehi:
i hope it will be ok this time!

Client/mods/deathmatch/logic/CClientRenderElementManager.h
MobinYengejehi: fixed!

thisdp:
Maybe implementing gifSetShowingFrame instead of gifNavigateToThumbnail will be better.
And isGifPlaying should be gifIsPlaying

MobinYengejehi:
for `gifSetShowingFrame` i think gif player needs some functions like `gifGetShowingFrame` & `gifGetFrameCount`
but i removed them cuz your friends said it should be simple
i can return them if you want

and `isGifPlaying` changed to `gifIsPlaying`!

thisdp:
> ```lua
> gifPlay
> ```

then, maybe we can specify the first frame in gifPlay or gifStop, in this way gifNavigateToThumbnail can be removed.

MobinYengejehi:
then can i return `gifGetFrameCount` and `gifGetShowingFrame`?

thisdp:
> then can i return `gifGetFrameCount` and `gifGetShowingFrame`?

maybe gifGetDetails, or gifGetProperty/gifSetProperty

MobinYengejehi:
ah yeah your right this one is better
i will do it now

thisdp:
> ah yeah your right this one is better i will do it now

gifGetProperty(gif,"delay")
gifSetProperty(gif,"delay",100)

MobinYengejehi:
delay works per frame
i mean each frame has a different delay like `gifSetFrameDelay(gif, frame, delay)`

MobinYengejehi:
what should we do for it?

thisdp:
> delay works per frame i mean each frame has a different delay like `gifSetFrameDelay(gif, frame, delay)`

then, maybe gifSetProperty(gif,frame,"delay",100) for frame property.
If frame is not specified, the property will be global

MobinYengejehi:
aha ok
after 30 minutes i finish them

MobinYengejehi:
and i think it is better we set showing frame with `gifSetProperty(gif,"showing_frame",frameindex)` too
whats your opinion? isn't it better?

thisdp:
> and i think it is better we set showing frame with `gifSetProperty(gif,"showing_frame",frameindex)` too whats your opinion?

Good idea

thisdp:
In this way, gifCreate, gifPlay, gifStop and gifIsPlaying will be mainly used functions. For rarely used functions, we can just use gifSetProperty/gifGetProperty.

thisdp:
Better to use new lua parser. Refer to #2535

MobinYengejehi:
`gifGetProperty` return types are string or number
how should i return something like that when im using new lua parser?

thisdp:
> `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?

for example:
 std::variant<bool, CLuaMultiReturn<char, char>>

and you can even use this for passing arguments:
std::variant<bool, float> argumentN

MobinYengejehi:
new functions:
```lua
gifGetProperty(element [gif (dx material)], [int (frame index)] string [property]);
gifSetProeprty(element [gif (dx material)], [int (frame index)] string [property], string or number [value]);
```

gif get properties:
    `Frame`:
        `delay` : gets the frame delay
        `default_delay` : gets the frame default delay
    `Global`:
        `showing_frame` : gets the showing frame
        `frame_count` : gets the loaded frames count
        `format` : gets the gif format `GIF89a` or `GIF87a`
        `tick` : gets last updated tick

gif set properties:
    `Frame`:
        `delay` : sets frame delay
    `Global`:
        `showing_frame` : sets showing frame

MobinYengejehi:
> > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> 
> for example: std::variant<bool, CLuaMultiReturn<char, char>>
> 
> and you can even use this for passing arguments: std::variant<bool, float> argumentN

ok im working on it

MobinYengejehi:
> > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> 
> for example: std::variant<bool, CLuaMultiReturn<char, char>>
> 
> and you can even use this for passing arguments: std::variant<bool, float> argumentN

and what about arguments

for example in the `gifSetProperty` function if second argument be number it will set property for a single frame but if it be string it will be for global
how should i make it like that with new lua parser ?

thisdp:
> > > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> > 
> > 
> > for example: std::variant<bool, CLuaMultiReturn<char, char>>
> > and you can even use this for passing arguments: std::variant<bool, float> argumentN
> 
> and what about arguments
> 
> for example in the `gifSetProperty` function if second argument be number it will set property for a single frame but if it be string it will be for global how should i make it like that with new lua parser ?

refer to "SetColPolygonHeight" definition

MobinYengejehi:
> > > > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> > > 
> > > 
> > > for example: std::variant<bool, CLuaMultiReturn<char, char>>
> > > and you can even use this for passing arguments: std::variant<bool, float> argumentN
> > 
> > 
> > and what about arguments
> > for example in the `gifSetProperty` function if second argument be number it will set property for a single frame but if it be string it will be for global how should i make it like that with new lua parser ?
> 
> refer to "SetColPolygonHeight" definition

and another thing is reading file buffer
for that i need to have resource and lua main
how should i get this one?

thisdp:
> > > > > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> > > > 
> > > > 
> > > > for example: std::variant<bool, CLuaMultiReturn<char, char>>
> > > > and you can even use this for passing arguments: std::variant<bool, float> argumentN
> > > 
> > > 
> > > and what about arguments
> > > for example in the `gifSetProperty` function if second argument be number it will set property for a single frame but if it be string it will be for global how should i make it like that with new lua parser ?
> > 
> > 
> > refer to "SetColPolygonHeight" definition
> 
> and another thing is reading file buffer for that i need to have resource and lua main how should i get this one?

maybe refer to "svgCreate"

MobinYengejehi:
> > > > > > `gifGetProperty` return types are string or number how should i return something like that when im using new lua parser?
> > > > > 
> > > > > 
> > > > > for example: std::variant<bool, CLuaMultiReturn<char, char>>
> > > > > and you can even use this for passing arguments: std::variant<bool, float> argumentN
> > > > 
> > > > 
> > > > and what about arguments
> > > > for example in the `gifSetProperty` function if second argument be number it will set property for a single frame but if it be string it will be for global how should i make it like that with new lua parser ?
> > > 
> > > 
> > > refer to "SetColPolygonHeight" definition
> > 
> > 
> > and another thing is reading file buffer for that i need to have resource and lua main how should i get this one?
> 
> maybe refer to "svgCreate"

tnx a lot
now it is using new lua parser too

and `gifGetProperty` & `gifSetProperty` have new definitions:
```lua
gifGetProperty(element [gif (dx material)], string [property], [int frame]); -- last argument is optional
gifSetProperty(element [gif (dx material)], string [property], int or string [value], [int frame]); -- last argument is optional

File: Client/core/Graphics/CRenderItemManager.cpp [hunk 14]
```diff
@@ -171,6 +171,32 @@ CVectorGraphicItem* CRenderItemManager::CreateVectorGraphic(uint width, uint hei
...
+//
+//
+//
+////////////////////////////////////////////////////////////////
+CGifItem* CRenderItemManager::CreateGif(uint width, uint height)
+{
+    if (!CanCreateRenderItem(CGifItem::GetClassId()))
+        return nullptr;
+
+    CGifItem* pVectorItem = new CGifItem;
```
Lpsd: `pVectorItem` -> `pGifItem`

Client/mods/deathmatch/logic/CClientGif.cpp
```diff
@@ -0,0 +1,355 @@
...
+ *****************************************************************************/
+
+#include "StdInc.h"
+#include "CClientGif.h"
+
+#define GIF_ALLOCATE(pointer,size,release) pointer = (uint8_t*)realloc((release) ? 0 : pointer,(release) ? size : 0UL)
+#define BIGE 0
+#define _SWAP(h) ((BIGE) ? ((uint16_t)(h << 8) | (h << 8)) : h)
+
+CClientGif::CClientGif(CClientManager* pManager, ElementID ID, CGifItem* p_GifItem)
```
Lpsd: `p_GifItem` -> `pGifItem`

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,129 @@
...
+    CClientManager*                            m_pManager;
+    std::unique_ptr<CClientGifDisplay>         m_pGifDisplay;
+    SString                                    m_bFormat = "Undefined";
+    std::vector<std::vector<unsigned char>>    m_bFrames;
+    std::vector<int>                           m_bDefaultDelays;
+    std::vector<int>                           m_bDelays;
+    uint                                       m_bFrameCount = 0;
+    uint                                       m_bStride;
+    uint                                       m_bShowing = 1;
+    double                                     m_bTick = 0;
```
Lpsd: Incorrect usage of hungarian notation. `b` would indicate that it's a boolean value

```cpp
SString                                    m_strFormat = "Undefined";
std::vector<std::vector<unsigned char>>    m_vecFrames;
std::vector<int>                           m_vecDefaultDelays;
std::vector<int>                           m_vecDelays;
uint                                       m_uiFrameCount = 0;
uint                                       m_uiStride;
uint                                       m_uiShowing = 1;
double                                     m_dbTick = 0;
```

Client/mods/deathmatch/logic/CClientRenderElementManager.h
```diff
@@ -36,6 +37,7 @@ class CClientRenderElementManager
     CClientScreenSource*  CreateScreenSource(uint uiSizeX, uint uiSizeY);
     CClientWebBrowser*    CreateWebBrowser(uint uiSizeX, uint uiSizeY, bool bIsLocal, bool bTransparent);
     CClientVectorGraphic* CreateVectorGraphic(uint width, uint height);
+    CClientGif*           CreateGif(uint width,uint height);
```
Lpsd: Formatting is kind of all over the place (for example, there should be a space after each comma for the parameters.

Are you using the clang-format included with the repo? Visual Studio _should_ pick this up and allow you to format the current working file with the default shortcut. (`Ctrl + K, Ctrl + D`).

You could also run [win-apply-clang-format.bat](https://github.com/multitheftauto/mtasa-blue/blob/master/utils/win-apply-clang-format.bat) but it's probably going to create a ton of unrelated formatting changes so I'd advise against it for now.

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.cpp
```diff
@@ -0,0 +1,205 @@
...
+    }else{
+        if (property == "showing_frame") {
+            return (int)gif->GetShowingFrame();
+        }else if (property == "frame_count") {
+            return (int)gif->GetFrameCount();
+        }else if (property == "format") {
+            return (std::string)gif->GetFormat();
+        }else if (property == "tick") {
+            return (int)gif->GetTick();
+        }else{
```
Lpsd: Again, need the clang formatting running for all of the changed files in this PR.

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.h
```diff
@@ -0,0 +1,26 @@
...
+{
+public:
+    static void                                  LoadFunctions();
+    static void                                  AddClass(lua_State*);
+    static CClientGif*                           GifCreate(lua_State*, std::string);
+    static bool                                  GifPlay(CClientGif*);
+    static bool                                  GifStop(CClientGif*);
+    static bool                                  GifSetProperty(lua_State*, CClientGif*, std::string, std::variant<int, std::string>, std::optional<int>);
+    static std::variant<bool, std::string, int>  GifGetProperty(lua_State*, CClientGif*, std::string, std::optional<int>);
+    static bool                                  GifIsPlaying(CClientGif*);
```
Lpsd: Why do these parameters have no names?

Lpsd (dismissed):

Client/core/Graphics/CRenderItemManager.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientGif.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/CClientRenderElementManager.h
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.cpp
MobinYengejehi: fixed!

Client/mods/deathmatch/logic/luadefs/CLuaGifDefs.h
MobinYengejehi: fixed!

MobinYengejehi:
i hope it will be ok this time

MobinYengejehi:
@Lpsd  and you requested to change a file but i don't know how to accept that
can you help me in this way?

MobinYengejehi:
😔

lopezloo:
Do we really need this?

CrosRoad95:
> Do we really need this?

actualy, no. If you for example want to have gif in mta, i bet there is a program that can turn it into atlas texture with every frame.
If you want to have dynamic gifs, let's say gif avatars then use for example: https://www.npmjs.com/package/gif-to-png

dmi7ry:
> > Do we really need this?
> 
> actualy, no. If you for example want to have gif in mta, i bet there is a program that can turn it into atlas texture with every frame. If you want to have dynamic gifs, let's say gif avatars then use for example: https://www.npmjs.com/package/gif-to-png

Keep in mind that GIF is not just a collection of pictures displayed one after the other. GIF allows you to set pauses between frames and other things. If such GIF is directly converted to a png set, then the size of the textures will be much MUCH larger.
And to avoid this, you still have to write a sequencer that will take into account all the pauses between frames, etc. I think that this will be somewhat difficult for many developers.

CrosRoad95:
> take into account all the pauses between frames, etc. 

Probably along with frames you can use some tool to extract metadata.

dmi7ry:
> > take into account all the pauses between frames, etc.
> Probably along with frames you can use some tool to extract metadata.

Me personally prefer to use Rive and Spine. But how about not very experienced deveopers? Of course this means that just no one will use animations, and therefore the interfaces will look less interesting than they could be.

MobinYengejehi:
> > Do we really need this?
> 
> actualy, no. If you for example want to have gif in mta, i bet there is a program that can turn it into atlas texture with every frame. If you want to have dynamic gifs, let's say gif avatars then use for example: https://www.npmjs.com/package/gif-to-png

well then you won't be able to use shaders to control the texture pixels
this feature is not that important cuz developers can use browsers which is much better
but my experience says CEF sometimes doesn't work properly on weak cpus
actually it happened to me multiple times and it wasn't only me who experienced that
mta created browser and also `onClientBrowserCreated` event was triggered but my display was empty
it seemed like cef didn't change browser texture pixels and all of those was 0x00000000
so i decided to contribute mta and add this feature that is fast and allows developers save some frames inside a gif and use that frames the way they want
and also because this gif element is a dx texture developers can use shaders for it and also apply this shaders on models
for example animated labels on vehicles or something like that
and the way to use it is like this:
```lua
local gif = gifCreate("label.gif");

local shader = dxCreateShader("shader.fx");
assert(shader, "couldn't create the shader!");

dxSetShaderValue(shader, "gif", gif);

addCommandHandler("startanim", function()
    gifPlay(gif);
);

addCommandHandler("stopanim", function()
    gifStop(gif);
end);

addCommandHandler("setlabel", function()
    -- engineApplyShaderToWorldTexture ...
end);
```
```glsl
texture gif;
sampler2D gifSampler{
    Texture = gif;
};

float4 pixels(float2 coords : TEXCOORD0) : COLOR0 {
    // and for example i want to remove green background here
    float4 color = tex2D(gifSampler, coords);
    
    if (color == float4(0, 1, 0, 1)) {
        return (float4)0;
    }
    
    return color;
}

technique Render{
    pass p0 {
        PixelShader = compile ps_2_a pixels();
    }
}
```
this feature has less fps drop than the render target or other possible methods
with this api also the clients who use weak systems can experience this animated labels on their vehicles

PlatinMTA:
The shader idea is kinda cool, I can see myself using it. I would use it to display some forum avatars aswell. 

``this feature has less fps drop than the render target or other possible methods``
Just out of curiosity, do you have some benchmarks about that? Just having the easy implementation is enough, but having it be faster is a no brainer if you need to use .gifs.

MobinYengejehi:
> The shader idea is kinda cool, I can see myself using it. I would use it to display some forum avatars aswell.
> 
> `this feature has less fps drop than the render target or other possible methods` Just out of curiosity, do you have some benchmarks about that? Just having the easy implementation is enough, but having it be faster is a no brainer if you need to use .gifs.

actually i was thinking about playing videos on mta client
i know we can play videos or other medias with CEF
but as i mentioned mta sometimes can't load browser textures
so i decided to somehow create my own library to load videos but i knew lua isn't that fast that i can load my webm buffer and display it on screen
i thought maybe if we had some features in mta that we could allocate memory to save frames and use them can help me.
at first i used textures for that
when i tested it, it took 30 seconds to load only 50 frames and after all frames loaded when i displayed them on screen it was using 8% of my cpu
and most important thing was that creating the texture occupied the memory of graphic card
so i thought if i create only one texture and allocate some bytes from computer memory to save my frames and send that bytes to graphic card memory this can fix my problem (and this is exactly what this gif player does)
then i came up with this gif player idea and decided to make it and ask mta team if they like my idea add it to the launcher
when i used gif player to load my webm frames, it took less than 100 ms to load more than 700 frames and on display, it was 
only using something like 0.5% of my cpu
and you can control your player to display frames with `gifSetProperty` and `gifGetProperty` functions

> new functions:
> 
> ```lua
> gifGetProperty(element [gif (dx material)], [int (frame index)] string [property]);
> gifSetProperty(element [gif (dx material)], [int (frame index)] string [property], string or number [value]);
> ```
> 
> gif get properties: 
> `Frame`: 
> > `delay` : gets the frame delay
> > `default_delay` : gets the frame default delay 
>
> `Global`: 
> > `showing_frame` : gets the showing frame 
> > `frame_count` : gets the loaded frames count 
> > `format` : gets the gif format `GIF89a` or `GIF87a`
> > `tick` : gets last updated tick
> 
> gif set properties: 
> `Frame`: 
> > `delay` : sets frame delay 
>
> `Global`: 
> > `showing_frame` : sets showing frame

you know i don't want to open another application in my computers background and create other vm s to display a simple gif

tederis:
@MobinYengejehi Please, resolve the conflicts.

MobinYengejehi:
@tederis conflicts have been resolved

Client/mods/deathmatch/logic/CClientRenderElementManager.h
```diff
@@ -36,6 +37,7 @@ class CClientRenderElementManager
     CClientScreenSource*  CreateScreenSource(uint uiSizeX, uint uiSizeY);
     CClientWebBrowser*    CreateWebBrowser(uint uiSizeX, uint uiSizeY, bool bIsLocal, bool bTransparent);
     CClientVectorGraphic* CreateVectorGraphic(uint width, uint height);
+    CClientGif*           CreateGif(uint width, uint height);
```
FileEX, line 40: uint -> std::uint32_t

Client/mods/deathmatch/logic/CClientRenderElementManager.h
```diff
@@ -47,6 +49,7 @@ class CClientRenderElementManager
     uint GetScreenSourceCount() { return m_uiStatsScreenSourceCount; }
     uint GetWebBrowserCount() { return m_uiStatsWebBrowserCount; }
     uint GetVectorGraphicCount() { return m_uiStatsVectorGraphicCount; }
+    uint GetGifCount() { return m_uiStatsGifCount; }
```
FileEX, line 52: uint -> std::uint32_t

Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
```diff
@@ -273,6 +275,35 @@ CClientVectorGraphic* CClientRenderElementManager::CreateVectorGraphic(uint widt
...
+//
+// CClientRenderElementManager::CreateGif
+//
+//
+//
+////////////////////////////////////////////////////////////////
+CClientGif* CClientRenderElementManager::CreateGif(uint width, uint height)
+{
+    // Create the item
+    CGifItem* pGifItem = m_pRenderItemManager->CreateGif(width, height);
```
FileEX, line 288: We no longer use Hungarian notation.

pGifItem -> gifItem

Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
```diff
@@ -273,6 +275,35 @@ CClientVectorGraphic* CClientRenderElementManager::CreateVectorGraphic(uint widt
...
+
+    // Check create worked
+    if (!pGifItem)
+    {
+        return nullptr;
+    }
```
FileEX, line 294: You can remove braces here

Client/mods/deathmatch/logic/CClientGifDisplay.h
```diff
@@ -0,0 +1,33 @@
...
+    ~CClientGifDisplay() = default;
+    eDisplayType GetType() { return DISPLAY_GIF; }
+    void         Render();
+    void         UpdateTexture();
+    void         ClearTexture();
+    bool         IsCleared() { return m_bIsCleared; }
+
+private:
+    CClientGif* m_pGif;
+    bool        m_bIsCleared;
```
FileEX, line 32: We no longer use Hungarian notation.

Client/mods/deathmatch/logic/CClientGifDisplay.h
```diff
@@ -0,0 +1,33 @@
...
+#include "CClientDisplayManager.h"
+
+class CClientGifDisplay final : public CClientDisplay
+{
+    friend class CClientDisplayManager;
+
+public:
+    CClientGifDisplay(CClientDisplayManager* pDisplayManager, CClientGif* pGif, int ID = DISPLAY_GIF);
+    ~CClientGifDisplay() = default;
+    eDisplayType GetType() { return DISPLAY_GIF; }
```
FileEX, line 24: const noexcept

Client/mods/deathmatch/logic/CClientGifDisplay.h
```diff
@@ -0,0 +1,33 @@
...
+    void         Render();
+    void         UpdateTexture();
+    void         ClearTexture();
+    bool         IsCleared() { return m_bIsCleared; }
```
FileEX, line 28: const neoxcept

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include "StdInc.h"
+#include "CClientGifDisplay.h"
+#include "CClientGif.h"
+
+CClientGifDisplay::CClientGifDisplay(CClientDisplayManager* pDisplayManager, CClientGif* pGif, int ID) : CClientDisplay(pDisplayManager, ID)
+{
+    m_pGif = pGif;
+    m_bVisible = true;
+    m_bIsCleared = false;
```
FileEX, line 18: We no longer use Hungarian notation.

Client/mods/deathmatch/logic/CClientGifDisplay.cpp
```diff
@@ -0,0 +1,113 @@
...
+        }
+        return;
+    }
+    else if (IsCleared())
+    {
+        m_bIsCleared = false;
+    }
+    if (m_pGif->IsPlaying())
+    {
+        if ((double)GetTickCount64_() >= m_pGif->GetTick() + m_pGif->GetFrameDelay(-1))
```
FileEX, line 41: Use `static_cast`

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    SString                                 m_strFormat = "Undefined";
+    std::vector<std::vector<unsigned char>> m_vecFrames;
+    std::vector<int>                        m_vecDefaultDelays;
+    std::vector<int>                        m_vecDelays;
+    uint                                    m_uiFrameCount = 0;
+    uint                                    m_uiStride;
+    uint                                    m_uiShowing = 1;
+    double                                  m_dbTick = 0;
+    bool                                    m_bPlaying = false;
+    bool                                    m_bIsDestoryed = false;
```
FileEX, line 139: We no longer use Hungarian notation.

SString -> std::string?
uint -> std::uint32_t

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    void SetFrame(const int& frame) { if (frame <= GetImageCount()) m_uiShowing = 1; }
+
+    void UpdateTick() { m_dbTick = (double)GetTickCount64_(); }
+    void Play() { m_bPlaying = true; }
+    void Stop() { m_bPlaying = false; }
+    void Next();
+    void NavigateToThumbnail() { m_uiShowing = 1; }
+
+    bool IsPlaying() const { return m_bPlaying; }
+    bool IsDestoryed() const { return m_bIsDestoryed; }
```
FileEX, line 124: noexcept

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    uint                                    GetFrameCount() const { return m_uiFrameCount; }
+
+    void SetFrameDelay(const uint& id, const uint32_t& delay) { m_vecDelays[(id < 1 ? m_uiShowing : (id > GetImageCount() ? (m_uiShowing) : id)) - 1] = delay; }
+    void SetFormat(const SString& fmt) { if (!fmt.empty()) m_strFormat = fmt; }
+    void SetFrameCount(const uint& count) { m_uiFrameCount = count; }
+    void SetFrame(const int& frame) { if (frame <= GetImageCount()) m_uiShowing = 1; }
+
+    void UpdateTick() { m_dbTick = (double)GetTickCount64_(); }
+    void Play() { m_bPlaying = true; }
+    void Stop() { m_bPlaying = false; }
```
FileEX, line 119: `static_cast` + noexcept

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    long           m_lgBackground = 0;
+    long           m_lgTransparent = 0;
+    long           m_lgInterlaced = 0;
+    long           m_lgDelay = 0;
+    long           m_lgFrameNumber = 0;
+    long           m_uiFrameCount = 0;
+    long           m_lgMode = CGif_None;
+    unsigned long  m_ulgLast = 0;
+    CClientGifRect m_rcRect = {0, 0, 0, 0};
+    bool           m_bError = false;
```
FileEX, line 82: We no longer use Hungarian notation.

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    static long SkipChunk(uint8_t** buffer, long size);
+    static long LoadHeader(unsigned globalFlags, uint8_t** buffer, void** pallete, unsigned state, long* size, long length);
+    static long LoadFrame(uint8_t** buffer, long* size, uint8_t* address, uint8_t* blen);
+
+private:
+    struct
+    {
+        uint8_t r;
+        uint8_t g;
+        uint8_t b;
```
FileEX, line 62: std::uint8_t

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    }*             m_pPallete;
+    SString        m_strFormat = "Undefined";
+    uint8_t*       m_pBuffer = nullptr;
+    uint8_t*       m_pAddress = nullptr;
```
FileEX, line 66: std::uint8_t

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+
+    void     CreateFrame(std::vector<std::vector<unsigned char>>& frames, std::vector<int>& delays);
+    void     Load(std::vector<std::vector<unsigned char>>& frames, std::vector<int>& delays, long skip);
+    uint8_t* GetBuffer() { return m_pBuffer; }
+    long&    GetWidth() { return m_lgWidth; }
+    long&    GetHeight() { return m_lgHeight; }
+    long&    GetDelay() { return m_lgDelay; }
+    long&    GetFrameNumber() { return m_lgFrameNumber; }
+    long&    GetFrameCount() { return m_uiFrameCount; }
+    SString  GetFormat() { return m_strFormat; }
```
FileEX, line 50: const + noexcept

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    CClientGifRect() = default;
+    CClientGifRect(const uint& l, const uint& t, const uint& w, const uint& h) : left(l), top(t), width(w), height(h) { valid = true; }
+};
+
+class CClientGifLoader
+{
+public:
+    CClientGifLoader() = default;
+    CClientGifLoader(SString& data);
+    CClientGifLoader(uint8_t* data, unsigned long dataSize);
```
FileEX, line 39: std::uint8_t

Client/mods/deathmatch/logic/CClientGif.h
```diff
@@ -0,0 +1,140 @@
...
+    CGif_Previous = 3
+};
+
+class CClientGifRect
+{
+public:
+    uint left = 0;
+    uint top = 0;
+    uint width = 0;
+    uint height = 0;
```
FileEX, line 28: std::uint32_t

Client/core/Graphics/CRenderItemManager.h
```diff
@@ -34,6 +34,7 @@ class CRenderItemManager : public CRenderItemManagerInterface
                                                    bool bForce = false);
     virtual CScreenSourceItem*  CreateScreenSource(uint uiSizeX, uint uiSizeY);
     virtual CVectorGraphicItem* CreateVectorGraphic(uint width, uint height);
+    virtual CGifItem*           CreateGif(uint width,uint height);
```
FileEX, line 37: std::uint32_t

Client/core/Graphics/CRenderItem.Gif.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+
+////////////////////////////////////////////////////////////////
+//
+// CGifItem::IsValid
+//
+// Check underlying data is present
+//
+////////////////////////////////////////////////////////////////
+bool CGifItem::IsValid()
```
FileEX, line 51: const + noexcept?

FileEX (changes requested):
The code generally looks very messy. There are missing line breaks, everything is stacked together, and inside functions there are definitions and structures inserted. It all looks very untidy and needs cleaning up for better readability.

It might be worth reviewing the guidelines:
https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md

tederis:
@MobinYengejehi Please, address the conversations given above.

PR #2849, opened 2022-12-25
Possible crash fix for libcef(0x3705563)

tederis:
After some research and having access to some player's dumps I've found the source of crashes within libcef.dll(at 0x3705563). These crashes appear when GOOGLE API keys are registered in OS. You can reproduce the crash by specifying a value for system registry keys GOOGLE_API_KEY, GOOGLE_DEFAULT_CLIENT_ID and GOOGLE_DEFAULT_CLIENT_SECRET. For some reason key signin.allowed_on_next_startup is not registered in the chromium preferences and chrome is trying to dereference a null pointer. I tried different ways to add signin.allowed_on_next_startup into preferences but failed at everything. The only way I found to bypass this is using CEF command line to override allow-browser-signin switch. It works for me, but comprehensive tests required.

### DIFF
--- a/Client/cefweb/CWebApp.cpp
+++ b/Client/cefweb/CWebApp.cpp
@@ -32,6 +32,9 @@ void CWebApp::OnBeforeCommandLineProcessing(const CefString& process_type, CefRe
     // command_line->AppendSwitch("disable-d3d11");
     command_line->AppendSwitch("enable-begin-frame-scheduling");
 
+    // browser-signin switch(or lack thereof) produces crashes when GOOGLE API keys are present in the OS registry
+    command_line->AppendSwitchWithValue("allow-browser-signin", "false");
+
     if (process_type.empty())
     {
         command_line->AppendSwitchWithValue("autoplay-policy", "no-user-gesture-required");


sbx320:
It might be worth reporting this issue upstream with CEF

patrikjuvonen:
I found this thread https://groups.google.com/a/chromium.org/g/chromium-packagers/c/SG6jnsP4pWM?pli=1

> if you set a client_id and secret, you'll have to make sure that the Google signin code is disabled, e.g., by passing the flag --allow-browser-signin=false as the signin integration does not fail gracefully once the quota on the API is exceeded.
https://groups.google.com/a/chromium.org/g/chromium-packagers/c/SG6jnsP4pWM/m/c4YTWPixAgAJ

Seems that the change in this PR is required.

Dutchman101:
This is 1 of the 2 biggest CEF crashes on MTA, judging by our crash stats.. it's very nice of TEDERIs to dig into it, i vouch for adding this workaround as soon as possible for short term relief, and then notifying CEF authors of the issue.

Top CEF crashes in MTA (that don't seem to fade away over time.. as upstream fixes a thing here and there) always come down to either CEF bugs, Chromium bugs, or MTA's implementation of CEF being flawed. The recent PR's #2831 and #2844 did good work for the part of our implementation, and now i think the 2 top crashes that are left, this one 03705563, and 013f8fd9 (note: offsets shift with each CEF version update) may be pure CEF bugs. I've forwarded a dump for the other Top 2 CEF crash (013f8fd9) to TEDERIs a minute ago as well, should he be interested to figure that one out as well, and then CEF on MTA was made much more stable this week, with this PR included

* For reference to upstream maintainers: all references made are against our current version of CEF, `108.4.13+ga98cd4c+chromium-108.0.5359.125`

cc @magreenblatt (CEF maintainer), maybe we will appear on your tracker soon with an issue report. Maybe by then we'll also have collected information on the other top crash (013f8fd9 as mentioned above)

patrikjuvonen (approved):
Tested working, thanks for the pull request.

Dutchman101:
Possible fix to fix.. it's confirmed in a few support cases, where users presented with the crash and were asked to install a nightly build that includes this PR

I suspect that this was a new CEF crash, brought on by `108.4.13+ga98cd4c+chromium-108.0.5359.125` in some way, yet it doesn't happen to everyone, likely limited to having the stuff you mentioned in OP on their PC (due to using other Google products?) but to who it happens, it seems to do so on all MTA servers. If someone wants to file a bug report upstream, im not available for that yet

PR #2855, opened 2022-12-28
Tweak mouse cursor startup behaviour

lopezloo:
This PR:
- Disables rendering of MTA (CEGUI) cursor during game startup (because it's frozen anyway).
- Disables changing system cursor position to the center on game startup if game is being launched in windowed mode or user has multiple monitors and full screen minimize is disabled (it's annoying). 
- Enables rendering of system cursor within MTA window during game startup if game is being launched in windowed mode or user has multiple monitors and full screen minimize is disabled (feels less _claustrophobic_ this way). 
- Removes some strange undocumented condition from `CLocalGUI::InputGoesToGUI()`.

At first I wanted to disable this centering and show system mouse cursor only when game is being launched in windowed mode. But then I started testing it more and it feels right on fullscreen mode too to me. I literally feels less _claustrophobic_ this way to me. You see your cursor and you can switch to other screen for example while this initial credits screen is showing. Games usually don't do that so we can discuss but for me - it simply feels good.

(startup in this context means initial SA credits screen technically)


patrikjuvonen:
From what I understand cursor centering is useful in fullscreen because sometimes on multi-display setups your cursor is "stuck" somewhere outside the game and if you click once you'll end up having to alt-tab back into the game, so it prevents that back and forth. For windowed mode this is less of a problem.

turret001:
I'm not sure if this is considered offtopic but as a user, I don't really care where my mouse is when I launch the game. Centering it automatically feels very restrict-y and makes me feel like I'm being trapped in a Chinese finger lock.

PlatinMTA:
can't really tell what's the benefit considering I always play on windowed mode

Client/core/CCommandFuncs.cpp
```diff
@@ -21,12 +21,6 @@ extern bool   g_bBoundsChecker;
 void CCommandFuncs::Help(const char* szParameters)
 {
     CConsoleInterface* pConsole = g_pCore->GetConsole();
-
-    if (szParameters && atoi(szParameters) == 1)
-        CCore::GetSingleton().GetMultiplayer()->AllowWindowsCursorShowing(true);
-    else
-        CCore::GetSingleton().GetMultiplayer()->AllowWindowsCursorShowing(false);
-
```
qaisjp: These changes come from initial commit, seems reasonable to remove 👍

Client/core/CCommandFuncs.cpp
```diff
@@ -47,10 +41,6 @@ void CCommandFuncs::Exit(const char* szParameters)
 
 void CCommandFuncs::Ver(const char* szParameters)
 {
-    ShowCursor(TRUE);
-    HCURSOR hc = LoadCursor(NULL, IDC_ARROW);
-    SetCursor(hc);
-
```
qaisjp: These changes come from initial commit, seems reasonable to remove 👍

Client/core/CGUI.cpp
```diff
@@ -748,7 +742,8 @@ void CLocalGUI::UpdateCursor()
             CCore::GetSingleton ().GetGame ()->GetPad ()->Clear ();*/
 
             // Restore the mouse cursor to its old position
-            SetCursorPos(m_StoredMousePosition.x, m_StoredMousePosition.y);
+            if (m_StoredMousePosition.x != -1 && m_StoredMousePosition.y != -1)
+                SetCursorPos(m_StoredMousePosition.x, m_StoredMousePosition.y);
```
qaisjp, line 755: I wonder if this is what reverts the cursor to its original position when you toggle `showCursor` while ingame?

Client/core/CCommandFuncs.cpp
lopezloo: Merged in 0cdc4d2d53dd7f829767e1f790ea47b29fc1ff31.

Client/core/CCommandFuncs.cpp
lopezloo: Merged in 0cdc4d2d53dd7f829767e1f790ea47b29fc1ff31.

lopezloo:
I have tweaked it now to show system cursor and do not center cursor only if game is being launched in windowed mode or user has multiple monitors and full screen minimize is disabled. It basically ensures that clicking outside game window won't minimize the game.

Things which remains controversial is the removal of one check from `CLocalGUI::InputGoesToGUI()`. It was required to make this PR work. I haven't found any issues with that but there may be though.

Client/core/CGUI.cpp
lopezloo: Most likely.

lopezloo:
By the way, even in current Windows 10 update screen (fullscreen) cursor is visible and unlocked. And it feels great.

Dutchman101:
PR was reverted in b71b222 due to a bug ([#3622](https://github.com/multitheftauto/mtasa-blue/issues/3622)), it may be re-submitted in a fixed state.

PR #2864, opened 2023-01-03
Update mainmenu version image

Haxardous:
This PR updates the mainmenu's version image to 1.6
image

Would be great to update mainmenu's images too or even redesign them.


patrikjuvonen:
Thanks but we already got it in our [bump-version branch](https://github.com/multitheftauto/mtasa-blue/compare/master...bump-version) :P

PR #2870, opened 2023-01-15
Fix crash on disconnect

### LINKED ISSUE #2473: MTA crashes when streamed in object with custom model is deleted (0x000C4BB0)

TheNormalnij:
Reopened #1842
Fixes #2473

WTR:
1. load [resources](https://www.mediafire.com/file/ys140v56wqzef5m/nandoSampmaploader.zip/file)
2. start samp_maploader
3. wait
4. run in console `gotomap 3`
5. esc => disconnect


botder:
https://github.com/multitheftauto/mtasa-blue/blob/05a7a744eb710b3e4b4943ada7a71e92afcb06c5/Client/mods/deathmatch/logic/CResourceManager.cpp#L20-L27

Please add the change after `Remove(pResource)` instead of doing it in `CResource::~CResource`.

TheNormalnij:
@botder 

I think, there we can find some related issue later. At least resource with a lot of custom models can produce some bugs after restarting.

In optimization purpose i can make this call optional and make only one call on disconnect and always after stopping/restarting.

botder:
Fixed in 5df6d1f9e3c9bf5568150f206062ae4f276ac36b

Issue #2411, opened 2021-10-12
libcef.dll crash on wine with mtasa 1.5.9


Sommmmy:
**Describe the bug**

libcef.dll crash upon entering into options > web browser or entering some servers.

**To reproduce**

1. Go to settings > web browser
2. Wait a couple seconds (notice strange lag)
3. Game crash

**Expected behaviour**

For it not to crash

**Screenshots**

![image](https://user-images.githubusercontent.com/59640424/136882320-ea312383-aeaa-42f8-bdfc-bcc868b3dac8.png)

**Version**

client 1.5.9
Wine(Lutris) on linux
**Additional context**

kolezka:
happens also to me
![image](https://user-images.githubusercontent.com/24277846/148098731-d20d61b2-f213-4b29-bbe6-6e9df3748798.png)

jolty1:
Fix CEF section in link below worked for me

[Fix CEF](https://gist.github.com/qaisjp/778ed99ac8a8dcb93f797a28ba5f8303)

TheNormalnij:
This is old wine issue, was fully fixed once in wine 5.12, but latest wine is broken again.

kolezka:
> Fix CEF section in link below worked for me
> 
> [Fix CEF](https://gist.github.com/qaisjp/778ed99ac8a8dcb93f797a28ba5f8303)

Does not help for me :(

kolezka:
https://github.com/multitheftauto/mtasa-blue/issues/1000 workaround helps me

Lpsd:
I'm pretty sure the gist (Fix CEF) has the same fix listed as #1000 - it just creates a symbolic link instead of moving the files.

Did you successfully create the symbolic links? Did you receive a different error after this?

Zorono:
> I'm pretty sure the gist (Fix CEF) has the same fix listed as #1000 - it just creates a symbolic link instead of moving the files.
> 
> Did you successfully create the symbolic links? Did you receive a different error after this?

already tried any workaround possible to solve this situation but nothing made it work... once i try to launch MTA 1.5.9 under WINE 7 it complains about dxgi related errors and then crashes... while it says that GTA SA is already running and i have to terminate it...

[core.log](https://github.com/multitheftauto/mtasa-blue/files/8009479/core.log)
[logfile.txt](https://github.com/multitheftauto/mtasa-blue/files/8009480/logfile.txt)
[...]

Issue #1000, opened 2019-06-18
CEF on Wine


qaisjp:
**Describe the bug**

CEF does not work on Wine as expected. 

**To reproduce**

1. Run `start webbrowser`
2. Open the webbrowser.
3. Nothing appears in place of the MTA website

**Version**

Multi Theft Auto v1.5.6-release-18664

**Additional context**

You will receive the following warnings the Wine (Windows) console:
```
0177:err:module:import_dll Library libcef.dll (which is needed by L"C:\\Program Files\\MTA San Andreas 1.5\\MTA\\CEF\\CEFLauncher_DLL.dll") not found
0177:err:module:import_dll Library CEFLauncher_DLL.dll (which is needed by L"C:\\Program Files\\MTA San Andreas 1.5\\MTA\\CEF\\CEFLauncher.exe") not found
0177:err:module:LdrInitializeThunk Importing dlls for L"C:\\Program Files\\MTA San Andreas 1.5\\MTA\\CEF\\CEFLauncher.exe" failed, status c0000135
```

You must **COPY** the following files from `MTA` to `MTA/CEF` to work around this issue:

```bash
cp libcef.dll chrome_elf.dll icudtl.dat natives_blob.bin v8_context_snapshot.bin snapshot_blob.bin CEF
```

Thank you to @TheNormalnij for discovering this solution.

Note that you do _**not**_ need `snapshot_blob.bin` from the MTA directory if you set `WINARCH=win32`. This additional `bin` file is only required in x64.

Without the DLL files in the `MTA/` directory, the game WILL crash. This has been tested and is reproducible. You can't test this without performing a game restart.

Without the DLL files in the `CEF/` directory, the game will silently output an error to the Windows (Wine) console. Nothing will appear ingame. This can be tested by restarting the `webbrowser` resource.

Without the other `.bin` data files in the `CEF/` directory, CEFLauncher will crash, but the game will not. Nothing will appear in game. This can be tested by restarting the `webbrowser` resource. This also happens if you remove only the `icudtl.dat`. Partial removal of the `.bin` files have not been tested.

**Roadmap**

- [ ] Determine if this is a bug in MTA or a bug in Wine.
- If a bug in MTA: fix it.
- If a bug in Wine:
    - File a Wine report.
    - Add a fix into our installer by simply symlinking these important files.
[...]

PR #2875, opened 2023-01-17
Fix CEF crashing on wine

vahook:
See https://github.com/multitheftauto/mtasa-blue/issues/2411 (and also the original issue https://github.com/multitheftauto/mtasa-blue/issues/1000 )

Currently, `CEFLauncher_DLL.dll` is linked against `libcef.dll` **implicitly**. Therefore, `libcef.dll` must be available at the time of loading `CEFLauncher_DLL.dll`. However, contrary to what the implementation in `CEFLauncher_DLL` might suggest, this is achieved by calling `SetDllDirectory` **in the MTA process**, as on Windows `SetDllDirectory` also [affects the children](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-setdlldirectorya#:~:text=for%20win32%20processes%20that%20are%20not%20running%20a%20packaged%20or%20protected%20process%2C%20calling%20this%20function%20will%20also%20affect%20the%20dll%20search%20order%20of%20the%20children%20processes%20started%20from%20the%20process%20that%20has%20called%20the%20function.) (and [implicit linking](https://pretentiousname.com/setdlldirectory_inheritance/index.html)). Sadly, wine does not implement this at the moment, hence `CEFLauncher.exe` crashes.

This PR introduces [delay-loading](https://learn.microsoft.com/en-us/cpp/build/reference/delayload-delay-load-import?view=msvc-170), which is a possible solution to the issue. Alternatively, we could wait for a fix in wine.

Also, feedback from other wine players would be appreciated.


jason3855:
Could you describe how to do that? I don't quite know how to do this loading delay

Lpsd:
This is a pull request to resolve the issue in core code. If you're unsure what this means, then you don't need to worry. It _may_ be available as a fix at some point. Until then, you'll just need to wait @jason3855

jason3855:
Ok!

Pirulax:
We need someone to test this and it can be merged.
@jason3855 If you have a Wine installation to test, you can download it from [here](https://github.com/multitheftauto/mtasa-blue/suites/10407663185/artifacts/514868749)
It's not an official build, so you won't be able to connect to public servers, just a local one.

gamelaster:
@Pirulax I'm playing on Wine, so I will test it within today / tomorrow (I will delete the CEF files which are used as workaround right now).

gamelaster:
Hey @Pirulax , I tried the CI build, sadly, it seems there are some issues, some related to Wine, some seems to be build misconfiguration:
1) loader.dll built by CI can't find `WSCAPI.dll` and fails on it (release loader.dll works fine)
```
0104:err:module:import_dll Library WSCAPI.dll (which is needed by L"C:\\MTA-Test\\mta\\loader.dll") not found
0104:err:module:import_dll Library WSCAPI.dll (which is needed by L"C:\\MTA-Test\\mta\\loader.dll") not found
```
2) After using release loader.dll and ignoring integrity errors, local server is not working.
```
[2023-02-19 15:56:05] [Output] : Error: Could not start local server. [Could not load network library]
```
3) Running MTA Server.exe throws following error:
```
MTA:BLUE Server for MTA:SA

ERROR: 'net.dll' library version is '1.5.9-5.21530.0' (Expected '1.5.9-5.00000.0')

(If this is a custom build,
 1. Set MTASA_VERSION_BUILD in version.h to 21530
 2. Rebuild)


ERROR: Loading network library (net.dll) failed!
Press Q to shut down the server!
```

3) and probably 2) would be possible to fix by building it by myself and configuring it as error says, but 1), I'm clueless how is possible that it works on release, but not on CI build.

botder:
Loader shouldn't depend on wscapi.dll since https://github.com/multitheftauto/mtasa-blue/commit/44ac68b2931e0298b8616f389ba0d4099c746833

gamelaster:
With the help of @botder , the CI build is now working, and can confirm, that this PR works perfectly on my setup with wine8.1-staging in 32-bit bottle.
image

gamelaster:
@botder @Pirulax anything else needed to test before merge? Would be really nice to have this :blush:

botder (approved):

PR #2879, opened 2023-01-19
Fix crash after disconnect #2877

### LINKED ISSUE #2877: Crash after disconnect after using `engineSetModelPhysicalPropertiesGroup` on custom models
### Describe the bug

This crash only happens when using ``engineSetModelPhysicalPropertiesGroup`` on a custom model, and only after disconnecting from the server. It will not happen for default models, and it will not happen if the resource is stopped. It will also still happen after the function has been triggered, even if the model was unloaded.

You can use this resource to test this crash: 
[crash_property.zip](https://github.com/multitheftauto/mtasa-blue/files/10452301/crash_property.zip)

The offsets varies on the version, but the function responsible (according to Dutchman) is https://github.com/multitheftauto/mtasa-blue/blob/master/Client/game_sa/CModelInfoSA.cpp#L1733

Offsets are:

0x0003EB37 on r21519
0x0003EC30 on r21488
0x00003E73 on r21476

I didn't test any other versions, but this bug is probably there since the introduction of custom models for objects. No one really uses ``engineSetModelPhysicalPropertiesGroup`` that much.

### Steps to reproduce

1. Load an empty server (with play/freeroam).
2. Load the resource provided
3. Run the command /replaceCrash
4. Disconnect

You should crash.

### Version

Client: v1.5.9-release-21519
Server: v1.5.9-release-21480

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #2877

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1730,7 +1730,9 @@ void CModelInfoSA::RestoreAllObjectsPropertiesGroups()
 {
     for (const auto& pair : ms_OriginalObjectPropertiesGroups)
     {
-        pGame->GetModelInfo(pair.first)->GetInterface()->usDynamicIndex = pair.second;
+        CBaseModelInfoSAInterface* pInterface = pGame->GetModelInfo(pair.first, true)->GetInterface();
+        if (pInterface)
+            pInterface->usDynamicIndex = pair.second;
     }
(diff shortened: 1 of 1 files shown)


botder (approved):

patrikjuvonen (approved):

patrikjuvonen:
Thank you.

Issue #2489, opened 2021-12-27
A way to upload files to a server


PlatinMTA:
**Is your feature request related to a problem? Please describe.**
Currently the only way you can upload files from your MTA into a server is by copying the files to one of your resources in ``MTA/mods/deathmach/resources`` folder or via web.

MTA could have a way to let users upload files to the server. You could call a function like ``requestFile`` (both client-sided and server-sided) that open a GUI where you could search for any file on your system and upload it. 

![image](https://user-images.githubusercontent.com/23291573/147445035-68f631dd-d73b-4b8f-8329-88a242d8e6b0.png)
How CS:Source's UI looks for reference.

The window needs to work similar to how ``requestBrowserDomains`` works. You could disable the option to upload any files to servers from the Configuration, and the function could return some value so scripters can ask the client to change this configuration in order to upload the file (similar to ``takePlayerScreenshot``).

Servers will not be able to see the client file-tree, ever. Servers will not be able to ask for a file that is on X directory. The user should search for the file by himself. The option to upload files to a server should be disabled by default, and when changed it should pop-up a warning.

I really can't see any harm done to have this added, apart from bugs. Maybe I'm not seeing some important security issues a thing like this would bring.

**Describe the solution you'd like**
A function like ``requestFile`` to be added client-side and server-side.

Client-side should upload the data of the file into a variable (like how ``fileRead(file, fileGetSize(file))`` would) and let the resource manage it. Maybe a callback would be needed, but I don't think it will.

Server-side should do the same but on an event or callback. The server function should allow a way to select the size of the file aswell as the upload velocity, similar to ``takePlayerScreenshot``.

This functionality should come disabled by default and need to be enabled from Main Menu > Configuration. 

**Describe alternatives you've considered**
Copying the file into a temporary folder would also work wonders.

If this is too risky then the alternative would be to not have this at all.

**Additional context**
This would come in handy for uploading of images in particular. Models could also be used for customization, but this would be harder than managing some images.

The main focus should be security. And oversight could mean the world for malicious server owners that want to harm users. I guess this is why there is no such feature like this yet.

Fernando-A-Rocha:
This is really clever! I use the old fashioned file uploading way (player places file in downloaded resource folder and gives file name) on various systems, this would do wonders!!

lopezloo:
Such function could open modal to ask user if s/he actually want to open file explorer. Maybe even with optional tickbox like "Remember this choice for current server".

[...]

Issue #1953, opened 2020-12-29
Accessing File System and privacy prompts


prnxdev:
**Is your feature request related to a problem? Please describe.**
For now developers don't have an ability to access local files so if developer wants users allow to upload some files like avatars or maps they have to use some other methods like web page or application.

**Describe the solution you'd like**
I think there should be some API to access local files. I know that there is a lot about keeping everything secure etc. and I thought about it. For me good solution would be to give player solid information that script/resource/server wants to access files and do you allow it. It would look like in mobile apps where you get that type of question for every app you are installing or lunching. Example code:

```lua
--[[ 
it would prompt info I mentioned earlier 
or just return true/false if it was accepted earlier or not.
For security reason we could only allow to read files or just save files
using dialog mentioned later
]]
local permitted = resourceAskForPermission('file_system')

if permitted then
   --[[
   it would open files dialog
   and return all selected files.
   Each element of array would  be path to selected file or files
   and we could use File.open methods along with other File methods.
   It can return only one path (still in table) if first param is true 
   (so dialog should allow only one file to be selected)
   and only files with specific extension provided in second param
   ]]
   local files = selectLocalFiles(true, '.jpg')

   for i, file in ipairs(files) do
      file.open() -- it works
   end
end
```

And I thought further: we could use `resourceAskForPermission` not only for File System but also for example to access System Informations, microphone (I know that now there is voice chat but no one knows when mic is used etc.), devices connected to same network or to open URLs in browser (currently it is only allowed using CEF).

**Describe alternatives you've considered**
n/a

**Additional context**
It would add more flexibility for servers.

prnxdev:
Opening URLs in browser should be also always prompted with
```
This resource wants to open url: [and here we got this url]. Do you want to allow it?
```
There are also APIs to check if URL is safe so we could just add additional info that this URL is safe or not.

CrosRoad95:
```
local files = selectLocalFiles(true, '.jpg')

   for i, file in ipairs(files) do
      file.open() -- it works
   end
```
code like this clientside will cause mta to hang, you must use callback in this case

prnxdev:
Ok, true so that should look like this:
```lua
selectLocalFiles(true, '.jpg', function (filesPaths)
   for i, path in ipairs(filesPaths) do
      File.open(path) -- i forgot it is static method
   end
end)
```
or
```lua
local localFilesSelectionManager = selectLocalFiles(true, '.jpg')

[...]

PR #2887, opened 2023-02-04
Drag and drop files to import them into MTA

CrosRoad95:
This feature basicly adds a new event "onFilesDrop", it is triggered when you drag and drop file onto mta window from any location of your computer. Behind  the scene it using fileOpen, fileRead and fileClose to read content. Event gives you table of all droppped files [`fileName] = content`, fileName does not contain full path, just file name. You can't drop folders ( not implemented ).

```lua
local textures = {}
  for i,v in ipairs(textures)do
    dxDrawImage(200 + i * 128, 200, 128, 128, v)
  end
end)

addEventHandler("onFilesDrop", localPlayer, function(files)
  for fileName, data in pairs(files)do
    outputChatBox(string.format("Dropped file: %s of size: %i", fileName, string.len(data)))
    local extension = gettok (fileName, 2, "." )
    if(extension == "png" or extension == "jpg")then
      local texture = dxCreateTexture(data)
      if(texture)then
        textures[#textures + 1] = texture
      end
    elseif(extension == "txt")then
      outputChatBox("File content: "..tostring(data))
    end
  end
end)
```

how it works in action:
https://www.youtube.com/watch?v=5w5G3snKY1g

Related issues #1953 #2489


Fernando-A-Rocha:
What about Fullscreen users? Maybe an alternative way for them to pass files to the window? 🤔

Vadya963:
I think you can make it easier ctrl+c and ctrl+v, copied the picture, went into the game and inserted it and call the event onFilesClipboard

CrosRoad95:
> I think you can make it easier ctrl+c and ctrl+v, copied the picture, went into the game and inserted it and call the event onFilesClipboard

had idea to make something like that, clipboard where you get raw string, and second argument would be some kind of rich content of what you copied. For example if you copy paste files you get files names and content, if you copy html you get all tags ect

lopezloo:
Such thing could be useful for script/model/map development but it needs to be implemented carefully IMO.

My suggestions:
1. Add `onFilesDragStart` and `onFilesDragEnd` events which would be called when user starts and stops hovering files over game window. Could be used to draw appropriate UI.
`onFilesDragStart` parameters could have some basic information such as amount of files (and/or folders).

2. Implement creation of modal to confirm user choice asking if s/he really want to upload files (notify user that the dropped file contents and name will be read by the game and can be uploaded to remote server). Add an optional checkbox which allow to skip creation of next modals for current server session. Example:



3. Add informational client console message containing file paths after files being dropped.
Also send similar message if drop was blocked because user disallowed it.

Related issue: #2489

CrosRoad95:
> Implement creation of modal to confirm user choice ...

i will add it if you provide me at least one example from the past of people accidently uploading their files to someones website ( a news article, a website that bait people to upload their computer to their servers )

lopezloo:
> i will add it if you provide me at least one example from the past of people accidently uploading their files to someones website ( a news article, a website that bait people to upload their computer to their servers )

IMO browsers should create similar modal on file drop. I don't know how often such thing happen but it's definitely possible to accidently drop file to wrong window.

CrosRoad95:
i accidently can paste my password on public chat, and what? nothing. if you want to treat people as babies all the time, you can implement such dialog :)

PlatinMTA:
> > Implement creation of modal to confirm user choice ...
> 
> i will add it if you provide me at least one example from the past of people accidently uploading their files to someones website ( a news article, a website that bait people to upload their computer to their servers )

Considering this is similar to #2489, you should prompt the user anyway

Shuubaru:
This is awesome but I agree a confirmation prompt is needed. In that way you can avoid throwing files by accident (yes, it happens) and you can also avoid the client/server to process unwanted files, it's just a QoL feature.

CrosRoad95:
two security improvments i could introduce:
1. make impossible to drag and drop files from user documents and simillar locations 
2. add option in settings that enables all formats ( by default disabled ), by default allow text files, lua scripts, models, .xml files, html files so you can't drop your "cv.doc" by default or any other documents

PlatinMTA:
> 1. make impossible to drag and drop files from user documents and simillar locations

That seems rather unwanted, I would be annoyed if I had a file on my user documents I want to share but I can't because of this

> 2. add option in settings that enables all formats ( by default disabled ), by default allow text files, lua scripts, models, .xml files, html files so you can't drop your "cv.doc" by default or any other documents

This would be cool

>Implement creation of modal to confirm user choice asking if s/he really want to upload files (notify user that the dropped file contents and name will be read by the game and can be uploaded to remote server). Add an optional checkbox which allow to skip creation of next modals for current server session.

But why not this tho?

CrosRoad95:
image
don't want to overengineer it, added a dropdown in settings that let you specify what do you want, by default first option is selected: allow mta specific files only which means: allow formats: ".txt", ".lua", ".luac", ".txd", ".dff", ".col", ".xml", ".json", ".png", ".jpg", ".svg", ".map" ( probably i missed some extensions ).

PR #2891, opened 2023-02-07
Set `MTASA_VERSION_TYPE` back to `VERSION_TYPE_CUSTOM`

lopezloo:
Sets `MTASA_VERSION_TYPE` back to `VERSION_TYPE_CUSTOM`. It was changed in 19831ae4b19e7279575ae23ed9df720e1d695ff5. According to documentation `VERSION_TYPE_UNSTABLE` is meant for MTA forks. But by default we develop MTA, not MTA forks, right? Currently freshly built MTA server asks user to change version number which may be confusing for development.

### DIFF
--- a/Shared/sdk/version.h
+++ b/Shared/sdk/version.h
@@ -27,7 +27,7 @@
 #define MTASA_VERSION_MAJOR         1
 #define MTASA_VERSION_MINOR         5
 #define MTASA_VERSION_MAINTENANCE   9
-#define MTASA_VERSION_TYPE          VERSION_TYPE_UNSTABLE
+#define MTASA_VERSION_TYPE          VERSION_TYPE_CUSTOM
 #define MTASA_VERSION_BUILD         0
 
 #include "../build_overrides.h"


Dutchman101:
I've previously explained you why it is how it is. Relevant dev discord discussion: https://discord.com/channels/801330706252038164/801330706252038170/1057761559511842916

If what you say about server is correct, please open a PR to solve that instead. If i already won't do that on short term. Also, please link the documentation page you're referring to

You can put the answers in a new issue, or open a PR for the server change and reply about that documentation

lopezloo:
> Also, please link the documentation page you're referring to

I'm referring to this: https://github.com/multitheftauto/mtasa-blue/blob/af30d18e2df3216dcdeedaddff0bf01e726adc9d/Shared/sdk/version.h#L5-L25

> If what you say about server is correct, please open a PR to solve that instead.

This is what freshly compiled server (on master - af30d18e2df3216dcdeedaddff0bf01e726adc9d) console prints:
```
MTA:BLUE Server for MTA:SA

ERROR: 'net.dll' library version is '1.5.9-5.21612.0' (Expected '1.5.9-5.00000.0')

(If this is a custom build,
 1. Set MTASA_VERSION_BUILD in version.h to 21612
 2. Rebuild)


ERROR: Loading network library (net.dll) failed!
Press Q to shut down the server!
```

Well it basically tells user what to do so it could be automated.

Maybe `win-create-projects.bat` script could fetch that build number and automatically write to `Shared/build_overrides.h`:
```c++
#define MTASA_VERSION_BUILD {{NUMBER}}
```
And we could add `Shared/build_overrides.h` to `.gitignore` then.

Issue #2880, opened 2023-01-21
Add ability to get debug setting in dxGetStatus


ghost:
### Is your feature request related to a problem? Please describe.

As the title says. Currently this is achievable in hacky workaround given below, but i believe it should be natively supported by using aforementioned function. For people who are not aware of what it does, it allows you to completely disable replaced textures (# 6734 Graphics to be clear)

### Describe the solution you'd like

Simple, efficient and not hacky one. Add it as **SettingDebugMode** which would be returned by [dxGetStatus](https://wiki.multitheftauto.com/wiki/DxGetStatus)

### Describe alternatives you've considered

Ugly, hacky workaround:

```lua
if #engineGetVisibleTextureNames() < 1 then
	outputChatBox("graphics debug")
end
```

### Additional context

![obraz](https://user-images.githubusercontent.com/51768772/213881370-ede690b2-9069-4539-a83f-305643f3c594.png)

(This setting is available in advanced tab)

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

Issue #2650, opened 2022-06-25
Merge isVehicleOnGround and isPedOnGround into one function "isElementOnGround"


CrosRoad95:
### Is your feature request related to a problem? Please describe.

not really, just for api quality reason

### Describe the solution you'd like

merge isVehicleOnGround and isPedOnGround into one function "isElementOnGround"

### Describe alternatives you've considered

/

### Additional context

if you have function suitable for more than one type of element, it should have name beginnin with "isElement ""setElement" ect

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

PR #2900, opened 2023-02-16
Add ability to get debug setting in dxGetStatus and merge isVehicleOnGround and isPedOnGround into one function "isElementOnGround"

vyn666:
Referred to issue #2880 (https://github.com/multitheftauto/mtasa-blue/issues/2880)


vyn666:
Merge isVehicleOnGround and isPedOnGround into one function "isElementOnGround" referred to issue #2650 (https://github.com/multitheftauto/mtasa-blue/issues/2650)

Allerek:
This might be confusing for elements like objects and many more in my opinion.

vyn666:
At this point there is no way to check if, for example, the object is on the ground, to prevent confusing in my opinion could add a note on wikipedia that this function only supports elements of the type ped & vehicle.

ghost:
> This might be confusing for elements like objects and many more in my opinion.

I disagree. We have some functions called setElement*/getElement* (to name a few - attachElements, setElementAlpha) which do not support each of available element types. This is mentioned on wiki, and you can conclude this from function return itself (e.g it returns false in case of failure). Unified function sounds much better, than having to keep (use) different methods for each element.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1551,6 +1554,29 @@ int CLuaElementDefs::IsElementFrozen(lua_State* luaVM)
...
+    argStream.ReadUserData(pEntity);
+
+    if (!argStream.HasErrors())
+    {
+        bool bOnGround;
+        if (CStaticFunctionDefinitions::IsElementOnGround(*pEntity, bOnGround))
+        {
+            lua_pushboolean(luaVM, bOnGround);
+            return 1;
+        }
```
CrosRoad95: first of all, use new parser, second: if "isElementOnGround" return false -> throw ""error, element type not supported""

Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
```diff
@@ -1728,6 +1728,43 @@ int CLuaDrawingDefs::DxGetStatus(lua_State* luaVM)
...
+            case EDiagnosticDebug::RESIZE_NEVER_0000:
+                diagnosticDebugString = "#0000 Resize never";
+                break;
+            default:
+                diagnosticDebugString = "Default";
+                break;
+        }
+
+        lua_pushstring(luaVM, "SettingDebugMode");
+        lua_pushstring(luaVM, diagnosticDebugString.c_str());
```
Pirulax:
```suggestion
        lua::Push(luaVM, "SettingDebugMode");
        lua::Push(luaVM, []{
            switch (g_pCore->GetDiagnosticDebug())
            {
            case EDiagnosticDebug::GRAPHICS_6734:
                return "#6734 Graphics";
            case EDiagnosticDebug::D3D_6732:
                return "#6732 D3D";
            case EDiagnosticDebug::LOG_TIMING_0000:
                return "#0000 Log timing";
            case EDiagnosticDebug::JOYSTICK_0000:
                return "#0000 Joystick";
            case EDiagnosticDebug::LUA_TRACE_0000:
                return "#0000 Lua trace";
            case EDiagnosticDebug::RESIZE_ALWAYS_0000:
                return "#0000 Resize always";
            case EDiagnosticDebug::RESIZE_NEVER_0000:
                return "#0000 Resize never";
            default:
                return "Default";
            }
        }());
```

Also, side-note, instead of default maybe we should return `None` or something similar.

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -1551,6 +1554,29 @@ int CLuaElementDefs::IsElementFrozen(lua_State* luaVM)
     return 1;
 }
 
+int CLuaElementDefs::IsElementOnGround(lua_State* luaVM)
```
Pirulax: Replace this with:
```cpp
bool CLuaElementDefs:::IsElementOnGround(CEntity* entity) {
    switch (Entity.GetType())
    {
    case CCLIENTPLAYER:
    case CCLIENTPED:
        return static_cast<CClientPed*>(entity)->IsOnGround();
    case CCLIENTVEHICLE:
        return static_cast<CClientVehicle*>(entity)->IsOnGround();
    default:
        throw std::invalid_argument{"Element type not supported"}; // TODO: Add `InvalidElementType{entity}` or something similar for the parser (so the exact type can be reported)
    }
}
```
And get rid of the function in `CStaticFunctionDefinitions`

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.h
```diff
@@ -66,6 +66,7 @@ class CLuaElementDefs : public CLuaDefs
     LUA_DECLARE(IsElementDoubleSided);
     LUA_DECLARE(GetElementCollisionsEnabled);
     LUA_DECLARE(IsElementFrozen);
+    LUA_DECLARE(IsElementOnGround);
```
Pirulax: You'll have to forward declare the function instead: `bool IsElementOnGround(CEntity* entity);`

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -64,6 +64,7 @@ void CLuaElementDefs::LoadFunctions()
         {"isElementDoubleSided", IsElementDoubleSided},
         {"getElementCollisionsEnabled", GetElementCollisionsEnabled},
         {"isElementFrozen", IsElementFrozen},
+        {"isElementOnGround", IsElementOnGround},
```
Pirulax: Use `ArgumentParser<IsElementOnGround>` here

Pirulax (commented):
Also, we don't remove old functions (Backwards compatibility).
I see that's what you did.
Not sure if others agree, but you could just use `CLuaElementDefs:::IsElementOnGround` for `isPedOnGround` and `isVehicleOnGround` instead of the old functions. Yes, it is hacky, but it will work just fine (Though, old functions only accepted peds/vehicles respectively, so... might not be ideal)

vyn666:
Finally all updated.

vyn666:
Updated `CLuaElementDefs.cpp` and `CLuaElementDefs.h` (changed **CEntity** to **CClientEntity**, because it was showing errors).

Lpsd:
Please consider submitting multiple pull requests for unrelated changes. In this case, one PR for dxGetStatus, and another for the isElementOnGround addition.

Also, you can consider marking `isVehicleOnGround` and `isPedOnGround` as deprecated by defining them here: https://github.com/multitheftauto/mtasa-blue/blob/master/Client/mods/deathmatch/logic/luadefs/CLuaCompatibilityDefs.cpp#L16

PR #2914, opened 2023-03-03
Correct Visual Studio studio version in Readme

lopezloo:
Corrects Visual Studio studio version in Readme.


patrikjuvonen:
It appears to me that the 2017 is intended. The plugin adds support for .editorconfig files, which is only necessary for old Visual Studio installs.

A better choice would be to remove the stale line now, since most people have likely upgraded their Visual Studio anyway to native editorconfig support.

patrikjuvonen (approved):

PR #2915, opened 2023-03-03
Tweak Readme header

lopezloo:
Drops `San Andreas` from project header in Readme and enlarges it. `Multi Theft Auto` alone branding looks better in my opinion. This PR is up to discussion (and is super freaky minor change but why not). MTA in many places doesn't use `San Andreas` anymore (home page and main menu for example has big centered `Multi Theft Auto` text).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Multi Theft Auto: San Andreas
+# Multi Theft Auto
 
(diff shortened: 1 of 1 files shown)


Fernando-A-Rocha:
-1
Repo name is gonna stay "mtasa" so this is one reason to not remove "San Andreas" from the readme.
Also, having "San Andreas" is better for finding the project via search engine (SEO) as it's mainly related to GTA: San Andreas.

MTA as a whole is Multi Theft Auto, sure, but this is one specific SA project that turns out to be the main one, imo.

lopezloo:
> Repo name is gonna stay "mtasa" so this is one reason to not remove "San Andreas" from the readme.
> Also, having "San Andreas" is better for finding the project via search engine (SEO) as it's mainly related to GTA: San Andreas.

Well it also says blue. It's like codename, right?

I think it's valid point but I like "Multi Theft Auto" branding though. There are no other active projects under this brand (like MTA:VC) so there won't be possibility to confuse them. And MTA is now an engine basically. You don't really need MTA:VC because you can load Vice City map into MTA. Personally I would rename it to "Multi Theft Auto" everywhere. It's minor change but after all I like minor changes.

And currently there is incosistency in naming. For example MTA executable is named "Multi Theft Auto" while application window title is "MTA: San Andreas".

For SEO, I think text like "Multi Theft Auto is multiplayer mod for Grand Theft Auto: San Andreas" would be good enough.

patrikjuvonen:
I think we should open a discussion or issue about this instead of a pull request, since pull requests are ultimately meant to resolve an issue, not act as an issue. I personally disagree with this change, I don't see the reason to change it. It is however worth creating an issue to address the inconsistencies across the board and see what the ultimate solution for this is. Thanks.

Issue #2906, opened 2023-02-21
warpPedIntoVehicle causing C++ runtime assertion failure


Dark-Dragon:
### Describe the bug

Some obscure, mostly trailer-like vehicles that can technically have a driver currently misbehave when you attempt to warp someone inside on a passenger seat using warpPedIntoVehicle, causing network trouble for the player at best and a Microsoft Visual C++ Runtime assertion failure on the server at worst (see below).

Example Vehicle IDs: 610, 611, and a couple more

### Steps to reproduce

```
run veh = createVehicle(611,0,0,3) plyr = getRandomPlayer() warpPedIntoVehicle(plyr,veh,0) -- works as intended, contrary to...

run veh = createVehicle(611,0,0,3) plyr = getRandomPlayer() warpPedIntoVehicle(plyr,veh,1) -- warpPedIntoVehicle will falsely return true but warp the player on top of the vehicle instead, causing network trouble after a while

run veh = createVehicle(611,0,0,3) plyr = getRandomPlayer() warpPedIntoVehicle(plyr,veh,10) -- to cause the assertion failure
```

### Version

MTA:SA Server v1.5.9-release-21554 (Windows 10, 64 bit)

### Additional context

Possibly related: #2907 getVehicleMaxPassengers returns 'false' for those instead of the expected 0

### Relevant log output

```shell
---------------------------
Microsoft Visual C++ Runtime Library
---------------------------
Assertion failed!

Program: ... Andreas 1.5\server\mods\deathmatch\deathmatch.dll
File: C:\BuildAgent\work\675e5b8e8f135823...\SharedU....Misc.h
Line: 1571

Expression: uiIndex < SIZE

For information on how your program can cause an assertion
failure, see the Visual C++ documentation on asserts

(Press Retry to debug the application - JIT must be enabled)
---------------------------
Abbrechen   Wiederholen   Ignorieren   
---------------------------
```

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

ghost:
I don't think it's related to certain vehicle models. I remember having network trouble when calling warpPedIntoVehicle right after vehicle creation, if you delay warp (by setTimer) - it should be fine, hacky - but working temporary fix.

Dark-Dragon:
[...]

PR #2918, opened 2023-03-06
Broken PR - (Fixed issue #2906)

TracerDS:

* [X] [your code follows the coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines)
* [X] your commit messages are informative ("update file.cpp" is unhelpful. "fix missing model in getVehicleNameFromModel" is helpful)

Description:
- When `GetMaxPassengers()` is 255, we know its an error (`VEHICLE_PASSENGERS_UNDEFINED`; trailers shouldn't be driveable anyway)
- If `GetMaxPassengers()` is 255 (`VEHICLE_PASSENGERS_UNDEFINED`) then we can return from that function to not cause any issues
- Also changed some if-hells to more suitable return-early
- Also added safe casting to some arguments


CrosRoad95:
you probably turn on some formatting, have white spaces at the end of all lines ( may be because you created pr by uploading file what broke it )

TracerDS:
> you probably turn on some formatting, have white spaces at the end of all lines ( may be because you created pr by uploading file what broke it )

Oops, sorry. First time using PR honestly

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -4136,6 +4136,9 @@ bool CStaticFunctionDefinitions::WarpPedIntoVehicle(CPed* pPed, CVehicle* pVehic
     assert(pPed);
     assert(pVehicle);
 
+    if(pVehicle->GetMaxPassengers() == VEHICLE_PASSENGERS_UNDEFINED)
```
TracerDS, line 4139: `VEHICLE_PASSENGERS_UNDEFINED` = 255;
`GetMaxPassengers()` returns 255 if the seat amount is not defined **or** if the vehicle is not drive-able (like trailers)

Issue #2923, opened 2023-03-12
Make it possible for a player to permanently be the syncer of a vehicle/ped


MegadreamsBE:
### Is your feature request related to a problem? Please describe.

I'm trying to implement a element pool system for my Sphene project. Basically the server generates a specific amount of instances of vehicles & peds that are available client-side before the SCM runtime needs them. Then instead of asking the server to create those elements when needed (and temporarily keeping track of the instructions SCM has run in order to not pause the runtime continuously) it can bring them over to the right location and do whatever it needs to do immediately. The current method causes these elements to pop in later than they should and causes all sorts of weird behavior we need to correct due to actions being delayed etc. 

However, as the server is swapping out the syncers for these elements it still requires a lot more calls to the server in order to ensure that the player can actually enter the vehicle after having teleported it (otherwise it'll teleport back to the last location). Once again causing all sorts of unintended behaviors that are hard to correct.

### Describe the solution you'd like

I'd like to have the possibility of permanently assigning a specific player to be the syncer of a vehicle/ped no matter the distance and for the server to not re-assign it. Basically I want full control over who syncs which elements.

### Describe alternatives you've considered

Sending event calls to the server to teleport the vehicle or modify the syncer after moving the vehicle closer client-side (this works often but not always). However I just can't accept the delays of making these server calls in my project.

Another solution is further improving the already very complex logic I have wrapped around these elements in order to correct any issues that arise and accepting the pop in issue but this is far from ideal.

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

MegadreamsBE:
This was merged in so I'll close this issue.

PR #2925, opened 2023-03-12
Add persist parameter to setElementSyncer

MegadreamsBE:
Related to issue #2923 

Adds a parameter "persist" to setElementSyncer which causes the selected player to always be the syncer of the vehicle/ped/object until the player either leaves the server or setElementSyncer is called again without persist. This parameter is only available if a player is passed and not when passing true/false as the first parameter.

This gives servers more finer control over who syncs which vehicle/ped/object if so desired.


MegadreamsBE:
[syncer.zip](https://github.com/multitheftauto/mtasa-blue/files/10951457/syncer.zip)

Attached is a resource I used to test this functionality. There are multiple commands synctest1 to synctest7 testing different scenario's. When a player leaves the syncer is reset and default behavior continues. Elements with persistent syncers do not fall through the map even if they are far out of view.

CrosRoad95:
what if persistant syncer left server? what i i'm only syncer ( persistant syncer ) and i left the game?

MegadreamsBE:
> what if persistant syncer left server? what i i'm only syncer ( persistant syncer ) and i left the game?

It'll go back to the default behavior of dynamically figuring out who should be the syncer. Thankfully all I needed to do was disable persist in StopSync as this is already called when a player leaves. When you're the only syncer and leave the game then it'll stop sycning these elements (which is also the current behavior) until someone suitable is in the server

MegadreamsBE:
I did some testing with a second player, here's the results:

- When the syncer is far away the vehicle will not fall through the map or act weird.
- If the syncer from a far distance moves the vehicle into the air (but not close enough to be streaming it) it will fall down on the other players screen but sync back into place once the vehicle is in streaming distance. - This is completely fine
- If the second player touches or enters the vehicle they will become the syncer breaking the persistence. This behavior can easily be changed with some if-statements BUT this behavior is fine as the second player is actively interacting with the vehicle. This can be up for discussion but is fine as-is in my opinion.
- Peds cannot be damaged by the second player unless the syncer is in streamable range.

Nothing seems to be out of the ordinary as everything is in line with the purpose of this persist boolean based on my testing.

MegadreamsBE:
I've added extra checks that ensure that in the scenarios outlined in my previous comment in which the syncer does modify the server will no longer do so. Do note that the player actively driving the vehicle will actually be in charge of syncing but the player set by setElementSyncer will resume doing so after this player leaves.

CrosRoad95:
how about an option to switch to "manual" synchronization? so when change syncer i will never be changed by mta itself?

MegadreamsBE:
In addition to persist you mean? Since that's what this persist parameter already does.

CrosRoad95:
`setElementSynchronizationMode(element, "auto"/"manual")`

MegadreamsBE:
> `setElementSynchronizationMode(element, "auto"/"manual")`

I mean, we certainly could add a completely separate function to control it but I personally do prefer just having a boolean as part of setElementSyncer (as it is a cleaner solution and directly related to setElementSyncer) to control the behavior. But this is something that can be up for discussion.

botder (approved):

PR #2934, opened 2023-03-25
Fix #2917: Fix Get/Set vehicle model wheel size memory leak

### LINKED ISSUE #2917: Memory leak. Vehicle models are loaded into memory forever
### Describe the bug

If you get/set wheels size of a vehicle that is not loaded into memory, then this model will be forcibly loaded into memory forever
image

Affected functions:
```
CModelInfoSA::GetVehicleWheelSize
CModelInfoSA::SetVehicleWheelSize
CModelInfoSA::GetVehicleDummyDefaultPosition
CModelInfoSA::GetVehicleDummyPosition
CModelInfoSA::SetVehicleDummyPosition
```

### Steps to reproduce

```
local veh = createVehicle(any_model, any_position)
veh.dimension = 666
addVehicleUpgrade(veh, 1083)
```

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #2917

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1020,11 +1020,14 @@ CVector CModelInfoSA::GetVehicleDummyDefaultPosition(eVehicleDummies eDummy)
         }
     }
 
-    if (!IsLoaded())
-        Request(BLOCKING, "GetVehicleDummyDefaultPosition");
+    ModelAddRef(BLOCKING, "GetVehicleDummyDefaultPosition");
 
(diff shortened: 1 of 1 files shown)


AlexTMjugador:
Thanks for submitting this PR! :heart:

Do you know whether just calling `GetInterface` here is correct? In other words, under what circumstances can `ppModelInfo` hold null model info pointers, and how is MTA affected by them?

TheNormalnij:
I think `ppModelInfo` can have nullptr only during MTA loading. Maybe it's wrong DI in CModelInfoSA and it should be always `GetInterface` without m_pInterface member.

PR #2942, opened 2023-04-03
Function as argument

MobinYengejehi:
hi guys
i could made a method that help lua to pass a function as argument and send it to other resources
and you can use it in both server and client sides
you can pass function as argument in `triggerEvent` or `exports.resource:func_name` or `setElementData(if you set 4th argument to false)` and send it to another resource and call it
and also you can send functions in table (but you can't share metamethods)
for example:
```lua
-- resource : A
local player = nil;

local library = {};

function library.SetPlayer(element)
    if isElement(element) then
        player = element;
    end
end

function library.GetPlayer()
    return player;
end

addEvent("getlibrary", false);
addEventHandler("getlibrary", localPlayer, function(send)
    iprint("function is:", send, type(send)); -- type(send) = table
    send(library);
    print("library sent to ", getResourceName(send.resource));
    send:free(); -- cleanup function reference at resource B (this cleans up the memory. if you don't call this method it stays in memory untill program destroy lua stack like stopping resource)
end);
```
```lua
-- resource : B
local library;

triggerEvent("getlibrary", getLocalPlayer(), function(resourceALibrary)
    library = resourceALibrary;
    print("resource B got the library!");
end);

if library then
    print("player is : ", library.GetPlayer()); -- nil
    library.SetPlayer(getLocalPlayer());
    print("player is : ", library.GetPlayer()); -- local player
end
```
when you pass a function as an argument it will look like a table
and has these values:
-- function reference:
    .resource : the resource that owns function
    .reference : the reference id
    .free : this method deletes the reference. if you call it, means that you won't be able to use that function anymore

note 1 : you can store a function reference in other resource if you want to let it be active otherwise if you call that function in render event or sth like that and you don't use `:free` method, after seconds your game or server console crashs because it occupies your memory
note 2 : remember when you want to cleanup your function call it as method, like `function:free` (not like `function.free`) because `free` needs a callable value as first argument


CrosRoad95:
couldn't you just implicit convert functions to refs using `ref` and `deref`

MobinYengejehi:
> couldn't you just implicit convert functions to refs using `ref` and `deref`

i think i couldn't understand well what you mean
but if you mean why those are table
its because it doesn't cause confusion when passing a function as an argument in triggerEvent and other functions and you can check in other resources that where does this function come from

salwador:
No news about this PR? I can't imagine a cases for what need store functions in elementData, but for exports it very useful. 🤔

Fernando-A-Rocha:
This reminds me of #3551

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
```diff
@@ -446,6 +517,29 @@ void CLuaArgument::Push(lua_State* luaVM, CFastHashMap<CLuaArguments*, int>* pKn
...
                 break;
             }
+
+            case LUA_TFUNCTION:
+            {
+                if (m_pResource)
+                {
+                    lua_newtable(luaVM);            // create callable
+
+                    lua_pushresource(luaVM, m_pResource);            // push function inside callable
```
tederis, line 527: When a referenced resource become stopped this pointer will turn into the dangling one. From what I was able to see there is no an invalidation mechanism for functions.

tederis:
I consider this feature as excessive and the implementation as dangerous.

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
MobinYengejehi: i agree

MobinYengejehi: i have some cleaner and better version in my mind for this
i will work on it and update its PR to use that instead

MobinYengejehi:
i need time to refactor this PR

MegadreamsBE:
I'm opposed to this. There's no good reason for this to exist, it can be dangerous AND it can cause easy memory leaks.

FileEX:
I think that’s a **good** idea.
With more complex resources, it would often be useful to send something like a callback instead of having to trigger, for example.

~~Is it safe?~~
~~Yes and no — it depends on how it’s used. We can’t halt project development because of fear or poor script management by the community. We should, however, provide them with tools to manage such a feature, like sending functions.
It is indeed dangerous in the sense that if a potential cheater finds a trigger that accepts a function as an argument and, worse, executes it without checking, they could run arbitrary code on the server. But does that immediately disqualify this PR? Absolutely **not**.~~

~~It’s ultimately the scripters’ responsibility to secure their server. We can help them, but we can’t do it for them.~~

~~Imo, the ability to send functions via a trigger or export should have a toggle in mtaserver.conf - but only for the client-to-server direction.
Meaning: the server can always send functions to the client, but the reverse is only possible if the option is enabled in mtaserver.conf. If disabled, the server ignores the function and shows an appropriate warning as it does now.~~

As for element data, I’m not sure if it has any real use here, but if the server ignores a value that’s a function during synchronization (since a cheater could force this even with the 4th sync argument set to true), then that’s fine.

Rilot06:
> It is indeed dangerous in the sense that if a potential cheater finds a trigger that accepts a function as an argument and, worse, executes it without checking, they could run arbitrary code on the server.
> 
> but only for the client-to-server direction.
> Meaning: the server can always send functions to the client, but the reverse is only possible if the option is enabled in mtaserver.conf. If disabled, the server ignores the function and shows an appropriate warning as it does now.

I think you really misunderstand this PR. How would a client send a function's reference to the server or vice versa? That's not how it works. This is to send functions between resources on the same script side with either local events or exports.

FileEX:
> > It is indeed dangerous in the sense that if a potential cheater finds a trigger that accepts a function as an argument and, worse, executes it without checking, they could run arbitrary code on the server.
> > but only for the client-to-server direction.
> > Meaning: the server can always send functions to the client, but the reverse is only possible if the option is enabled in mtaserver.conf. If disabled, the server ignores the function and shows an appropriate warning as it does now.
> 
> I think you really misunderstand this PR. How would a client send a function's reference to the server or vice versa? That's not how it works. This is to send functions between resources on the same script side with either local events or exports.

I don’t know why I kept thinking it was about sending a function between sides, something like a string and reading it through loadstring, lol. Maybe it’s just after work and the effect of being sleep-deprived.

But still... I think a properly refined solution is acceptable. Currently, cheaters can execute any code on the client side, so I don’t think sending a function from resource X to resource Y would create any huge threat here. On the client side, nothing is and never will be truly secure. A cheater can do whatever they want, so sending their own function to some resource doesn’t really give them any additional capability to cause harm.

Rilot06:
> But still... I think a properly refined solution is acceptable. Currently, cheaters can execute any code on the client side, so I don’t think sending a function from resource X to resource Y would create any huge threat here. On the client side, nothing is and never will be truly secure. A cheater can do whatever they want, so sending their own function to some resource doesn’t really give them any additional capability to cause harm.

I don't think megadreams meant the "dangerous" part as in cheaters. More like easy memory leaks, segfaults, etc. But I think if you are careful enough, it can be implemented in a safe way

PR #2964, opened 2023-04-12
Addendum to 1c00ef9b - fix crash happening on debug builds

lopezloo:
Fixes crash happening on debug builds introduced in 1c00ef9b13dae5b8b23abcb94b236790252ecd42 (#2746).


patrikjuvonen (approved):
Thank you!

PR #2976, opened 2023-04-23
dxCreateRenderTarget custom surface format

CrosRoad95:
This pull request adds overload for dxCreateRenderTarget function:
```lua
element dxCreateRenderTarget( int sizeX, int sizeY, SurfaceFormat surfaceFormat )
```
where surface format is one of https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dformat value


Einheit-101:
For backwards compatibility, what format is the original "withAlpha"? I guess A8R8G8B8? And default is R8G8B8?

tederis:
It also should be taken into account that most of these D3D surface types will not be supported by the decent part of players' hardware. And it's important to have an ability to check whether a concrete type is available on client's GPU.

CrosRoad95:
true/false still work as worked, so no change here, i assume if you attempt to create surface not supported by player, this function just return "false"

tederis:
> true/false still work as worked, so no change here, i assume if you attempt to create surface not supported by player, this function just return "false"

Are you sure that checking resource availability by creating a resource itself is a good strategy? Because texture creation is a driver specific task and it's quite common when texture creation failure leads to crashes on certain configurations.

CrosRoad95:
probably i should also add a condition that check if format is valid, https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3d9-checkdeviceformat and maybe a helper function that checks if surface format is supported on player hardware `isSurfaceFormatSupporter(surface-format format)`

tederis:
> probably i should also add a condition that check if format is valid, https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3d9-checkdeviceformat and maybe a helper function that checks if surface format is supported on player hardware `isSurfaceFormatSupporter(surface-format format)`

Yes, it would be great to have a function for checking format availability

Shared/sdk/version.h
botder, line 1: This shouldn't be here.

Shared/sdk/version.h
```diff
@@ -28,8 +28,8 @@
 #define MTASA_VERSION_MAJOR         1
 #define MTASA_VERSION_MINOR         6
 #define MTASA_VERSION_MAINTENANCE   0
-#define MTASA_VERSION_TYPE          VERSION_TYPE_UNSTABLE
-#define MTASA_VERSION_BUILD         0
+#define MTASA_VERSION_TYPE          VERSION_TYPE_CUSTOM
+#define MTASA_VERSION_BUILD         21765
```
botder: Don't change this file.

botder (commented):
Pull request looks good, but what use case did you have in mind? When would a scripter need a different surface format?

CrosRoad95:
for example shaders like this that adds shadows would benefit from using float textures
image

tederis:
At least shadows and HDR require specific texture formats. Without the PR most of the modern effects cannot be implemented without resorting to different tricks reducing the overall effectiveness and quality.

botder (approved):

samr46:
It's already merged. But I'm wondering why you created new enum for this instead of extending existing [ERenderFormat](https://github.com/multitheftauto/mtasa-blue/blob/a19f06ef8b35a3152e6a4fe91d4f8d7ccea458d0/Client/sdk/core/CRenderItemManagerInterface.h#L46). ERenderFormat is [D3DFORMAT](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dformat).

tederis:
There can be also crashes on some hardware because of not checking texture format availability as I mentioned above. Unfortunately this was ignored.

botder:
@CrosRoad95 Can you tackle the issues above?

CrosRoad95:
from what i know, "ERenderFormat" is different than "D3DFORMAT", also: directx9 is almost 20 years old, how old computer you must have to do not support all added formats? you have to have windows10/11 to play mta anyway, windows probably already require from you to have non-potato pc

samr46:
> from what i know, "ERenderFormat" is different than "D3DFORMAT", also: directx9 is almost 20 years old, how old computer you must have to do not support all added formats? you have to have windows10/11 to play mta anyway, windows probably already require from you to have non-potato pc

ERenderFormat is used in dxCreateTexture and all related code as surface format (D3DFORMAT). Extending it and using it for render targets would be consistent with existing codebase. And dxCreateTexture would benefit from new formats as well (although I think it may be already possible to load prepared dds file in any format by using "unknown" as format).

PR #2980, opened 2023-04-25
Update CVehicleDefaultColors.cpp

Lord-Henry:
The vehicle **"Street Clean Trailer (611)"** should use the same colors from its truck **"Utility Van (552)"** instead of being always fully black.

Also, there's an old pull request that changes another file, fixing the same issue.
[pull/2432](https://github.com/multitheftauto/mtasa-blue/pull/2432)


CrosRoad95 (commented):
i trust you

botder (approved):

PR #2982, opened 2023-04-27
Remove not needed includes in server side "Deathmatch" project

CrosRoad95:
As in title, i went through most of server side deathmatch project and removed what possible. refactor is minimal just to decrease amount of code, includes, not needed includes.


tederis:
Headers like CSphere.h or CBox.h are ubiquitous and I think it would be better to put them into PCH.

CrosRoad95:
box is used in ~3 files, sphere in ~10, meanwhile CVector is used everywhere

Server/mods/deathmatch/logic/CResourceMapItem.h
```diff
@@ -10,8 +10,6 @@
  *****************************************************************************/
 #pragma once
 
-#include "CResourceFile.h"
-
```
botder, line 14: The class in this file inherits from that base class. Why do you want to remove the include for it here?

Server/mods/deathmatch/logic/packets/CPedWastedPacket.h
```diff
@@ -12,7 +12,6 @@
 #pragma once
 
 #include "CPacket.h"
-#include <CVector.h>
```
botder, line 15: Are you removing the include because it's already in the precompiled header?

CrosRoad95: yes

Server/mods/deathmatch/logic/CResourceMapItem.h
CrosRoad95: it compiles, i suspect it is included in different file

botder: IMHO a header must include everything it needs to "stand on its own". A precompiled header is just a tool to help with compilation times and, again my own opinion, headers should not be written in a style where such precompiled header exists.

CrosRoad95: i didn't say it coming from stdinc.h

botder: Yes, but that's not the problem I see with it.

CrosRoad95: what's the problem? how can i solve it?

PR #2985, opened 2023-04-29
Remove unused CClientVehicle variables

tederis:
Removes unused CClientVehicle member variables that obviously do nothing but spending memory(120 bytes) and CPU time by calculating a wrong speed.


CrosRoad95 (commented):
looking good :)

tederis:
> looking good :)

Anything that removes a junk is looking good. Thanks for your thorough review.

botder (approved):

PR #276, opened 2018-07-28
New world functions

samr46:
Fixes #385 
Related Mantis issues at https://bugs.mtasa.com/view.php?id=8113 (see relations)

Added new property for [setWorldSpecialPropertyEnabled](https://wiki.multitheftauto.com/wiki/SetWorldSpecialPropertyEnabled):

"**watercreatures**" toggle randomly created water creatures (default: true) ([picture 1](https://i.imgur.com/v3PYmIF.jpg))

Added new function "**addBirds**" ([picture 2](https://i.imgur.com/g37GgfR.png)).

> Syntax: addBirds ( float startX, float startY, float startZ, float destX, float destY, float destZ, int count, int birdType, bool checkObstacles )

count: 1 - 6
birdType: 0 - 2

Returns the number of created birds if the function is successful, false otherwise.

Note: GTA:SA can render only 6 birds at the same time. You can use removeBirds before calling addBirds if you need to create 6 birds in one call.

Added new function "**removeBirds**" which removes all existing birds in game. It can be useful with **addBirds** function.

> Syntax: removeBirds ( )

Returns true if the function is successful, false otherwise.

Changed behavior of [setBirdsEnabled](https://wiki.multitheftauto.com/wiki/SetBirdsEnabled) to allow addBirds create new birds even if it's disabled.

---

Added new functions to control default **time cycle** (24 weathers, 8 cycles).

> Syntax: setTimeCycle ( string property, [float value = -1], [int weatherID = -1], [int cycle = -1] )

property:
- AmbientRed
- AmbientGreen
- AmbientBlue
- AmbientObjectRed
- AmbientObjectGreen
- AmbientObjectBlue
- SkyTopRed
- SkyTopGreen
- SkyTopBlue
- SkyBottomRed
- SkyBottomGreen
- SkyBottomBlue
- SunCoreRed
- SunCoreGreen
- SunCoreBlue
- SunCoronaRed
- SunCoronaGreen
- SunCoronaBlue
- SunSize
- SpriteSize
- SpriteBrightness
- ShadowStrength
- LightShadowStrength
- PoleShadowStrength
- FarClip
- FogStart
- LightsOnGroundBrightness
- LowCloudsRed
- LowCloudsGreen
- LowCloudsBlue
- BottomCloudsRed
- BottomCloudsGreen
- BottomCloudsBlue
- WaterRed
- WaterGreen
- WaterBlue
- WaterAlpha
- ColourCorrectRed
- ColourCorrectGreen
[...]

PR #2989, opened 2023-05-02
Add new world properties (time cycle and weather related features)

### LINKED ISSUE #385: Ability to switch off street lights
This is a request from "zencro":

> It's currently impossible to turn off the lights at night time.

> Perhaps a setAmbientLightingEnabled/toggleAmbientLighting()?

> I checked to see if this is possible with shaders and it's not. Disabling `coronastar` makes the lampposts not shine, but it doesn't get rid of the actual light on the ground and also ends up disabling the stars and the sun. Maybe there's a more complex shader solution than that though.
>
> I think a function that makes everything darker would be more useful than something that disables street lights because even with all the street lights removed the world is still way too bright at night time and the only way to get close to proper darkness is one of those super dark weathers. Though this is almost realistic dark:
>
> ```
> crun setWeather(159)
> crun setSkyGradient(0, 0, 0, 0, 0, 0)
> ```

**Additional context**
> It's precisely for roleplay, and mainly for ambience. If a roleplay is set during a blackout, it would be nice to actually not have lights.

From https://bugs.mtasa.com/view.php?id=5543

### LINKED ISSUE #2723: "New" shadows flicker when they are far away from the camera, add an option to change them

On #1699, TFP worked on a setting that let you lower the shadows in game. Original SA has the shadows a bit too high. This fix works and makes shadows not appear floatty, but this makes them flicker.

There were plans to add a function to change the shadows height, or at least let the client decided whenever they would want the fix or not.

#2018 added this, but without an option to toggle it, since botder argued that this was a general improvement and no one would want it to be disabled. I disagree, because since this change I have been noticing a lot of flickering on far away shadows. I'm 90% sure this didn't happen before, but I could be wrong since this change was added so far ago.

https://user-images.githubusercontent.com/23291573/187042864-c3c0673d-73a6-4b91-88a7-4a3501a00bb1.mp4

We should have an option to enable or disable this, or even better, set the shadow height to whatever we want.

Consider adding the setting proposed on #1699.

Alternatively add a clientside function in which we can set the shadows height.

### Additional context

This visual error COULD be happening for other reason that #2018 (particulary that broken traffic light), I didn't test older versions and I didn't try to compile MTA with the default shadows. If you have encountered this issue, please let me know.

### Security Policy

### LINKED ISSUE #2778: Make volumetric shadows slighty darker

Volumetric shadows look ugly IMO. These would look a lot better if were blurred and little darker like blob shadows and dynamic ped shadows. Blurring them may be hard but changing their color is one byte change.

Make volumetric shadows a bit darker, maybe? It would be a _breaking_ (?) change, like, it would be different than it is in single player, so we could discuss that but I think these would seriously look better.

_No response_

### Additional context

This is how blob shadows look: https://i.imgur.com/IdIy20b.png
This is how it looks now (initial alpha set to 50): https://i.imgur.com/Mu1gEVL.png
This is how it looks with alpha set to 95:
https://i.imgur.com/xpB4Eeq.png
https://i.imgur.com/WqZ61bd.png

This is how it looks with alpha set to 85:
https://i.imgur.com/FKnUPBS.png

### Security Policy

### LINKED ISSUE #2785: Allow custom interval for setWeatherBlended

Currently if you change minute duration setWeatherBlended is unuseable.

```
setWeather(0)
setWeatherBlended(9, 10000) -- changes blended weather in 10 secs
```

```
setWeatherBlendTime()
resetWeatherBlendTime()
```

### Additional context

_No response_

### Security Policy

samr46:
This adds `getWorldProperty`, `setWorldProperty`, `resetWorldProperty` to modify remaining time cycle and weather related properties. Unlike `setTimeCycle` from stale PR #276 this implementation is consistent with existing related functions and adds more features.

Imgur album: https://imgur.com/a/jeUxx7L

| Property | Description |
| --- | --- |
| AmbientColor | The color of ambient light on map objects (including custom objects) |
| AmbientObjColor | The color of ambient light on dynamically created elements (peds, vehicles) |
| DirectionalColor | The color of direct light on dynamically created elements (peds, vehicles) |
| Illumination | Multiplier for the directional light (DirectionalColor) |
| SpriteSize | Point lights corona size (looks like only traffic lights are affected) |
| SpriteBrightness | Point lights corona brightness (also affects light on ground for light poles) |
| LightsOnGround | Point lights ground reflection brightness (looks like only traffic lights are affected) |
| PoleShadowStrength | Pole shadows alpha (used if volumetric shadows are disabled) |
| ShadowStrength | Shadows alpha (all shadows) |
| ShadowsOffset | Shadows height<br>Added for convenience to adjust shadows height on demand. Can be used to resolve #2723 |
| BottomCloudsColor | Bottom (normal) clouds color |
| LowCloudsColor | Low (skyline) clouds color (seems to be dependant on game hours) |
| CloudsAlpha | Bottom (normal) clouds alpha |
| | |
| WetRoads | Wet roads weather effect (mostly noticeable during driving)<br>Turns on lights on ground during daytime |
| Foggyness | Adds light fog effect for headlights and turns on lights on ground during daytime<br>Also affects skyline clouds and shadows visibility |
| Fog | Fog weather effect alpha |
| RainFog | Rain fog weather effect alpha (different fog; used in rain weathers) (reset is smooth) |
| WaterFog | Water fog alpha |
| Rainbow | Rainbow alpha |
| Sandstorm | Sandstorm sound volume (reset is smooth) |

Some properties like `AmbientColor` are float by default and adjusted to regular RGB for usability (consistent with timecyc.dat).

<hr>

### Discarded properties:

`ExtraSunnyness`, `CloudCoverage`
> Both properties work with `Foggyness` to determine visiblity of skyline clouds. Might actually be necessary if someone wants to use default skyline clouds and alter this behavior of default weathers. But most servers tend to use skydomes with custom clouds these days. And if needed I'd rather we add separate property that could change skyline clouds alpha.

`CloudsAlpha2`, `CloudsAlpha3`
> Appears to be useless. Changing `CloudsAlpha2` doesn't produce any visible results. `CloudsAlpha3` slightly changes water fog alpha which is redundant.

### Tested and can be added:

`TrafficLightsBrightness` or `TrafficLights`
> Changes traffic lights ground reflection brightness. By default game uses `WetRoads`, `Foggyness`, `Rain level` and game hour properties to determine if light reflections should be visible. So this could act like main override.

<hr>

Fixes #385
> setWorldProperty("SpriteBrightness", 0)

Fixes #2778
> setWorldProperty("ShadowStrength", 500)

Fixes #2785 (kind of)
> This function will allow developers to create custom weathers utilizing all time cycle and weather related properties (along with existing functions). So custom blending, 24h time cycles, anything is possible.

<hr>

Using new parser for this was tricky. But I think it looks good.

It shouldn't be too hard to review. Offsets are used to make memory address validation easier. Test resource with GUI.

Test resource: [worldprops.zip](https://github.com/multitheftauto/mtasa-blue/files/11375576/worldprops.zip)


Fernando-A-Rocha:
Wow amazing 😄

FileEX:
I really don't understand why no one has looked into this for a year. It's a pity that so few people have write access

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
     MemCpy((LPVOID)0x72BC92, &originalFstp1, 6);
     MemCpy((LPVOID)0x72C686, &originalFstp2, 6);
 }
+
+float CWeatherSA::GetWetRoads()
+{
+    return *(float*)0xC81308;
```
TracerDS, line 68: Maybe make it a `#define` so it could be reused later?

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+    MemCpy((LPVOID)(0x72BB9F + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BBB7 + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BBD0 + 1), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BBD7 + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetFoggyness()
+{
+    return *(float*)0xC81300;
```
TracerDS, line 94: Above

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+    BYTE originalCodes[3] = {0x00, 0x13, 0xC8};
+    MemCpy((LPVOID)(0x72BDF5 + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BDDD + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BE13 + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetFog()
+{
+    return *(float*)0xC812FC;
```
TracerDS, line 118: Above

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+    BYTE originalCodes[3] = {0xFC, 0x12, 0xC8};
+    MemCpy((LPVOID)(0x72BE37 + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BE1F + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BE4F + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetRainFog()
+{
+    return *(float*)0xC81410;
```
TracerDS, line 142: Above

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+{
+    BYTE originalCodes[3] = {0x10, 0x14, 0xC8};
+    MemCpy((LPVOID)(0x72ADD8 + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72ADE4 + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetWaterFog()
+{
+    return *(float*)0xC81338;
```
TracerDS, line 164: Above

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+    BYTE originalCodes[3] = {0x38, 0x13, 0xC8};
+    MemCpy((LPVOID)(0x72C35C + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72C38E + 2), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72C36F + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetSandstorm()
+{
+    return *(float*)0xC812F4;
```
TracerDS, line 188: Above

Client/game_sa/CWeatherSA.cpp
```diff
@@ -62,3 +62,169 @@ void CWeatherSA::ResetAmountOfRain()
...
+    BYTE originalCodes[3] = {0xF4, 0x12, 0xC8};
+    MemCpy((LPVOID)(0x72A4B6 + 1), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BCEB + 1), &originalCodes, 3);
+    MemCpy((LPVOID)(0x72BD0B + 2), &originalCodes, 3);
+    return true;
+}
+
+float CWeatherSA::GetRainbow()
+{
+    return *(float*)0xC812E4;
```
TracerDS, line 212: Above

Client/game_sa/CWeatherSA.h
```diff
@@ -26,4 +26,39 @@ class CWeatherSA : public CWeather
...
+    bool  SetWaterFog(float fAmount);
+    bool  ResetWaterFog();
+
+    float GetSandstorm();
+    bool  SetSandstorm(float fAmount);
+    bool  ResetSandstorm();
+
+    float GetRainbow();
+    bool  SetRainbow(float fAmount);
+    bool  ResetRainbow();
```
TracerDS: Please add `const` specifier for getters

Client/game_sa/CWeatherSA.h
```diff
@@ -26,4 +26,39 @@ class CWeatherSA : public CWeather
...
+
+private:
+    static unsigned char* VAR_CWeather__ForcedWeatherType;
+    static unsigned char* VAR_CWeather__OldWeatherType;
+    static unsigned char* VAR_CWeather__NewWeatherType;
```
TracerDS, line 61: `std::uint8_t` instead of `unsigned char`

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -2060,3 +2063,174 @@ bool CLuaWorldDefs::ResetCoronaReflectionsEnabled()
...
+        {
+            float red, green, blue;
+            g_pMultiplayer->GetAmbientObjectColor(red, green, blue);
+            return std::make_tuple((int16)(red * 255), (int16)(green * 255), (int16)(blue * 255));
+        }
+        case eWorldProperty::DIRECTIONAL_COLOR:
+        {
+            float red, green, blue;
+            g_pMultiplayer->GetDirectionalColor(red, green, blue);
+            return std::make_tuple((int16)(red * 255), (int16)(green * 255), (int16)(blue * 255));
```
TracerDS, line 2087: `std::int16_t` instead of `int16`

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -324,6 +324,32 @@ BYTE  ucSkyGradientTopB = 0;
...
+
+bool  bUsingCustomLowCloudsColor = false;
+int16 iLowCloudsColorR = 0;
+int16 iLowCloudsColorG = 0;
+int16 iLowCloudsColorB = 0;
+
+bool  bUsingCustomBottomCloudsColor = false;
+int16 iBottomCloudsColorR = 0;
+int16 iBottomCloudsColorG = 0;
+int16 iBottomCloudsColorB = 0;
```
TracerDS, line 354: `std::int16_t` instead of `int16`

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2224,6 +2250,302 @@ void CMultiplayerSA::ResetWater()
     MemPut<BYTE>(0x7051D7, 184);
 }
 
+void CMultiplayerSA::GetAmbientColor(float& red, float& green, float& blue)
+{
+    if (bUsingCustomAmbientColor)
+        red = fAmbientColorR, green = fAmbientColorG, blue = fAmbientColorB;
+    else
+        red = *(float*)0xB7C4A0, green = *(float*)0xB7C4A4, blue = *(float*)0xB7C4A8;
```
TracerDS, line 2268: `#define` for mem addresses maybe?

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2224,6 +2250,302 @@ void CMultiplayerSA::ResetWater()
...
+    bUsingCustomAmbientColor = false;
+    return true;
+}
+
+void CMultiplayerSA::GetAmbientObjectColor(float& red, float& green, float& blue)
+{
+    if (bUsingCustomAmbientObjectColor)
+        red = fAmbientObjectColorR, green = fAmbientObjectColorG, blue = fAmbientObjectColorB;
+    else
+        red = *(float*)0xB7C4AC, green = *(float*)0xB7C4B0, blue = *(float*)0xB7C4B4;
```
TracerDS, line 2291: Above

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2224,6 +2250,302 @@ void CMultiplayerSA::ResetWater()
...
+    bUsingCustomAmbientObjectColor = false;
+    return true;
+}
+
+void CMultiplayerSA::GetDirectionalColor(float& red, float& green, float& blue)
+{
+    if (bUsingCustomDirectionalColor)
+        red = fDirectionalColorR, green = fDirectionalColorG, blue = fDirectionalColorB;
+    else
+        red = *(float*)0xB7C4B8, green = *(float*)0xB7C4BC, blue = *(float*)0xB7C4C0;
```
TracerDS, line 2314: Above

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2224,6 +2250,302 @@ void CMultiplayerSA::ResetWater()
...
+{
+    MemPut<BYTE>(0x55FC21, 0xD9);
+    MemPut<BYTE>(0x55FC22, 0x5E);
+    MemPut<BYTE>(0x55FC23, 0x40);
+    return true;
+}
+
+float CMultiplayerSA::GetSpriteBrightness()
+{
+    return *(float*)0xB7C4E4;
```
TracerDS, line 2357: Above

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2224,6 +2250,302 @@ void CMultiplayerSA::ResetWater()
...
+{
+    MemPut<BYTE>(0x55FC34, 0xD9);
+    MemPut<BYTE>(0x55FC35, 0x5E);
+    MemPut<BYTE>(0x55FC36, 0x44);
+    return true;
+}
+
+int16 CMultiplayerSA::GetPoleShadowStrength()
+{
+    return *(int16*)0xB7C4EC;
```
TracerDS, line 2380: Above

Client/multiplayer_sa/CMultiplayerSA.h
```diff
@@ -195,6 +196,58 @@ class CMultiplayerSA : public CMultiplayer
...
+    bool  ResetBottomCloudsColor();
+
+    float GetCloudsAlpha1();
+    bool  SetCloudsAlpha1(float alpha);
+    bool  ResetCloudsAlpha1();
+
+    float GetIllumination();
+    bool  SetIllumination(float illumination);
+    bool  ResetIllumination();
+
```
TracerDS: `const` qualifier for getters, `std::int16_t` instead of `int16`

TracerDS (changes requested):
Good job

samr46:
@TracerDS
I didn't put memory addresses into definitions (and don't want to) to keep consistency within the file and with current implementation of similar functions. I don't see any reason to use #define for memory addresses in this case because it will make code messy and review harder.

I will make changes to getters and data types later today. Thank you.
Originally I didn't use `const` and `override` to make methods as similar as possible to other methods within the file (again for consistency). It actually looks weird right now after we started to use it in newer PRs without proper refactor of old code.

TracerDS:
> @TracerDS I didn't put memory addresses into definitions (and don't want to) to keep consistency within the file and with current implementation of similar functions. I don't see any reason to use #define for memory addresses in this case because it will make code messy and review harder.

Magic values are already a hindrance in mta source code. Having them as a definition makes it easier to understand

> I will make changes to getters and data types later today. Thank you. Originally I didn't use `const` and `override` to make methods as similar as possible to other methods within the file (again for consistency). It actually looks weird right now after we started to use it in newer PRs without proper refactor of old code.

👌

samr46:
> Having them as a definition makes it easier to understand

I actually agree with this and it's useful for reusing address within getter and setter. But it's harder to copy memory address during review / code inspection. And the last one is important since we have PRs without review for years due limited interest / manpower.

TracerDS:
> But it's harder to copy memory address during review / code inspection.

How so?

samr46:
> How so?

Double click and Ctrl + C is faster and more intuitive than hovering and clicking Copy or switching directly to #define value. The only benefit is that it can be easier to understand in some cases and less risk of using wrong address in multiple places. In this case all addresses are related to method names so we don't really have any problem with understanding what it will change.

TracerDS:
> Double click and Ctrl + C is faster and more intuitive than hovering and clicking Copy or switching directly to #define value. The only benefit is that it can be easier to understand in some cases and less risk of using wrong address in multiple places. In this case all addresses are related to method names so we don't really have any problem with understanding what it will change.

Thats just laziness that doesnt benefit anyone. Memory addresses dont change, definitions do.

samr46:
> Memory addresses dont change, definitions do.

In our case nothing will really change. This is why it's better to keep it consistent imo.

Client/game_sa/CWeatherSA.h
samr46: By the way, I'm not the author of these changes. It was merged from master in a1db8f4 (it looks like some private changes, I don't really know where it originated in and who is the author). It's not even used in public files as far as I know.

Client/multiplayer_sa/CMultiplayerSA.cpp
samr46: Technically `std::int16_t` and `int16` are the same (short is signed by default) 🤔
`int16` is more clean and consistent with reversed code

TracerDS: > Technically `std::int16_t` and `int16` are the same (short is signed by default) 🤔 `int16` is more clean and consistent with reversed code

No. They are **not** the same. `std::int16_t` is crossplatform while `int16` is not. Please use `std` variant instead

Client/game_sa/CWeatherSA.h
TracerDS: > By the way, I'm not the author of these changes. It was merged from master in [a1db8f4](https://github.com/multitheftauto/mtasa-blue/commit/a1db8f45913c7bc4b4f7c76fcd0163c9495961f1) (it looks like some private changes, I don't really know where it originated in and who is the author). It's not even used in public files as far as I know.

But a refactor in this case would be better

TracerDS:
> In our case nothing will really change. This is why it's better to keep it consistent imo.

Thats not consistency. Going with harder to understand codeblocks just because is not a good approach.
Having it defined as a preprocessor variable costs us nothing but benefits exceed the cons.

Magic values are **NEVER** the way to go

FileEX (approved):
LGTM. Debating whether to use a definition for addresses should not be part of this PR. The same applies to discussions about whether to use std library types (recommended for C++) or types from the global namespace (typical for C and available in C++ as well). Such debates should not be part of this PR, as they only lengthen its review. This is not a code refactoring PR. We all know that the MTA code is long spaghetti, but separate PRs are used for refactoring (by the way, they have been waiting for further steps for a long time). Imo code review concerns how the code was written in terms of conventions and consistency with the project, and in this case, consistency was maintained. Mixing it up so that half the code uses definitions and the other half uses direct addresses will only create even more spaghetti.

TracerDS:
Its better to not create problems that would add up in the future. That way there would be less code refactoring PRs.
Better to make changes earlier while there is still chance for merge

Client/multiplayer_sa/CMultiplayerSA.cpp
samr46: I'd use it for shared or server-side code. But this code is strictly client-side and not cross-platform. It uses the same definition on Windows.

To be honest with you, I don't agree with almost all code suggestions from you. It's not necessary at all and breaks consistency in modified files. I decided to make some changes only because I'm polite and this useful PR was ignored for a year.
For example I don't want to use #define for memory addresses in this particular PR. I strongly oppose not consistent code in files because it makes it harder to work with in the long run. If this is actually required to merge this PR then I'd like to get a comment from Member / someone who is interested in merging this PR.

TracerDS: So, you'd rather stick to unknown addresses without documenting them...?
Its alright. Your PR, your call, just trying to make the source code more reliable and less messy

ghost:
> Its better to not create problems that would add up in the future. That way there would be less code refactoring PRs. Better to make changes earlier while there is still chance for merge

I agree with you on this, the only problem with it is that these are not clarified in the community, they are not stipulated in the coding guidelines.

If these were all clarified and prescribed in a format readable by anyone, there would basically be no problem with it - it would be really useful in the case of a new implementation, preventing further refactoring. 

I personally think that to avoid future problems it would be worthwhile to use std namespace types, but it takes a lot of people's opinions to decide that, and this PR is not the place for that.

Client/multiplayer_sa/CMultiplayerSA.cpp
FileEX: Maybe let's make a small compromise and just add comments above the addresses for now?

TracerDS: Its alright. If it would bother other people then a refactoring PR will be posted.

FileEX:
To start with, the coding guidelines haven't been updated for several years. No official documentation is currently up to date regarding the development of the MTA. There is no one who would take care of it and develop it at the appropriate level. Arguing now on every PR will not solve this problem. We should seriously consider developing coding guidelines and establishing some clear community standards.

Client/multiplayer_sa/CMultiplayerSA.cpp
samr46: It's documented in reversed code. And this is actually the only place where it's useful.

There is a reason why other similar functions almost don't use definitions for addresses. For example you don't #define addresses when you NOP the same variable in different places because it makes code messy and hard to work with.
It makes sense to #define the address for overwritten variable. But you can also just put a new line there and keep this overwritten variable separated. And this is actually how it's used in all similar functions right now because it's easy to read and compare with reversed code.

samr46:
> To start with, the coding guidelines haven't been updated for several years. No official documentation is currently up to date regarding the development of the MTA. There is no one who would take care of it and develop it at the appropriate level. Arguing now on every PR will not solve this problem. We should seriously consider developing coding guidelines and establishing some clear community standards.

It's true. This is why in all my PRs I prefer to make code consistent with existing codebase even when it's not really good in some cases. Without proper refactor many things look out of place.

Client/multiplayer_sa/CMultiplayerSA.cpp
TracerDS: By reversed code you mean...?

samr46: > By reversed code you mean...?

I'm obviously talking about private `.idb` files of `gta_sa.exe` used for creating features like this.

TracerDS: > > By reversed code you mean...?
> 
> I'm obviously talking about private `.idb` files of `gta_sa.exe` used for creating features like this.

Thought you were talking about gta-reversed for a moment there. Its okay 👍

ghost (approved):
LGTM 🎉

TracerDS (approved):
lgtm

samr46:
If someone is interested in merging this for testing in nightly, I will add information to wiki within several hours.

TracerDS:
Resolve conversations, please 😃

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -58,6 +58,7 @@ void CLuaWorldDefs::LoadFunctions()
                                                                              {"getFPSLimit", GetFPSLimit},
                                                                              {"getBirdsEnabled", GetBirdsEnabled},
                                                                              {"getCoronaReflectionsEnabled", ArgumentParser<GetCoronaReflectionsEnabled>},
+                                                                             {"getWorldProperty", ArgumentParserWarn<false, GetWorldProperty>},
```
Lpsd: Is this an entirely new Lua function/definition? If so, use `ArgumentParser`, not `ArgumentParserWarn`

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -98,6 +99,7 @@ void CLuaWorldDefs::LoadFunctions()
                                                                              {"setMoonSize", SetMoonSize},
                                                                              {"setFPSLimit", SetFPSLimit},
                                                                              {"setCoronaReflectionsEnabled", ArgumentParser<SetCoronaReflectionsEnabled>},
+                                                                             {"setWorldProperty", ArgumentParserWarn<false, SetWorldProperty>},
```
Lpsd: Same here (`ArgumentParser`)

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -122,6 +124,7 @@ void CLuaWorldDefs::LoadFunctions()
                                                                              {"resetSunSize", ResetSunSize},
                                                                              {"resetMoonSize", ResetMoonSize},
                                                                              {"resetBlurLevel", ResetBlurLevel},
+                                                                             {"resetWorldProperty", ArgumentParserWarn<false, ResetWorldProperty>},
```
Lpsd, line 127: Same here (`ArgumentParser`)

Lpsd (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
samr46: This is new Lua function. However, if I use `ArgumentParser` then it will throw an error on incorrect / missing property name, wouldn't it?
It's not a problem for me personally but other similar functions throw warnings right now. Also this function is likely to be used in a loop (may cause unwanted loop breaks if we add new properties in the future and scripter forget to add client version checks)

Lpsd: All new functions should use `ArgumentParser` and are hard error on **usage** failure. Even when modifying existing functions, they should be switched to use `ArgumentParserWarn` if possible.

`ArgumentParserWarn` is just for compatibility so upgraded definitions produce a warning instead of error.

Currently you don't actually `throw` in the function, so you'll want to introduce a `default` case to the switch statement and throw in there, ie:

```cpp
default:
    throw std::invalid_argument{"Property not supported"};
```

You can also remove the `return false;` on the last line of the method, as it's not necessary by introducing this.

We shouldn't worry about user scripts failing because they used it wrong, they should worry about using it correctly ;)

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -2060,3 +2063,174 @@ bool CLuaWorldDefs::ResetCoronaReflectionsEnabled()
...
+
+std::variant<bool, float, CLuaMultiReturn<float, float, float>> CLuaWorldDefs::GetWorldProperty(eWorldProperty property)
+{
+    switch (property)
+    {
+        case eWorldProperty::AMBIENT_COLOR:
+        {
+            float red, green, blue;
+            g_pMultiplayer->GetAmbientColor(red, green, blue);
+            return std::make_tuple((int16)(red * 255), (int16)(green * 255), (int16)(blue * 255));
```
TracerDS, line 2075: can you do `return { red * 255, green * 255, blue * 255 };`?

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
samr46: It seems that default error message produced by ArgumentParser is more than enough to understand what is wrong in this case. So no need to throw custom error message. Thank you for this information regarding new argument parser.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -2060,3 +2063,174 @@ bool CLuaWorldDefs::ResetCoronaReflectionsEnabled()
...
+        }
+        case eWorldProperty::SPRITE_SIZE:
+            return g_pMultiplayer->GetSpriteSize();
+        case eWorldProperty::SPRITE_BRIGHTNESS:
+            return g_pMultiplayer->GetSpriteBrightness();
+        case eWorldProperty::POLE_SHADOW_STRENGTH:
+            return (float)g_pMultiplayer->GetPoleShadowStrength();
+        case eWorldProperty::SHADOW_STRENGTH:
+            return (float)g_pMultiplayer->GetShadowStrength();
```
TracerDS, line 2096: use static_cast instead, pleasee

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
samr46: Doesn't appear to work with `CLuaMultiReturn` as `std::variant`.

Dutchman101:
Regarding the coding guidelines, there's been some discussion on dev discord, and contributors took a draft initiative: [https://wiki.multitheftauto.com/wiki/New_coding_guidelines](https://wiki.multitheftauto.com/wiki/New_coding_guidelines)

This PR has been reviewed a bunch of times by good people, merging. Thanks to @samr46 for his work to get this highly demanded set of features into MTA.

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Dutchman101: If this is validated, it can be done by addendum.

PR #2990, opened 2023-05-03
Refactor, extract common part of CResource to CSharedResource

CrosRoad95:
goal of this pr is to just extract most common part of CResource to common class for client and server


botder:
Sorry, but you only extracted one member field to the common base class. This makes the code unnecessary complex.

PR #2992, opened 2023-05-03
Add rustlerbombs and speedometer to default startup resources

F2BShady:
Adds rustlerbombs and speedometer to default startup resources

### DIFF
--- a/Server/mods/deathmatch/mtaserver.conf
+++ b/Server/mods/deathmatch/mtaserver.conf
@@ -284,6 +284,8 @@
     <resource src="resourcebrowser" startup="1" protected="1" default="true"/>
     <resource src="resourcemanager" startup="1" protected="1"/>
     <resource src="scoreboard" startup="1" protected="0"/>
+    <resource src="rustlerbombs" startup="1" protected="0"/>
+    <resource src="speedometer" startup="1" protected="0"/>
     <resource src="spawnmanager" startup="1" protected="0"/>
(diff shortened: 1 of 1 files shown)


MegadreamsBE:
Any issue this resolves specifically? Any reason why those two resources should be run by default? I feel this should have been a discussion point in issues beforehand.

F2BShady:
> Any issue this resolves specifically? Any reason why those two resources should be run by default? I feel this should have been a discussion point in issues beforehand.

it may be needed for speedometer, but I can't say anything about rustlerbombs, the player can activate it at will, it's true what you said, but I think speedomeeter can be a good choice.

botder:
Adding speedometer sounds good, but I don't think we need rustlerbombs

Server/mods/deathmatch/mtaserver.conf
```diff
@@ -284,6 +284,8 @@
     <resource src="resourcebrowser" startup="1" protected="1" default="true"/>
     <resource src="resourcemanager" startup="1" protected="1"/>
     <resource src="scoreboard" startup="1" protected="0"/>
+    <resource src="rustlerbombs" startup="1" protected="0"/>
+    <resource src="speedometer" startup="1" protected="0"/>
```
botder:
```suggestion
    <resource src="speedometer" startup="1" protected="0"/>
```

botder (approved):

PR #2995, opened 2023-05-03
Fix ase description default value in mtaserver.conf and mtaserver.conf.template

muratdogdas:


patrikjuvonen (approved):
Thank you. That is correct.

PR #2997, opened 2023-05-05
Remove whitespaces surrounding some tags

androksi:
It was bothering me a lot, since I had to remove every time I install a MTA Server (yes). Probably doesn't affect anything, but I gone ahead and removed.


CrosRoad95 (commented):
its fine for me

F2BShady:
I tested it too and it's fine with me

patrikjuvonen (dismissed):
Thank you.

PR #3003, opened 2023-05-12
Feature: replace atomic models with clump models

CrosRoad95:
Goal of this pull request is to make possible to replace atomic models ( model with one submodel ) with clump models ( model with multiple submodels ).
To do this, i created new function that converts model `engineConvertAtomicToClumpModel( modelId )`
Only "atomic" models can be converted to clumps, otherwise function return false
Probably better solution would be to add function to change model type, but in practice you want "atomic" or "clump" objects, if you want new vehicles, peds models just request new id's.
In my opinion it should be separete function because attempt to put it into `engineReplaceModel` may lead to a lot of confusions.
Also, i consider another function to get model type, so for example by default: 370 returns "weapon", 404 "vehicle", 3425 "clump", 1337 "atomic".

```lua
engineConvertAtomicToClumpModel(1337)
a = createObject(1337, 4, 10, 5)
b = createObject(370, 15, 10, 5) -- don't need to convert to clump because this is model that already is clump
dff = engineLoadDFF ( "nt_windmill.dff" )
txd = engineLoadTXD ( "farmstuff.txd" )
engineImportTXD ( txd, 1337 )
engineImportTXD ( txd, 370 )
engineReplaceModel ( dff, 1337 )
engineReplaceModel ( dff, 370 )

addCommandHandler("restore", function()
  engineRestoreModel ( 1337 )
  engineRestoreModel ( 370 )
end)
```
before, only last atomic is visible:
image
after, all (2) atomics are visible:
image
1337 model on the left, 370 jetpack on the right

This pull request is still in progress, assume there are some memory leaks, i consider if it is important to revert model back to atomic


lopezloo:
> i consider if it is important to revert model back to atomic

It must be done on disconnect and `engineRestoreModel`. Is it already implemented?

Is the new API (`engineConvertAtomicToClumpModel`) really needed? Couldn't it be done automatically in `engineReplaceModel` (convert model to clump if specified model has multiple atomics)?

CrosRoad95:
> > i consider if it is important to revert model back to atomic
> 
> It must be done on disconnect and `engineRestoreModel`. Is it already implemented?
> 
> Is the new API (`engineConvertAtomicToClumpModel`) really needed? Couldn't it be done automatically in `engineReplaceModel` (convert model to clump if specified model has multiple atomics)?

it may be a case where people already have had models with multiple atomics/clump models but they didn't know and this change may cause old code to not be fully compatible thats why i added this function.

"engineRestoreModel" is working,
and yea, maybe i should keep track which model has been converted to clump, and on disconnect revert changes

CrosRoad95:
in ~2 weeks i'm going add functionality where models changes using "engineConvertAtomicToClumpModel(1337)" will be reverted back to atomic when you left the server. if anyone has any objectives, let me know

CrosRoad95:
Now it works a bit differently, instead of "convert" you change internal type of model, for now only availiable options are "clump" and "atomic", limitation is that you can change from atomic to clump, but not from clump to atomic until you previously set it to clump. I have made it to prevent potential crashes because i'm 100% sure that crash will happen if you convert ped to atomic, atomic to vehicle ect.

```
engineSetModelInfoType(1337, "clump") -- valid because it was atomic
engineSetModelInfoType(1337, "atomic") -- valid because it originally was "atomic"
engineSetModelInfoType(3425, "atomic") -- invalid because it originally was "clump"
engineSetModelInfoType(370, "clump") -- not supported, throw error because this is "weapon" and for now this conversion is not supported
```
there's second function "engineGetModelInfoType" that return internal type of model, so for example `engineGetModelInfoType(370)` is weapon, `engineGetModelInfoType(1337)` is atomic, `engineGetModelInfoType(3425)` is clump
now if you leave server, all models are reverted back to atomic if they were changed to clump.
Someone in future can implement support for other type change so for example you will be able to convert clump to atomic ( what is kinda useless )

first you have to change type, then replace it ( or probably restream world when type change )

there are things i, you need to test:
1. changing custom allocated models
2. maybe there are streaming issues?
3. maybe some models after conversion fails?

lopezloo:
I think we should discuss more about possibility of merging this into existing API (`engineReplaceModel`). It's up to server owners to take care of models. If they have models with multiple atomics then it's their liability.

CrosRoad95:
> I think we should discuss more about possibility of merging this into existing API (`engineReplaceModel`). It's up to server owners to take care of models. If they have models with multiple atomics then it's their liability.

removed all new lua functions, now if you replace model it get converted to clump automatically, restore will convert model back to atomic if needed, same when you leave the server.
Maybe functions responsible for conversion could be improved but apart of that i think this pr is ready to finish

Pirulax:
What would be the use-case for this though?
Could you provide a few examples?

CrosRoad95:
mostly for future pull requests, it opens doors to many features.

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -496,6 +496,50 @@ bool AtomicsReplacer(RpAtomic* pAtomic, void* data)
...
+#define RpClumpGetFrame(_clump)      RpClumpGetFrameMacro(_clump)
+
+static auto CModelInfo_ms_modelInfoPtrs = (CBaseModelInfoSAInterface**)ARRAY_ModelInfo;
+
+static void CVisibilityPlugins_SetAtomicId(RpAtomic* pRpAtomic, int id)
+{
+    return ((void(__cdecl*)(RpAtomic*, int))0x732230)(pRpAtomic, id);
+}
+
+typedef struct
```
tederis: Perhaps it's better not to use C-style struct declaration.

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -70,8 +70,50 @@ bool CClientModel::Allocate(ushort usParentID)
...
 }
 
+bool CClientModel::MakeAtomicModel()
+{
+    if (!m_bWasConvertedToClump)
+        return false;
+
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
+
+    if (!pModelInfo->IsValid() || m_eModelType != eClientModelType::OBJECT)
```
tederis: **GetModelInfo** can return **nullptr** what should be taken into account in this conditional expression.

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -70,8 +70,50 @@ bool CClientModel::Allocate(ushort usParentID)
...
+    if (pModelInfo->MakeAtomicModel(m_iModelID))
+        m_bWasConvertedToClump = false;
+    return true;
+}
+
+bool CClientModel::MakeClumpModel()
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
+
+    if (!pModelInfo->IsValid() || m_eModelType != eClientModelType::OBJECT)
```
tederis: The same

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+{
+    RpAtomic** firstAtomicPtr = (RpAtomic**)data;
+
+    if (*firstAtomicPtr == nullptr)
+        *firstAtomicPtr = atomic;
+
+    return false;
+}
+
+bool CModelInfoSA::MakeAtomicModel(ushort usModelId)
```
tederis: What is the purpose of usModelId? Why don't just use m_dwModelID instead? Or usModelId can differ?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+    return true;
+}
+
+bool CModelInfoSA::MakeClumpModel(ushort usModelId)
+{
+    if (GetModelType() != eModelInfoType::ATOMIC)
+        return false;
+
+    CBaseModelInfoSAInterface*  pBaseObjectInfo = (CBaseModelInfoSAInterface*)ppModelInfo[usModelId];
+    CClumpModelInfoSAInterface* m_pInterface = new CClumpModelInfoSAInterface();
```
tederis: m_ prefix should correspond only to class members

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+    CClumpModelInfo_SA_VTBL* vfbl = (CClumpModelInfo_SA_VTBL*)vftable_CClumpModelModelInfo;
+    m_pInterface->VFTBL = vfbl;
+    RpClump*  pClump = RpClumpCreate();
+    RpAtomic* pAtomic = (RpAtomic*)m_pInterface->pRwObject;
+    RpSetFrame(pClump, RwFrameCreate());
+    RpClumpAddAtomic(pClump, pAtomic);
+
+    ((void(__thiscall*)(CClumpModelInfoSAInterface * pThis, RpClump * clump)) vfbl->SetClump)(m_pInterface, pClump);
+
+    ppModelInfo[m_dwModelID] = m_pInterface;
```
tederis: I see a potential memory leak in here. Because you're creating a new instance of CClumpModelInfoSAInterface, but the previous one keeps living in memory.

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+    RpAtomic* firstAtomic = nullptr;
+    RpClumpForAllAtomics(pClump, GetFirstAtomic, &firstAtomic);
+    RwFrame* pFrame = RwFrameCreate();
+    RpAtomic* pClonedAtomic = RpAtomicClone(firstAtomic);
+    RpAtomicSetFrame(pClonedAtomic, pFrame);
+
+    m_pInterface->pRwObject = nullptr;
+    ((void(__thiscall*)(CAtomicModelInfoSAInterface * pThis, RpAtomic * pAtomic)) vfbl->SetAtomic)(m_pInterface, pClonedAtomic);
+
+    ppModelInfo[m_dwModelID] = m_pInterface;
```
tederis: The same potential memory leak

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+    return false;
+}
+
+bool CModelInfoSA::MakeAtomicModel(ushort usModelId)
+{
+    if (GetModelType() != eModelInfoType::CLUMP)
+        return false;
+
+    CBaseModelInfoSAInterface*   pBaseObjectInfo = (CBaseModelInfoSAInterface*)ppModelInfo[usModelId];
+    CAtomicModelInfoSAInterface* m_pInterface = new CAtomicModelInfoSAInterface();
```
tederis: The same prefix issue

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1728,6 +1734,69 @@ void CModelInfoSA::MakeVehicleAutomobile(ushort usBaseID)
...
+    MemCpyFast(m_pInterface, pBaseObjectInfo, sizeof(CBaseModelInfoSAInterface));
```
tederis: Copying of a whole class object is quite dangerous. I doubt that this class is a POD structure that can be copied without any issues.

CrosRoad95: this pattern already exists in "void CModelInfoSA::MakeObjectModel(ushort usBaseID)" function

Client/game_sa/CAtomicModelInfoSAInterface.h
```diff
@@ -0,0 +1,21 @@
...
+
+class CAtomicModelInfo_SA_VTBL : public CBaseModelInfo_SA_VTBL
+{
+public:
+    DWORD SetAtomic;            // (RpAtomic*)
+};
+
+class CAtomicModelInfoSAInterface : public CBaseModelInfoSAInterface
+{
+};
```
TheNormalnij, line 21: This struct has additional field
```
unsigned short usAnimId;
```

PR #3013, opened 2023-05-19
Remove some strange undocumented logic from `Client/core/CCommandFuncs.cpp`

lopezloo:
Removes some strange undocumented cursor-related logic from `Client/core/CCommandFuncs.cpp`.


CrosRoad95:
remove "AllowWindowsCursorShowing" method too because it is only place from where it is called

lopezloo:
> remove "AllowWindowsCursorShowing" method too because it is only place from where it is called

Could be useful. In fact, I'm using it in PR #2855.

patrikjuvonen (approved):
Tested lightly. Can't see any reason for these indeed.

PR #3017, opened 2023-05-19
Fix #418: addCommandHandler: Returning False for Existing Built-in Commands

### LINKED ISSUE #418: AddCommandHandler problem
**Describe the bug**
If you trying to add command example for `showmemstat` command, addCommandHandler returning `true` even command isn't added.

**To Reproduce**
clientside
```lua
created = addCommandHandler("showmemstat", function()
  outputChatBox( "showmemstat working?") -- will never displayed
end)
outputChatBox( tostring(created) ) -- true
```
if you execute command `showmemstat`, only mem stat window appear.

**Expected behavior**
Should return false for some exeptions

**Screenshots**
-

Multi Theft Auto v1.5.5-release-13968

MTA:SA Server v1.5.5-release-11751

**Additional context**
-

CrosRoad95:
Fixes #418
```lua
iprint(addCommandHandler("help", function()end)) -- false
iprint(addCommandHandler("asd", function()end)) -- true
```

+ fix tiny bug where server console commands are added after resource starts what would make "help" command in example above to return true for first time

### DIFF
--- a/Client/mods/deathmatch/logic/CRegisteredCommands.cpp
+++ b/Client/mods/deathmatch/logic/CRegisteredCommands.cpp
@@ -27,6 +27,9 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (g_pCore->GetCommands()->Get(szKey))
+        return false;
+
     // Check if we already have this key and handler
     SCommand* pCommand = GetCommand(szKey, pLuaMain);
     if (pCommand)
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -660,6 +660,8 @@ bool CMainConfig::LoadExtended()
         }
     } while (pNode);
 
+    RegisterCommands();
+
     // Handle the <resource> nodes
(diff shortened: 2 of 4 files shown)


patrikjuvonen (approved):
Thanks, seems fine to me.

samr46:
This broke "radar" bind. No longer possible to replace default Map using this command.

Previously it was possible if "radar" control is disabled. And it's useful.

```
toggleControl("radar", false)
addCommandHandler("radar", function(cmd) outputChatBox(cmd) end)
```

CrosRoad95:
<https://github.com/multitheftauto/mtasa-blue/pull/3031#issuecomment-1558097143> - maybe it should be just backwards incompatible change

samr46:
> maybe it should be just backwards incompatible change

What's the point of removing simple and clean way of using "radar" command to replace Map? The alternative would be tracking user binds or binding custom map regardless of user selected binds. Unnecessary additional logic.

I'm also wondering what was the point of fixing this original minor issue in a way that breaks this. I think if you really want to fix it then you have to do it right and not simply block all default MTA commands. Because in fact it was always possible to use some of default commands (the ones that share control / default MTA binds) in `addCommandHandler`.

patrikjuvonen:
People are angry about the change, hence I've reverted it now.

If you want this implemented in the future, please gather more input first and make sure everyone is happy.

PR #3018, opened 2023-05-20
Feature: Refactor UTF defs to use new lua parser

CrosRoad95:
as in title


patrikjuvonen (approved):
Thanks, good work. See my additional commits for future reference.

PR #3019, opened 2023-05-20
Add new account events: onAccountCreate/onAccountRemove & new function getAccountType

TracerDS:
New events:
- `onAccountCreate` (takes 1 parameter: account)
- `onAccountRemove` (takes 1 parameter: account)

New function:
- `getAccountType`


CrosRoad95:
looks okey for me

botder:
What kind of use case do you have for these two events? Also, assuming you've tested these events, can you even process the events for "console accounts" and does it even make sense to trigger that event for guest accounts?

TracerDS:
> What kind of use case do you have for these two events?

Well, we have some events for creation of elements (`onClientProjectileCreation`, `onClientBrowserCreated`), why not have one for creation of accounts? 

> can you even process the events for "console accounts"

Yes, sir.

For guest accounts, its a debatable thing in my opinion. It may or may not be necessary to fire an event when someone joins but its better to have more than less. or maybe other event (onGuestAccountCreated) might be fired specifically for guest accounts?

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -891,46 +892,44 @@ bool CAccountManager::GetAllAccountData(CAccount* pAccount, lua_State* pLua)
     m_pDatabaseManager->QueryWithResultf(m_hDbConnection, &result, "SELECT key,value,type from userdata where userid=?", SQLITE_INTEGER, iUserID);
 
     // Do we have any results?
-    if (result->nRows > 0)
+    if (result->nRows <= 0)
+        return false;
+
+    // Loop through until i is the same as the number of rows
+    for (const auto& row : *result.operator->())
```
tederis: I bet `for (const auto& row : result->Data)` would fit better here

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -891,46 +892,44 @@ bool CAccountManager::GetAllAccountData(CAccount* pAccount, lua_State* pLua)
...
     {
-        // Loop through until i is the same as the number of rows
-        for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+        // Get our key
+        strKey = reinterpret_cast<const char*>(row[0].pVal);
+        // Get our type
+        int iType = static_cast<int>(row[2].nVal);
```
tederis, line 902: It's better to use something more readable instead of magic indices in `row[0], row[1], ...`

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -939,10 +938,8 @@ void CAccountManager::GetAccountsBySerial(const SString& strSerial, std::vector<
     CRegistryResult result;
     m_pDatabaseManager->QueryWithResultf(m_hDbConnection, &result, "SELECT name FROM accounts WHERE serial = ?", SQLITE_TEXT, strSerial.c_str());
 
-    for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+    for (const auto& row : *result.operator->())
```
tederis: The same

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -955,10 +952,8 @@ void CAccountManager::GetAccountsByIP(const SString& strIP, std::vector<CAccount
     CRegistryResult result;
     m_pDatabaseManager->QueryWithResultf(m_hDbConnection, &result, "SELECT name FROM accounts WHERE ip = ?", SQLITE_TEXT, strIP.c_str());
 
-    for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+    for (const auto& row : *result.operator->())
```
tederis: The same

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -970,10 +965,8 @@ CAccount* CAccountManager::GetAccountByID(int ID)
     CRegistryResult result;
     m_pDatabaseManager->QueryWithResultf(m_hDbConnection, &result, "SELECT name FROM accounts WHERE id = ?", SQLITE_INTEGER, ID);
 
-    for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+    for (const auto& row : *result.operator->())
```
tederis: The same

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -988,39 +981,53 @@ void CAccountManager::GetAccountsByData(const SString& dataName, const SString&
                                          "SELECT acc.name FROM accounts acc, userdata dat WHERE dat.key = ? AND dat.value = ? AND dat.userid = acc.id",
                                          SQLITE_TEXT, dataName.c_str(), SQLITE_TEXT, value.c_str());
 
-    for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+    for (const auto& row : *result.operator->())
```
tederis: The same

Server/mods/deathmatch/logic/CAccountManager.cpp
```diff
@@ -1173,9 +1174,8 @@ void CAccountManager::LoadAccountSerialUsage(CAccount* pAccount)
                                          " WHERE userid=?",
                                          SQLITE_INTEGER, pAccount->GetID());
 
-    for (CRegistryResultIterator iter = result->begin(); iter != result->end(); ++iter)
+    for (const auto& row : *result.operator->())
```
tederis: The same

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -79,6 +79,8 @@ class CAccount
     const SString& GetName() { return m_strName; }
     void           SetName(const std::string& strName);
 
+    EAccountType   GetType() { return m_AccountType; }
```
tederis: **const** qualifier required

tederis:
As you decided to refactor the existing code I would suggest to take a look at **CRegistryResultCell**. It stores **nVal**, **fVal** and **pVal** all the time but only one can be in use at one time. So its better to use **union** to save memory and promote true sharing while looping.

botder:
> For guest accounts, its a debatable thing in my opinion. It may or may not be necessary to fire an event when someone joins but its better to have more than less. or maybe other event (onGuestAccountCreated) might be fired specifically for guest accounts?

I can understand having the even trigger for player accounts, but I don't see any value for console or guest accounts - that's just noise.

TMachadoDev (commented):
I think its all good, in fact perfect.

Dutchman101:
This has had enough feedback and eyes, thanks!

@TracerDS please add the 2 events and function to the wiki

PR #3020, opened 2023-05-21
Use DXSDK_DIR env variable instead of full path

TracerDS:
Use `DXSDK_DIR` env variable instead of providing full path to DX SDK dir;
That way MTA project wont be affected when `DXSDK_DIR` will be changed.

When creating projects via `premake5`, path to the `DXSDK_DIR` will be an absolute path. That might break if you decide to change location of the dir. You would have to change the location in the project settings too which might be annoying.


patrikjuvonen:
Build is failing.

TracerDS:
> Build is failing.

Thats probably because github cant detect `DXSDK_DIR` env variable, thus its failing to find these files.
But it fails only in win32 build...?

patrikjuvonen:
> Thats probably because github cant detect `DXSDK_DIR` env variable, thus its failing to find these files. But it fails only in win32 build...?

Nothing has changed here though: https://github.com/multitheftauto/mtasa-blue/blob/master/.github/workflows/build.yaml#L42

It fails for win32 only because only win32 builds the client projects that make use of DX.

TracerDS:
> > Thats probably because github cant detect `DXSDK_DIR` env variable, thus its failing to find these files. But it fails only in win32 build...?
> 
> Nothing has changed here though: https://github.com/multitheftauto/mtasa-blue/blob/master/.github/workflows/build.yaml#L42
> 
> It fails for win32 only because only win32 builds the client projects that make use of DX.

Cant reproduce it :/ It works perfectly for me but not for the github job

It looks like github has some [problems](https://github.com/multitheftauto/mtasa-blue/actions/runs/5037294023/jobs/9034029656#step:8:125) with env variables and dont recognize `DXSDK_DIR` variable

botder:
I don't quiet understand what the point of this pull request is? The premake script is already using the environment variable and I don't think there is a requirement for an absolute path there.

TracerDS:
> I don't quiet understand what the point of this pull request is? The premake script is already using the environment variable and I don't think there is a requirement for an absolute path there.

Absolute path is **currently** being provided to VS projects. The point of this PR was to provide an env variable to VS include path. You can always change the env variable, but changing absolute path in >6 projects is kinda... annoying.

botder:
I still don't understand the issue. It works for me when I am using both absolute and relative path to DXFiles.

For example, this will generate projects each with a relative path (`..\utils\DXFiles`) to DXFiles.
```ps
$env:DXSDK_DIR = "utils/DXFiles"
.\win-create-projects.bat
```

TracerDS:
> I still don't understand the issue. It works for me when I am using both absolute and relative path to DXFiles.
> 
> For example, this will generate projects each with a relative path (`..\utils\DXFiles`) to DXFiles.
> 
> ```postscript
> $env:DXSDK_DIR = "utils/DXFiles"
> .\win-create-projects.bat
> ```

Is your DX dir path in VS directories set to `$(DXSDK_DIR)` or to an absolute one?

botder:
> Is your DX dir path in VS directories set to `$(DXSDK_DIR)` or to an absolute one?

It's relative.

image

TracerDS:
> > Is your DX dir path in VS directories set to `$(DXSDK_DIR)` or to an absolute one?
> 
> It's relative.
> 
> image

Yeah, and my PR's purpose is to replace that path with `$(DXSDK_DIR)` env variable.
```xml
<AdditionalIncludeDirectories>..\vendor;$(DXSDK_DIR)\Include;..\vendor\...</AdditionalIncludeDirectories>
```

win-create-projects.bat
```diff
@@ -9,6 +9,13 @@ utils\premake5.exe install_unifont
 rem Generate solutions
 utils\premake5.exe vs2022
 
+rem Check utils folder
+echo ------------------
+dir /B .\utils\
+echo ------------------
+dir /B .\utils\DXFiles
+echo ------------------
+
```
botder: This looks like debug code. It's not needed.
```suggestion
```

premake5.lua
```diff
@@ -45,7 +45,9 @@ workspace "MTASA"
 	symbols "On"
 	flags "MultiProcessorCompile"
 
-	dxdir = os.getenv("DXSDK_DIR") or ""
+	dxsdk = os.getenv("DXSDK_DIR") and "$(DXSDK_DIR)" or ""
+	print(('DXSDK: [%s] [%s]'):format(dxsdk, os.getenv("DXSDK_DIR")))
+
 	includedirs {
```
botder:
```suggestion
	includedirs {
```

premake5.lua
```diff
@@ -110,10 +112,10 @@ workspace "MTASA"
 		defines { "WIN32", "_WIN32", "_WIN32_WINNT=0x601", "_MSC_PLATFORM_TOOLSET=$(PlatformToolsetVersion)" }
 		buildoptions { "/Zc:__cplusplus" }
 		includedirs {
-			path.join(dxdir, "Include")
+			dxsdk.."Include"
 		}
 		libdirs {
-			path.join(dxdir, "Lib/x86")
+			dxsdk.."Lib/x86"
 		}
```
botder: This is enough:
```lua
		includedirs { "$(DXSDK_DIR)/Include" }
		libdirs { "$(DXSDK_DIR)/Lib/x86" }
```

botder (changes requested):
Can you show a warning if the environment variable `DXSDK_DIR` is empty or unset (only on Windows)?

win-create-projects.bat
TracerDS: > This looks like debug code. It's not needed.

Yeah, I tried to pinpoint the culprit. Couldnt do it without commiting ;/

botder:
You should also change [build.yaml](https://github.com/multitheftauto/mtasa-blue/blob/master/.github/workflows/build.yaml#L42) and try removing the `./` prefix there

botder:
There is no reason to change the current code. It also works with a relative path.

PR #2979, opened 2023-04-24
mtaserver.conf fps change

efexc:
For more ability and performance, it can be 60. I think it will be very useful for those who are new on the MTA Server adventure. Thanks. :)

PR #3023, opened 2023-05-21
Raise default fps limit to 74

Dutchman101:
I propose to use FPS limit of 74 over 60 (as added in b9f5c64), because 74 is the FPS with consensus that it avoids the majority of framerate-related GTA bugs just as well as other high values (lower than 74) do.

Despite that PR #2979 didn't have to mention it to go through, there has been plenty of discussion about the safe FPS (74..) in development discord over the past 3 years, besides that some of this knownledge and testing results made its way to being documented at [https://wiki.multitheftauto.com/wiki/SetFPSLimit](https://wiki.multitheftauto.com/wiki/SetFPSLimit) under "Issues when increasing FPS". An fps of 74 was called the breaking point after which GTA bugs really begin to appear, long before notable framerate bugfixes were made by contributors such as Merlin, so in reality the safe limit is definately beyond 74 (To be exact, it was the point after which "Aiming while moving sideways" bug first appeared, but it's since been [fixed](https://github.com/multitheftauto/mtasa-blue/commit/e64d311f62de2bd848c07b59f4f53a30826c1bed)). Although someone who has time for testing post-framerate fixes can determine a new base limit that gets labeled as the highest safe value, let's be moderate and go to 74 for now.

TL;DR - The jump from 60 to 74 FPS has no additional impact when it comes to framerate related bugs, but it does give that final push of smoothness that improves gameplay experience, this is a jump that your eyes can still perceive in SA.

Opinions differ a lot, some people believe that every increase in FPS worsens effects of higher framerates on things like vehicle physics for racing, but consider this proposal in the light that the leap from previous value (36) to 60 already did what those people complained about, and that my point is: it would've been better if the merged PR had initially settled on 74. It's key to realize this, as well as ` .. The jump from 60 to 74 FPS has no additional impact .. ` and put things into perspective like this, to avoid having a battle of opinions. We are to use higher FPS by default in MTA 1.6, and that already commits the 'evil' those people see, but it's very proportional as tiny effects they seem to perceive on racing physics vs. offering the best possible configuration out of the box, i.e visual smoothness and game experience. If racing server owners think that these higher FPS limits cause issues, they can always customize the value.

### DIFF
--- a/Server/mods/deathmatch/editor.conf
+++ b/Server/mods/deathmatch/editor.conf
@@ -190,8 +190,8 @@
     <filter_duplicate_log_lines>1</filter_duplicate_log_lines>
 
     <!-- Specifies the frame rate limit that will be applied to connecting clients.
-         Available range: 25 to 32767. Default: 60. You can also use 0 for uncapped fps. -->
-    <fpslimit>60</fpslimit>
+         Available range: 25 to 32767. Default: 74. You can also use 0 for uncapped fps. -->
+    <fpslimit>74</fpslimit>
 
     <!-- This parameter specifies whether or not to enable player voice chat in-game
          Values: 0 - disabled , 1 - enabled -->
--- a/Server/mods/deathmatch/local.conf
+++ b/Server/mods/deathmatch/local.conf
@@ -190,8 +190,8 @@
     <filter_duplicate_log_lines>1</filter_duplicate_log_lines>
 
     <!-- Specifies the frame rate limit that will be applied to connecting clients.
-         Available range: 25 to 32767. Default: 60. You can also use 0 for uncapped fps. -->
-    <fpslimit>60</fpslimit>
+         Available range: 25 to 32767. Default: 74. You can also use 0 for uncapped fps. -->
+    <fpslimit>74</fpslimit>
 
     <!-- This parameter specifies whether or not to enable player voice chat in-game
          Values: 0 - disabled , 1 - enabled -->
--- a/Server/mods/deathmatch/mtaserver.conf
+++ b/Server/mods/deathmatch/mtaserver.conf
@@ -190,8 +190,8 @@
     <filter_duplicate_log_lines>1</filter_duplicate_log_lines>
 
     <!-- Specifies the frame rate limit that will be applied to connecting clients.
-         Available range: 25 to 32767. Default: 60. You can also use 0 for uncapped fps. -->
-    <fpslimit>60</fpslimit>
+         Available range: 25 to 32767. Default: 74. You can also use 0 for uncapped fps. -->
+    <fpslimit>74</fpslimit>
 
     <!-- This parameter specifies whether or not to enable player voice chat in-game
          Values: 0 - disabled , 1 - enabled -->


Dark-Dragon:
I understand the reasoning, but in my opinion 60 makes a lot more sense because it's such a common refresh rate for screens. In other words 74 will not benefit nearly as many people as the previous increase did. The complaint that many issues progressively get worse the more FPS you run at is valid from my experience. Server owners unknowingly crippling their players FPS also doesn't appear to be an issue from my experience, as usually if they stick to something lower they have a reason for it.

Dutchman101:
> Server owners unknowingly crippling their players FPS also doesn't appear to be an issue from my experience, as usually if they stick to something lower they have a reason for it

Less than ideal experience (where a gain in visual smoothness is possible between 60 and 74) isn't crippling, no. It is just something that can be done better. My point regarding that isn't about server owners' actions or whether or not they are providing a less than ideal experience to players, but about MTA providing ideal configuration out of the box. We can even put it like it'll be for the sake of ideal configuration for most servers, where server owners that feel like an exception, such as those who think racing is impacted, can easily just modify the FPS limit as they desire.. not to speak of that we're not forcing existing servers to update their FPS limit here. Anyone that'll be setting up a new server will be able to make their own choice, but FPS of 74 over 60 benefits the masses who play the most prevalent gamemodes, which is the goal here.

Regarding the refresh rate: an increasing share of players has more capable monitors, and they will only increase. Even without vsync, those who play at 74fps on 60hz monitors won't see more artifacts, it's not enough of a difference for that.

Dark-Dragon:
If we're talking about the "ideal experience" I'd go as far as saying increasing the default to 60 before taking care of the frame rate issues was the wrong call. IMO Accuracy for physics etc is more important than visual smoothness, but I can see why others would disagree with this.

samr46:
> Accuracy for physics etc is more important than visual smoothness

This is why it's not 100 or unlimited. 70-74 is current safe limit. It will not affect physics much compared to 60 but will be beneficial for players with high refresh rate monitors.

Ultimately most potentially affected servers (race) use custom FPS limit anyway. And it wouldn't affect any existing server unless they manually revert server configs to updated default (and even in this case it would take a second to edit FPS limit).

Dutchman101:
> If we're talking about the "ideal experience" I'd go as far as saying increasing the default to 60 before taking care of the frame rate issues was the wrong call.

I already see it.. you lack the background on these FPS things (being in the loop on what it is, the extremely trivial impact of certain things, and what the current situation is. All of which are neccesary to judge proportionality and relativity). You got the wrong idea entirely, we should try to keep this on track. There have been many fixes for framerate issues in the last 2 years, and now higher fps limits are more than feasible & free from noticable disruptions to gameplay. Yet, the proposed limit of 74 is to be moderate and avoid sliding down the scale of risk taking for predisposition to FPS-related GTA bugs and their increase in noticability when going towards the 'extremely high FPS values'. Like towards 100, you may really feel altered physics in races. 

It may be useful to check [https://github.com/multitheftauto/mtasa-blue/projects/14](https://github.com/multitheftauto/mtasa-blue/projects/14)

And the reason that 60 already went through is due to the people involved in approving it, having a clear picture of how things are as a result of the discussion and knownledge i referred to. It's not the wrong call, and in regards to bug safety there'll be no difference between going from 36 to 60 as for going from 36 to 74 (i think ideal).

Dark-Dragon:
I mean you can question my expertise if you'd like, but I've played the game for a pretty long time, and while I can't contribute to addressing the issues, or could point at the code and tell you where the issues stems from, I know a lot of them and I've seen a lot of them at various FPS limits.

All I'm saying is that 60 is a pretty well established FPS limit that probably covers a large majority of the player base perfectly and that going from 60 to 74 is a pretty minor improvement in smoothness.

But I'm not here to stop you or anyone from achieving their vision for MTA, in the end I'm not directly affected by this. I just thought it would be valuable to have some kind of feedback.

AlexTMjugador:
> Regarding the refresh rate: an increasing share of players has more capable monitors, and they will only increase. Even without vsync, those who play at 74fps on 60hz monitors won't see more artifacts, it's not enough of a difference for that.

Out of curiosity, do MTA stats collect monitor refresh rates? My gut feeling is that the bump indeed probably doesn't matter much for the vast majority of players _right now_. 75 Hz used to be a common refresh rate for 1280x1024 screens, but that's no longer the case with modern low-end 1920x1080 displays. And indeed running the game at 75 FPS on a 60 Hz display may cause artifacts, even though they are so minor to notice.

Edit: by the way, players with high-end, high refresh rate monitors should probably be already using variable refresh rate features to achieve the smoothest experience.

PR #3029, opened 2023-05-21
Update local.conf

F2BShady:
I'm also someone who prefers the FPS limit to be 74


patrikjuvonen:
Please stop creating new pull requests when there is a pull request already at #3023

PR #3035, opened 2023-05-24
Increase main menu FPS limit

lopezloo:
Increases main menu (when not connected to any server) FPS limit from 88 to 240. GUI is a lot smoother on higher framerates.

### DIFF
--- a/Client/core/CModManager.cpp
+++ b/Client/core/CModManager.cpp
@@ -285,7 +285,7 @@ void CModManager::DoPulsePostFrame()
     if (m_pClientBase)
         CCore::GetSingleton().EnsureFrameRateLimitApplied();            // Catch missed frames
     else
-        CCore::GetSingleton().ApplyFrameRateLimit(200);            // Limit when not connected
+        CCore::GetSingleton().ApplyFrameRateLimit(240);            // Limit when not connected
 
     // Load/unload requested?
     if (m_bUnloadRequested)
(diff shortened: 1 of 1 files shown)


Dutchman101 (changes requested):
Please record 3 video's of how it looks on different refresh rates: 60Hz, 144Hz and if you can also 240Hz. Move around a few menu's, scroll them and change the position of a few main interfaces like settings GUI and server list, also play the credits slideshow. Now we can do a side by side comparison.

Take them with a physical camera instead of software, we will be looking for adverse effects such as artifacts or excessive blurring. 

Either way i dont think this PR can be merged until weeks after the release of 1.6, because we'll need to have others test the results on older monitors (like actual bricks, not one where you can easily configure refresh rates) in a nightly.

lopezloo:
Thinking about it now, we actually could limit it to monitor Hz and apply higher than 88 FPS limit only if it would make sense.

Dutchman101:
> Thinking about it now, we actually could limit it to monitor Hz and apply higher than 88 FPS limit only if it would make sense.

yes, I agree that's the best approach

Olya-Marinova:
I think this can be implemented in CMainMenu.cpp

Dutchman101:
> Thinking about it now, we actually could limit it to monitor Hz and apply higher than 88 FPS limit only if it would make sense.

@lopezloo If you are ready to do that, please open a new PR

PR #3036, opened 2023-05-24
Render cursor in current frame

lopezloo:
Currently mouse cursor seems to be queued for render in the next frame. This change disables render queueing for the mouse cursor thus it renders in the current frame now.

https://github.com/multitheftauto/mtasa-blue/blob/76d350ba29d3416c8af21dc21c270664ea66f33b/Client/core/DXHook/CDirect3DEvents9.cpp#L163-L174


botder (approved):
I hope this doesn't have any significant negative impact on performance.

Issue #3022, opened 2023-05-21
Feature: add constants to add more meaning to magic numbers


CrosRoad95:

Fernando-A-Rocha:
Reopen ?
Lua constants would be nice to have
like enums in Roblox Lua

PR #3039, opened 2023-05-24
New Lua constants (#3022)

TracerDS:
This PR is a direct response to the #3022 feature which adds new global variable: [`constants`](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-249795227c2ee9f3fb91954f46dd7537e4850ec38ec61eb896d531010bdb87f6R10) and assigns new fields to it:

- [ElementType](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-48a934fdd631a5bdfaae26e47688253e269075ff084aeeb96d8ede34eb94b253R10)
- [TriggerPriority](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R10)
- [DebugMessageLevels](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R15)
- [KeyState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R22)
- [KeyNames](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R27)
- [ControlNames](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R135)
- [AnimationBlocks](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R182)
- [BodyPartName](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R312)
- [ClothesType](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R321)
- [Weather](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-f1f884dc24bc7971313a71f0291a353aa6c7e2029098e350d20b44d4fd2d3388R341)
- [PedStats](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-61dbc1054ae477c980adf2ad12c03a2e43a0e6bc63bae666cadb39ef03d399a8R10)
- [PedWalkingStyle](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-61dbc1054ae477c980adf2ad12c03a2e43a0e6bc63bae666cadb39ef03d399a8R174)
- [PedFightingStyle](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-61dbc1054ae477c980adf2ad12c03a2e43a0e6bc63bae666cadb39ef03d399a8R215)
- [HudComponents](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-7b0eaeb61603a5c36e06de322488bd0191b2c5cc829ef14aefab30edead36422R10)
- [BlipIcons](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-7b0eaeb61603a5c36e06de322488bd0191b2c5cc829ef14aefab30edead36422R26)
- [MarkerTypes](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-7b0eaeb61603a5c36e06de322488bd0191b2c5cc829ef14aefab30edead36422R92)
- [PickupTypes](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-7b0eaeb61603a5c36e06de322488bd0191b2c5cc829ef14aefab30edead36422R99)
- [ModelPickupTypes](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-7b0eaeb61603a5c36e06de322488bd0191b2c5cc829ef14aefab30edead36422R105)
- [VehicleLights](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R10)
- [VehicleLightState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R16)
- [VehicleDoorType](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R20)
- [VehicleDoorState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R28)
- [VehicleLightOverride](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R35)
- [CarPanels](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R40)
- [PlanePanels](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R49)
- [VehiclePanelState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R58)
- [VehicleWheelState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R64)
- [VehicleSeat](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R70)
- [TrafficLightState](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-2f1691532679512edca1e3260f82f714d5ed70ff6277c7bc945e7e27fa8b2aa8R76)
- [WeaponSlot](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-bc32986768f32997595693b2a6fdee7e51f49f9a9fc30fd38c15372e3babca63R10)
- [WeaponSkillLevel](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-bc32986768f32997595693b2a6fdee7e51f49f9a9fc30fd38c15372e3babca63R25)
- [WeaponSkillName](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-bc32986768f32997595693b2a6fdee7e51f49f9a9fc30fd38c15372e3babca63R30)
- [WeaponProperties](https://github.com/multitheftauto/mtasa-blue/pull/3039/files#diff-bc32986768f32997595693b2a6fdee7e51f49f9a9fc30fd38c15372e3babca63R35)


tederis:
I think that the length of these constants makes them cumbersome and difficult to use. I doubt that long constructions like `constants.VehicleLightOverride.Disable` can be usefull. I would suggest to make them shorter.

tederis:
In addition, I would suggest to make the table read-only.

CrosRoad95 (commented):
Fine for me, just maybe other people may have concern about names, but apart of that it is good to go

Shared/mods/deathmatch/logic/luascripts/constants_other.lua.h
```diff
@@ -0,0 +1,378 @@
...
+    VehicleLookRight = 'vehicle_look_right',
+    VehicleLookBehind = 'vehicle_look_behind',
+    VehicleMouseLook = 'vehicle_mouse_look',
+    SpecialControlLeft = 'special_control_left',
+    SpecialControlRight = 'special_control_right',
+    SpecialControlDown = 'special_control_down',
+    SpecialControlUp = 'special_control_up',
+    EnterExit = 'enter_exit',
+}
+local _animationBlocks = {
```
Pirulax: why is the casing different here that in the other enums?

Pirulax (changes requested):
Also, whats the point of not directly assigning the values to `constants.*` but instead making a `local` var first?
If there's no good reason, then we should just assign it directly [this also reduces the chance of interfering with existing scripts]

Also, there's a chance that some scripts might break if these constants are loaded into every script, we should have a `meta.xml` toggle for it, that toggles this feature per-resource [as theres one LuaVM per resource]

Also... I'm not sure if `constants` is the solution.
Perhaps we should just push them into the global namespace? [Like `KeyStates.Up` rather than `constants.KeyStates.up`.]
[I'm sure some people would prefer the one over the other]

And, as @tederis noted, making the tables read only might be useful [though Lua is somewhat limited in that regard]

Shared/mods/deathmatch/logic/luascripts/constants_other.lua.h
TracerDS: ~~Its more readable to create a local variable than to have everything in a static table. And it wont interfere with anything~~

I misunderstood. Thats a better approach

Pirulax:
Also something I just realized, these constants are pretty useless on their own.
There should be a way to convert to/from name/value.
Given:
```lua
local MyEnum = {
    First = 1
}
```
`MyEnum['First']` should return `1`. `MyEnuim(1)` should return `First`
This is similar to how Python work (Though Python returns special EnumValue objects that we can't afford in terms of performance, so we're going with the simpler solution here).

TracerDS:
> I think that the length of these constants makes them cumbersome and difficult to use. I doubt that long constructions like `constants.VehicleLightOverride.Disable` can be usefull. I would suggest to make them shorter.

What would we rename them to? 🤔

> Also something I just realized, these constants are pretty useless on their own. There should be a way to convert to/from name/value. Given:
> 
> ```lua
> local MyEnum = {
>     First = 1
> }
> ```
> 
> `MyEnum['First']` should return `1`. `MyEnuim(1)` should return `First` This is similar to how Python work (Though Python returns special EnumValue objects that we can't afford in terms of performance, so we're going with the simpler solution here).

Implemented. Thanks for the suggestion ❤

Shared/mods/deathmatch/logic/luascripts/constants_main.lua.h
```diff
@@ -0,0 +1,37 @@
+namespace EmbeddedLuaCode
+{
+    const char* const constantsMain = R"~LUA~(
+--[[
+    SERVER AND CLIENT.
+    Defines a constant variables available for server and client.
+--]]
+
+metatable = {
```
TheNormalnij: Name it "readonly_table_meta" or something like this

Fernando-A-Rocha (approved):

TracerDS:

PR #3057, opened 2023-06-04
Update ZIP vendor

TracerDS:
Main purpose of this PR is to move all zip functionality into one shared class with easy to use API.
The old vendor is... not easy to use


Server/mods/deathmatch/utils/CZipMaker.cpp
```diff
@@ -66,14 +88,23 @@ bool CZipMaker::IsValid()
...
+bool CZipMaker::Close() noexcept
 {
-    if (!m_uzFile)
+    if (!IsValid())
         return false;
 
-    int iResult = zipClose(m_uzFile, NULL);
-    m_uzFile = NULL;
-    return iResult == ZIP_OK;
+    zip_close(m_uzFile);
```
tederis, line 101: m_uzFile is not nulled after the file closing

Server/mods/deathmatch/utils/CZipMaker.h
```diff
@@ -9,19 +9,66 @@
  *****************************************************************************/
 #pragma once
 
+#include <zip/zip.h>
+#include <memory>
+
 class CZipMaker
 {
+    struct ZipCloser
```
tederis: Unused struct

Server/mods/deathmatch/utils/CZipMaker.cpp
```diff
@@ -133,35 +164,112 @@ bool CZipMaker::InsertDirectoryTree(const SString& strInSrc, const SString& strI
...
+    if (offset > size)
+        return entry;
+
+    if (offset < 0)
+        offset += size;
+
+    if (zip_entry_openbyindex(m_uzFile, offset))
+        return entry;
+
+    entry.name = zip_entry_name(m_uzFile);
```
tederis, line 274: It is unsafe to store a pointer to a string this way. When zip file get closed all these strings become invalid. And I can see that it is already a problem in here since m_vecEntries won't be cleared when ::Close called.

Server/mods/deathmatch/utils/CZipMaker.cpp
```diff
@@ -133,35 +164,112 @@ bool CZipMaker::InsertDirectoryTree(const SString& strInSrc, const SString& strI
...
+{
+    CZipEntry entry;
+    if (!m_uzFile)
+        return entry;
+
+    const auto size = zip_entries_total(m_uzFile);
+    if (offset > size)
+        return entry;
+
+    if (offset < 0)
```
tederis: offset is unsigned and cannot be less then zero. Thus, this expression makes no sense.

Server/mods/deathmatch/utils/CZipMaker.cpp
```diff
@@ -133,35 +164,112 @@ bool CZipMaker::InsertDirectoryTree(const SString& strInSrc, const SString& strI
...
+{
+    if (!m_uzFile)
+        return false;
 
-    int iResult = zipOpenNewFileInZip(m_uzFile, strDest, &zfi, NULL, 0, NULL, 0, NULL, Z_DEFLATED, Z_DEFAULT_COMPRESSION);
+    return zip_extract(m_strZipPath, strDirPath, NULL, NULL) == 0;
+}
 
-    if (iResult == ZIP_OK)
+std::vector<CZipMaker::CZipEntry> CZipMaker::ListEntries() const noexcept
```
tederis, line 247: It is worth to check m_uzFile in this function.

Server/mods/deathmatch/logic/CZipFile.cpp
```diff
@@ -0,0 +1,66 @@
...
+    m_strFilename = szFilename ? szFilename : "";
+    m_pResource = nullptr;
+}
+
+CZipFile::CZipFile(const uint uiScriptId, const char* szFilename, const CZipMaker::Mode mode) : CElement(NULL)
+{
+    Init(uiScriptId, szFilename);
+    m_zipMaker = CZipMaker(szFilename, mode);
+}
+CZipFile::CZipFile(const uint uiScriptId, const char* szFilename, const char mode) : CElement(NULL)
```
tederis: It's better to construct m_zipMaker within a member initializer list

Server/mods/deathmatch/logic/CResourceChecker.cpp
```diff
@@ -840,297 +840,26 @@ bool CResourceChecker::RenameBackupFile(const string& strOrigFilename, const str
...
-    unz_global_info glob_info;
-    if (unzGetGlobalInfo(szip, &glob_info) != UNZ_OK)
-    {
-        zipClose(dzip, NULL);
-        unzClose(szip); /*free(tmp_name);*/
-        return 0;
-    }
+    CZipMaker srcZip(strSrcZip.c_str(), CZipMaker::Mode::READ);
+    if (!srcZip)
+        return false && srcZip.Close();
```
tederis: What is going on here? What is the purpose of this ugly logical construction?

Shared/sdk/version.h
```diff
@@ -28,8 +28,8 @@
 #define MTASA_VERSION_MAJOR         1
 #define MTASA_VERSION_MINOR         6
 #define MTASA_VERSION_MAINTENANCE   0
-#define MTASA_VERSION_TYPE          VERSION_TYPE_UNSTABLE
-#define MTASA_VERSION_BUILD         0
+#define MTASA_VERSION_TYPE          VERSION_TYPE_CUSTOM
+#define MTASA_VERSION_BUILD         21832
```
tederis: I suppose, it was mistakenly left

Server/mods/deathmatch/logic/luadefs/CLuaZipDefs.cpp
```diff
@@ -0,0 +1,195 @@
...
+    pZip->Close();
+    m_pElementDeleter->Delete(pZip);
+
+    // Success. Return true
+    lua_pushboolean(luaVM, true);
+    return 1;
+}
+
+/*
+bool CLuaZipDefs::zipClose(CZipFile* pZip)
```
tederis: Commented out code is not encouraged

tederis:
I don't have time to go through all the checks but I think that CZipFile won't be deleted after the resource stop. On the contrary, fileOpen creates an element that will be destroyed as resource get stopped. I think you should take a look at how default file functions are implemented and follow the same idea.

Server/mods/deathmatch/utils/CZipMaker.cpp
TracerDS: It is nulled internally

TracerDS:
> I don't have time to go through all the checks but I think that CZipFile won't be deleted after the resource stop. On the contrary, fileOpen creates an element that will be destroyed as resource get stopped. I think you should take a look at how default file functions are implemented and follow the same idea.

I did try following it but due to how mta is designed i couldnt follow up with the implementation ;/

Server/mods/deathmatch/utils/CZipMaker.cpp
tederis: > It is nulled internally

How? CZipMaker::m_uzFile does not change in this function.

TracerDS: zipClose nulls it internally

tederis: zip_close does not change the m_uzFile. It is simply impossible because zip_close obtains an argument by value, not by a pointer of a reference. But even if that would be possible it is a bad design allows dependent objects modify internal states of a parent.

PR #3086, opened 2023-06-25
Functions to control the streaming memory/buffer sizes [Fixes #3084]

### LINKED ISSUE #3065: Using IMG feature causes map loading issue
### Describe the bug

After started using .img feature, map doesn't load like before. 1.5.9 is working fine, 1.6 too with old script, but when we load one .img map will be blinking.
https://streamable.com/umrhmt

### Steps to reproduce

Use custom img (~700mb)

### Version

Multi Theft Auto v1.6-release-21834

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

### LINKED ISSUE #3084: Increase streaming memory

MTA currently have maximum 256mb streaming memory, that's not enough if you want to use new features as well. (Like: https://github.com/multitheftauto/mtasa-blue/issues/3065)
If you load eg. an img which is around 200-250mb map will be blinking if streaming memory is on full, **so new IMG feature is useless now**.

Something like engineSetStreamingMemory( [megabytes] 256-768 )
768mb should be enough for all servers, which have eg. large model pack (like mentioned in #3065)

I've tested with this, and have no issues (original discussion: https://github.com/multitheftauto/mtasa-blue/pull/2264):
image

_No response_

### Additional context

_No response_

### Security Policy

Pirulax:
This PR aims to implement the function as discussed [here](https://github.com/multitheftauto/mtasa-blue/pull/2323#issuecomment-1602088547)

- [x] `engineStreamingSetMemorySize`
  - All positive values set the streaming size [=the maximum memory that can be used for models] to the value given [in Bytes] 
- [x] `engineStreamingGetMemorySize`
    - Get the current streaming memory size [in Bytes]
- [x] `engineStreamingRestoreMemorySize`
    - Restore the streaming memory size to the cvar setting [Usually 256 MB]
- [ ] `engineStreamingSetBufferSize`
    - Set the streaming buffer size - The larger it is, the more models can be loaded in one go [Though, the limiting factor is still the GTA streamer] - Can help a bit with custom IMGs loading speed [reduces pop-in]
- [x] `engineStreamingGetBufferSize`
    - Getter for the above...

Also added a new class `EngineStreaming` for all streaming-related functions, and 2 new class variables:
- `memorySize` - The streaming memory size [in bytes] - Can also be set [uses `engineStreamingSetMemorySize` interally]
- `usedMemory` - Amount of memory used right now for streaming

Example usage [and unit test]:
```lua
local initialSz = EngineStreaming.memorySize

EngineStreaming.memorySize = 1024 * 1024 * 1024
assert(EngineStreaming.memorySize == 1024 * 1024 * 1024)

EngineStreaming.restoreMemorySize()
assert(EngineStreaming.memorySize == initialSz)

assert(not EngineStreaming.setBufferSize(0)) -- Should fail, as the buffer size is too small
assert(EngineStreaming.setBufferSize(50 * 1024 * 1024)) -- Should not fail, this is a big enough buffer for the default gta3.img files
```

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -2318,19 +2318,22 @@ SString CCore::GetBlueCopyrightString()
 }
 
 // Set streaming memory size override [See `engineStreamingSetMemorySize`]
-void CCore::SetCustomStreamingMemory(size_t szMB) {
-    m_CustomStreamingMemoryLimitMB = szMB;
+// Use `0` to turn it off, and thus restore the value to the `cvar` setting
+void CCore::SetCustomStreamingMemory(size_t sizeBytes) {
+    // NOTE: The override is applied to the game in `CClientGame::DoPulsePostFrame`
+    // There's no specific reason we couldn't do it here, but we wont
+    m_CustomStreamingMemoryLimitBytes = sizeBytes;
 }
 
 bool CCore::IsUsingCustomStreamingMemorySize()
 {
-    return m_CustomStreamingMemoryLimitMB != 0;
+    return m_CustomStreamingMemoryLimitBytes != 0;
 }
 
-// Streaming memory size used [In MB]
+// Streaming memory size used [In Bytes]
 size_t CCore::GetStreamingMemory()
 {
     return IsUsingCustomStreamingMemorySize()
-        ? m_CustomStreamingMemoryLimitMB
-        : CVARS_GET_VALUE<size_t>("streaming_memory", 0);
+        ? m_CustomStreamingMemoryLimitBytes
+        : CVARS_GET_VALUE<size_t>("streaming_memory") * 1024 * 1024; // MB to B conversion
 }
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -2316,3 +2316,27 @@ SString CCore::GetBlueCopyrightString()
     SString strCopyright = BLUE_COPYRIGHT_STRING;
     return strCopyright.Replace("%BUILD_YEAR%", std::to_string(BUILD_YEAR).c_str());
 }
+
+// Set streaming memory size override [See `engineStreamingSetMemorySize`]
+void CCore::SetCustomStreamingMemory(size_t szMB) {
+    m_CustomStreamingMemoryLimitMB = szMB;
+}
+
+bool CCore::IsUsingCustomStreamingMemorySize()
+{
+    return m_CustomStreamingMemoryLimitMB != 0;
+}
+
+// Streaming memory size used [In MB]
+size_t CCore::GetStreamingMemory()
+{
+    // If custom override is set, use that
+    if (IsUsingCustomStreamingMemorySize()) {
+        return m_CustomStreamingMemoryLimitMB;
+    }
+
+    // Otherwise client CVar setting
+    size_t streamingMemSzCVar;
+    CVARS_GET_VALUE("streaming_memory", streamingMemSzCVar);
+    return streamingMemSzCVar;
+}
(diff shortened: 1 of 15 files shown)


TheNormalnij:
Some notices:
* There is no error/waring when you call
```lua
engineStreamingSetMemorySize(9000 * 1024 * 1024)
-- or
engineStreamingSetMemorySize(-1)
```

* GTA will not free memory immediately after call engineStreamingSetMemorySize (should be in doc)

Pirulax:
That it something the argparser should do, not me.
I did make a [PR] (https://github.com/multitheftauto/mtasa-blue/pull/2114), but nobody cared to review it for 2 years :D
Also, the streamer will [or, well, should anyways] free up memory, as it won't be allowed to load a new model before freeing up memory [in case it doesn't have enough memory everything will start flickering]

CrosRoad95:
"That it something the argparser should do, not me."
sounds like a bug, size_t is unsigned but probably it hasn't been checked while converting from lua_Number to size_t, but still you have variable upper bound, if someone have X GB of vram, you probably should not be able to set max streaming memory to more than that, i don't know what can happen ( maybe crash ) if you do this, probably gta itself care of it.

Pirulax:
> sounds like a bug, size_t is unsigned but probably it hasn't been checked while converting from lua_Number to size_t

That's exactly what #2114 aims to fix.

> ...but still you have variable upper bound...

You might say it's the amount of RAM the PC has - But it isn't as the OS can just use the swap file [slow, but would still work].
Also, this is an advanced feature, I expect scripters who use it to be smart about it [and not try to set it to a trillion gigabytes]

CrosRoad95:
wonder what happen if you set max memory stream to 0 😂

Pirulax:
Whole map starts flickering, nothing interesting

TheNormalnij:
> Some notices:
> 
> * There is no error/waring when you call
> 
> ```lua
> engineStreamingSetMemorySize(9000 * 1024 * 1024)
> -- or
> engineStreamingSetMemorySize(-1)
> ```
> 
> * GTA will not free memory immediately after call engineStreamingSetMemorySize (should be in doc)

This is MTA issue. We don't perform this checks in new argument parser now. Can be fixed in next backward incompatible  release via #2114. So this should not block this PR.

TheNormalnij (approved):

Pirulax:
NOTE: Wiki pages need to be added

Dutchman101:
> NOTE: Wiki pages need to be added

@Pirulax I hope you can add these functions to the wiki this soon, it's actually a responsibility of the feature's contributor and should have been checked off (or with a clear plan to work on it immediately) around merge time.

If not done, it will be a waste as usage of these functions will be minimal. There is even a chance of revert if this requirement can't be checked off on.

Pirulax:
I've added the pages. Please check them out, and let me know if there are any errors:
[engineStreamingSetMemorySize](https://wiki.multitheftauto.com/wiki/engineStreamingSetMemorySize)
[engineStreamingGetMemorySize](https://wiki.multitheftauto.com/wiki/engineStreamingGetMemorySize)
[engineStreamingRestoreMemorySize](https://wiki.multitheftauto.com/wiki/engineStreamingRestoreMemorySize)
[engineStreamingGetBufferSize](https://wiki.multitheftauto.com/wiki/engineStreamingGetBufferSize)
[engineStreamingSetBufferSize](https://wiki.multitheftauto.com/wiki/engineStreamingSetBufferSize)

Also, it seems like I didn't add a `engineStreamingRestoreBufferSize`  function [nor did I restore it on disconnect].
I'll add it tomorrow.

PR #3090, opened 2023-06-29
Shortening of the game_sa project code

Moris-Onz:
My little fixes in the game_sa project that I think are necessary


sbx320:
I don't really see a reason why this would be useful at all. It's a bunch of questionable type renamings and a major loss of information due to losing function names in #defines

PR #3092, opened 2023-06-30
Workaround Fix #1326: Screen flickering on high memory usage

### LINKED ISSUE #1326: Visual glitches on high memory usage
**Describe the bug**
https://streamable.com/vlgd1

I don't know what causes this, nor how it starts. There are just a few things I know about it. This already exists for years, moving camera can remove/restore the flickering. Based on this camera thing I believe that the cause could be something 3D related like a vehicle or a ped/player. Not sure if modded elements affect this flickering.

Looking more thoroughly and carefully through the video, before i start flying to that part of SF Airport, you can notice a black thing as possible source, and as i fly towards, only fitting element is a player (LOKO) flying with superman using default CJ model with some clothes from GTA. 

Actually looking even more closer, the flickering moved the same way a player with superman script would, suggesting that it's probably coming from that player.

**To reproduce**
Unknown

**Expected behaviour**
To not happen

**Screenshots**
Provided a video above in the description

**Version**
Any

Zangomangu:


Pirulax:
Fixing the bug instead of this would be a better solution imo.

Zangomangu:
Good point, let's wait for someone to do that.

lopezloo:
What results does this patch produce? It eliminates flickering but holes in peds still occur since it skips rendering corrupted matrices, right?

Zangomangu:
yeah a ped will be missing a limb for 1 frame, which is what you would otherwise see stretched across your screen as a visual glitch

lopezloo:
Could you dump screenshot exactly when this bug happen to see how it looks? What if instead skipping rendering, we would set corrupted matrice value to 0?

Pirulax:
I'll reopen this, for one reason: Waiting for someone to fix this bug would take a while. So let's merge this, and later fix the actual bug.

Zangomangu:
> Could you dump screenshot exactly when this bug happen to see how it looks? What if instead skipping rendering, we would set corrupted matrice value to 0?

it looks like this - it's easy to reproduce, check the stuff posted under the issue.

image

lopezloo (approved):
While it's not proper fix, it's better than current state. I'm merging it then. Thank you.

PR #3093, opened 2023-07-02
Refactor vehicle logic

Moris-Onz:
Refactor vehicle logic, corrections


Pirulax (changes requested):
While I approve most of the changes, I don't think there's any pros to use `ushort` over `ulong`, so please revert those changes.

CrosRoad95:
take a look:
```cpp
    static eVehicleType GetVehicleType(unsigned short usModel);
    static unsigned int GetMaxPassengers(unsigned int uiVehicleModel);
```
in both cases we are dealing with same id, id of vehicle, in some cases it is okey to have range 0-65k, in other not? we should agree that "ushort" is type of model id no matter what.

Moris-Onz:
What exactly is wrong here?

> Хотя я одобряю большинство изменений, я не думаю, что есть какие-либо плюсы для использования `ushort` вместо `ulong`, поэтому, пожалуйста, отмените эти изменения.

sbx320 (changes requested):
I don't really see any point in changing ´unsigned char/short` to `uchar/ushort` to be honest...

Moris-Onz:
> Честно говоря, я действительно не вижу никакого смысла в изменении unsigned char / short`to`uchar /ushort`...

Source code reduction, this is the same as unsigned char and unsigned short, there is nothing wrong or questionable

multimaks2:
what is the problem with accepting this branch if this type of data allows you to expand the transport ID?

Moris-Onz:
Why was it closed?

sbx320:
See above. "Source code reduction" isn't a useful thing if it's just bulk renaming some types without reducing code complexity.

Moris-Onz:
> Смотрите выше. "Сокращение исходного кода" бесполезно, если это просто массовое переименование некоторых типов без уменьшения сложности кода.

Here I removed the additional ranges of the car using the IsStandardModel function, and removed the additional functions with the old gta ranges

Moris-Onz:
The code has become easier, why cancel it?

Moris-Onz:
The refusal is very controversial and unpleasant

PR #3094, opened 2023-07-02
Update mtaserver.conf.template

efexc:
I've edited it, including the recent changes (pr #3078) and also added scripts that will start automatically.

### DIFF
--- a/Server/mods/deathmatch/mtaserver.conf.template
+++ b/Server/mods/deathmatch/mtaserver.conf.template
@@ -264,5 +264,34 @@ R"=====(
          *NOTE* This only protects resources which use dbConnect with mysql
          Values: 0 - Off, 1 - Enabled.  Default - 1 -->
     <database_credentials_protection>1</database_credentials_protection>
+<!-- Specifies the module(s) which are loaded with the server. To load several modules, add more <module>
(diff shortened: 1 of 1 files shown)


sbx320:
The template is used to fill missing defaults in existing server configs (please don't ask me why this is done in such an odd manner...). 

However resources shouldn't be automatically added to a server config, as those resources might not even be present in the local server installation.

TracerDS:
> However resources shouldn't be automatically added to a server config, as those resources might not even be present in the local server installation.

Not true. All default resources are present if you install mta server. They are also added to server's config regardless if you chose to install them or not.

sbx320:
Yes. However a server owner may decide to not install any resources or remove default resources (and remove them from the mtaserver.conf as well). 

With the change from this PR, the resources would be readded to the mtaserver.conf, which would cause the server to fail.

TracerDS:
Adding resources to config doesnt cause the server to crash or shutdown. It only alerts about lack of resources but you can continue with it.
And the only way to not include resources during install is by building server from source and running property bat file. But thats obviously out of scope.
If the owner doesnt want any resources then there is no issue with removing lines from mtaserver.conf file, if he cares about the messages that is

PR #3099, opened 2023-07-03
Add texture hit info to `processLineOfSight`

### LINKED ISSUE #645: ProcessLineOfSight returns with UV Coords
The function ProcessLineOfSight should return with the elements UV coordinate where the hit was detected. It would be very helpful when using shaders and render targets. Of course, that will only work with correct UV Maps.

Pirulax:
Fixes #645 

This PR aims to add texture related info to hits [Such as hit texture UV, and name], and a more precise "clump hit" position [that is the intersection point of the clump and the line segment]
It uses the hit entity's clump to do so [As it contains all the necessary UV information]

### DIFF
--- a/Client/game_sa/CCollisionSA.cpp
+++ b/Client/game_sa/CCollisionSA.cpp
@@ -1,3 +1,14 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        game_sa/CCollisionSA.cpp
+ *  PURPOSE:     Implementation of `CCollision` - collision detection related functions
+ *
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
 #include "StdInc.h"
 
 #include "CCollisionSA.h"
--- a/Client/game_sa/CCollisionSA.cpp
+++ b/Client/game_sa/CCollisionSA.cpp
@@ -1,0 +1,7 @@
+#include "StdInc.h"
+
+#include "CCollisionSA.h"
+
+bool CCollisionSA::TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) {
+    return reinterpret_cast<bool(__cdecl*)(const CColLineSA&, const CColSphereSA&)>(0x417470)(line, sphere);
+}
--- a/Client/game_sa/CCollisionSA.cpp
+++ b/Client/game_sa/CCollisionSA.cpp
@@ -2,6 +2,7 @@
 
 #include "CCollisionSA.h"
 
-bool CCollisionSA::TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) {
+bool CCollisionSA::TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) const
+{
     return reinterpret_cast<bool(__cdecl*)(const CColLineSA&, const CColSphereSA&)>(0x417470)(line, sphere);
 }
--- a/Client/game_sa/CCollisionSA.h
+++ b/Client/game_sa/CCollisionSA.h
@@ -1,0 +1,8 @@
+#pragma once
+
+#include "game/CCollision.h"
+
+class CCollisionSA : CCollision {
+public:
+    bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) override;
+};
--- a/Client/game_sa/CCollisionSA.h
+++ b/Client/game_sa/CCollisionSA.h
@@ -4,5 +4,5 @@
 
 class CCollisionSA : CCollision {
 public:
-    bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) override;
+    bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) const override;
 };
--- a/Client/game_sa/CCollisionSA.h
+++ b/Client/game_sa/CCollisionSA.h
@@ -1,3 +1,14 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        game_sa/CCollisionSA.h
+ *  PURPOSE:     Header file for `CCollision` - collision detection related functions
+ *
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
 #pragma once
 
 #include "game/CCollision.h"
--- a/Client/game_sa/CWorldSA.cpp
+++ b/Client/game_sa/CWorldSA.cpp
@@ -382,8 +382,6 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
         c.minHitDistSq = c.dirOS.LengthSquared();            // By setting it to this value we avoid collisions that would be detected after the line segment
                                                              // [but are still ont the ray]
 
-        // const auto lineMagSq = (ctx.originOS - ctx.endOS).LengthSquared();
-
         // Do raycast against the DFF to get hit position material UV and name
         // This is very slow
         // Perhaps we could paralellize it somehow? [OpenMP?]
@@ -469,7 +467,7 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
             RpClumpForAllAtomics(targetEntity->m_pRwObject, ProcessOneAtomic, &c);
         }
         else
-        {            // Object is an atomic, so process directly
+        {            // Object is a single atomic, so process directly
             ProcessOneAtomic(reinterpret_cast<RpAtomic*>(targetEntity->m_pRwObject), &c);
         }
 
--- a/Client/game_sa/CWorldSA.cpp
+++ b/Client/game_sa/CWorldSA.cpp
@@ -15,6 +15,9 @@
 #include "CGameSA.h"
 #include "CPoolsSA.h"
 #include "CWorldSA.h"
+#include "CColModelSA.h"
+#include "gamesa_renderware.h"
+#include "CCollisionSA.h"
 
 extern CCoreInterface* g_pCore;
 extern CGameSA*        pGame;
@@ -252,7 +255,7 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
 }
 
 bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd, CColPoint** colCollision, CEntity** CollisionEntity,
-                                  const SLineOfSightFlags flags, SLineOfSightBuildingResult* pBuildingResult)
+                                  const SLineOfSightFlags flags, SLineOfSightBuildingResult* pBuildingResult, SProcessLineOfSightMaterialInfoResult* outMatInfo)
 {
     DWORD dwPadding[100];            // stops the function missbehaving and overwriting the return address
     dwPadding[0] = 0;                // prevent the warning and eventual compiler optimizations from removing it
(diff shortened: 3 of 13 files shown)


CrosRoad95:
Take an account that often collision is not the same as model ( as documentation note ) what mean that if i raycast i can get texture of for example collisionless leaves.

What if for example there is no collision hit, but raycast went through some collisionless mesh?
and does it work for vehicles, peds?

tederis:
It would be nice to get normals as well. Because a collision does not always match the actual surface of a mesh and normals can point in a completely different direction. Sometimes it's important to have "real" normals.

CrosRoad95:
Dies it work for rotated objects? And maybe its better to make standalone function to raycast mesh rather bloat process line of sight?

Pirulax:
Yeah, perhaps adding a new function would make more sense, to give the scripter more freedom [And room to optimize], in case they want to do a raycasting against a given entity.
What would you suggest as the name?

The return arguments would be:
- hit pos x, y, z
- hit normal x, y, z
- texture U, V
- texture name
- atomic index
- geometry index
- triangle index

Last three I'm not sure about, but one might find them useful, so might as well add it.

Pirulax:
> Take an account that often collision is not the same as model ( as documentation note ) what mean that if i raycast i can get texture of for example collisionless leaves.
> 
> What if for example there is no collision hit, but raycast went through some collisionless mesh? and does it work for vehicles, peds?

Good point.
Sadly, a simple raycast against a given entity is all I need.
This can later be extended as necessary.

> Dies it work for rotated objects? And maybe its better to make standalone function to raycast mesh rather bloat process line of sight?

Yes it does work for rotated objects too, as it uses the entity's inverse matrix to transform world position points into object [model] space.

CrosRoad95:
```
atomic index
geometry index
```
this is i'm working on, to get dff atomics, geometry ect. now if you add index instead of userdata what i'm planning, it may cause compatibility problem on my or your side.

I think it should be separete function, raycast against atomic/clump(dff)

Pirulax:
raycast agianst a clump
are you adding clump as a userdata too?

tederis:
The problem is now you are doing the work twice: raycast against a collision mesh and DFF geometry after it. There is a lot of unnecessary heavy calculations. I think you should get rid of a standart collision processing part and write a simple ray casting procedure against AABBs. It's not as difficult as it might seem, because this part of game is already reversed and most of the code can be reused.

Pirulax:
I guess with enough parallelization the mesh raycast will be ultimately faster than the mesh one
Thus, in the end we could actually improve the performance of PLOS [and isLineOfSightClear too]
But, since all this is in the `game_sa` project I don't have easy access to any threading facilities [AsyncScheduler for example].
The best way to deal with this would be OpenMP, but we don't have it enabled [though, I guess we could enable it?]
Anyways, I find it out-of-scope for now, as for my needs this is enough.

And yeah, writing a simple line/triangle collision test function is quite easy at this point [and I do have experience with it too]

As for the function to raycast against a specific element ( @CrosRoad95 ) instead of adding it to PLOS.
If this wouldn't be available in PLOS you couldn't process native GTA world enetities [ones without an MTA representation].
So, I think, instead of *moving* it, we could make a separate function too + keep in in PLOS

Also, I ran into a problem.
Kam's scripts seem to export incorrectly, as they don't set the triangle material ID [in the geometry], but rather just use the binmesh extension [plugin] to store it. This is an issue, because the data stored in `RpMesh` is optimized for *rendering*, not for collision detection.
This causes the function to return incorrect texture names. UV values are fine though.

Pirulax:
Okay, so it's *kinda* working.
Regular objects work just fine, and vehicles do somewhat too.
The issue is that vehicle components [such as doors, the bonnet, etc] seem to not exist? [[Demo](https://youtu.be/WPSC_OcJ8XQ)]
Notice how the ray doesn't seem to hit the door at all, but instead something else behind it.
I'm not sure if it's because that atomic [the door] isn't part of the vehicles clump or something else.

tederis:
> Okay, so it's _kinda_ working. Regular objects work just fine, and vehicles do somewhat too. The issue is that vehicle components [such as doors, the bonnet, etc] seem to not exist? [[Demo](https://youtu.be/WPSC_OcJ8XQ)] Notice how the ray doesn't seem to hit the door at all, but instead something else behind it. I'm not sure if it's because that atomic [the door] isn't part of the vehicles clump or something else.

Maybe there is something that I overlooked, but I can't find a transformation of atomics in your code. It seems like all parts(except the chassis) are in the zero position(or in the vehicle's origin). This may very well be the cause of this erroneous behavior.

Pirulax:
Yeah, that would explain the issue, since all frames can have a different transform.
I will look into it, thank you.
Only that I'm not sure how the transformations are supposed to be applied? Do they "stack" [like bone transformations], or applying the frame transform [LTM] to the position is enough?

Okay, so, the transforms *were* missing indeed, but the situation is still pretty much the same.
Issue is that I'm unsure how doors [and other components] are rendered.
Pretty sure each one is just a frame in the clump [and I'm also pretty sure it's the `_dummy` frames], but even when a dummy frame is hit, the texture info is completely off [and when hovering the mouse over a door, noit the whole door has a collision]...

Pirulax:
@tederis you're a genius dude, ly
https://www.youtube.com/watch?v=M5FLVLkWcOo

Client/sdk/game/CWorld.h
```diff
@@ -51,6 +51,14 @@ struct SLineOfSightBuildingResult
     CEntitySAInterface* pInterface;
 };
 
+struct SProcessLineOfSightMaterialInfoResult {
+    CVector2D   uv;            // On-texture UV coordinates of the intersection point
+    const char* textureName;   // GTA texture name
+    const char* frameName;     // The name of the frame the hit geometry belongs to
+    CVector     hitPos;        // Precise hit position on the clump [World space]
+    bool        valid{};       // Data found in this struct is only valid if this is `true`!
```
botder: Please use `//<` for the documentation strings of member fields.

Client/sdk/game/RenderWare.h
```diff
@@ -321,25 +321,35 @@ struct RwGeometry
...
-    unsigned int unknown2;
-    float        unknown3;
-    float        unknown4;
-    float        unknown5;
+    int32 flags; /**< flags */
+    int16 startMorphTarget; /**< startMorphTarget */
+    int16 endMorphTarget; /**< endMorphTarget */
+    float time; /**< time */
+    float recipTime; /**< recipTime */
+    float position; /**< position */
```
botder: Please use `//<` for the documentation strings of member fields.
Also: use fixed size integer types with the `_t` suffix.

Shared/sdk/CVector.h
```diff
@@ -261,4 +267,6 @@ class CVector
     }
 
     bool operator!=(const CVector& param) const noexcept { return !(*this == param); }
+
+    float operator[](size_t i) const noexcept { return ((float*)this)[i]; }
```
botder, line 271:
```suggestion
    float operator[](size_t i) const { return ((float*)this)[i]; }
```

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -286,50 +289,70 @@ int CLuaWorldDefs::ProcessLineOfSight(lua_State* luaVM)
...
+                    lua::Push(L, matInfo.hitPos.fY);
+                    lua::Push(L, matInfo.hitPos.fZ);
+                } else {
+                    for (auto i = 2 + 1 + 1 + 3; i-- > 0;) {
+                        lua_pushnil(L);
+                    }
                 }
-                return 11;
+
+                return 11 + 8 + 7;
```
botder, line 349: Maybe use variables instead? I don't expect anyone to change this code in the foreseeable time, but this seems odd.

Client/game_sa/CWorldSA.cpp
```diff
@@ -350,6 +388,157 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
...
+            const auto localEnd = ObjectToLocalSpace(c->endOS);
+
+#if 0
+            if (!CCollisionSA::TestLineSphere(
+                CColLineSA{localOrigin, 0.f, localEnd, 0.f},
+                reinterpret_cast<CColSphereSA&>(*RpAtomicGetBoundingSphere(a)) // Fine for now
+            )) {
+                return true; // Line segment doesn't touch bsp
+            }
+#endif
```
botder, line 434: Is there any reason to keep this code?

Client/game_sa/CWorldSA.cpp
```diff
@@ -350,6 +388,157 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
...
+        targetEntity->Placeable.matrix->ConvertToMatrix(c.entMat);
+        c.entInvMat = c.entMat.Inverse();
+
+        // ...to transform the line origin and end into object space
+        c.originOS     = c.entInvMat.TransformVector(*vecStart);
+        c.endOS        = c.entInvMat.TransformVector(*vecEnd);
+        c.dirOS        = c.endOS - c.originOS;
+        c.minHitDistSq = c.dirOS.LengthSquared(); // By setting it to this value we avoid collisions that would be detected after the line segment [but are still ont the ray]
+
+        //const auto lineMagSq = (ctx.originOS - ctx.endOS).LengthSquared();
```
botder: Can we remove this?

Client/game_sa/CWorldSA.cpp
```diff
@@ -350,6 +388,157 @@ bool CWorldSA::ProcessLineOfSight(const CVector* vecStart, const CVector* vecEnd
...
+
+        // Do raycast against the DFF to get hit position material UV and name
+        // This is very slow
+        // Perhaps we could paralellize it somehow? [OpenMP?]
+        const auto ProcessOneAtomic = [](RpAtomic* a, void* data) {
```
botder: Please format the entire block properly.

Client/game_sa/CWorldSA.cpp
```diff
@@ -227,6 +239,32 @@ void CWorldSA::RemoveReferencesToDeletedObject(CEntitySAInterface* entity)
...
+
+void _rpAtomicResyncInterpolatedSphere(RpAtomic* atomic) {
+    ((void(__cdecl*)(RpAtomic*))0x7491F0)(atomic);
+}
+
+/* NB "RpAtomicGetBoundingSphere(atomic++)" will break it */
+#define RpAtomicGetBoundingSphere(_atomic)                              \
+    ((((_atomic)->interpolator.flags & rpINTERPOLATORDIRTYSPHERE)?      \
+      _rpAtomicResyncInterpolatedSphere(_atomic), 0: 0),                \
+      &((_atomic)->boundingSphere))
```
botder: This belongs in the RenderWare header.

Client/game_sa/CWorldSA.cpp
```diff
@@ -26,6 +28,16 @@ namespace
...
 
+class CCollisionSA {
+public:
+    static bool TestLineSphere(
+        const CColLineSA& line,
+        const CColSphereSA& sphere
+    ) {
+        return reinterpret_cast<bool(__cdecl*)(const CColLineSA&, const CColSphereSA&)>(0x417470)(line, sphere);
+    }
+};
```
botder: Move this to a [new] collision-related file.

botder (changes requested):

tederis:
Despite the fact that it works the efficiency of an algorithm is terrible. Given that it's expected to be used with vehicles, it's important to notice that those vehicles are often replaced with custom ones which have much more triangles. It definetely requires a parallelization and SIMD optimizations of operations with matrices and vectors (I think SIMD is even more important than a parallelization).

Client/sdk/game/CWorld.h
```diff
@@ -51,6 +51,14 @@ struct SLineOfSightBuildingResult
     CEntitySAInterface* pInterface;
 };
 
+struct SProcessLineOfSightMaterialInfoResult {
+    CVector2D   uv;            // On-texture UV coordinates of the intersection point
+    const char* textureName;   // GTA texture name
+    const char* frameName;     // The name of the frame the hit geometry belongs to
```
tederis: `textureName` and `frameName` should be initialized I guess.

Pirulax:
Yeah I do agree that the algorithm's performance isn't really all that good.
I could parallelize it, but the facilities [Thread pool, etc] necessary to do so aren't present in the `game_sa` project .
SIMD would be great though, but I do expect the compiler to use some SIMD by itself [Though, actually enabling it to use AVX would be great, as right now it only uses SSE... And this is why using a libary that does all this SIMD stuff with run-time dispatch would be great)

Client/sdk/game/CWorld.h
Pirulax:
```cpp
bool valid{}; // Data found in this struct is only valid if this is `true`!
```

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
Pirulax: For the number of args?

Client/game_sa/CWorldSA.cpp
Pirulax: Yeah, hopefully, eventually, it'll be fixed

Client/game_sa/CWorldSA.cpp
Pirulax: The context stuff? No because of `RpClumpForAllAtomics`

Client/game_sa/CWorldSA.cpp
Pirulax: Should I just run clang-format on it?

Shared/sdk/CVector.h
Pirulax: All other operators are noexcept too, why change it?

Client/game_sa/CWorldSA.cpp
Pirulax: Ah, okay you meant the 417th line

tederis:
> Yeah I do agree that the algorithm's performance isn't really all that good. I could parallelize it, but the facilities [Thread pool, etc] necessary to do so aren't present in the `game_sa` project . SIMD would be great though, but I do expect the compiler to use some SIMD by itself [Though, actually enabling it to use AVX would be great, as right now it only uses SSE... And this is why using a libary that does all this SIMD stuff with run-time dispatch would be great)

Unfortunately, the quality of vectorization is insufficient in today's compilers(even with -O3 and enabled vectorization). You can easily see it by the inspection of an assembly. That's why you can find SIMD intrinsics in every math library(including game engine math libs): Eigen, glm, numpy(scipy), UE linear algebra, CryEngine math, and many others.

Pirulax:
Yes I know it's not perfect, but it's better than plain old x87 arithmetic.

tederis:
> Yes I know it's not perfect, but it's better than plain old x87 arithmetic.

x87 for what? SSE for example has nothing to do with x87. SSE intrinsics (see Intel® Intrinsics Guide) are necessary for any math library(including MTA's one). It's not possible to write an efficient and performant math without it.

Shared/sdk/CVector.h
botder: Because you can access an element out-of-range, and even the STL does not use `noexcept` for `operator[]` in its classes.

Client/game_sa/CCollisionSA.cpp
```diff
@@ -0,0 +1,7 @@
+#include "StdInc.h"
```
botder, line 12: Copyright header is missing here.

Client/game_sa/CCollisionSA.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
botder, line 12: Copyright header is missing here.

Client/sdk/game/CCollision.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
botder, line 12: Copyright header is missing here.

Client/sdk/game/CCollision.h
```diff
@@ -0,0 +1,9 @@
...
+
+struct CColLineSA;
+struct CColSphereSA;
+
+class CCollision {
+public:
+    virtual bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) = 0;
```
botder:
```suggestion
    virtual bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) const = 0;
```

Client/game_sa/CCollisionSA.h
```diff
@@ -0,0 +1,8 @@
...
+
+#include "game/CCollision.h"
+
+class CCollisionSA : CCollision {
+public:
+    bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) override;
```
botder:
```suggestion
    bool TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) const override;
```

Client/game_sa/CCollisionSA.cpp
```diff
@@ -0,0 +1,7 @@
...
+
+#include "CCollisionSA.h"
+
+bool CCollisionSA::TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) {
+    return reinterpret_cast<bool(__cdecl*)(const CColLineSA&, const CColSphereSA&)>(0x417470)(line, sphere);
+}
```
botder:
```suggestion
bool CCollisionSA::TestLineSphere(const CColLineSA& line, const CColSphereSA& sphere) const
{
    return reinterpret_cast<bool(__cdecl*)(const CColLineSA&, const CColSphereSA&)>(0x417470)(line, sphere);
}
```

Client/sdk/game/RenderWare.h
```diff
@@ -321,25 +321,35 @@ struct RwGeometry
...
-    unsigned int unknown2;
-    float        unknown3;
-    float        unknown4;
-    float        unknown5;
+    int32 flags;
+    int16 startMorphTarget;
+    int16 endMorphTarget;
+    float time;
+    float recipTime;
+    float position;
```
botder:
```suggestion
    int32_t flags;
    int16_t startMorphTarget;
    int16_t endMorphTarget;
    float   time;
    float   recipTime;
    float   position;
```

Client/sdk/game/RenderWare.h
```diff
@@ -321,25 +321,35 @@ struct RwGeometry
     unsigned char  unknown1[14];
     unsigned short refs;
 };
-struct RpInterpolation
+
+/* Interpolator flags */
+enum RpInterpolatorFlag : int32
```
botder:
```suggestion
enum RpInterpolatorFlag : int32_t
```

Client/sdk/game/RenderWare.h
```diff
@@ -412,9 +429,35 @@ struct RpGeometry
...
+inline auto RpAtomicGetFrame(RpAtomic* atomic) {
+    return (RwFrame*)atomic->object.object.parent;
+}
+
+inline auto RwFrameGetParent(RwFrame* f) {
+    return (RwFrame*)rwObjectGetParent((RwObject*)f);
+}
+
+inline auto RwFrameGetMatrix(RwFrame* f) {
+    return &f->modelling;
```
botder, line 448: I can't tell what the return type here is.

Client/sdk/game/RenderWare.h
```diff
@@ -412,9 +429,35 @@ struct RpGeometry
...
+}
+
+inline void _rpAtomicResyncInterpolatedSphere(RpAtomic* atomic) {
+    ((void(__cdecl*)(RpAtomic*))0x7491F0)(atomic);
```
botder:
```suggestion
    reinterpret_cast<void(__cdecl*)(RpAtomic*)>(0x7491F0)(atomic);
```

botder (changes requested):
How heavy is the performance hit?

Pirulax:
There's no performance hit to PLOS unless this feature is enabled, otherwise it depends on the complexity of the mesh. [Obviously, the more complex, the slower it is].
As you might expect, since we don't utilize SIMD, it's not fast by any means, and it can be improved, but I'd argue we can do that later too [I don't expect this feature to be used very often, and devs already know that PLOS is *slow*]

botder (approved):

Pirulax:
[Wiki page](https://wiki.multitheftauto.com/wiki/ProcessLineAgainstMesh) added

PR #3100, opened 2023-07-04
Remove min-version lua function from old MTA versions

Moris-Onz:
Remove min-version lua function from old MTA versions


Server/mods/deathmatch/logic/CResourceChecker.Data.h
```diff
@@ -324,106 +111,6 @@ namespace
...
-        {"setColPolygonPointPosition", "1.5.7-0.20397"},
-        {"setColShapeRadius", "1.5.7-0.20397"},
-        {"setColShapeSize", "1.5.7-0.20397"},
-        {"isResourceProtected", "1.5.7-9.20468"},
-        {"hasElementData", "1.5.7-9.20477"},
-        {"addElementDataSubscriber", "1.5.7-9.20477"},
-        {"hasElementDataSubscriber", "1.5.7-9.20477"},
-        {"removeElementDataSubscriber", "1.5.7-9.20477"},
-
         // Features added in 1.5.9
```
sbx320: 1.5.9 additions should also be removed (as master is 1.6 now)

Server/mods/deathmatch/logic/CResourceChecker.Data.h
```diff
@@ -22,219 +22,6 @@ namespace
...
-        {"engineGetModelTextures", "1.5.7-9.20416"},
-        {"dxGetTextSize", "1.5.7-9.20447"},
-        {"hasElementData", "1.5.7-9.20447"},
-        {"resetBlurLevel", "1.5.7-9.20450"},
-        {"setVehicleModelWheelSize", "1.5.7-9.20642"},
-        {"getVehicleModelWheelSize", "1.5.7-9.20642"},
-        {"setVehicleWheelScale", "1.5.7-9.20642"},
-        {"getVehicleWheelScale", "1.5.7-9.20642"},
-
         // Features added in 1.5.9
```
sbx320: 1.5.9 additions should also be removed (as master is 1.6 now)

sbx320 (changes requested):
Looks good. We should also remove the 1.5.9 additions as we are shipping 1.6 now.

Server/mods/deathmatch/logic/CResourceChecker.Data.h
Moris-Onz: > 1.5.9 additions should also be removed (as master is 1.6 now)

fixed

sbx320 (approved):

PR #3103, opened 2023-07-05
Refactor the protocol 39

Moris-Onz:
Refactor the protocol 39, vehicle


Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -448,12 +448,10 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
...
                     BitStream.Write(&rotationDegrees);
 
-                    // Vehicle id as a char
-                    // I'm assuming the "-400" is for adjustment so that all car values can
-                    // fit into a char?  Why doesn't someone document this?
-                    //
-                    // --slush
-                    BitStream.Write(static_cast<unsigned char>(pVehicle->GetModel() - 400));
+                    // Write the vehicle
+                    unsigned short usVehicleModel = pVehicle->GetModel();
```
tederis, line 452: What will happened if a vehicle model is not valid? In the previous version a packet is always correct, even when a model is invalid. But now a packet can be corrupted.

multimaks2:
this has happened, we are abandoning the 400-unit offset)🫡

tederis:
Given that a server-side model allocation can be added in future, using a 2-byte range might make sense. But these changes break the backward compatibility. Thus a bitstream version must be corrected. As well as the packet corruption has to be fixed.

multimaks2:
As for modding, it is better to abandon the offset now, because today we can use the transport identifier from 0 to 255, but today this is not enough. And user projects that do not know that a shift is taking place - they will study for a long time "why the wrong identifier comes from the server".   This transition can be made as a last resort with the release of a new official version of the MTA. The fact that the server will send in a packet instead of 1 byte - 2 is not so critical, given that the Internet is more or less developed. (translated)

Lpsd:
Regardless, most changes to packets will need to make use of BitStream versioning in order to retain backwards compatibility. See https://github.com/multitheftauto/mtasa-blue/wiki/BitStream

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
Moris-Onz: IDs from 400 to 611 are allowed, ``if (CVehicleManager::IsValidModel(usVehicleModel))``

tederis: > IDs from 400 to 611 are allowed, `if (CVehicleManager::IsValidModel(usVehicleModel))`

Exactly this conditional expression can break a packet. You must always write something to a packet, even when a model is invalid. Or not send a packet at all. Or if you are pretty sure that a model will always be valid don't test it(but use `assert(CVehicleManager::IsValidModel(usVehicleModel))`).

Olya-Marinova: You can make sure that the package is not sent at all if the ID of the vehicle is invalid

Moris-Onz: It can be done that an invalid vehicle ID will not be sent

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -3196,23 +3196,13 @@ void CPacketHandler::Packet_EntityAdd(NetBitStreamInterface& bitStream)
...
-                    unsigned short usModel = ucModel + 400;
-                    if (!CClientVehicleManager::IsValidModel(usModel))
-                    {
-                        RaiseEntityAddError(39);
-                        return;
-                    }
+                    // Read out the vehicle
+                    unsigned short usModel = 0;
+
+                    if (CClientVehicleManager::IsStandardModel(usModel))
```
tederis: This condition will never be true. And a vehicle will never be spawned.

botder:
This pull request is non-sensical.

PR #3104, opened 2023-07-05
Refactor modules

TracerDS:
Modules are kinda limited (and dead really). Upgrading them will remove some limitations, etc.

This upgrade **breaks** backwards compatibility due to removed and reformatted functions

### DIFF
--- a/Server/mods/deathmatch/logic/lua/CLuaModule.cpp
+++ b/Server/mods/deathmatch/logic/lua/CLuaModule.cpp
@@ -308,17 +308,17 @@ bool CLuaModule::GetResourceName(lua_State* luaVM, std::string& strName) const n
     return true;
 }
 
-std::string CLuaModule::GetResourceName(lua_State* luaVM) const noexcept {
+std::optional<std::string> CLuaModule::GetResourceName(lua_State* luaVM) const noexcept {
     if (!luaVM)
-        return "";
+        return std::nullopt;
         
     CLuaMain* pLuaMain = m_pLuaModuleManager->GetLuaManager()->GetVirtualMachine(luaVM);
     if (!pLuaMain)
-        return "";
+        return std::nullopt;
         
     CResource* pResource = pLuaMain->GetResource();
     if (!pResource)
-        return "";
+        return std::nullopt;
         
     return pResource->GetName();
 }
--- a/Server/mods/deathmatch/logic/lua/CLuaModule.cpp
+++ b/Server/mods/deathmatch/logic/lua/CLuaModule.cpp
@@ -19,41 +19,38 @@
 
 extern CGame* g_pGame;
 
-CLuaModule::CLuaModule(CLuaModuleManager* pLuaModuleManager, CScriptDebugging* pScriptDebugging, const char* szFileName, const char* szShortFileName)
-{
-    // Set module manager
-    m_pLuaModuleManager = pLuaModuleManager;
-    // Set script debugging
-    m_pScriptDebugging = pScriptDebugging;
-    // set module path
-    m_szFileName = SString("%s", szFileName);
-    m_szShortFileName = SString("%s", szShortFileName);
-    // set as uninitialised
-    m_bInitialised = false;
-}
+CLuaModule::CLuaModule(CLuaModuleManager* pLuaModuleManager, CScriptDebugging* pScriptDebugging,
+    const char* szFileName, const char* szShortFileName
+) noexcept : m_pLuaModuleManager(pLuaModuleManager), m_pScriptDebugging(pScriptDebugging),
+    m_szFileName(SString("%s", szFileName)), m_szShortFileName(SString("%s", szShortFileName)),
+    m_bInitialised(false)
+{}
 
-CLuaModule::~CLuaModule()
+CLuaModule::~CLuaModule() noexcept
 {
-    if (m_hModule)
+    if (!m_hModule)
+        return;
+        
+    if (m_bInitialised)
     {
-        if (m_bInitialised)
-        {
-            // Shutdown module
-            m_FunctionInfo.ShutdownModule();
+        // Shutdown module
+        m_FunctionInfo.ShutdownModule();
 
-            // Unregister Functions
-            _UnregisterFunctions();
+        // Unregister Functions
+        UnregisterFunctions();
 
-            CLogger::LogPrintf("MODULE: Unloaded \"%s\" (%.2f) by \"%s\"\n", m_FunctionInfo.szModuleName, m_FunctionInfo.fVersion, m_FunctionInfo.szAuthor);
-        }
-
-        // Unload Module
-        _UnloadModule();
+        CLogger::LogPrintf(
+            "MODULE: Unloaded \"%s\" (%.2f) by \"%s\"\n",
+            m_FunctionInfo.szModuleName, m_FunctionInfo.fVersion, m_FunctionInfo.szAuthor
+        );
     }
+
+    // Unload Module
+    UnloadModule();
 }
 
-#ifdef WIN32
-bool IsModule32Bit(const SString& strExpectedPathFilename)
+#ifdef _WIN32
+bool IsModule32Bit(const SString& strExpectedPathFilename) noexcept
 {
     FILE* fh = File::Fopen(strExpectedPathFilename, "rb");
     fseek(fh, 60, SEEK_SET);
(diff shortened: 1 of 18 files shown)


Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -297,59 +292,56 @@ bool CLuaModule::RegisterFunction(lua_State* luaVM, const char* szFunctionName,
     return true;
 }
 
-bool CLuaModule::GetResourceName(lua_State* luaVM, std::string& strName)
+std::string CLuaModule::GetResourceName(lua_State* luaVM)
```
tederis: In my opinion, std::optional would fit here better

Server/mods/deathmatch/logic/lua/CLuaModule.h
```diff
@@ -47,32 +47,36 @@ struct FunctionInfo
...
-    const char*   GetVersionName();
-    unsigned long GetNetcodeVersion();
-    const char*   GetOperatingSystemName();
-    lua_State*    GetResourceFromName(const char* szResourceName);
-    // GetResourceName above might not work if module and MTA were compiled with different compiler versions
-    bool GetResourceName(lua_State* luaVM, char* szName, size_t length) override;
-    bool GetResourceFilePath(lua_State* luaVM, const char* fileName, char* path, size_t length) override;
+    ulong       GetVersion();
+    const char* GetVersionString();
+    const char* GetVersionName();
```
tederis: Getters must be usually const

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -199,16 +199,16 @@ void CLuaModule::_UnregisterFunctions()
     list<CLuaMain*>::const_iterator liter = m_pLuaModuleManager->GetLuaManager()->IterBegin();
     for (; liter != m_pLuaModuleManager->GetLuaManager()->IterEnd(); ++liter)
     {
-        lua_State*                luaVM = (*liter)->GetVM();
-        vector<SString>::iterator iter = m_Functions.begin();
-        for (; iter != m_Functions.end(); ++iter)
+        lua_State* luaVM = (*liter)->GetVM();
+
+        for (auto iter : m_Functions)
```
tederis: In additional, name **iter** is not the best name for a string

Moris-Onz:
good job

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -199,16 +199,16 @@ void CLuaModule::_UnregisterFunctions()
     list<CLuaMain*>::const_iterator liter = m_pLuaModuleManager->GetLuaManager()->IterBegin();
```
tederis: As long as you cannot completely get rid of iterators I think it's a good idea to hide their types by using auto.

TracerDS:
> good job

well, i didnt do much but much can be achieved with this pr.

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
Pirulax: should be `const auto&` - we don't want to make string copies unnecessarily

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -229,24 +229,19 @@ void CLuaModule::_ResourceStopped(lua_State* luaVM)
     if (m_FunctionInfo.ResourceStopped)
         m_FunctionInfo.ResourceStopped(luaVM);
 
-    vector<SString>::iterator iter = m_Functions.begin();
-    for (; iter != m_Functions.end(); ++iter)
+    for (auto iter : m_Functions)
```
Pirulax: Ditto

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -297,59 +292,56 @@ bool CLuaModule::RegisterFunction(lua_State* luaVM, const char* szFunctionName,
...
+    if (!luaVM)
+        return CChecksum();
+
+    CLuaMain* pLuaMain = m_pLuaModuleManager->GetLuaManager()->GetVirtualMachine(luaVM);
+    if (!pLuaMain)
+        return CChecksum();
+
+    CResource* pResource = pLuaMain->GetResource();
+    if (!pResource)
+        return CChecksum();
```
Pirulax: use `return {}`

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -297,59 +292,56 @@ bool CLuaModule::RegisterFunction(lua_State* luaVM, const char* szFunctionName,
...
+
```
Pirulax: Ditto

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -297,59 +292,56 @@ bool CLuaModule::RegisterFunction(lua_State* luaVM, const char* szFunctionName,
...
-                list<CResourceFile*>::iterator iter = pResource->IterBegin();
-                for (; iter != pResource->IterEnd(); ++iter)
-                {
-                    if (strcmp((*iter)->GetName(), szFile) == 0)
-                        return (*iter)->GetLastChecksum();
-                }
-            }
-        }
+        if (!strcmp((*iter)->GetName(), szFile))
+            return (*iter)->GetLastChecksum();
```
Pirulax: You could use a range for here too, no?

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
```diff
@@ -383,47 +375,44 @@ lua_State* CLuaModule::GetResourceFromName(const char* szResourceName)
...
-{
-    std::string resourceName;
-    if (GetResourceName(luaVM, resourceName))
-    {
-        std::strncpy(szName, resourceName.c_str(), length);
-        return true;
-    }
-
-    return false;
+    return pLuaMain ? pLuaMain->GetVM() : NULL;
```
Pirulax: nullptr

Server/mods/deathmatch/logic/lua/CLuaModule.h
```diff
@@ -47,32 +47,36 @@ struct FunctionInfo
...
-    void      DebugPrintf(lua_State* luaVM, const char* szFormat, ...);
-    bool      RegisterFunction(lua_State* luaVM, const char* szFunctionName, lua_CFunction Func);
-    bool      GetResourceName(lua_State*   luaVM,
-                              std::string& strName);            // This function might not work if module and MTA were compiled with different compiler versions
+    void Printf(const char* szFormat, ...);
+    void ErrorPrintf(const char* szFormat, ...);
+    void DebugPrintf(lua_State* luaVM, const char* szFormat, ...);
+
+    bool        RegisterFunction(lua_State* luaVM, const char* szFunctionName, lua_CFunction Func);
+    std::string GetResourceName(lua_State* luaVM);
```
Pirulax: I do feel like that the old function was better though.

Pirulax (changes requested):
Also, in order for the manager stuff to work you have to provide a DLL interface for those classes too.

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
TracerDS: No, you cant. Range loop work only on default `begin` and `end` functions

Server/mods/deathmatch/logic/lua/CLuaModule.h
TracerDS: The old function caused additional copies, new one should be a better approach

TracerDS:
> Also, in order for the manager stuff to work you have to provide a DLL interface for those classes too.

Yes, im aware of it but I have no idea how to do that without modifying literally everything :/

Server/mods/deathmatch/logic/lua/CLuaModule.cpp
CrosRoad95: are you sure it wont break modules? as you can see in ILuaModuleManager interface class, this function has signature:
```cpp
virtual bool                GetResourceName         ( lua_State * luaVM, std::string &strName ) = 0;
```

tederis: > are you sure it wont break modules? as you can see in ILuaModuleManager interface class, this function has signature:
> 
> ```c++
> virtual bool                GetResourceName         ( lua_State * luaVM, std::string &strName ) = 0;
> ```

It will definitely break modules. It also breaks the ABI.

TracerDS: > are you sure it wont break modules? as you can see in ILuaModuleManager interface class, this function has signature:
> 
> ```c++
> virtual bool                GetResourceName         ( lua_State * luaVM, std::string &strName ) = 0;
> ```

yes it will break other modules

Server/mods/deathmatch/logic/lua/CLuaModule.h
TracerDS: i didnt change it. It was like that for years. But ofc, they need to be const

Server/modules/IAccountManager.hpp
```diff
@@ -0,0 +1,17 @@
+#pragma once
```
TheNormalnij: Header with license information is missing

TracerDS: Thats the least of my problems here. We need to make it work im the first place

Dutchman101:
@TracerDS please resolve conflicts

PR #3107, opened 2023-07-06
Refactor: merge g_pClientGame and g_pGame into g_pSharedGame 

CrosRoad95:
as in title


botder (commented):
Wouldn't it be easier to just rename (and to understand) `g_pClientGame` to `g_pGame`?

PR #3110, opened 2023-07-08
Fix CStreamingSA::GetUnusedStreamHandle

tederis:
Seems like CStreamingSA::GetUnusedArchive returns not what is declared in the method's name. It always returns 0 that leads to an unexpected behaviour when engineAddImage called. engineAddImage is always adding an image to a slot 6. And furthermore, it can break the default SA streaming because of overriding of the first streaming handle.

### DIFF
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -210,7 +210,7 @@ unsigned char CStreamingSA::GetUnusedStreamHandle()
 {
     for (size_t i = 0; i < VAR_StreamHandlersMaxCount; i++)
     {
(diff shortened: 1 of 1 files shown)


CrosRoad95:
maybe thats why streaming flickering and this can fix issue mentioned here #3092

Zangomangu:
> maybe thats why streaming flickering and this can fix issue mentioned here #3092

nah the flickering/artifacting fixed in that PR is caused specifically by a bug in the bone animation code which is just exacerbated by higher memory usage (which is why it never occurred on the PS2)

TheNormalnij:
lgtm, tested

TheNormalnij (approved):

PR #3111, opened 2023-07-08
IMG count extension

tederis:
An attempt to extend a number of allowed IMG archives that can be simultaneously used and added through **engineAddImage**. The maximum possible number now is 245 in total(and 239 accessible from Lua). It allocates archives on demand which minimize the memory footprint. It also includes changes from #3110.

### DIFF
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -20,8 +20,6 @@ extern CCoreInterface* g_pCore;
 
 // count: 26316 in unmodified game
 CStreamingInfo (&CStreamingSA::ms_aInfoForModel)[26316] = *(CStreamingInfo(*)[26316])0x8E4CC0;
-HANDLE (&CStreamingSA::m_aStreamingHandlers)[32] = *(HANDLE(*)[32])0x8E4010;                // Contains open files
-CArchiveInfo (&CStreamingSA::ms_aAchiveInfo)[8] = *(CArchiveInfo(*)[8])0x8E48D8;            // [8][0x30]
 HANDLE* phStreamingThread = (HANDLE*)0x8E4008;
 uint32(&CStreamingSA::ms_streamingHalfOfBufferSize) = *(uint32*)0x8E4CA8;
 void* (&CStreamingSA::ms_pStreamingBuffer)[2] = *(void* (*)[2])0x8E4CAC;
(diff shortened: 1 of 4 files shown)


Moris-Onz:
good

Olya-Marinova:
If the maximum number of archives is 8, then the maximum number of archives is 7

Moris-Onz:
The maximum of archives should be 32 as fastman92 limit adjuster says

tederis:
255 archives is the maximum possible number because of the size of CStreamingInfo::archiveId(8 bit field). I didn't get a chance to test all 255 images but it works with 36 without any noticeable issues.

Client/game_sa/CStreamingSA.cpp
```diff
@@ -51,13 +49,179 @@ namespace
             memoryUsed = *(DWORD*)VAR_CStreaming_memoryUsed;
         }
     };
-}            // namespace
+
+    constexpr size_t MAX_IMAGES_NUM = 0xFF;
+    constexpr size_t MAX_STREAMS_NUM = 0xFF;
```
TheNormalnij: I remember steam handlers used for game audio. So MAX_IMAGES_NUM can't be 255

Client/game_sa/CStreamingSA.cpp
```diff
@@ -51,13 +49,179 @@ namespace
...
+    std::memcpy(m_Imgs.data(), defaultAchiveInfo, sizeof(CArchiveInfo) * VAR_DefaultMaxArchives);
+}
+
+void CStreamingSA::SetArchivesNum(size_t imagesNum)
+{
+    if (imagesNum > MAX_IMAGES_NUM)
+        return;    
+
+    /*
+        IMGs
```
TheNormalnij: If we use SOLID principles, this should be in different class (S)

tederis: What exactly should be in a different class? Method SetArchivesNum? Are you sure that a dedicated class can make it better? SOLID is awesome but it is not something that should be applied everywhere.

Client/game_sa/CStreamingSA.cpp
tederis: It is quite possible, but I cannot find an evidence of this. If you could indicate the location of this, I would appreciate it.

Client/game_sa/CStreamingSA.cpp
```diff
@@ -51,13 +49,179 @@ namespace
...
+    // Copy the default data
+    HANDLE (&defaultStreamingHandlers)[32] = *(HANDLE(*)[32])0x8E4010;
+    SStreamName (&defaultStreamingNames)[32] = *(SStreamName(*)[32])0x8E4098;
+    CArchiveInfo (&defaultAchiveInfo)[8] = *(CArchiveInfo(*)[8])0x8E48D8;
+
+    assert(m_StreamHandles.size() == VAR_DefaultStreamHandlersMaxCount);
+    assert(m_StreamNames.size() == VAR_DefaultStreamHandlersMaxCount);
+    assert(m_Imgs.size() == VAR_DefaultMaxArchives);
+
+    std::memcpy(m_StreamHandles.data(), defaultStreamingHandlers, sizeof(HANDLE) * VAR_DefaultStreamHandlersMaxCount);
```
Pirulax: Why would you memcpy?

Client/game_sa/CStreamingSA.cpp
Pirulax: I think we're way past the point where any kind of principles can be applied to this codebase 😂

Pirulax:
I feel like that the amount of memory you save by allocating on-demand is so small that it's more of an overhead [to allocate on-demand] than anything.

Client/game_sa/CStreamingSA.cpp
tederis: Because it is the only way to preserve the default data. As you can see SetArchivesNum replaces those three arrays and without the copying they would be empty. At the moment of copying some of the arrays contains a data that will be used later for the streaming.

tederis:
> I feel like that the amount of memory you save by allocating on-demand is so small that it's more of an overhead [to allocate on-demand] than anything.

Without dynamic allocation it would take around 30 kB of memory. I agree that it is a relatively little quantity, but this can be important on some of the "potato" systems(or for heavily loaded gamemodes), though. It preallocates with a margin that makes the overhead negligible.

Pirulax:
Okay, since you've already made it, lets keep it like that I guess.

Also, if you want to fiddle with it, it'd be possible to store 4 bytes for the `archieveId` by using the existing 1 byte and the 3 bytes of padding at the end of the struct

tederis:
Although it can be tempting but poses much more difficulties than you can think. Even if **archieveId** gets extended there is an another problem area: **CArchiveInfo::uiStreamHandleId**. It stores only 8 bits for a stream ID which makes the possible extension of archieveId meaningless. Of course this problem also can be overcome but would go beyond the boundaries of practical needs and rationality.

Client/game_sa/CStreamingSA.cpp
TheNormalnij: GTA can't handle more that 246 IMG archives.
By default it uses 16 stream handlers and 6 IMG archives. 10 stream handlers used for audio. So 256 - 10 = 246
So MAX_IMAGES_NUM should be 245, it there isn't any other limitations.

tederis: I have done a bit of research and can tell you that you're right. Perhaps with the exception of the amount of audio streams(it's 9 streams). Actually SA uses 10 reserved handles(9 for audio + 1 for GTA3.IMG(yes, there are two GTA3's handles). The full list of handles in the default game: "MODELS\\GTA3.IMG", "AUDIO\\SFX\\FEET", "AUDIO\\SFX\\GENRL", "AUDIO\\SFX\\PAIN_A", "AUDIO\\SFX\\SCRIPT", "AUDIO\\SFX\\SPC_EA", "AUDIO\\SFX\\SPC_FA", "AUDIO\\SFX\\SPC_GA", "AUDIO\\SFX\\SPC_NA", "AUDIO\\SFX\\SPC_PA", "MODELS\\GTA3.IMG", "MODELS\\GTA_INT.IMG", "DATA\\PATHS\\CARREC.IMG ", "DATA\\SCRIPT\\SCRIPT.IMG ", "MODELS\\CUTSCENE.IMG ",  "MODELS\\PLAYER.IMG". I've taken all the points in the last commit.

Pirulax:
imo even 2-3 images are enough, 256 is going to be plenty, that was just a thought.

Pirulax (approved):
LGTM - I didn't deep-dive neither into the code, or bug testing, but that's what nightlies are for..

Pirulax:
Okay let's merge it, and see what the nightly brings us.

PR #3112, opened 2023-07-08
Fix a desync state after aborted carjacking

Zangomangu:
Sometimes driver and sometimes passenger is desynced on the screen of a remote player after a carjacking. The driver is stuck to the car but not actually inside. Sometimes the vehicle goes flying.
Happens with high latency or lag.

Looks like this. The desync remains even after jacking is aborted.

https://github.com/multitheftauto/mtasa-blue/assets/59518113/94e1b166-d9ae-42ce-a57d-d0677ab14c40

With the fix. The desync is corrected when the jacking is aborted.

https://github.com/multitheftauto/mtasa-blue/assets/59518113/ffea112f-bb8b-4e4a-86d7-964c61ba1226

### DIFF
--- a/Client/mods/deathmatch/logic/CPacketHandler.cpp
+++ b/Client/mods/deathmatch/logic/CPacketHandler.cpp
@@ -1791,6 +1791,7 @@ void CPacketHandler::Packet_Vehicle_InOut(NetBitStreamInterface& bitStream)
                                         // For bikes and cars where jacked through passenger door, warp the passenger back in if desynced
                                         if (ucSeat == 0) {
                                             CClientPed* pPassenger = pJacked->GetOccupiedVehicle()->GetOccupant(1);
+                                            // Is the passenger a remote player or ped and is he physically outside but supposed to be in
                                             if (pPassenger &&
                                                 !pPassenger->IsLocalPlayer() &&
                                                 !pPassenger->IsSyncing() &&
--- a/Client/mods/deathmatch/logic/CPacketHandler.cpp
+++ b/Client/mods/deathmatch/logic/CPacketHandler.cpp
@@ -1777,9 +1777,33 @@ void CPacketHandler::Packet_Vehicle_InOut(NetBitStreamInterface& bitStream)
                             CClientPed* pJacked = pVehicle->GetOccupant(ucSeat);
 
                             // If it's the local player or syncing ped getting jacked, reset some stuff
-                            if (pJacked && (pJacked->IsLocalPlayer() || pJacked->IsSyncing()))
-                            {
-                                pJacked->ResetVehicleInOut();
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -1777,9 +1777,33 @@ void CPacketHandler::Packet_Vehicle_InOut(NetBitStreamInterface& bitStream)
                             CClientPed* pJacked = pVehicle->GetOccupant(ucSeat);
 
                             // If it's the local player or syncing ped getting jacked, reset some stuff
-                            if (pJacked && (pJacked->IsLocalPlayer() || pJacked->IsSyncing()))
-                            {
-                                pJacked->ResetVehicleInOut();
+                            if (pJacked)
```
tederis: It seems that there should be an opening bracket.

tederis: In any case, nested conditionals without brackets is a kind of attribute of the bad code style.

Zangomangu: Fixed

lopezloo:
Brace style doesn't match the rest of the code.

lopezloo:
This video confuses me. Why that player jacks vehicle from passenger side? In MTA you always jack from driver side if passenger seat is occupied.

Zangomangu:
No you don't always jack from driver side, you jack from passenger side if driver door is blocked.
And it doesnt matter, the same desync can happen if you only jack the driver.

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -1777,9 +1777,31 @@ void CPacketHandler::Packet_Vehicle_InOut(NetBitStreamInterface& bitStream)
...
+                                        if (ucSeat == 0) {
+                                            CClientPed* pPassenger = pJacked->GetOccupiedVehicle()->GetOccupant(1);
+                                            if (pPassenger &&
+                                                !pPassenger->IsLocalPlayer() &&
+                                                !pPassenger->IsSyncing() &&
+                                                pPassenger->GetOccupiedVehicle() &&
+                                                !pPassenger->GetRealOccupiedVehicle())
+                                            {
+                                                pPassenger->WarpIntoVehicle(pPassenger->GetOccupiedVehicle(), pPassenger->GetOccupiedVehicleSeat());
+                                            }
```
botder: Could you describe this wonky if-statement with a comment?

Zangomangu: added a comment

botder (approved):

PR #3116, opened 2023-07-11
Add function `engineStreamingSetModelCacheLimits`

Pirulax:
Currently the img streaming functionality is quite unusable, as the models just end up being cached forever by `CModelCacheManager`.
So, the memory usage was the same, or sometimes even higher than with the OG method of replacing models.
This function gives some control over it, and thus makes the img streaming functionality actually useable.
Ideally, we'd make it more intelligent, but since it's just a workaround [that I hope to remove], I don't think spending more time than absolutely necessary is worth it.

The original purpose of it [the cache] was to reduce/prevent pop-in. While it really did/does help with that, it's a workaround for the SA streaming being horribly inefficient, and slow.
The solution would be to rewrite the streamer [or at least tune it to PC hardware (and not CDs)], but that's a topic for another time [If anyone is interested, I have a few branches that do work, they just have to be ported overt to MTA and **tested**].


Client/core/CModelCacheManager.cpp
```diff
@@ -229,13 +255,15 @@ void CModelCacheManagerImpl::DoPulse()
...
     //  256MB streaming = 16+8 MB for peds & vehicles       72 peds + 56 veh
     //
-    int iStreamingMemoryAvailableKB = *(int*)0x08A5A80;
-
-    SSamplePoint<float> pedPoints[] = {{65536, 9}, {98304, 18}, {131072, 36}, {262144, 72}};
-    SSamplePoint<float> vehPoints[] = {{65536, 7}, {98304, 28}, {131072, 56}, {262144, 56}};
-
-    m_uiMaxCachedPedModels = (int)EvalSamplePosition<float>(pedPoints, NUMELMS(pedPoints), (float)iStreamingMemoryAvailableKB);
-    m_uiMaxCachedVehicleModels = (int)EvalSamplePosition<float>(vehPoints, NUMELMS(vehPoints), (float)iStreamingMemoryAvailableKB);
+    const auto iStreamingMemoryAvailableKB = *(int*)0x08A5A80;
```
botder, line 258: You shouldn't access game memory directly in the **core** project. Also, why not cast the value to **float** here?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -14,6 +14,13 @@
 #include <game/CObjectGroupPhysicalProperties.h>
 #include <game/CStreaming.h>
 #include <lua/CLuaFunctionParser.h>
+#include "../../../core/CModelCacheManager.h"
```
botder: That's just lazy. Please use an interface class for this (place it in `Client/sdk/core/` and inherit from it).

botder (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
Pirulax: That's how it was done before me too.

Client/core/CModelCacheManager.cpp
Pirulax: I didn't write this code, it was like this.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
Pirulax: See `CClientModelCacheManager.cpp`

botder (approved):

botder:
@Pirulax Please create a wiki page for this function.

Meine1:
How to use this function?

Xenius97:
> How to use this function?

engineStreamingSetModelCacheLimits(vehicle, ped)

We're using engineStreamingSetModelCacheLimits(0, 0) which fully disables cache. Reduced crashes 80% by this.

Meine1:
> Reduced crashes 80% by this.

it really does! I love this new function, it turned the IMG functions that were nearly useless due to the cache into the best feature of MTA 1.6, I get amazing memory performance even when replacing every single ped and the game feels faster too. I hope this one makes it into the stable build soon. It works on peds and vehicles, but does this also work on replaced objects? That would be really useful for total conversions, like the ones that add the GTA Vice City map in a separate dimension but take a lot of time to load with the old functions.

Meine1:
Is it possible to expand this function so that we can define how many objects can be kept in streaming memory as well?

PR #3118, opened 2023-07-12
Add Tracy, a frame profiler

Pirulax:
Adds [Tracy](https://github.com/wolfpld/tracy) v0.9.1: A real time, nanosecond resolution, remote telemetry, hybrid frame and sampling profiler

I've added tracing to quite a few functions.
Updated the readme with instructions on how to use Tracy. [TODO]

The performance impact isn't too bad.

To test, just download the Tracy v0.9.1 server from [here](https://github.com/tycho/tracy/releases/tag/v0.9.1)
We're using v0.9.1 in particular because I'm too lazy to compile the latest Tracy server.

I'd like to use submodules, but not sure if our current infrastructure would work with it.
I also get linker errors when enabling the callstack feature (Regarding `RtlWalkStackChain`), so if someone knows what the issue is, pls fix it, ty.

A possible issue could be security, as tracy must be in a separate dll, and uses sockets to communicate.

NOTE:
When connecting to the client [from the tracy server] **first connect to an MTA server** asTracy really doesn't like modules being unloaded while it's running [and it's going to crash].


tederis:
Nice. I have been using it for at least 3 years with MTA. But you should be aware of the macro TRACY_ON_DEMAND. Without using it a profiler will spend all the process memory in situations when it has no connections.

Pirulax:
Ah, so you've been using it with MTA.
I kinda have  trouble getting it to compile with callstack support, it's missing an external symbol. `__tracy_RtlWalkStackChain` [or whatever the fuck the name is].
No clue how to solve it. Mind showing your premake for it?

tederis:
Nothing special:
```
project "Tracy"
	language "C++"
	kind "StaticLib"
	targetname "tracy"

	defines { "TRACY_ENABLE=1", "TRACY_ON_DEMAND" }

	vpaths {
		["Headers/*"] = "**.h",
		["Headers/*"] = "**.hpp",
		["Sources/*"] = "**.cpp",
		["*"] = "premake5.lua"
	}

	files {
		"premake5.lua",
		"*.h",
		"*.hpp",
		"*.cpp"
	}

```

tederis:
Okay, I just tried out your branch. From what I can see, the problem is you're exporting DLL in all projects even when import is being required. You should remove `workspace "*"` from **vender/tracy/premake5.lua** and manually add these macros in every project that uses Tracy. Every project except the profiler itself must contain **TRACY_IMPORTS** macro.

tederis:
I also saw an error `C2065: 'ZoneScoped'`, but the solution is trivial I guess(add missing headers to PCHs).

Pirulax:
So I can't be lazy.
I also wonder if there's a better solution than adding that macro to every project I want to use tracy in manually?
CMake seems to *just work*, so there must be a good solution in premake too, no?

botder:
> I also wonder if there's a better solution than adding that macro to every project I want to use tracy in manually?

Premake doesn't support that. CMake [propagates requirements](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#target-usage-requirements) of a target to dependents automatically.

Adding Tracy is a good idea.

Pirulax:
Okay Ill work on this soon
Won't be adding server profiling for now, I don't think that's necessary yet.

CArg22:
Do you need live profiler? no
Do you need additional library? no

Instead of tracy just use debugging capabilities of browsers debugger, generate json that can be loaded to chrome devtools. 99999 times more safe than your approach. Mta already supporting json, add nice functions like "profilerStart", "profilerStop"  and maybe other

clientside:
```lua
addCommandHandler("profiler", function()
  profilerStart()
  setTimer(function()
    local json = profilerStop();
    local file = fileOpen("out.json")
    fileWrite(file, json)
    fileClose(file)
  end, 5000, 1)
end
```
then drag and drop that file onto chrome devtools

Issue #3032, opened 2023-05-23
engineImageLinkDFF and engineImageLinkTXD don't work correctly on some IDs


Meine1:
### Describe the bug

When replacing peds or vehicles using a IMG file with engineImageLinkDFF and engineImageLinkTXD, some models don't get replaced 100% of the time as they should. The more models you replace, the more likely these IDs won't work, the most problematic IDs I have found are 280, 60, 58, 44, 22, 17, 13, 7 but there's likely more as I didn't test all of them.

### Steps to reproduce

1. Use MTA 1.6.
2. Start the attached resource (VCskins) [VCskins.zip](https://github.com/multitheftauto/mtasa-blue/files/11543454/VCskins.zip)
3. Change skin to the replaced models (IDs  1, 7, 13, 20, 60, 90, 280, 281 )
4. They should all be GTA Vice City characters, almost all IDs work most of the time but ID 7 and ID 280 rarely work.
5. If they are all correctly replaced, restart the resource until the problem appears.

### Version

Client: v1.6-release-21902
Server v1.6-release-21902

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

Pirulax:
ID 7 (male01) is a very special model, as it's used as some kind of a base for all other models, and it must be loaded at all times [otherwise the game just crashes].
The issues with model 280 are caused by [this](https://github.com/multitheftauto/mtasa-blue/issues/3162#issuecomment-1693954333).

TheNormalnij:
engineImageLinkDFF doesn't change a model immediately. You should use `engineRestreamWorld` to reload models.

PR #3120, opened 2023-07-15
Additional API to change model TXD

TheNormalnij:
```
bool engineSetModelTXDID( number modelId, number txdId )
bool engineResetModelTXDID( number modelId )
```


Client/game_sa/CModelInfoSA.cpp
```diff
@@ -34,6 +34,7 @@ std::unordered_map<std::uint32_t, std::map<eVehicleDummies, CVector>> CModelInfo
 std::map<CTimeInfoSAInterface*, CTimeInfoSAInterface*>                CModelInfoSA::ms_ModelDefaultModelTimeInfo;
 std::unordered_map<DWORD, unsigned short>                             CModelInfoSA::ms_OriginalObjectPropertiesGroups;
 std::unordered_map<DWORD, std::pair<float, float>>                    CModelInfoSA::ms_VehicleModelDefaultWheelSizes;
+std::map<unsigned short, int>                                         CModelInfoSA::ms_DefaultTxdIDMap;
```
Pirulax, line 37: You should consider using a fixed-size array, it's faster, and uses less memory.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1256,7 +1258,34 @@ int CLuaEngineDefs::EngineRemoveShaderFromWorldTexture(lua_State* luaVM)
 
 uint CLuaEngineDefs::EngineGetModelTXDID(uint uiModelID)
 {
-    return g_pGame->GetRenderWare()->GetTXDIDForModelID(uiModelID);
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(uiModelID);
+
+    if (uiModelID >= 20000 || !pModelInfo)
```
Pirulax: Don't use magic numbers [Use an existing function or something of similar nature]

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1256,7 +1258,34 @@ int CLuaEngineDefs::EngineRemoveShaderFromWorldTexture(lua_State* luaVM)
...
+        throw std::invalid_argument("Expected a valid model ID in range [0-19999] at argument 1");
+
+    return pModelInfo->GetTextureDictionaryID();
+}
+
+bool CLuaEngineDefs::EngineSetModelTXDID(uint uiModelID, unsigned short usTxdId)
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(uiModelID);
+
+    if (uiModelID >= 20000 || !pModelInfo)
```
Pirulax: Ditto

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1256,7 +1258,34 @@ int CLuaEngineDefs::EngineRemoveShaderFromWorldTexture(lua_State* luaVM)
...
+
+    pModelInfo->SetTextureDictionaryID(usTxdId);
+    return true;
+}
+
+bool CLuaEngineDefs::EngineResetModelTXDID(uint uiModelID)
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(uiModelID);
+
+    if (uiModelID >= 20000 || !pModelInfo)
```
Pirulax: Ditto

Pirulax (changes requested):

CrosRoad95:
see how for example "EngineStreamingFreeUpMemory" lua function is implemented, move y our implementation above "Loadfunctions" so you can get rid of 2 lines of code in CLuaEngineDefs.h

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
TheNormalnij: It is only draft =)
Will be changed

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Fixed-size array will use at least 320 KB memory

Pirulax: Memory fragmentation >>> usage

TheNormalnij: This struct will be empty on 95% servers. I see no reason to optimize this struct with memory cost for now.

Pirulax:
I don't see you increasing the ref count of the TXDs anywhere, you should do that

TheNormalnij:
> I don't see you increasing the ref count of the TXDs anywhere, you should do that

@Pirulax
Are your sure? This will prevent txd from unloading

Pirulax:
A model [when loaded] will increase the refcount of a txd, and decrease it when it's unloaded.
So, when changing the txd's of a model you should manually unref the previous txd, and ref the new one.

TheNormalnij:
> A model [when loaded] will increase the refcount of a txd, and decrease it when it's unloaded.
> So, when changing the txd's of a model you should manually unref the previous txd, and ref the new one.

Makes sense. Thanks you

Pirulax (approved):
lgtm

Xenius97:
I've tested `engineSetModelTXDID` and `engineRequestTXD`, look good to me, but I noticed a possible memory leak. When you restart the resource, the requested ID won't be freed anymore (maybe if I manually call `engineFreeModel`). The same applies to `engineRequestTXD` and `engineFreeTXD`.

https://youtu.be/VoTWDzTKSjs
(Some default skins not working, See #3032)

Pirulax:
Depends on #1802

Pirulax:
Make sure to add the Wiki pages.

PR #3128, opened 2023-07-24
Add unsafe resource marker

botder:
This pull request introduces an `<unsafe>true</unsafe>` marker requirement for every resource that utilizes either the `load` or `loadstring` function. Furthermore, an unsafe resource is not allowed to be granted the `general.ModifyOtherObjects` permission. An unsafe resource can still use `general.ModifyOtherObjects.<resourceName>` to receive access to specific resources, if desired.

Without this unsafe marker, these two functions will not work, not on the server neither on the client (if the bitstream version supports it). If a player with a newer client joins a server which doesn't support the unsafe marker yet, the client will default to unsafe mode and thus allow the usage of  `load` and `loadstring` functions.

For developers I have added the `isUnsafeResource` function, which returns a boolean whether the current resource is marked as unsafe. For security reasons this function should never be capable of returning the unsafe marker for another resource. Now, if you want to use either `load` or `loadstring` in your resource, you need to:

- add `<unsafe>true</unsafe>` to the meta.xml
- grant `function.loadstring` and/or `function.load` permissions for the resource
- ensure that `general.ModifyOtherObjects` is not granted to the resource

At last, the server now also restricts the functions `load` and `loadstring` by default. That means the permission to use either of those must be explicitly granted by the ACL.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
@@ -433,6 +433,14 @@ int CLuaResourceDefs::GetResourceState(lua_State* luaVM)
 
 int CLuaResourceDefs::LoadString(lua_State* luaVM)
 {
+    CLuaMain& luaMain = lua_getownercluamain(luaVM);
+
+    if (!luaMain.IsUnsafe())
+    {
+        lua_pushboolean(luaVM, false);
+        return 1;
+    }
+
     //  func,err loadstring( string text[, string name] )
     SString strInput;
     SString strName;
@@ -483,6 +491,14 @@ int CLuaResourceDefs::LoadString(lua_State* luaVM)
 
 int CLuaResourceDefs::Load(lua_State* luaVM)
 {
+    CLuaMain& luaMain = lua_getownercluamain(luaVM);
+
+    if (!luaMain.IsUnsafe())
+    {
+        lua_pushboolean(luaVM, false);
+        return 1;
+    }
+
     //  func,err load( callback callbackFunction[, string name] )
     CLuaFunctionRef iLuaFunction;
     SString         strName;
@@ -498,18 +514,19 @@ int CLuaResourceDefs::Load(lua_State* luaVM)
         // Should apply some limit here?
         SString       strInput;
         CLuaArguments callbackArguments;
-        CLuaMain*     pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
-        while (pLuaMain)
+
+        while (true)
         {
             CLuaArguments returnValues;
-            callbackArguments.Call(pLuaMain, iLuaFunction, &returnValues);
+            callbackArguments.Call(&luaMain, iLuaFunction, &returnValues);
+
             if (returnValues.Count())
             {
                 CLuaArgument* returnedValue = *returnValues.IterBegin();
                 int           iType = returnedValue->GetType();
+
                 if (iType == LUA_TNIL)
                     break;
-
                 else if (iType == LUA_TSTRING)
                 {
                     std::string str = returnedValue->GetString();
@@ -520,6 +537,7 @@ int CLuaResourceDefs::Load(lua_State* luaVM)
                     continue;
                 }
             }
+
             break;
         }
 
--- a/Server/mods/deathmatch/logic/CResource.cpp
+++ b/Server/mods/deathmatch/logic/CResource.cpp
@@ -105,6 +105,7 @@ bool CResource::Load()
     m_bClientFiles = true;
 
     m_bOOPEnabledInMetaXml = false;
+    m_isUnsafeInMetaXml = false;
 
     m_pVM = nullptr;
     // @@@@@ Set some type of HTTP access here
@@ -199,12 +200,14 @@ bool CResource::Load()
                 m_bSyncMapElementDataDefined = true;
             }
 
-            m_bOOPEnabledInMetaXml = false;
-            CXMLNode* pNodeClientOOP = pRoot->FindSubNode("oop", 0);
-
-            if (pNodeClientOOP)
+            if (CXMLNode* node = pRoot->FindSubNode("oop", 0); node != nullptr)
             {
-                m_bOOPEnabledInMetaXml = StringToBool(pNodeClientOOP->GetTagContent().c_str());
+                m_bOOPEnabledInMetaXml = StringToBool(node->GetTagContent().c_str());
+            }
+
+            if (CXMLNode* node = pRoot->FindSubNode("unsafe", 0); node != nullptr)
+            {
+                m_isUnsafeInMetaXml = StringToBool(node->GetTagContent().c_str());
             }
 
             m_iDownloadPriorityGroup = 0;
@@ -842,8 +845,20 @@ bool CResource::Start(std::list<CResource*>* pDependents, bool bManualStart, con
     // Set the Resource Element name
     m_pResourceElement->SetName(m_strResourceName.c_str());
 
+    // Consider every resource to be safe by default.
+    m_isUnsafe = false;
+
+    if (m_isUnsafeInMetaXml)
+    {
+        bool canModifyOtherObjects = g_pGame->GetACLManager()->CanObjectUseRight(m_strResourceName.c_str(), CAccessControlListGroupObject::OBJECT_TYPE_RESOURCE,
+                                                                                 "ModifyOtherObjects", CAccessControlListRight::RIGHT_TYPE_GENERAL, false);
+
+        // We only grant unsafe to resources which can't access nor modify all other resources.
+        m_isUnsafe = (canModifyOtherObjects == false);
+    }
+
     // Create the virtual machine for this resource
-    CreateVM(m_bOOPEnabledInMetaXml);
+    CreateVM(m_bOOPEnabledInMetaXml, m_isUnsafe);
 
     // We're now active
     CLogger::LogPrintf(LOGLEVEL_LOW, "Starting %s\n", m_strResourceName.c_str());
@@ -1117,11 +1132,11 @@ bool CResource::Stop(bool bManualStop)
     return true;
 }
 
-bool CResource::CreateVM(bool bEnableOOP)
+bool CResource::CreateVM(bool bEnableOOP, bool isUnsafe)
 {
     if (!m_pVM)
     {
-        m_pVM = g_pGame->GetLuaManager()->CreateVirtualMachine(this, bEnableOOP);
+        m_pVM = g_pGame->GetLuaManager()->CreateVirtualMachine(this, bEnableOOP, isUnsafe);
         m_pResourceManager->NotifyResourceVMOpen(this, m_pVM);
     }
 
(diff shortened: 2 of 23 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -483,6 +491,14 @@ int CLuaResourceDefs::LoadString(lua_State* luaVM)
 
 int CLuaResourceDefs::Load(lua_State* luaVM)
 {
+    CLuaMain& luaMain = lua_getownercluamain(luaVM);
+
+    if (!luaMain.IsUnsafe())
+    {
+        lua_pushboolean(luaVM, false);
```
sbx320, line 498: Also here

Server/mods/deathmatch/logic/CResource.cpp
```diff
@@ -842,8 +845,20 @@ bool CResource::Start(std::list<CResource*>* pDependents, bool bManualStart, con
...
+    // Consider every resource to be safe by default.
+    m_isUnsafe = false;
+
+    if (m_isUnsafeInMetaXml)
+    {
+        bool canModifyOtherObjects = g_pGame->GetACLManager()->CanObjectUseRight(m_strResourceName.c_str(), CAccessControlListGroupObject::OBJECT_TYPE_RESOURCE,
+                                                                                 "ModifyOtherObjects", CAccessControlListRight::RIGHT_TYPE_GENERAL, false);
+
+        // We only grant unsafe to resources which can't access nor modify all other resources.
+        m_isUnsafe = (canModifyOtherObjects == false);
```
sbx320, line 857: This should probably also cause a warning log

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -433,6 +433,14 @@ int CLuaResourceDefs::GetResourceState(lua_State* luaVM)
 
 int CLuaResourceDefs::LoadString(lua_State* luaVM)
 {
+    CLuaMain& luaMain = lua_getownercluamain(luaVM);
+
+    if (!luaMain.IsUnsafe())
```
sbx320, line 438: We should throw an error here / print an error log, to indicate why this has failed.

sbx320 (changes requested):

sbx320:
This seems like an odd place to break backwards compatibility. Most issues around `loadscript` are with server owners who have no clue what they are doing. Those will 

- Complain everywhere that we're breaking their scripts, and then 
- Enable `unsafe` for their resources, negating any benefit of this.

They are unlikely to actually fix any of their issuses due to a lack of Lua skills. 

Also the security benefits regarding `general.ModifyOtherObjects.` are rather doubtful, since any `loadscript`-capable resource is likely also running in an acl-admin context. As such it could just grant itself access to the relevant other resource on-demand.

There are also more ways to run arbitrary Lua scripts than just `loadscript` that aren't affected by this change. 

> For security reasons this function should never be capable of returning the unsafe marker for another resource.

Which security reasons? I don't really see an attack vector where it would be relevant if another resource is running in unsafe mode. 

> At last, the server now also restricts the functions load and loadstring by default. That means the permission to use either of those must be explicitly granted by the ACL.

That seems like a reasonable thing to do (regardless of the rest of this PR). (Edit: `loadstring` handled in https://github.com/multitheftauto/mtasa-blue/pull/3127)

The whole `unsafe` notion seems to me like it would cause more pain than benefit for everyone involved.

botder:
> This seems like an odd place to break backwards compatibility. Most issues around `loadscript` are with server owners who have no clue what they are doing. Those will
> 
> * Complain everywhere that we're breaking their scripts, and then
> * Enable `unsafe` for their resources, negating any benefit of this.
> 
> They are unlikely to actually fix any of their issuses due to a lack of Lua skills.

The plan is to catch these developers/server operators, provide a dedicated wiki page and further explanations in our support channel(s), because the idea here is to protect them from being exploited due to lack of knowledge in security aspects and ACL. It's being annoying on purpose, because we can't expect everyone to update their resources and ACL.

> Also the security benefits regarding `general.ModifyOtherObjects.` are rather doubtful, since any `loadscript`-capable resource is likely also running in an acl-admin context. As such it could just grant itself access to the relevant other resource on-demand.

The solution here would be to disable `aclSetRight` for unsafe resources, no? I also noticed that map editor wouldn't work with these changes (see edf/scriptreader.lua)

> There are also more ways to run arbitrary Lua scripts than just `loadscript` that aren't affected by this change.

Alright, but this pull request is specifically for exploitation of loadstring and load. What else do you have on your mind?

> Which security reasons? I don't really see an attack vector where it would be relevant if another resource is running in unsafe mode.

I thought you shouldn't be able to loop over all resources to find the unsafe ones with a convenient function.

> > At last, the server now also restricts the functions load and loadstring by default. That means the permission to use either of those must be explicitly granted by the ACL.
> 
> That seems like a reasonable thing to do (regardless of the rest of this PR). (Edit: `loadstring` handled in #3127)

Yeah. My pull request is restricting these two functions in C++ by default, the linked pull request is just being explicit about it.

xLive:
The restriction of `loadstring` will break scripts that call [other resources](https://github.com/search?q=loadstring%28exports.&type=code) to import their OOP classes or functions. And it's also needed for other purposes as well. For example, [DGS](https://wiki.multitheftauto.com/wiki/Resource:DGS) has [G2D](https://wiki.multitheftauto.com/wiki/DgsG2DLoadHooker#Example), which hooks every CEGUI function/event and replaces it with DGS by adding just two lines.

The easiest solution for many people would be marking their resources as unsafe without knowing what it actually does, which ruins the goal of this PR.

botder:
> The restriction of `loadstring` will break scripts that call [other resources](https://github.com/search?q=loadstring%28exports.&type=code) to import their OOP classes or functions. And it's also needed for other purposes as well. For example, [DGS](https://wiki.multitheftauto.com/wiki/Resource:DGS) has [G2D](https://wiki.multitheftauto.com/wiki/DgsG2DLoadHooker#Example), which hooks every CEGUI function/event and replaces it with DGS by adding just two lines.
> 
> The easiest solution for many people would be marking their resources as unsafe without knowing what it actually does, which ruins the goal of this PR.

I am not going to merge this pull request, but I don't think the loadstring approach is the correct way to accomplish what you want - sharing code between resources should be easier/safer.

thisdp:
> > The restriction of `loadstring` will break scripts that call [other resources](https://github.com/search?q=loadstring%28exports.&type=code) to import their OOP classes or functions. And it's also needed for other purposes as well. For example, [DGS](https://wiki.multitheftauto.com/wiki/Resource:DGS) has [G2D](https://wiki.multitheftauto.com/wiki/DgsG2DLoadHooker#Example), which hooks every CEGUI function/event and replaces it with DGS by adding just two lines.
> > The easiest solution for many people would be marking their resources as unsafe without knowing what it actually does, which ruins the goal of this PR.
> 
> I am not going to merge this pull request, but I don't think the loadstring approach is the correct way to accomplish what you want - sharing code between resources should be easier/safer.

The loadstring approach is used to eliminate prefix in the code. For example "exports.dgs:dgsSetText()" is really an annoying usage.

botder:
> The loadstring approach is used to eliminate prefix in the code. For example "exports.dgs:dgsSetText()" is really an annoying usage.

That problem is easily solved without loadstring, though:
```lua
function import(resourceName, into)
    into = into or _G
    local resource = getResourceFromName(resourceName)
    local exports = resource and getResourceExportedFunctions(resource) or {}

    for _, name in pairs(exports) do
        into[name] = function (...)
            return call(resource, name, ...)
        end
    end
end

import "mapmanager"

addCommandHandler("mmtest", function (_, _, name)
    changeGamemode(getResourceFromName(name))
end)
```

thisdp:
> > The loadstring approach is used to eliminate prefix in the code. For example "exports.dgs:dgsSetText()" is really an annoying usage.
> 
> That problem is easily solved without loadstring, though:
> 
> ```lua
> function import(resourceName, into)
>     into = into or _G
>     local resource = getResourceFromName(resourceName)
>     local exports = resource and getResourceExportedFunctions(resource) or {}
> 
>     for _, name in pairs(exports) do
>         into[name] = function (...)
>             return call(resource, name, ...)
>         end
>     end
> end
> 
> import "mapmanager"
> 
> addCommandHandler("mmtest", function (_, _, name)
>     changeGamemode(getResourceFromName(name))
> end)
> ```

but you still need to define "import" in other resources.

botder:
> but you still need to define "import" in other resource

Of course, but 11 lines are not going to annoy anyone and you don't have a loadstring-backdoor.

thisdp:
> > but you still need to define "import" in other resource
> 
> Of course, but 11 lines are not going to annoy anyone and you don't have a loadstring-backdoor.

using loadstring, just 1 line

botder:
> using loadstring, just 1 line

If we were to add this kind of `import` function directly to MTA (like we do with `exports`), then it also would be 1 line without the security nightmare.

Edit: This is not really necessary, you can achieve this (more or less) with `exports` already:
```lua
mapmanager = exports.mapmanager

addCommandHandler("change_gamemode", function (_, _, name)
    mapmanager:changeGamemode(getResourceFromName(name))
end)
```

thisdp:
> > using loadstring, just 1 line
> 
> If we were to add this kind of `import` function directly to MTA (like we do with `exports`), then it also would be 1 line without the security nightmare.

that's the best way

xLive:
We still need a way to import OOP classes from one resource to another, along with a method to import wrapper functions. Currently, the only options are either using `loadstring` or asking the scripters to manually add the script to their resources. The latter is difficult to manage, especially when the resource gets updated and it's used by many resources.

tederis:
**loadstring** is heavily used in big projects. It is a heart of custom modules systems. This PR will complicate the life of those projects. But safety is important and these changes should have happeded a long ago.

botder:
> **Note**
> I am not going to merge this pull request. It's going to break map editor and the effort is not worth it.

botder:
> We still need a way to import OOP classes from one resource to another, along with a method to import wrapper functions. Currently, the only options are either using `loadstring` or asking the scripters to manually add the script to their resources. The latter is difficult to manage, especially when the resource gets updated and it's used by many resources.

We need a way to share scripts from resource to resource then. How should that look like?

Maybe a combination of `<script export="true" .../>` in the *source* resource,
and `<script from="resourceName" src="relative/path/here.lua"/>` in the *target* resource?

ghost:
I believe it should work similar to [Lua's require](https://www.lua.org/pil/8.1.html), which is widely present, e.g: in GMod/Project Zomboid. However **require** is disabled in MTA for security reasons.

Example from Project Zomboid:
obraz
ItemTweaker_Core.lua (different mod/resource):
obraz
Another mod, which fetch game shipped .lua:
obraz


Now i can't really explain how it works in-depth (as-in i've didn't used Lua outside MTA), but i've seen some people complaining about it missing. Creating new function like **mtaRequire** or **requireLua**, which is consistent with how **require** behaves (but with taking care of possible security vulnerabilities - which i am not aware of) seems like a good move. Especially for those people, who could come to MTA and knew Lua before it.

thisdp:
When require called, hash of the required part should be checked

PR #3042, opened 2023-05-25
Fix #3037: Add ResetPedVoice function

TracerDS:
Fixes #3037

Added new function: `resetPedVoice`.
It resets ped voice to default.
Syntax: `bool resetPedVoice ( ped thePed )`

PR #3129, opened 2023-07-25
Fix #3037: Add ResetPedVoice function - Bug fix

### LINKED ISSUE #3037: Add resetPedVoice function

we have now
setPedVoice
and
getPedVoice

... i wanted to make a ped mute ( even in pain he doesn't say much )
so i did setPedVoice( "disabled bank name" )

and it worked, the ped is perfectly silent... problem is when i want to undo this, at first i expected that changing my skin to another skin would override this and restore my voice, but it didn't..

ped remained mute until rejoin

i did think about using getPedVoice to store a backup of what i had before muting... but then when i changed the skin to a female and was previously male... the backup values no longer matched !

i would like resetPedVoice to be introduced as a native function that resets the voice bank to your skin IDs' native original bank !

_No response_

### Additional context

_No response_

### Security Policy

TracerDS:
The function was bugged; it was never actually resetting ped's voice due to internal mta's logic.
Now should be fixed, function returns `true` when the voice was successfully reset, returns `false` if the voice was not changed (ie. player joins the server, player executes `resetPedVoice` without ever calling `setPedVoice`)

The old PR (#3042) was merged, so I had to make new PR


Client/game_sa/CPedSA.h
```diff
@@ -265,24 +264,24 @@ class CPedSA : public virtual CPed, public virtual CPhysicalSA
 
 private:
     CWeaponSA*          m_pWeapons[WEAPONSLOT_MAX];
-    CPedIKSA*           m_pPedIK;
-    CPedIntelligenceSA* m_pPedIntelligence;
-    CPedSAInterface*    m_pPedInterface;
-    CPedSoundSA*        m_pPedSound;
+    CPedIKSA*           m_pPedIK = NULL;
```
sbx320: nullptr

sbx320 (changes requested):

Client/game_sa/CPedSoundSA.h
```diff
@@ -70,15 +70,22 @@ class CPedSoundSA : public CPedSound
     CPedSoundSAInterface* m_pInterface;
 
 public:
+    CPedSoundSA(CPedSoundSAInterface&& pInterface) noexcept
```
tederis: What is the purpose of rvalue reference here?

Client/game_sa/CPedSA.cpp
```diff
@@ -94,7 +88,7 @@ void CPedSA::Init()
     CPedIntelligenceSAInterface* m_pPedIntelligenceInterface = (CPedIntelligenceSAInterface*)(dwPedIntelligence);
     m_pPedIntelligence = new CPedIntelligenceSA(m_pPedIntelligenceInterface, this);
     m_pPedSound = new CPedSoundSA(&pedInterface->pedSound);
-    m_pDefaultPedSound = new CPedSoundSA(&pedInterface->pedSound);
+    m_pDefaultPedSound = new CPedSoundSA(std::move(pedInterface->pedSound));
```
tederis: What exactly are you trying to move here? CPedSoundSA constructor below doesn't implement any move.

Client/game_sa/CPedSoundSA.h
TracerDS: to distinguish it from the normal sound

Client/game_sa/CPedSA.cpp
TracerDS: `std::move` transforms pedSound into a rvalue. pedSound need to be a copy and by `std::move`ing it, it kinda achieves it but I could've better written it now that I think of it

Client/game_sa/CPedSoundSA.h
tederis: Nonsence! When you're using a rvalue reference you should use it as a rvalue reference. 

```
CPedSoundSA(CPedSoundSAInterface* pInterface, bool bInstantiate = false) :
        m_pInterface(bInstantiate ? new CPedSoundSAInterface() : pInterface)
{
        if (bInstantiate)
        {
                SetVoiceID(pInterface->m_sVoiceID);
                SetVoiceTypeID(pInterface->m_sVoiceType);
        }
}
```
would be much better

Client/game_sa/CPedSA.cpp
tederis: This is counterintuitive and breaks principles of the C++ standard.

PR #3130, opened 2023-07-25
Add onExplosion event

botder:
This pull request adds the `onExplosion` event to the server-side and it can be cancelled. The event callback receives the absolute coordinates of the explosion, the explosion type, and the explosion origin (like a vehicle), if there is one.

I haven't tested this event in a two-or-more player scenario yet, but it worked locally for me.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1619,6 +1619,7 @@ void CGame::AddBuiltInEvents()
     // Other events
     m_Events.AddEvent("onSettingChange", "setting, oldValue, newValue", NULL, false);
     m_Events.AddEvent("onChatMessage", "message, element", NULL, false);
+    m_Events.AddEvent("onExplosion", "x, y, z, type, origin", nullptr, false);
 
     // Weapon events
     m_Events.AddEvent("onWeaponFire", "", NULL, false);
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2627,112 +2628,120 @@ void CGame::Packet_DestroySatchels(CDestroySatchelsPacket& Packet)
...
+
+        CVector cameraPosition;
+        player->GetCamera()->GetPosition(cameraPosition);
+
+        // Is this players camera close enough to send?
+        if (IsPointNearPoint3D(explosionPosition, cameraPosition, MAX_EXPLOSION_SYNC_DISTANCE))
+        {
+            sendList.push_back(player);
         }
     }
```
TheNormalnij, line 2733: I suggest extract method here

sbx320:
Wouldn't this cause a desync for the player causing the explosion if cancelled (since it will have already happened on the source client)?

botder:
> Wouldn't this cause a desync for the player causing the explosion if cancelled (since it will have already happened on the source client)?

The client waits for the server packet to actually create the explosion.

Server/mods/deathmatch/logic/CGame.cpp
botder: Is that logic reused somewhere?

sbx320 (approved):

Server/mods/deathmatch/logic/CGame.cpp
TheNormalnij: To improve readability and self-documenting.

botder:
Well, this was too quick, I wanted to add the event to createExplosion function on serverside. I forgot to mention this.

botder:
Also, the origin element was not done right: it uses the closest player to an explosion, which is not the expected value in this event.

PR #3131, opened 2023-07-25
Allow allocating clump models

TheNormalnij:
This fixes issues with missing animations on custom models

```lua
engineRequestModel( 'clump', parentId )
```

You can replace a new model using `engineReplaceModel` or IMG functions.

Test resource:
[clump_model_test.zip](https://github.com/multitheftauto/mtasa-blue/files/12165405/clump_model_test.zip)


Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1710,6 +1710,22 @@ void CModelInfoSA::MakeTimedObjectModel(ushort usBaseID)
     CopyStreamingInfoFromModel(usBaseID);
 }
 
+void CModelInfoSA::MakeClumpModel(ushort usBaseID)
+{
+    CClumpModelInfoSAInterface* m_pInterface = new CClumpModelInfoSAInterface();
```
tederis: **m_** prefix should correspond only to class members

CrosRoad95:
if newly allocated models replaced with clump model has all atomics visible then this pull request in my opinion should be merged

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: I  agree with you. It's copy/pasted from bad code

Client/game_sa/CModelInfoSA.h
```diff
@@ -243,6 +243,16 @@ struct CTimeInfoSAInterface
     short m_wOtherTimeModel;
 };
 
+class CClumpModelInfoSAInterface : public CBaseModelInfoSAInterface
+{
+public:
+    union
+    {
+        char*  m_animFileName;
+        uint32 m_nAnimFileIndex;
```
botder:
```suggestion
        char*    m_animFileName;
        uint32_t m_nAnimFileIndex;
```

botder (changes requested):

botder (approved):

PR #3132, opened 2023-07-27
Fix build after CEF update

TheNormalnij:
As description.

Maybe it's better to use templates. But i'm not sure about that


tederis:
I would use templates, yes. Something like `template<typename T, typename U = typename std::enable_if<sizeof(T) == sizeof(wchar_t)>::type>`

TheNormalnij:
I was trying different templates but nothing works

Dutchman101 (approved):

PR #3133, opened 2023-07-29
Clean up multiplayer_sa project round 1

Merlin:
This PR removes COffsetsMP and all EU addresses


Client/multiplayer_sa/CMultiplayerSA_Direct3D.cpp
```diff
@@ -151,6 +141,6 @@ void _declspec(naked) HOOK_PostCreateDevice()
 //////////////////////////////////////////////////////////////////////////////////////////
 void CMultiplayerSA::InitHooks_Direct3D()
 {
-    EZHookInstall(PreCreateDevice);
```
TheNormalnij, line 154: What changes this?

Merlin: Previously, EZHookInstall was saving bytes to restore the hooks if a variable with the name "RESTORE_Bytes_HookName" existed for US/EU hooks. However, this functionality was only used for the PreCreateDevice hook. Therefore, I moved it into its own macro so we don't have unnecessary code generated for all hooks.

botder (approved):

PR #3140, opened 2023-08-03
Fixed setElementDimension resets attachElements

### LINKED ISSUE #3114: setElementInterior/Dimension resets attachElements
### Describe the bug

Attached elements will reset if element changes interior/dimension. Need to re-attach.

### Steps to reproduce

attachElements()
then setElementInterior/Dimension()

### Version

Multi Theft Auto v1.6-release-21889 (all prev versions too)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

adem-hosni:
Fixes #3114 
when use attachElements with setElementDimension all the attached elements are reset


Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -614,6 +614,10 @@ void CClientEntity::SetDimension(unsigned short usDimension)
     m_usDimension = usDimension;
 
     CLuaArguments Arguments;
+    for (uint i = 0; i < GetAttachedEntityCount(); i++)
+    {
+        GetAttachedEntity(i)->SetDimension(usDimension);
+    }
```
TracerDS, line 620: Use iterators instead + `auto` or `std::size_t` instead of `uint`

CrosRoad95 (commented):
lgtm

botder:
My problem with this solution is the recursive event trigger of `onClientElementDimensionChange`.

Proxy-99:
> My problem with this solution is the recursive event trigger of `onClientElementDimensionChange`.

@HYPER-00 can u work on this and add function like 
`GetAttachedEntity(i)->SetAttachElemenstDimension(usDimension);`

I think you should include server side too 
Server\mods\deathmatch\logic\CElement.cpp

TracerDS (changes requested):

PR #3144, opened 2023-08-06
Cleanup multiplayer_sa project

CrosRoad95:


Client/multiplayer_sa/StdInc.h
```diff
@@ -11,28 +11,16 @@
 #include "SharedUtil.MemAccess.h"
 #include <stdio.h>
 
-#include <algorithm>
-#include <list>
-#include <map>
-#include <string>
-#include <vector>
-
```
botder, line 19: Keep standard C++ headers here if they're used in the project.

Client/multiplayer_sa/multiplayersa_init.cpp
```diff
@@ -10,9 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
-#include <net/CNet.h>
 #define DECLARE_PROFILER_SECTION_multiplayersa_init
-#include "profiler/SharedUtil.Profiler.h"
```
botder, line 15: This is necessary for the profiler, keep it please.

botder (changes requested):

Client/multiplayer_sa/StdInc.h
CrosRoad95: they are not used

Client/multiplayer_sa/multiplayersa_init.cpp
CrosRoad95: this is also unused so it can be removed, "#define DECLARE_PROFILER_SECTION_multiplayersa_init" probably too

botder: Profiler integration is done on the build server, don't ask me why, but it's there. Please keep the lines.

botder (approved):

PR #3146, opened 2023-08-06
Fix #3143: Ped revives when syncer changes

### LINKED ISSUE #3143: Ped revives when syncer changes
### Describe the bug

isPedDead = true on server
isPedDead = false for new syncer
isPedDead = true for remote player

Ped is standing on screen of both players after the change

https://github.com/multitheftauto/mtasa-blue/assets/59518113/aaa813af-651a-49d7-8fb7-b29d8ada39b4

### Steps to reproduce

1. Kill ped
2. Change syncer

### Version

Client and server: 1.6.0, 1.5.9, 1.5.8 
Not tested in earlier versions

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

Zangomangu:
Small fix for #3143
When ped is killed and synced by player with some latency to server, the bug occurs.

The wasted packet is sent immediately as ped dies which increments sync context on server.
Before the new sync context is bounced back, the client has sent ped sync containing old sync context but the new health value (0). This is ignored by server.

The health value (0) is not re-sent because only changed values are sent.
So on server the ped is dead but ped has health value > 0 forever, this is sent to clients (and new syncer in video example) causing them to revive clientside.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1970,6 +1970,8 @@ void CGame::Packet_PedWasted(CPedWastedPacket& Packet)
     if (pPed && !pPed->IsDead())
     {
         pPed->SetIsDead(true);
+        pPed->SetHealth(0.0f);
+        pPed->SetArmor(0.0f);
         pPed->SetPosition(Packet.m_vecPosition);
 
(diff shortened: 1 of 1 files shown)


botder (approved):

botder:
Fixes #3143

Issue #578, opened 2018-09-24
Changing vehicle model from doorless or "doorful" causes doors to fall off


ArranTuna:
**Describe the bug**
Changing vehicle model from doorless or "doorful" causes doors to fall off

**To reproduce**
1) Take a tractor.
2) Set its model to a 4 door vehicle.
3) Doors fall off.

**Expected behaviour**

**Screenshots**

**Version**

**Additional context**

From https://bugs.mtasa.com/view.php?id=9915

Issue #1318, opened 2020-03-25
Caddy (vehicle) plays incorrect door/entry animation


tut001:
**Describe the bug**
When entering Golf Caddy ID 457, it plays a door opening animation both when entering from passenger and driver side.

**To reproduce**
1. Spawn ID 457
2. Enter from either passenger or driver seat
3. Notice you're opening an invisible door

**Expected behaviour**
The player should enter the vehicle by directly stepping inside.

**Screenshots**
N/A

**Version**
Client: Multi Theft Auto v1.5.7-release-20460

**Additional context**
Qaisjp suspects the vs2019 issue may be related
Strix later found the answer, https://github.com/multitheftauto/mtasa-blue/issues/1318#issuecomment-669616427

zo3n:
Tractor as well
by the way noticed something about this invisible door:

you know when you normally enter a vehicle, and door remains open, so when you let go of W, the player closes the door automatically by also doing animation of doing it?
if this happens with invisible door of tractor, ped will attempt to close the door always, whenever letting go of W key, but it'll never close because it doesn't exist

zo3n:
Just to show it off better, here's a video demonstration:
https://streamable.com/u7p9m

qaisjp:
Thank you so much for the video! Please can you install https://nightly.mtasa.com/mtasa-1.5.7-rc-20417-20200301.exe and check if that still happens?

zo3n:
That fixed it. Both Caddy and Tractor act normal now.

qaisjp:
Thanks for confirming. Okay, definitely a VS 2019 problem. Release the kraken!

zo3n:
Uh, problem reappeared in version 20488

Dutchman101:
> 
> 
> Uh, problem reappeared in version 20488

No.. it just hasn't been fixed yet, the build you tested with (which you said "fixed" it) was simply an old build from before upgrading to VS2019.

This is probably undefined behavior somewhere that gets exposed with VS2019 only, and it needs to be located in order to get fixed.

ccw808:
Does the bug occur with a custom (debug or release) build?

ghost:
> 
> 
> Does the bug occur with a custom (debug or release) build?

I can reproduce it on both, and sometimes this happens: https://cdn.discordapp.com/attachments/366384007535001612/725386799610069083/2020-06-24_21-23-39.mp4
[...]

PR #3150, opened 2023-08-09
Fix vehicle doors after change model #578

FileEX:
This PR fixes issue #578


botder:
Why were the previous PRs/commits reverted?

FileEX:
> Why were the previous PRs/commits reverted?

https://github.com/multitheftauto/mtasa-blue/issues/1318 - It was 3 years ago, when I had very little experience in programming and instead of investigating why this is happening, I just removed the switch that was responsible for setting "missing doors" for vehicles i.e. tractor, caddy, etc. Now as you can see this the code is significantly different from that of the reverted PR.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1003,9 +1003,13 @@ void CClientVehicle::SetModelBlocking(unsigned short usModel, unsigned char ucVa
...
-        // Are we swapping from a vortex or skimmer?
-        bool bResetWheelAndDoorStates = (m_usModel == VT_VORTEX || m_usModel == VT_SKIMMER ||
-                                         (m_eVehicleType == CLIENTVEHICLE_PLANE && m_eVehicleType != CClientVehicleManager::GetVehicleType(usModel)));
+        // Are we swapping from a vehicle without doors?
+        bool bResetWheelAndDoorStates =
+            ((m_usModel == VT_VORTEX || m_usModel == VT_TRACTOR || m_usModel == VT_CADDY || m_usModel == VT_BAGGAGE || m_usModel == VT_BANDITO ||
+              m_usModel == VT_BFINJECT || m_usModel == VT_DOZER || m_usModel == VT_FORKLIFT || m_usModel == VT_KART || m_usModel == VT_MOWER ||
+              m_usModel == VT_QUAD || m_usModel == VT_RCBANDIT || m_usModel == VT_RCCAM || m_usModel == VT_RCGOBLIN || m_usModel == VT_RCRAIDER ||
+              m_usModel == VT_RCTIGER || m_usModel == VT_BLOODRA) ||
+             m_eVehicleType != CClientVehicleManager::GetVehicleType(usModel));
```
TheNormalnij: Why don't you use `!HasDoors(m_usModel)` ?

FileEX: It's a good idea

botder (approved):

PR #3154, opened 2023-08-11
Fixed wheel visibility when using setVehicleWheelStates #1556

### LINKED ISSUE #1556: setVehicleWheelStates reset component visible
**Describe the bug**
When hide some wheel components and change state of other wheel, the component will be show.

**To reproduce**
1. Hide wheels component
```lua
for k,v in pairs(getVehicleComponents(vehicle)) do
    if (k ~= 'wheel_rf_dummy') then
        setVehicleComponentVisible(vehicle,k,false)
   end
end
```
2. Change one wheel state
```lua
setVehicleWheelStates(-1,-1,0,-1)
```
All invisible wheels will be visibled

**Expected behaviour**
setVehicleWheelStates shouldn't affect for no-changed wheels visible state.

**Screenshots**

**Version**
1.5.7

**Additional context**
When use fixVehicle function, this bug doesn't occur.

FileEX:
This PR fixes #1556 

Unfortunately, I don't know much about synchronization yet, so I don't know how this affects synchronization. I will be grateful for any tests


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1464,6 +1464,30 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
 
                 // Update the wheel's visibility
                 m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING));
+
+                // restart wheels component visibility (fix #1556)
+                std::map<SString, SVehicleComponentData>::iterator iter = m_ComponentData.begin();
```
tederis: There can be used auto type deduction.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1464,6 +1464,30 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
...
+
+                for (; iter != m_ComponentData.end(); iter++)
```
tederis: Why not use a structured binding instead of iterators?

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1464,6 +1464,30 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
...
+                {
+                    SString strComponentName = (*iter).first;
+                    if (strComponentName.BeginsWith("wheel_"))
```
tederis: You're creating a new SString object for `"wheel_"` in every iteration. To eliminate this you can create a static string object somewhere and use it here.

Client/mods/deathmatch/logic/CClientVehicle.cpp
FileEX: I based it on a code fragment that does the same thing (fixes wheel visibility) but within the fixVehicle function. Wanted to maintain consistency throughout the code and the rest of the project.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1464,6 +1464,29 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
 
                 // Update the wheel's visibility
                 m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING));
+
+                // restart wheels component visibility (fix #1556)
+                static SString strComponentName;
+                static unsigned char ucComponentWheel;
```
tederis: In the conversation above I meant that `"wheel_"` implicitly creates a new string every time `BeginWith` called. And it's better to create a static object for `"wheel_"`. Something like: `const static SString WHEEL_PREFIX("wheel_");`. There is not need for static variables `strComponentName `and `ucComponentWheel`.

FileEX: Okay, done!

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1464,6 +1464,30 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
 
                 // Update the wheel's visibility
                 m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING));
+
+                // restart wheels component visibility (fix #1556)
+                static const SString strWheelPrefix = "wheel_";
+
+                for (const auto& [key, value] : m_ComponentData)
+                {
+                    SString strComponentName = key;
```
tederis: `strComponentName` is just a copy of `key`. This copy is unnecessary, you can use `key` directly. I also recommend to change `key` and `value` to more reasonable names.

FileEX: I found a much simpler way to fix this bug 😄

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1463,7 +1463,18 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
                 m_pVehicle->GetDamageManager()->SetWheelStatus((eWheelPosition)(ucWheel), ucGTAStatus);
 
                 // Update the wheel's visibility
-                m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING));
+                SString componentName;
+
+                if (ucWheel == FRONT_LEFT_WHEEL)
```
TheNormalnij: Typically we use switch case for this.
Also this part can be described as function 'GetComponentNameForWheel'. This will improve readability, because SetWheelStatus will have less code/context.

FileEX: Done

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1472,6 +1473,32 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
...
+//
+SString CClientVehicle::GetComponentNameForWheel(unsigned char ucWheel)
+{
+    SString componentName;
+
+    switch (ucWheel)
+    {
+        case FRONT_LEFT_WHEEL:
+            componentName = "wheel_lf_dummy";
+            break;
```
TheNormalnij: This works?
`return "wheel_lf_dummy"'

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -282,6 +282,7 @@ class CClientVehicle : public CClientStreamElement
     int           GetWheelFrictionState(unsigned char ucWheel);
     unsigned char GetPanelStatus(unsigned char ucPanel);
     unsigned char GetLightStatus(unsigned char ucLight);
+    SString       GetComponentNameForWheel(unsigned char ucWheel);
```
TheNormalnij:
```suggestion
    const SString GetComponentNameForWheel(unsigned char ucWheel) const noexcept;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
FileEX: You mean returning in switch?

Pirulax: Yeah that would be a lot nicer.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1463,7 +1463,8 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
                 m_pVehicle->GetDamageManager()->SetWheelStatus((eWheelPosition)(ucWheel), ucGTAStatus);
 
                 // Update the wheel's visibility
-                m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING));
+                SString componentName = GetComponentNameForWheel(ucWheel);
+                m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, (ucStatus != DT_WHEEL_MISSING ? m_ComponentData[componentName].m_bVisible : false));
```
Pirulax:
```suggestion
                m_pVehicle->SetWheelVisibility((eWheelPosition)ucWheel, ucStatus != DT_WHEEL_MISSING && m_ComponentData[GetComponentNameForWheel(ucWheel)].m_bVisible);
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1472,6 +1473,32 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
...
+        case REAR_LEFT_WHEEL:
+            componentName = "wheel_lb_dummy";
+            break;
+        case REAR_RIGHT_WHEEL:
+            componentName = "wheel_rb_dummy";
+            break;
+    }
+
+    return componentName;
+}
```
Pirulax: I really think this shouldn't be in `CClientVehicle`.
Do we have a more generic class for these kinds of utility functions?

FileEX: There are many utility functions that are in this file. So why has no one thought of another class for functions like this before? - In addition, this function probably won't be used **anywhere** outside of this bugfix.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1472,6 +1472,24 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
...
+    {
+        case FRONT_LEFT_WHEEL:
+            return "wheel_lf_dummy";
+        case FRONT_RIGHT_WHEEL:
+            return "wheel_rf_dummy";
+        case REAR_LEFT_WHEEL:
+            return "wheel_lb_dummy";
+        case REAR_RIGHT_WHEEL:
+            return "wheel_rb_dummy";
+    }
```
tederis, line 1493: All code-paths must return a value

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1472,6 +1472,24 @@ void CClientVehicle::SetWheelStatus(unsigned char ucWheel, unsigned char ucStatu
     }
 }
 
+//
+// Returns component name for eWheelPosition enum
+//
+const SString CClientVehicle::GetComponentNameForWheel(unsigned char ucWheel) const noexcept
```
tederis: It is strongly advised to return values as non-const

FileEX: I don't know who to listen to. TheNormalnij wrote that I should give const. You write that it would be better not to const. I don't want to keep playing like this because everyone has a different opinion.

Client/mods/deathmatch/logic/CClientVehicle.cpp
FileEX: In this case, the function will never get anything other than a wheel, so I don't see any reason to return anything with the default value.

Client/mods/deathmatch/logic/CClientVehicle.cpp
tederis: `const` disables move semantics and should be avoided. In addition, it can break RVO and NRVO in some conditions.

Client/mods/deathmatch/logic/CClientVehicle.cpp
tederis: A value-returning function must return a value from all code paths; otherwise, it will result in undefined behavior. This is requirements of the standard.

Dutchman101:
It's a small patch, and it was code reviewed by 3 maintainers, so it's good to merge.

PR #3158, opened 2023-08-14
General refactor of MTA's code

TracerDS:
I changed most of `#ifdef WIN32` to `#ifdef _WIN32` because `WIN32` was not always defined in some files. `_WIN32` however is always *active*


Server/mods/deathmatch/logic/CMapManager.cpp
```diff
@@ -239,42 +224,35 @@ void CMapManager::SendMapInformation(CPlayer& Player)
...
+        EntityPacket.Add(pColShape);
     }
 
     marker.Set("ColShapes");
 
     // Add the water polys to the packet
-    CWaterManager*                pWaterManager = g_pGame->GetWaterManager();
-    list<CWater*>::const_iterator iterWater = pWaterManager->IterBegin();
-    for (; iterWater != pWaterManager->IterEnd(); iterWater++)
+    for (const auto& pWater : *g_pGame->GetWaterManager())
```
CrosRoad95, line 254: are you in this place by accident dereferecing CWaterManager that can cause unnecessary copy??

CrosRoad95:
i took a look at what did you do and i have few things to say:

```diff
-    reply << (unsigned char)(temp.str().length() + 1);
+   reply << (uchar)(temp.str().length() + 1);
 ```
 we should probably use types from https://en.cppreference.com/w/cpp/types/integer if we are going to refactor something, so  in this case uchar should be uint8_t

I'm not good at windows headers so i have a bit concern about changes you have made regarding that

If function has depth of 1, maybe 2 levels we probably should keep it as is.

Even if you compiled it, and it works, you should probably try to run some bigger gamemode to test more scenarious

I think if you are doing such big refactor, split it into smaller parts, sections or however you name them, so in one pr you refactor all types like "unsigned short" to "ushort", in the next you decreasing depth of functions ect. it's a general good advice so reviewers have easier time

**Comment by TracerDS** [comment]
> i took a look at what did you do and i have few things to say:
> 
> ```diff
> -    reply << (unsigned char)(temp.str().length() + 1);
> +   reply << (uchar)(temp.str().length() + 1);
> ```
> 
> we should probably use types from https://en.cppreference.com/w/cpp/types/integer if we are going to refactor something, so in this case uchar should be uint8_t
> 
> I'm not good at windows headers so i have a bit concern about changes you have made regarding that
> 
> If function has depth of 1, maybe 2 levels we probably should keep it as is.
> 
> Even if you compiled it, and it works, you should probably try to run some bigger gamemode to test more scenarious
> 
> I think if you are doing such big refactor, split it into smaller parts, sections or however you name them, so in one pr you refactor all types like "unsigned short" to "ushort", in the next you decreasing depth of functions ect. it's a general good advice so reviewers have easier time

well, then just a quick change from all `utype` to `u/int8-64` would fix that

File: Server/mods/deathmatch/logic/CMapManager.cpp (hunk 1 above)
Comment (TracerDS) [reply Server/mods/deathmatch/logic/CMapManager.cpp:254]: > are you in this place by accident dereferecing CWaterManager that can cause unnecessary copy??

this wont create a copy, it will only dereference WaterManager in order to get `begin` and `end` functions

File: Client/mods/deathmatch/logic/CLocalServer.cpp [hunk 2]
```diff
@@ -198,18 +198,16 @@ bool CLocalServer::Load()
...
-                {
-                    m_pResourcesCur->SetItemText(m_pResourcesCur->AddRow(), m_hResourcesCur, src->GetValue().c_str());
-                }
-            }
+            if (pNode->GetTagName().compare("resource"))
+                continue;
+
+            CXMLAttribute* src = pNode->GetAttributes().Find("src");
+            if (!src || !src->GetValue()[1])
+                continue;
```
tederis: Too many `continue` statements. In this particular case it makes things worse and absolutely redundant.
```
if (src && src->GetValue()[1])
    m_pResourcesCur->SetItemText(m_pResourcesCur->AddRow(), m_hResourcesCur, src->GetValue().c_str());
```
much better

Client/mods/deathmatch/logic/CLocalServer.cpp
```diff
@@ -344,17 +342,15 @@ bool CLocalServer::Save()
...
-        for (; iter != pRoot->ChildrenEnd(); ++iter)
+        for (const auto& pNode : *pRoot)
         {
-            CXMLNode* pNode = reinterpret_cast<CXMLNode*>(*iter);
-            if (pNode->GetTagName().compare("resource") == 0)
-            {
-                pRoot->DeleteSubNode(pNode);
-                iter = pRoot->ChildrenBegin();
-            }
+            if (pNode->GetTagName().compare("resource"))
```
tederis: The same here and below. `continue` statements overdose.

tederis:
The derefencing such as `for (const auto& pWater : *g_pGame->GetWaterManager())` is counterintuitive and makes it difficult to read the code. It's better to add a dedicated method for getting inner objects like `for (const auto& pWater : g_pGame->GetWaterManager()->GetWaters())`.

Client/mods/deathmatch/logic/CLocalServer.cpp
TracerDS: How does it makes things worse?
the line is already too big. And its easily changeable

tederis: `continue` should not be used in everywhere. And it's better to avoid the negation in expressions. It's simply difficult to read.

TracerDS: > `continue` should not be used in everywhere. And it's better to avoid the negation in expressions. It's simply difficult to read.

I beg to differ. Chained `if` statements are far worse than `continue` everywhere. Well, we can already see the results of it by a poor readability in current implementation.

tederis: There is no nested statements in the example above. It's okay when you `continue` in the beginning of a cycle, but only in the beginning. 
```
{
        ...

        CXMLAttribute* src = pNode->GetAttributes().Find("src");
	if (!src || !src->GetValue()[1])
                continue;

	m_pResourcesCur->SetItemText(m_pResourcesCur->AddRow(), m_hResourcesCur, src->GetValue().c_str());
}
```
is defenitely the overuse of `continue`. It takes more code lines and makes it difficult to read the code.

TracerDS: A bigger line doesn't increase readability. `continue` increases it by removing the additional tab there.
Negations in `if` statements are trivial. Trivial to read, write and understand. Though it should be refactored a little bit more than just removing nested `if`

tederis: > A bigger line doesn't increase readability. `continue` increases it by removing the additional tab there. Negations in `if` statements are trivial. Trivial to read, write and understand. Though it should be refactored a little bit more than just removing nested `if`

There is no "bigger line". Only considerations of clean and readable code. 
P.S. How many programmers, so many opinions, for sure. But I bet that most of the programmers have the same opinion that I have. But I do not exclude the possibility of the wrongness of such opinion though. My business is merely to communicate suggestions.

TracerDS: > There is no "bigger line". 

There literally is. A tab significantly increases line's length. And bigger lines are less readable.

> How many programmers, so many opinions, for sure. But I bet that most of the programmers have the same opinion that I have. But I do not exclude the possibility of the wrongness of such opinion though.

I agree with it. But I don't agree with removing `continue`

Pirulax: imo it's fine either way

Pirulax:
> The derefencing such as `for (const auto& pWater : *g_pGame->GetWaterManager())` is counterintuitive and makes it difficult to read the code. It's better to add a dedicated method for getting inner objects like `for (const auto& pWater : g_pGame->GetWaterManager()->GetWaters())`.

I agree on this one, it's not clear 100% clear what `*g_pGame->GetWaterManager()` is going to be.

Shared/XML/CXMLNodeImpl.cpp
```diff
@@ -123,80 +123,75 @@ void CXMLNodeImpl::DeleteAllSubNodes()
...
+    uint uiTemp = 0;
+    for (const auto& pNode : m_Children)
     {
-        if (dynamic_cast<CXMLNodeImpl*>((*iter))->GetNode()->ValueStr() == szTagName)
-        {
-            if (uiTemp == uiIndex)
-            {
-                return *iter;
-            }
+        if (dynamic_cast<CXMLNodeImpl*>(pNode)->GetNode()->ValueStr() != szTagName)
```
Pirulax: Do we need `dynamic_cast` here?

Shared/XML/CXMLNodeImpl.cpp
```diff
@@ -372,56 +367,41 @@ bool CXMLNodeImpl::CopyChildrenInto(CXMLNode* pDestination, bool bRecursive)
 
     // Iterate through our children if we have. Otherwize just copy the content
     if (m_Children.size() > 0)
```
Pirulax: !m_Children.empty()

Client/core/CSettings.cpp
```diff
@@ -3876,26 +3876,24 @@ void CSettings::SetChatColorValues(eChatColorType eType, CColor pColor)
...
+    CXMLNode* pPresetsRoot = pPresetsFile->GetRootNode();
+    if (!pPresetsRoot)
+        pPresetsRoot = pPresetsFile->CreateRootNode(CHAT_PRESETS_ROOT);
+
+    for (const auto& pNode : *pPresetsRoot)
     {
-        CXMLNode* pPresetsRoot = pPresetsFile->GetRootNode();
-        if (!pPresetsRoot)
-            pPresetsRoot = pPresetsFile->CreateRootNode(CHAT_PRESETS_ROOT);
+        if (pNode->GetTagName().compare("preset"))
```
Pirulax: might as well get rid of that `compare`:
`(pNode->GetTagName() != "compare"`

Server/core/CCrashHandler.cpp
```diff
@@ -12,14 +12,14 @@
...
-#ifdef WIN32
+#ifdef _WIN32
     #include "CExceptionInformation_Impl.h"
 #else
     #include <execinfo.h>
 #endif
 
 // clang-format off
-#ifndef WIN32
+#ifndef _WIN32
```
Pirulax, line 22: Why change `WIN32` to `_WIN32`?

Server/core/CCrashHandler.cpp
```diff
@@ -212,103 +209,103 @@ void CCrashHandler::DumpMiniDump(_EXCEPTION_POINTERS* pException, CExceptionInfo
...
+            // End of unhandled exception
+            fprintf(pFile, "%s", "** -- End of unhandled exception -- **\n\n\n");
+
+            // Close the file
+            fclose(pFile);
+        }
     }
+
+    // Free the DLL again
+    FreeLibrary(hDll);
```
Pirulax: you could also just wrap `hDll` into an `unique_ptr`.
That way you can add even more early outs

Server/core/CDynamicLibrary.cpp
```diff
@@ -11,37 +11,28 @@
...
 
-#ifdef WIN32
+#ifdef _WIN32
 
     // Copy the filename
     char szFilenameBuffer[MAX_PATH];
-    strncpy(szFilenameBuffer, szFilename, MAX_PATH);
+    // for "safer" version on windows
+    strncpy_s(szFilenameBuffer, szFilename, MAX_PATH);
+    szFilenameBuffer[MAX_PATH - 1] = '\0';
```
Pirulax: I'm pretty sure `strncpy_s` will alway snull terminate it

Server/core/CDynamicLibraryInterface.h
```diff
@@ -0,0 +1,74 @@
...
+#endif
+
+#include <cstdlib>
+#include <cstring>
+#include <cstdarg>
+#include <cstdio>
+
+using FuncPtr_t = void (*)();
+
+class CDynamicLibraryInterface
```
Pirulax: What's this?

Server/core/CServerImpl.cpp
```diff
@@ -329,101 +327,91 @@ int CServerImpl::Run(int iArgumentCount, char* szArguments[])
...
+
+    if (!m_XMLLibrary.Load(PathJoin(m_strServerPath, SERVER_BIN_PATH, szXMLLibName)))
     {
         // Couldn't load it
-        Print("ERROR: Loading network library (%s) failed!\n", szNetworkLibName);
+        Print("ERROR: Loading XML library (%s) failed!\n", szXMLLibName);
         Print("Press Q to shut down the server!\n");
         WaitForKey('q');
         DestroyWindow();
         return ERROR_NO_NETWORK_LIBRARY;
```
Pirulax: I know you didn't change it, but that's a misleading error code.

Server/core/CServerImpl.cpp
```diff
@@ -329,101 +327,91 @@ int CServerImpl::Run(int iArgumentCount, char* szArguments[])
...
+    {
+        // Couldn't find the InitNetServerInterface func
+        Print("ERROR: Initialization functions failed!\n");
+        Print("Press Q to shut down the server!\n");
+        WaitForKey('q');
+        DestroyWindow();
+        return ERROR_NETWORK_LIBRARY_FAILED;
+    }
+    // Make the modmanager load our mod
+    if (m_pModManager->Load("deathmatch", iArgumentCount, szArguments)) // Hardcoded for now
```
Pirulax: *Hardcoded for now*
(Been like that since 2005)

Pirulax (commented):
This is too big of a PR....
And I've been reviewing it for a good 20-30 minutes

Server/core/CCrashHandler.cpp
TracerDS: ~~Because `WIN32` might not be defined in some files. `_WIN32` is always defined on Windows. I had issues with `WIN32` not being defined~~

Because `_WIN32` is changed to `WIN32` in some projects which causes troubles with some functions

Server/core/CDynamicLibraryInterface.h
TracerDS: Replacement for `typedef`. Its just a C++ version of function alias

Server/core/CDynamicLibrary.cpp
TracerDS: > I'm pretty sure `strncpy_s` will alway snull terminate it

Based on [cppreference](https://en.cppreference.com/w/c/string/byte/strncpy):
> If count is reached before the entire array src was copied, the resulting character array is not null-terminated.

So yeah, null the last character

Shared/XML/CXMLNodeImpl.cpp
TracerDS: > Do we need `dynamic_cast` here?

it can be replaced with `static_cast` here

PR #3163, opened 2023-08-26
Streaming buffer restore + and fixes to `EngineStreaming.bufferSize`

Pirulax:
Addendum to #3086
- Now the code correctly restores the buffer size on reconnect.
- Added a function `engineStreamingRestoreBufferSize`
- Fixed `EngineStreming.bufferSize` class var's getter function
- Also refactored `SetStreamingBufferSize` [In an attempt to handle out-of-memory situations correctly, but it turns out the original memory alloc. code just crashes instead]


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -128,7 +131,7 @@ void CLuaEngineDefs::LoadFunctions()
         {"engineStreamingRestoreMemorySize", ArgumentParser<EngineStreamingRestoreMemorySize>},
         {"engineStreamingSetBufferSize", ArgumentParser<EngineStreamingSetBufferSize>},
         {"engineStreamingGetBufferSize", ArgumentParser<EngineStreamingGetBufferSize>},
-        {"engineStreamingSetModelCacheLimits", ArgumentParser<EngineStreamingSetModelCacheLimits>},
```
botder, line 131: Why was this one removed?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -186,10 +189,10 @@ void CLuaEngineDefs::AddClass(lua_State* luaVM)
         lua_classfunction(luaVM, "restoreMemorySize", "engineStreamingRestoreMemorySize");
         lua_classfunction(luaVM, "getBufferSize", "engineStreamingGetBufferSize");
         lua_classfunction(luaVM, "setBufferSize", "engineStreamingSetBufferSize");
-        lua_classfunction(luaVM, "setModelCacheLimits", "engineStreamingSetModelCacheLimits");
+        lua_classfunction(luaVM, "restoreBufferSize", "engineStreamingSetBufferSize");
```
botder: Why did you remove `setModelCacheLimits`?
Also, shouldn't you use `"engineStreamingRestoreBufferSize"` here?

botder (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
Pirulax: Um, perhaps I forgot to pull master.
I'll fix it.

botder (approved):

PR #3167, opened 2023-09-02
Discord Rich Presence support

znvjder:
Based on fork #2577 

I decided to copy the code from the base fork and rework it due to compilation problems.
What has changed:
- Most features tweaked, class rebuilt a bit. 
- Added enable/disable from client menu for Discord rich presence sync - disabled by default. 
- Any server owner who wants their own logo must create their own application (we need to create a tutorial on how to do this).
- Added a feature to manage from the server, but on the client side.
- Added the ability to display buttons to the discord-rpc code.
- Support emojis

Below are some of the features:
```lua
setDiscordApplicationID(string applicationid)
setDiscordRichPresenceDetails(string details)
setDiscordRichPresenceState(string state)
setDiscordRichPresenceAsset(string assetImage, string text)
setDiscordRichPresenceSmallAsset(string assetImage, string text)
setDiscordRichPresenceButton(number index, string text, string url)
resetDiscordRichPresenceData()
isDiscordRichPresenceConnected()
```

Example on Discord:
drpc

All that was left was to add a default resource to show additional data.
I have no idea what else to add, so I'm open to suggestions.


CrosRoad95:
here's our favourite issue back again, vendor libraries inflate pull requests, either: 1. download sources from some server, 2. use submodule
because now reviewing it is hard due to laggy browser :)

znvjder:
> here's our favourite issue back again, vendor libraries inflate pull requests, either: 1. download sources from some server, 2. use submodule
> 
> because now reviewing it is hard due to laggy browser :)

Sorry, but due to the fact that I have an edited version of discord-rpc I had to add it.

If you want i can redo it.

CrosRoad95:
i don't want to discuss the same issue over and over for every case, i want general solution

Dutchman101:
@CrosRoad95 Please leave comments like that to the project maintainers (MTA team/write access). Avoid dragging contributors down on their PR's, if those responsible want to update coding guidelines then they can do it separately and in its own discussion. Don't reply to this post, let's keep it on topic, thanks.

@znvjder Don't mind it, vendor\discord-rpc in this PR is fine

Who uses a mobile browser to review code anyways, im on desktop and it scrolls smoothly.
Nice PR, and heavily awaited!

Fernando-A-Rocha:
Commenting for support and exposure. Nice PR, thank you!

Client/core/CDiscordRichPresence.h
```diff
@@ -0,0 +1,65 @@
...
+    void SetAsset(const char* szAsset, const char* szAssetText, bool bIsLarge = false);
+    void SetAssetLargeData(const char* szAsset, const char* szAssetText);
+    void SetAssetSmallData(const char* szAsset, const char* szAssetText);
+
+    bool ResetDiscordData();
+    bool SetPresenceState(const char* szState, bool bCustom = false);
+    bool SetPresenceDetails(const char* szDetails, bool bCustom = false);
+    bool SetPresenceButtons(unsigned short int iIndex, const char* szName, const char* szUrl);
+    bool SetDiscordRPCEnabled(bool bEnabled);
+    bool IsDiscordRPCEnabled();
```
Lpsd: `IsDiscordRPCEnabled` can be made `const`

Client/sdk/core/CDiscordInterface.h
```diff
@@ -0,0 +1,30 @@
...
+    virtual bool ResetDiscordData() = 0;
+    virtual bool SetPresenceState(const char* szState, bool bCustom) = 0;
+    virtual void SetAssetLargeData(const char* szAsset, const char* szAssetText) = 0;
+    virtual void SetAssetSmallData(const char* szAsset, const char* szAssetText) = 0;
+    virtual void SetPresenceStartTimestamp(const unsigned long ulStart) = 0;
+    virtual bool SetPresenceButtons(unsigned short int iIndex, const char* szName, const char* szUrl) = 0;
+    //virtual void SetPresenceEndTimestamp(const unsigned long ulEnd) = 0;
+
+    virtual bool SetDiscordRPCEnabled(bool bEnabled = false) = 0;
+    virtual bool IsDiscordRPCEnabled() = 0;
```
Lpsd: remember to update `const` here

Lpsd (commented):
Mostly seems good to me, just a couple of small comments.

You also mentioned "due to the fact that I have an edited version of discord-rpc I had to add it." - in this case we would want to commit those changes to https://github.com/multitheftauto/discord-rpc and restore the `install_discord` script. You should be able to make a pull request to the `master` branch of that repository with the changes.

This way we retain the `install_discord` premake script & exclude the vendor files from the `mtasa-blue` repo, which is definitely preferable.

It would be really nice to have a test resource showing examples of all functions. They will need to be written for the wiki anyway.

Good work and thanks for picking up the branch!

Client/mods/deathmatch/logic/luadefs/CLuaDiscordDefs.cpp
```diff
@@ -0,0 +1,172 @@
...
+        return false;
+
+    return true;
+}
+
+bool CLuaDiscordDefs::SetAppID(std::string strAppID)
+{
+    int appIDLength = strAppID.length();
+
+    if (appIDLength > 32)
```
Lpsd: Need to check for `appIDLength < 1` here

Allerek: > Need to check for `appIDLength < 1` here

Discord APPids are at least 17 characters long

TracerDS: > > Need to check for `appIDLength < 1` here
> 
> Discord APPids are at least 17 characters long

That's exactly why you need to check if its `< 17` or empty (`< 1`)

Lpsd: Either way, the expression needs to match the invalid argument exception message. Either option is fine but I would prefer not to be bound to limits which Discord could change, so a minimum of `1` character is preferable here.

znvjder: As for the number of characters in the application ID, it varies, sometimes it is 17, 18 or even 20. So I would stick with 32.

Client/sdk/core/CDiscordInterface.h
znvjder: Done, sure we can do that, I don't see a problem. I will do it in a moment.

Lpsd (dismissed):

Lpsd:
To be merged along with this: https://github.com/multitheftauto/discord-rpc/pull/1

Lpsd:
The branch got borked by our simultaneous commits, it's fixed now. I've kept the `v` in the tag for consistency (I made a mistake when creating `3.4.2`).

Client/core/CDiscordRichPresence.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+#include "StdInc.h"
+#include "discord_rpc.h"
+#include "CDiscordRichPresence.h"
+
+constexpr char DEFAULT_APP_ID[] = "1139666477813866546";
+constexpr char DEFAULT_APP_ASSET[] = "mta_logo_round";
+constexpr char DEFAULT_APP_ASSET_TEXT[] = "Multi Theft Auto";
+constexpr char DEFAULT_APP_ASSET_SMALL[] = "";
+constexpr char DEFAULT_APP_ASSET_SMALL_TEXT[] = "";
```
AlexTMjugador: Nice work and PR, this feature has been awaited for a long time! :heart:

As far as I can see, scripts can customize rich presence information without changing the app ID, which in the eyes of Discord would establish a link between MTA and whatever server owners decide to use rich presence for. Should we force scripts to use their own app ID if they want to change the default rich presence data?

Also, out of curiosity, who manages the default app ID? Could it get unexpectedly deleted or misused by its owner?

znvjder: I aimed to enable the scripts to establish their own application, hence create their image (emblem, logo?) in Rich Presence and other essential information.

AlexTMjugador: Right. It is my understanding that application IDs should not be treated as private information, and any third party can try to extract a rich presence application ID and use it for a completely unrelated application, so these concerns probably don't mean much from a security standpoint. But I still find them interesting.

Lpsd: > Should we force scripts to use their own app ID if they want to change the default rich presence data?

Good thinking. We should stop customization unless the server is using their own application ID.

> Also, out of curiosity, who manages the default app ID?

I think that's @znvjder's app ID. We'll get one created under the MTA discord, nice spot.

Lpsd: Please address the first comment above @znvjder  - unless the server has set their own application ID they shouldn't be able to customize anything.

znvjder: The given ID in the default application was for the time of testing, sorry for that, so MTA team has to give their own ID.

Generally there is a variable to lock it, just need to add some conditions to the functions. But in general - do we need to block all features if we don't want to use a custom application?

Lpsd: Yeah I'd say block everything

Allerek: > The given ID in the default application was for the time of testing, sorry for that, so MTA team has to give their own ID.
> 
> Generally there is a variable to lock it, just need to add some conditions to the functions. But in general - do we need to block all features if we don't want to use a custom application?

I think every server should set-up their own application id.

znvjder: Already all features are blocked till we set up own application. :)

Lpsd (approved):

PR #3170, opened 2023-09-03
[Feature/Bug fix] Make possible to replace clump models

CrosRoad95:
This pr make possible to replace clump model with another clump model
```lua
addCommandHandler("replace", function()
    local model = 3425
    txd = engineLoadTXD("clumpmodel.txd")
    engineImportTXD(txd, model)
    local dff = engineLoadDFF("clumpmodel.dff")
    engineReplaceModel(dff, model)
    local obj = createObject(model,x + 5,y,z)
end)
```
Test resource:
[replaceclump.zip](https://github.com/multitheftauto/mtasa-blue/files/12505834/replaceclump.zip)
Blender project:
[blender project.zip](https://github.com/multitheftauto/mtasa-blue/files/12505836/blender.project.zip)

image


Client/game_sa/CRenderWareSA.cpp
```diff
@@ -496,29 +496,87 @@ bool AtomicsReplacer(RpAtomic* pAtomic, void* data)
...
 bool CRenderWareSA::ReplaceAllAtomicsInModel(RpClump* pNew, unsigned short usModelID)
 {
     CModelInfo* pModelInfo = pGame->GetModelInfo(usModelID);
 
     if (pModelInfo)
     {
-        RpAtomic* pOldAtomic = (RpAtomic*)pModelInfo->GetRwObject();
+        RwObject*                  pRwObject = pModelInfo->GetRwObject();
+        static auto                CModelInfo_ms_modelInfoPtrs = (CBaseModelInfoSAInterface**)ARRAY_ModelInfo;
+        CBaseModelInfoSAInterface* pBaseModelInfo = CModelInfo_ms_modelInfoPtrs[usModelID];
```
TheNormalnij: Can you use CGameSA::GetModelInfo here?

botder:
Why do you have to move all atomics from the old clump to the new one? Wouldn't replacing just the RwObject pointer be enough?

CrosRoad95:
> Why do you have to move all atomics from the old clump to the new one? Wouldn't replacing just the RwObject pointer be enough?

from what i know, just swapping pointer wont be enough

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -496,29 +496,86 @@ bool AtomicsReplacer(RpAtomic* pAtomic, void* data)
     return true;
 }
 
+static auto CModelInfo_ms_modelInfoPtrs = (CBaseModelInfoSAInterface**)ARRAY_ModelInfo;
```
TheNormalnij, line 499:
```suggestion
```

CrosRoad95:
as i'm testing it  i encountered potential problem, unsupported feature, idk. Because i created scene in blender that involve using models hierarchy, connected few cubes together and exported them, but all frames are now called "Cube" even they have different names in blender ( maybe while exporting, exporter took some name and used it for all? )
 
The question is: 
1. can [dragondff](https://github.com/Parik27/DragonFF) export models with preserved hierarchy? Maybe models in gta are not inteded to have it?
2. Should it be supported in mta? as you can see my later pr let you move frames so having hierarchy can be beneficial in some cases

CrosRoad95:
@botder I'd like to ask for feedback

botder:
> I'd like to ask for feedback

I don't have the necessary knowledge about models/RenderWare to judge this.

CrosRoad95:
@Pirulax and @tederis may know more about it

tut001:
@CrosRoad95 
> 1. can [dragondff](https://github.com/Parik27/DragonFF) export models with preserved hierarchy? Maybe models in gta are not inteded to have it?

Max's DFF plugin does have a 'user data' export setting for frames, a 'node names' setting for object names I am guessing (or dummy helpers), and a separate 'object names' setting which I assume does the same

To my knowledge GTA uses multiclump objects together with object IFP's to animate things like soda factory signs and oil pumps in the desert

It does look like it preserves object names when inspected with RW Analyze
1.dff = node names false
2.dff = node names true
3.dff = node names true, object names true
4.dff = node names true, user-data set to 'frame' true
[name test.zip](https://github.com/multitheftauto/mtasa-blue/files/12705774/name.test.zip)

PR #3172, opened 2023-09-03
Feature: dxDrawModel3d

CrosRoad95:
This pull request adds function to draw arbitrary model in simillar fasion as other dxDraw* functions

syntax: `bool dxDrawModel3d(model, position, rotation, optional scale)` ( draw in onClientRender/onClientPreRender loop )

You can draw as many objects as you want, it has no limits both amount of models and draw distance.
Rendered models has no collision, they look exactly the same as objects created using "createObject" ( possibly transparency will be an issue )
use cases:
1. draw vegetation
2. draw ped/vehicles visual accessories
3. draw miscellaneous objects on ground as an items player can pick up

example: `dxDrawModel3d(3276, -719.64984, 951.26685, 12.13281, 0, 0,90)` - as you can see, it look 1:1 as createObject
image
image

No limits:
image
image


Einheit-101:
This is nice. I guess you trade performance for free object streaming slots?

CrosRoad95:
you can call it that

Pirulax:
Eh I'd argue this is faster than having objects, as they have a lot of overhead.
These don't have collisions either, so there's that.

Client/mods/deathmatch/logic/luadefs/CLuaDrawingDefs.cpp
```diff
@@ -64,6 +75,7 @@ void CLuaDrawingDefs::LoadFunctions()
         {"dxSetAspectRatioAdjustmentEnabled", DxSetAspectRatioAdjustmentEnabled},
         {"dxIsAspectRatioAdjustmentEnabled", DxIsAspectRatioAdjustmentEnabled},
         {"dxSetTextureEdge", DxSetTextureEdge},
+        {"dxDrawModel3d", ArgumentParser<DxDrawModel3d>},
```
Lpsd: rename to `dxDrawModel3D` for case consistency with other functions

Lpsd (changes requested):

Lpsd (approved):

tederis:
There is a lot of commented out lines. I doubt that it should be merged in this "dirty" condition. But now it is too late to worry anyway.

Lpsd:
It's never too late to worry. https://github.com/multitheftauto/mtasa-blue/pull/3210

onikyannn:
Object is rendered at zero coordinates in any dimension other than 0.

Xenius97:
What about dxDrawModel2D?
Non shadered version of https://wiki.multitheftauto.com/wiki/Resource:Object_preview

tederis:
> What about dxDrawModel2D? Non shadered version of https://wiki.multitheftauto.com/wiki/Resource:Object_preview

It will be enough just to call dxDrawModel3d inside a render target.

PlatinMTA:
> It will be enough just to call dxDrawModel3d inside a render target.

wiki says no rendertarget support

tederis:
> > It will be enough just to call dxDrawModel3d inside a render target.
> 
> wiki says no rendertarget support

Of course there is no RT support now. But it can be easily added.

Dutchman101:
Reverted in 5b44a09 as a result of dev discord feedback:

_TEDERIs — Today at 2:24 PM
I respect the efforts of CrosRoad95. But I think #3172 (dxDrawModel3d feature) should be revoked. It's very buggy now, has a lot of internal problems and even cause crashes. It renders an object randomly (due to the nature of CClientGame::GameEntityRenderHandler) which leads to an unpredictable appearience(and can even disappear). Even after #3210 it contains unused code(see HOOKPOS_CRenderer__RenderFadingInEntities_MID). And it can crash the game(see model handling in DxDrawModel3D). This is nice PR, but it still has a long way to go before it is ready for production use_

PR #3174, opened 2023-09-05
Feature: Changing model geometry at runtime

CrosRoad95:
# The goal:
Add option to do basic modeling inside mta without any external tools. Provide for people tools  that help them easier and better integrate their models with rest of the world

# Changes in api

## New:
```lua
model-frame-hierarchy engineModelGetFrameHierarchy(modelId) -- return frame hierarchy in format { { frame, parentFrame}, ... }, returns empty table `{}` for atomic models
geometry-info engineModelGetFrameGeometry(modelId, frameName) -- return information about frame geometry, vertices count, triangles count and more
geometry-data engineModelGetFrameGeometry(modelId, frameName) -- return geometry data, vertices position, triangles
bool engineModelFrameSetVertexPosition(modelId, frameName, vertexIndex, x,y,z) -- queue setVertexPosition change for given model and frame
bool engineModelFrameSetVertexColor(modelId, frameName, vertexIndex, color) -- queue setVertexColor change for given model and frame
bool engineModelFlushChanges(modelID [, frameName]) -- updates visually model, don't use it more often than once every frame. If FrameName not provided, entire model get updated
dff-raw-data engineExportModel( modelID ) -- return raw data of the model

TODO:
1. Add function check if model is clump, or atomic. If atomic -> you don't need to provide frame name
2. Changing uv maps, night lighting, material id and rest of properties
3. Verify if it is required to update model bounding box
4. Add function to list all materials
5. (maybe) Add function to change lighting for all vertices with option to filter it by material id
6. Probably engineModelFlushChanges could return at second argument how many changes were commited
7. Test it with custom models
```

## Changes:
```
- ProcesLineOfSight now can return triangle index and it's vertices indices, 4 more return arguments
  - TODO: return frame name
```

# Scope of this pull request:
1. Getting basic informations about models,
8. Just modyfing model basic geometry properties, like vertices positions, color, materials ect.
9. Removing/Adding functions will be added in future pull requests

# Why api looks like above?
Wanted to avoid adding tons of functions, don't want to make api looks like:
```lua
local atomics = engineGetModelAtomics(1337)
local geometry = engineGetAtomicGeometry(atomics[1])
local ... = engineGetGeometrySomething(geometry)
```
instead, i treat atomic, geometry and frame as one thing, i'm aware that 1 frame can refer multiple geometries, atomics, atomics can have multiple of geometries ect ect ect.

# Examples
## Example 1
Changes windmill vertices to random positions +-10 units:
```lua
local hierarchy = engineModelGetFramesHierarchy(3425);        
local frame = hierarchy[1][1]
for i=1,200 do
  engineModelFrameSetVertexPosition(3425, frame, i, math.random(-10, 10), math.random(-10, 10), math.random(-10, 10))
end
engineModelFlushChanges(3425, frame)
```
and your model will look like on video:
https://www.youtube.com/watch?v=HYqlAipt5pI

## Example 2
If you want to vertex paint model using mouse, use:
```lua
function render()
    if(not getKeyState("mouse1") and not getKeyState("mouse2"))then
        return
    end

    local cx,cy = getCursorPosition()
    local swx,swy,swz = getWorldFromScreenPosition(cx * sx, cy * sy, 3)
    local ewx,ewy,ewz = getWorldFromScreenPosition(cx * sx, cy * sy, 100)
    local data = {processLineOfSight ( swx,swy,swz, ewx,ewy,ewz, true, true, true, true, true, false, false, false, nil, false, false, true)}

    local color = tocolor(255, 255, 255, 255)
    if(getKeyState("mouse2"))then
        color = tocolor(math.random(255),math.random(255),math.random(255),math.random(255))
    end

    if(data[5] and data[24])then
        local model = getElementModel(data[1])
        engineModelFrameSetVertexColor(model, "", data[24], color)
        engineModelFrameSetVertexColor(model, "", data[25], color)
        engineModelFrameSetVertexColor(model, "", data[26], color)
        engineModelFlushChanges(model, "")
    end
end
```
It gives following effect: https://www.youtube.com/watch?v=lQQb7Uo8urw

## Example 3
```lua
local hierarchy = engineModelGetFramesHierarchy(3425);        
local frame = hierarchy[1][2]
for i=1,200 do
    engineModelFrameSetVertexPosition(3425, frame, i, math.random(-1, 1), math.random(-1, 1), math.random(-1, 1))
end
engineModelFlushChanges(3425, frame)

local rawData = engineExportModel(3425)
local f;
if(not fileExists("out.dff"))then
    f = fileCreate("out.dff")
else
    f = fileOpen("out.dff")
end
fileWrite(f, rawData)
fileClose(f)
```
Modyfing first 200 vertices of windmill bottom frame, exports modified model to .dff
image

# What do i need?
- how api should looks like? i use "Frame" as frame/atomic/geometry right now for simplification, don't want to add functions such a "getFrameAtomic", "getAtomicGeometry" ect. If 1 frame has multiple geometries then current solution wont work
- initial code review, if current approach is okey. Created some class for update simillar to batchers
- maybe some feedback? :)
- feedback
- and feeedback
- Tederis, Piralux, TheNormalnij, heard you have some experience in that area? :)

### DIFF
--- a/Client/game_sa/CRenderWareSA.cpp
+++ b/Client/game_sa/CRenderWareSA.cpp
@@ -1029,16 +1029,16 @@ void RwFrameDump(RwFrame* parent, std::vector<std::vector<std::string>>& frames)
     }
 }
 
-void CRenderWareSA::GetFrameHierarchy(RpClump* pRoot, std::vector<std::vector<std::string>>& frames)
+void CRenderWareSA::GetFrameHierarchy(RwObject* rwObject, std::vector<std::vector<std::string>>& frames)
 {
-    if (RpGetFrame(pRoot) == nullptr)
+    if (RpGetFrame(rwObject) == nullptr)
         return;
-    RwFrameDump(RpGetFrame(pRoot), frames);
+    RwFrameDump(RpGetFrame(rwObject), frames);
 }
 
-bool CRenderWareSA::GetFrameGeometryInfo(RpClump* pRoot, std::string& frameName, SFrameGeometryInfo& info)
+bool CRenderWareSA::GetFrameGeometryInfo(RwObject* rwObject, std::string& frameName, SFrameGeometryInfo& info)
 {
-    RpAtomic* pAtomic = GetAtomicFromFrameName(pRoot, frameName);
+    RpAtomic* pAtomic = GetAtomicFromFrameName(rwObject, frameName);
     if (pAtomic == nullptr)
         return false;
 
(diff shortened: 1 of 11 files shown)


Client/game_sa/CRenderWareSA.cpp
```diff
@@ -986,3 +1013,236 @@ void CRenderWareSA::RwMatrixSetScale(RwMatrix& rwInOutMatrix, const CVector& vec
...
+    m_mapGeometryFrameUpdate[frameName].m_vecVertexSetColor.clear();
+    RpGeometryUnlock(pGeometry);
+}
+
+typedef uint32_t(__cdecl* RpGeometryRegisterPluginStream_t)(int pluginID, void* readBC, void* writeCB, void* getSizeCB);
+
+#define RWPLUGINOFFSET(_type, _base, _offset) ((_type*)((unsigned char*)(_base) + (_offset)))
+
+
+bool CRenderWareSA::ExportModel(uint16_t usModelId, std::string& outString)
```
Pirulax, line 1227: You might want to refactor/remove this
Also, use uint32 for model ids

Client/game_sa/CRenderWareSA.h
```diff
@@ -21,6 +22,56 @@ struct RpAtomic;
...
 struct STexInfo;
 struct STexTag;
+struct RpGeometry;
+
+struct SFrameGeometryInfo
+{
+public:
+    int texCoordsCount;
+    int trianglesCount;
+    int verticesCount;
```
Pirulax, line 32: use unsigned if possible

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -331,7 +331,7 @@ int CLuaWorldDefs::ProcessLineOfSight(lua_State* L)
                     }
                 }
 
-                if (bIncludeExtraMateriaInfo && matInfo.valid)  { // 7 args
+                if (bIncludeExtraMateriaInfo && matInfo.valid)  { // 7+4 args
```
Pirulax, line 334: you also need to change the return at the end to include the additional 4 args

PR #3178, opened 2023-09-08
Add `ProcessLineAgainstMesh`

Pirulax:
This PR adds a new function `ProcessLineAgainstMesh`, that is supposed to do a raycast against an element's renderable mesh model [not the collision model!].
The same functionality is already present in #3099, but the latter is a little buggy due to the fact that the collision model is always simplified, and not exactly the same as the mesh, and this leads to situations where no hit is detected, even though the visible mesh is hittable.
Also, when one is interested in a specific element the overhead is a lot smaller [as we can skip all the collision detection done by `processLineOfSight`]
The reason I didn't make it be a standalone function in the beginning is that not all game entitites can be represented as `CClientEntity` [For example IPL objects].

```lua
float u, float v, string textureName, string frameName, float hitX, float hitY, float hitZ = processLineAgainstMesh(Element targetElement, float startX, float startY, startZ, float endX, float endY, float endZ)
```


CrosRoad95:
does it take an account animated models like 3425? i expect that when i use this function the results will match what i see. also, does it work for peds and vehicles?

Pirulax:
No clue, haven't tried.
It does work for vehicles, and I guess it should work for peds too.
afaik animations don't affect the geometry directly, but only what is rendered, so the answer is no I believe

Client/game_sa/CWorldSA.cpp
```diff
@@ -254,6 +254,167 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
...
+    c.dirOS = c.endOS - c.originOS;
+    c.minHitDistSq = c.dirOS.LengthSquared();            // By setting it to this value we avoid collisions that would be detected after the line segment
+    // [but are still ont the ray]
+
+    // Do raycast against the DFF to get hit position material UV and name
+    // This is very slow
+    // Perhaps we could parallelize it somehow? [OpenMP?]
+    const auto ProcessOneAtomic = [](RpAtomic* a, void* data)
+        {
+            const auto c = (Context*)data;
```
botder: Make this one a C++-style cast, please.

Client/game_sa/CWorldSA.cpp
```diff
@@ -254,6 +254,167 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
...
+                };
+
+            // Atomic not visible
+            if (!a->renderCallback || !(a->object.object.flags & 0x04 /*rpATOMICRENDER*/))
+            {
+                return true;
+            }
+
+            // Sometimes atomics have no geometry [I don't think that should be possible, but okay]
+            const auto geo = a->geometry;
```
botder: This comment applies to the code below too. Please only use `auto` where I don't have to guess the type in a code review.

Client/game_sa/CWorldSA.cpp
```diff
@@ -254,6 +254,167 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
...
+            const auto localEnd = ObjectToLocalSpace(c->endOS);
+
+#if 0
+            if (!CCollisionSA::TestLineSphere(
+                CColLineSA{localOrigin, 0.f, localEnd, 0.f},
+                reinterpret_cast<CColSphereSA&>(*RpAtomicGetBoundingSphere(a)) // Fine for now
+            )) {
+                return true; // Line segment doesn't touch bsp
+            }
+#endif
```
botder, line 343: Do we really need this block?

Client/game_sa/CWorldSA.cpp
```diff
@@ -254,6 +254,167 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
...
+
+    // It might be false if the collision model differs from the clump
+    // This is completely normal as collisions models are meant to be simplified
+    // compared to the clump's geometry
+    if (ret.valid = c.hitGeo != nullptr)
+    {
+        // Now, calculate texture UV, etc based on the hit [if we've hit anything at all]
+        // Since we have the barycentric coords of the hit, calculating it is easy
+        ret.uv = {};
+        for (auto i = 3u; i-- > 0;)
```
botder: Wouldn't a normal incrementing loop be enough?

Client/game_sa/CWorldSA.cpp
```diff
@@ -254,6 +254,167 @@ void ConvertMatrixToEulerAngles(const CMatrix_Padded& matrixPadded, float& fX, f
...
+            // Now, just interpolate
+            ret.uv += CVector2D{vtxUV->u, vtxUV->v} * c.hitBary[i];
+        }
+
+        // Find out material texture name
+        // For some reason this is sometimes null
+        const auto tex = c.hitGeo->materials.materials[c.hitTri->materialId]->texture;
+        ret.textureName = tex ? tex->name : nullptr;
+
+        const auto frame = RpAtomicGetFrame(c.hitAtomic);            // `RpAtomicGetFrame`
```
botder: Use the property type instead of `auto` + remove that comment.

botder (changes requested):

Client/game_sa/CWorldSA.cpp
Pirulax: I really don't think the type of it is important at all.
But okay.

Client/game_sa/CWorldSA.cpp
Pirulax: It doesn't bother anyone imho

Pirulax:
I've applied the requested suggestions.

botder (approved):

PR #3179, opened 2023-09-08
Add new world special property (fireballdestruct)

samr46:
Suggested by @Dutchman101 

This property toggles destruction of airplane / helicopter model on explosion (default is `true`). By default game destroys aircraft model when you collide with hard surface at too high speed or after direct missile hit (also if [blowVehicle](https://wiki.multitheftauto.com/wiki/BlowVehicle) is used with driver inside).

It should be easy to review. 2 nops.
~~I will increment sync version if it's not going to be possible to merge it before new build.~~

This PR includes small changes intended to fix `AC #1` false positive kicks introduced by adding server-side setWorldSpecialPropertyEnabled.

> setWorldSpecialPropertyEnabled("fireballdestruct", false)

Before (`true`):
before
After (`false`):
after

### DIFF
--- a/Client/game_sa/CGameSA.h
+++ b/Client/game_sa/CGameSA.h
@@ -215,6 +215,9 @@ public:
     bool IsBurnFlippedCarsEnabled() const noexcept { return m_isBurnFlippedCarsEnabled; }
     void SetBurnFlippedCarsEnabled(bool isEnabled);
 
+    bool IsFireballDestructEnabled() const noexcept { return m_isFireballDestructEnabled; }
+    void SetFireballDestructEnabled(bool isEnabled);
+
     unsigned long GetMinuteDuration();
     void          SetMinuteDuration(unsigned long ulTime);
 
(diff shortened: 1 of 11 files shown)


Client/sdk/game/CGame.h
```diff
@@ -213,6 +213,9 @@ class __declspec(novtable) CGame
     virtual bool IsBurnFlippedCarsEnabled() const = 0;
     virtual void SetBurnFlippedCarsEnabled(bool isEnabled) = 0;
 
+    virtual bool IsFireballDestructEnabled() const = 0;
```
botder:
```suggestion
    virtual bool IsFireballDestructEnabled() const noexcept = 0;
```

Client/game_sa/CGameSA.h
```diff
@@ -215,6 +215,9 @@ class CGameSA : public CGame
     bool IsBurnFlippedCarsEnabled() const noexcept { return m_isBurnFlippedCarsEnabled; }
     void SetBurnFlippedCarsEnabled(bool isEnabled);
 
+    bool IsFireballDestructEnabled() const noexcept { return m_isFireballDestructEnabled; }
+    void SetFireballDestructEnabled(bool isEnabled);
```
botder:
```suggestion
    bool IsFireballDestructEnabled() const noexcept override { return m_isFireballDestructEnabled; }
    void SetFireballDestructEnabled(bool isEnabled) override;
```

botder (changes requested):

botder (approved):

PR #3180, opened 2023-09-09
Add new function: aclObjectGetGroups

TracerDS:
This function will return all groups of an object.
Syntax: `table aclObjectGetGroups ( string theObject )`

Function returns `table` on success, `false` if something went wrong


botder:
Please use the new argument parser.

TracerDS:
> Please use the new argument parser.

Should I return a vector as a table then?

Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp
```diff
@@ -1049,3 +1051,36 @@ int CLuaACLDefs::OOP_isObjectInACLGroup(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+std::variant<std::vector<CAccessControlListGroup*>, bool> CLuaACLDefs::aclObjectGetGroups(SString strObject)
```
CrosRoad95: change type from `SString` to `std::string`, convert it later to "SString" if needed

Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.cpp
```diff
@@ -1049,3 +1051,37 @@ int CLuaACLDefs::OOP_isObjectInACLGroup(lua_State* luaVM)
...
+        objectType = CAccessControlListGroupObject::OBJECT_TYPE_RESOURCE;
+    }
+    else if (StringBeginsWith(szObjectAfterDot, "user."))
+    {
+        szObjectAfterDot += 5;
+        objectType = CAccessControlListGroupObject::OBJECT_TYPE_USER;
+    }
+    else
+    {
+        return false;
```
botder: This function should never anything else other than a list of groups. Throw an exception here.

Server/mods/deathmatch/logic/luadefs/CLuaACLDefs.h
```diff
@@ -48,4 +48,7 @@ class CLuaACLDefs : public CLuaDefs
     LUA_DECLARE(aclGroupRemoveObject);
     LUA_DECLARE_OOP(isObjectInACLGroup);
     LUA_DECLARE(hasObjectPermissionTo);
+
+    static std::variant<std::vector<CAccessControlListGroup*>, bool>
+        aclObjectGetGroups(std::string strObject);
```
botder: This function returns a vector on every return path now, get rid of the variant.

botder (approved):

PR #3189, opened 2023-09-18
Add new functions: `pathListDir`, `pathIsFile`, `pathIsDirectory`

### LINKED ISSUE #346: Getting files and directories in specified path
Years/two ago i made function for it but was rejected.
I want function to this serverside that return every file in path even isn't inside meta.xml and slightly different function clientside, this return only every file belonging to selected resource.

If function return path as raw string, then should be added function to convert pathes into better format ( example table ).

TracerDS:
Also refactored SharedUtil a little (`WIN32` might not be defined. `_WIN32` will always be defined)

Syntax: ~~`table fileListDir ( string path )`~~ `table pathListDir ( string path )`
returns table of entries on success, `nil` on failure

### Additional functions
~~`bool isFile ( string path )`~~ `bool pathIsFile ( string path )`
Returns true if path is a file, false if a directory or path doesn't exist

~~`bool isDirectory ( string path )`~~ `bool pathIsDirectory ( string path )`
Returns true if path is a directory, false if a file or path doesn't exist

### OOP functions
`path.listDir` => `pathListDir`
`path.isFile` => `pathIsFile`
`path.isDirectory` => `pathIsDirectory`

---

Resolves: #346


Pirulax:
Is this supposed to return a list of files in a directory?

TracerDS:
> Is this supposed to return a list of files in a directory?

Relative to a current resource, yes

botder:
Is the Unix code block functional with `if (!(dir = opendir("c:\\src\\")))`?
Also, don't you exclude the directory items in the `< C++17` block?
And last, how can the scripter differentiate between file and directory?

TracerDS:
> Is the Unix code block functional with `if (!(dir = opendir("c:\\src\\")))`? Also, don't you exclude the directory items in the `< C++17` block? And last, how can the scripter differentiate between file and directory?

1. Changed path to the actual path. Should be functional now
2. Now including all files in a directory without adding the directory name before it too
3. via `isFile` and `isDirectory` functions (comment edited)

botder:
Wouldn't it make more sense to append a directory separator for directory items?
Checking each item whether it's a directory or file seems expensive?

TracerDS:
> Wouldn't it make more sense to append a directory separator for directory items?

It might cause some issues when messing with directories directly via that function (unless mta takes care of that already)

> Checking each item whether it's a directory or file seems expensive?

No, it's not really expensive. 

If MTA doesn't have problems with directories ending with `\` | `/` then sure, I can add a separator for directory entries

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -1052,9 +1066,55 @@ int SharedUtil::File::Delete(const char* szFilename)
...
+    std::vector<std::string> entries;
+#if _HAS_CXX17
+    namespace fs = std::filesystem;
+    if (!DirectoryExists(szPath))
+        return {};
+
+    for (const auto& entry : fs::directory_iterator(szPath))
+    {
+        entries.push_back(entry.path().filename().string());
+    }
```
botder: This is going to include non-directory and non-regular-file file types (block file, character file, socket file, symlinks) if I am not wrong. How about we only allow `entry.is_regular_file()` and `entry.is_directory()` here?

Also, you're using the throwing filesystem functions.

TracerDS: > This is going to include non-directory and non-regular-file file types (block file, character file, socket file, symlinks) if I am not wrong. How about we only allow `entry.is_regular_file()` and `entry.is_directory()` here?
> 
> Also, you're using the throwing filesystem functions.

sure, we can do that. With throwing we can just use `try-catch` here

Pirulax:
I don't think `fileListDir` makes a lot of sense as a name.
Perhaps something like `getDirectoryEntries` or (or, I'd prefer to separate them, `getDirectoryFiles` and `getDirectoryFolders`, both with a `bool recursive` parameter as well.
Or, could even go with `pathGetFolders`, `pathGetFiles` [if we want to follow the `file` prefix convention thingy]

TracerDS:
> I don't think `fileListDir` makes a lot of sense as a name. Perhaps something like `getDirectoryEntries` or (or, I'd prefer to separate them, `getDirectoryFiles` and `getDirectoryFolders`, both with a `bool recursive` parameter as well. Or, could even go with `pathGetFolders`, `pathGetFiles` [if we want to follow the `file` prefix convention thingy]

its better to have one function that gets all entries that you can sort by using `isFile`/`isDirectory` functions.

DNL291:
There's the module FileSystem which already does the same and has some more features/functions btw.
Although could be added nativelly yes if the only purpose is to get files from a directory instead of using the module just for that.

TracerDS:
> There's the module FileSystem which already does the same and has some more features/functions btw.
> Although could be added nativelly yes if the only purpose is to get files from a directory instead of using the module just for that.

Using module for that will decrease performance (since you have to load the module, load its functions and events not to mention modules are very limited and you cant use almost any MTA functions there at all without refactoring the whole mta server).
Also that would only make the command work in server scripts only.

Implementing it directly in MTA solves all these issues

DNL291:
> > There's the module FileSystem which already does the same and has some more features/functions btw.
> > Although could be added nativelly yes if the only purpose is to get files from a directory instead of using the module just for that.
> 
> Using module for that will decrease performance (since you have to load the module, load its functions and events not to mention modules are very limited and you cant use almost any MTA functions there at all without refactoring the whole mta server). Also that would only make the command work in server scripts only.
> 
> Implementing it directly in MTA solves all these issues

Uhm I'm not sure if we really have these issues in the (FileSystem) module. But I do agree that having the functions client-side and being directly implemented into MTA is better.

tederis:
> > There's the module FileSystem which already does the same and has some more features/functions btw.
> > Although could be added nativelly yes if the only purpose is to get files from a directory instead of using the module just for that.
> 
> Using module for that will decrease performance (since you have to load the module, load its functions and events not to mention modules are very limited and you cant use almost any MTA functions there at all without refactoring the whole mta server). Also that would only make the command work in server scripts only.
> 
> Implementing it directly in MTA solves all these issues

I doubt that modules can affect performance in a context of filesystem functions. The main disadvantage that modules have - they don't exist on a client side. So the only way to add proposed in this PR functions is to make them a part of MTA.

But I personally would not say that these functions are really needed on a client side. Typically, server(and therefore client scripts) know which files exist on a client and how to reach them, which makes sense from the security perspective.

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -1052,9 +1066,60 @@ int SharedUtil::File::Delete(const char* szFilename)
...
+        {
+            if (entry.is_regular_file() || entry.is_directory())
+                entries.push_back(entry.path().filename().string());
+        }
+    }
+    catch (const fs::filesystem_error& exc) {}
+#else
+    #ifdef _WIN32
+    std::string search_path = szPath;
+    if (search_path.back() != '/')
```
tederis, line 1101: When a string is empty method `back` has undefined behaviour. I would recommend to add a check `!search_path.empty()`.

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -53,14 +60,21 @@ bool SharedUtil::FileExists(const SString& strFilename)
 //
 bool SharedUtil::DirectoryExists(const SString& strPath)
 {
-#ifdef WIN32
+#if _HAS_CXX17
+    namespace fs = std::filesystem;
+    return fs::is_directory(static_cast<const std::string&>(strPath));
```
tederis: Is static_cast really needed here?

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -29,22 +30,28 @@
...
 
 //
 // Returns true if the file exists
 //
 bool SharedUtil::FileExists(const SString& strFilename)
 {
-#ifdef WIN32
+#if _HAS_CXX17
+    namespace fs = std::filesystem;
+    return fs::is_regular_file(static_cast<const std::string&>(strFilename));
```
tederis: Is static_cast really needed here?

Shared/sdk/SharedUtil.File.h
```diff
@@ -102,6 +102,8 @@ namespace SharedUtil
     WString FromUTF8(const SString& strPath);
     SString ToUTF8(const WString& strPath);
 
+    std::vector<std::string> ListDir(const char* szPath) noexcept;
```
tederis, line 105: `noexcept` is dangerous here because `ListDir` method can throw an exception(bad_alloc for example).

TracerDS: Bad_alloc can happen all the time, doesnt mean we should surround all mta code into multiple huge try-catch blocks. Other than that, no exceptions are thrown in ListDir function (except the possibility of an empty string but its easily solvable)

tederis: `bad_alloc` is just an example of possible exceptions that can be thrown. `noexpect` in this case leads to `std::terminate` without even possibility to catch an exception somewhere upper in the stack. This is an example of so called ill-formed program.

tederis: `noexcept` shall not be placed on a method that can even hypothetically throw an exception.

TracerDS: Bad_alloc is the only exception here that can throw. All other are handled (fs functions throw under the fs_error class).
C functions dont throw at all. So there is nothing to worry about. If `bad_alloc` is thrown then mta cannot do anything anymore either way

Shared/sdk/SharedUtil.File.hpp
TracerDS: to `std::string`? yes. Or we can use `.c_str()` but I think it would perform slower because it has to convert it to string first internally.

Shared/sdk/SharedUtil.File.hpp
TracerDS: as below

Shared/sdk/SharedUtil.File.h
tederis: As I already said, `ill-formed, no diagnostic required`.

Shared/sdk/SharedUtil.File.hpp
tederis: This cannot be slower because the string copying will occur in any way. By using `c_str` you will achieve the same behaviour as you already have. `static_cast` is something that you should not use if you can.

Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    SString strAbsPath;
+
+    CResource* pResource = pLuaMain->GetResource();
+    if (!CResourceManager::ParseResourcePathInput(path, pResource, &strAbsPath))
+        return std::nullopt;
+
+    if (!DirectoryExists(strAbsPath))
+        return std::nullopt;
+
```
TheNormalnij: Warnings will be useful here

Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+    if (!pLuaMain)
+        return false;
+
+    SString strAbsPath;
+
+    CResource* pResource = pLuaMain->GetResource();
+    if (!CResourceManager::ParseResourcePathInput(path, pResource, &strAbsPath))
+        return false;
```
TheNormalnij, line 88: Warning

Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp
```diff
@@ -0,0 +1,93 @@
...
```
TheNormalnij, line 107: Warning

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -1052,9 +1068,63 @@ int SharedUtil::File::Delete(const char* szFilename)
...
+    while ((ent = readdir(dir)))
+    {
+        entries.push_back(ent->d_name);
+    }
+    closedir(dir);
+    #endif
+#endif
+
+    return entries;
+}
```
TheNormalnij, line 1130:
```suggestion

#if __cplusplus >= 201703L
std::vector<std::string> SharedUtil::ListDir(const char* szPath) noexcept
{
    std::vector<std::string> entries;
    namespace fs = std::filesystem;
    if (!DirectoryExists(szPath))
        return {};

    try
    {
        for (const auto& entry : fs::directory_iterator(szPath))
        {
            if (entry.is_regular_file() || entry.is_directory())
                entries.push_back(entry.path().filename().string());
        }
    }
    catch (...) {} // catch all possible errors and ignore them

    return entries;
}
#endif
```
I think there is no reason to add dead code to the project.

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -29,38 +30,53 @@
...
-    return std::filesystem::is_regular_file(static_cast<const std::string&>(strFilename), ec);
+    return !(dwAtr & FILE_ATTRIBUTE_DIRECTORY);
+    #else
+    struct stat s;
+    if (!stat(strFilename.c_str(), &s))
+        return false;
+    return s.st_mode & S_IFREG;
+    #endif
 #endif
 }
```
TheNormalnij, line 57: Can we use 'fs::is_regular_file' only here?

Shared/sdk/SharedUtil.File.hpp
```diff
@@ -29,38 +30,53 @@
...
-    return std::filesystem::is_directory(static_cast<const std::string&>(strPath), ec);
+    return (dwAtr & FILE_ATTRIBUTE_DIRECTORY) != 0;
+    #else
+    struct stat s;
+    if (!stat(strPath.c_str(), &s))
+        return false;
+    return s.st_mode & S_IFDIR;
+    #endif
 #endif
 }
```
TheNormalnij: Same question for  "fs::is_directory"

Shared/sdk/SharedUtil.File.h
```diff
@@ -20,8 +20,8 @@ namespace SharedUtil
     //
     // Returns true if the file/directory exists
     //
-    bool FileExists(const SString& strFilename);
-    bool DirectoryExists(const SString& strPath);
+    bool FileExists(const SString& strFilename) noexcept;
+    bool DirectoryExists(const SString& strPath) noexcept;
```
TheNormalnij: SString -> std::string

TracerDS: Whats the point of SString if you dont use it anywhere?

SString inherits from std::string. Wouldnt make much difference

Shared/sdk/SharedUtil.File.hpp
TracerDS: Some mta projects are still limited to c++14. Better safe than sorry

Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp
TracerDS: What kind of warning here?

TheNormalnij: "Unable to parse the path."
"Directory does not exist."
These warnings will be helpful during debugging and show potential issues.

Shared/sdk/SharedUtil.File.h
TheNormalnij: We are trying to use only std::string in the new API. As far as I know, std::string -> SString casts have an allocation problem. So it's safe to use std::string here

Shared/sdk/SharedUtil.File.hpp
TheNormalnij: But this code is unused and should be removed when we get C++ 17 there.
The codebase becomes more complex due to dead code. I would recommend keeping only the necessary code that can be tested.

Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp
TracerDS: > "Unable to parse the path." "Directory does not exist." These warnings will be helpful during debugging and show potential issues.

sorry, but I've only done simple things with the parser. How would I throw a warning in ArgumentParser?

TheNormalnij: Did you try `m_pScriptDebugging->LogWarning(luaVM, "message");` ?

TracerDS: > Did you try `m_pScriptDebugging->LogWarning(luaVM, "message");` ?

`luaVM` is not accessible via ArgumentParser

TheNormalnij:
```
CLuaPathDefs::pathListDir(lua_State* luaVM, std::string path)
                                     ^^^^^^
 ```

Comment (TracerDS) [reply Shared/mods/deathmatch/logic/luadefs/CLuaPathDefs.cpp]: > ```
> CLuaPathDefs::pathListDir(lua_State* luaVM, std::string path)
>                                      ^^^^^^
> ```

oops, sorry my bad. I think Im too tired with it all 😄

File: Shared/sdk/SharedUtil.File.hpp (hunk 9 above)
Comment (TheNormalnij) [reply Shared/sdk/SharedUtil.File.hpp:1130]: After the most recent force push, have you made these changes back?

Comment (TracerDS) [reply Shared/sdk/SharedUtil.File.hpp:1130]: > After the most recent force push, have you made these changes back?

Nope. The code is used by `multiplayer_sa`. Cannot remove it.

Once `multiplayer_sa` gets converted to C++17+ we can remove old code

File: Client/mods/deathmatch/logic/CResource.h [hunk 13]
```diff
@@ -69,6 +69,7 @@ class CResource
     void                 SetResourceEntity(CClientEntity* pEntity) { m_pResourceEntity = pEntity; }
     class CClientEntity* GetResourceDynamicEntity() { return m_pResourceDynamicEntity; }
     void                 SetResourceDynamicEntity(CClientEntity* pEntity) { m_pResourceDynamicEntity = pEntity; }
+    SString              GetResourceDirectoryPath() { return GetResourceDirectoryPath(eAccessType::ACCESS_PUBLIC, ""); }
     SString              GetResourceDirectoryPath(eAccessType accessType, const SString& strMetaPath);
```
ghost, line 73: How about using std::string instead of SString?

ghost (changes requested):

Client/mods/deathmatch/logic/CResource.h
TracerDS: ~~Oops, yeah forgot about that part.
Sure~~

Cant do that yet. Lets just merge these changes and SString will be taken care of in the next PR that will refactor these functions

ghost: Okay, fair enough.
I will accept the modifications if you also solve the other conversations, as suggested by TheNormalnij.

TracerDS: As far as I see all conflict have been resolved

ghost (approved):
LGTM :shipit:

Dutchman101:
There's been many code reviews, so merging.
@TracerDS thanks, please add it to wiki though

PR #3190, opened 2023-09-18
Fix engineFreeModel regression caused by #1802

TheNormalnij:
engineFreeModel doesn't do any thing.
WTR:

```
crun engineRequestModel('vehicle')
crun engineFreeModel(74)
```


Dutchman101:
Fast-tracking this PR to get a stable build. It's only a few lines, and appears to resolve the issue without introducing another problem.

PR #3195, opened 2023-09-23
Fix engineFreeModel crashes

### LINKED ISSUE #3191: engineFreeModel Crashes client if object destroyed
### Describe the bug

The client just crashes when trying to call engineFreeModel if allocated object was be destroyed

### Steps to reproduce

1. Request model with engineRequestModel
2. Load and import any txd/dff with requested model
3. Сreate an object with imported model near you
4. Destroy the object
5. use engineFreeModel

### Version

v1.6-release-22215

### Additional context

[crash_test.zip](https://github.com/multitheftauto/mtasa-blue/files/12691238/crash_test.zip)

Test resource. Just use command `testCrash`

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #3191

### DIFF
--- a/Client/mods/deathmatch/logic/CClientModel.cpp
+++ b/Client/mods/deathmatch/logic/CClientModel.cpp
@@ -78,30 +78,49 @@ bool CClientModel::Allocate(ushort usParentID)
     return false;
 }
 
+// You can call it only in destructor for DFF.
 bool CClientModel::Deallocate()
 {
     if (!m_bAllocatedByUs)
         return false;
+
+    SetParentResource(nullptr);
+
     CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
     if (!pModelInfo || !pModelInfo->IsValid())
         return false;
 
-    SetParentResource(nullptr);
+    if (m_eModelType != eClientModelType::TXD)
+    {
+        // Remove model info
+        pModelInfo->DeallocateModel();
+    }
+
+    return true;
+}
+
+void CClientModel::RestoreEntitiesUsingThisModel()
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
+    if (!pModelInfo || !pModelInfo->IsValid())
+        return;
 
     switch (m_eModelType)
     {
         case eClientModelType::PED:
         case eClientModelType::OBJECT:
         case eClientModelType::VEHICLE:
-            return DeallocateDFF(pModelInfo);
+            RestoreDFF(pModelInfo);
+            return;
         case eClientModelType::TXD:
-            return DeallocateTXD(pModelInfo);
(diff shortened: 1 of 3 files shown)


Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -78,30 +78,49 @@ bool CClientModel::Allocate(ushort usParentID)
     return false;
 }
 
+// You can call it only in destructor for DFF.
```
botder, line 81: Shouldn't you mark this method private in such case?

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -78,30 +78,49 @@ bool CClientModel::Allocate(ushort usParentID)
...
 bool CClientModel::Deallocate()
 {
     if (!m_bAllocatedByUs)
```
botder, line 84: Looking over the class...wouldn't it make more sense if `m_bAllocatedByUs = true;` (line 28) was done only in the case where we actually allocated a new model? You won't need that if-block around `pModelInfo->DeallocateModel();`.

Client/mods/deathmatch/logic/CClientModel.cpp
```diff
@@ -78,30 +78,49 @@ bool CClientModel::Allocate(ushort usParentID)
...
+        // Remove model info
+        pModelInfo->DeallocateModel();
+    }
+
+    return true;
+}
+
+void CClientModel::RestoreEntitiesUsingThisModel()
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(m_iModelID, true);
```
botder, line 104: Don't you also need an `if (!m_bAllocatedByUs)` here?

TheNormalnij: ~~I think we can remove m_bAllocatedByUs from this class completely~~

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: It is still possible to use it with TXD

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: Some classes CClientPed/CClientVehicle/CClientObject have field `std::shared_ptr<CClientModel> m_clientModel`. I think it is used to make deallocation after unloading a model from stream.

Client/mods/deathmatch/logic/CClientModel.cpp
TheNormalnij: Here can be check. But this is unnecessary for now. This code used only for deallocation.

botder (approved):

PR #3205, opened 2023-10-12
Discord RPC tweaks (from nightly feedback)

Lpsd:
- Added server name to the default presence details
- Fixed a couple of cases where the presence wouldn't display properly (or at all)
    - When disconnecting from a server
    - When enabling rich presence in MTA settings
- Added the following functions:
    - `bool setDiscordRichPresenceStartTime(int seconds)`
    - `bool setDiscordRichPresenceEndTime(int seconds)`
    - `bool setDiscordRichPresencePartySize(int size, int max)`
- Added some default states when roaming around San Andreas. These are updated every 15 seconds:
    - On foot: `Walking around <zone name>`
    - In vehicle ([prefixes](https://github.com/multitheftauto/mtasa-blue/blob/2c7932398ad561d7af8b7bf5b9a6c5dc0e52a58f/Client/mods/deathmatch/logic/CClientVehicle.h#L150)): `<prefix> <zone name>` - e.g: `Flying around Los Santos`
    - Dead: `In a ditch`, `En-route to hospital`, `Meeting their maker`, `Regretting their decisions`, `Wasted`
    - Many more states based on player task, see https://github.com/multitheftauto/mtasa-blue/blob/a5e70f55cd73dc83ac05c8b23fd44a7772e63404/Client/mods/deathmatch/logic/CClientTask.h#L36-L54
- Allow state, details, assets and button functions to be passed an empty string in order to reset to defaults.

Easter egg: when zone name is "Unknown" (out of normal SA bounds) it will show as Area 51 👽 

image

### DIFF
--- a/Client/core/CCore.h
+++ b/Client/core/CCore.h
@@ -286,7 +286,7 @@ public:
     bool   IsUsingCustomStreamingMemorySize();
     size_t GetStreamingMemory();
 
-    SString GetLastConnectedServerName() const;
+    SString GetLastConnectedServerName() const { return m_strLastConnectedServerName; }
     void    SetLastConnectedServerName(SString strServerName) { m_strLastConnectedServerName = strServerName; }
 
 private:
--- a/Client/core/CCore.h
+++ b/Client/core/CCore.h
@@ -285,6 +285,10 @@ public:
     void   SetCustomStreamingMemory(size_t szMB);
     bool   IsUsingCustomStreamingMemorySize();
     size_t GetStreamingMemory();
+
+    SString GetLastConnectedServerName() const;
+    void    SetLastConnectedServerName(SString strServerName) { m_strLastConnectedServerName = strServerName; }
+
 private:
     void ApplyCoreInitSettings();
 
@@ -385,6 +389,8 @@ private:
     bool                 m_bDummyProgressUpdateAlways;
     bool                 m_bIsRenderingGrass;
     bool                 m_bFakeLagCommandEnabled;
+
+    SString m_strLastConnectedServerName{};
 
     // Command line
     static void                        ParseCommandLine(std::map<std::string, std::string>& options, const char*& szArgs, const char** pszNoValOptions = NULL);
--- a/Client/core/CCore.h
+++ b/Client/core/CCore.h
@@ -45,30 +45,30 @@ class CDiscordInterface;
 #define DIRECTINPUT_VERSION 0x0800
 #include <dinput.h>
 
-#define BLUE_VERSION_STRING     "Multi Theft Auto v" MTA_DM_BUILDTAG_LONG
-#define BLUE_COPYRIGHT_STRING   "Copyright (C) 2003 - %BUILD_YEAR% Multi Theft Auto"
+#define BLUE_VERSION_STRING   "Multi Theft Auto v" MTA_DM_BUILDTAG_LONG
+#define BLUE_COPYRIGHT_STRING "Copyright (C) 2003 - %BUILD_YEAR% Multi Theft Auto"
 
 // Configuration file path (relative to MTA install directory)
-#define MTA_CONFIG_PATH             "mta/config/coreconfig.xml"
-#define MTA_SERVER_CACHE_PATH       "mta/config/servercache.xml"
-#define MTA_CONSOLE_LOG_PATH        "mta/logs/console.log"
-#define MTA_CONSOLE_INPUT_LOG_PATH  "mta/logs/console-input.log"
-#define CONFIG_ROOT                 "mainconfig"
-#define CONFIG_NODE_CVARS           "settings"                  // cvars node
-#define CONFIG_NODE_KEYBINDS        "binds"                     // keybinds node
-#define CONFIG_NODE_JOYPAD          "joypad"
-#define CONFIG_NODE_UPDATER         "updater"
-#define CONFIG_NODE_SERVER_INT      "internet_servers"          // backup of last successful master server list query
-#define CONFIG_NODE_SERVER_FAV      "favourite_servers"         // favourite servers list node
-#define CONFIG_NODE_SERVER_REC      "recently_played_servers"   // recently played servers list node
-#define CONFIG_NODE_SERVER_OPTIONS  "serverbrowser_options"     // saved options for the server browser
-#define CONFIG_NODE_SERVER_SAVED    "server_passwords"    // This contains saved passwords (as appose to save_server_passwords which is a setting)
-#define CONFIG_NODE_SERVER_HISTORY  "connect_history"
-#define CONFIG_INTERNET_LIST_TAG    "internet_server"
-#define CONFIG_FAVOURITE_LIST_TAG   "favourite_server"
-#define CONFIG_RECENT_LIST_TAG      "recently_played_server"
-#define CONFIG_HISTORY_LIST_TAG     "connected_server"
-#define IDT_TIMER1 1234
+#define MTA_CONFIG_PATH            "mta/config/coreconfig.xml"
+#define MTA_SERVER_CACHE_PATH      "mta/config/servercache.xml"
+#define MTA_CONSOLE_LOG_PATH       "mta/logs/console.log"
+#define MTA_CONSOLE_INPUT_LOG_PATH "mta/logs/console-input.log"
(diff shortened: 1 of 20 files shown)


Fernando-A-Rocha:
Very creative with the unique default states! Good job

Client/core/CCore.h
```diff
@@ -285,6 +285,10 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
     void   SetCustomStreamingMemory(size_t szMB);
     bool   IsUsingCustomStreamingMemorySize();
     size_t GetStreamingMemory();
+
+    SString GetLastConnectedServerName() const { return m_strLastConnectedServerName; }
```
tederis: Must return `const SString&`

Client/core/CCore.h
```diff
@@ -285,6 +285,10 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
...
+    void    SetLastConnectedServerName(SString strServerName) { m_strLastConnectedServerName = strServerName; }
```
tederis: Unnecessary string copy when passing it as an argument. Can be eliminated by using a const reference or move semantics.

Client/core/CDiscordRichPresence.cpp
```diff
@@ -125,13 +140,13 @@ void CDiscordRichPresence::SetAsset(const char* szAsset, const char* szAssetText
 {
     if (isLarge)
     {
-        m_strDiscordAppAsset = (szAsset && *szAsset) ? szAsset : DEFAULT_APP_ASSET;
-        m_strDiscordAppAssetText = (szAssetText && *szAssetText) ? szAssetText : DEFAULT_APP_ASSET_TEXT;
+        m_strDiscordAppAsset = (std::string(szAsset).length() > 0 && szAsset && *szAsset) ? szAsset : DEFAULT_APP_ASSET;
```
tederis: It is better to use std::strlen. There you have a string copy without a sufficient reason.

Client/core/CDiscordRichPresence.cpp
```diff
@@ -125,13 +140,13 @@ void CDiscordRichPresence::SetAsset(const char* szAsset, const char* szAssetText
...
+        m_strDiscordAppAssetText = (std::string(szAssetText).length() > 0 && szAssetText && *szAssetText) ? szAssetText : DEFAULT_APP_ASSET_TEXT;
```
tederis: The same

Client/core/CDiscordRichPresence.cpp
```diff
@@ -125,13 +140,13 @@ void CDiscordRichPresence::SetAsset(const char* szAsset, const char* szAssetText
...
     }
     else
     {
-        m_strDiscordAppAssetSmall = (szAsset && *szAsset) ? szAsset : DEFAULT_APP_ASSET_SMALL;
-        m_strDiscordAppAssetSmallText = (szAssetText && *szAssetText) ? szAssetText : DEFAULT_APP_ASSET_SMALL_TEXT;
+        m_strDiscordAppAssetSmall = (std::string(szAsset).length() > 0 && szAsset && *szAsset) ? szAsset : DEFAULT_APP_ASSET_SMALL;
```
tederis: The same

Client/core/CDiscordRichPresence.cpp
```diff
@@ -125,13 +140,13 @@ void CDiscordRichPresence::SetAsset(const char* szAsset, const char* szAssetText
...
+        m_strDiscordAppAssetSmallText = (std::string(szAssetText).length() > 0 && szAssetText && *szAssetText) ? szAssetText : DEFAULT_APP_ASSET_SMALL_TEXT;
```
tederis: The same

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -958,6 +962,96 @@ void CClientGame::DoPulsePostFrame()
...
+
+                    if (task && task->IsValid())
+                    {
+                        auto taskType = task->GetTaskType();
+
+                        // Check for states which match our primary task
+                        std::vector<STaskState> taskStates{};
+                        for (auto it = g_playerTaskStates.begin(); it != g_playerTaskStates.end(); it++)
+                        {
+                            if (it->first == taskType)
```
tederis: Isn't it better to use a structured binding here?

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -958,6 +962,96 @@ void CClientGame::DoPulsePostFrame()
...
+
+                        discord->SetPresenceState(discordState.c_str(), false);
+                    }
+                }
+                else
+                {
+                    discord->SetPresenceState("In-game", false);
+                }
+
+                m_timeLastDiscordStateUpdate = time(nullptr);
```
tederis: Why not use GetTickCount64_ instead of a calendar time here?

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -5553,6 +5649,14 @@ void CClientGame::DoWastedCheck(ElementID damagerID, unsigned char ucWeapon, uns
             // Send the packet
             g_pNet->SendPacket(PACKET_ID_PLAYER_WASTED, pBitStream, PACKET_PRIORITY_HIGH, PACKET_RELIABILITY_RELIABLE_ORDERED);
             g_pNet->DeallocateNetBitStream(pBitStream);
+
+            auto discord = g_pCore->GetDiscord();
+            if (discord->IsDiscordRPCEnabled())
+            {
+                std::vector<std::string> states{"In a ditch", "En-route to hospital", "Meeting their maker", "Regretting their decisions", "Wasted"};
```
tederis: I think it's better to make the vector static.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -5553,6 +5649,14 @@ void CClientGame::DoWastedCheck(ElementID damagerID, unsigned char ucWeapon, uns
...
+                std::string              state = states[rand() % states.size()];
```
tederis: It should be `const std::string& state = `.

Client/mods/deathmatch/logic/CClientVehicle.h
```diff
@@ -145,6 +145,12 @@ struct SVehicleComponentData
     bool    m_bVisible;
 };
 
+static const std::map<eClientVehicleType, std::string> g_vehicleTypePrefixes{
```
tederis: std::map is excessive here. There are 11 enum entries that can be fully covered by **std::array<std::string, 11>**.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -958,6 +962,96 @@ void CClientGame::DoPulsePostFrame()
...
+                            std::srand(GetTickCount64_());
+                            int  index = (std::rand() % stateCount);
+                            auto taskState = taskStates[index];
+
+                            discordState = taskState.strState;
+                            useZoneName = taskState.bUseZone;
+                        }
+
+                        zoneName = (zoneName == "Unknown") ? "Area 51" : zoneName;
+                        discordState = useZoneName ? SString(std::string(discordState + " %s").c_str(), zoneName.c_str()) : discordState;
```
tederis, line 1041: I would suggest to append a zone name
```
if (useZoneName)
    discordState.append(" " + zoneName);
```

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -958,6 +962,96 @@ void CClientGame::DoPulsePostFrame()
...
+                                ++it;
+                        }
+
+                        // Choose a random task state (if we have any)
+                        int stateCount = taskStates.size();
+                        if (stateCount > 0)
+                        {
+                            std::srand(GetTickCount64_());
+                            int  index = (std::rand() % stateCount);
+                            auto taskState = taskStates[index];
```
tederis, line 1034: I'd use a reference here to avoid unnecessary STaskState struct copy.

tederis:
This PR is no more relevant due to #3216.

Dutchman101:
Closed by ef26810

PR #3209, opened 2023-10-14
Use MTA temp path for client tests

Lpsd:
Resolves an issue where the temp tests file fails to create on the client (debug build), when launching from mtasa:// protocol

```
Assertion failed!

Program: mtasa-blue\Bin\mta\loader_d.dll
File: mtasa-blue\Shared\sdk\SharedUtil.Tests.hpp
Line: 977

Expression: strResult == result
```


Lpsd:
Merging fix into #3211 instead

PR #3210, opened 2023-10-14
Cleanup comments from #3172

Lpsd:


tederis:
I think that the PR is no longer relevant.

PR #3211, opened 2023-10-15
Support mtasa:// protocol connect args

Lpsd:
### Changes
- Allows arguments to be supplied to the mtasa:// protocol and received by the server in Lua via `onPlayerConnect` event parameters. 
- Adds an extra argument to `redirectPlayer` which allows you to pass a table to achieve the same result (all keys, values are automatically converted to strings, in order to form the 'argument path').
- A fix for some of the SharedUtil tests (#3209) has been modified and merged into this PR.

---

### Additional Info 
Total length of arguments is limited to 1024 including `/` delimiter per key/value pair (arguments are treated as path string for 99% of lifetime)

_Subject to change for typical url parameters e.g `?a=1&b=2`_

This feature is useful for the Discord RPC functionality, servers can create a party system using this or give rewards for joining via Discord (of course this can extend to other platforms).

---

### Example
Have serverside script running:
```lua
SERVER_IP, SERVER_PORT, SERVER_PASS = "localhost", 22003, ""
DEFAULT_ARGS = {["foo"] = "abc", ["bar"] = "def"}

local storedArgs

function connect(nick, ip, username, serial, version, versionString, args)
    storedArgs = args
    iprint(args)
end
addEventHandler("onPlayerConnect", root, connect)

function redirect(player)
    redirectPlayer(player, SERVER_IP, SERVER_PORT, SERVER_PASS, (type(storedArgs) == "table") and storedArgs or DEFAULT_ARGS)
end
addCommandHandler("redirect", redirect)
```

| Action        | Result |
| ------------- |-----|
| Player connects with `mtasa://123.456.78.9:22003/foo/123/bar/456/baz/789`      | image |
| Player uses `redirect` command     | Same as above     | 

---

## Notes
- To test mtasa:// protocol with debug build, edit registry `Computer\HKEY_CLASSES_ROOT\mtasa\shell\open\command` and set to your mtasa build path e.g `"C:\mtasa-blue\Bin\Multi Theft Auto_d.exe"%1`

## To-do

- [x] Handle `/` in user provided keys/values, either remove character or URL encode? (`/` is our delimiter so would affect converting to pairs)
    - `/` will be removed from user provided args
- [x] Fix argument parser issue when using `std::unordered_map<type, type>` for Lua args
    - See #3246
- [x] Determine whether keys/values (protocol parameters) should be changed to `?a=1&b=2` format

## The method used for URL parameters (key/value pairs delimited by slash) should be changed to use the standard URI parameter format `?foo=bar&baz=boz`

This is going to require a lot of time to investigate how this is currently being used by various services, both public and private (internal). It will affect all areas, to do with downloads, updates, server connections, etc. This is the reason why I initially opted to use the key/value pairs delimited by slash, as it causes the least disruption to the client. This will be a major change, that fittingly should go into a major release.


Inder00:
marge :shipit:

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -335,6 +335,9 @@ class CPlayer final : public CPed, public CClient
     const SString& GetQuitReasonForLog() { return m_strQuitReasonForLog; }
     void           SetQuitReasonForLog(const SString& strReason) { m_strQuitReasonForLog = strReason; }
 
+    void SetProtocolConnectArgs(SString args)  { m_strProtocolConnectArgs = args; }
+    SString GetProtocolConnectArgs() const { return m_strProtocolConnectArgs; }
```
tederis: const SString&

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -335,6 +335,9 @@ class CPlayer final : public CPed, public CClient
     const SString& GetQuitReasonForLog() { return m_strQuitReasonForLog; }
     void           SetQuitReasonForLog(const SString& strReason) { m_strQuitReasonForLog = strReason; }
 
+    void SetProtocolConnectArgs(SString args)  { m_strProtocolConnectArgs = args; }
```
tederis: const SString& args

tederis: Or even SString&& args to leverage the move sematics.

Client/core/CCommandFuncs.cpp
```diff
@@ -217,7 +217,10 @@ void CCommandFuncs::Connect(const char* szParameters)
     if (!strncmp(szBuffer, "mtasa://", 8))
     {
         // Using a mtasa:// URI to connect
-        SString strArguments = g_pCore->GetConnectCommandFromURI(szBuffer);
+        std::string strQueryParams;
+
+        SString strArguments = g_pCore->GetConnectCommandFromURI(szBuffer, strQueryParams);
+        g_pCore->SetProtocolConnectArgs(strQueryParams);
```
tederis: I think we can use std::move(strQueryParams) here.

Client/core/CCore.cpp
```diff
@@ -1266,7 +1267,11 @@ void CCore::DoPostFramePulse()
                 // Does it begin with mtasa://?
                 if (m_szCommandLineArgs && strnicmp(m_szCommandLineArgs, "mtasa://", 8) == 0)
                 {
-                    SString strArguments = GetConnectCommandFromURI(m_szCommandLineArgs);
+                    std::string strQueryParams;
+
+                    SString strArguments = GetConnectCommandFromURI(m_szCommandLineArgs, strQueryParams);
+                    g_pCore->SetProtocolConnectArgs(strQueryParams);
```
tederis: The same. I would be better to use a move sematics.

Client/core/CCore.h
```diff
@@ -285,6 +287,10 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
     void   SetCustomStreamingMemory(size_t szMB);
     bool   IsUsingCustomStreamingMemorySize();
     size_t GetStreamingMemory();
+
+    void        SetProtocolConnectArgs(std::string args) { m_strProtocolConnectArgs = args; }
```
tederis: Move sematics or const reference.

Client/core/CCore.h
```diff
@@ -285,6 +287,10 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
...
+    std::string GetProtocolConnectArgs() const { return m_strProtocolConnectArgs; }
```
tederis: const std::string&

Client/core/CCore.cpp
tederis: The same below

fresholia:
Also it should work with `connect` console command, as I see i can't connect the game with connect command.

Fernando-A-Rocha:
What about also changing [redirectPlayer](https://wiki.multitheftauto.com/wiki/RedirectPlayer)?

Fernando-A-Rocha:
well done, lgtm

Lpsd:
the redirectPlayer functionality has uncovered an issue with the argument parser which needs resolving before this can be merged. See [discussion on dev discord](https://discord.com/channels/801330706252038164/801411628600000522/1174814630992945202)

Lpsd:
> Also it should work with `connect` console command, as I see i can't connect the game with connect command.

Works fine for me, can you confirm how you're using the command and what errors you're seeing? @fresholia

Lpsd:
We can't use `std::unordered_map<std::string, std::string>` to retrieve table arguments in a Lua definition (because a script providing the wrong type ie `{1, 2, 3}` will cause a crash, and I couldn't find a solution for this).

For this reason I've changed to use `CLuaArgument` in the unordered_map, as it can handle any value from the Lua side, then [convert this to a string](https://github.com/multitheftauto/mtasa-blue/blob/ea878efd205fef0ab325231e8c4a9ea64b7ee835/Server/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp#L817-L851) using the already available `CLuaArgument::GetAsString` method

However, I did have to implement a hash function for `CLuaArgument` in order to use it with `std::unordered_map` - could do with some opinions on this; not sure if it's correct (approach and implementation-wise). This is an issue we inevitably need to tackle at some point anyway (accepting tables of any value using new arg parser)...

https://github.com/multitheftauto/mtasa-blue/blob/c96a8fffa063577b1273f033ee8728d7a71c0b9f/Client/mods/deathmatch/logic/lua/CLuaArgument.h#L68-L78

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -2063,6 +2065,18 @@ namespace SharedUtil
         return false;
     }
 
+    std::vector<std::string> Explode(std::string const& s, char delim)
```
TheNormalnij: Is "Split" common name for this operation?

Server/mods/deathmatch/logic/packets/CPlayerJoinDataPacket.cpp
```diff
@@ -26,13 +26,16 @@ bool CPlayerJoinDataPacket::Read(NetBitStreamInterface& BitStream)
     m_bOptionalUpdateInfoRequired = BitStream.ReadBit();
 
     if (BitStream.Read(m_ucGameVersion) && BitStream.ReadStringCharacters(m_strNick, MAX_PLAYER_NICK_LENGTH) &&
-        BitStream.Read(reinterpret_cast<char*>(&m_Password), 16) && BitStream.ReadStringCharacters(m_strSerialUser, MAX_SERIAL_LENGTH))
+        BitStream.Read(reinterpret_cast<char*>(&m_Password), 16) && BitStream.ReadStringCharacters(m_strSerialUser, MAX_SERIAL_LENGTH) &&
+        BitStream.ReadStringCharacters(m_strConnectArgs, MAX_CONNECT_ARGS_LENGTH))
     {
```
TheNormalnij: Does this change breaks old clients?

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
```diff
@@ -371,6 +371,42 @@ CClientEntity* CLuaArgument::GetElement() const
...
+        case LUA_TNIL:
+            strBuffer = "nil";
+            break;
+        case LUA_TBOOLEAN:
+            if (m_bBoolean)
+                strBuffer = "true";
+            else
+                strBuffer = "false";
+            break;
+        case LUA_TTABLE:
```
TheNormalnij: Shouldn't we allow only string and number arguments?

Client/core/CCore.cpp
```diff
@@ -26,6 +26,7 @@
 #include <SharedUtil.Detours.h>
 #include <ServerBrowser/CServerCache.h>
 #include "CDiscordRichPresence.h"
+#include <curl/include/curl/curl.h>
```
TheNormalnij: Add "../../vendor/curl/include" to "includedirs" in  premake5.lua for Client Core.  This allows you to use "#include <curl/curl.h>" here.

Shared/sdk/SharedUtil.Misc.hpp
Lpsd: "Explode" would be the correct equivalent in PHP (sorry, I use PHP daily). It's called split in JavaScript, so either is fine imo (but explode sounds cooler 😎 )

Server/mods/deathmatch/logic/packets/CPlayerJoinDataPacket.cpp
Lpsd: No, because we are checking the bitstream versions on both sides, if the bitstream is not updated, the packet remains the same.

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
Lpsd: Because of the issue raised in Discord with arg parser, we cannot specify exact types to `std::unordered_map` when receiving the Lua arguments, we have to use `CLuaArgument` which accepts any value. So I think this is fine, if they supply any bad values (such as function, or userdata) then it's simply not included.

Lpsd: Also this function already exists, it's just copied from Server Deathmatch project (we don't have shared SDK for arg parser unfortunately)

Server/mods/deathmatch/logic/packets/CPlayerJoinDataPacket.cpp
Lpsd: Sorry, I thought this was some other code 😅 You're right, I will update this to check player bitstream version

tederis:
> We can't use `std::unordered_map<std::string, std::string>` to retrieve table arguments in a Lua definition (because a script providing the wrong type ie `{1, 2, 3}` will cause a crash, and I couldn't find a solution for this).
> 
> For this reason I've changed to use `CLuaArgument` in the unordered_map, as it can handle any value from the Lua side, then [convert this to a string](https://github.com/multitheftauto/mtasa-blue/blob/ea878efd205fef0ab325231e8c4a9ea64b7ee835/Server/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp#L817-L851) using the already available `CLuaArgument::GetAsString` method
> 
> However, I did have to implement a hash function for `CLuaArgument` in order to use it with `std::unordered_map` - could do with some opinions on this; not sure if it's correct (approach and implementation-wise). This is an issue we inevitably need to tackle at some point anyway (accepting tables of any value using new arg parser)...
> 
> https://github.com/multitheftauto/mtasa-blue/blob/c96a8fffa063577b1273f033ee8728d7a71c0b9f/Client/mods/deathmatch/logic/lua/CLuaArgument.h#L68-L78

I think it would be better to fix a crash instead of constructing hash related complicated things. I hope my PR #3246 will help.

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
Lpsd: No longer relevant due to #3246

Client/core/CCore.h
```diff
@@ -286,6 +288,9 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
     bool   IsUsingCustomStreamingMemorySize();
     size_t GetStreamingMemory();
 
+    void               SetProtocolConnectArgs(const std::string&& args) { m_strProtocolConnectArgs = args; }
```
tederis, line 291: Unfortunately, const rvalue reference doesn't make sense. Even without a const qualifier it wont work as expected because in the function's body `args` is actually lvalue reference. So you should remove `const` and cast `args` to the rvalue explicitly: `SetProtocolConnectArgs(std::string&& args) { m_strProtocolConnectArgs = std::move(args); }`

Client/sdk/core/CCoreInterface.h
```diff
@@ -190,6 +190,9 @@ class CCoreInterface
     virtual bool   IsUsingCustomStreamingMemorySize() = 0;
     virtual size_t GetStreamingMemory() = 0;
 
+    virtual void        SetProtocolConnectArgs(const std::string&& args) = 0;
```
tederis, line 193: The same: const rvalue has no sense

Server/mods/deathmatch/logic/packets/CPlayerJoinDataPacket.h
```diff
@@ -44,6 +44,9 @@ class CPlayerJoinDataPacket final : public CPacket
 
     bool IsOptionalUpdateInfoRequired() { return m_bOptionalUpdateInfoRequired; }
 
+    const char* GetConnectArgs() { return m_strConnectArgs; };
```
tederis, line 47: Why don't you return a string object reference?

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1735,6 +1735,11 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
         pPlayer->SetBitStreamVersion(Packet.GetBitStreamVersion());
         g_pNetServer->SetClientBitStreamVersion(Packet.GetSourceSocket(), Packet.GetBitStreamVersion());
 
+        const char* connectArgs = Packet.GetConnectArgs();
```
tederis, line 1742: This related to a CPlayerJoinDataPacket class problem: it's probably better to use a string.

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -335,6 +335,9 @@ class CPlayer final : public CPed, public CClient
     const SString& GetQuitReasonForLog() { return m_strQuitReasonForLog; }
     void           SetQuitReasonForLog(const SString& strReason) { m_strQuitReasonForLog = strReason; }
 
+    void           SetConnectArgs(const SString&& args) { m_strConnectArgs = args; }
```
tederis, line 338: The same problem with const rvalue ref

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1735,6 +1735,11 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
...
+
+        if (std::strlen(connectArgs) > 0)
+            pPlayer->SetConnectArgs(std::move(Packet.GetConnectArgs()));
```
tederis, line 1745: I would recommend to remove `std::move` here because Packet_PlayerJoinData(and even CGame) method doesn't own this string.

Client/mods/deathmatch/logic/rpc/CInputRPCs.cpp
```diff
@@ -240,6 +240,20 @@ void CInputRPCs::ForceReconnect(NetBitStreamInterface& bitStream)
 
         if (bitStream.Read(usPort))
         {
+            char* szArgs = new char[MAX_PROTOCOL_CONNECT_ARGS_LENGTH + 1];
```
tederis, line 243: It seems that we have a memory leak here. Because `szArgs` isn't deleted below.
Maybe a preallocated `std::string` is the solution? It's null terminated and guaranteed to contiguous since C++11 and we have non-const `::data` member since C++17(please, use it).

tederis: But it's even better to use `bitStream.ReadString()` (with the modification of a packet itself).

Client/core/CCore.h
Pirulax: I agree with this.
Also, that custom hash function for `CLuaArgument` isn't good at all, and should be avoided at all costs.
Actually, and kind of comparasion/hashing of it should be avoided.

Dutchman101:
Here's the last status update on this PR:

**Dutchman101 — 01/07/2024 1:42 AM**
If you address the final Dec 1 code reviews on https://github.com/multitheftauto/mtasa-blue/pull/3211, it can be merged 

**lopsi — 01/07/2024 3:04 PM**
there is more to do than that
I would like to use other arg format, ?foo=bar&baz=boz
I know it's possible, I already started working on it, but ran into few issues
MTA updater also uses this format in some weird way, I would like to change it too

ghost:
Any updates on this? Would be a really nice feature.

Inder00:
Bump

ghost:
Bump?

arisudesu:
Why would you use this weird syntax with slashes? Aren't usual ampersands enough, as seen in URLs?

What if an argument is specified twice, what value will be in the table - first or last? What if it is specified thrice? E.g. a `/arg/1/arg/2/arg/3` or a `/arg[]/1/arg[]/2/arg[]/3`?

Are all the values of `arg[]` possible to be extracted? Real-world examples which I know are:
 * PHP where repeated keys suffixed with `[]` in the query string form an array;
 * Golang, we have `net/url.Values` which is a map of keys to subarrays, that is similar to PHP but doesn't require `[]`;
 * JavaScript we have [URLSearchParams.getAll](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/getAll)

and so on.

What if an argument does not have value, then how will it be combined with the arguments following it - like this: `/novaluearg//valuearg/1`?

Why not pass your "query" string as is and let the server/client script decide how to parse it? And to implement general querystring-like parsing, just add a new `fromURLQuery` [utility function](https://wiki.multitheftauto.com/wiki/Server_Scripting_Functions#Utility_functions).

Don't reinvent the wheel please, we all (mostly) are used to web URLs and they do their job well, use them.

arisudesu:
> `/` will be removed from user provided args

What is syntax if one wants to encode slash to an argument key or a value? Is it possible at all, and if it is, how? A usual percent-encoding, as in web URLs?

Lpsd:
> Why would you use this weird syntax with slashes? Aren't usual ampersands enough, as seen in URLs?

Requires much more thought and effort to ensure compatibility with existing services which utilize the standard protocol/parameters, such as updates. We'll have to change code surrounding that which hasn't changed in a while.

Also please refer to existing comments to save time https://github.com/multitheftauto/mtasa-blue/pull/3211#issuecomment-1913764845

I haven't had much interest in this lately so, if you feel motivated, feel free to pick this up!

Lpsd:
All of your other questions can be answered by taking a glance at `CLuaPlayerDefs::ArgsToString` and testing using what's provided. Not that it is relevant, as the intention was already to use standard URI parameters instead.

Lpsd:
I've converted to draft and updated the description so there's no confusion on the state of this PR.

Fernando-A-Rocha:
Would be nice to revive this PR

Issue #3236, opened 2023-11-10
Discord RPC gets reset when ANY resource stops


VulpyWags:
### Describe the bug

Due to changes in https://github.com/multitheftauto/mtasa-blue/pull/3224 when any resource stops the default Discord RPC starts being used and any custom implementation is gone.

Based on the PR this looks like intended behaviour but I don't think this is expected behaviour for server scripters. 

Can this be removed or improved upon?

### Steps to reproduce

1. Start your Discord RPC resource implementation
2. Stop any resource
3. Your Discord RPC is gone

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

matheustkn:
Do you have any code examples? It seems to be a problem with how it is being used.

EDIT: The problem really is happening.

VulpyWags:
No it isn't. Due to this commit: https://github.com/multitheftauto/mtasa-blue/pull/3224/commits/b51f43d93d124c54eda83ee6ad4c3fc2ce46385e when ANY resource stops the custom Discord RPC implementation is gone

This is my current implementation, it worked pefectly before mentioned PR: https://github.com/MrGreenGaming/MTA-Resources/tree/master/resources/%5Bmisc%5D/%5Bdiscord%5D/discord_rich_presence

znvjder:
I must have missed that, my bad - fix is already in the PR. Thank you for reporting this!

VulpyWags:
Can confirm that the bug has been solved with the latest nightly build.

PR #3237, opened 2023-11-11
Fix Discord resetting whenever any resource stops

znvjder:
This PR introduces a fix for issue #3236


VulpyWags:
Much appreciated for the quick fix!

Dutchman101 (approved):
This is important, builds with the bug are circulating; if anyone has code feedback it can be left afterwards (or add any tweaks through a new PR)

PR #3238, opened 2023-11-11
Trigger onClientPlayerWeaponSwitch for Remote Players

znvjder:
This PR is a duplicate of #2181 and deals with issue #1141.

For those unfamiliar - it allows triggering event **onClientPlayerWeaponSwitch** for players. However, currently, it's only works for the local player.
It also resolves an issue that was impeding code merging. This specifically involves sending the event to all players, which could impact performance with network traffic on high players counts. Now, it's refined to send event solely to nearby (streamed) players.


Lpsd (approved):
Looks good to me, assuming `CPlayerManager::Broadcast` handles the bitstream version checks (_looks_ like it does).

Dutchman101:
Reverted in e242f22

Even if this works while testing in a custom build, the game is broken in nightly r22348; after connecting to a server, all world objects/terrain will be invisible and certain resources functionality won't work (like you can't open admin panel. I didn't debug it further but it may indicate client/server calls are broken). PR must have gotten incorrect bitstream code..

@znvjder

znvjder:
Okay. I will take a closer look at it. Thanks for report.

PR #3241, opened 2023-11-13
Add server side projectile events

Zangomangu:
Adds two events

`onPlayerProjectileCreation ( int weaponType, float posX, float posY, float posZ [, float force, element target, float rotX, float rotY, float rotZ, float velX, float velY, float velZ ] )`
**source**: player who created the projectile.
This event is triggered when player fires a projectile (e.g. a grenade, satchel, rocket)
or when a projectile is created through scripting function [createProjectile](https://wiki.multitheftauto.com/wiki/CreateProjectile).

If cancelled the projectile won't appear for other players on the server.

`onPlayerDetonateSatchels ( )`
**source**: player who wants to detonate his own satchels.
This event is triggered when a player attempts to detonate satchels with a detonator, or through scripting function [detonateSatchels](https://wiki.multitheftauto.com/wiki/DetonateSatchels).

If cancelled the satchels won't detonate - neither on client or remote players

Note to reviewer: The packet change only renames a previously unused variable and removes 10+ yr old bitstream checks.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1583,6 +1583,8 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerNetworkStatus", "type, ticks", NULL, false);
     m_Events.AddEvent("onPlayerScreenShot", "resource, status, file_data, timestamp, tag", NULL, false);
     m_Events.AddEvent("onPlayerResourceStart", "resource", NULL, false);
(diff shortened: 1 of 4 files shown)


tederis:
Looks good to me

ghost:
I'd suggest event name change, from onPlayerProjectileCreation to onProjectileCreation, to keep it consistent with https://wiki.multitheftauto.com/wiki/OnClientProjectileCreation and other events (e.g onClientExplosion & onExplosion)

PR #3244, opened 2023-11-18
Discord-RPC: IO Thread Disablement and minor security fixes

znvjder:
This pull request introduces several fixes and security enhancements. 

The main goal of this pull request is to disable the IO thread and manually update Rich Presence. Will improve interoperability and compatibility with the MTA client - hopefully reducing the number of asserts triggered, for example, by the load on the iostream thread by MTA (resource fetching, etc.).

### DIFF
--- a/Client/core/CDiscordRichPresence.h
+++ b/Client/core/CDiscordRichPresence.h
@@ -69,7 +69,7 @@ private:
 
     std::optional<std::tuple<std::pair<std::string, std::string>, std::pair<std::string, std::string>>> m_aButtons;
 
-    std::mutex    m_mutexThread;
+    std::mutex m_threadSafetyMutex;
 
     unsigned long m_uiDiscordAppStart;
     unsigned long m_uiDiscordAppEnd;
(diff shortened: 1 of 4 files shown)


Lpsd (approved):

Client/core/CDiscordRichPresence.h
```diff
@@ -69,6 +69,8 @@ class CDiscordRichPresence : public CDiscordInterface
 
     std::optional<std::tuple<std::pair<std::string, std::string>, std::pair<std::string, std::string>>> m_aButtons;
 
+    std::mutex m_threadSafetyMutex;
```
tederis, line 72: It doesn't seem like this mutex is really needed. Is this a memory region that can be accessed from two threads at the same time?

znvjder: By adding a mutex, I wanted to further secure the initialization and updates of Rich Presence. This way, it becomes possible to effectively control access to Discord-RPC, preventing simultaneous modifications by different threads.

PR #3246, opened 2023-11-19
Fix std::unordered_map<std::string, std::string> parsing

tederis:
This PR is inspired by the problem appeared in #3211. It fixes a crash when `std::unordered_map<std::string, std::string>` is used as an argument for the new Lua parser.


Lpsd (approved):
lgtm

PR #3251, opened 2023-11-26
Add ability to limit client triggered events (via triggerServerEvent)

Lpsd:
Currently, one of the only ways to limit the amount of events being sent from the client to server (via triggerServerEvent) is to track the calls manually for every event handler you add (in Lua) which results in a valid `client` variable being set, upon the handler function being executed.

That's fine for registered events (although a bit of a pain to add for every single event). However, unregistered events cannot be tracked this way. The only way to track unregistered events being called is by using `onDebugMessage` and catching a log entry such as below, which isn't ideal:

```
Client (lopsi) triggered serverside event myTestEvent, but event is not added serverside
```

This is also the same for events which do exist, but aren't remotely triggerable:

```
Client (lopsi) triggered serverside event myTestEvent, but event is not marked as remotely triggerable
```

A client spamming serverside events can cause lag, so I've added a way to deal with this in MTA itself. 

This PR adds a new serverside event, `onPlayerTriggerEventThreshold`, as well as two new mtaserver.conf settings:

- `player_triggered_event_interval`
- `max_player_triggered_events_per_interval`

These options can be set using `setServerConfigSetting` in Lua, or directly via mtaserver.conf before server startup.

Providing the player has triggered more than `max_player_triggered_events_per_interval` events per `player_triggered_event_interval` time (in milliseconds), the event will be fired, giving full control to the server how they deal with such players (log, kick, ban, etc).

This event counts ALL events triggered by the client, whether they are registered, unregistered or not remotely triggerable. Basically any usage of `triggerServerEvent` by the client will be counted.

Usage is fairly self-explanatory:

```lua
function processPlayerTriggerEventThreshold()
    kickPlayer(source, "Event spam")
end
addEventHandler("onPlayerTriggerEventThreshold", root, processPlayerTriggerEventThreshold)
```

I've also changed the way mtaserver.conf.template is loaded in the code, as adding these options took it over the 16,384 char string literal limit. That file will now be copied into the server deathmatch directory and read as XML directly, instead of `const char* szTemplateText = #include MTA_SERVER_CONF_TEMPLATE;` _shivers_

As it's currently possible for a client to abuse this event spamming, it's important that all server owners become aware of this and take necessary measures to implement (atleast) the example script above, once this PR has been merged. 

Until then, I have attached a Lua script which will prevent unregistered events from being spammed (default max 10 per 500 ms). Feel free to export the `registerEventUsage` function and track your own registered events in other resources too. 

Download the resource/script here: [events.zip](https://github.com/multitheftauto/mtasa-blue/files/13467266/events.zip)

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -4714,19 +4714,16 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
     auto iter = m_mapClientTriggeredEvents.find(pPlayer);
 
     if (iter == m_mapClientTriggeredEvents.end())
-        iter = m_mapClientTriggeredEvents.insert({pPlayer, 0}).first;
+        m_mapClientTriggeredEvents.insert({pPlayer, 0});
 
-    iter->second++;
+    m_mapClientTriggeredEvents[pPlayer]++;
 }
 
 void CGame::ProcessClientTriggeredEventSpam()
 {
-    for (const auto& pair : m_mapClientTriggeredEvents)
+    for (const auto& [player, count]: m_mapClientTriggeredEvents)
     {
-        CPlayer* player = pair.first;
-        int      count = pair.second;
-
-        if (player && count > m_iMaxClientTriggeredEventsPerInterval)
+        if (player && player->IsPlayer() && !player->IsBeingDeleted() && count > m_iMaxClientTriggeredEventsPerInterval)
             player->CallEvent("onPlayerTriggerEventThreshold", {});
     }
 
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -551,6 +551,9 @@ void CGame::DoPulse()
     PrintLogOutputFromNetModule();
     m_pScriptDebugging->UpdateLogOutput();
 
+    if (GetTickCount64_() + m_iClientTriggeredEventsIntervalMs > m_lClientTriggeredEventsLastCheck)
+        ProcessClientTriggeredEventSpam();
+
     // Unlock the critical section again
     Unlock();
 }
@@ -1585,6 +1588,7 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerResourceStart", "resource", NULL, false);
     m_Events.AddEvent("onPlayerProjectileCreation", "weaponType, posX, posY, posZ, force, target, rotX, rotY, rotZ, velX, velY, velZ", nullptr, false);
     m_Events.AddEvent("onPlayerDetonateSatchels", "", nullptr, false);
+    m_Events.AddEvent("onPlayerTriggerEventThreshold", "", nullptr, false);
 
     // Ped events
     m_Events.AddEvent("onPedVehicleEnter", "vehicle, seat, jacked", NULL, false);
@@ -2568,6 +2572,8 @@ void CGame::Packet_LuaEvent(CLuaEventPacket& Packet)
         }
         else
             m_pScriptDebugging->LogError(NULL, "Client (%s) triggered serverside event %s, but event is not added serverside", pCaller->GetNick(), szName);
+
+        RegisterClientTriggeredEventUsage(pCaller);
     }
 }
 
@@ -4698,4 +4704,32 @@ void CGame::HandleCrashDumpEncryption()
         }
     }
 #endif
+}
+
+void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
+{
+    if (!pPlayer)
+        return;
+
+    auto iter = m_mapClientTriggeredEvents.find(pPlayer);
+
+    if (iter == m_mapClientTriggeredEvents.end())
+        iter = m_mapClientTriggeredEvents.insert({pPlayer, 0}).first;
+
+    iter->second++;
+}
+
+void CGame::ProcessClientTriggeredEventSpam()
+{
+    for (const auto& pair : m_mapClientTriggeredEvents)
+    {
+        CPlayer* player = pair.first;
+        int      count = pair.second;
+
+        if (player && count > m_iMaxClientTriggeredEventsPerInterval)
+            player->CallEvent("onPlayerTriggerEventThreshold", {});
+    }
+
+    m_mapClientTriggeredEvents.clear();
+    m_lClientTriggeredEventsLastCheck = GetTickCount64_();
 }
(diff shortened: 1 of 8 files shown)


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4699,3 +4705,31 @@ void CGame::HandleCrashDumpEncryption()
...
+{
+    if (!pPlayer)
+        return;
+
+    auto iter = m_mapClientTriggeredEvents.find(pPlayer);
+
+    if (iter == m_mapClientTriggeredEvents.end())
+        iter = m_mapClientTriggeredEvents.insert({pPlayer, 0}).first;
+
+    iter->second++;
```
TheNormalnij: You don't decrease this value with time. It's works not how you describe/expecting. It cause _random_ triggering after any random spam in time.

TheNormalnij:
I didn't try my code. But i can simply implement token bucket in Lua:
```lua

local UPDATE_INTERVAL = 100
local LIMIT_PER_SECOND = 100

local REFRESH_COUNT = LIMIT_PER_SECOND / 1000 * UPDATE_INTERVAL

local limitsData = {}

local function update()
    for player, count in pairs(limitsData) do
        local newCount = count + REFRESH_COUNT
        if newCount > LIMIT_PER_SECOND then
            limitsData[player] = LIMIT_PER_SECOND
        else
            limitsData[player] = newCount
        end
    end
end
setTimer(update, UPDATE_INTERVAL, 0)

local function init()
    for _, player in pairs(getElementsByType("player")) do
        limitsData[player] = LIMIT_PER_SECOND
    end
end

init()

addEventHandler("onPlayerJoin", root, function()
    limitsData[source] = LIMIT_PER_SECOND
end)

addEventHandler("onPlayerQuit", root, function()
    limitsData[source] = nil
end)

addDebugHook("postEvent", function(sourceResource, eventName, eventSource, eventClient)
    if eventClient then
        local newCount = limitsData[eventClient] - 1
        if newCount >= 0 then
            limitsData[eventClient] = newCount
        else
            outputDebugString("Kick player during event spam. Latest event: " .. eventName, 3)
            kickPlayer(eventClient, "Event spam")
        end
    end
end)
```

Server/mods/deathmatch/logic/CGame.cpp
Lpsd: The map gets cleared in `CGame::ProcessClientTriggeredEventSpam` every interval.

There is a possibility the event will trigger if they go over the threshold before the interval is over (because they were added to the map _after_ it was last checked), but it would be getting triggered later anyway... 

Once the map is cleared that player is 'in sync' with the interval again, which should be low anyway so that doesn't _really_ matter.

Lpsd:
You still have the problem that unregistered events and non-remotely triggerable events will not be caught by your code.

Server/mods/deathmatch/logic/CGame.cpp
TheNormalnij: My bad. I see this now

tederis: I would suggest replace it with `m_mapClientTriggeredEvents[pPlayer]++;`

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4699,3 +4705,31 @@ void CGame::HandleCrashDumpEncryption()
...
+}
+
+void CGame::ProcessClientTriggeredEventSpam()
+{
+    for (const auto& pair : m_mapClientTriggeredEvents)
+    {
+        CPlayer* player = pair.first;
```
tederis: What will happen if a player leaves a server in between of RegisterClientTriggeredEventUsage and ProcessClientTriggeredEventSpam? Formally pointer will be valid(non-null) but technically it is a segfault.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4699,3 +4705,31 @@ void CGame::HandleCrashDumpEncryption()
...
+
+    if (iter == m_mapClientTriggeredEvents.end())
+        iter = m_mapClientTriggeredEvents.insert({pPlayer, 0}).first;
+
+    iter->second++;
+}
+
+void CGame::ProcessClientTriggeredEventSpam()
+{
+    for (const auto& pair : m_mapClientTriggeredEvents)
```
tederis: Additionally, server side deatchmatch project has C++17 enabled. And you can use the structured binding: `for (const auto& [player, count] : m_mapClientTriggeredEvents)`

Server/mods/deathmatch/logic/CGame.cpp
Lpsd: Referencing [CPedSync::UpdateNearPlayersList](https://github.com/multitheftauto/mtasa-blue/blob/abec07aeb90eef19ed47396baed3b190994be54a/Server/mods/deathmatch/logic/CPedSync.cpp#L310-L322), my latest changes should suffice

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4699,3 +4705,28 @@ void CGame::HandleCrashDumpEncryption()
...
+
+void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
+{
+    if (!pPlayer)
+        return;
+
+    auto iter = m_mapClientTriggeredEvents.find(pPlayer);
+
+    if (iter == m_mapClientTriggeredEvents.end())
+        m_mapClientTriggeredEvents.insert({pPlayer, 0});
```
tederis, line 4717: What is the purpose of this insert here?

tederis: See https://en.cppreference.com/w/cpp/container/map/operator_at. [] operator inserts an entry if key doesnt not already exist.

Lpsd: This has been updated in master

Server/mods/deathmatch/logic/CGame.cpp
tederis: Sorry, but I'm not convinced. The thing is **ProcessClientTriggeredEventSpam** is placed after both Lua packet processing(**g_pNetServer->DoPulse()**) and the element deletor(**m_ElementDeleter.DoDeleteAll()**). The situation: you create an entry within m_mapClientTriggeredEvents, then delete a player and only then call **ProcessClientTriggeredEventSpam** with a pointer that points to an invalid memory region.

tederis: Probably you can crash a server by kicking a player somewhere in between.

Lpsd: So, how about moving the call after line 506 `CLOCK_CALL1(m_pACLManager->DoPulse(););`

https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/logic/CGame.cpp#L506

tederis: I would suggest to move it just before **m_ElementDeleter.DoDeleteAll()** and additionally test `pPlayer->IsBeingDeleted() != true`.

Lpsd: Done (#3252). I know a couple servers who will use this so we'll find out soon enough anyway

PR #3252, opened 2023-11-27
Addendum to #3251

Lpsd:


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4708,7 +4708,7 @@ void CGame::HandleCrashDumpEncryption()
 
 void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
 {
-    if (!pPlayer)
+    if (!pPlayer && pPlayer->IsPlayer() && !pPlayer->IsBeingDeleted())
```
tederis, line 4711: It looks like it should be `if (!pPlayer || !pPlayer->IsPlayer() || pPlayer->IsBeingDeleted())`
.

tederis: And `!pPlayer->IsBeingDeleted()` check must be added inside a loop within **ProcessClientTriggeredEventSpam** method. Because we don't want to process a player when it's about to be deleted(which is highly possible since ProcessClientTriggeredEventSpam is called just before the deleter).

Lpsd: Yeah brainfart sorry. and we already have that check in ProcessClientTriggeredEventSpam

Dutchman101: 86846f1 (authored by Lpsd)

PR #3253, opened 2023-11-28
More changes for #3251

Lpsd:
Currently it's possible that the player spams 100 events, but half of them are caught in one interval, and the rest in another (since the whole map gets cleared each interval). This means if your threshold is 100, it's only going to count as 50.

These changes make it so the map is processed every pulse, and each entry has its interval tracked individually. No more lost counts

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -4711,17 +4711,35 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
     if (!pPlayer || !pPlayer->IsPlayer() || pPlayer->IsBeingDeleted())
         return;
 
-    m_mapClientTriggeredEvents[pPlayer]++;
+    // If key/player doesn't exist in map, store time of entry
+    if (m_mapClientTriggeredEvents.find(pPlayer) == m_mapClientTriggeredEvents.end())
+        m_mapClientTriggeredEvents[pPlayer].second = GetTickCount64_();
+
+    m_mapClientTriggeredEvents[pPlayer].first++;
 }
 
 void CGame::ProcessClientTriggeredEventSpam()
 {
-    for (const auto& [player, count]: m_mapClientTriggeredEvents)
+    for (const auto& [player, pair]: m_mapClientTriggeredEvents)
     {
-        if (player && player->IsPlayer() && !player->IsBeingDeleted() && count > m_iMaxClientTriggeredEventsPerInterval)
-            player->CallEvent("onPlayerTriggerEventThreshold", {});
+        if (player && player->IsPlayer() && !player->IsBeingDeleted())
+        {
+            if (GetTickCount64_() - pair.second >= m_iClientTriggeredEventsIntervalMs)
+            {
+                if (pair.first > m_iMaxClientTriggeredEventsPerInterval)
+                {
+                    player->CallEvent("onPlayerTriggerEventThreshold", {});
+                    m_mapClientTriggeredEvents.erase(player);
+                }
+
+                m_mapClientTriggeredEvents.erase(player);
+            }
+        }
+        else
+        {
+            m_mapClientTriggeredEvents.erase(player);
+        }
     }
 
-    m_mapClientTriggeredEvents.clear();
     m_lClientTriggeredEventsLastCheck = GetTickCount64_();
 }
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -4724,8 +4724,12 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
 
 void CGame::ProcessClientTriggeredEventSpam()
 {
-    for (const auto& [player, data] : m_mapClientTriggeredEvents)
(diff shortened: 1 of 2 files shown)


Server/mods/deathmatch/logic/CGame.h
```diff
@@ -665,7 +665,6 @@ class CGame
 
     int m_iMaxClientTriggeredEventsPerInterval = 100;
     int m_iClientTriggeredEventsIntervalMs = 1000;
-    long long m_lClientTriggeredEventsLastCheck = 0;
 
-    std::map<CPlayer*, int> m_mapClientTriggeredEvents;
+    std::map<CPlayer*, std::pair<int, int>> m_mapClientTriggeredEvents;
```
tederis: In order to make it more readable I would suggest you to create a new structure instead of using a pair.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4711,17 +4710,34 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
     if (!pPlayer || !pPlayer->IsPlayer() || pPlayer->IsBeingDeleted())
         return;
 
-    m_mapClientTriggeredEvents[pPlayer]++;
+    int now = GetTickCount64_();
```
tederis: GetTickCount64_ returns `long long` which is 8 bytes long. But `int` is only 4 bytes long. That means that when your timer will cover INT_MAX your logic gets broken.

tederis: For example: 
```
struct 
{
    long long m_llTicks;
    uint32_t m_uiCounter;
};
```

Server/mods/deathmatch/logic/CGame.h
tederis: For example:

```
struct 
{
    long long m_llTicks;
    uint32_t m_uiCounter;
};
```

Lpsd: I've added this

Server/mods/deathmatch/logic/CGame.cpp
Lpsd: Resolved

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -665,7 +665,12 @@ class CGame
 
     int m_iMaxClientTriggeredEventsPerInterval = 100;
     int m_iClientTriggeredEventsIntervalMs = 1000;
-    long long m_lClientTriggeredEventsLastCheck = 0;
 
-    std::map<CPlayer*, int> m_mapClientTriggeredEvents;
+    struct ClientTriggeredEventsInfo
+    {
+        long long m_llTicks;
```
tederis: One more thing. I know that std::map::operator[] gives you a value-initialized object and fields are filled with zero. But there can be a situation when you wants to create an object of ClientTriggeredEventsInfo manually. And probably you will forget that those fields are not initialized by default. And this will be a problem. So I recommend you to initialize fields by default values.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4711,17 +4710,34 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
...
+        if (player && player->IsPlayer() && !player->IsBeingDeleted())
+        {
+            if (GetTickCount64_() - data.m_llTicks >= m_iClientTriggeredEventsIntervalMs)
+            {
+                if (data.m_uiCounter > m_iMaxClientTriggeredEventsPerInterval)
+                    player->CallEvent("onPlayerTriggerEventThreshold", {});
 
-    m_mapClientTriggeredEvents.clear();
-    m_lClientTriggeredEventsLastCheck = GetTickCount64_();
+                m_mapClientTriggeredEvents.erase(player);
```
tederis: And I just noticed that you have an undefined behaviour here. Because you cannot erase an element while iterating this way. You have to use iterators. See first erase method: https://en.cppreference.com/w/cpp/container/map/erase. When you need to erase an element you write: `it = map.erase(it)`.

tederis: If there was c++20 I would recommend std::erase_if, but.. we are still using C++17.

Lpsd:
Committed changes from review

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4711,17 +4711,43 @@ void CGame::RegisterClientTriggeredEventUsage(CPlayer* pPlayer)
...
 void CGame::ProcessClientTriggeredEventSpam()
 {
-    for (const auto& [player, count]: m_mapClientTriggeredEvents)
+    for (auto it = m_mapClientTriggeredEvents.begin(); it != m_mapClientTriggeredEvents.end();)
     {
-        if (player && player->IsPlayer() && !player->IsBeingDeleted() && count > m_iMaxClientTriggeredEventsPerInterval)
-            player->CallEvent("onPlayerTriggerEventThreshold", {});
-    }
+        CPlayer* player = it->first;
+        auto     data = it->second;
```
tederis: `const auto& [player, data] = *it;` would be neater.

tederis:
LGTM

Dutchman101 (approved):
Code reviewed by TEDERIs, merge requested by Lpsd

PR #3261, opened 2023-12-05
Fix new vehicle upgrades

Serius41:
This PR resolves issue #2198 by fixing vehicle upgrades on the client for added vehicles (engineRequestModel).

Video:
https://streamable.com/zngjok


Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,6 +64,7 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
         return false;
 
     unsigned short usModel = m_pVehicle->GetModel();
+
```
TheNormalnij:
```suggestion
    unsigned short usModel = m_pVehicle->GetModel();
    usModel = g_pGame->GetModelInfo(usModel)->GetParentID() || usModel;
```
This do same thing, but less code.

Client/multiplayer_sa/CMultiplayerSA_1.3.cpp
```diff
@@ -1435,7 +1435,7 @@ void                       CVehicleModelInterface_SetClump()
 {
     // Loop through all vehicles and find the vehicle id that this interface belongs to
     CModelInfo* pModelInfo = NULL;
-    for (int i = 400; i < 612; i++)
+    for (int i = 400; i < 80000000; i++)
```
TheNormalnij: Why do you need that?

Serius41: It adds more than one parts. Here you specified only the GTA vehicle ids for (int i = 400; i < 612; i++), but it was not checking when there were more ids, so I increased the number. This is a temporary solution. I could not connect the deatmach file to the multiplayer:sa file to capture the parent id. I did something like this for him and it worked.

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
tederis: > This do same thing, but less code.

Your expression will produce 1 all the time.

Client/multiplayer_sa/CMultiplayerSA_1.3.cpp
tederis: It is extremely inefficient and can lead to freezes when spawning vehicles.

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
tederis:
```
if (auto* pModelInfo = g_pGame->GetModelInfo(usModel); pModelInfo && pModelInfo->GetParentID() != 0)
        usModel = static_cast<uint16_t>(pModelInfo->GetParentID());
```
Would be better I guess.

Client/multiplayer_sa/CMultiplayerSA_1.3.cpp
Serius41: How else can I prevent it? I already thought of this I put it as a temporary solution and it worked

Serius41: How else can I prevent it? I already thought of this I put it as a temporary solution and it worked

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
TheNormalnij: Lol. I forgot about this in c++. Ty @tederis

Client/multiplayer_sa/CMultiplayerSA_1.3.cpp
tederis: From what I can see, there is two possible solutions. The first one is to use one of the associative containers to link model ptr to a model ID at the game's start. And the second one is to store model ID in `CBaseModelInfo` class(but it will require reverse engineering skills). Maybe there is something else, it requires additional study.

Serius41: @tederis 
I updated it again, can you check it? I tested it and it worksworks

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,376 +289,28 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
...
-            if (ucSlot == 0 || ucSlot == 1)
-            {
-                if (supportedUpgrades.m_bBonnet == true)
-                {
-                    if (us == 1142 || us == 1144)
-                    {
-                        // We add both left and right for some reason seems to be SA doing it
-                        bReturn &= supportedUpgrades.m_bBonnet_Left;
-                        bReturn &= supportedUpgrades.m_bBonnet_Left_dam;
+    int modelRequestIDParent = g_pGame->GetModelInfo(usModel)->GetParentID();
```
tederis, line 298: Unused variable

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,376 +289,28 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
...
-        return true;
 
     // No nitro or other upgrades for planes
     if (vehicleType == CLIENTVEHICLE_PLANE)
         return false;
 
-    if (us == VEHICLEUPGRADE_NITRO_5X || us == VEHICLEUPGRADE_NITRO_2X || us == VEHICLEUPGRADE_NITRO_10X || us == VEHICLEUPGRADE_HYDRAULICS)
-        return true;
-
     bool bReturn = false;
```
tederis, line 297: Unused variable

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,376 +289,28 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
...
-            }
-            else if (ucSlot == 16)
-            {
-                bReturn = supportedUpgrades.m_bMisc;
-            }
-        }
-    }
-    return bReturn;
+   if (auto* pModelInfo = g_pGame->GetModelInfo(usModel); pModelInfo && pModelInfo->GetParentID() != 0)
+    usModel = static_cast<uint16_t>(pModelInfo->GetParentID());
```
tederis, line 301: Indentation problem

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,376 +289,28 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
...
+
+    size_t vehicleIndex = (usModel - 400);
+    size_t upgradeIndex = (usUpgrade - 1000);
+
+    const VehicleUpgradeSlots& slots = vehicleUpgradeSlots[vehicleIndex];
```
tederis, line 306: This is unsafe since potentially `usModel`(and `usUpgrade`) can be extented in future and you will get out of the range access problem. Maybe you should use `at` method instead of [] to properly handle with the help of exceptions? Or at least add `assert` for the early diagnosis.

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -64,376 +289,28 @@ bool CVehicleUpgrades::IsUpgradeCompatible(unsigned short usUpgrade)
...
+    size_t                     slotIndex = upgradeIndex / 32;
+    uint32_t                   slot = slots[slotIndex];
```
tederis, line 308: The same (line 308)

Fernando-A-Rocha (commented):
@tederis 's reviews make sense

Dutchman101:
Closed in favour of #4114 (a favour i requested, because @Serius41 abandoned the original PR)
All now-passed code reviews and the planned merge of this feature will carry over to the new PR. Consider it done.

PR #3267, opened 2023-12-10
Add UserID Sharing Capability with Consent Dialog (Related with Discord-RPC)

znvjder:
This PR introduces the ability to share discord user IDs with servers if the user grants permission. 
A special dialog has been added (look below), prompting the user (only once) to allow sharing their **UserID**, specifically when '**Allow connecting with Rich Presence**' is enabled, and also when re-enabling this option.

Additionally, this PR adds the function:
`string getDiscordRichPresenceUserID()`

The function returns the user ID as a string if the user has granted permission; Otherwise, an empty string ("") is returned.

rpc

### DIFF
--- a/Client/core/CDiscordRichPresence.cpp
+++ b/Client/core/CDiscordRichPresence.cpp
@@ -280,7 +280,7 @@ void CDiscordRichPresence::SetPresencePartySize(int iSize, int iMax, bool bCusto
     }
 }
 
-void CDiscordRichPresence::SetDiscordUserID(const char* strUserID)
+void CDiscordRichPresence::SetDiscordUserID(const std::string& strUserID)
 {
     if (CVARS_GET_VALUE<bool>("discord_rpc_share_data"))
         m_strDiscordUserID = strUserID;
@@ -291,7 +291,7 @@ std::string CDiscordRichPresence::GetDiscordUserID() const
     if (CVARS_GET_VALUE<bool>("discord_rpc_share_data"))
         return m_strDiscordUserID;
 
-    return "";
+    return {};
 };
 
 #ifdef DISCORD_DISABLE_IO_THREAD
--- a/Client/core/CDiscordRichPresence.cpp
+++ b/Client/core/CDiscordRichPresence.cpp
@@ -280,6 +280,20 @@ void CDiscordRichPresence::SetPresencePartySize(int iSize, int iMax, bool bCusto
     }
 }
 
+void CDiscordRichPresence::SetDiscordUserID(const char* strUserID)
+{
+    if (CVARS_GET_VALUE<bool>("discord_rpc_share_data"))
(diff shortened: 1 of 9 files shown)


Client/core/CDiscordRichPresence.cpp
```diff
@@ -280,6 +280,20 @@ void CDiscordRichPresence::SetPresencePartySize(int iSize, int iMax, bool bCusto
     }
 }
 
+void CDiscordRichPresence::SetDiscordUserID(const char* strUserID)
+{
+    if (CVARS_GET_VALUE<bool>("discord_rpc_share_data"))
+        m_strDiscordUserID = strUserID;
+}
+
+std::string CDiscordRichPresence::GetDiscordUserID() const
```
tederis, line 289: Isn't it better to return a const reference instead of a string copy?

Client/mods/deathmatch/logic/luadefs/CLuaDiscordDefs.cpp
```diff
@@ -266,3 +266,13 @@ bool CLuaDiscordDefs::IsDiscordRPCConnected()
 
     return discord->IsDiscordClientConnected();
 }
+
+std::string CLuaDiscordDefs::GetDiscordUserID()
+{
+    auto discord = g_pCore->GetDiscord();
+
+    if (!discord || !discord->IsDiscordRPCEnabled())
+        return "";
```
tederis: I think `return {};` would be preferable.

1Lorenzo-1:
What's so private about initiating a dialogue to ask players if they want to share their Discord ID? It's just a Discord ID; if the player has allowed RPC, that's sufficient to obtain their Discord ID.

Similarly, if I set a custom code for the player in RPC and then retrieve it using an internal discord, bot that searches for members with a custom ID in their RPC. 
I can get the dis-id too!

znvjder:
> What's so private about initiating a dialogue to ask players if they want to share their Discord ID? It's just a Discord ID; if the player has allowed RPC, that's sufficient to obtain their Discord ID.
> 
> Similarly, if I set a custom code for the player in RPC and then retrieve it using an internal discord, bot that searches for members with a custom ID in their RPC. I can get the dis-id too!

You are partially correct, but we discussed about this some time ago on the developer channel. In general, Multi Theft Auto is open-source software that includes the option for synchronization with Rich Presence, and third parties can manage this (display various data).

Since the Discord Identifier (userid) is treated as personal data, obtaining and sharing it with third parties without explicit user consent may violate privacy policies and Discord's terms.

Therefore, we decided to be transparent with users and ask for their consent to share the identifier through a pop-up window. We should respect the privacy of users.

Dutchman101 (approved):
The design with privacy in mind is good (and as said, what we want). It's not a big change and has been code reviewed. Thanks

PR #3272, opened 2023-12-16
Added server side weapon related checks

NanoBob:
- Added a check to see if a player has a certain weapon before forwarding a bullet sync packet
- Clear the weapon related data from a puresync packet if the player doesn't currently have that weapon

The reasoning for both of these changes has to do with people running hacks and/or Lua injectors client sided. In cases like these (which I've verified with a client running such cheats) clients will trigger puresync packets and/or bullet sync packets for weapons the player doesn't actually have.

While the server doesn't process these puresync packets, they still relay them to other players. Meaning these weapons are visible to other players (even though according to the server the player will not have the weapons).

For the bullet sync packet the server would blindly forward them, meaning a compromised client would be able to fire any weapon, without even having that weapon.

This PR aims to add two relatively simple checks to prevent this from happening.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -45,7 +45,7 @@ jobs:
       - name: Create build artifacts
         run: utils\premake5 compose_files
 
-      - uses: actions/upload-artifact@master
+      - uses: actions/upload-artifact@v3
         with:
           name: InstallFiles
           path: InstallFiles/
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -2443,6 +2443,10 @@ void CGame::Packet_Bulletsync(CBulletsyncPacket& Packet)
     CPlayer* pPlayer = Packet.GetSourcePlayer();
     if (pPlayer && pPlayer->IsJoined())
     {
+        // Early return when the player attempts to fire a weapon they do not have
+        if (!pPlayer->HasWeaponType(Packet.m_WeaponType))
+            return;
+
         // Relay to other players
         RelayNearbyPacket(Packet);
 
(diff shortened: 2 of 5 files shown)


PlatinMTA:
So that was the problem all this time along. Lets hope this gets merged asap.

Lpsd (approved):
lgtm :shipit:

PlatinMTA:
https://streamable.com/uc3250

Seems to keep happening to some extent. Version r22335 (Linux x64). 

Useful information:
* The server knows the amount of bullets the user has in the slot, but returns 0 as a weapon.
* The server does not know when the player is shooting ("onPlayerWeaponFire" is not working, probably because weapon == 0).

That's about it.

Pirulax:
What about explosions/damage caused by those shots (rpgs for example)?
We should stop those from being relayed too (if not already, not sure how the weapon sync works).

.github/workflows/build.yaml
```diff
@@ -45,7 +45,7 @@ jobs:
       - name: Create build artifacts
         run: utils\premake5 compose_files
 
-      - uses: actions/upload-artifact@master
```
Pirulax, line 48: is this intentional?

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -175,6 +175,12 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
             // Set weapon slot
             if (bWeaponCorrect)
                 pSourcePlayer->SetWeaponSlot(uiSlot);
+            else
+            {
+                // remove invalid weapon data to prevent this from being relayed to other players
```
Pirulax, line 180: I think it would be nice to add a comment explaining what could cause this (Cheaters, etc).

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2443,6 +2443,10 @@ void CGame::Packet_Bulletsync(CBulletsyncPacket& Packet)
     CPlayer* pPlayer = Packet.GetSourcePlayer();
     if (pPlayer && pPlayer->IsJoined())
     {
+        // Early return when the player attempts to fire a weapon they do not have
```
Pirulax, line 2446: I think it would be nice to add a comment explaining what could cause this (Cheaters, etc).

Issue #2755, opened 2022-09-19
Corona reflections do not render on custom placed objects


lopezloo:
### Describe the bug

Corona reflections do not render on custom placed objects.

### Steps to reproduce

```lua
crun setCoronaReflectionsEnabled(2) Object(5501, 0, 0, 10) me.position = Vector3(0, 0, 10 + 1) marker = Marker(0, 0, 10 + 1, "corona", 1.0) marker:setCoronaReflectionEnabled(true)
```

### Additional context

Similar to #450.

lopezloo:
Most likely fixed by #3274. Feedback appreciated.

Fernando-A-Rocha:
I just confirmed with this code that #3274 resolved this! This issue can be closed.

```lua
setElementPosition(getPedOccupiedVehicle(localPlayer) or localPlayer, 0, 0, 11)
setCoronaReflectionsEnabled(2) -- enable even during day

-- This building will reflect light
createBuilding(5501, 0, 0, 10, 0, 0, 0, 0)
local marker = createMarker(0, 0, 11, "corona", 1.0)
setCoronaReflectionEnabled(marker, true)

-- This object won't
createObject(5501, 0, 20, 10, 0, 0, 0)
local marker2 = createMarker(0, 20, 11, "corona", 1.0, 255, 255, 0)
setCoronaReflectionEnabled(marker2, true)
```

![image](https://github.com/user-attachments/assets/a6f5c1d4-4ecd-4d46-be75-8a1d5be385c2)

Fernando-A-Rocha:
> Corona reflections do not render on custom placed objects.

Solution: use buildings instead, to have the intended behavior for world objects :P

lopezloo:
Fixed in #3274.

Issue #1921, opened 2020-12-15
Player vehicle camera ignores collisions of placed objects when moving above certain speed


LeXDDD:
**Describe the bug**

Camera ignores any placed object collisions when player in vehicle reaches certain speed.

**To reproduce**

1. Create any building/road or any other, preferably big, object on map in clear space.
2. Get into any vehicle that can accelerate and move beyond 40km/h.
3. While in vehicle, move close to the placed object, rotate camera in the direction of placed object.
5. When reached 40 km/h, camera will ignore created object collision and clip inside/past object.

**Expected behaviour**

Camera wont ignore objects collisions and will close up to player/move in front of the object, as it would when player moving in vehicle nearby any base game world objects. 
 
**Screenshots**

![0355](https://user-images.githubusercontent.com/62285061/102171085-c412ed00-3ea6-11eb-9adb-9042f1fc7930.png)
![0352](https://user-images.githubusercontent.com/62285061/102171092-c7a67400-3ea6-11eb-85d1-89a102eb1221.png)

**Version**

1.5.7-9.20358.0, 1.5.8
Windows 7

**Additional context**

https://www.youtube.com/watch?v=GsedO_5iSBY
This issue only occurs when player in vehicle, and total sum of velocity values in both `x`, and `y` is above ~0.4. 
While on foot this doesnt seem to happen, despite player moving with same or higher velocity.
Any `setCameraClip` values dont affect issue in any way. 
Seems like same effect, but affecting only player vehicle, can be spotted in base game when moving above 40km/h, which will make camera ignore player vehicle collision and clip through. Other than that one instance i was unable to find same effect affecting anything else in base game.

xLive:
Duplicate of #1885

LeXDDD:
https://community.multitheftauto.com/?p=resources&s=details&id=18365
Quick resource made for testing purpouses, includes: 
objects created on server and client sides(left side objects with hill creation scripted in client script, right side objects with buildings in server script); 
separate building and object `isLineOfSightClear` triggers between player vehicle and camera positions; 
`setCameraClip` switch by using `/clip5` which only seems to affect base game world objects; 
additional info box with all necessary parameters. 
Teleport to testing area `/cake5`.
Perhaps introducing _building_ attribute to `object` class , might solve this issue, since all of the MTA created objects, camera wise, are treated same way as dynamic base game objects such as lamp poles and boxes.

lopezloo:
[...]

PR #3274, opened 2023-12-18
Lua API for buildings

### LINKED ISSUE #450: Created objects don't reflect lights / shadows

TheNormalnij:
Fixes #450, #2755, #1921

API:
```
element-building createBuilding(number modelId, float x, float, y, float z, float rx, float ry, float rz [,  number interior = 0])
```

### DIFF
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -28,6 +28,11 @@ void CFileLoaderSA::StaticSetHooks()
     HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
 }
 
+CBuildingSAInterface* CFileLoaderSA::LoadFileObjectInstance(SFileObjectInstance* obj)
+{
+    return ((CBuildingSAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
+}
+
 class CAtomicModelInfo
 {
 public:
@@ -201,7 +206,7 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
     return atomic;
 }
 
-CEntitySAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
+CBuildingSAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
 {
     char                szModelName[24];
     SFileObjectInstance inst;
@@ -217,5 +222,5 @@ CEntitySAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
     if (fLenSq > 0.0f && std::fabs(fLenSq - 1.0f) > std::numeric_limits<float>::epsilon())
         inst.rotation /= std::sqrt(fLenSq);
 
-    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(&inst);
+    return ((CBuildingSAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(&inst);
 }
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -28,9 +28,9 @@ void CFileLoaderSA::StaticSetHooks()
     HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
 }
 
-CBuildingSAInterface* CFileLoaderSA::LoadFileObjectInstance(SFileObjectInstance* obj)
+CEntitySAInterface* CFileLoaderSA::LoadFileObjectInstance(SFileObjectInstance* obj)
 {
-    return ((CBuildingSAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
+    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
 }
 
 class CAtomicModelInfo
@@ -206,7 +206,7 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
     return atomic;
 }
 
-CBuildingSAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
+CEntitySAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
 {
     char                szModelName[24];
     SFileObjectInstance inst;
@@ -222,5 +222,5 @@ CBuildingSAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
     if (fLenSq > 0.0f && std::fabs(fLenSq - 1.0f) > std::numeric_limits<float>::epsilon())
         inst.rotation /= std::sqrt(fLenSq);
 
-    return ((CBuildingSAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(&inst);
+    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(&inst);
 }
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -28,7 +28,7 @@ void CFileLoaderSA::StaticSetHooks()
     HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
 }
 
-CEntitySAInterface* CFileLoaderSA::LoadFileObjectInstance(SFileObjectInstance* obj)
+CEntitySAInterface* CFileLoaderSA::LoadObjectInstance(SFileObjectInstance* obj)
 {
     return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
 }
--- a/Client/game_sa/CFileLoaderSA.cpp
+++ b/Client/game_sa/CFileLoaderSA.cpp
@@ -30,7 +30,8 @@ void CFileLoaderSA::StaticSetHooks()
 
 CEntitySAInterface* CFileLoaderSA::LoadObjectInstance(SFileObjectInstance* obj)
 {
-    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
+    // Second argument is model name. It's unused in the function
+    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*, const char*))0x538090)(obj, nullptr);
 }
 
 class CAtomicModelInfo
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -421,6 +421,9 @@ void CPoolsSA::RemoveBuilding(CBuilding* pBuilding)
     // Delete it from memory
     delete pBuildingSA;
 
+    // Remove building from SA pool
+    (*m_ppBuildingPoolInterface)->Release(dwElementIndexInPool);
+
     // Decrease the count of elements in the pool
     --m_buildingPool.ulCount;
 }
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -374,8 +374,7 @@ CBuilding* CPoolsSA::AddBuilding(CClientBuilding* pClientBuilding, uint16_t mode
     // Fix strange SA rotation 
     instance.rotation.fW = -instance.rotation.fW;
 
-    CFileLoaderSA loader{};
-    auto pBuilding = static_cast<CBuildingSAInterface*>(loader.LoadObjectInstance(&instance));
+    auto pBuilding = static_cast<CBuildingSAInterface*>(CFileLoaderSA::LoadObjectInstance(&instance));
 
     // Disable lod and ipl
     pBuilding->m_pLod = nullptr;
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -360,6 +360,9 @@ inline bool CPoolsSA::AddBuildingToPool(CClientBuilding* pClientBuilding, CBuild
 
 CBuilding* CPoolsSA::AddBuilding(CClientBuilding* pClientBuilding, uint16_t modelId, CVector *vPos, CVector4D *vRot, uint8_t interior)
 {
+    if (!HasFreeBuildingSlot())
+        return nullptr;
+
     // Load building
     SFileObjectInstance instance;
     instance.modelID = modelId;
@@ -421,6 +424,11 @@ void CPoolsSA::RemoveBuilding(CBuilding* pBuilding)
 
     // Decrease the count of elements in the pool
     --m_buildingPool.ulCount;
+}
+
+bool CPoolsSA::HasFreeBuildingSlot()
+{
+    return (*m_ppBuildingPoolInterface)->GetFreeSlot() != -1;
 }
 
 //////////////////////////////////////////////////////////////////////////////////////////
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -364,8 +364,7 @@ inline bool CPoolsSA::AddBuildingToPool(CClientBuilding* pClientBuilding, CBuild
 
 CBuilding* CPoolsSA::AddBuilding(class CClientBuilding* pClientBuilding, uint16_t modelId, CVector vPos, CVector4D vRot, uint8_t interior)
 {
-    CFileLoaderSA loader{};
-
+    // Load building
     SFileObjectInstance instance;
     instance.modelID = modelId;
     instance.lod = -1;
@@ -373,15 +372,23 @@ CBuilding* CPoolsSA::AddBuilding(class CClientBuilding* pClientBuilding, uint16_
     instance.position = vPos;
     instance.rotation = vRot;
 
+    CFileLoaderSA loader{};
     auto pBuilding = loader.LoadFileObjectInstance(&instance);
 
+    // Disable lod and ipl
     pBuilding->m_pLod = nullptr;
     pBuilding->m_iplIndex = 0;
 
-    pBuilding->AddRect();
+    // Always stream model collosion
+    // TODO We can setup collison bounding box and use GTA streamer for it
+    auto modelInfo = pGame->GetModelInfo(modelId);
+    modelInfo->AddColRef();
 
+    // Add building in world
     auto pBuildingSA = new CBuildingSA(pBuilding);
+    pGame->GetWorld()->Add(pBuildingSA, CBuildingPool_Constructor);
 
+    // Add CBuildingSA object in pool
     AddBuildingToPool(pClientBuilding, pBuildingSA);
 
     return pBuildingSA;
@@ -399,6 +406,14 @@ void CPoolsSA::RemoveBuilding(CBuilding* pBuilding)
         return;
     }
 
+    // Remove building from world
+    pGame->GetWorld()->Remove(pInterface, CBuildingPool_Destructor);
+
+    // Remove col reference
+    auto modelInfo = pGame->GetModelInfo(pBuilding->GetModelIndex());
+    modelInfo->RemoveColRef();
+
+    // Remove from BuildingSA pool
     auto* pBuildingSA = m_buildingPool.arrayOfClientEntities[dwElementIndexInPool].pEntity;
     m_buildingPool.arrayOfClientEntities[dwElementIndexInPool] = {nullptr, nullptr};
 
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -362,15 +362,18 @@ inline bool CPoolsSA::AddBuildingToPool(CClientBuilding* pClientBuilding, CBuild
     return true;
 }
 
-CBuilding* CPoolsSA::AddBuilding(class CClientBuilding* pClientBuilding, uint16_t modelId, CVector vPos, CVector4D vRot, uint8_t interior)
+CBuilding* CPoolsSA::AddBuilding(CClientBuilding* pClientBuilding, uint16_t modelId, CVector *vPos, CVector4D *vRot, uint8_t interior)
 {
     // Load building
     SFileObjectInstance instance;
     instance.modelID = modelId;
     instance.lod = -1;
     instance.interiorID = interior;
-    instance.position = vPos;
-    instance.rotation = vRot;
+    instance.position = *vPos;
+    instance.rotation = *vRot;
+
+    // Fix strange SA rotation 
+    instance.rotation.fW = -instance.rotation.fW;
 
     CFileLoaderSA loader{};
     auto pBuilding = loader.LoadFileObjectInstance(&instance);
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -375,7 +375,7 @@ CBuilding* CPoolsSA::AddBuilding(CClientBuilding* pClientBuilding, uint16_t mode
     instance.rotation.fW = -instance.rotation.fW;
 
     CFileLoaderSA loader{};
-    auto pBuilding = static_cast<CBuildingSAInterface*>(loader.LoadFileObjectInstance(&instance));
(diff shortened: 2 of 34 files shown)


TheNormalnij:
image

tederis:
I think Euler angles should be used instead of quaternions. Quaternions are less intuitive and will be a nightmare for the most of beginners. Or maybe there should be an enbedded function for the Euler <-> Quaternion conversion. In addition, it would be easier to move previously created maps on `createBuilding` without an effort.

TheNormalnij:
> I think Euler angles should be used instead of quaternions. Quaternions are less intuitive and will be a nightmare for the most of beginners. Or maybe there should be an enbedded function for the Euler <-> Quaternion conversion. In addition, it would be easier to move previously created maps on `createBuilding` without an effort.

Yes, i think so. I didn't found  euler to quaternion functions.

Shared/mods/deathmatch/logic/Utils.h
```diff
@@ -246,6 +246,21 @@ inline float GetSmallestWrapUnsigned(float fValue, float fHigh)
...
+    double sy = sin(vecFrom.fZ * 0.5);
+    double cr = cos(vecFrom.fX * 0.5);
+    double sr = sin(vecFrom.fX * 0.5);
+    double cp = cos(vecFrom.fY * 0.5);
+    double sp = sin(vecFrom.fY * 0.5);
+
+    vecTo.fW = cy * cr * cp + sy * sr * sp;
+    vecTo.fX = cy * sr * cp - sy * cr * sp;
+    vecTo.fY = cy * cr * sp + sy * sr * cp;
+    vecTo.fY = sy * cr * cp - cy * sr * sp;
```
Zangomangu: This doesn't look right

Client/game_sa/CPoolsSA.cpp
```diff
@@ -329,6 +332,101 @@ void CPoolsSA::DeleteAllObjects()
...
+inline bool CPoolsSA::AddBuildingToPool(CClientBuilding* pClientBuilding, CBuildingSA* pBuilding)
+{
+    // Grab the new object interface
+    CBuildingSAInterface* pInterface = pBuilding->GetBuildingInterface();
+
+    if (!pInterface)
+    {
+        return false;
+    }
+    else
```
Zangomangu: can remove `else`

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -28,6 +28,11 @@ void CFileLoaderSA::StaticSetHooks()
     HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
 }
 
+CBuildingSAInterface* CFileLoaderSA::LoadFileObjectInstance(SFileObjectInstance* obj)
```
Zangomangu: LoadObjectInstance?

Zangomangu:
We're going from euler, to quaternion, to matrix now
Could we also please allow skipping euler ?

```
element-building createBuilding(number modelId, float x, float, y, float z, float rx, float ry, float rz, float rw [, number interior = 0])
```
If `rw` is nil interpret as Euler otherwise quaternion

Client/game_sa/CFileLoaderSA.cpp
TheNormalnij: This is common name for this function. It is used in gta-reverse-modern and  ida ipb

Zangomangu: yeah i mean it should be LoadObjectInstance instead of LoadFileObjectInstance to match

TheNormalnij:
> Could we also please allow skipping euler ?

We can add quaternion -> euler utility function

Zangomangu:
Well the point would be to get rid of that conversion altogether (where some euler angles are problematic) since the GTA function accepts a quaternion.

But i'm not too steadfast on it, since it can be worked around by setting matrix.
I can see the point where scripters are used to euler angles and we want to keep it similar.

Client/game_sa/CFileLoaderSA.cpp
TheNormalnij: It loads SFileObjectInstance object. So function name describes it. idk

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -201,7 +206,7 @@ RpAtomic* CFileLoader_SetRelatedModelInfoCB(RpAtomic* atomic, SRelatedModelInfo*
     return atomic;
 }
 
-CEntitySAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
+CBuildingSAInterface* CFileLoader_LoadObjectInstance(const char* szLine)
```
tederis: Why do you need this change from **CEntitySAInterface** to **CBuildingSAInterface**? From what I can see this function should return CEntity*(or CEntitySAInterface* in terms of MTA).

Client/mods/deathmatch/logic/CClientBuilding.cpp
```diff
@@ -0,0 +1,83 @@
...
+    : ClassInit(this),
+      CClientEntity(ID),
+      m_pBuildingManager(pManager->GetBuildingManager()),
+      m_usModelId(usModelId),
+      m_vPos(pos),
+      m_vRot(rot),
+      m_interior(interior)
+{
+    m_pBuilding = nullptr;
+    m_pManager = pManager;
```
tederis, line 23: It's better to move the initialization of these members to the member initializer list.

Client/mods/deathmatch/logic/CClientBuilding.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+void CClientBuilding::SetRotationRadians(const CVector& vecRadians)
+{
+    m_vRot = vecRadians;
+    Recreate();
+}
+
+void CClientBuilding::SetInterior(uint8_t ucInterior)
+{
+    m_interior = ucInterior;
```
tederis, line 78: The method **Recreate** is heavy and we could avoid it by adding a comparison: `if (m_interior == ucInterior) return;` The same is true for other methods.

Client/mods/deathmatch/logic/CClientBuilding.cpp
```diff
@@ -0,0 +1,83 @@
...
+    m_pManager = pManager;
+    SetTypeName("building");
+    m_pBuildingManager->AddToList(this);
+    Create();
+}
+
+CClientBuilding::~CClientBuilding()
+{
+    m_pBuildingManager->RemoveFromList(this);
+    if (m_pBuilding)
```
tederis: This check is already exists within `Destroy` method. I think double check is unnecessary.

Client/mods/deathmatch/logic/CClientBuilding.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <game/CBuilding.h>
+
+class CClientBuilding : public CClientEntity
+{
+    DECLARE_CLASS(CClientBuilding, CClientEntity)
+    friend class CClientBuildingManager;
+
+public:
+    CClientBuilding(class CClientManager* pManager, ElementID ID, uint16_t usModelId, CVector &pos, CVector &rot, uint8_t interior);
```
tederis: I see the lack of `const` qualifiers in front of `&pos` and `&rot`. Was this done intentionally?

Client/mods/deathmatch/logic/CClientBuildingManager.cpp
```diff
@@ -0,0 +1,66 @@
...
+    RemoveAll();
+}
+
+void CClientBuildingManager::RemoveAll()
+{
+    // Make sure they don't remove themselves from our list
+    m_bRemoveFromList = false;
+
+    // Run through our list deleting the buildings
+    auto iter = m_List.begin();
```
tederis: Is the choice in favor of iterators intentional? Maybe better to use a range based for loop? And I see that after removing all objects the list it is still contains pointers to them. I think `m_List.clear()` chould be called.

Client/mods/deathmatch/logic/CClientBuildingManager.cpp
```diff
@@ -0,0 +1,66 @@
...
+        delete *iter;
+    }
+
+    // Allow list removal again
+    m_bRemoveFromList = true;
+}
+
+bool CClientBuildingManager::Exists(CClientBuilding* pBuilding)
+{
+    // Matches given DFF?
```
tederis: DFF?

Client/mods/deathmatch/logic/CClientBuildingManager.cpp
```diff
@@ -0,0 +1,66 @@
...
+    auto iter = m_List.begin();
+    for (; iter != m_List.end(); iter++)
```
tederis: Is the choice in favor of iterators intentional? Maybe better to use a range based for loop? Or even `std::find`?

Client/mods/deathmatch/logic/CClientBuildingManager.h
```diff
@@ -0,0 +1,41 @@
...
+    friend class CClientBuilding;
+
+public:
+    CClientBuildingManager(class CClientManager* pManager);
+    ~CClientBuildingManager();
+
+    void RemoveAll();
+    bool Exists(CClientBuilding* pBuilding);
+
+    std::list<CClientBuilding*>::const_iterator IterBegin() { return m_List.begin(); }
```
tederis: Maybe better to just return a const ref to a list?

Shared/mods/deathmatch/logic/Utils.h
```diff
@@ -246,6 +246,21 @@ inline float GetSmallestWrapUnsigned(float fValue, float fHigh)
 
 void RotateVector(CVector& vecLine, const CVector& vecRotation);
 
+inline void ConvertEulersToQuaternion(const CVector& vecFrom, CVector4D vecTo)
```
tederis: Looks like `vecTo` is expected to be passed by reference.

Client/game_sa/CPoolsSA.cpp
```diff
@@ -703,6 +805,18 @@ DWORD CPoolsSA::GetObjectPoolIndex(std::uint8_t* pInterface)
     return ((pInterface - pTheObjects) / dwAlignedSize);
 }
 
+DWORD CPoolsSA::GetBuildingPoolIndex(std::uint8_t* pInterface)
+{
+    DWORD         dwAlignedSize = 412;
```
tederis: It should be 56 bytes

Client/mods/deathmatch/logic/CClientManager.h
```diff
@@ -96,6 +97,7 @@ class CClientManager
     CClientEffectManager*        GetEffectManager() { return m_pEffectManager; }
     CClientPointLightsManager*   GetPointLightsManager() { return m_pPointLightsManager; }
     CClientIMGManager*           GetIMGManager() { return m_pImgManager; }
+    CClientBuildingManager*      GetBuildingManager() { return m_pBuildingManager; }
```
tederis: I think **const** qualifier should be added for this method.

Shared/mods/deathmatch/logic/Utils.h
```diff
@@ -246,6 +246,21 @@ inline float GetSmallestWrapUnsigned(float fValue, float fHigh)
...
+{
+    double cy = cos(vecFrom.fZ * 0.5);
```
tederis: Trigonometric functions accept radians. But from what I can see `vecFrom` contains degrees.

tederis:
**createBuilding** function must check if **modelId** argument is actually a building model. It can be done by looking at **usDynamicIndex** field: `auto modelInfo = pGame->GetModelInfo(modelId);
if (modelInfo->GetInterface()->usDynamicIndex != 0xffff)
    return nullptr;`

Otherwise GTA will create a dummy object instead of a building and this type of object will never be deleted. You can check what will happened if create a building with model ID 3797.

tederis:
I also noticed that shadows and grass are still on the map even after a resource restart. But they disappear after the respawn(I guess shadows and grass are spawned once after the building creation).

tederis:
There also a problem with the rotation. When you spawn an object and a building with the same coordinates and the same rotation, their actual transformation differs.

mta-screen_2023-12-25_16-05-51

Client/mods/deathmatch/logic/CClientBuilding.cpp
TheNormalnij: m_pManager is member of parent class

Shared/mods/deathmatch/logic/Utils.h
TheNormalnij: Oh, yes. This is a mess

Client/mods/deathmatch/logic/CClientBuilding.cpp
tederis: Oh, ok. Didn't notice that.

TheNormalnij: But i moved m_pBuilding. It was useful

Client/game_sa/CFileLoaderSA.cpp
TheNormalnij: I think current variant describes this function better. CFileLoader_LoadObjectInstance always returns a building object and not a another class object inherited from CEntitySAInterface

Client/mods/deathmatch/logic/CClientBuildingManager.cpp
TheNormalnij: Yep :p

Client/game_sa/CFileLoaderSA.cpp
tederis: LoadObjectInstance can return either CBuilding or CDummyObject as can be verified by looking at reversed SA. You can pass model 3797 into this function and it will return CDummyObject which is not a building.

TheNormalnij:
> also noticed that shadows and grass are still on the map even after a resource restart. But they disappear after the respawn(I guess shadows and grass are spawned once after the building creation).

This can be fixed manually via `setWorldSpecialPropertyEnabled('randomfoliage', false)`.

TheNormalnij:
> There also a problem with the rotation. When you spawn an object and a building with the same coordinates and the same rotation, their actual transformation differs.

Fixed

tederis:
> > also noticed that shadows and grass are still on the map even after a resource restart. But they disappear after the respawn(I guess shadows and grass are spawned once after the building creation).
> 
> This can be fixed manually via `setWorldSpecialPropertyEnabled('randomfoliage', false)`.

This disables the grass no doubt. But what if I want to keep grass on the default SA objects? And what about the shadows?
mta-screen_2023-12-25_19-29-14

tederis:
> > > also noticed that shadows and grass are still on the map even after a resource restart. But they disappear after the respawn(I guess shadows and grass are spawned once after the building creation).
> > 
> > 
> > This can be fixed manually via `setWorldSpecialPropertyEnabled('randomfoliage', false)`.
> 
> This disables the grass no doubt. But what if I want to keep grass on the default SA objects? And what about the shadows? mta-screen_2023-12-25_19-29-14

My opinion is that we should restream(reload or recreate) grass and shadows somehow at each change in the buildings. Or even better leave a mark that grass(and shadows) should be recreated and do it in the next frame to eliminate tons of excessive calculations.

TheNormalnij:
> My opinion is we should restream grass and shadows somehow at each change in the buildings.

It causes performance degradation when you have significant amount of buildings. I prefer additional `restreamShadows` function.

tederis:
> > My opinion is we should restream grass and shadows somehow at each change in the buildings.
> 
> It causes performance degradation when you have significant amount of buildings. I prefer additional `restreamShadows` function.

See my additions to the comment: `Or even better leave a mark that grass(and shadows) should be recreated and do it in the next frame to eliminate tons of excessive calculations.` In my opinion a deletion procedure should leave the game as it was before the building creation. It should be by default. But function `restreamShadows` can be added as well.

TheNormalnij:
> **createBuilding** function must check if **modelId** argument is actually a building model. It can be done by looking at **usDynamicIndex** field: `auto modelInfo = pGame->GetModelInfo(modelId); if (modelInfo->GetInterface()->usDynamicIndex != 0xffff) return nullptr;`

Fixed too

TheNormalnij:
> In my opinion a deletion procedure should leave the game as it was before the building creation.

Yes. This is right. But i'd rather have manual control in API to prevent unnecessary glass or shadows flickering.

I think this is not a big deal for now. Glass and shadows disappear itself after some distance

Client/game_sa/CFileLoaderSA.cpp
TheNormalnij: Ah... I see

Client/game_sa/CPoolsSA.cpp
```diff
@@ -329,6 +332,105 @@ void CPoolsSA::DeleteAllObjects()
...
+    instance.lod = -1;
+    instance.interiorID = interior;
+    instance.position = *vPos;
+    instance.rotation = *vRot;
+
+    // Fix strange SA rotation 
+    instance.rotation.fW = -instance.rotation.fW;
+
+    CFileLoaderSA loader{};
+    auto pBuilding = static_cast<CBuildingSAInterface*>(loader.LoadObjectInstance(&instance));
```
tederis: What is the reason for using a non-static **LoadObjectInstance** method? Isn't it better to make it static and just `CFileLoaderSA::LoadObjectInstance(&instance)`? From what I can see, this method is static in SA(see reversed SA or IDA).

Client/game_sa/CFileLoaderSA.cpp
```diff
@@ -28,6 +28,11 @@ void CFileLoaderSA::StaticSetHooks()
     HookInstall(0x538690, (DWORD)CFileLoader_LoadObjectInstance, 5);
 }
 
+CEntitySAInterface* CFileLoaderSA::LoadObjectInstance(SFileObjectInstance* obj)
+{
+    return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*))0x538090)(obj);
```
tederis: This function (0x538090) is expecting a second `const char`* argument. See the function's prologue:
modelName

tederis: If fact, this argument is no actually used. So you can just rewrite it the following way: `return ((CEntitySAInterface * (__cdecl*)(SFileObjectInstance*, const char*))0x538090)(obj, nullptr);`

Client/game_sa/CPoolsSA.cpp
TheNormalnij: For some architectural reasons i prefer loader classes over functions.
I can make it simpler with using static function, because we don't have DI / tests for this code

Client/game_sa/CFileLoaderSA.cpp
TheNormalnij: Didn't notice that

Client/game_sa/CPoolsSA.cpp
tederis: This is OK. But I prefer to keep the code as close to SA as possible (within the bounds of game_sa module).

Zangomangu:
Nice work on this pr
Getting 30 % less cpu usage with createBuilding over createObject

Zangomangu:
#1921 is also fixed with this PR

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Zangomangu, line 1: Missing getElementModel
I think you should go through the getter/setters here related to CClientObject and determine which should also work for CClientBuilding.

E.g.
- get/setElementCollisionsEnabled
- getElementBoundingBox
- getElementRadius
- getElementDistanceFromCentreOfMassToBaseOfModel
- get/setElementAlpha
- "CollidableWith" functions

TheNormalnij: getElementModel is fixed
All another functions will be added later. I keep this PR small to make review and testing much easier.

tederis:
There are still problems with the rotation(it doesn't match the rotation that objects created via createObject have). I believe that this is an important point because most of the players will use buildings as a substitute for objects.
mta-screen_2023-12-28_12-02-21

TheNormalnij:
Oh. rx between 90 and 270 is broken

TheNormalnij:
I think there is no more issues with rotation

Shared/mods/deathmatch/logic/Utils.h
```diff
@@ -246,19 +246,20 @@ inline float GetSmallestWrapUnsigned(float fValue, float fHigh)
...
-    double cr = cos(vecFrom.fX * 0.5);
-    double sr = sin(vecFrom.fX * 0.5);
-    double cp = cos(vecFrom.fY * 0.5);
-    double sp = sin(vecFrom.fY * 0.5);
-
-    vecTo.fW = cy * cr * cp + sy * sr * sp;
-    vecTo.fX = cy * sr * cp - sy * cr * sp;
-    vecTo.fY = cy * cr * sp + sy * sr * cp;
-    vecTo.fZ = sy * cr * cp - cy * sr * sp;
+    const float c1 = cos(vecFrom.fX / 2.0f);
```
tederis: I suggest you use a multiplication instead of the division due to the performance issues.

TheNormalnij: fixed

Client/mods/deathmatch/logic/CClientBuildingManager.cpp
```diff
@@ -0,0 +1,84 @@
...
+        return false;
+    }
+
+    eModelInfoType eType = pModelInfo->GetModelType();
+    return (eType == eModelInfoType::CLUMP || eType == eModelInfoType::ATOMIC || eType == eModelInfoType::WEAPON || eType == eModelInfoType::TIME);
+}
+
+bool CClientBuildingManager::IsValidPosition(const CVector& pos) noexcept
+{
+    return (pos.fX >- -3000.0f && pos.fX <= 3000.0f && pos.fY >= -3000.0f && pos.fY <= 3000.0f);
```
tederis: `pos.fX >- -3000.0f` misprint

tederis: Aside of it, I would use a constexpr variable here.

TheNormalnij: fixed

tederis:
The PR is mature enough to be merged I think. Problems related to glass and shadows(see my comments above) can be addressed in the next PRs. LODs can be also implemented in future commits.

tederis (approved):
The PR is mature enough to be merged I think. Problems related to grass and shadows(see my comments above) can be addressed in the next PRs. LODs can also be implemented in future commits.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
tederis: I propose to move it to the next commits and merge the PR as is. I hope @TheNormalnij will be able to address proposed changes shortly after the merge.

Zangomangu (approved):
agree with tederis

BenhurPaz:
Merge this!

MrSol0:
I come here to add my voice with the previous people and say please merge this ASAP as we need it in our upcoming project.

Einheit-101:
What about:
-LOD Distance (How far can we see buildings? Infinite draw distance like in MTA:Eir?)
-Shaders - should we be able to restrict shaders to buildings?
-Can buildings be altered with setElementPosition and other functions?

TheNormalnij:
> -LOD Distance (How far can we see buildings? (Infinite draw distance, like in MTA:Eir?)

Draw distance is limited by GTA SA only. Buildings have equal draw distance with SA world world elements.

> -Shaders - should we be able to restrict shaders to buildings?

I think only global shaders  work. Personally, i don't prioritize shaders over dynamic pool size / LODs  / dynamic map box.

> -Can buildings be altered with setElementPosition and other functions?

setPostion / setRotation / setMatrix functions  work, but they are expensive. I suggest not use them

Dutchman101:
@TheNormalnij [https://wiki.multitheftauto.com/wiki/CreateBuilding](https://wiki.multitheftauto.com/wiki/CreateBuilding) doesn't exist 2 weeks after merge, please add it

PR #3152, opened 2023-08-11
Add event onElementHealthChange 

znvjder:
This PR adds the serverside event **onElementHealthChange**

I think this is an insignificant change for many people, but it will certainly make life easier for ours scripters.

Refers to #3136

PR #3278, opened 2023-12-29
Add onElementHealthChange & onClientElementHealthChange events

1Lorenzo-1:
## Features:

- **Server-Side Event:** `onElementHealthChange`
  - Triggered when the health of an element changes on the server side.
  - **Parameters:**
    - `source`: The affected element.
    - `arg1`: Old health value.
    - `arg2`: New health value.
    - `client` (optional): The client responsible for the change (if initiated from the client side).

- **Client-Side Event:** `onClientElementHealthChange`
  - Triggered when the health of an element changes on the client side.
  - **Parameters:**
    - `source`: The affected element.
    - `arg1`: Old health value.
    - `arg2`: New health value.

## Example Usage:

```lua
-- Server-Side
addEventHandler("onElementHealthChange", root,
    function (oldHealth, newHealth)
        if (client and client~=source and newHealth > 100) then 
            kickPlayer(client, "AC", "CHEATER")
        end 
    end
)

-- Client-Side
addEventHandler("onClientElementHealthChange", root,
    function (oldHealth, newHealth)
        if source==localPlayer then 
            iprint("your new/health: " .. newHealth)
        end 
    end
)
```


TracerDS:
#3152 duplicate

1Lorenzo-1:
> #3152 duplicate

I should have looked up before starting it..

znvjder:
Sure, it might be a duplicate, but I can see that the health status is updated through the packet - that's actually a good idea. My PR has been abandoned... I don't think I even have local access to it (it's actually on the cloud and rotting).

znvjder:
I have run some test cases. The event is only triggered through function setElementHealth. Generally, we would like the event to be triggered in any case of health change.

1Lorenzo-1:
> I have run some test cases. The event is only triggered through function setElementHealth. Generally, we would like the event to be triggered in any case of health change.

Tbh we got already (on[client]PlayerDamge)  on[client]VehicleDamge,  

So I dont think adding it would be nice 
Then we should remove both others events?

znvjder:
No, no... These events must stay. I'll allow myself to quote Lpsd comment from my PR.

> As Pirulax mentioned, I think it would be nice for onElementHealthChange to trigger whenever element health changes, not just via setElementHealth, but also when changed in-game (from fall damage, explosion, etc). There should be a parameter for the event to indicate whether it was changed via script.

1Lorenzo-1:
> No, no... These events must stay. I'll allow myself to quote Lpsd comment from my PR.
> 
> > As Pirulax mentioned, I think it would be nice for onElementHealthChange to trigger whenever element health changes, not just via setElementHealth, but also when changed in-game (from fall damage, explosion, etc). There should be a parameter for the event to indicate whether it was changed via script.

It's such a nice thing.. but why?
i do still think we don't need it.. but i can do it

ghost:
The events you've mentioned are meant only for streamed players/vehicles. Why exactly it should be as znvjder explained above? To avoid shenanigans of using extra logic, from event you'd expect to cover any health change, in any case. Otherwise it will be very inconvenient for scripter. Removing them doesn't make sense, as they have different purpose - either tweak how it works, disable damage or completely overhaul it.

1Lorenzo-1:
> The events you've mentioned are meant only for streamed players/vehicles. Why exactly it should be as znvjder explained above? To avoid shenanigans of using extra logic, from event you'd expect to cover any health change, in any case. Otherwise it will be very inconvenient for scripter. Removing them doesn't make sense, as they have different purpose - either tweak how it works, disable damage or completely overhaul it.

well i'll start rn with it doing what you want

1Lorenzo-1:
> The events you've mentioned are meant only for streamed players/vehicles. Why exactly it should be as znvjder explained above? To avoid shenanigans of using extra logic, from event you'd expect to cover any health change, in any case. Otherwise it will be very inconvenient for scripter. Removing them doesn't make sense, as they have different purpose - either tweak how it works, disable damage or completely overhaul it.

I believe it will only be server-side events. Otherwise, client-side damage is challenging to work with. I have tried everything in `Client/game_sa/CPedSA.cpp`, but nothing worked. Btw I'll try my best to make it. If someone has any idea how it could be done, that's going to help me much.

Pirulax:
You could do a check each frame perhaps.
It's impossible to track every instance of health change down, so that's your best bet.
Yes, it's going to be a frame late but oh well....

ghost:
Any updates on this?

1Lorenzo-1:
> Any updates on this?

No, somehow I can't do it... I've tried it in various ways, but there are so many FPS drops when a PLAYER got health changed. Maybe I'll close it, hoping that next time someone with more experience will make better one. 👍

botder:
Great amount of time passed, I am closing this then.

PR #3281, opened 2024-01-04
Fix reversed disable_backface_culling flag

### LINKED ISSUE #3279: Setting flag disable_backface_culling doesn't work
### Describe the bug

using `engineSetModelFlag` on this flag makes no difference.
engineGetModelFlag works fine.

### Steps to reproduce

Example location: 2305.15, -1089.17, 48.12

Changing through LAhills.ide works fine:
`13743, CEmulwire01, mulveg2lahills, 200, 2097284`
`13743, CEmulwire01, mulveg2lahills, 200, 132`

But this makes no difference
`engineSetModelFlag(13743, "disable_backface_culling", false)`
`engineSetModelFlag(13743, "disable_backface_culling", true)`

Other flags seem to work fine from my testing

### Version

v1.6-release-22345

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #3279
This PR can break some double-sided models, when engeineSetModelFlags is used

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -518,10 +518,10 @@ void CModelInfoSA::SetIdeFlags(unsigned int uiFlags)
     tIdeFlags ideFlags;
     ideFlags.uiFlags = uiFlags;
 
-    // Default value is 0xC0 (bIsColLoaded + bIsBackfaceCulled)
+    // Default value is 0xC0 (bIsColLoaded + bIsBackfaceCullingDisabled)
     // But bIsColLoaded should not be changed
     m_pInterface->usFlags &= 0x80;            // Reset all flags except bIsColLoaded
-    m_pInterface->bIsBackfaceCulled = true;
+    m_pInterface->bIsBackfaceCullingDisabled = true;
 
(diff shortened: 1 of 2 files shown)


Zangomangu:
The flag actually isn't reversed - with this PR we would need to call it enable_backface_culling?

> The game applies this flag before object spawning. 

This is why i couldn't get it to work. Is this the same for all the flags? I'll add a note on the wiki
So if you need to change a flag for an existing world model you need to remove and re-create the instances

TheNormalnij:
> Is this the same for all the flags

No. Most of them work immediately. 

> The flag actually isn't reversed - with this PR we would need to call it enable_backface_culling?

Community know this flag as [DISABLE_BACKFACE_CULLING](https://gtamods.com/wiki/Item_Definition#GTA_San_Andreas). I think we should provide same API. 
This flag confuses me. Does it have expected result for objects? Is it reversed?

Zangomangu:
> This flag confuses me. Does it have expected result for objects? Is it reversed?

Yes it has expected result and it matches with the original flag name
When you set the flag the model is drawn on both sides (culling is **disabled**)
I will close the issue and add a note on wiki

PR #3284, opened 2024-01-05
Fix `SetStreamingBufferSize`possibly  accessing memory out-of-bounds

Pirulax:
`numBlocks` was rounded upwards after the allocation was done, this could've caused OOB reads/writes.

### DIFF
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -431,6 +431,8 @@ void CStreamingSA::RemoveArchive(unsigned char ucArchiveID)
 
 bool CStreamingSA::SetStreamingBufferSize(uint32 numBlocks)
 {
+    numBlocks &= ~1; // Make it be even [Otherwise it can't be split in half properly]
+    
     // Check if the size is the same already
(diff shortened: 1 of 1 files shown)


Client/game_sa/CStreamingSA.cpp
```diff
@@ -431,6 +431,8 @@ void CStreamingSA::RemoveArchive(unsigned char ucArchiveID)
 
 bool CStreamingSA::SetStreamingBufferSize(uint32 numBlocks)
 {
+    numBlocks &= ~1; // Make it be even [Otherwise it can't be split in half properly]
+
```
TheNormalnij: Does it make less blocks?

TheNormalnij: Yes. if `numBlocks` is 41 it allocates only 40 (42 is expected)

Pirulax:
```suggestion
    numBlocks += numBlocks % 2; // Make sure number is even by "rounding" it upwards. [Otherwise it can't be split in half properly]
    
```

TheNormalnij (approved):

Issue #3286, opened 2024-01-09
Protected Element Data


TriangleToo:
### Is your feature request related to a problem? Please describe.

Nowadays, when Lua injectors have become commonplace, and considering how a large portion of MTA servers use Element Data without knowing its security weaknesses/not being able to secure them, I think it would be a good idea to make this small - but facilitating change to help secure such servers.

### Describe the solution you'd like

My suggestion is to add an additional parameter to the `setElementData` function - `protected=false/true` (default: `false`). If the parameter is set to `true` then all changes to the given element data from clientside will be ignored and printed in the debugscript. Eg. by rejecting network packets.

`bool setElementData ( element theElement, string key, var value [, var syncMode="broadcast", var protected=false] )`

### Describe alternatives you've considered

For the time being, the only solution if you don't want to give up Element Data (and with the introduction of subscriptions, it's a very good feature for optimally storing very little data) is to use the `onElementDataChange` event and undo the changes and then kick/ban the invasive user. Unfortunately, this solution has a significant drawback - this event is triggered already AFTER the element data change (which could have already performed some actions, for example), so we are forced to undo it. 

### Additional context

In my opinion, this is a fairly simple and quick solution to help secure an important element datas on MTA servers that should not be modified from the client side. In addition, we can send such a flag to the client, so that if you try to send such a change, no network traffic is sent to the server (which can try to overload it) - I realize that everything can be bypassed by cheaters and such traffic can still go out (so the the server side must verify such a change by the protected flag), but it will already be much more difficult for them.

I realize that servers can secure the current Element Data system with the above-mentioned event or by switching to their own data system sent by events, but I think that such a change, because of its simplicity, will positively affect the security of smaller servers or servers created by less experienced scripters who, for some reason, do not want to use the above-mentioned solutions. I think that everyone is also familiar with the performance of the `onElementDataChange` event, so the introduction of this feature will also increase the performance of servers, because they will be able to avoid using this event too intensively by cheaters.

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

PlatinMTA:
Maybe make ``onElementDataChange`` work with ``cancelEvent()`` instead?
The idea is good nonetheless, but maybe it's just too convoluted. ElementData was always regarded as unsafe and also slow, and is useful only in certain scenarios.

People that are already miss using elementdatas will not use ``protected = true`` btw.

TriangleToo:
[...]

PR #3287, opened 2024-01-10
Element data optimizations

tederis:
This PR is intended to improve the element data performance. The current implementation is far too inefficient because of a lot of string copying and the scourge of **strlen** invocations. There is also an excessive amount of access to **std::map** which makes it worse. Prior to the PR each **setElementData** call was producing at least 5 string copies on client and at least 7 copies(and a multiple of _player count_ in worst case when using an element data subscription) on server side. There were also 14 **strlen** invocations(each has O(n) complexity). 

Among the other improvements an additional ElementData's value check(newValue != oldValue) in method **Packet_CustomData** which can protect a server from **PACKET_ID_CUSTOM_DATA** packets spamming and the same for **CElementRPCs::SetElementData** method which also can protect a client from erroneous changes.

Simple(preliminary) measurements(local set/getElementData) show the performance gain nearly 2(in some cases 5, depends on an element data value) times.


tederis:
Fun fact: **80%** of the `setElementData` execution time is `on(Client)ElementDataChange` event processing with just one simple event handler.

TriangleToo:
> Fun fact: **80%** of the `setElementData` execution time is `on(Client)ElementDataChange` event processing with just one simple event handler.

lol. I would understand it for an event that can be canceled, but not so much for an event that is not cancellable (like this one). Is there any way to optimize this? Just wondering..

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -295,22 +293,22 @@ CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData
     // If none, try returning parent's custom data
     if (bInheritData && m_pParent)
     {
-        return m_pParent->GetCustomData(szName, true, pbIsSynced);
+        return m_pParent->GetCustomData(strName, true, pbIsSynced);
     }
 
     // None available
-    return NULL;
+    return {};
```
TheNormalnij, line 300: Isn't `nullptr` better in such cases?

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -279,12 +279,10 @@ void CClientEntity::SetID(ElementID ID)
     }
 }
 
-CLuaArgument* CClientEntity::GetCustomData(const char* szName, bool bInheritData, bool* pbIsSynced)
+CLuaArgument* CClientEntity::GetCustomData(const SString& strName, bool bInheritData, bool* pbIsSynced)
```
TheNormalnij: We are truing to not use SString in new code. [Discussion](https://discord.com/channels/278474088903606273/366384007535001612/796592491700158494)

Client/mods/deathmatch/logic/CClientEntity.cpp
tederis: {} is more universal and reliably. Just allow the compiler to deduce it, it knows better. This is also come in handy when a function's return argument is changed(say, from a pointer to bool). Then you don't need find and fix all places where the function returns. This is a matter of a code style of course, but {} is preferable in generic C++.

Client/mods/deathmatch/logic/CClientEntity.cpp
tederis: Yes, I know. But in this partucular casу `SString` was chosen for the reason that it was already used in almost all associative containers. By using lets say `std::string` you would get an additional string copy (see `SString(const std::string& strText`) constructor) and potential problems with key mathching. So the usage of std::string would exceed the boundaries of this PR.

tederis:
> > Fun fact: **80%** of the `setElementData` execution time is `on(Client)ElementDataChange` event processing with just one simple event handler.
> 
> lol. I would understand it for an event that can be canceled, but not so much for an event that is not cancellable (like this one). Is there any way to optimize this? Just wondering..

Yes, events can be slightly optimized. But to get a noticeable performance gain there would need a lot of work(probably, a full reengineering of the event system).

tederis:
I think function `setElementDataSilence(element, key, silent)`(suggest the better name) can be added for blocking element data events. Because in most of the cases those events are not needed but they still consume a lot of CPU.

tederis:
Apart from Lua events (see notes about their performance above), the element data is as fast as possible now.

tederis:
In theory `google::dense_hash_map` may give even better results but I'm afraid it will consume too much memory

Allerek:
> I think function `setElementDataSilence(element, key, silent)`(suggest the better name) can be added for blocking element data events. Because in most of the cases those events are not needed but they still consume a lot of CPU.

better to just add argument to setElementData that tells if it should trigger an event

Allerek:
> In theory `google::dense_hash_map` may give even better results but I'm afraid it will consume too much memory

These days we have a lot of memory in our computers. 16GB is the minimum for new pcs in few last years, 8GB is still common. I think people can handle that really. Lets not be so paranoic. Main problem of element-datas are their CPU usage.

tederis:
> > I think function `setElementDataSilence(element, key, silent)`(suggest the better name) can be added for blocking element data events. Because in most of the cases those events are not needed but they still consume a lot of CPU.
> 
> better to just add argument to setElementData that tells if it should trigger an event

This brings up a question: what should we do when a client sends data to a server? Send this flag as well and dictate whether server should invoke an event? It can produce the following circle: client -> server -> other clients. One of the clients may decide to silently change the data without even possibility for detecting it at endpoints. Or we shouldn't allow clients do that?

ghost:
> These days we have a lot of memory in our computers. 16GB is the minimum for new pcs in few last years, 8GB is still common. I think people can handle that really. Lets not be so paranoic. Main problem of element-datas are their CPU usage.

But you are aware that MTA/GTA is only 32 bit?

TriangleToo:
> > > I think function `setElementDataSilence(element, key, silent)`(suggest the better name) can be added for blocking element data events. Because in most of the cases those events are not needed but they still consume a lot of CPU.
> > 
> > 
> > better to just add argument to setElementData that tells if it should trigger an event
> 
> This brings up a question: what should we do when a client sends data to a server? Send this flag as well and dictate whether server should invoke an event? It can produce the following circle: client -> server -> other clients. One of the clients may decide to silently change the data without even possibility for detecting it at endpoints. Or we shouldn't allow clients do that?

IMO this should definitely be an additional argument to `setElementData` instead of another function. It makes no sense to have it as a separate function (at least none that I can think of). The client should have this argument ignored when syncMode is set to true (synced with server) and same on the server logic - it should ignore such flag if sent from client. 

> One of the clients may decide to silently change the data without even possibility for detecting it at endpoints

This can be also solved by #3286 😎

Client/mods/deathmatch/logic/CClientEntity.cpp
TheNormalnij: All your arguments are good for template/library code. 
> {} is more universal and reliably.

Specific instruments are always better and reliably. There is no reasons to make some parts of code more universal than they need to be.
> Just allow the compiler to deduce it, it knows better.

We write code for humans not the compiler. And we optimize code for programmers first. There is only one possible way to deduce return value in such cases. Does the compiler have some secrets? Or is there a reason to hide the value type from developers?

> This is also come in handy when a function's return argument is changed(say, from a pointer to bool). Then you don't need find and fix all places where the function returns.

This doesn't ever look like a significant time save. Developers waste more time when they read code. Additional context makes code easier to read.

> This is a matter of a code style of course, but {} is preferable in generic C++.

I checked UE5 and GTA5 code. UE5 uses `return {}` a lot ever for ints. GTA5 never uses this. [Stack overflow](https://stackoverflow.com/questions/39487065/what-does-return-statement-mean-in-c11) users prefer to use {} only for structs with a default initializer.

> This is a matter of a code style of course

It looks more like this. Do we want allow {} initialization for primitives in our code style?

Pirulax:
`newValue != oldValue` doesn't work all that well.
Checking whenever 2 `LuaArguments` are the same isn't trivial, and the implementation is currently very shallow.
So keep that in mind.

Client/mods/deathmatch/logic/CClientEntity.cpp
Pirulax: I feel like this is really just nitpicking. It isn't a major change in the style, so it's perfectly fine imo.

Client/mods/deathmatch/logic/CClientEntity.cpp
Pirulax: Why not use `std::string_view` all the way, and only in the end (when inserting) create the `SString`?
This way for the getter you can get away without creating a string at all.
Also, since the code limits the string length to 128, you could use some kind of a fixed-length string and store the data directly inside it (So Caching works better).

TheNormalnij: std::string_view uses 8 bytes.
Pointers and (I believe) references use 4 bytes.
So function use less registers. And the code can be optimized better.

Pirulax: I think not allocating any memory until absolutely necessary far outweighs using 2 registers instead of 1.......

TheNormalnij: Can you be more specific, please?
When we use 'getElementData' or 'setElementData', string is already created. 
Maybe, only 'CClientCClientEntity::SetCustomData' can be overloaded with 'std::string_view' key argument if we allow using data directly from network packets.
Where do you suggest using caching?

Pirulax: The string is created within Lua, yes, so you can use a `string_view`, no need for `std::string` (Or `SString`).
Eventually when storing the key in the `std::map` the string has to be allocated.
But for getters `std::map` supports transparent lookup, so you can pass in `std::string_view` even if the key type is `std::string`.
(You just have to pass in `std::less<>` as the comparator instead of using the default one)

TheNormalnij: Yeah. It seems `std::string_view` works without making a string copy for function arguments, but still makes a copy for output (this is expected). MTA has a very few usages of `std::string_view` in parser. So it wasn't expected, that this thing works.
getElementData can be optimized 100%.
Maybe `const char*` will be ever faster in this case.

Pirulax: the argument parser correctly handles `std::string_view`.
`const char*` wont be faster, because it has to calculate the length, just use `string_view` that takes the length from Lua....
Why are you so against `string_view` I don't understand.

tederis: Okay, I'll take into account your suggestions. std::string_view should fit in.

TheNormalnij: > Why are you so against `string_view` I don't understand.

Im not totally against `string_view`. If you can pass simpler value, you should use it. Basic string compare function doesn't require string length. It's similar to removing unused data from function arguments.

tederis: > > Why are you so against `string_view` I don't understand.
> 
> Im not totally against `string_view`. If you can pass simpler value, you should use it. Basic string compare function doesn't require string length. It's similar to removing unused data from function arguments.

`std::hash` struct requires string len. That is `std::unordered_map` will have to construct a new string from the characters buffer(const char*) which is obviously not the best behaviour. `std::string_view` does not have this drawback.

TheNormalnij: > `std::hash` function requires string len.

This is the answer. Thx

Pirulax: > > > Why are you so against `string_view` I don't understand.
> > 
> > 
> > Im not totally against `string_view`. If you can pass simpler value, you should use it. Basic string compare function doesn't require string length. It's similar to removing unused data from function arguments.
> 
> `std::hash` struct requires string len. That is `std::unordered_map` will have to construct a new string from the characters buffer(const char*) which is obviously not the best behaviour. `std::string_view` does not have this drawback.

If you use std::map you can use take advantage of transparent lookup.
Sadly unordered_map doesn't support this.
Another way of doing this would be to use a fixed length string buffer as the key to avoid heap allocation (and by using an `unodered_map` we could increase cache locality)

tederis: > > > > Why are you so against `string_view` I don't understand.
> > > 
> > > 
> > > Im not totally against `string_view`. If you can pass simpler value, you should use it. Basic string compare function doesn't require string length. It's similar to removing unused data from function arguments.
> > 
> > 
> > `std::hash` struct requires string len. That is `std::unordered_map` will have to construct a new string from the characters buffer(const char*) which is obviously not the best behaviour. `std::string_view` does not have this drawback.
> 
> If you use std::map you can use take advantage of transparent lookup. Sadly unordered_map doesn't support this. Another way of doing this would be to use a fixed length string buffer as the key to avoid heap allocation (and by using an `unodered_map` we could increase cache locality)

Heterogeneous lookups for unordered maps should be available since C++20. Regretfully we're still with C++17.

As you must be aware we are using both (unordered hash and tree-based maps) in MTA. Thus the solution must be selected according to the specifics of the both. Fixed length strings cannot be efficient here due to the unpredictable length of such strings.

Pirulax: > As you must be aware we are using both (unordered hash and tree-based maps) in MTA. Thus the solution must be selected according to the specifics of the both. Fixed length strings cannot be efficient here due to the unpredictable length of such strings.

See `MAX_CUSTOMDATA_NAME_LENGTH`.
So no name can be longer than 128 chars, so a 128 + 1 byte buffer would be enough.
A cache line is typically 64B, so keep that in mind. (But I think 128 + 1 for the char buffer, and uint8 for the length should be enough, this way the whole struct would be 132 bytes (I think, or 136 on 64b).
I'm unsure why you think fixed length strings would be inefficient. Tree based maps have extreme memory overhead for small values (std::string + CLuaArgument is barely a few bytes).
[Also, memory footprint can be lowered by merging my CLuaArgument refactor that uses `std::variant` (Thus the final size is quite a bit less)]

> Heterogeneous lookups for unordered maps should be available since C++20. Regretfully we're still with C++17

So we can upgrade to c++20 (For the server/client dm project at least).

Pirulax: Also, we could experiment with custom hash functions (that generate less collisions for our use-case), if we decide to use a hashmap.

tederis: > See MAX_CUSTOMDATA_NAME_LENGTH.
So no name can be longer than 128 chars, so a 128 + 1 byte buffer would be enough.
A cache line is typically 64B, so keep that in mind. (But I think 128 + 1 for the char buffer, and uint8 for the length should be enough, this way the whole struct would be 132 bytes (I think, or 136 on 64b).
I'm unsure why you think fixed length strings would be inefficient. Tree based maps have extreme memory overhead for small values (std::string + CLuaArgument is barely a few bytes).
[Also, memory footprint can be lowered by merging my CLuaArgument refactor that uses std::variant (Thus the final size is quite a bit less)]

128 bytes is still a lot actually. On large servers such keys can eat up to 10 megabytes(or even more) of the ram. You might say that 10 mb does not deserve attention, but on client side it may be significant (especially on high-load custom maps).

Personally I would prefer the StringName(as like in Godot Engine, or FName in UE) approach that implies just one buffer for each unique string. This way we only need a unique number that can be used as a key within the maps.

Pirulax: That is an awesome idea, never thought of that, you're a genius!
So we could both save RAM, and CPU!
Just have to have a string interning system (That is global to all elements).
We also need to use some kind of reference counting (`shared_ptr` perhaps? Though the control block is allocated separately, that's kinda a waste of memory in our case) - To free unused strings.

Not sure how good of a key the serial index of the string would be, in terms of load factor (we want to avoid list traversals of course).

Not sure how efficient the same interning would be for `CLuaArgument`.

Though, I do have a concern, we currently use case-sensitive strings, right? `FName`'s are case-insensitive (which makes sense, less variation)

tederis:
Although this PR is fully functional I'm forced to admit that most of the new suggestions cannot be addressed without the C++20 support (see the discussion in comments above). A brief attempt to enable C++20 shows that a lot of things(even the singular **deathmatch** module) depend on previous standards. So there is many things that should be resolved before. I hope that work will be done in the foreseeable future.

Einheit-101:
Am i correct that we can save a lot of performance if we just delete all "onClientElementDataChange" event handlers in all running scripts? I use this event a lot in my old WW2 server because its very handy but the performance is terrible, even on a Ryzen 5900X.

tederis:
> Am i correct that we can save a lot of performance if we just delete all "onClientElementDataChange" event handlers in all running scripts? I use this event a lot in my old WW2 server because its very handy but the performance is terrible, even on a Ryzen 5900X.

Events are the heavier part of the element data. Thus, we can get a significant performance gain by disabling them. My original idea was to add setElementDataSilence(element, key, silent) function for this. If you're asking if you can get that gain right now, then yes. The less event handles you have the less they impact the performance.

Dutchman101:
Are there any updates or foreseeable plans to continue this? Element data optimizations would be huge for MTA.
@tederis

tederis:
> Are there any updates or foreseeable plans to continue this? Element data optimizations would be huge for MTA. @tederis

As was already mentioned above, most of the suggestions cannot be addressed without the C++20 support. For example, one of the suggestions was using std::string instead of SString, but this would lead to the performance degradation without C++20. But there could be some workarounds that I'm going to consider. Perhaps, it would require an intermediate PR (for the string interning implementation). I'm also interested in the element data optimization, so I'm going to finish it this month.

Fernando-A-Rocha:
@tederis 
I support the addition of:

`setElementDataSilent( element theElement, bool status [, string key ] )`

This would allow the developer to disable `on(Client)ElementDataChange` event triggers for certain elements & their children, as well as specific keys.

Inspired by #3839

Dutchman101:
> As was already mentioned above, most of the suggestions cannot be addressed without the C++20 support. For example, one of the suggestions was using std::string instead of SString, but this would lead to the performance degradation without C++20. But there could be some workarounds that I'm going to consider. Perhaps, it would require an intermediate PR (for the string interning implementation).

@tederis 

Alright, that sounds like a good summary after looking at all passed and pending code reviews.
Let's merge this, so you can work on a follow-up PR for C++ 20 readiness and final implementations next, while most of the changes (that got in from this PR) can be stress tested during the upcoming mass-testing phase that will be organized very soon to prepare for the release of MTA 1.6.1.

Dutchman101:
Please resolve conflicts, so we can go ahead with that course of action.

tederis:
This PR will be reimplemented and reopened soon.

PR #3291, opened 2024-01-14
Trigger onClientPlayerWeaponSwitch Event for Remote Players (fix #1141)

### LINKED ISSUE #1141: onClientPlayerWeaponSwitch should trigger for remote players

znvjder:
This PR is a follow-up of https://github.com/multitheftauto/mtasa-blue/pull/3238 and adds Bitstream handling. It's also an iteration of, and a replacement for, a few older PR's that were intended to do this, e.g #2181 and #1723.

It is to fix issue #1141 

Tested on debug and release project.
a13221


Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -178,6 +179,28 @@ void CRPCFunctions::PlayerWeapon(NetBitStreamInterface& bitStream)
...
+
+            if (!nearList.empty())
+            {
+                static std::vector<CPlayer*> playersInNearList;
+                playersInNearList.reserve(nearList.size());
+                playersInNearList.clear();
+
+                for (const auto& player : nearList)
+                    if (player.first->CanBitStream(eBitStreamVersion::OnPlayerWeaponSwitch_Remote))
+                        playersInNearList.push_back(player.first);
```
botder: Can you use brackets **{ }** for multi-line *blocks*?

TracerDS (commented):
early return would be better here

Client/mods/deathmatch/logic/rpc/CPlayerRPCs.cpp
```diff
@@ -168,3 +169,22 @@ void CPlayerRPCs::TakePlayerScreenShot(NetBitStreamInterface& bitStream)
...
+    {
+        CClientPlayer* pPlayer = m_pPlayerManager->Get(pSource->GetID());
+        if (IS_REMOTE_PLAYER(pPlayer))
+        {
+            CLuaArguments Arguments;
+            Arguments.PushNumber(lastSlot);
+            Arguments.PushNumber(currentSlot);
+            pPlayer->CallEvent("onClientPlayerWeaponSwitch", Arguments, true);
+        }
+    }
```
TracerDS: Early return would be better here

Server/mods/deathmatch/logic/CRPCFunctions.cpp
znvjder: Yes, done.

Client/mods/deathmatch/logic/rpc/CPlayerRPCs.cpp
znvjder: Sure, that's a good idea. Done.

Dutchman101:
I rewrote OP like this, to clarify:

This PR is a follow-up of https://github.com/multitheftauto/mtasa-blue/pull/3238 and adds Bitstream handling. It's also an iteration of, and a replacement for, a few older PR's that were intended to do this, e.g #2181 and #1723.

It is to fix issue #1141

Dutchman101:
> Tested on debug and release project

@znvjder The [same bug](https://github.com/multitheftauto/mtasa-blue/pull/3238#issuecomment-1879953715) showed up in nightly, sorry but had to revert (032dbd3) it again. It doesn't affect all servers, but you can reproduce easily on many, like FFS-Gaming > join derby.. you won't see the world

Example of how it looks there:
[https://i.imgur.com/2yNuYns.png](https://i.imgur.com/2yNuYns.png)

Dutchman101:
> The [same bug](https://github.com/multitheftauto/mtasa-blue/pull/3238#issuecomment-1879953715) showed up in nightly, sorry but had to revert ([032dbd3](https://github.com/multitheftauto/mtasa-blue/commit/032dbd310a6dae53303ce730a3c4080a2961dc80)) it again. 

Update.. it looks like caused by not putting the enum at end of list.
The mistake:
Untitled


https://github.com/multitheftauto/mtasa-blue/blob/bea48237d137ed4a51d8f6bac1dec88edddb512a/Shared/sdk/net/rpc_enums.h#L17

It's in the documentation as ^ but somehow the net bugs only appear in official MTA, not custom/debug builds. Leading to testing complications and spotting the bugs only once a nightly has been built with merge (if documentation was missed). Clearly, no amount of code reviews spotted the wrong place that the enum was inserted, so code reviewers aren't really aware yet, there should be a self-test/assert for this i guess.

Exactly the same type of bug just got fixed with another PR (#3550) through correcting the order: #3571
We'll just re-introduce your changes soon, @znvjder

Issue #2002, opened 2021-01-14
Base32 support


Inder00:
**Describe the solution you'd like**
Add `base32` support programically.
Base32 is used by two-factor-authentication (ex. TOTP) to generating secret tokens.

Shared functions:
- `string base32Encode( string value )`
- `string base32Decode( string value )`,
- `string encodeString( string "base32", string value )`
- `string decodeString( string "base32", string value )`

Pirulax:
I think we should add a `baseEncode/Decode` function instead.
The syntax would look like this:
```cpp
baseEncode(string value, int radix = 64)
```
Radix is the base, so 32, 64, 128, etc...

Nico8340:
#3312

ghost:
This issue can be closed, as #3312 was merged.

PR #3312, opened 2024-02-13
Improved Base Encoding & Decoding

### LINKED ISSUE #3311: Improved Base Encoding & Decoding

_No response_

Integration into encodeString and decodeString to be available asynchronously, support for base64 and base32, also for other base variants.

_No response_

### Additional context

_No response_

### Security Policy

ghost:
Adds base encoding and decoding to the existing encodeString and decodeString functions, supports base64, base32, and url and hex variants _asynchronously_, deprecates the old base64Encode and base64Decode functions (The code remains)
Closes #3311 and #2002

Syntax:
```lua
encodeString("base64", "myString")
encodeString("base64", "myString", {variant = "URL"})
encodeString("base32", "myString")
encodeString("base32", "myString", {variant = "Hex"})
```

Syntax (asynchronous)
```lua
encodeString("base32", "myString", {variant = "Hex"},
	function(callbackString)
		iprint(callbackString)
	end
)
```


ghost:
If this Pull Request gets merged, I will update the wiki pages.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -589,6 +596,150 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
             }
+            case StringEncodeFunction::BASE64:
+            {
+                SString& variant = options["variant"];
+
+                if (!variant.empty())
+                {
+                    variant = variant.ToUpper();
+
+                    if (variant != "URL")
```
tederis: Can be compacted down to 
```
const SString variant = options["variant"].ToUpper();
if (!variant.empty() && variant != "URL")
```
The same applies to the lines below.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -402,10 +402,17 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
     CScriptArgReader argStream(luaVM);
     argStream.ReadEnumString(algorithm);
     argStream.ReadString(data);
-    argStream.ReadStringMap(options);
 
-    argStream.ReadFunction(luaFunctionRef, LUA_REFNIL);
-    argStream.ReadFunctionComplete();
+    if ((algorithm != StringEncodeFunction::BASE64 && algorithm != StringEncodeFunction::BASE32) || argStream.NextIsTable())
```
tederis, line 406: This makes `base64` and `base32` algorithms syntactically inconsistent with the older ones. For example, you can omit the `options` argument for new algorithms, while must preserve it for, let's say, `tea`. This is confusing.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -402,10 +402,17 @@ int CLuaCryptDefs::EncodeString(lua_State* luaVM)
...
+    {
+        argStream.ReadStringMap(options);
+    }
+
+    if ((algorithm != StringEncodeFunction::BASE64 && algorithm != StringEncodeFunction::BASE32) || argStream.NextIsFunction())
```
tederis: This forces the use of async `function` for the existing algorithms while omit the requirement for the new ones. This may also confuse and break the backward compatibility.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -614,10 +765,17 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
     CScriptArgReader argStream(luaVM);
     argStream.ReadEnumString(algorithm);
     argStream.ReadString(data);
-    argStream.ReadStringMap(options);
 
-    argStream.ReadFunction(luaFunctionRef, LUA_REFNIL);
-    argStream.ReadFunctionComplete();
+    if ((algorithm != StringEncodeFunction::BASE64 && algorithm != StringEncodeFunction::BASE32) || argStream.NextIsTable())
```
tederis, line 756: This makes `base64` and `base32` algorithms syntactically inconsistent with the older ones. For example, you can omit the `options` argument for new algorithms, while must preserve it for, let's say, `tea`. This is confusing.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
```diff
@@ -614,10 +765,17 @@ int CLuaCryptDefs::DecodeString(lua_State* luaVM)
...
+    {
+        argStream.ReadStringMap(options);
+    }
+
+    if ((algorithm != StringEncodeFunction::BASE64 && algorithm != StringEncodeFunction::BASE32) || argStream.NextIsFunction())
```
tederis: This forces the use of async `function` for the existing algorithms while omit the requirement for the new ones. This may also confuse and break the backward compatibility.

tederis: Although, this condition makes no sense because of the `LUA_REFNIL` argument.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
tederis: Although, this condition makes no sense because of the `LUA_REFNIL` argument.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: Good suggestion, done.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: Thank you, done.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: Thank you, done.

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: There was some conversation on Discord about this, @botder said its fine.
https://discord.com/channels/801330706252038164/801411628600000522/1206737515940155534
https://discord.com/channels/801330706252038164/801411628600000522/1206757450162774079

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: There was some conversation on Discord about this, botder said its fine.
https://discord.com/channels/801330706252038164/801411628600000522/1206737515940155534
https://discord.com/channels/801330706252038164/801411628600000522/1206757450162774079

Shared/mods/deathmatch/logic/luadefs/CLuaCryptDefs.cpp
ghost: What do you think, it is acceptable in this case? Or do you have any better ideas?

tederis: Can be marked as solved, according to the last conversation in Discord. https://discord.com/channels/801330706252038164/801411628600000522/1208271751214276639

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -25,18 +26,66 @@ namespace SharedUtil
...
+        {
+            CryptoPP::StringSource ss(data, true, new CryptoPP::Base64Encoder(new CryptoPP::StringSink(result), false));
+        }
+
+        return result;
+    }
+
+    inline SString Base64decode(const SString& data, const SString& variant = SString())
+    {
+        std::string result;
```
tederis: This breaks the NRVO optimization. A type of the returning value shall match the return type of a function. The same for functions below.

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -25,18 +26,66 @@ namespace SharedUtil
...
+        }
+
+        return result;
+    }
+
+    inline SString Base32encode(const SString& data, const SString& variant = SString())
+    {
+        std::string result;
+
+        if (variant == "HEX")
```
tederis, line 65: This is debatable but I would prefer a enum here.

tederis: I know that this will require a conversion from string to enum. I'm just pointing out the possible improvements. Due to a small string length the current implementation is not that bad actually.

Shared/sdk/SharedUtil.Crypto.h
ghost: Done.

Shared/sdk/SharedUtil.Crypto.h
ghost: Since base64 and base32 have separate variants (base64 - url, base32 - hex) and there are no others, they should be hardcoded if we don't want to create two separate enum classes, which doesn't make any sense, the variants wouldn't be expandable anyway, because they don't exist. If this is not enough reason to keep it the current way, please write a comment in the next review.

Shared/sdk/SharedUtil.Crypto.h
```diff
@@ -25,18 +26,66 @@ namespace SharedUtil
...
     };
 
-    inline SString Base64encode(const SString& data)
+    inline SString Base64encode(const SString& data, const SString& variant = SString())
     {
-        SString                result;
-        CryptoPP::StringSource ss(data, true, new CryptoPP::Base64Encoder(new CryptoPP::StringSink(result), false));            // Memory is freed automatically
+        SString result;
+
+        if (variant == "URL")
```
tederis, line 33: Okay, we came to the conclusion that a string representation is better then enum in this particular case. I would just suggest a tiny improvement: move `"HEX"` and `"URL"` literals into an unnamed namespace and make them `const std::string`. In this way, we will be able to eliminate calculation of string length every time.

ghost: To avoid breaking NRVO, they should remain SStrings, but I've moved them to a separate namespace as constants.

tederis: In the comparison statements the NRVO doesn't exist because it has no returning values(at least within the bounds of an expression), except bool. So there is no need to keep using SString. But SString is also okay, fine.

ghost: Okay, didn't know about that it only applies if the return is inside an expression.

ghost: Other arguments are SStrings too, i didn't wanna mix them tbh

tederis: Oh, I didn't notice that those functions are within a header file. In this case unnamed namespace is unacceptable. So you can either revert the last commit or create SharedUtil.Crypto.hpp and put the functions(along with the namespace) into it (you can see how the hpp's are included in SharedUtil.cpp). Once this is done I approve the PR.

ghost: > Oh, I didn't notice that those functions are within a header file. In this case unnamed namespace is unacceptable. So you can either reverse the last commit or create SharedUtil.Crypto.hpp and put the functions(along with the namespace) into it (you can see how the hpp's are included in SharedUtil.cpp). Once this is done I approve the PR.

Done.
@tederis

tederis (dismissed):
LGTM

Client/mods/deathmatch/logic/luadefs/CLuaCompatibilityDefs.cpp
```diff
@@ -71,6 +71,8 @@ void CLuaCompatibilityDefs::LoadFunctions()
         {"getControlState", CLuaPedDefs::GetPedControlState},
         {"setCameraShakeLevel", ArgumentParserWarn<false, CLuaCameraDefs::SetCameraDrunkLevel>},
         {"getCameraShakeLevel", ArgumentParserWarn<false, CLuaCameraDefs::GetCameraDrunkLevel>},
+        {"base64Encode", CLuaCryptDefs::EncodeString},
+        {"base64Decode", CLuaCryptDefs::DecodeString},
```
botder: What about `{"base64Encode", ArgumentParserWarn<false, Base64encode>},` in CLuaCryptDefs.cpp?

ghost: I deleted them from here because we talked on Discord about keeping the functions forever.

botder (approved):

PR #3321, opened 2024-02-21
Fix #3320: Incorrect colors for 3D draws

### LINKED ISSUE #3320: Incorrect colors for 3D draws
### Describe the bug

With any 3D DirectX drawing function, if we specify our exact color code using tocolor, but do not draw as postgui, we get an unexpected result: The color will be different from the specified color.

### Steps to reproduce

1. Test resource - [Download](https://www.file.io/A8Nt/download/yVb11KPRcxsc)
2. Test resource - Run
3. See the difference between the colors

### Version

Client: v1.6.0-release-22408 (Windows 11, 64-bit)
Server: v1.6.0-release-22408 (Windows 11, 64-bit)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

ghost:
Fixes #3320 (an old bug caused by an incomplete, unfinished piece of code)

### DIFF
--- a/Client/core/Graphics/CLine3DBatcher.cpp
+++ b/Client/core/Graphics/CLine3DBatcher.cpp
@@ -79,16 +79,8 @@ void CLine3DBatcher::Flush()
     for (std::vector<SLine3DItem>::iterator iter = m_LineList.begin(); iter != m_LineList.end(); ++iter)
     {
         const SLine3DItem& item = *iter;
-        SColor color = item.ulColor;
 
-        if (m_bPreGUI)
-        {
-            color.R /= 2;
-            color.G /= 2;
-            color.B /= 2;
-        }
-
-        const ulong    ulColor = color;
+        const ulong    ulColor = item.ulColor;
         const float    fWidth = item.fWidth;
         const CVector& vecA = item.vecFrom;
         const CVector& vecB = item.vecTo;
--- a/Client/core/Graphics/CLine3DBatcher.cpp
+++ b/Client/core/Graphics/CLine3DBatcher.cpp
@@ -19,9 +19,9 @@
 //
 //
 ////////////////////////////////////////////////////////////////
-CLine3DBatcher::CLine3DBatcher(bool bZTest)
+CLine3DBatcher::CLine3DBatcher(bool bPreGUI)
 {
-    m_bZTest = bZTest;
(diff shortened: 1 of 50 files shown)


tederis:
Although I assume it works, the origin of these coefficients isn't clear for me. Can you elaborate on how they were found?

ghost:
> Although I assume it works, the origin of these coefficients isn't clear for me. Can you elaborate on how they were found?

The bug only appears for 3D lines, not for 3D material lines, because the code was already there. The solution comes from the code of 3D material lines.
@tederis

tederis:
> > Although I assume it works, the origin of these coefficients isn't clear for me. Can you elaborate on how they were found?
> 
> The bug only appears for 3D lines, not for 3D material lines, because the code was already there. The solution comes from the code of 3D material lines. @tederis

I see now, you just copied the behaviour from `CMaterialLine3DBatcher`. But I still cannot understand how exactly it fixes the problem. Why should it be divided by two? This is not directed at you, I just want to comprehend.

ghost:
> I see now, you just copied the behaviour from `CMaterialLine3DBatcher`. But I still cannot understand how exactly it fixes the problem. Why should it be divided by two? This is not directed at you, I just want to comprehend.

This is a very good question, because if we draw as a postGUI, then this is not necessary. This is probably caused by a bug / behavior of DirectX.
@tederis

tederis:
After a little research I have found that the problem is indeed comes from the color correction pass (as I pointed out in #3320). This pass isn't fixed and it changes the colors every frame. So you can see that the division by 2 doesn't fix that, but only masks the problem. Moreovere, there is no constant that can fix it. 
Colors

That means that 3D material lines still experience the problem. I think that the division should be removed from CMaterialLine3DBatcher as it makes colors inconsistent with other dx functuin and makes it difficult to correct colors using getColorFilter function. Unfortunately this PR also doesn't fix the issue.

ghost:
I removed the division from both batchers.

Einheit-101:
Finally no more getColorFilter to fix the dx rendering? Nice.

ghost:
> Finally no more getColorFilter to fix the dx rendering? Nice.

This patch only removes divide by two from the line batcher, so it doesn't provide a real solution to the problem, it can be used uniformly and correctly with getColorFilter, but a future patch may include a real solution to the problem.

Issue #3305, opened 2024-01-30
Add new event: onResourceStateChange


TracerDS:
### Is your feature request related to a problem? Please describe.

Following #3304 there should be an event for tracking state changes of all resources. `onResourceLoadStateChange` only triggers for load changes (aka removing and creating a resource)

### Describe the solution you'd like

`onResourceStateChange` would trigger when a resource has been stopped, started or refreshed

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

R4ven47:
How is this event going to behave any different than onResourceStart/onResourceStop?

TracerDS:
> How is this event going to behave any different than onResourceStart/onResourceStop?

it would be much easier to communicate between one event handler and 2 separate event handlers.
With that function you will have more flexibility with managing your resources

PR #3325, opened 2024-02-29
Add event "OnResourceStateChange"

TracerDS:
#3305

Some might say this event is pointless. But it might be a good alternative for `onResourceLoadStateChange`.
It will also help resources that monitor changing resources. Connecting that event with tables and other data would be easier than tracking 3 separate events.


znvjder:
I'm in favor of accepting this, but removing the `onResourceLoadStateChange` event. It's entirely pointless and unnecessary.

TracerDS:
> I'm in favor of accepting this, but removing the `onResourceLoadStateChange` event. It's entirely pointless and unnecessary.

For backwards compatibility its better to leave it for now and deprecate it when this PR gets merged

ghost (approved):
LGTM :shipit:

Lpsd (approved):
LGTM. Please add the relevant wiki page once the nightly build has been created, following the "new feature" format ([example](https://wiki.multitheftauto.com/wiki/OnPlayerProjectileCreation))

PR #3330, opened 2024-03-02
Fix crashes caused by createBuilding with engineRequestModel

TheNormalnij:
Fixes issues:
* Destroyed building leaves grass
* Crash occurs when a resource that uses createBuilding with engineRequestModel is stopped.

### DIFF
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -416,8 +416,8 @@ void CPoolsSA::RemoveBuilding(CBuilding* pBuilding)
     coverList->RemoveItem(pInterface);
 
     // Remove plant
-    typedef CEntitySAInterface*(__cdecl * CPlantColEntEntry_Remove)(CEntitySAInterface * item);
-    ((CPlantColEntEntry_Remove)0x5DBEF0)(pInterface);
+    using CPlantColEntry_Remove = CEntitySAInterface* (*)(CEntitySAInterface*);
+    ((CPlantColEntry_Remove)0x5DBEF0)(pInterface);
 
     // Remove col reference
     auto modelInfo = pGame->GetModelInfo(pBuilding->GetModelIndex());
--- a/Client/game_sa/CPoolsSA.cpp
+++ b/Client/game_sa/CPoolsSA.cpp
@@ -26,6 +26,7 @@
 #include "CWorldSA.h"
 #include "CKeyGenSA.h"
 #include "CFileLoaderSA.h"
+#include "CPtrNodeSingleListSA.h"
 
 extern CGameSA* pGame;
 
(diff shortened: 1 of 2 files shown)


Client/game_sa/CPoolsSA.cpp
```diff
@@ -410,6 +411,14 @@ void CPoolsSA::RemoveBuilding(CBuilding* pBuilding)
     // Remove building from world
     pGame->GetWorld()->Remove(pInterface, CBuildingPool_Destructor);
 
+    // Remove building from cover list
+    CPtrNodeSingleListSAInterface<CBuildingSAInterface>* coverList = reinterpret_cast<CPtrNodeSingleListSAInterface<CBuildingSAInterface>*>(0xC1A2B8);
+    coverList->RemoveItem(pInterface);
+
+    // Remove plant
+    typedef CEntitySAInterface*(__cdecl * CPlantColEntEntry_Remove)(CEntitySAInterface * item);
```
TracerDS: Why not use `using` instead of `typedef`?

TheNormalnij: In this case, there is no difference. Do i miss something?

TracerDS: > In this case, there is no difference. Do i miss something?

not really a difference. Just that in C++ its better to use `using` than `typedef`

Client/game_sa/CPtrNodeSingleListSA.h
```diff
@@ -0,0 +1,45 @@
...
+    void RemoveAllItems();
+
+private:
+    CPtrNodeSingleLink<T>* m_pList;
+};
+
+template <class T>
+void CPtrNodeSingleListSAInterface<T>::RemoveItem(T* item)
+{
+    typedef void(__thiscall * CPtrNodeSingleList_RemoveItem_t)(CPtrNodeSingleListSAInterface<T> * pLinkList, void* item);
```
TracerDS: `using` instead of `typedef`

TracerDS (changes requested):

Client/game_sa/CPoolsSA.cpp
Lpsd: Would be better to provide actual details on the differences
> A typedef-name can also be introduced by an alias-declaration. The identifier following the using keyword becomes a typedef-name and the optional attribute-specifier-seq following the identifier appertains to that typedef-name. It has the same semantics as if it were introduced by the typedef specifier. In particular, it does not define a new type and it shall not appear in the type-id.

> The alias declaration is compatible with templates, whereas the C style typedef is not.

TracerDS: > Would be better to provide actual details on the differences
> 
> > A typedef-name can also be introduced by an alias-declaration. The identifier following the using keyword becomes a typedef-name and the optional attribute-specifier-seq following the identifier appertains to that typedef-name. It has the same semantics as if it were introduced by the typedef specifier. In particular, it does not define a new type and it shall not appear in the type-id.
> 
> > The alias declaration is compatible with templates, whereas the C style typedef is not.

`using` is the same as `typedef` with the exception of templates. Doesn't really matter. `typedef` is a C thing. We have `using` in C++ so let's start using `using`

Lpsd: Yeah, I didn't make an opinion either way, just providing the relevant details so that others can accurately come to a conclusion. It's better than simply saying "there's not really a difference" when there are actual differences.

Client/game_sa/CPtrNodeSingleListSA.h
TheNormalnij: resolved

ghost (approved):
LGTM

Fernando-A-Rocha:
ready to merge? 👀

TheNormalnij:
Yes, this fixes most common issues.

PR #3334, opened 2024-03-09
Syncable serverside Vehicles

### LINKED ISSUE #3333: Add a "sync" argument for vehicles, as is done for peds

In the case of peds, when we create static peds that only stand, we can disable their synchronization by setting the "sync" argument to false, which is good for performance.

I think this possibility would also be useful in the case of vehicles, when we create some static vehicles that just stand and do not require synchronization.

I know - You can create a vehicle on the client side, but a ped can also be created on the client side. Why add such a synchronization argument for vehicles? Why not? Since it was added for peds, it would also be useful for vehicles, especially since sometimes a vehicle needs to be created, e.g. from a database, and this is a better solution than sending data from the database to the client.

Add a "sync" argument for vehicles in createVehicle, as was done for peds in createPed

_No response_

### Additional context

_No response_

### Security Policy

znvjder:
This PR adds the sync argument to createVehicle function on the server side.

Resolves #3333


ghost (approved):
LGTM :shipit:

botder (approved):

T-MaxWiese-T:
For your information, it only does not synchronize if the vehicle is not occupied. I don't know if this is a desired behavior. Because it is actually assumed that there is no synchronization at all.

PR #3357, opened 2024-04-02
Fix missing restore handlers for TIMED_OBJECT and CLUMP

### LINKED ISSUE #3354: Constant crash with engineRequestModel for timed-object
### Describe the bug

When restarting a resource or doing a reconnect, it keeps crashing - Every time or every other time. It occurs in everyone, not just me.
image

Then a message appears that there was a problem loading the model with id 74.
image

I believe this is a problem with timed-object because I've never had this problem with regular objects or peds. Currently, I only have 1 custom id on the server, so no other script causes the crash.

@TheNormalnij 

### Steps to reproduce

1. Request ID for timed-object
2. Load/replace model & txd
3. Restart the resource several times or do recononect several times
4. Crash

### Version

Multi Theft Auto v1.6-release-22396

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
FIxes #3354

Repro resource is [here](https://github.com/multitheftauto/mtasa-blue/issues/3354#issuecomment-2030425963)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientModel.cpp
+++ b/Client/mods/deathmatch/logic/CClientModel.cpp
@@ -109,6 +109,8 @@ void CClientModel::RestoreEntitiesUsingThisModel()
     {
         case eClientModelType::PED:
         case eClientModelType::OBJECT:
+        case eClientModelType::CLUMP:
+        case eClientModelType::TIMED_OBJECT:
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM :shipit:

TracerDS (approved):
Great 👍

PR #3363, opened 2024-04-14
Global variable and default value for resource names

### LINKED ISSUE #2927: Make getResourceName take current resource as a default argument

As in the title. Make `getResourceName` take current resource as a default argument.

 Make `getResourceName` take current resource as a default argument.
Simple change but may come in handy

_No response_

### Additional context

_No response_

### Security Policy

ghost:
### Description
This pull request creates a global variable named `resourceName`, which always contains the name of the current resource, and sets the current resource as the default value to the function `getResourceName`, so if it is not defined as an argument, it is based on that, like in case of other similar functions. Closes #2927.

### Example
```lua
addEventHandler("onResourceStart", resourceRoot,
    function()
        local myResourceName = resourceName or getResourceName()
        local myResourceMessage = string.format("Resource successfully started: %s", myResourceName)

        outputChatBox(myResourceMessage)
    end
)
```


Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -225,10 +225,20 @@ int CLuaResourceDefs::GetResourceName(lua_State* luaVM)
     // Verify arguments
     CResource*       pResource = NULL;
     CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource);
+    argStream.ReadUserData(pResource, NULL);
```
TracerDS: Use `nullptr` instead of `NULL`

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -903,12 +903,31 @@ int CLuaResourceDefs::getResourceName(lua_State* luaVM)
     CResource* pResource;
 
     CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource);
+    argStream.ReadUserData(pResource, NULL);
```
TracerDS: Same as above

TracerDS (changes requested):

ghost:
@TracerDS 
Thanks for the suggestions. 
Although I didn't write this piece of code, I updated it according to modern standards.

TracerDS:
> @TracerDS Thanks for the suggestions. Although I didn't write this piece of code, I updated it according to modern standards.

You could also format it to use ArgumentParser, but I think it would be best for a separate PR somewhere in the future

TracerDS (approved):
lgtm

Fernando-A-Rocha (approved):
Good job

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -223,12 +223,22 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
+        if (!pResource)
+        {
+            // Find our vm and get the root
+            CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
+            if (pLuaMain)
+            {
+                pResource = pLuaMain->GetResource();
+            }
+        }
+
```
TracerDS: Argument parser

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -901,14 +901,32 @@ int CLuaResourceDefs::getResourceLoadTime(lua_State* luaVM)
...
+                lua_pushboolean(luaVM, false);
+                return 1;
+            }
+        }
+
+        if (pResource->IsActive())
+        {
+            lua_pushstring(luaVM, pResource->GetName().c_str());
+            return 1;
+        }
```
TracerDS: Argument parser

TracerDS (changes requested):
If you still want to merge these changes, please use ArgumentParser.
A small refactor wouldnt be so bad here

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Refactored

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Refactored

Fernando-A-Rocha (approved):
:shipit:

TracerDS (approved):
LGTM 👌

Fernando-A-Rocha (approved):

Fernando-A-Rocha:
Bump 😈

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -18,7 +20,7 @@ void CLuaResourceDefs::LoadFunctions()
         {"call", Call},
         {"getThisResource", GetThisResource},
         {"getResourceConfig", GetResourceConfig},
-        {"getResourceName", GetResourceName},
+        {"getResourceName", ArgumentParser<GetResourceName>},
```
tederis: This makes the function backward incompatible. Because there're many scripts that expect this function returns false in case of failure. You should you `ArgumentParserWarn<false, GetResourceName>` here.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -54,7 +54,7 @@ void CLuaResourceDefs::LoadFunctions()
         {"getResourceLoadFailureReason", getResourceLoadFailureReason},
         {"getResourceLastStartTime", getResourceLastStartTime},
         {"getResourceLoadTime", getResourceLoadTime},
-        {"getResourceName", getResourceName},
+        {"getResourceName", ArgumentParser<GetResourceName>},
```
tederis: This makes the function backward incompatible. Because there're many scripts that expect this function returns false in case of failure. You should you `ArgumentParserWarn<false, GetResourceName>` here.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -898,23 +898,33 @@ int CLuaResourceDefs::getResourceLoadTime(lua_State* luaVM)
...
 }
 
-int CLuaResourceDefs::getResourceName(lua_State* luaVM)
+std::string CLuaResourceDefs::GetResourceName(lua_State* luaVM, std::optional<CResource*> resourceElement)
 {
-    CResource* pResource;
+    if (!resourceElement)
+    {
+        CLuaMain*  localVM = m_pLuaManager->GetVirtualMachine(luaVM);
+        CResource* localResource;
```
tederis: It's C++, not C89, there's no need in variables declaration before being use. You can declare this varible and initialize it in one shot at line 911.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -220,34 +222,30 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
-            if (szName)
-            {
-                lua_pushstring(luaVM, szName);
-                return 1;
-            }
-        }
-        else
-            m_pScriptDebugging->LogBadPointer(luaVM, "resource", 1);
+        CLuaMain*  localVM = m_pLuaManager->GetVirtualMachine(luaVM);
+        CResource* localResource;
```
tederis: It's C++, not C89, there's no need in variables declaration before being use. You can declare this varible and initialize it in one shot at line 235.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -220,34 +222,30 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    // Failed
-    lua_pushboolean(luaVM, false);
-    return 1;
+    std::string resourceName = (*resourceElement)->GetName();
+
+    if (resourceName.empty())
+        throw std::runtime_error("The resource name is empty");
```
tederis: There's no need in this check because you simply cannot create a resource with empty name in MTA.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -898,23 +898,33 @@ int CLuaResourceDefs::getResourceLoadTime(lua_State* luaVM)
...
 
-    lua_pushboolean(luaVM, false);
-    return 1;
+    if (!(*resourceElement)->IsActive())
+        throw std::runtime_error("The resource isn't active");
+
+    std::string resourceName = (*resourceElement)->GetName();
+
+    if (resourceName.empty())
+        throw std::runtime_error("The resource name is empty");
```
tederis: There's no need in this check because you simply cannot create a resource with empty name in MTA.

tederis (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: This is my preferred way, but it doesn't matter to me.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: This is my preferred way, but it doesn't matter to me.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -220,34 +222,24 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
+
+        resourceElement = localResource;
     }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    // Failed
-    lua_pushboolean(luaVM, false);
-    return 1;
+    return (*resourceElement)->GetName();
```
TracerDS: Early return + do you want to check if `resourceElement` is an invalid pointer or it wasnt provided?

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -898,23 +898,27 @@ int CLuaResourceDefs::getResourceLoadTime(lua_State* luaVM)
...
     }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    lua_pushboolean(luaVM, false);
-    return 1;
+    if (!(*resourceElement)->IsActive())
+        throw std::runtime_error("The resource isn't active");
+
+    return (*resourceElement)->GetName();
```
TracerDS: Same as above

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.h
```diff
@@ -47,14 +47,15 @@ class CLuaResourceDefs : public CLuaDefs
...
-    LUA_DECLARE(getResourceName);
     LUA_DECLARE(getResourceRootElement);
     LUA_DECLARE(getResourceDynamicElementRoot);
     LUA_DECLARE(getResourceMapRootElement);
     LUA_DECLARE(getResourceExportedFunctions);
     LUA_DECLARE(getResourceOrganizationalPath);
     LUA_DECLARE(isResourceArchived);
+
     static bool isResourceProtected(CResource* const resource);
+    static std::string GetResourceName(lua_State* luaVM, std::optional<CResource*> resourceElement);
```
TracerDS, line 58: ~~`noexcept`~~

TracerDS (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.h
tederis: This function throws exceptions

TracerDS: Ah, didnt notice, my bad.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: Thank you.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -220,34 +222,24 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
-                lua_pushstring(luaVM, szName);
-                return 1;
-            }
-        }
-        else
-            m_pScriptDebugging->LogBadPointer(luaVM, "resource", 1);
+        CLuaMain* localVM = m_pLuaManager->GetVirtualMachine(luaVM);
+
+        if (!localVM)
+            throw std::runtime_error("Couldn't find the virtual machine");
```
tederis: There's another problem. std::runtime_error isn't properly handled within the argument parser. That is when an expection is thrown, the game will be terminated. The argument parser is capable to catch only std::invalid_argument exceptions. But I don't think that it's the case when this type of expressions should be used.

tederis (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
ghost: I should use std::invalid_argument here then?
I have no other ideas currently.
@tederis

tederis: I would get rid of exceptions at all. Because the chance of them being thrown is negligible and doesn't depend on scripts logic.

ghost: Okay

TracerDS: > I should use std::invalid_argument here then? I have no other ideas currently. @tederis

If an error needs to be thrown, `std::invalid_argument` should be used.
ArgumentParser sees it as a custom error anyway, though it is indeed confusing.

ghost: Then I think it's better to keep the throws with std::invalid_argument, because I don't want to return empty strings, and I don't want to return error messages in strings either, because this is not the expected operation of the function.

tederis: It's important to note that `!localVM` is something that will almost never happen. So why then you have to print any message? For me it's better to return false in this case.

ghost: > It's important to note that `!localVM` is something that will almost never happen. So why then you have to print any message? For me it's better to return false in this case.

Removed throws.

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -220,34 +222,24 @@ int CLuaResourceDefs::GetResourceConfig(lua_State* luaVM)
...
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    // Failed
-    lua_pushboolean(luaVM, false);
-    return 1;
+    CLuaMain* localVM = m_pLuaManager->GetVirtualMachine(luaVM);
+
+    if (!localVM)
+        return false;
```
tederis: The returning type is std::string. You cannot return bool here. Maybe you should use std::variant<std::string, bool> for the returning type. I know that's clumsy. The argument parser shall be improved before. So, okay, maybe it's better to choose std::invalid_argument, if you want to.

TracerDS: `std::variant` is fine. Its long but its profitable. `std::invalid_argument` is also fine. Whichever you like

ghost: Reverted for now.

Fernando-A-Rocha (approved):
OK LGTM

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
tederis: If #3493 get merged you will be able to use std::logic_error instead of std::invalid_argument.

Dutchman101:
Enough code reviews and approvals have been submitted, thanks!

PR #3365, opened 2024-04-15
Add extendedwatercannons special world property

### LINKED ISSUE #3315: Only 3 water cannons on Fire Truck can be used at the same time
### Describe the bug

idk if this was intended but only 3 water cannons are spraying out the water
image

### Steps to reproduce

1. Spawn 5 Fire Trucks
2. Warp 5 peds into each Fire Trucks
3. Use setPedControlState(ped, "vehicle_fire", true)
4. Only 3 water cannons are spraying out the water

Or use this resource:
[watercannon.zip](https://github.com/multitheftauto/mtasa-blue/files/14298367/watercannon.zip)
1. Goto coordinate 0, 0, 0,
2. Start the resource

### Version

Client: Multi Theft Auto v1.6-release-22388 (Windows)
Server: MTA:SA Server v1.6-release-22388 (Windows)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Added ``extendedwatercannons`` special world property, that allows the maximum number of water cannons to be extended at the same time. I set the new limit to 30 because that number seems to be enough, perhaps even too much. 

image

Thanks to @Merlin for help with crashes

Resolved #3315


Dark-Dragon:
Would there be any disadvantage to just increasing this limit by default? I don't think this is a case where the limited default behavior is desirable at all, therefore being able to toggle between the two seems rather unnecessary to me.

FileEX:
> Would there be any disadvantage to just increasing this limit by default? I don't think this is a case where the limited default behavior is desirable at all, therefore being able to toggle between the two seems rather unnecessary to me.

The same can be said about vegetation (``randomfoliage``) and underwater creatures - toggle them seems unnecessary, but you can toggle them anyway. I believe we can set this property to enabled by default, but I still think leaving the option to disable it isn't a bad thing. It is better when the user has choice and freedom than predefined settings that he cannot change.

In addition, it seems to me that few people really need an increased limit. In 12 years of playing, I never even noticed that such a limit even existed, and no one had ever reported it before.

Dark-Dragon:
Well there could be had a conversation about what the reasons might be it wasn't reported in the first place, but having it on by default seems like a fine compromise. Still I wouldn't be surprised if it will take more than 12 years for someone to specifically need to disable it, whereas I can see scenarios where foliage and water creatures could be desired or undesired. I'm just thinking the special world properties are slowly starting to get rather crowded, but that's not for me to decide, so I just want to point it out whenever I can.

FileEX:
I think it's good that we get new world properties. Thanks to this, server creators on MTA have more and more freedom and more opportunities. If someone doesn't want foliages, they can disable it, if someone doesn't want flipped cars to burn, they can disable this property. The same in this case, if for some reason someone really does not need such a limit extension, then disable this property. I think it's better to have this option than not to have it.

Lpsd:
Please fix conflicts and we can merge this

FileEX:
> Please fix conflicts and we can merge this

Done

lopezloo:
I think this limit should be increased by default.

FileEX:
> I think this limit should be increased by default.

As you can see https://github.com/multitheftauto/mtasa-blue/pull/3365/commits/bff7007fe674e6c75f129f1ac68b39ed9a2a2d05

The limit is increased by default

Client/game_sa/CGameSA.h
```diff
@@ -347,6 +364,7 @@ class CGameSA : public CGame
     bool         m_areWaterCreaturesEnabled{true};
     bool         m_isBurnFlippedCarsEnabled{true};
     bool         m_isFireballDestructEnabled{true};
+    bool         m_isExtendedWaterCannonsEnabled;
```
botder:
```suggestion
    bool         m_isExtendedWaterCannonsEnabled{false};
```
You were using the default initializer, which uses a random value at whatever address this member ends up at. In a Debug build this is probably set to `false` anyway, so you would never notice this.

Client/game_sa/CGameSA.h
```diff
@@ -82,6 +82,20 @@ extern unsigned int OBJECTDYNAMICINFO_MAX;            // default: 160
 #define CHEAT_NEVERWANTED           "neverwanted"
 #define CHEAT_HEALTARMORMONEY       "healtharmormoney"
 
+#define FUNC_CAESoundManager_CancelSoundsOwnedByAudioEntity 0x4EFCD0
+#define STRUC_CAESoundManager                               0xB62CB0
```
botder: Shouldn't this be `STRUCT_`?

Shared/sdk/net/bitstream.h
```diff
@@ -540,8 +540,17 @@ enum class eBitStreamVersion : unsigned short
     // 2023-10-12
     CPlayerJoinCompletePacket_ServerName,
 
+
+    // Add "extendedwatercannons" to setWorldSpecialPropertyEnabled
+    // 2024-05-23
+    WorldSpecialProperty_ExtendedWaterCannons,
```
botder, line 549: This should be added at the end of the enum.

Shared/sdk/net/bitstream.h
```diff
@@ -540,8 +540,17 @@ enum class eBitStreamVersion : unsigned short
...
+    
+    // Send weapon switch for other players
+    // 2023-01-13
+    OnPlayerWeaponSwitch_Remote,
+
     // Add "roadsignstext" to setWorldSpecialPropertyEnabled
-    // 2024-05-17
+    // 2024-05-23
```
botder: Do not modify the date of this enum field.

Shared/sdk/net/SyncStructures.h
```diff
@@ -2065,6 +2069,7 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
         if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_FireballDestruct))
             bitStream.WriteBits(reinterpret_cast<const char*>(&data2), BITCOUNT2);
 
+        if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_ExtendedWaterCannons))
         if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_RoadSignsText))
             bitStream.WriteBits(reinterpret_cast<const char*>(&data3), BITCOUNT3);
```
botder: Merge error? Looks like you're missing a WriteBits call for the first if-statement.

Shared/sdk/net/SyncStructures.h
```diff
@@ -2097,6 +2102,7 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
 
     struct
     {
+        bool extendedwatercannons : 1;
         bool roadsignstext : 1;
```
botder: Your new bit should be in a new struct `data4`.

Shared/sdk/net/SyncStructures.h
```diff
@@ -2034,7 +2034,7 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
     };
     enum
     {
-        BITCOUNT3 = 1
+        BITCOUNT3 = 2
```
botder: You must use `BITCOUNT4`, don't change `BITCOUNT3` at all.

Shared/mods/deathmatch/logic/Enums.h
```diff
@@ -88,6 +88,7 @@ enum class WorldSpecialProperty
     WATERCREATURES,
     BURNFLIPPEDCARS,
     FIREBALLDESTRUCT,
+    EXTENDEDWATERCANNONS,
     ROADSIGNSTEXT,
```
botder: Move your new field to the bottom of the list.

botder (changes requested):

Shared/sdk/net/SyncStructures.h
FileEX: Some strange things happened during conflict resolution. I've already fixed everything

Shared/sdk/net/bitstream.h
```diff
@@ -539,11 +539,19 @@ enum class eBitStreamVersion : unsigned short
     // Send server name to player in CPlayerJoinCompletePacket
     // 2023-10-12
     CPlayerJoinCompletePacket_ServerName,
+    
+    // Send weapon switch for other players
+    // 2023-01-13
+    OnPlayerWeaponSwitch_Remote,
```
botder: This does not belong here

botder (changes requested):

Shared/sdk/net/bitstream.h
FileEX: Fixed

botder (approved):

PR #3369, opened 2024-04-17
Fix Lua array detection and order preservation when using toJSON

Pieter-Dewachter:
User _Inder00_ on the development Discord server found an interesting bug when using _toJSON_ in combination with Lua arrays that have explicitely defined indices. Reproducing this issue is very straightforward and was also confirmed by _NanoBob_.

`local array = { [1] = "1", [2] = "2", [3] = "3", [4] = "4" }`
`outputChatBox(toJSON(array))`

The above piece of code will output:
`[ { "1": "1", "2": "2", "4": "4", "3": "3" } ]`

We can see that the order of the array was not preserved and the produced JSON is actually an object rather than an array.

Now consider the following piece of code:

`local array = { "1", "2", "3", "4" }`
`outputChatBox(toJSON(array))`

You would expect the same output right? Nope! Now we actually get the correct JSON array with the correct order preserved:
`[ [ "1", "2", "3", "4" ] ]`

When looking into the code, MTA actually assumes that the order is preserved when reading an array from Lua in C++. This is however not guaranteed (Lua uses a hash table internally) as proven by the above examples. My pull request does a two things to properly mitigate this issue:

- The detection of a Lua array is improved by allowing out-of-order indices. An array in Lua is defined as a table with numerical indices that start at 1 and increment by 1 continuously without gaps in between. This is now properly reflected in the code, without an assumtion of the order in which the keys are presented to C++.
- _toJSON_ will now preserve the order of a Lua array correctly, by sorting its elements based on the numerical key before constructing the actual JSON array.


Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
-            CLuaArgument* pArgument = *iter;
+            bIsArray = vecSortedArguments.empty();            // an empty table is also considered an array
+        }
+    }
+
+    if (bIsArray)            // the table is definitely an array
+    {
+        json_object*                                              my_array = json_object_new_array();
+        vector<pair<unsigned int, CLuaArgument*>>::const_iterator iter = vecSortedArguments.begin();
+        for (; iter != vecSortedArguments.end(); ++iter)
```
TracerDS: Use for-range loop instead

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
-            CLuaArgument* pArgument = *iter;
+            bIsArray = vecSortedArguments.empty();            // an empty table is also considered an array
+        }
+    }
+
+    if (bIsArray)            // the table is definitely an array
+    {
+        json_object*                                              my_array = json_object_new_array();
+        vector<pair<unsigned int, CLuaArgument*>>::const_iterator iter = vecSortedArguments.begin();
+        for (; iter != vecSortedArguments.end(); ++iter)
```
TracerDS: Same as above

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
-        for (; iter != m_Arguments.end(); ++iter)
+        // sort the table based on the keys (already handled correctly by std::pair)
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        unsigned int const iFirstKey = vecSortedArguments.front().first;
+        unsigned int const iLastKey = vecSortedArguments.back().first;
+
+        unsigned int const iFirstArrayPos = 1;            // lua arrays are 1 based
+        unsigned int const iLastArrayPos = static_cast<unsigned int>(vecSortedArguments.size());
```
TracerDS: Cant you just use auto here?

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
-        for (; iter != m_Arguments.end(); iter++)
+        // sort the table based on the keys (already handled correctly by std::pair)
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        unsigned int const iFirstKey = vecSortedArguments.front().first;
+        unsigned int const iLastKey = vecSortedArguments.back().first;
+
+        unsigned int const iFirstArrayPos = 1;            // lua arrays are 1 based
+        unsigned int const iLastArrayPos = static_cast<unsigned int>(vecSortedArguments.size());
```
TracerDS: Cant you just use auto here?

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
     {
-        json_object*                          my_array = json_object_new_array();
-        vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
-        for (; iter != m_Arguments.end(); iter++)
+        // sort the table based on the keys (already handled correctly by std::pair)
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        unsigned int const iFirstKey = vecSortedArguments.front().first;
+        unsigned int const iLastKey = vecSortedArguments.back().first;
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
 
     pKnownTables->insert(std::make_pair(this, pKnownTables->size()));
 
-    bool                                  bIsArray = true;
-    unsigned int                          iArrayPos = 1;            // lua arrays are 1 based
-    vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
+    bool                                      bIsArray = true;
+    vector<pair<unsigned int, CLuaArgument*>> vecSortedArguments;            // lua arrays are not necessarily sorted
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Also use `std::vector` instead of `vector`.
`using namespace std` should be banned

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
+    vector<CLuaArgument*>::const_iterator     iter = m_Arguments.begin();
     for (; iter != m_Arguments.end(); iter += 2)
     {
         CLuaArgument* pArgument = *iter;
         if (pArgument->GetType() == LUA_TNUMBER)
         {
             double       num = pArgument->GetNumber();
             unsigned int iNum = static_cast<unsigned int>(num);
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
TracerDS: `std::uint32_t` instead of `unsigned int`

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
-            iter++;            // skip the key values
-            CLuaArgument* pArgument = *iter;
+            bIsArray = vecSortedArguments.empty();            // an empty table is also considered an array
+        }
+    }
+
+    if (bIsArray)            // the table is definitely an array
+    {
+        json_object*                                              my_array = json_object_new_array();
+        vector<pair<unsigned int, CLuaArgument*>>::const_iterator iter = vecSortedArguments.begin();
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
 
     pKnownTables->insert(std::make_pair(this, pKnownTables->size()));
```
TracerDS: Why not `{this, pKnownTables->size()}`?

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
 
-    bool                                  bIsArray = true;
-    unsigned int                          iArrayPos = 1;            // lua arrays are 1 based
-    vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
+    bool                                      bIsArray = true;
+    vector<pair<unsigned int, CLuaArgument*>> vecSortedArguments;            // lua arrays are not necessarily sorted
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
+    vector<CLuaArgument*>::const_iterator     iter = m_Arguments.begin();
     for (; iter != m_Arguments.end(); iter += 2)
     {
         CLuaArgument* pArgument = *iter;
         if (pArgument->GetType() == LUA_TNUMBER)
         {
             double       num = pArgument->GetNumber();
             unsigned int iNum = static_cast<unsigned int>(num);
```
TracerDS: `std::uint32_t` instead of `unsigned int`

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -611,46 +611,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
     {
-        json_object*                          my_array = json_object_new_array();
-        vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
-        for (; iter != m_Arguments.end(); ++iter)
+        // sort the table based on the keys (already handled correctly by std::pair)
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        unsigned int const iFirstKey = vecSortedArguments.front().first;
+        unsigned int const iLastKey = vecSortedArguments.back().first;
```
TracerDS: `std::uint32_t` instead of `unsigned int`

TracerDS (changes requested):
Overall good job, but please dont use `using namespace std` features (implicit scope).

Also lets not use default `unsigned` and signed values as they are platform dependent

FileEX:
Won't this break backwards compatibility?

MegadreamsBE:
> Won't this break backwards compatibility?

Could always make it a flag for a while and warn scripters that the old method will no longer work down the line.

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -535,46 +535,51 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
+    if (bIsArray)            // the table could possibly be an array
     {
-        json_object*                          my_array = json_object_new_array();
-        vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
-        for (; iter != m_Arguments.end(); iter++)
+        // sort the table based on the keys (already handled correctly by std::pair)
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        unsigned int const iFirstKey = vecSortedArguments.front().first;
```
tederis: What if `vecSortedArguments` is empty?

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
Pieter-Dewachter: I left this code as-is in my first commit, changed it now with the second commit

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
Pieter-Dewachter: You're right, I was unsure which syntax was preferred in the mta codebase

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
Pieter-Dewachter: That's an oversight on my part, I used to have everything inline in the if-statement which worked because of lazy evaluation. Thanks for the catch!

Pieter-Dewachter:
> > Won't this break backwards compatibility?
> 
> Could always make it a flag for a while and warn scripters that the old method will no longer work down the line.

Personally, I consider this to be an actual bug that needs fixing. Backwards compatibily shouldn't really be an issue due to the undefined behaviour (order of a hash table) that is causing the issue in the first place. When scripting (either in MTA Lua or something else), both cases (correct and incorrect behaviour) have to be handled anyway so that should not be a problem.

Currently there is an inconsistency between `toJSON` and `fromJSON`where you will not get the original table back if you provide it the table given in the first example. The "keys" (actually indici in this case) will be converted to a string, which causes Lua to handle the table completely different (# operator, ipairs vs pairs, ...).

That's of course my personal opinion, I'm by no means entitled to such decisions.

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -533,49 +533,52 @@ json_object* CLuaArguments::WriteTableToJSONObject(bool bSerialize, CFastHashMap
...
+        std::sort(vecSortedArguments.begin(), vecSortedArguments.end());
+
+        // only the first and last element are checked, everything else is correct by default because the vector was sorted
+        auto const iFirstKey = vecSortedArguments.front().first;
+        auto const iLastKey = vecSortedArguments.back().first;
+
+        auto const iFirstArrayPos = 1U;            // lua arrays are 1 based
+        auto const iLastArrayPos = static_cast<std::uint32_t>(vecSortedArguments.size());
+
+        if (iFirstKey != iFirstArrayPos || iLastKey != iLastArrayPos)
```
AlexTMjugador: I'd add a comment stating that checking that the last array position matches the length of the sorted keys vector is important to handle tables like `{ [1] = "foo", [3] = "bar" }` well, which are array-like but contain gaps that force a JSON object representation.

AlexTMjugador (commented):
The end result of this PR looks fine to me, although I'm slightly concerned about the need for an auxiliary sorted numeric key vector for reliably detecting arrays. Does the Lua API expose some other way to detect if a table is indeed an array, hopefully one that's more efficient?

About backwards compatibility, I think we should explore exactly what conditions make tables be unordered vs. ordered. If a script author wrote a table with the more verbose `{ [1] = "1", [2] = "2", [3] = "3", [4] = "4" }` syntax and, due to some Lua implementation detail, was lead to believe that such verbose syntax is an approved way to let tables be serialized as JSON objects, I could imagine this change causing some real pain, even if the new behavior is more correct. On the other hand, if that explicit syntax could also happen to generate ordered JSON arrays a significant fraction of the time, I'd be inclined to think that the old behavior was broken beyond repair and we shouldn't really cater for it, as it could never be relied upon in the first place.

Pieter-Dewachter:
> The end result of this PR looks fine to me, although I'm slightly concerned about the need for an auxiliary sorted numeric key vector for reliably detecting arrays. Does the Lua API expose some other way to detect if a table is indeed an array, hopefully one that's more efficient?
> 
> About backwards compatibility, I think we should explore exactly what conditions make tables be unordered vs. ordered. If a script author wrote a table with the more verbose `{ [1] = "1", [2] = "2", [3] = "3", [4] = "4" }` syntax and, due to some Lua implementation detail, was lead to believe that such verbose syntax is an approved way to let tables be serialized as JSON objects, I could imagine this change causing some real pain, even if the new behavior is more correct. On the other hand, if that explicit syntax could also happen to generate ordered JSON arrays a significant fraction of the time, I'd be inclined to think that the old behavior was broken beyond repair and we shouldn't really cater for it, as it could never be relied upon in the first place.

The interesting thing is that it seems to trigger only in certain cases as you feared already. For example this table will already serialize to an array correctly before my changes: `{ [1] = "1", [2] = "2", [3] = "3" }`. So the behaviour seems to be quite broken already unfortunately.

AlexTMjugador:
> So the behaviour seems to be quite broken already unfortunately.

Right, that's useful to know. Calling `fromJSON` of a `toJSON` result should deserialize to the same Lua value no matter if the array or object representation is chosen, right? If so, then this change won't introduce a backwards-incompatible change with how MTA parses its own JSON data, which I think it's all we should cater for, as separate applications never were able to exchange arrays reliably anyway (programmers of correct and smart separate applications probably designed them to allow both representations).

Pieter-Dewachter:
> > So the behaviour seems to be quite broken already unfortunately.
> 
> Right, that's useful to know. Calling `fromJSON` of a `toJSON` result should deserialize to the same Lua value no matter if the array or object representation is chosen, right? If so, then this change won't introduce a backwards-incompatible change with how MTA parses its own JSON data, which I think it's all we should cater for, as separate applications never were able to exchange arrays reliably anyway (programmers of correct and smart separate applications probably designed them to allow both representations).

Before my changes, certain Lua arrays (those represented as an object in JSON, which seems to be "random") would actually deserialize to a Lua table containing the keys as a string rather than a number. After my change, this is no longer the case for any Lua array. This means calling `fromJSON` of a `toJSON` becomes fully deterministic* unlike before.

* Array-like Lua tables with gaps in them will still deserialize to a Lua table with string keys, that's a limitation of the JSON specification as far as I know.

TheNormalnij:
> * Array-like Lua tables with gaps in them will still deserialize to a Lua table with string keys, that's a limitation of the JSON specification as far as I know.

This behavior is specific to the current implementation.  It's technically impossible to implement a universal Lua -> JSON serializer in one function for an unknown table.
If you need a specific JSON structure, you might use object-oriented JSON builders.

Dutchman101:
Would like a few more reviewers to focus on the backwards compatibility/existing scripts aspect

Fernando-A-Rocha (approved):

ghost:
Bump?

botder (approved):

PR #3370, opened 2024-04-21
onPlayerTeamChange Event

esmail9900:
implemented the onPlayerTeamChange event for the server-side. this event includes two arguments: oldTeam and newTeam. if the player doesn't have an old team, oldTeam will be false.


Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9203,13 +9203,23 @@ bool CStaticFunctionDefinitions::SetTeamColor(CTeam* pTeam, unsigned char ucRed,
...
+        // Call the Event
+        CLuaArguments Arguments;
+        if (currentTeam){
+            Arguments.PushString(currentTeam->GetTeamName());
+        } else {
+            Arguments.PushBoolean(false);
+        }
+        Arguments.PushString(pTeam->GetTeamName());
+        pPlayer->CallEvent("onPlayerTeamChange", Arguments);
+
```
TracerDS: Use early return please.

TracerDS (changes requested):

tederis:
How about enabling cancelEvent() for this new event?

Citizen01:
Hello @esmail9900 and thank you for your PR.

Your PR gave me some strong flashbacks from when I attempted to add this event and lost motivation.

Please view the comments on this PR https://github.com/multitheftauto/mtasa-blue/pull/10 and check if your code complies with all that have been said in this PR.

Thank you.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9203,22 +9203,30 @@ bool CStaticFunctionDefinitions::SetTeamColor(CTeam* pTeam, unsigned char ucRed,
...
-    if (pTeam != pPlayer->GetTeam())
-    {
-        // Change his team
-        pPlayer->SetTeam(pTeam, true);
-
-        // Tell everyone his new team
-        CBitStream BitStream;
-        BitStream.pBitStream->Write(pTeam ? pTeam->GetID() : INVALID_ELEMENT_ID);
-        m_pPlayerManager->BroadcastOnlyJoined(CElementRPCPacket(pPlayer, SET_PLAYER_TEAM, *BitStream.pBitStream));
+    if (pTeam == currentTeam){
```
Citizen01: The comment above should be removed or updated.

I would comply with the code around you in this file and not use brackets {} for if conditions with only one statement in it.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9203,22 +9203,30 @@ bool CStaticFunctionDefinitions::SetTeamColor(CTeam* pTeam, unsigned char ucRed,
...
+    // Change his team
+    pPlayer->SetTeam(pTeam, true);
 
-        return true;
+    // Call the Event
+    CLuaArguments Arguments;
+    if (currentTeam){
+        Arguments.PushString(currentTeam->GetTeamName());
+    } else {
+        Arguments.PushBoolean(false);
```
Citizen01: can you push the team element instead of the names ? (for oldTeam and newTeam)
I feel like it would be easier for lua devs to work with the elements instead of the names.

for oldTeam I was expecting to get `nil` instead of `false` if there is no "currentTeam")

One minor thing and i don't really know the latest rules about it but I think we should keep the formating style of the current file for consistency and add a new line for every brackets:
```c++
    if (currentTeam)
    {
        Arguments.PushString(currentTeam->GetTeamName());
    }
    else
    {
        Arguments.PushBoolean(false);
    }
```
(Personnaly I prefer your version except it lacks a space before the 1st opening bracket)

Citizen01 (changes requested):
some stuff needs to be adressed.

I think the this event do not need to be cancelable right away, this can be added later with another PR as you will surely encounter some tricky problems to overcome to make it work properly.

Note: What happens if you delete the team ?
(Expectected behaviour would be that this event gets called for every players in the team and before onElementDestroy gets called so that it is already at the right place if you or someone else decides to make your event cancelable)

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -333,14 +333,30 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
     {
         return false;
     }
+    // Its team trigger onPlayerTeamChange for each player in the team
+    else if (iType == CElement::TEAM) 
+    {
```
Citizen01: Here you can remove the "else" part because it's kind of implicit. And put a blank line at 336 to seperate both conditions
```suggestion
    {
        return false;
    }
    
    // Its team trigger onPlayerTeamChange for each player in the team
    if (iType == CElement::TEAM) 
    {
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -333,14 +333,30 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
...
+        CTeam* pTeam = static_cast<CTeam*>(pElement);
+
+        auto iterBegin = pTeam->PlayersBegin();
+        auto iterEnd = pTeam->PlayersEnd();
+
+        for (auto iter = iterBegin; iter != iterEnd; ++iter){
```
Citizen01:
```suggestion
        for (auto iter = iterBegin; iter != iterEnd; ++iter)
        {
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9203,22 +9219,32 @@ bool CStaticFunctionDefinitions::SetTeamColor(CTeam* pTeam, unsigned char ucRed,
 bool CStaticFunctionDefinitions::SetPlayerTeam(CPlayer* pPlayer, CTeam* pTeam)
 {
     assert(pPlayer);
-
+    CTeam* currentTeam = pPlayer->GetTeam();
```
Citizen01: a blank line after the assert would be nice #cosmetic

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -333,14 +333,30 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
...
+    {
+        CTeam* pTeam = static_cast<CTeam*>(pElement);
+
+        auto iterBegin = pTeam->PlayersBegin();
+        auto iterEnd = pTeam->PlayersEnd();
+
+        for (auto iter = iterBegin; iter != iterEnd; ++iter){
+            CPlayer* player = *iter;
+            CLuaArguments arguments;
+            arguments.PushNil(); // Current team going to delete so return nil
```
Citizen01: damn, I don't know about this one. I'd love to have the real team element there. Even though it's about to be deleted.

CallEvent are synchronous so I think we should still send it, all the event handlers will do what they have to do with the team element being destroyed and I think it will do it.

PS: If anyone sees a reason not to send the team element being destroyed at that stage, please add your input.

Citizen01 (changes requested):
looks pretty good. Added some more comments.

but this need to be tested when I have time (or someone else ? be my guest)

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
esmail9900: at first it was team element, but i thought it will be better to return nil, but i think return team element its actually better

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
esmail9900: yeah, i will make it better

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -344,7 +342,7 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
         for (auto iter = iterBegin; iter != iterEnd; ++iter){
             CPlayer* player = *iter;
             CLuaArguments arguments;
-            arguments.PushNil(); // Current team going to delete so return nil
+            arguments.PushNil(pTeam); // Return team element as oldteam
```
Citizen01:
```suggestion
            arguments.PushElement(pTeam); // Return team element as oldteam
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -329,9 +329,24 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
     // We can't destroy the root or a player/remote client/console
     int iType = pElement->GetType();
     if (pElement == m_pMapManager->GetRootElement() || iType == CElement::PLAYER || iType == CElement::CONSOLE ||
-        g_pGame->GetResourceManager()->IsAResourceElement(pElement))
-    {
+        g_pGame->GetResourceManager()->IsAResourceElement(pElement)){
```
Citizen01: revert this change, it was fine

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -9204,21 +9219,32 @@ bool CStaticFunctionDefinitions::SetPlayerTeam(CPlayer* pPlayer, CTeam* pTeam)
...
+    // Call the Event
+    CLuaArguments Arguments;
+    if (currentTeam)
     {
-        // Change his team
-        pPlayer->SetTeam(pTeam, true);
+        Arguments.PushElement(currentTeam);
+    } else {
+        Arguments.PushNil(); // No oldTeam return nil
+    }
```
Citizen01: Not resolved. Please do it like so:
```c++
    if (currentTeam)
    {
        Arguments.PushElement(currentTeam);
    } 
    else
    {
        Arguments.PushNil(); // No oldTeam return nil
    }
```

Citizen01 (changes requested):
Tested it locally and it works fine. Cancelling the event works too and the event is called when the team is deleted as well.

I think this PR will be ready for merge once you resolve the latest things.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Citizen01: not resolved

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Citizen01: not resolved

Citizen01 (changes requested):
You forgot some of my suggestions.

If they have been left on purpose, please explain why.

esmail9900:
> You forgot some of my suggestions.
> 
> If they have been left on purpose, please explain why.

esmail9900:
> You forgot some of my suggestions.
> 
> If they have been left on purpose, please explain why.

mb, i just confused, im new to github :D

Citizen01 (dismissed):
Looks okay for me. Congrats on your 1st PR.

I'll ask for another review just in case. It can take a bit of time so don't worry if nothing happens in the next few days.

Thank you

ghost (approved):
LGTM :shipit:

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -328,19 +328,35 @@ bool CStaticFunctionDefinitions::DestroyElement(CElement* pElement)
...
+    if (iType == CElement::TEAM) { // Its team trigger onPlayerTeamChange for each player in the team
+        CTeam* pTeam = static_cast<CTeam*>(pElement);
+
+        auto iterBegin = pTeam->PlayersBegin();
+        auto iterEnd = pTeam->PlayersEnd();
+
+        for (auto iter = iterBegin; iter != iterEnd; ++iter)
+        {
+            CPlayer* player = *iter;
+            CLuaArguments arguments;
```
tederis: Potentially a number of players in a team can be significant. And you can handle this situation better by taking `CLuaArguments arguments` out of the loop and calling `arguments.DeleteArguments()` on an every iteration.

ghost (approved):

tederis (dismissed):

TracerDS (approved):
Nice work 🎉

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Citizen01: Didn't know it was somewhat expensive (good to know)

Citizen01 (dismissed):

ghost (approved):

Dutchman101:
There are enough passed code reviews, thanks for your contribution @esmail9900

PR #3378, opened 2024-05-01
Add isObjectMoving to server side

FileEX:
This PR adds the ``isObjectMoving`` function also on the server side, making it available as a shared function.

Test server-side code
```lua
local object

addCommandHandler('obj', function(plr,cmd)
	local x,y,z = getElementPosition(plr);
	object = createObject(1337, x,y,z);

	moveObject(object, 20*1000, 0,0,5);
end);

addCommandHandler('objm', function(plr,cmd)
	print(isObjectMoving(object));
end);
```

### DIFF
--- a/Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
@@ -55,7 +55,7 @@ void CLuaObjectDefs::AddClass(lua_State* luaVM)
 
     lua_classvariable(luaVM, "scale", "setObjectScale", "getObjectScale");
     lua_classvariable(luaVM, "breakable", "setObjectBreakable", "isObjectBreakable");
-    lua_classvariable(luaVM, "moving", NULL, "isObjectMoving");
+    lua_classvariable(luaVM, "moving", nullptr, "isObjectMoving");
 
     lua_registerclass(luaVM, "Object", "Element");
 }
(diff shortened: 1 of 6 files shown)


ghost (approved):
Looks good to me

ghost (approved):

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -222,6 +225,11 @@ int CLuaObjectDefs::SetObjectScale(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaObjectDefs::IsObjectMoving(CObject* const pObject)
```
TracerDS, line 228: Could you add `const` to the parameter itself?
```cpp
bool CLuaObjectDefs::IsObjectMoving(const CObject* const pObject)
```

TracerDS (changes requested):
Your commitment is extraordinary, good job!

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
FileEX: No, because the methods in the CObject class are not constant and it would require editing many methods in this class because the IsMoving method is used by other functions

TheNormalnij (approved):

Lpsd:
Please add the relevant wiki page for this, including the "new feature" template (once the nightly build has been created).

See other functions/events - for example: https://wiki.multitheftauto.com/wiki/OnPlayerProjectileCreation

PR #3379, opened 2024-05-01
Add serverside objects move events

FileEX:
We have ``onClientObjectMoveStart`` and ``onClientObjectMoveStop`` events - This PR adds these events also for the server side.

The ``stoppedByScript`` parameter determines whether the object has been stopped using stopObject or has reached its destination


Server/mods/deathmatch/logic/CObject.cpp
```diff
@@ -346,9 +346,12 @@ void CObject::Move(const CPositionRotationAnimation& a_rMoveAnimation)
         SetPosition(positionRotation.m_vecPosition);
         SetRotation(positionRotation.m_vecRotation);
     }
+
+    CLuaArguments Arguments;
+    this->CallEvent("onObjectMoveStart", Arguments);
```
tederis: Is `this->` really needed?

FileEX:
I had to change the way the ``onObjectStop`` event was triggered. This must be synchronized via a packet on the client side because on the server side StopMovement is not called when the object reaches its destination as this happens on the client side.

I also added the ``stoppedByScript`` parameter for the ``onClientObjectMoveStop`` event on the client side.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -2772,7 +2772,7 @@ void CClientGame::AddBuiltInEvents()
     m_Events.AddEvent("onClientObjectDamage", "loss, attacker", NULL, false);
     m_Events.AddEvent("onClientObjectBreak", "attacker", NULL, false);
     m_Events.AddEvent("onClientObjectMoveStart", "", NULL, false);
-    m_Events.AddEvent("onClientObjectMoveStop", "", NULL, false);
+    m_Events.AddEvent("onClientObjectMoveStop", "stoppedByScript", NULL, false);
```
TracerDS: `nullptr` instead of `NULL`

Client/mods/deathmatch/logic/CDeathmatchObject.cpp
```diff
@@ -89,7 +89,12 @@ void CDeathmatchObject::_StopMovement(bool a_bUnregister)
         delete m_pMoveAnimation;
         m_pMoveAnimation = NULL;
         CLuaArguments Arguments;
-        this->CallEvent("onClientObjectMoveStop", Arguments, true);
+        Arguments.PushBoolean(bStoppedByScript);
+
+        CallEvent("onClientObjectMoveStop", Arguments, true);
+
+        if (!IsLocalEntity())
+            SendObjectStopPacket(bStoppedByScript);
```
TracerDS: Early return

Client/mods/deathmatch/logic/CDeathmatchObject.cpp
```diff
@@ -89,7 +89,12 @@ void CDeathmatchObject::_StopMovement(bool a_bUnregister)
         delete m_pMoveAnimation;
         m_pMoveAnimation = NULL;
```
TracerDS: `nullptr` instead of `NULL`

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1546,6 +1552,8 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerVoiceStop", "", NULL, false);
 
     // Object events
+    m_Events.AddEvent("onObjectMoveStart", "", NULL, false);
+    m_Events.AddEvent("onObjectMoveStop", "stoppedByScript", NULL, false);
```
TracerDS: `nullptr` instead of `NULL`

TracerDS (changes requested):
Good job

Client/mods/deathmatch/logic/CDeathmatchObject.cpp
FileEX: This is not my code. There are a lot of NULLs throughout the file, as well as throughout the project. Better to do a separate PR with refactoring

TracerDS: > This is not my code. There are a lot of NULLs throughout the file, as well as throughout the project. Better to do a separate PR with refactoring

Well, you are changing the file already so the change would not be significant

FileEX: Okay, so let's do it

TheNormalnij:
A packet from a client to notify server about this event is overkill.
onObjectMoveStart and onObjectMoveStop are pointless on server, because server has full control over created objects in scripts, and setTimer replaces onObjectMoveStop.
This realization still request setTimer, because onObjectMoveStop doesn't work on empty server. This can cause frustrating behavior and strange bugs in server scripts

FileEX:
So how to improve it?

PR #3381, opened 2024-05-02
Add breakObject to server-side

FileEX:
This PR adds the ``breakObject`` function also on the server side, making it available as a shared function

```lua
local object

addCommandHandler('obj', function(plr,cmd)
	local x,y,z = getElementPosition(plr);
	object = createObject(1226, x,y,z);
end);

addCommandHandler('objb', function(plr)
	breakObject(object);
end);
```


Client/mods/deathmatch/logic/rpc/CObjectRPCs.cpp
```diff
@@ -126,3 +127,11 @@ void CObjectRPCs::SetObjectBreakable(CClientEntity* pSource, NetBitStreamInterfa
         pObject->SetBreakable(bitStream.ReadBit());
     }
 }
+
+void CObjectRPCs::BreakObject(CClientEntity* pSource, NetBitStreamInterface& bitStream)
+{
+    CDeathmatchObject* pObject = static_cast<CDeathmatchObject*>(m_pObjectManager->Get(pSource->GetID()));
```
TracerDS: you could use `auto` here to somehow increase readability.
Its clear that the output type is going to be `CDeathmatchObject*`

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -8235,6 +8235,24 @@ bool CStaticFunctionDefinitions::StopObject(CElement* pElement)
...
+        CObject* pObject = static_cast<CObject*>(pElement);
+
+        if (pObject != nullptr)
+        {
+            CBitStream BitStream;
+            m_pPlayerManager->BroadcastOnlyJoined(CElementRPCPacket(pObject, BREAK_OBJECT, *BitStream.pBitStream));
+        }
+    }
+
+    return false;
```
TracerDS: Early return

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -297,3 +299,8 @@ bool CLuaObjectDefs::SetObjectBreakable(CObject* const pObject, const bool bBrea
 {
     return CStaticFunctionDefinitions::SetObjectBreakable(pObject, bBreakable);
 }
+
+bool CLuaObjectDefs::BreakObject(CObject* const pObject)
```
TracerDS, line 311: Could you add `const` to the `CObject` itself too?
```cpp
bool CLuaObjectDefs::BreakObject(const CObject* const pObject)
```

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.h
```diff
@@ -34,4 +34,5 @@ class CLuaObjectDefs : public CLuaDefs
     static bool SetObjectBreakable(CObject* const pObject, const bool bBreakable);
     LUA_DECLARE(MoveObject);
     LUA_DECLARE(StopObject);
+    static bool BreakObject(CObject* const pObject);
```
TracerDS, line 38: Above

TracerDS (changes requested):
Good job

Shared/sdk/net/rpc_enums.h
```diff
@@ -115,6 +115,7 @@ enum eElementRPCFunctions
     SET_OBJECT_ROTATION,
     MOVE_OBJECT,
     STOP_OBJECT,
+    BREAK_OBJECT,
```
TheNormalnij: Should be the last item on the list.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -8235,6 +8235,24 @@ bool CStaticFunctionDefinitions::StopObject(CElement* pElement)
     return false;
 }
 
+bool CStaticFunctionDefinitions::BreakObject(CElement* pElement)
+{
+    RUN_CHILDREN(BreakObject(*iter));
```
TheNormalnij, line 8258: Add min client version check.
Server should not call this when the minimal client version doesn't support this event to avoid desync problems.

FileEX: You mean adding min. version in CResourceChecker?

botder:
This ignores the `object->IsBreakable()` value on server-side.

FileEX:
> This ignores the `object->IsBreakable()` value on server-side.

Fixed

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -3078,6 +3079,9 @@ void CPacketHandler::Packet_EntityAdd(NetBitStreamInterface& bitStream)
                         if (bitStream.Read(&health))
                             pObject->SetHealth(health.data.fValue);
 
+                        if (bitStream.ReadBit())
+                            pObject->Break();
```
botder: This is going to fail if either the server is running an older version (and doesn't send this bit) or when the client is older (and the server sends a bit here). You have to check if the bitstream supports this.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -297,6 +297,9 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
                     health.data.fValue = pObject->GetHealth();
                     BitStream.Write(&health);
 
+                    // is object break?
+                    BitStream.WriteBit(pObject->GetHealth() <= 0);
```
botder: Again: you have to check if the bitstream supports this.

botder (changes requested):

Client/mods/deathmatch/logic/CPacketHandler.cpp
botder: https://github.com/multitheftauto/mtasa-blue/blob/06d1f1daf55a97bddd2fdf2ce80ad4a01085fbb0/Shared/sdk/net/bitstream.h#L397-L400

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
FileEX: I understand client-side checking because the server may not send this bit, but why server-side checking? If the server is older, it won't have this piece of code, right?

botder: If an older client doesn't support this, and the newer server sends this bit, the client will read garbage.

FileEX: Yes, I know on the client side why we check. But why do we check this with WriteBit? If the server is older, it simply does not have this line of code

TheNormalnij: New server breaks old clients with unexpected data in stream

FileEX: So checking BitStream::Can also checks whether the client (who will receive the packet) has the same version as the server? Because I still don't understand why we check on the server side whether the bit can be sent, when the older server simply won't send it because it doesn't have this piece of code

FileEX: @botder ?

botder: You should add it. Try to imagine what will happen, when the server sends the extra-bit here, and an older client reads the bitstream.

FileEX: > So checking BitStream::Can also checks whether the client (who will receive the packet) has the same version as the server?

So checking BitStream::Can (on the server-side) also checks whether the client (who will receive the packet) has the same version as the server?

botder: I don't know what the server does - I would have to look it up - I assume that it creates a BitStream for every player and then uses the Packet::Write method for every instance.

FileEX: Okey, done.

botder (approved):

PR #3382, opened 2024-05-02
Add message parameter to onPlayerPrivateMessage 

### LINKED ISSUE #2135: onPlayerPrivateMessage returns bad "message" argument
**Describe the bug**
onPlayerPrivateMessage event returns recipient nick + message as "message" event

**To reproduce**
```lua
addEventHandler('onPlayerPrivateMessage', root, function(msg, rec)
    print(msg, getPlayerName(rec))
    cancelEvent();
    return;
end);
```

i executed command "/msg Adams e"

Result: 
image

**Expected behaviour**
message argument should be only message without recipient name

**Screenshots**

**Version**
1.5.8

**Additional context**

FileEX:
This PR adds a parameter containing only the message content.

``string fullMessage, player recipient, string content``

Resolved #2135

### DIFF
--- a/Server/mods/deathmatch/logic/CConsoleCommands.cpp
+++ b/Server/mods/deathmatch/logic/CConsoleCommands.cpp
@@ -596,8 +596,10 @@ bool CConsoleCommands::Msg(CConsole* pConsole, const char* szInArguments, CClien
 
                                         // Send the message and player pointer to the script
                                         CLuaArguments Arguments;
-                                        Arguments.PushString(szArguments);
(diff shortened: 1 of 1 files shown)


ghost (approved):

Lpsd (approved):
Looks good, please update https://wiki.multitheftauto.com/wiki/OnPlayerPrivateMessage with the new parameter once the nightly build has been generated. 

You may want to use the "New feature" block ([example](https://wiki.multitheftauto.com/wiki/OnPlayerProjectileCreation)), or at-least mention somewhere that the `content` parameter is only available since r<insert_revision_number>.

Issue #672, opened 2018-10-27
Some traffic signs have no text


ClipperBled:
**Describe the bug**
Text missing from some traffic signs

**To reproduce**

Steps to reproduce the behaviour:
1. Go around the map
2. Look for missing text on traffic signs

**Expected behaviour**
Expecting text to be on such signs, like singleplayer.

**Screenshots**
![mta-screen_2018-10-27_19-41-06](https://user-images.githubusercontent.com/42635966/47606838-461ece00-da21-11e8-862b-9eaca76ed2ac.png)
![mta-screen_2018-10-27_19-41-16](https://user-images.githubusercontent.com/42635966/47606839-461ece00-da21-11e8-81d8-19d4c1a50fcd.png)
![mta-screen_2018-10-27_19-41-50](https://user-images.githubusercontent.com/42635966/47606840-461ece00-da21-11e8-836a-878c52cf97ff.png)

**Version**
Multi Theft Auto v1.5.6-release-14664.4
Copyright (C) 2003 - 2018 Multi Theft Auto

MegadreamsBE:
If added it would be nice this could also be disabled. This way custom text can be added to the signs.

ccw808:
Can be caused by modified GTA files

ClipperBled:
I have vanilla game, no mods. And the bug can be seen by everyone from the server,

Dutchman101:
> Can be caused by modified GTA files

I also thought so, but after checking (without any mods myself) I can confirm that atleast some signs in the direct surrounding of where he took the screenshot, are empty.

FileEX:
This is GTA SA bug

https://www.youtube.com/watch?v=lTC8D7z2f0M (2:52:00, 3:14:40)

botder:
Use [mapfixes](https://github.com/multitheftauto/mtasa-resources/tree/master/%5Bgameplay%5D/mapfixes) to resolve the issue.

PR #3385, opened 2024-05-12
Add roadsignstext special world property

FileEX:
Added ``roadsignstext`` special world property, which allows you to disable text rendering on road signs. Thanks to this, we have the opportunity to make our own road signs, which was suggested e.g. in #672 

Enabled:
image

Disabled:
image

### DIFF
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -729,6 +729,17 @@ void CGameSA::SetFireballDestructEnabled(bool isEnabled)
     m_isFireballDestructEnabled = isEnabled;
 }
 
+void CGameSA::SetRoadSignsTextEnabled(bool isEnabled)
+{
+    if (isEnabled == m_isRoadSignsTextEnabled)
+        return;
+
+    // Skip JMP to CCustomRoadsignMgr::RenderRoadsignAtomic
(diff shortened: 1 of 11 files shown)


ghost (approved):

samr46:
Just wanna let you know again that current implementation is not backwards compatible (my comment with details in PR #3365)

FileEX:
> Just wanna let you know again that current implementation is not backwards compatible (my comment with details in PR #3365)

I don't see your comment in that PR

samr46:
> I don't see your comment in that PR

You need to create new struct for this property and add new [eBitStreamVersion](https://github.com/multitheftauto/mtasa-blue/blob/1f2c6e75fb71b01f0053f151e766a232ed33692b/Shared/sdk/net/bitstream.h#L535-L537) in order for it to be backwards compatible. Please look at how "fireballdestruct" is implemented. There are also references to this struct in CMapInfoPacket.cpp and CPacketHandler.cpp

Lpsd (approved):
Looks good to me

Lpsd:
Please update https://wiki.multitheftauto.com/wiki/SetWorldSpecialPropertyEnabled once the next nightly build has been created (after r22420).

PR #3391, opened 2024-05-16
Fix ability to move server-side vehicles that are far away from the player

### LINKED ISSUE #3389: Client-side setElementPosition syncs for server-side vehicles
### Describe the bug

As in title, it doesn't happen for other element types. It doesn't account whether player is syncer of vehicle. The way to prevent it is to use:
```lua
setElementSyncer(vehicleElement, false, true)
```
And set syncer manually when needed.

### Steps to reproduce

```lua
-- server

local vehicleElement = createVehicle(434, 0, 0, 3)

setElementID(vehicleElement, "testVeh")

function checkVehiclePosition()
	local vehX, vehY, vehZ = getElementPosition(vehicleElement)

	iprint(vehX, vehY, vehZ)
end
setTimer(checkVehiclePosition, 500, 0)

-- client

local playerX, playerY, playerZ = getElementPosition(localPlayer)
local vehicleElement = getElementByID("testVeh")

setElementPosition(vehicleElement, playerX, playerY, playerZ)
```

### Version

Multi Theft Auto v1.6-release-22390

### Additional context

Ideally, it shouldn't affect server created vehicles at all. At the very least it should be limited to syncer only.

### Relevant log output

_No response_

### Security Policy

MegadreamsBE:
fixes https://github.com/multitheftauto/mtasa-blue/issues/3389

Adds extra checks to ensure that players cannot use setElementPosition client-side to move vehicles that are far away from them. Essentially closes loopholes that made it possible for players to become the syncer of said vehicles.

### DIFF
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -1455,6 +1455,7 @@ const std::vector<SIntSetting>& CMainConfig::GetIntSettingList()
         {true, true, 10, 50, 1000, "update_cycle_messages_limit", &m_iUpdateCycleMessagesLimit, &CMainConfig::ApplyNetOptions},
         {true, true, 50, 100, 400, "ped_syncer_distance", &g_TickRateSettings.iPedSyncerDistance, &CMainConfig::OnTickRateChange},
         {true, true, 50, 130, 400, "unoccupied_vehicle_syncer_distance", &g_TickRateSettings.iUnoccupiedVehicleSyncerDistance, &CMainConfig::OnTickRateChange},
+        {true, true, 0, 30, 130, "vehicle_contact_sync_radius", &g_TickRateSettings.iVehicleContactSyncRadius, &CMainConfig::OnTickRateChange},
         {false, false, 0, 1, 2, "compact_internal_databases", &m_iCompactInternalDatabases, NULL},
         {true, true, 0, 1, 2, "minclientversion_auto_update", &m_iMinClientVersionAutoUpdate, NULL},
         {true, true, 0, 0, 100, "server_logic_fps_limit", &m_iServerLogicFpsLimit, NULL},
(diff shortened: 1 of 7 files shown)


ghost (approved):
Looks good to me 🚀

Lpsd:
I think it would be good to make `MAX_CONTACT_SYNC_RADIUS` value configurable via the Lua API

MegadreamsBE:
> I think it would be good to make `MAX_CONTACT_SYNC_RADIUS` value configurable via the Lua API

In that case it'd probably be better to link it with the syncer radius and make them configurable together. On its own there's not much of a reason to be able to configure this radius

EDIT: Adds the ability to disable it which is actually pretty useful, I stand corrected and thus implemented it

ghost (approved):
Looks good 🎉

MegadreamsBE:
> I think it would be good to make `MAX_CONTACT_SYNC_RADIUS` value configurable via the Lua API

Added a configurable server setting for it. You can configure it as follows during runtime:
`setServerConfigSetting('vehicle_contact_sync_radius', 30)`

Which should satisfy this.

haron4igg:
Just to be sure...

If client will set position for vehicle locally - will car appear locally to him?
If so - when player will enter the car - it will be forcely synced to new location anyway?

MegadreamsBE:
> Just to be sure...
> 
> If client will set position for vehicle locally - will car appear locally to him? If so - when player will enter the car - it will be forcely synced to new location anyway?

If you're not the syncer, even if the vehicle appears to be near you on your client, you cannot enter it. Instead attempting to the enter the vehicle will reset its position to the position stored on the server. It has always been like this and the PR does not change that.

Lpsd (approved):
LGTM!

Lpsd:
Thanks for this PR, please update https://wiki.multitheftauto.com/wiki/SetServerConfigSetting and https://wiki.multitheftauto.com/wiki/Server_mtaserver.conf where necessary.

PR #3393, opened 2024-05-17
Update acl.xml - add runcode loadstring permission

Fernando-A-Rocha:
Default MTA Resource `runcode` requires function.loadstring access.

With default acl.xml, it kept printing a warning when you start the MTA server.

### DIFF
--- a/Server/mods/deathmatch/acl.xml
+++ b/Server/mods/deathmatch/acl.xml
@@ -49,8 +49,8 @@
    <group name="DevGroup">
       <acl name="DevACL" />
    </group>
-   <group name="runcodeACLGroup">
-      <acl name="runcodeACL" />
+   <group name="RuncodeGroup">
+      <acl name="RuncodeACL" />
       <object name="resource.runcode" />
    </group>
    <acl name="Default">
@@ -282,7 +282,7 @@
         <right name="resource.performancebrowser.http" access="true"></right>
         <right name="resource.ajax.http" access="true"></right>
    </acl>
-   <acl name="runcodeACL">
+   <acl name="RuncodeACL">
       <right name="function.loadstring" access="true" />
    </acl>
 </acl>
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/acl.xml
```diff
@@ -49,6 +49,10 @@
    <group name="DevGroup">
       <acl name="DevACL" />
    </group>
+   <group name="runcodeACLGroup">
```
ghost: I think we should stick with the previous naming conventions, RuncodeGroup would be a better name.

Server/mods/deathmatch/acl.xml
```diff
@@ -278,4 +282,7 @@
         <right name="resource.performancebrowser.http" access="true"></right>
         <right name="resource.ajax.http" access="true"></right>
    </acl>
+   <acl name="runcodeACL">
```
ghost: I think we should stick with the previous naming conventions, RuncodeACL would be a better name.

ghost (changes requested):

jlillis:
I don't think we should give `runcode` any rights by default. It would be better to make admins explicitly give it `loadstring` access using an aclrequest.

Fernando-A-Rocha:
> I don't think we should give `runcode` any rights by default. It would be better to make admins explicitly give it `loadstring` access using an aclrequest.

Sure, it should only be used by server developers that know what they're doing with the code ran...

Fernando-A-Rocha:
Naming changes were made in case you guys want to implement this

ghost (commented):
Naming is suitable for me, it fits into our scheme, but the team has to make the decision in this matter for security reasons.

botder:
Personally, I would not add this to the default acl.xml

Dutchman101:
Thanks for the suggestion, but no

PR #3397, opened 2024-05-23
Fix permission check in File.create method

theSarrum:
As the title says, this PR fixes permission check in `File.create` method.


ghost (approved):
Looks good 🎉

Issue #460, opened 2018-09-13
Object sync


ArranTuna:
**Describe the solution you'd like**
@ccw808 said:
> Sync objects that move and or can be damaged.
> Object sync is currently disabled in the code because of issues. Someone should fix them and become a hero.

**Additional context**

> Issue #5886 was fixed by disabling a part of object sync
> https://code.google.com/p/mtasa-blue/source/detail?r=2846
> 
> Issue #6436 was fixed by disabling object sync completely
> https://code.google.com/p/mtasa-blue/source/detail?r=3119

From https://bugs.mtasa.com/view.php?id=8840

qaisjp:
r2846: https://github.com/multitheftauto/mtasa-blue/commit/eb30ddf1b0dc98f3339e4626c9290a177ac55f35

Could not recover r3119 (seems to be in branch 1.1.0 but that branch doesn't exist)... could only find the merge commit. See https://github.com/multitheftauto/mtasa-blue/commit/043acda9296c16c8c14e77ff964d059edcbecfe3

Other information about r3119 that might help find the commit: http://pastebin.com/c62u5b0M

lopezloo:
Issue 5886: http://web.archive.org/web/20200923182852/https://bugs.mtasa.com/view.php?id=5886 (`[Editor] Objects' positions gets resetted and setted when moving the camera`)
Issue 6436: http://web.archive.org/web/20160911210356/https://bugs.mtasa.com/view.php?id=6436 (`bugs in the maps object streamer server side`)

PlatinMTA:
> Issue 5886: http://web.archive.org/web/20200923182852/https://bugs.mtasa.com/view.php?id=5886 (`[Editor] Objects' positions gets resetted and setted when moving the camera`)

I can confirm that one

Issue #2752, opened 2022-09-18
Dynamic objects have sync issues


PlatinMTA:
### Describe the bug

If you move a server-sided dynamic object and then you step on it, the server will think that you are on top of it but on it's original position. This will cause issues.

https://youtu.be/zwEvcKlGng0

0:06 - The player that gets on top of the dumpster is already badly synced, since his position is actually a little more to the left (because the dumpster on his client is a little to the left aswell). I understand that the actual position of the dumpster is not synced, but I can't get a grip of why would the player position be dependent on the others clients dumpster position.

The recording player tries to shoot the dumpster one, but he won't receive damage since, on his client, he is a little far away and bullets are passing in front of him.

1:06 - Now we have the dumpster player POV.

1:27 - He runs a serversided command which adds 0.25z to his position. Since the server thinks he is on top of the dumpster, he gets teleported.

1:38 - He runs another command which sets a marker on his position. Again, since the server thinks he is on top of the dumpster, the marker is created elsewhere.

``BandwidthReductionMode`` was set to ``medium``.

### Steps to reproduce

Create a server-sided dynamic object (in the video they used ``1344``). Move it, then step on it. Now, get your coordinates clientside and serverside, they will be different.

``run local x,y,z = getElementPosition(me) createObject(1344, x, y, z)``
``run iprint(getElementPosition(me))``
``crun iprint(getElementPosition(me))``

### Version

Client: Multi Theft Auto v1.5.9-release-21299
Server: Multi Theft Auto v1.5.9-release-21318

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

lopezloo:
Reproduced.

Better repro:

1. 
```lua
run me.position = Vector3(0, -2.5, 3) obj = Object(1344, 0, 0, 3)
```

2. Step on this object.
3.
```lua
run me.position
crun me.position
```
4.
```lua
run obj.velocity = Vector3(0, 0.15, 0.2)
```
5. Step on this object again.
6.
```lua
run me.position
crun me.position
```

I have set `BandwidthReductionMode` to `none`.

Related: #460

Issue #386, opened 2018-09-03
Sync broken objects


ArranTuna:
**Is your feature request related to a problem? Please describe.**

LooooP said:

> Broken objects aren't synchronized. This is bad for some maps as you can't pass certain sections. You can't pass some certain sections because the object is broken for you, but other players can pass peacefully because that object isn't showed as broken for them. 
>
> DDs with breakable objects too. For you, the object is broken, but for other player isn't, so he can come to you and break your object (because he will stay less in your vision because he is floating)

**Describe the solution you'd like**
Broken object sync

**Additional context**
I said:

> Object sync was created by iztas about 7 years ago but he went inactive and it was reverted as it wasn't completed.

From https://bugs.mtasa.com/view.php?id=5537

PR #3405, opened 2024-05-26
Object sync - fixes and improvements

FileEX:
I undertook the repair and improvement of object synchronization, which has been disabled since 2011.

I was unable to reproduce the bugs described in #460 

I realize that this is quite a big PR, so I will be very grateful for any kind of feedback and reliable tests to eliminate any bugs and to safely merge this PR. I tested everything myself on the local server and I don't have the opportunity to test it with other players, so I'm not sure if all the problems have been eliminated.

### Fixes
- Code refactoring
- Small changes
### New events
- Added ``onObjectDamage`` event based on synchronization
- Added ``onObjectBreak`` event based on synchronization
- Added ``onObjectMoveStart`` event
- Added ``onObjectMoveStop`` event based on synchronization
### Changes
- Added almost all objects functions to server-side: ``setObjectProperty``, ``getObjectProperty``
- Added ``getElementVelocity`` & ``getElementAngularVeloity`` functions for objects on the server-side
- Added ``isElementInWater`` function for objects (shared)
- Move ``eObjectProperty`` enum to the shared

Resolved issues: #460, #2752, #386 (probably)

Parts of this PR:
- #3450 ✓
- #3714 ✓


Server/mods/deathmatch/logic/CObjectSync.cpp
```diff
@@ -196,43 +193,106 @@ void CObjectSync::Packet_ObjectSync(CObjectSyncPacket& Packet)
...
+        if (pData->ucFlags & 0x80)
+            pObject->SetTurnMass(pData->fTurnMass);
+        if (pData->ucFlags & 0x100)
+            pObject->SetAirResistance(pData->fAirResistance);
+        if (pData->ucFlags & 0x200)
+            pObject->SetElasticity(pData->fElasticity);
+        if (pData->ucFlags & 0x400)
+            pObject->SetBuoyancyConstant(pData->fBuoyancyConstant);
+        if (pData->ucFlags & 0x800)
+            pObject->SetCenterOfMass(pData->vecCenterOfMass);
```
TheNormalnij: I think we don't need to sync static data

Server/mods/deathmatch/logic/CObjectSync.cpp
```diff
@@ -196,43 +193,106 @@ void CObjectSync::Packet_ObjectSync(CObjectSyncPacket& Packet)
...
+        // We check whether the object is marked as "moving" and whether it is actually still moving
+        if (pObject->m_bIsMoving && !pObject->IsMoving())
+        {
+            CLuaArguments Arguments;
+            pObject->CallEvent("onObjectMoveStop", Arguments);
 
-        // Tell everyone
-        m_pPlayerManager->BroadcastOnlyJoined(Packet, pPlayer);
+            pObject->m_bIsMoving = false;
+        }
```
TheNormalnij: No. This event is broken by design.
Objects without syncer won't trigger it. Real "stop time" and "trigger time" may differ significantly as a result of this.

Server/mods/deathmatch/logic/CObjectSync.cpp
FileEX: Data is only sent if it has changed. If the object is created on the server side, you must first get its properties, which are available on the client side, otherwise the properties on the server and client side would be different

Server/mods/deathmatch/logic/CObjectSync.cpp
FileEX: This is true, so an appropriate note should be added to the wiki about it (if it ever gets merged).

Events such as onPedDamage, onPedWasted, etc. will also not be triggered without a syncer. If you place a ped next to a burning vehicle and leave the server, the ped damage event will not be triggered

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: But why do we need that? 
If you create object on server, you should change properties on server.

FileEX: This is illogical. When creating an object on the server side, its properties will be default (-1, zero vector), and on the client side the properties will have some values.

When we create a vehicle on the server side, it also has handling, plate and other data just like on the client side. It's as if upgrades were only on the client side by default, and there were no upgrades on the server side because they had to be set on the server side as well.

This is strange behavior imo

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: > If you place a ped next to a burning vehicle and leave the server, the ped damage event will not be triggered

But a ped doesn't deal damage without a syncer, and an object stops without a syncer. Why do you need API with such bad behavior? I see no reason to use this event instead of setTimer.

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: > When creating an object on the server side, its properties will be default (-1, zero vector)

So server should have a config. All object properties should be available for a new object immediately.

> When we create a vehicle on the server side, it also has handling, plate and other data just like on the client side.

Server has configs for that

FileEX: If you know where and how to do it in a different way as you write, give me a hint and tell me

TheNormalnij: GTA has 160 physical groups. You can store them on server and add a hashmap `modelId` -> `groupId`
You can use `engineGetObjectGroupPhysicalProperty` and `engineGetModelPhysicalPropertiesGroup` to generate data.
Horewer, I have no idea what we should return for objects without a physical group. Does gta use some default parameters for them?

FileEX: This requires a lot of additional work for the same effect, which is not profitable. Why waste time doing something in a more complex way when the current way works well and doesn't cause problems? I see no reason to waste time doing the same thing in a more difficult way

Server/mods/deathmatch/logic/CObjectSync.cpp
FileEX: The server does not check whether the object has reached its destination, this happens on the client side. So you can't clearly say that an object stops on an empty server, because no object physically exists on an empty server. Its status is set at the time of joining the server in CEntityAddPacket.

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: No, your solution adds uninitialized state for server objects. So it's much complex, that the solution with serverside config

FileEX: This is normal behavior. The object has its properties set at the time of synchronization and this is normal, because on the client side if the object is not streamed, its properties are also default (-1, zero vector). The properties have the appropriate values ​​when the object is streamed, and if the object is streamed, there is a good chance that it is also synchronized. I don't understand why you should complicate such a trivial thing as object properties. If this is really such a bad way, then I will remove access to the server-side properties, because implementing the same in a more difficult way requires investigation and a lot of work for the same effect

FileEX:
Ok, this is quite a large PR, so I decided to split it into several smaller PRs to make the review easier.

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: This is not normal and unnecessary complex

Server/mods/deathmatch/logic/CObjectSync.cpp
TheNormalnij: > So you can't clearly say that an object stops on an empty server

You can check `IsMoving`, so server knows when an object moving and when it stops

FileEX: @TheNormalnij What do you think about the new logic? https://github.com/multitheftauto/mtasa-blue/pull/3405/commits/843af2d030042c413f6f8d855396fccf8359fd1c

Now the event is sent even on an empty server

TheNormalnij: Yeah, it's better

Server/mods/deathmatch/logic/CObjectSync.cpp
Dutchman101: > This is not normal and unnecessary complex

Adding feedback label due to this, cc @FileEX to please provide counterarguments or chat with eachother in dev discord about your POV's. Only after that comes to a conclusion, this can be marked as resolved.

FileEX: I will work on it. Now I'm breaking the entire PR into several smaller ones to make it easier to review. One of them is already ready and waiting to be merged #3450

lopezloo:
Maybe split this work into:
1. Health and broken state of object sync.
2. Position, velocity sync.
3. Server controlled object properties.

ghost:
Bump?

PR #3408, opened 2024-05-27
Add onObjectMoveStart and onObjectMoveStop Events for Server-Side

esmail9900:
This PR adds the onObjectMoveStart and onObjectMoveStop events for server-side, with support for cancelEvent.


FileEX:
The ``onObjectMoveStop`` event is only triggered using the ``stopObject`` function, but what if the object reaches its destination and stops?

I add these events in my last PR #3405

esmail9900:
> The `onObjectMoveStop` event is only triggered using the `stopObject` function, but what if the object reaches its destination and stops?
> 
> I add these events in my last PR #3405

oh im really sorry mb. im so dumb, i saw your pr yesterday and forgot 1 day later

PR #3402, opened 2024-05-25
Render stages for 3D primitives

tederis:
This PR is aiming to fix a series of problems related to the purity of color of 3D lines and primitives(see #3399 and the discussion below #3321). It adds a new argument `"stage"` that can hold three values: `"prefx"`, `"postfx"` and `"postgui"`. 

The syntax:
```Lua
bool dxDrawLine3D ( float startX, float startY, float startZ, float endX, float endY, float endZ [, int color = 0xFFFFFFFF, float width = 1.0, bool postGUI = false / string stage = "prefx,postfx,postgui" ] )
```
```Lua
bool dxDrawMaterialLine3D ( float startX, float startY, float startZ, float endX, float endY, float endZ, [ bool flipUV = false, ] element material, float width, [ int color = white, [ bool postGUI = false / string stage = "prefx,postfx,postgui", ] float faceTowardX, float faceTowardY, float faceTowardZ ] )
```
```Lua
bool dxDrawMaterialSectionLine3D ( float startX, float startY, float startZ, float endX, float endY, float endZ, float u, float v, float usize, float vsize, [ bool flipUV = false, ] element material, float width, [ int color = white, [ bool postGUI = false / string stage = "prefx,postfx,postgui", ] float faceTowardX, float faceTowardY, float faceTowardZ ] )
```
```Lua
bool dxDrawMaterialPrimitive3D ( primitiveType pType, mixed material, bool postGUI / string stage = "prefx,postfx,postgui", table vertex1 [, table vertex2, ...] )
```
```Lua
bool dxDrawPrimitive3D ( string primitiveType, bool postGUI / string stage = "prefx,postfx,postgui", table vertex1, table vertex2, table vertex3 [, table vertex4, ...] )
```

To make colors correct out of the box the stage `"postfx"` is used by default. For the sake of backward compatibility the `"postGUI"` argument was preserved. 

The meaning of render stages:
- prefx - Primitives are renderer before the color correction which leads to distorted primitive's colors. This render stage was known as `postGUI = false`.
- postfx - Primitives are renderer after the color correction. This is a new render stage that should be used in most situations as it fixes the colors distortion.
- postgui - Primitives are rendered after GUI. This render stage was known as `postGUI = true`.

Colors

The color that is used for these primitives: #4199d7.

PR #3412, opened 2024-05-29
Fix render states spillover

tederis:
Fixes the problem reported here: https://github.com/multitheftauto/mtasa-blue/pull/3402#issuecomment-2136843658


ghost (approved):
Looks good 🎉

botder (approved):

PR #3416, opened 2024-05-30
Update mbedTLS to 3.6.0

Lpsd:
https://github.com/Mbed-TLS/mbedtls/releases/tag/v3.6.0


botder:
Isn't this enough?
https://github.com/multitheftauto/mtasa-blue/blob/06d1f1daf55a97bddd2fdf2ce80ad4a01085fbb0/vendor/curl/lib/vtls/mbedtls.c#L125-L127
https://github.com/multitheftauto/mtasa-blue/blob/06d1f1daf55a97bddd2fdf2ce80ad4a01085fbb0/vendor/curl/lib/vtls/mbedtls.c#L514-L522

Lpsd:
Yeah, I've updated my initial message as it seems we don't need to make any further changes.

PR #3421, opened 2024-05-31
Add disable option for engineSetModelPhysicalPropertiesGroup

TheNormalnij:
`engineSetModelPhysicalPropertiesGroup( modelId, -1 )` disables model physic.
So you can use this id for builings


Fernando-A-Rocha:
So if you set model group to -1 on this door model it will no longer have the "door opening physics"?

image

TheNormalnij:
Yes, but only for new objects.

ghost (approved):
Looks good 🎉

Issue #472, opened 2018-09-16
setElementHealth in onClientPedDamage can cause crash


ArranTuna:
**Describe the bug**
setElementHealth inside onClientPedDamage can cause crash.

**To Reproduce**
```lua
function handleDamage(attacker,weapon,bodypart,loss)
    setElementHealth(source,1)
end
addEventHandler("onClientPedDamage",root,handleDamage)
```

**Expected behavior**
No crash.

**MTA Client (please complete the following information):**
v1.5.5-release-13192

**Additional context**
[client_1.5.5-release-13192.0.000_gtasa_00332b2a_5_QPrMb_0100007F_55F3_178_4564C_20180722_2146.dmp](https://bugs.mtasa.com/file_download.php?file_id=1423&type=bug)

From https://bugs.mtasa.com/view.php?id=8926

CrosRoad95:
it also happen for player,
for peds crash is caused in "RpAnimBlendClumpGetFirstAssociation" function because ped on damage is recreated - see CClientPed::ReCreateModel
possible solutions:
1. `v2 = *(_DWORD **)(&a1->object.type + *(_DWORD *)(0xFB18C562 - (loc_45BA91 + 1)));` - crash here ( in RpAnimBlendClumpGetFirstAssociation ), add additional check for nullptr because from what i see
2. fix recreateModel, probably something is missing, is being added after for example one frame
3. simplest: make setElementHealth in damage don't recreate model, or make setElementHealth not work in ped damage ( but i assume it can break something )

why do we recreate ped?
```
function handleDamage(attacker,weapon,bodypart,loss)
    setElementHealth(source,1)
end
addEventHandler("onClientPedDamage",root,handleDamage)
addEventHandler("onClientPlayerDamage", root, handleDamage)
```

TracerDS:
Cant reproduce in 1.6. Probably fixed?

ArranTuna:
Still crashed for me.

All I did was runcode that code then shot a ped and instantly crashed.

Version = 1.6-release-21890.0.000
Time = Mon Jul 24 16:40:59 2023
Module = C:\ProgramData\MTA San Andreas All\1.6\GTA San Andreas\gta_sa.exe
Code = 0xC0000005
Offset = 0x00332B2A

EAX=00000000  EBX=00000000  ECX=00000034  EDX=047ED0DC  ESI=118906F4
[...]

PR #3428, opened 2024-06-03
New function: `hasPlayerSpawned`

TracerDS:
Returns `true` when player has been spawned by using `spawnPlayer`. `false` otherwise


Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -731,6 +731,8 @@ void CClientPed::Spawn(const CVector& vecPosition, float fRotation, unsigned sho
     {
         SetFrozenWaitingForGroundToLoad(true);
         m_iLoadAllModelsCounter = 10;
+
+        static_cast<CClientPlayer&>(*this).SetPlayerSpawned(true);
```
tederis: I would avoid the expressions like this. Looks clumsy and violates the design principles. Have you thought about moving the entire method CClientPed::Spawn under the CClientPlayer class? Or

TracerDS: Cant move it to the player. Players inherit functions from peds and `Spawn` manage the ped's being alive. If I move it to CClientPlayer then I will break CPed

tederis: The only place from which this method is invoked is Packet_PlayerSpawn which on its own is dedicated for players only(not peds). It cannot break non-player peds.

TracerDS: > The only place from which this method is invoked is Packet_PlayerSpawn which on its own is dedicated for players only(not peds). It cannot break non-player peds.

Maybe instead of moving the entire method, I'll move only `SetPlayerSpawned` there
```cpp
pPlayer->Spawn(vecPosition, fRotation, usPlayerModelID, ucInterior);
pPlayer->SetPlayerSpawned(true);
```

tederis: That's better. But breaks the object's state coherency. Because someone in the future can invoke Spawn but forget to call SetPlayerSpawned and boom, object's state is invalid now.

TracerDS: Okay, I'll move it to `CClientPlayer`

tederis: I would suggest you to leave it as is. But after the merge create a new PR with the corresponding refactor.

TracerDS: If it wont cause any issues then its better to do it now instead of waiting for it 'til merge

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -465,4 +468,6 @@ class CPlayer final : public CPed, public CClient
 
     ushort  m_usPrevDimension;
     SString m_strQuitReasonForLog;
+
+    bool m_bHasPlayerSpawned;
```
tederis: There already exists field CPed::m_bSpawned on server side. I know it behaves a bit different, but maybe there could be a way to avoid that collision. Because two fields with almost the same name may be confusing.

Server/mods/deathmatch/logic/CMapManager.cpp
```diff
@@ -592,6 +592,7 @@ void CMapManager::SpawnPlayer(CPlayer& Player, const CVector& vecPosition, float
     }
 
     // Update the player data
+    Player.SetPlayerSpawned(true);
```
tederis: I see that SetPlayerSpawned is called only ones. Should't there be a place where you call `SetPlayerSpawned(false)`?

TracerDS: > I see that SetPlayerSpawned is called only ones. Should't there be a place where you call `SetPlayerSpawned(false)`?

You can't despawn a player

Server/mods/deathmatch/logic/CPlayer.h
TracerDS: > There already exists field CPed::m_bSpawned on server side. I know it behaves a bit different, but maybe there could be a way to avoid that collision. Because two fields with almost the same name may be confusing.

`m_bHasPlayerSpawned` will change to `false` when player dies. If I were to make these 2 variables into 1, I'd need to make a complete overhaul of the spawn system

Server/mods/deathmatch/logic/CMapManager.cpp
tederis: > > I see that SetPlayerSpawned is called only ones. Should't there be a place where you call `SetPlayerSpawned(false)`?
> 
> You can't despawn a player

What will happened when player is died then?

TracerDS: > What will happened when player is died then?

When player dies, `m_bHasPlayerSpawned` will change to `false` (indicating that the player has been despawned) but the player itself wont despawn, he will retain every property of a spawned player.
Its kinda confusing tbf. Dead players should be dead, not despawned.

TracerDS: Actually, maybe I should just update `SetSpawned` to only work for spawning player and retain `SetDead` for killing/reviving players/peds? That would make more sense, no?

tederis: > > What will happened when player is died then?
> 
> When player dies, `m_bHasPlayerSpawned` will change to `false` (indicating that the player has been despawned) but the player itself wont despawn, he will retain every property of a spawned player. Its kinda confusing tbf. Dead players should be dead, not despawned.

But how exactly the value of m_bHasPlayerSpawned can become false? There's no place where it's set to false in your code.

TracerDS: > > > What will happened when player is died then?
> > 
> > 
> > When player dies, `m_bHasPlayerSpawned` will change to `false` (indicating that the player has been despawned) but the player itself wont despawn, he will retain every property of a spawned player. Its kinda confusing tbf. Dead players should be dead, not despawned.
> 
> But how exactly the value of m_bHasPlayerSpawned can become false? There's no place where it's set to false in your code.

Oops, sorry. I meant `m_bSpawned` (the original variable that changes when player dies)

`m_bHasPlayerSpawned` should only be set to false when player joins and leaves.
Since there is no direct way to despawn a player, it will never be set to false once a player has been spawned.

tederis: > Actually, maybe I should just update `SetSpawned` to only work for spawning player and retain `SetDead` for killing/reviving players/peds? That would make more sense, no?

Yeah. I think `m_bSpawned` could be renamed to m_bDead (with the corresponding refactor, of course). When player connects m_bDead is false, when he's dead m_bDead is true (obviously), and when he's spawned after it m_bDead is false again. In this case we can avoid the collision and there will no any confusion.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -3735,8 +3734,8 @@ void CGame::Packet_VehicleTrailer(CVehicleTrailerPacket& Packet)
     CPlayer* pPlayer = Packet.GetSourcePlayer();
     if (pPlayer && pPlayer->IsJoined())
     {
-        // Spawned?
-        if (pPlayer->IsSpawned())
+        // Alive?
+        if (!pPlayer->IsDead())
```
tederis, line 3809: I see you has taken care of two call sites where IsSpawned was used. But there are more of them: 
spawned

TracerDS: I didnt know if I should replace them since these lines operate on alive peds

tederis: That's important because you have changed the meaning of `m_bSpawned`.

tederis: It's become complicated. What if you revert the last commit and just rename m_bHasPlayerSpawned with something like m_bAlive? Simple and no collisions.

TracerDS: > That's important because you have changed the meaning of `m_bSpawned`.

I can change `IsSpawned` to `IsDead`. Based on the usage of the `SetSpawned` before the PR, `SetSpawned` is the same as `SetIsDead(not IsDead())`

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4296,9 +4295,6 @@ void CGame::PlayerCompleteConnect(CPlayer* pPlayer)
 
     // Send him the join details
     pPlayer->Send(CPlayerConnectCompletePacket());
-
-    // The player is spawned when he's connected, just the Camera is not faded in/not targetting
-    pPlayer->SetSpawned(true);
```
tederis, line 4372: You have replaced IsSpawned with !IsDead(). But at this line you can see that they are not interchangeable. Because m_bIsDead is only set when player is wasted. But m_bSpawned is true when player's connection is completed, not only when he's spawned via spawnPlayer.

TracerDS: that line of code itself doesnt make sense. When player joins a server (lets say without any resources) then he only joined but he is not spawned. Only `spawnPlayer` will spawn a player. Currently, mta sets `m_bSpawned` to the opposite of `m_bIsDead` on every dying related event or function.

tederis: Okay. But it should be noted that these changes will break the backward compatibility. Lets take a look at `setElementHealth` for example. Currently `setElementHealth` is valid even if it's called before `spawnPlayer`. After your changes it will not be true anymore.

tederis: On the other hand, who the hell would want to use setElementHealth if player is not even spawned, right?

TracerDS: > On the other hand, who the hell would want to use setElementHealth if player is not even spawned, right?

It also *might potentially* fix #472 but I really doubt it.

Backwards compatibility wouldn't really be a problem here. A quick refactor of `setElementHealth` to return `false` when the player is not spawned

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.h
```diff
@@ -47,6 +47,7 @@ class CLuaPlayerDefs : public CLuaDefs
     LUA_DECLARE(GetPlayerVersion);
     LUA_DECLARE(GetPlayerACInfo);
     static unsigned int GetPlayerScriptDebugLevel(CPlayer* const player);
+    static bool HasPlayerSpawned(CPlayer* player);
```
FileEX: player can be const?

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -1459,6 +1460,11 @@ unsigned int CLuaPlayerDefs::GetPlayerScriptDebugLevel(CPlayer* const player)
     return player->GetScriptDebugLevel();
 }
 
+bool CLuaPlayerDefs::HasPlayerSpawned(CPlayer* player)
```
FileEX: player can be const?

Client/mods/deathmatch/logic/CClientPlayer.cpp
```diff
@@ -205,6 +205,61 @@ bool CClientPlayer::IsOnMyTeam(CClientPlayer* pPlayer)
...
+    }
+
+    // Remove any animation
+    KillAnimation();
+
+    // Give him the correct model
+    SetModel(usModel);
+
+    // Detach from any entities
+    AttachTo(NULL);
```
FileEX: nullptr?

Client/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -309,6 +310,13 @@ int CLuaPlayerDefs::GetPlayerWantedLevel(lua_State* luaVM)
     return 1;
 }
 
+
+bool CLuaPlayerDefs::HasPlayerSpawned(CClientPlayer* player)
```
FileEX: noexcept?

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
FileEX: noexcept?

FileEX (approved):

Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.h
TracerDS: ~~When I marked any pointer `const`, ArgumentParser rejected these functions and threw a compile time error~~

Right, thanks

Server/mods/deathmatch/logic/CGame.cpp
Dutchman101: I've unresolved this review because as it stands, it looks like the backwards compatibility concern hasn't been tackled.
`A quick refactor of setElementHealth to return false when the player is not spawned` (@TracerDS)
Then please open a PR to do that, which will be a blocker for this to resolve and merge.

@tederis Do you have anything to add?

Dutchman101: @TracerDS I see you're updating the PR, but not acknownledging the above review.
Can you clarify why, because otherwise merge isn't gonna happen.

TracerDS: > @TracerDS I see you're updating the PR, but not acknownledging the above review. Can you clarify why, because otherwise merge isn't gonna happen.

`setElementHealth` currently checks if player was spawned.
The value of `IsSpawned()` is the same as `!IsDead()`.
They were interchangeable. `m_bIsDead` is set to `false` in `CPed`'s constructor.
`m_bIsSpawned` is set to `true` **only** here.
So in current mta's code its always set to `true`.

In this PR you cannot set player's health when he is not connected to the server yet.
If it stayed the same then it might cause some issues.
Checking against `IsDead` instead fixes these issues and returns false when ped is dead/not spawned.
However, that would also mean that using `setElementHealth` on a dead ped will fail (return false). If that is a concern, I could remove the whole `if` statement here since its always set to `true` in previous code

Fernando-A-Rocha (approved):
Good job

TracerDS:

Issue #3435, opened 2024-06-07
Created markers has a mismatch collision 


Proxy-99:
### Describe the bug
types arrow,cylidner,ring, checkpoints
marker trigger before touch it 
col not matching the visuals 

see:
[https://youtu.be/53tw5ihuWzM](https://youtu.be/53tw5ihuWzM)

### Steps to reproduce

create a marker of type cylinder 
attach the marker to a onClientMarkerHit or onMarkerHit
event will trigger before reaching to the marker

### Version

1.6

### Additional context

client and server side are given same bug

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

Proxy-99:
related to #331

Issue #870, opened 2019-04-15
isElementWithinMarker returns false also in serverside


javinotfound:
**Describe the bug**
In some cases, isElementWithinMarker returns false. However, it should return true. This occurs with markers created serverside, and isElementWithinMarker created also serverside (I say that because it could be related with #516 

**To reproduce**
1. Change your MTA nickname to "Test"
2. Execute next code

function myTestBug()
	local player = getPlayerFromName("Test")
	local x,y,z = getElementPosition(player)
	local marker = createMarker( x,y,z, "cylinder", 20.0, 255, 0, 0, 255)
	if isElementWithinMarker(player, marker) then
		outputChatBox("Works fine!", player, 0, 255, 0)
	else
		outputChatBox("Fail detected!", player, 255, 0, 0)
	end
end

**Expected behaviour**
If isElementWithinMarker is working properly, you should see Works fine!, but in chatbox it showed Fail detected!

**Screenshots**
N/A

**Version**
Server 1.5.6 Linux
Client mtasa-1.5.6-rc-16588

**Additional context**
N/A

Issue #331, opened 2018-08-21
Marker size problem


xAbdulKariM:
hi
i am have problem with markers

the problem is if i am set marker size for example 100 and if i am  used isElementWithinMarker or "onMarkerHit" or "onMarkerLeave"

will isElementWithinMarker return true and i am not in it

or events will run if i am not in it

and i m made function for solved the problem and Which it

[https://wiki.multitheftauto.com/wiki/IsElementInMarker](url)

and this function

[https://wiki.multitheftauto.com/wiki/GetElementsInMarker](url)

and i'm putting this functions in  https://wiki.multitheftauto.com/wiki/Useful_Functions Someone delete pages

and this pic explain what i'm talking about

![image](https://user-images.githubusercontent.com/923242/50401698-816f1c00-0788-11e9-974f-44e1d37cf93c.png)

the code :

```
testMarker = createMarker ( 2492 , -1666 , 12.5 , "cylinder" , 100 , 255 , 255 , 255 , 120 )
setTimer( function( )
outputChatBox ( tostring ( isElementWithinMarker ( getElementsByType ( "player" ) [ 1 ] , testMarker ) ) )
end , 100 , 1 )
```

but if you used my function will the result is

![image](https://user-images.githubusercontent.com/923242/50401574-9dbe8900-0787-11e9-9c49-ee4f4ea1639d.png)

the code :

```
local testMarker = createMarker ( 2492 , -1666 , 12.5 , "cylinder" , 100 , 255 , 255 , 255 , 120 )
setTimer( function( )
outputChatBox ( tostring ( isElementInMarker ( getElementsByType ( "player" ) [ 1 ] , testMarker ) ) )
end , 100 , 1 )
```

i hope approval my functions

znvjder:
IsElementWithinMarker only works as a marker''ll be created and ped must be synced with client.

It works: TRUE
`local marker=createMarker( 2045.56,1533.65,10.67-0.99, "cylinder" , 10, 255, 0, 0, 120 )
local ped=createPed( 10, 2045.56,1533.65,12.67 )
setTimer( function( )
  outputChatBox( tostring( isElementWithinMarker( ped, marker ) ) )
end, 1000, 1 ) `

It doesn't work: FALSE
`local marker=createMarker( 2045.56,1533.65,10.67-0.99, "cylinder" , 10, 255, 0, 0, 120 )
local ped=createPed( 10, 2045.56,1533.65,12.67 )
[...]

PR #3436, opened 2024-06-07
Fix cylinder marker has a 2D colshape instead of 3D and Fix collision radius not matching visual

### LINKED ISSUE #516: Clientside cylinder marker has a 2D colshape instead of 3D
**Describe the bug**
Flying high above a **clientside** marker (with a jetpack or something) will trigger the marker, even when they are far above it.

Same if you create a marker on a bridge and walk under it.

Serverside markers don't have this problem.

**To Reproduce**
Unstated.

**Expected behavior**
Should only trigger if you walk into the marker.

**Additional context**

TAPL said:
> 
> The issue exists only on clientside `onClientMarkerHit` and `isElementWithinMarker`, doesn't matter which side the marker was created on.
> 
> Even if the marker was created serverside and attached to `onClientMarkerHit` the height still not checked, however `onMarkerHit` works perfectly. This applies to `isElementWithinMarker` as well.

From https://bugs.mtasa.com/view.php?id=7025

Proxy-99:
Fixes #516 #331 #3435 #870
using tubes for both client and server
by trying to adjust the size to fit the visual and focuses on improving accuracy 
this PR should improve isElementWithinMarker detect


tederis:
The spherical colshape requires a bit bigger radius than the circular one. Otherwise, there is a high chance of the miss in some of the edge conditions.

tederis:
My opinion is that instead of this PR it's better to extend the circle colshape and add function `setColCircleHeight` (as it has already done for polygons).

Proxy-99:
> The spherical colshape requires a bit bigger radius than the circular one. Otherwise, there is a high chance of the miss in some of the edge conditions.

the problem will need types of col which can fit into the drawn shapes correctly

and the ColCircle has no height limit it goes to infinity

see
#3435

tederis:
> > The spherical colshape requires a bit bigger radius than the circular one. Otherwise, there is a high chance of the miss in some of the edge conditions.
> 
> the problem will need types of col which can fit into the drawn shapes correctly of ring and arrow and cylinder
> 
> and the ColCircle has no height limit it goes to infinity
> 
> see #3435

Yeah, I see. But currently this PR is rather a temporary patch, not the comprehensive solution.

Proxy-99:
> > > The spherical colshape requires a bit bigger radius than the circular one. Otherwise, there is a high chance of the miss in some of the edge conditions.
> > 
> > 
> > the problem will need types of col which can fit into the drawn shapes correctly of ring and arrow and cylinder
> > and the ColCircle has no height limit it goes to infinity
> > see #3435
> 
> Yeah, I see. But currently this PR is rather a temporary patch, not the comprehensive solution.

I have make some testing I think it works but the problem the showcol will show the virual shape but I do not understand the code of draw
see [https://youtu.be/zl-qC0dzpps?si=yraYuZhTisqbsuv_](https://youtu.be/zl-qC0dzpps?si=yraYuZhTisqbsuv_)

Proxy-99:
@tederis I have used ColTube instead of of sphere, I did some math calculations to try to adjust the size 
so far I think it's good with sizes like 300 and 500 it still not perfect 
mta-screen_2024-06-10_22-23-26

tederis:
> @tederis I have used ColTube instead of of sphere, I did some math calculations to try to adjust the size so far I think it's good with sizes like 300 and 500 it still not perfect mta-screen_2024-06-10_22-23-26 mta-screen_2024-06-10_22-24-18

Looks pretty accurate. So maybe it's better to stick with tubes? Radiuses bigger than 100 are used rarely.

Proxy-99:
> > @tederis I have used ColTube instead of of sphere, I did some math calculations to try to adjust the size so far I think it's good with sizes like 300 and 500 it still not perfect mta-screen_2024-06-10_22-23-26 mta-screen_2024-06-10_22-24-18
> 
> Looks pretty accurate. So maybe it's better to stick with tubes? Radiuses bigger than 100 are used rarely.

https://github.com/multitheftauto/mtasa-blue/issues/331#issuecomment-831826963 
The radius adjustments works good with numbers like 5,17,20 etc but when passing float like 5,7, 30.2 it has few problems .

I can't making the collision matching the render visual with 100% accurate it's like 98% for intger numbers
float like 95% not always correct

I am not sure if there is such method to always fit them correctly seems the renderd visual not always increase with fixed value.

Proxy-99:
test video 
[https://youtu.be/XCtzKp5FHOM?si=vNxIUU42YOC5aWck](https://youtu.be/XCtzKp5FHOM?si=vNxIUU42YOC5aWck)

tederis:
I don't think that a collision must 100% match the corresponding visual part. As was mentioned above collisions are a bit larger in SA. So you can simplify the formula.

Fernando-A-Rocha (approved):

botder (approved):

PR #3437, opened 2024-06-07
New function fxCreateParticle

FileEX:
The new ``fxCreateParticle`` function allows you to create any particle with its own color and size. This can be useful for creating flares, toxic fumes, reward/prize effects, etc.

**Syntax**
```lua
bool fxCreateParticle(string particle, float posX, float posY, float posZ, float dirX, float dirY, float dirZ, float r, float g, float b, float a [, bool randomizeColors = false, int count = 1, float brightness = 1.0, float size = 0.3, bool randomSizes = false, float life = 1.0 ])
```

### Available particles:
- blood
- boat_splash
- bubble (only works in water)
- car_debris
- collision_smoke
- gunshell
- sand
- sand2
- huge_smoke
- smoke
- spark
- spark2
- splash
- wake
- water_splash
- wheel_dirt
- glass

**randomizeColors** - Specifies whether the color should be fixed (r,g,b) or randomly calculated for each particle based on the given color (this allows to create colorful effects).

**randomizeSizes** - Specifies whether all particles should be the same fixed size or each particle should have a random size.

### Example
Video: https://streamable.com/1iurzh


Client/sdk/game/Common.h
```diff
@@ -1599,3 +1599,24 @@ namespace eSoundEffectParams
         HIGH_FREQ_RT_RATIO,
     };
 }            // namespace eSoundEffectParams
+
+enum eFxParticleSystems
```
tederis: Better to stick to enum class

Client/sdk/game/CFx.h
```diff
@@ -34,4 +36,5 @@ class CFx
     virtual void TriggerWaterSplash(CVector& vecPosition) = 0;
     virtual void TriggerBulletSplash(CVector& vecPosition) = 0;
     virtual void TriggerFootSplash(CVector& vecPosition) = 0;
+    virtual void AddParticle(eFxParticleSystems eFxParticle, CVector& vecPosition, CVector& vecDirection, float fR, float fG, float fB, float fA, bool bRandomizeColors, int iCount, float fBrightness, float fSize, bool bRandomizeSizes, float fLife) = 0;
```
tederis: `const CVector& vecPosition` and `const CVector& vecDirection`. `int iCount` -> `std::uint32_t` or `uint32_t` (see which you prefer).

Client/mods/deathmatch/logic/luadefs/CLuaEffectDefs.h
```diff
@@ -37,4 +37,6 @@ class CLuaEffectDefs : public CLuaDefs
     LUA_DECLARE(GetEffectSpeed);
     LUA_DECLARE(SetEffectDensity);
     LUA_DECLARE(GetEffectDensity);
+
+    static bool FxCreateParticle(std::string strParticle, CVector vecPosition, CVector vecDirection, float fR, float fG, float fB, float fA, std::optional<bool> bRandomizeColors, std::optional<int> iCount, std::optional<float> fBrightness, std::optional<float> fSize, std::optional<bool> bRandomizeSizes, std::optional<float> fLife);
```
tederis: `std::optional<int> iCount `-> `std::uint32_t` or `uint32_t` (see which you prefer).

Client/game_sa/CFxSA.cpp
```diff
@@ -239,3 +239,93 @@ void CFxSA::TriggerFootSplash(CVector& vecPosition)
...
+        default:
+            fxParticleSystem = m_pInterface->m_fxSysBlood;
+    }
+
+    for (int i = 0; i < iCount; i++)
+    {
+        if (bRandomizeColors)
+        {
+            // 0x49EECB
+            fxPrt.m_color.red = (rand() % 10000) * 0.0001 * fR + 0.13;
```
tederis: Please, use floating point numbers instead of the doubles. For example: 0.0001 -> 0.0001f.

Client/game_sa/CFxSA.cpp
```diff
@@ -239,3 +239,93 @@ void CFxSA::TriggerFootSplash(CVector& vecPosition)
...
+            fxParticleSystem = m_pInterface->m_fxSysWheelDirt;
+            break;
+        case PRT_GLASS:
+            fxParticleSystem = m_pInterface->m_fxSysGlass;
+            break;
+        default:
+            fxParticleSystem = m_pInterface->m_fxSysBlood;
+    }
+
+    for (int i = 0; i < iCount; i++)
```
tederis: size_t

TracerDS (approved):
LGTM, good job 👏

TracerDS:
Would it be possible for you to sync `fxCreateParticle` to the server so that it would be a shared function or would that be too much of a hassle?

Client/game_sa/CFxSA.cpp
```diff
@@ -239,3 +239,93 @@ void CFxSA::TriggerFootSplash(CVector& vecPosition)
...
+            fxPrt.m_color.green = (rand() % 10000) * 0.0001 * fG + 0.12;
+            fxPrt.m_color.blue = (rand() % 10000) * 0.0001 * fB + 0.04;
+        }
+
+        if (bRandomizeSizes)
+            // 0x49EF21 - Calculate random size for each particle
+            fxPrt.m_fSize = (rand() % 10000) * 0.0001 * fSize + 0.3;
+
+        // 0x49EF4C - Calculate random direction for each particle
+        newDirection = CVector(vecDirection.fX * 4, vecDirection.fY * 4, vecDirection.fZ * 4);
```
tederis, line 323: Where did you find this algorithm? What does the multiplication by 4 means?

Client/mods/deathmatch/logic/luadefs/CLuaEffectDefs.cpp
```diff
@@ -639,3 +642,12 @@ int CLuaEffectDefs::SetEffectDensity(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+bool CLuaEffectDefs::FxCreateParticle(std::string strParticle, CVector vecPosition, CVector vecDirection, float fR, float fG, float fB, float fA, std::optional<bool> bRandomizeColors, std::optional<int> iCount, std::optional<float> fBrightness, std::optional<float> fSize, std::optional<bool> bRandomizeSizes, std::optional<float> fLife)
+{
+    eFxParticleSystems eParticleSystem;
+    if (!StringToEnum(strParticle, eParticleSystem))
```
tederis: Did you try to substitute eFxParticleSystems directly into the function's arguments?

FileEX:
> Would it be possible for you to sync `fxCreateParticle` to the server so that it would be a shared function or would that be too much of a hassle?

I can do it in a separate PR when this one gets merged

tederis:
> Would it be possible for you to sync `fxCreateParticle` to the server so that it would be a shared function or would that be too much of a hassle?

This is per-frame function like dxDrawLine and the like. It can only be client side.

FileEX:
> > Would it be possible for you to sync `fxCreateParticle` to the server so that it would be a shared function or would that be too much of a hassle?
> 
> This is per-frame function like dxDrawLine and the like. It can only be client side.

Theoretically this can be done as shared. It is enough to send a packet to the clients, but if particles are created frequently, this can be expensive

tederis:
> > Would it be possible for you to sync `fxCreateParticle` to the server so that it would be a shared function or would that be too much of a hassle?
> 
> This is per-frame function like dxDrawLine and the like. It can only be client side.

Well, it's actually not the same. But it's still a high frequency method.

Client/game_sa/CFxSA.cpp
tederis: Okay. I see that this logic is inherited from SA.

FileEX: I copied it from the original SA code, e.g. AddWood (0x49EE10).
image

tederis:
Maybe it's better to use ARGB color in the form of uint32_t number instead of separated floating points components?

FileEX:
> Maybe it's better to use ARGB color in the form of uint32_t number instead of separated floating points components?

What do you mean - tocolor? I don't sure if it makes any difference, but if you tell me how to reverse it to float (AddParticle takes float), I can do it.

ghost (approved):
Looks good 🎉

Dutchman101:
It's not a big change, and passed multiple code reviews. Suggestions were applied. Thanks

tederis:
> > Maybe it's better to use ARGB color in the form of uint32_t number instead of separated floating points components?
> 
> What do you mean - tocolor? I don't sure if it makes any difference, but if you tell me how to reverse it to float (AddParticle takes float), I can do it.



> > Maybe it's better to use ARGB color in the form of uint32_t number instead of separated floating points components?
> 
> What do you mean - tocolor? I don't sure if it makes any difference, but if you tell me how to reverse it to float (AddParticle takes float), I can do it.

Yeas, I meant that it could be better to choose the same semantics as we have already implemented in `dxDraw` functions. In addition it would allow to avoid constructions such as `fR/255`.
@FileEX

CrosRoad95:
> , but if you tell me how to reverse it to float (AddParticle takes float), I can do it.

It make difference, you have to sometimes use https://wiki.multitheftauto.com/wiki/Tocolor2rgba and sometimes not ... it should be just "uint color"

Einheit-101:
Small question, how hard would it be to implement speed argument for particles and should i create a feature request for it? Or just make setEffectSpeed compatible with this?

PR #3444, opened 2024-06-08
 Fix HUD size when playing on widescreen

### LINKED ISSUE #848: Widescreen Hud Placement
On MTA the original HUD takes too much of your screen, even if you turn on the option in settings that should adjust the proportions of it. It's just way too big, you can see in this thread how the HUD looks like on singleplayer or samp with widescreen fix made by thirteenag. https://forum.mtasa.com/topic/103042-widescreen-fix-for-mta/?_fromLogin=1

### LINKED ISSUE #2753: Ultrawidescreen (32:9) HUD and FOV issue
### Describe the bug

Camera FOV seems to be based off screen width instead of screen height, thus cropping view on the top and bottom when using a wider screen instead of extending the view on the sides.
HUD is also based off width to control the scaling, instead so instead of "extending" the HUD on the left and right of the screen while preserving the scaling of the height, it gets twice as big on a 32:9 compared to a 16:9.
Single player with ThirteenAG's wide screen fix doesn't have this issue. 
Using the FOV slider doesn't improve the situation, as it is clamped to 90° (horizontal), so to have a the same 59° of vertical FOV (for a 90° horizontal FOV on 16:9), it would need to be set at 127° (horizontal) on a 32:9.

Comparison:
GTA SA solo ultrawidescreen
MTA SA ultrawidescreen

### Steps to reproduce

1. Launch the game
2. Set the video resolution to a 32:9 aspect
3. Reboot the game to apply the resolution change
4. Launch a server or map editor
5. Observe the low FOV and big HUD.

### Version

Client: version 1.5.9-release-21324

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

henrymarques:
Fix #848, maybe fix #2753
There's nothing new, most part of the code is from ThirteenAG's widescreen fix (thanks).

This is a draft pull request because it needs more testing. If you can test it, please, send me feedback.
I need feedback about the code too.


Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+struct RsGlobal
```
FileEX: In the context of this PR, some fields are not used (frameLimit, quit, ps). In addition, this structure is incomplete relative to the SA code. If we do not need the remaining fields, use only those that are necessary.

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        0x582DC1,            // 14 Legend
+        0x582ED3,            // 15 Legend
+        0x582FA3,            // 16 Legend
+        0x582FF5,            // 17 Legend
+        0x583079,            // 18 Legend
+        0x5830F6,            // 19 Legend
+        0x58317B,            // 20 Legend
+        0x576119,            // 21 Legend
+        0x575EB7,            // 22 Map zones
+        NULL,                // 23
```
FileEX: What is the purpose of these empty indexes in arrays?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        0x719C58,            // 23 Font fixes
+        0x728864,            // 24 BarChart fixes
+        0x7288A9,            // 25 BarChart fixes
+        0x573F7D,            // 26 DrawWindow header
+        0x573FD6,            // 27 DrawWindow header
+    };
+
+    for (int i = 0; i < sizeof(m_dwMiscWidth) / sizeof(const void*); i++)
+    {
+        if (m_dwMiscWidth[i] != NULL)
```
FileEX: Why are there NULLs in the array that are checked here?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+float fHUDWidth[120], fHUDHeight[120];
+float fCameraWidth[2], fCameraHeight[2];
+
+float fWideScreenWidthScale = 1.0f;
+float fWideScreenHeightScale = 1.0f;
+
+static const float fHudHeightScale = 1.0f;
+static const float fHudWidthScale = 1.0f;
+static const float fRadarWidthScale = 1.0f;
+static const float fRadarHeightScale = 1.0f;
```
FileEX: constexpr will be better imo

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    fHUDHeight[16] /= 1.125f;
+}
+
+void UpdateScreenAspectRatio()
+{
+    fWideScreenWidthScale = 640.0f / (*(pfScreenAspectRatio) * 448.0f);
+    fWideScreenHeightScale = 448.0 / 448.0f;
+    fWideScreenWidthProperScale = static_cast<float>(pRsGlobal->width) / (*pfScreenAspectRatio * 448.0f);
+    fWideScreenHeightProperScale = static_cast<float>(pRsGlobal->height) / 448.0f;
+    fDefaultWidth = (4.0f / 3.0f);
```
FileEX: This variable can be defined at the top as a constant

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    fMiscHeight[28] = 0.002232143f * fWideScreenHeightScale * fHudHeightScale;
+}
+
+void UpdateScriptFixes()
+{
+    float w;
+    if (*pfScreenAspectRatio < fDefaultWidth)
+        w = static_cast<float>(pRsGlobal->width);
+    else
+        w = pRsGlobal->width * fDefaultWidth / *pfScreenAspectRatio;
```
FileEX: You can use the conditional operator (? :) here for better readability and make "w" as constant

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    fRadarWidth[1] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[2] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[3] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[4] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[5] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[6] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[7] = 0.0015625f * fWideScreenWidthScale * fRadarWidthScale;
+    fRadarWidth[8] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fRadarWidth[9] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fRadarWidth[10] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
```
FileEX: You can declare these magic numbers (0.0015625f, 0.0015625f, 0.002232143f) etc. as a constant variable

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        return (float)pRsGlobal->width * fValue * 0.0015625f;
+}
+
+float __stdcall StretchYHook(float fValue)
+{
+    if (pRsGlobal->height == 448)
+        return fValue;
+    else
+        return (float)pRsGlobal->height * fValue * 0.002232143f;
+}
```
FileEX: You can use the conditional operator (? :) here for better readability. What do you think about static_cast here?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        NULL,                // 33
+        NULL,                // 34
+        NULL,                // 35
+        NULL,                // 36
+        NULL,                // 37
+        0x5849FA,            // 38 DrawYouAreHereSprite
+        0x574096,            // 39 Message screen
+    };
+
+    for (int i = 0; i < sizeof(m_dwFrontendWidth) / sizeof(const void*); i++)
```
FileEX: size_t and you can use std::size. instead of sizeof. You also don't need sizeof(const void*)

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    {
+        if (m_dwFrontendWidth[i] != NULL)
+            MemPut<const void*>(m_dwFrontendWidth[i] + 0x2, &fFrontendWidth[i]);
+    }
+
+    for (int i = 0; i < sizeof(m_dwFrontendHeight) / sizeof(const void*); i++)
```
FileEX: Above

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        NULL,                // 36
+        NULL,                // 37
+        0x5849FA,            // 38 DrawYouAreHereSprite
+        0x574096,            // 39 Message screen
+    };
+
+    for (int i = 0; i < sizeof(m_dwFrontendWidth) / sizeof(const void*); i++)
+    {
+        if (m_dwFrontendWidth[i] != NULL)
+            MemPut<const void*>(m_dwFrontendWidth[i] + 0x2, &fFrontendWidth[i]);
```
FileEX, line 680: The pointer type here should be float because the values ​​at these addresses are of type float. In addition, why do you pass the address of the value to memory, instead of the values ​​themselves?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        if (m_dwFrontendHeight[i] != NULL)
+            MemPut<const void*>(m_dwFrontendHeight[i] + 0x2, &fFrontendHeight[i]);
+    }
+
+    // StretchXY Restoration
+    int m_dwCallWidth[] = {
+        0x57E3A5, 0x577CB7, 0x577CD0, 0x577D6F, 0x577D99, 0x577DAB, 0x577FA6, 0x578031, 0x578052, 0x5780E1, 0x5788DF, 0x5788FB,
+    };
+
+    int m_dwCallHeight[] = {
```
FileEX: If the array size is constant, it is worth specifying it

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    }
+
+    for (int i = 0; i < sizeof(m_dwFrontendHeight) / sizeof(const void*); i++)
+    {
+        if (m_dwFrontendHeight[i] != NULL)
+            MemPut<const void*>(m_dwFrontendHeight[i] + 0x2, &fFrontendHeight[i]);
+    }
+
+    // StretchXY Restoration
+    int m_dwCallWidth[] = {
```
FileEX: uintptr_t?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    {
+        if (m_dwCrosshairWidth[i] != NULL)
+            MemPut<const void*>(m_dwCrosshairWidth[i] + 0x2, &fMiscWidth[0]);
+    }
+
+    for (int i = 0; i < sizeof(m_dwCrosshairHeight) / sizeof(const void*); i++)
+    {
+        if (m_dwCrosshairHeight[i] != NULL)
+            MemPut<const void*>(m_dwCrosshairHeight[i] + 0x2, &fMiscHeight[0]);
+    }
```
FileEX: Same as above

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    {
+        if (m_dwRadarWidth[i] != NULL)
+            MemPut<const void*>(m_dwRadarWidth[i] + 0x2, &fRadarWidth[i]);
+    }
+
+    for (int i = 0; i < sizeof(m_dwRadarHeight) / sizeof(const void*); i++)
+    {
+        if (m_dwRadarHeight[i] != NULL)
+            MemPut<const void*>(m_dwRadarHeight[i] + 0x2, &fRadarHeight[i]);
+    }
```
FileEX: Same

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    // Lock Subtitles Width
+    static float fSubtitlesMult = 1.0f;
+    MemPut<const void*>(0x58C4E8 + 0x2, &fSubtitlesMult);
+    */
+
+    // Second player fix.
+    MemPut<const void*>(0x58F9A0 + 0x2, &fHUDWidth[110]);            // Weapon icon X
+    MemPut<const void*>(0x58F993 + 0x2, &fHUDWidth[16]);             // Weapon icon X
+    MemPut<const void*>(0x58F972 + 0x2, &fHUDHeight[16]);            // Weapon icon Y
+    MemPut<const void*>(0x58FA8E + 0x2, &fHUDWidth[17]);             // Ammo x
```
FileEX, line 1100: Use type float. Why do you put the addresses of the values ​​in memory instead of the values ​​themselves directly?

FileEX (changes requested):
It's nice that you are willing to do it. There are still a lot of things to improve, better readability, etc. Personally, I don't like these huge address tables. 

Keep it up!

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: I mean fDefaultWidth

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+        0x57A3EB,            // 13 Menu text
+        0x582DDB,            // 14 Legend
+        0x582E0D,            // 15 Legend
+        0x582EB0,            // 16 Legend
+        0x583019,            // 17 Legend
+        0x58309C,            // 18 Legend
+        0x582F82,            // 19 Legend
+        0x583128,            // 20 Legend
+        0x5831AD,            // 21 Legend
+        0x57613B,            // 22 Legend
```
FileEX: I don't think we need to apply the fix for menu, legend, subtitles, wasted, busted etc

FileEX (changes requested):

henrymarques:
I've read all the reviews, I will reply and work on the changes when I have a break

Fernando-A-Rocha:
thank you for your work, i hope this is finished :D 

hud looks like this without your PR, stretched and too big 😔 
image

Fernando-A-Rocha:
vs on samp with widescreen hud fix

image

DomenicF:
Thanks for working on this!

Just curious if this fix also covers FoV issues (reported in #2753) or if the fix would be a totally separate thing?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: honestly, I've kept this because I was not sure if it would work without this fields, I took it off and it works

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: changed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: removed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: removed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: changed and removed static

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: accepted

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: it will be not used anymore, static cast would be better tho

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: removed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: These magic numbers are also magic to me, I will try but there's a lot through the code

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: reviewing the code I realize this function is not used

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: Therefore, you do not need to document them. Just list the ones that are repeated many times as variables like "RadarWidth_Mult" etc

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: Static here makes sense, so don't remove it. Just change const to constexpr

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: agreed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: changed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+void UpdateMiscFixes();
+void UpdateHUDFixes();
+void UpdateScriptFixes();
+
+float AdjustFOV(float f, float ar)
+{
+    return std::round((2.0f * atan(((ar) / (4.0f / 3.0f)) * tan(f / 2.0f * ((float)M_PI / 180.0f)))) * (180.0f / (float)M_PI) * 100.0f) / 100.0f;
+}
+
+void __declspec(naked) CalculateAimingPoint()
```
FileEX: Unused

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+
+void InstallAspectRatioFixes();
+void InstallFrontendFixes();
+void InstallMiscFixes();
+void InstallHUDFixes();
+
+void UpdateFrontendFixes();
+void UpdateMiscFixes();
+void UpdateHUDFixes();
+void UpdateScriptFixes();
```
FileEX: These definitions can be removed because the functions are declared below

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,1111 @@
...
+    fHUDWidth[12] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[13] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[14] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[15] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[16] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[17] = 0.17343046f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[18] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[19] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[20] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
+    fHUDWidth[21] = 0.0015625f * fWideScreenWidthScale * fHudWidthScale;
```
FileEX: According to the coding guidelines, we no longer use Hungarian notation

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: I just put it back. Could you say more about `static`? it's lifecycle thing, I guess, but I don't get it

FileEX: The static keyword restricts the visibility of a variable or function to the source file and ensures that the variable is allocated at program start and remains in memory for the program's entire duration

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: those that repeated many times became two variables, WidthMult and HeightMult, the remaining still being a magic number on the code because they do not repeat more than twice

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: thanks

henrymarques:
> Just curious if this fix also covers FoV issues [...] or if the fix would be a totally separate thing?

Not yet, it's not totally separate thing but I think I will open another pull request to this

henrymarques:
Everything seems to work correctly, but I would like if another dev could test this

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x589D61,            // 45 Stats box
+        0x589E49,            // 46 Stats box
+        0x589F31,            // 47 Stats box
+        0x58A013,            // 48 Stats box
+        0x58A090,            // 49 Stats box
+        0x58A134,            // 50 Stats box
+        0x58C863,            // 51 SuccessFailed text
+        0x58D2DB,            // 52 MissionTitle text
+        0x58D459,            // 53 MissionTitle text
+        0x58CBC1,            // 54 WastedBusted text
```
FileEX: Stats box, mission title, wasted, busted, helptext are unused in MTA. Probably things like timers, menu system, OddJob, TripleSkip, CDarkel, CGarages are also unused in MTA. You should go through these addresses and leave only what makes sense for the MTA

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x5810EF,            // 74
+        0x581158,            // 75
+        0x5811CD,            // 76
+        0x58148A,            // 77
+        0x5814F7,            // 78
+        0x5815B1,            // 79
+        0x5815EB,            // 80
+        0x581633,            // 81
+        0x47AD2A,            // 82
+        0x5818CF,            // 83
```
FileEX: Undocumented

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x58BFCB,            // 64 Help text
+        0x580E11,            // 65 Menu system
+        0x580F85,            // 66
+        0x5810CC,            // 67
+        0x581132,            // 68
+        0x5811A1,            // 69
+        0x58147A,            // 70
+        0x5814E7,            // 71
+        0x581699,            // 72
+        0x581889,            // 73
```
FileEX: Undocumented

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x58B263,            // 56 Timers
+        0x58B5DE,            // 57 Timers
+        0x58B7BD,            // 58 Help text
+        0x58BA4C,            // 59 Help text
+        0x58BBA7,            // 60 Help text
+        0x58BD19,            // 61 Help text
+        0x58BE2B,            // 62 Help text
+        0x58BF1C,            // 63 Help text
+        0x58BFCB,            // 64 Help text
+        0x580E11,            // 65 Menu system
```
FileEX: Same as above

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x58A9BD,            // Radar point
+        0x5834EC,            // Radar point
+        0x586058,            // Radar centre
+        0x584346,            // Radar Trace 0
+        0x58440C,            // Radar Trace 0
+        0x58412B,            // Radar Trace 2
+        0x5841B0,            // Radar Trace 2
+        0x584207,            // Radar Trace 1
+        0x5842C6,            // Radar Trace 1
+        0x5876BC, 0x587733, 0x587916, 0x587A02, 0x587A92,
```
FileEX: Uncodumented

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+        0x5834BA,                      // Radar point
+        0x58603F,                      // Radar point
+        0x5886CC,                      // Radar centre
+        0x58439C,                      // Radar Trace 0
+        0x584434,                      // Radar Trace 0
+        0x58410B,                      // Radar Trace 2
+        0x584190,                      // Radar Trace 2
+        0x584249,                      // Radar Trace 1
+        0x5842E6,                      // Radar Trace 1
+        0x5876D4, 0x58774B, 0x58780A, 0x58788F, 0x58792E, 0x587A1A, 0x587AAA,
```
FileEX: Uncodumented

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,730 @@
...
+static constexpr float HudWidthScale = 1.0f;
+static constexpr float HudHeightScale = 1.0f;
+static constexpr float RadarWidthScale = 1.0f;
+static constexpr float RadarHeightScale = 1.0f;
+
+constexpr float WidthMult = 0.0015625f;
+constexpr float HeightMult = 0.002232143f;
+
+float* ScreenAspectRatio;
+float* ScreenFieldOfView;
```
FileEX: Unused variable - float* ScreenFieldOfView;

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: removed

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: added

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: added

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: unused

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: unused

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: unused

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: changed

henrymarques:
At this point, HUD size can be changed back switching aspect ratio to 4:3 on video settings, I don't know if this setting change could have side effects elsewhere than HUD. If this happen, please tell me, so I will think how to solve with "HUD match aspect ratio" setting checkbox. I will work on fov fix now

EDIT:
> [...] I don't know if this setting change could have side effects elsewhere than HUD.

It has, but I found a way to solve using "HUD match aspect ratio" checkbox. I don't think it's the better way but it works, also, unfortunetly the game needs to restart to apply changes. I can commit it as is to this PR if someone thinks it is needed

Fernando-A-Rocha:
Related to #3582 ?

henrymarques:
> Related to #3582 ?

You can see the changes [here](https://github.com/henrymarques/mtasa-blue/commit/4254ce23faccf7ea85fcc094b0573699465ee329), I don't think it's the better way to do this, so I didn't include in this PR

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+static constexpr float ScreenHeightScale = 1.0f;
+
+static constexpr float HudWidthScale = 1.0f;
+static constexpr float HudHeightScale = 1.0f;
+static constexpr float RadarWidthScale = 1.0f;
+static constexpr float RadarHeightScale = 1.0f;
+/* NOTE: Changing hud and radar scales may misalign elements */
+
+constexpr float WidthMult = 0.0015625f;
+constexpr float HeightMult = 0.002232143f;
```
FileEX: static?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
+float MiscWidth[30], MiscHeight[30];
+float RadarWidth[23], RadarHeight[23];
+float HUDWidth[39], HUDHeight[34];
+float CameraWidth[1];
+
+float                  ScreenWidthScale = 1.0f;
```
FileEX: static?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+#include "StdInc.h"
+
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
+float MiscWidth[30], MiscHeight[30];
+float RadarWidth[23], RadarHeight[23];
+float HUDWidth[39], HUDHeight[34];
+float CameraWidth[1];
```
FileEX: static?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    RadarWidth[13] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[14] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[15] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[16] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[17] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[18] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[19] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[20] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[21] = WidthMult * ScreenWidthScale * HudWidthScale;
+    RadarWidth[22] = WidthMult * ScreenWidthScale * HudWidthScale;
```
FileEX:
```cpp
for (std::size_t i = 0; i < std::size(RadarWidth); i++)
    RadarWidth[i] = (i == 6 ? 0.0015000f : WidthMult) * ScreenWidthScale * HudWidthScale
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    RadarHeight[12] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[13] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[14] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[15] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[16] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[17] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[18] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[19] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[20] = HeightMult * ScreenHeightScale * HudHeightScale;
+    RadarHeight[21] = HeightMult * ScreenHeightScale * HudHeightScale;
```
FileEX:
```cpp
for (std::size_t i = 0; i < std::size(RadarHeight); i++)
    RadarHeight[i] = HeightMult * ScreenHeightScale * HudHeightScale;
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    HUDWidth[30] = 0.0019350f * ScreenWidthScale;
+    HUDWidth[31] = WidthMult * ScreenWidthScale * HudWidthScale;
+    HUDWidth[32] = WidthMult * ScreenWidthScale * HudWidthScale;
+    HUDWidth[33] = 0.0019500f * ScreenWidthScale;
+    HUDWidth[34] = WidthMult * ScreenWidthScale * HudWidthScale;
+    HUDWidth[35] = WidthMult * ScreenWidthScale * HudWidthScale;
+    HUDWidth[36] = WidthMult * ScreenWidthScale * HudWidthScale;
+    HUDWidth[37] = WidthMult * ScreenWidthScale * HudWidthScale;
+
+    HUDWidth[38] = 1920.0f * 0.17343046f * ScreenWidthScale * HudWidthScale;
```
FileEX:
```cpp
for (std::size_t i = 0; i < std::size(HUDWidth); i++)
{
    if (i < 38)
    {
        float mult = (i == 17 || i == 22) ? 0.17343046f : ((i == 30 || i == 33) ? 0.0019500f : WidthMult);
        HUDWidth[i] = mult * ScreenWidthScale * HudWidthScale;
    }
    else
        HUDWidth[i] = 1920.0f * 0.17343046f * ScreenWidthScale * HudWidthScale;
}
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    HUDHeight[24] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[25] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[26] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[27] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[28] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[29] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[30] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[31] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[32] = HeightMult * ScreenHeightScale * HudHeightScale;
+    HUDHeight[33] = HeightMult * ScreenHeightScale * HudHeightScale;
```
FileEX:
```cpp
for (std::size_t i = 0; i < std::size(HUDHeight); i++)
    HUDHeight[i] = (i == 18 ? 0.002f : HeightMult) * ScreenHeightScale * HudHeightScale;
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    MemPut<BYTE>(0x713B6D, 0x38);
+    MemPut<BYTE>(0x713CFB, 0x38);
+    MemPut<BYTE>(0x713EFC, 0x30);
+    MemPut<BYTE>(0x714004, 0x38);
+
+    HookInstallCall(0x0053D695 - 0x1, (DWORD)UpdateScreenAspectRatio);
+    HookInstallCall(0x0053D7B2 - 0x1, (DWORD)UpdateScreenAspectRatio);
+    HookInstallCall(0x0053D967 - 0x1, (DWORD)UpdateScreenAspectRatio);
+    HookInstallCall(0x0053E771 - 0x1, (DWORD)UpdateScreenAspectRatio);
+    HookInstallCall(0x0053EB1A - 0x1, (DWORD)UpdateScreenAspectRatio);
```
FileEX:
```cpp
    DWORD callFunc = reinterpret_cast<DWORD>(UpdateScreenAspectRatio);

    HookInstallCall(0x0053D694, callFunc);
    HookInstallCall(0x0053D7B1, callFunc);
    HookInstallCall(0x0053D966, callFunc);
    HookInstallCall(0x0053E770, callFunc);
    HookInstallCall(0x0053EB19, callFunc);
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+}
+
+void InstallHUDFixes()
+{
+    uintptr_t crosshairWidthAddresses[10] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+        0x58E7CE, 0x58E7F8, 0x58E2FA, 0x58E4ED, 0x58E75B, 0x58E28B, 0x58E2AC, 0x58E2BA, 0x53E472, 0x53E4AE,
+    };
+
+    uintptr_t crosshairHeightAddresses[7] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    {
+        MemPut<float*>(crosshairWidthAddresses[i] + 0x2, &MiscWidth[0]);
+    }
+
+    for (size_t i = 0; i < std::size(crosshairHeightAddresses); i++)
+    {
+        MemPut<float*>(crosshairHeightAddresses[i] + 0x2, &MiscHeight[0]);
+    }
+
+    uintptr_t radarWidthAddresses[23] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+        0x58439C,                                                                        // Radar Trace 0
+        0x584434,                                                                        // Radar Trace 0
+        0x58410B,                                                                        // Radar Trace 2
+        0x584190,                                                                        // Radar Trace 2
+        0x584249,                                                                        // Radar Trace 1
+        0x5842E6,                                                                        // Radar Trace 1
+        0x5876D4, 0x58774B, 0x58780A, 0x58788F, 0x58792E, 0x587A1A, 0x587AAA,            // Radar blips
+    };
+
+    uintptr_t radarHeightAddresses[21] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    {
+        MemPut<float*>(radarWidthAddresses[i] + 0x2, &RadarWidth[i]);
+    }
+
+    for (size_t i = 0; i < std::size(radarHeightAddresses); i++)
+    {
+        MemPut<float*>(radarHeightAddresses[i] + 0x2, &RadarHeight[i]);
+    }
+
+    uintptr_t hudWidthAddresses[38] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+        0x58AD3A,            // 31 Area names scale
+        0x58AD65,            // 32 Area names ?
+        0x58AE4A,            // 33 Area names x pos
+        0x58C395,            // 34 Subs
+        0x58C41D,            // 35 Subs
+        0x58C4DC,            // 36 Subs
+        0x4E9F30,            // 37 RadioStation
+    };
+
+    uintptr_t hudHeightAddresses[34] = {
```
FileEX: std::uintptr_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+        0x58C4C6,            // 32 Subs
+        0x4E9F1A,            // 33 RadioStation
+    };
+
+    for (size_t i = 0; i < std::size(hudWidthAddresses); i++)
+    {
+        MemPut<float*>(hudWidthAddresses[i] + 0x2, &HUDWidth[i]);
+    }
+
+    for (size_t i = 0; i < std::size(hudHeightAddresses); i++)
```
FileEX: size_t -> std::size_t

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    MemSet((void*)0x50AD79, 0x90, 6);
+    MemPut<float*>(0x50AD59 + 0x2, &CameraWidth[0]);
+    MemPut<float*>(0x51498D + 0x2, &CameraWidth[0]);
+}
+
+void GetMemoryAddresses()
+{
+    ScreenAspectRatio = (float*)0xC3EFA4;
+    ScreenFieldOfView = (float*)0x8D5038;
+}
```
FileEX: You can remove this function and set the values ​​at the top of the file as static

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // Second player fix.
+    MemPut<float*>(0x58F9A0 + 0x2, &HUDWidth[38]);             // Weapon icon X
+    MemPut<float*>(0x58F993 + 0x2, &HUDWidth[16]);             // Weapon icon X
+    MemPut<float*>(0x58F972 + 0x2, &HUDHeight[16]);            // Weapon icon Y
+    MemPut<float*>(0x58FA8E + 0x2, &HUDWidth[17]);             // Ammo x
+}
+
+void InstallFieldOfViewFixes()
+{
+    HookInstall(0x6FF410, (DWORD)SetFOV, 5);
```
FileEX: Use reinterpret_cast

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    // Aiming point
+    HookInstall(0x51499E, (DWORD)CalculateAimingPoint, 5);
```
FileEX: Use reinterpret_cast

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    MemSet((void*)0x50AD79, 0x90, 6);
+    MemPut<float*>(0x50AD59 + 0x2, &CameraWidth[0]);
+    MemPut<float*>(0x51498D + 0x2, &CameraWidth[0]);
+}
+
+void GetMemoryAddresses()
+{
+    ScreenAspectRatio = (float*)0xC3EFA4;
+    ScreenFieldOfView = (float*)0x8D5038;
```
FileEX: Use reinterpret_cast

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+}
+
+void CMultiplayerSA::InitHooks_WidescreenFix()
```
FileEX, line 84: static?

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+}
+
+void InstallFieldOfViewFixes()
+{
+    HookInstall(0x6FF410, (DWORD)SetFOV, 5);
+    // Aiming point
+    HookInstall(0x51499E, (DWORD)CalculateAimingPoint, 5);
+    MemSet((void*)0x50AD79, 0x90, 6);
+    MemPut<float*>(0x50AD59 + 0x2, &CameraWidth[0]);
+    MemPut<float*>(0x51498D + 0x2, &CameraWidth[0]);
```
FileEX: 0x50AD59 + 0x2 = 0x50AD5B
0x51498D + 0x2 = 0x51498F

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    MemPut<float*>(0x58FA8E + 0x2, &HUDWidth[17]);             // Ammo x
+}
+
+void InstallFieldOfViewFixes()
+{
+    HookInstall(0x6FF410, (DWORD)SetFOV, 5);
+    // Aiming point
+    HookInstall(0x51499E, (DWORD)CalculateAimingPoint, 5);
+    MemSet((void*)0x50AD79, 0x90, 6);
+    MemPut<float*>(0x50AD59 + 0x2, &CameraWidth[0]);
```
FileEX: Add a comment what address this is. 
``CCamera::Find3rdPersonQuickAimPitch``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: Add a comment what address this is.
``CCamera::Find3rdPersonCamTargetVector``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    MemPut<float*>(0x58F972 + 0x2, &HUDHeight[16]);            // Weapon icon Y
+    MemPut<float*>(0x58FA8E + 0x2, &HUDWidth[17]);             // Ammo x
+}
+
+void InstallFieldOfViewFixes()
+{
+    HookInstall(0x6FF410, (DWORD)SetFOV, 5);
+    // Aiming point
+    HookInstall(0x51499E, (DWORD)CalculateAimingPoint, 5);
+    MemSet((void*)0x50AD79, 0x90, 6);
```
FileEX, line 79: Add a comment that skip division by ``CDraw::ms_fAspectRatio``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    for (size_t i = 0; i < std::size(hudHeightAddresses); i++)
+    {
+        MemPut<float*>(hudHeightAddresses[i] + 0x2, &HUDHeight[i]);
+    }
+
+    // Second player fix.
+    MemPut<float*>(0x58F9A0 + 0x2, &HUDWidth[38]);             // Weapon icon X
+    MemPut<float*>(0x58F993 + 0x2, &HUDWidth[16]);             // Weapon icon X
+    MemPut<float*>(0x58F972 + 0x2, &HUDHeight[16]);            // Weapon icon Y
+    MemPut<float*>(0x58FA8E + 0x2, &HUDWidth[17]);             // Ammo x
```
FileEX: Add a comment that these addresses point to ``CHud::DrawPlayerInfo``

0x58F9A0 + 0x2 = 0x58F9A2
0x58F993 + 0x2 = 0x58F995
0x58F972 + 0x2 = 0x58F974
0x58FA8E + 0x2 = 0x58FA90

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+    ScreenWidthScale = 640.0f / (*(ScreenAspectRatio) * 448.0f);
+
+    UpdateMiscFixes();
+    UpdateHUDFixes();
+}
+
+void InstallAspectRatioFixes()
+{
+    MemPut<BYTE>(0x745BD1, 0x7D);
+    MemPut<BYTE>(0x745BD9, 0x7C);
```
FileEX: Instead of put invalid instructions it is better to use MemSet and set these 2 bytes to 0x90 if you want to avoid the jumps.
image

Also add a comment that it disables jump instructions in ``GetVideoModeList``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    // Proportional coronas
+    MemSet((void*)0x6FB2C9, 0x90, 4);
```
FileEX, line 57: Add a comment that it skips multiplication by ScreenY in ``CCoronas::Render``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+    MemPut<BYTE>(0x6FB2BD, 0x6C);
+    MemPut<BYTE>(0x6FB2DC, 0x78);
+    MemPut<BYTE>(0x713BE5, 0x20);
+    MemPut<BYTE>(0x713B6D, 0x38);
+    MemPut<BYTE>(0x713CFB, 0x38);
+    MemPut<BYTE>(0x713EFC, 0x30);
+    MemPut<BYTE>(0x714004, 0x38);
```
FileEX: Add comments on what's going on here
```cpp
// Load ScreenY value from the stack
MemPut<BYTE>(0x6FB2BD, 0x6C); // CCoronas::Render
MemPut<BYTE>(0x6FB2DC, 0x78); // CCoronas::Render
MemPut<BYTE>(0x713BE5, 0x20); // CClouds::Render
MemPut<BYTE>(0x713B6D, 0x38); // CClouds::Render
MemPut<BYTE>(0x713CFB, 0x38); // CClouds::Render
MemPut<BYTE>(0x713EFC, 0x30); // CClouds::Render
MemPut<BYTE>(0x714004, 0x38); // CClouds::Render
```

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: It doesn't have to be an array, it can just be a float

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: This may cause a crash because the JMP instruction takes 5 bytes, there are only 4 bytes at the specified address, which means that the next instruction will be partially replaced with a random opcode (last byte of the offset)
image

The hook size should be 6 here

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+{
+    _asm
+    {
+        fstp    st
+        mov     edx, [ScreenAspectRatio]
+        fmul[edx]
+        mov     edi, [esp + 1Ch + 14h]
+        mov     edx, edi
+        mov     ebx, 5149A6h
+        jmp     ebx
```
FileEX: Replace these instructions with jmp

```cpp
static constexpr std::uintptr_t RETURN_CCamera_Find3rdPersonCamTargetVector = 0x5149A6;
```

```cpp
jmp RETURN_CCamera_Find3rdPersonCamTargetVector
```

FileEX: Additionally, are you sure that this hook works properly? It seems to me that the value at the top of the FPU register may be random here after doing fstp st, unless I missed something

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+{
+    return std::round((2.0f * atan(((ar) / (4.0f / 3.0f)) * tan(f / 2.0f * ((float)M_PI / 180.0f)))) * (180.0f / (float)M_PI) * 100.0f) / 100.0f;
+}
+
+void SetFOV(float factor)
+{
+    *ScreenFieldOfView = AdjustFOV(factor, *ScreenAspectRatio);
+}
+
+void __declspec(naked) CalculateAimingPoint()
```
FileEX: static

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,416 @@
...
+{
+    _asm
+    {
+        fstp    st
+        mov     edx, [ScreenAspectRatio]
+        fmul[edx]
```
FileEX: space

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: I mean CameraWidth

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: 0x6FF410 - hook size should be 9
Now this hook breaks the second mov instruction
image

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: I mean aboxe memset

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: above memset

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
FileEX: above memset

ghost:
Any updates on this? This is a long awaited feature, actually the issue is in the 1.6.1 milestone

ghost:
It would be very interesting if this pull returned

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: I can make it static

ghost:
I personally like the current "big" hud for 1080p (bad eyes) but I understand that it gets too big on 4k screens. An option where you could scale the hud is impossible?

henrymarques:
Hello! I was very sick last 4 months but I can continue this pr now, I do not remember what I was doing but I got issues with memory addresses as FileEX mentioned, maybe we could separate this into 2 pr since hud changes is ready (at least was before FileEX hud customizations but it seems easy to make it work again)

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: Rendering does not work properly if we ignore the value, I could be wrong, but I guess the value is a (partial?) tan, that is used to calculate where to position crosshair on screen. Without this hook crosshair does not stay in the right place
image

henrymarques:
> I personally like the current "big" hud for 1080p (bad eyes) but I understand that it gets too big on 4k screens. An option where you could scale the hud is impossible?

HUD match aspect ratio option will be kept, so you can use the big HUD simply unchecking that option, isn't the best solution but implementing HUD scale without making things weird needs more thinking

Fernando-A-Rocha:
Nice!
Will you add a client setting to keep the old hud? It should be disabled by default because your fixed version is nicer, imo

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,88 @@
...
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
+// NOTE: This code is based on ThirteenAG's fix, which is licensed under the MIT license.
+static float           CameraWidth = 0.01403292f;
```
FileEX: ``constexpr``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,88 @@
...
+static constexpr float WidthMult = 0.0015625f;
+static constexpr float HeightMult = 0.002232143f;
+static constexpr float SkyMultiFix = 10.0f;
+
+static float* ScreenAspectRatio = reinterpret_cast<float*>(0xC3EFA4);
+static float* ScreenFieldOfView = reinterpret_cast<float*>(0x8D5038);
+
+float AdjustFOV(float f, float ar)
+{
+    return std::round((2.0f * atan(((ar) / (4.0f / 3.0f)) * tan(f / 2.0f * ((float)M_PI / 180.0f)))) * (180.0f / (float)M_PI) * 100.0f) / 100.0f;
```
FileEX: ``std::atan``, ``std::tan``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,88 @@
...
+    HookInstall(0x6FF410, reinterpret_cast<DWORD>(SetFOV), 9);
+    HookInstall(0x51499E, reinterpret_cast<DWORD>(CalculateAimingPoint), 6);
+
+    // Skips division by CDraw::ms_fAspectRatio
+    MemSet((void*)0x50AD79, 0x90, 6);
+    MemPut<float*>(0x50AD5B, &CameraWidth); // CCamera::Find3rdPersonQuickAimPitch
+    MemPut<float*>(0x51498F, &CameraWidth); // CCamera::Find3rdPersonCamTargetVector
+}
+
+void CMultiplayerSA::InitHooks_WidescreenFix()
```
FileEX, line 84: Mark all functions as ``static``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,88 @@
...
+    MemPut<BYTE>(0x713CFB, 0x38);            // CClouds::Render
+    MemPut<BYTE>(0x713EFC, 0x30);            // CClouds::Render
+    MemPut<BYTE>(0x714004, 0x38);            // CClouds::Render
+}
+
+void InstallFieldOfViewFixes()
+{
+    // Fix sky blurring white in ultrawide screens
+    MemPut<const float*>(0x714843, &SkyMultiFix);
+    MemPut<const float*>(0x714860, &SkyMultiFix);
```
FileEX, line 73: You can change ``const float*`` to ``float*``

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,88 @@
...
+// NOTE: This code is based on ThirteenAG's fix, which is licensed under the MIT license.
+static float           CameraWidth = 0.01403292f;
+static constexpr float WidthMult = 0.0015625f;
+static constexpr float HeightMult = 0.002232143f;
+static constexpr float SkyMultiFix = 10.0f;
+
+static float* ScreenAspectRatio = reinterpret_cast<float*>(0xC3EFA4);
+static float* ScreenFieldOfView = reinterpret_cast<float*>(0x8D5038);
+
+float AdjustFOV(float f, float ar)
```
FileEX: Could you use more user-friendly names instead of 'f' and 'ar'?

Client/game_sa/CHudSA.cpp
```diff
@@ -19,8 +19,8 @@
 
 extern CGameSA* pGame;
 
-char szVehicleName[50] = {'\0'};
-char szZoneName[50] = {'\0'};
+static float radarAltimeterFix = 0.0014625f;                            // Changes altimeter width (and maybe x pos)
+static float aspectRatioMultiplicatorUntouched = 0.0015625f;            // (1 / 640)
```
FileEX: ``constexpr``

FileEX (changes requested):

Client/game_sa/CHudSA.cpp
henrymarques: forgotten that, thanks for remembering

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: Changing to `float*` give me a compilation error since `SkyMultiFix` is a constexpr, am I missing something?

FileEX: You're right, I didn't notice that. In that case, leave it as it is.

henrymarques:
> Nice! Will you add a client setting to keep the old hud? It should be disabled by default because your fixed version is nicer, imo

Yes, just uncheck "HUD match aspect ratio" checkbox on video options, it's checked by default (which means that will be using fixed hud)

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+// NOTE: This code is based on ThirteenAG's fix, which is licensed under the MIT license.
+
+static constexpr float CameraWidth = 0.01403292f;
+static constexpr float SkyMultiFix = 10.0f;
+
+static float* ScreenAspectRatio = reinterpret_cast<float*>(0xC3EFA4);
+static float* ScreenFieldOfView = reinterpret_cast<float*>(0x8D5038);
+
+static float AdjustFOV(float factor, float aspectRatio)
```
FileEX: noexcept

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
```diff
@@ -0,0 +1,89 @@
...
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
+// NOTE: This code is based on ThirteenAG's fix, which is licensed under the MIT license.
```
FileEX: This comment, imo, should be added to the comment section at the top of the file

Client/game_sa/CHudSA.cpp
```diff
@@ -56,23 +56,18 @@ CHudSA::CHudSA()
 {
     InitComponentList();
 
-    // Set the default values
-    m_fSniperCrosshairScale = 210.0f;
+    m_pfAspectRatioMultiplicatorX = (float*)VAR_AspectRatioMultX;
+    m_pfAspectRatioMultiplicatorY = (float*)VAR_AspectRatioMult;
```
FileEX: ``reinterpret_cast``

Client/game_sa/CHudSA.cpp
```diff
@@ -19,8 +19,8 @@
 
 extern CGameSA* pGame;
 
-char szVehicleName[50] = {'\0'};
-char szZoneName[50] = {'\0'};
+static float           radarAltimeterFix = 0.0014625f;                            // Changes altimeter width (and maybe x pos)
```
FileEX, line 22: ``constexpr``

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_WidescreenFix.cpp
henrymarques: yeah I was thinking about that before making the comment

FileEX (approved):

Fernando-A-Rocha (approved):
:shipit:

botder (approved):

PR #3446, opened 2024-06-08
Send command args in OnPlayerCommand

henrymarques:
Hello! It's a simple thing, onPlayerCommand used to receive only command, now it receives command and args. This change shouldn't break anything.

### DIFF
--- a/Server/mods/deathmatch/logic/CConsole.cpp
+++ b/Server/mods/deathmatch/logic/CConsole.cpp
@@ -57,6 +57,24 @@ bool CConsole::HandleInput(const char* szCommand, CClient* pClient, CClient* pEc
             CLuaArguments Arguments;
             Arguments.PushString(szKey);
 
+            if (szArguments)
+            {
+                // Create a copy and strtok modifies the string
+                char* szTempArguments = new char[strlen(szArguments) + 1];
+                strcpy(szTempArguments, szArguments);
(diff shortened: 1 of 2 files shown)


ghost (commented):
This would provide an opportunity to create keyloggers, so I don't think it's a good idea.

botder:
Why do you need this?

CrosRoad95:
> This would provide an opportunity to create keyloggers, so I don't think it's a good idea.

You already can do this

henrymarques:
> Why do you need this?

My entire gamemode is based on commands. This would help me a lot e.g. to deal with all variable args from /cars or /bikes. Even knowing keyloggers is already possible, I will understand if this change cannot be merged, due to /login or /register

PlatinMTA:
I don't get it tho, why can't you just deal with the arguments inside the function that's executed with the command?

Well I guess you are trying to do something like:

```lua
function onPlayerCommand(cmd, ...)
    local args = {...}
    if (cmd == "cars") then
        if (type(args[1]) ~= "number") then
            cancelEvent()
            return;
        end
    end
end
addEventHandler("onPlayerCommand", root, onPlayerCommand)

```

Dutchman101:
Thanks for the suggestion, but this PR has been rejected for reasons such as in the comments.

PR #3451, opened 2024-06-09
Fix #3445 - Allow comma separated list for scriptdebugloglevel

### LINKED ISSUE #3443: Allow differentiating custom-level debug messages

Debug level 3 shows all errors, warnings, info, and custom messages.
This can make the debug box quite messy and annoying if you don't care about some warnings and certain info messages.

Allow using levels higher than 4 for `outputDebugString`
With this change, the custom level can be used in `setPlayerScriptDebugLevel` to show only messages of that level.

- Level 3 would still show all messages (including the new custom-level ones).
- I'm unsure how many extra levels can be added, and whether there's some technical limitation or downside to allowing the addition of too many levels.

Another suggestion/solution on top of the above solution would be to make 'errors', 'warnings', and 'info' independent of each other when setting the player's debug level.

For example, if I could view only messages of types: `error + all custom`, or `error + custom level 5`.

However, I understand that this might make matters more confusing and counterintuitive. Maybe with the "right implementation" it can be near perfect, but I didn't think about it hard enough...yet

### Additional context

_No response_

### Security Policy

### LINKED ISSUE #3445: Allow comma separated list for scriptdebugloglevel

It would be great, if we could choose which debug messages will appear in log file. Currently it is achievable, but in a more restricted way.

```xml
<scriptdebugloglevel>1,2</scriptdebugloglevel>
```

For errors and warnings only.

```xml
<scriptdebugloglevel>0,3,4</scriptdebugloglevel>
```

For info messages and custom messages only.
Levels available here: https://wiki.multitheftauto.com/wiki/outputDebugString

none

### Additional context

```xml
<enablesd></enablesd>
```

Already does [similar approach.](https://wiki.multitheftauto.com/wiki/Anti-cheat_guide#%3Cenablesd%3E%3C/enablesd%3E)

### Security Policy

TracerDS:
Fixes #3445
Fixes #3443


Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -1394,6 +1393,41 @@ bool CMainConfig::SetSetting(const SString& strName, const SString& strValue, bo
...
+        // Merge
+        std::set<std::uint32_t> comboDebugLevelMap;
+        for (const auto& combo : curDebugLevelList)
+            MapInsert(comboDebugLevelMap, atoi(*combo));
+        for (const auto& combo : newDebugLevelList)
+            MapInsert(comboDebugLevelMap, atoi(*combo));
+
+        // Make a string
+        SString strComboResult;
+        for (const auto& uiId : comboDebugLevelMap)
```
tederis, line 1431: It's better to use an explicit type here because the type isn't obvious from the code.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -1394,6 +1393,41 @@ bool CMainConfig::SetSetting(const SString& strName, const SString& strValue, bo
...
+        ReadCommaSeparatedList(strCurDebugLevel, curDebugLevelList);
+
+        // Get new setting as as list of ids
+        std::vector<SString> newDebugLevelList;
+        ReadCommaSeparatedList(strValue, newDebugLevelList);
+
+        // Merge
+        std::set<std::uint32_t> comboDebugLevelMap;
+        for (const auto& combo : curDebugLevelList)
+            MapInsert(comboDebugLevelMap, atoi(*combo));
```
tederis: Please, use std::strtol.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -336,6 +336,21 @@ bool CMainConfig::Load()
...
 
+    {
+        SString strDebugLevel;
+        GetString(m_pRootNode, "scriptdebugloglevel", strDebugLevel);
+
+        std::vector<SString> debugLevelList;
+        ReadCommaSeparatedList(strDebugLevel, debugLevelList);
+        for (const auto& level : debugLevelList)
+        {
+            if (!isdigit(**level))
```
tederis: Maybe better to use `strtol` and the error checking?

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -1394,6 +1393,41 @@ bool CMainConfig::SetSetting(const SString& strName, const SString& strValue, bo
...
+        for (const auto& combo : curDebugLevelList)
+            MapInsert(comboDebugLevelMap, atoi(*combo));
+        for (const auto& combo : newDebugLevelList)
+            MapInsert(comboDebugLevelMap, atoi(*combo));
+
+        // Make a string
+        SString strComboResult;
+        for (const auto& uiId : comboDebugLevelMap)
+        {
+            if (!uiId)
```
tederis, line 1433: There have to be a check for a range, not only for the zero.

Server/mods/deathmatch/logic/CMainConfig.h
```diff
@@ -84,7 +84,10 @@ class CMainConfig : public CXMLConfig
     bool                        GetJoinFloodProtectionEnabled() { return m_bJoinFloodProtectionEnabled; };
     bool                        GetScriptDebugLogEnabled() { return m_bScriptDebugLogEnabled && !m_strScriptDebugLogFile.empty(); };
     const std::string&          GetScriptDebugLogFile() { return m_strScriptDebugLogFile; };
-    unsigned int                GetScriptDebugLogLevel() { return m_uiScriptDebugLogLevel; };
+
+    const std::set<SString>&    GetScriptDebugLogLevelMap() const noexcept { return m_scriptDebugLogLevelMap; };
+    std::set<SString>&          GetScriptDebugLogLevelMap() noexcept { return m_scriptDebugLogLevelMap; };
```
tederis: Do we really need the write access to this map from the outside of the class?

Server/mods/deathmatch/logic/CMainConfig.cpp
TracerDS: I copied the code from `enablesd` and modified it a little.
What about `stoi` instead of `strtol`?

Server/mods/deathmatch/logic/CMainConfig.h
TracerDS: Maybe. Lets leave it for now and we'll see later

Server/mods/deathmatch/logic/CMainConfig.cpp
tederis: > I copied the code from `enablesd` and modified it a little. What about `stoi` instead of `strtol`?

Yeah, `stoi` is even better.

Server/mods/deathmatch/logic/CMainConfig.cpp
TracerDS: Actually, wouldnt `isdigit` be more than perfect here? Max value is 4, if it goes above that it should ignore it.
Conversions would be a little expensive here (try-catch, performance costs (not that they matter a lot))

tederis: > Actually, wouldnt `isdigit` be more than perfect here? Max value is 4, if it goes above that it should ignore it. Conversions would be a little expensive here (try-catch, performance costs (not that they matter a lot))

Okay, sounds reasonable. But I suggest you to replace `**level` to `*level.c_str()`. Because ** is diffucult to understand without being immersed to the MTA codebase.

tederis: Or even better to replace `**level` to `level[0]`.

TracerDS: Alright

TracerDS:
The issue I currently have (and the reason why its a draft) is because of internal mta's logic I cannot easily change debugscript's logic.
Log files require a debug level integer instead of a vector of modes.
How could I go around that?

tederis:
> The issue I currently have (and the reason why its a draft) is because of internal mta's logic I cannot easily change debugscript's logic. Log files require a debug level integer instead of a vector of modes. How could I go around that?

What exactly makes it difficult to change the debugscript's logic?

TracerDS:
> What exactly makes it difficult to change the debugscript's logic?

How would I print (and save) debug strings considering the debug level is now a vector and not an integer?
The whole debugscript system needs to be changed

tederis:
> > What exactly makes it difficult to change the debugscript's logic?
> 
> How would I print (and save) debug strings considering the debug level is now a vector and not an integer? The whole debugscript system needs to be changed

There is literally one line you need to change in CScriptDebugging.cpp. You can pass the set of log levels to CScriptDebugging and test whether a message should be shown or not. At least looks like this.

Log

Fernando-A-Rocha:
Good job, still a draft?

TracerDS:
> Good job, still a draft?

Yeah, I kinda got stuck here.
Something about the internals of debugLevels (enablesd-like functionality) is off

Dutchman101:
> > Good job, still a draft?
> 
> Yeah, I kinda got stuck here. Something about the internals of debugLevels (enablesd-like functionality) is off

@TracerDS 
If you could elaborate, you can enlist help from other devs.
Btw, please resolve conflicts

PR #3452, opened 2024-06-09
Disable addDebugHook so cheaters have to use brain to hack the server

CrosRoad95:
Removes a goldmine for cheaters, make their life harder.
I know it breaks backwards compatibility with already existing cheats but we can deal with it. 😘

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaManager.cpp
@@ -226,7 +226,7 @@ void CLuaManager::LoadCFunctions()
         {"getPerformanceStats", CLuaFunctionDefs::GetPerformanceStats},
         {"setDevelopmentMode", CLuaFunctionDefs::SetDevelopmentMode},
         {"getDevelopmentMode", CLuaFunctionDefs::GetDevelopmentMode},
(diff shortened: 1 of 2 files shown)


Fernando-A-Rocha:
Tbh i agree. There's no real legitimate reason to use debug hook on a production server

TheNormalnij:
We can allow `addDebugHook` only in debug builds

TracerDS:
what about lua anticheat with `addDebugHook` used as a core function?

Lpsd:
So, you force cheaters to implement this functionality on their own (or similar alternatives) then what?

Xyrusek:
MTA should give an option to just disable debug hooks and loadstrings in server configuration file, cause in most cases it shouldn't be used on production servers.

TracerDS:
> MTA should give an option to just disable debug hooks and loadstrings in server configuration file, cause in most cases it shouldn't be used on production servers.

Loadstring and debug hooks dont make sense on the server either way. And you cannot disable loadstring on client.

Dutchman101:
Thanks for your suggestion, but it's denied. There are too many custom security solutions that rely on it for the opposite of cheating. We also know that for cheaters themselves, it's not so important to be able to use this; it doesn't offset the legitimate uses of debug hooks at all.

PR #3460, opened 2024-06-11
Fix outputDebugString with level 4 not showing

### LINKED ISSUE #3458: debugscript doesn't display outputDebugString level 4 at server-side
### Describe the bug

As in title.

### Steps to reproduce

```lua
outputDebugString("test level 4", 4, 255, 255, 255)
```

Client-side: displays, as it should be.
Server-side: it doesn't show.

### Version

r22505

### Additional context

```
debugscript: Modes available are 0 (None), 1 (Errors), 2 (Errors + Warnings), 3 (All)
```
#3440

### Relevant log output

_No response_

### Security Policy

ghost:
Fixes #3458


TracerDS:
~~Congratulations on your first PR 🥳~~
Nevermind, your second 😅

botder:
Instead of checking the value twice, why not do something like `uiMinimumDebugLevel = std::min(uiMinimumDebugLevel, 3)`

botder (approved):

PR #3462, opened 2024-06-11
Extract world removal class

TheNormalnij:
This PR extracts building removal logic from CWorldSA class to a new class CBuildingRemovalSA.


Dutchman101:
Replaced with 53e03d0

PR #3465, opened 2024-06-12
Fix Discord RPC and data types

G-Moris:
Fix Discord RPC and data types.
Something needs to be fixed?


Lpsd:
Why are some unsigned integer types being replaced for int64_t? (e.g `m_uiDiscordAppStart`)

G-Moris:
@Lpsd The structure treats these values as int64_t (long long), made for more convenient compatibility, if you are against such changes, I can change it to uint32_t

PR #3470, opened 2024-06-13
Fix createBuilding crash after engineSetPoolCapacity

TheNormalnij:
Fixes the missing resize operation for `m_buildingPool`. 
It was causing out-of-buffer writing and random crash.

### DIFF
--- a/Client/game_sa/CBuildingsPoolSA.cpp
+++ b/Client/game_sa/CBuildingsPoolSA.cpp
@@ -43,7 +43,7 @@ inline bool CBuildingsPoolSA::AddBuildingToPool(CClientBuilding* pClientBuilding
     m_buildingPool.entities[dwElementIndexInPool] = {pBuilding, (CClientEntity*)pClientBuilding};
 
     // Increase the count of objects
-    ++m_buildingPool.ulCount;
+    ++m_buildingPool.count;
 
     return true;
 }
@@ -118,7 +118,7 @@ void CBuildingsPoolSA::RemoveBuilding(CBuilding* pBuilding)
     (*m_ppBuildingPoolInterface)->Release(dwElementIndexInPool);
 
     // Decrease the count of elements in the pool
-    --m_buildingPool.ulCount;
+    --m_buildingPool.count;
 }
 
 void CBuildingsPoolSA::RemoveAllBuildings()
--- a/Client/game_sa/CBuildingsPoolSA.cpp
+++ b/Client/game_sa/CBuildingsPoolSA.cpp
@@ -40,7 +40,7 @@ inline bool CBuildingsPoolSA::AddBuildingToPool(CClientBuilding* pClientBuilding
     if (dwElementIndexInPool == UINT_MAX)
         return false;
 
-    m_buildingPool.arrayOfClientEntities[dwElementIndexInPool] = {pBuilding, (CClientEntity*)pClientBuilding};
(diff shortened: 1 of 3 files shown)


Client/game_sa/CPoolSAInterface.h
```diff
@@ -149,3 +149,16 @@ struct SPoolData
         }
     }
 };
+
+template <class T, class I>
```
botder: `class I` is unused?

Client/game_sa/CPoolSAInterface.h
```diff
@@ -149,3 +149,16 @@ struct SPoolData
...
+struct SVectorPoolData
+{
+    std::vector<SClientEntity<T>> entities;
+    unsigned long                 ulCount;
```
TracerDS: Can `ulCount` be a `std::uint32_t` instead of `unsigned long`?

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -210,15 +212,15 @@ bool CBuildingsPoolSA::Resize(int size)
     CBuildingSAInterface* newObjects = MemSA::malloc_struct<CBuildingSAInterface>(size);
     if (newObjects == nullptr)
```
TracerDS, line 213: `if(!newObjects)`

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -210,15 +212,15 @@ bool CBuildingsPoolSA::Resize(int size)
...
     {
-        Resize(curretnSize);
+        Resize(currentSize);
         return false;
     }
 
     tPoolObjectFlags* newBytemap = MemSA::malloc_struct<tPoolObjectFlags>(size);
     if (newBytemap == nullptr)
```
TracerDS, line 220: ~~`if(!newBytemap)`~~

TracerDS (changes requested):

Client/game_sa/CBuildingsPoolSA.cpp
tederis: This is not mentioned in the **Style Guide**. And depends on the reviewer's personal preferences. BTW, in UE codebase `== nullptr` is preferable.

TracerDS: `!` should be preferred as it takes less space and everyone knows its purpose. The same with `!strcmp`, but whatever

tederis: This is discussable. And I didn't want to joint it. I just pointed out that there are many codestyles with theirs advantages and disadvantages. And since this particular code part isn't referenced in the guidelines, there is no reason to push too much.

Client/game_sa/CPoolSAInterface.h
TheNormalnij: Yeah, i've removed this. I didn't understand why this exists

Client/game_sa/CPoolSAInterface.h
TheNormalnij: size_t

Client/game_sa/CBuildingsPoolSA.cpp
TheNormalnij: So, what do we keep here? I wrote this line automatically because all my commonly used languages require explicit null checks. Both variants are equal for me

TracerDS: Well, its up to you until a new coding guidelines appear

Issue #3457, opened 2024-06-10
dxDrawMaterialLine3D renders through buildings when heat haze is enabled and some other conditions are met


C0DELEVEL:
### Describe the bug

After the `postGUI` to `stage` changes (i think) it behaves strangely for some players (about 30% of ~400 unique players) - it renders through walls and gta sa buildings (behaves like postGUI=true) depending on the angle from which you look.

The issue exists when `stage` parameter is set to `false` (legacy) or `postfx`

### Steps to reproduce

1. Warp to `-2020.1213378906,-98.689750671387,35.1640625`
2. Start below script:
```    
    local txt = dxCreateTexture(128,128)
    local pixels = dxGetTexturePixels (txt)
    for ii=1,128 do
        for i=1,128 do
            dxSetPixelColor(pixels,i,ii,255,255,255)
        end
    end
    dxSetTexturePixels(txt, pixels)
    addEventHandler('onClientRender',root,function()
        dxDrawMaterialLine3D(-2023.3182373047,-105.71241760254,35.290626525879,-2023.3182373047+1,-105.71241760254+1,35.290626525879,txt,1,tocolor(255,255,255),false,-2020.4468994141,-99.970504760742,35.290626525879)
    end)
```
3. Look at the proper angle (pretty easy to achieve, there are only few angles that work **properly**)
4. You will see the white rectangle through the wall (just like with postGUI=true, but it's false/postfx)

### Version

Multi Theft Auto v1.6-release-22505

### Additional context

An example showing the annoyance of the issue:
[click to view image](https://cdn.discordapp.com/attachments/1242512557772767265/1249720778312912936/image.png?ex=666854ac&is=6667032c&hm=925c3cf0a8f8ef2711e35a6ff95e0c9aea0e73bbbdfd4866590f6d8305b2292b)

and result using the above example code:
![image](https://github.com/multitheftauto/mtasa-blue/assets/107777575/5958269e-3e95-4f65-831f-4c857a83f78f)

### Relevant log output

```shell
-
```

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

tederis:
Cannot reproduce the problem on 22512. Did you test it on a pure(without any custom modifications) freeroam server?

C0DELEVEL:
As I've already said above, it was hard to debug.. But now I have more details and it's not hardware dependent, but the graphic settings dependent.

[...]

PR #3478, opened 2024-06-16
Postfx stage fix

tederis:
Fixes the bug described in #3457. This PR changes the render order in the following way: 

1. Color filter, grain, radiosity, infrared and nightvision effects are still occured before the `"postfx"` stage.
2. Heat haze and CCTV effects are now rendered after the `"postfx"` stage.
This is due to specifics of the bug(see https://github.com/multitheftauto/mtasa-blue/issues/3457#issuecomment-2167423832).

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -2008,48 +2008,47 @@ void CCore::OnPreFxRender()
 //
 // OnPreHUDRender
 //
-void CCore::OnPreHUDRender()
+void CCore::OnPreHUDRender(PreHUDRenderStage stage)
 {
-    IDirect3DDevice9* pDevice = CGraphics::GetSingleton().GetDevice();
-
-    if (CGraphics::GetSingleton().HasLine3DPostFXQueueItems() || CGraphics::GetSingleton().HasPrimitive3DPostFXQueueItems())
+    if (stage == PreHUDRenderStage::PostColorFilter)
     {
-        /*
-            Although MTA render zones are expensive, we should use them twice in the bounds of the function
-            because some of render states from PostFX drain to the 2D part of the frame.
-        */
-        CGraphics::GetSingleton().EnteringMTARenderZone();
+        if (!CGraphics::GetSingleton().HasLine3DPostFXQueueItems() && !CGraphics::GetSingleton().HasPrimitive3DPostFXQueueItems())
+            return;
+    
+        CGraphics::GetSingleton().EnteringMTARenderZone();      
 
         CGraphics::GetSingleton().DrawPrimitive3DPostFXQueue();
-        CGraphics::GetSingleton().DrawLine3DPostFXQueue();
+        CGraphics::GetSingleton().DrawLine3DPostFXQueue();  
 
         CGraphics::GetSingleton().LeavingMTARenderZone();
     }
-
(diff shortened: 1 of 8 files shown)


Client/core/CCore.cpp
```diff
@@ -2008,48 +2008,47 @@ void CCore::OnPreFxRender()
 //
 // OnPreHUDRender
 //
-void CCore::OnPreHUDRender()
+void CCore::OnPreHUDRender(PreHUDRenderStage stage)
 {
-    IDirect3DDevice9* pDevice = CGraphics::GetSingleton().GetDevice();
-
-    if (CGraphics::GetSingleton().HasLine3DPostFXQueueItems() || CGraphics::GetSingleton().HasPrimitive3DPostFXQueueItems())
+    if (stage == PreHUDRenderStage::PostColorFilter)
```
TheNormalnij: Maybe just split function? This will remove unnecessary check

tederis: Do you mean
```
if (stage == PreHUDRenderStage::PostColorFilter)
   OnPostColor();
else
   OnPreHUD();
```
?

TheNormalnij: i mean
```cpp
CCore::OnPostColorFilterRender {}
CCore::OnPreHUDRender {}
```

tederis: As I've already mentioned in the description (https://github.com/multitheftauto/mtasa-blue/pull/3478#issue-2355528510) this would make the interface noisy. Because I cannot add a new method in the middle of the CCoreInterface.

TheNormalnij: Ah, i see. 
I would prefer to add a function, because it doesn't affects the logic or runtime. 
It seems that creating an additional enum and a check creates more mess than misplaced method does

Client/core/CCore.cpp
```diff
@@ -2008,48 +2008,47 @@ void CCore::OnPreFxRender()
...
-        */
-        CGraphics::GetSingleton().EnteringMTARenderZone();
+        if (!CGraphics::GetSingleton().HasLine3DPostFXQueueItems() && !CGraphics::GetSingleton().HasPrimitive3DPostFXQueueItems())
+            return;
+    
+        CGraphics::GetSingleton().EnteringMTARenderZone();      
 
         CGraphics::GetSingleton().DrawPrimitive3DPostFXQueue();
-        CGraphics::GetSingleton().DrawLine3DPostFXQueue();
+        CGraphics::GetSingleton().DrawLine3DPostFXQueue();
```
TheNormalnij: space

Client/core/CCore.cpp
tederis: I'm personally also inclined to the dedicated method. So, since there's no other opinion I'll do the thing.

botder:
Crashes on disconnect - fixed in a7e58e0c12eb9853f80987952b2d40eef76ed096

image

tederis:
> Crashes on disconnect - fixed in [a7e58e0](https://github.com/multitheftauto/mtasa-blue/commit/a7e58e0c12eb9853f80987952b2d40eef76ed096)
> 
> image

Oh, I forgot about the nulling. Thanks for the fix.

PR #800, opened 2019-02-02
Improve ped sync

tederis:
Originally MTA does nothing but send kinematics data and just applies it without any interpolation. So what we can see as a result is twitching among recieved points. This simple commit intended to mitigate these ped's jumps and brings a bit more lighter sync packets.

PR #3480, opened 2024-06-16
Ped sync improvements

tederis:
This PR is a reimplementation of #800 with additional fixes in rotation interpolation. 

The video below demonstrates that even without animations a ped is well synchronized (in comparison to the pre-PR). 
https://youtu.be/17vEh7k_RSI?si=8kxGwODs7lIn8OaF

Note that since peds are correctly syncronized now, there may be problems with old scripts relying on a ped's animation speed. These scripts will have to use animation without a root motion or use `setPedAnalogControlState`.

### DIFF
--- a/Shared/sdk/CVector.h
+++ b/Shared/sdk/CVector.h
@@ -26,11 +26,15 @@ private:
     static constexpr float FLOAT_EPSILON = 0.0001f;
 
 public:
-    float fX = 0.0f;
-    float fY = 0.0f;
-    float fZ = 0.0f;
+    float fX;
+    float fY;
+    float fZ;
 
-    constexpr CVector() = default;
+    struct NoInit{};
+
+    CVector(NoInit) {}
+
+    constexpr CVector() : fX(0.0f), fY(0.0f), fZ(0.0f) {}
 
     constexpr CVector(float x, float y, float z) : fX(x), fY(y), fZ(z) {}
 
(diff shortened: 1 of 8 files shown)


Fernando-A-Rocha:
Do you have a video of how it looks without this PR? 😁

tederis:
> Do you have a video of how it looks without this PR? 😁

There's the old video (from the previous PR) with the pre-PR behaviour: https://youtu.be/UcSOXd9PxhE?si=cbNeF8FVg-xf5tBa

Fernando-A-Rocha:
Nice, good job. The difference is notable.

TheNormalnij (dismissed):

TheNormalnij:
lgtm

Lpsd (approved):

Shared/sdk/CVector.h
```diff
@@ -26,11 +26,15 @@ class CVector
...
-    float fY = 0.0f;
-    float fZ = 0.0f;
+    float fX;
+    float fY;
+    float fZ;
 
-    constexpr CVector() = default;
+    struct NoInit{};
+
+    CVector(NoInit) {}
```
TracerDS, line 35: Whats the point behind it?

tederis: This is the call to a special constructor without vector's components initialization. For the sake of a performance.

tederis: See for example https://github.com/rheit/zdbsp/blob/master/tarray.h or https://docs.unrealengine.com/4.27/en-US/API/Runtime/Core/UObject/FNameEntryId/

botder (approved):

PR #3481, opened 2024-06-16
Refactoring Part 1 - CLuaArguments

TracerDS:


Fernando-A-Rocha:
Smaller refactor PRs like this that are labeled as something is good!  Better than mixing all the refactored code 👍🏼

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
```diff
@@ -176,20 +170,19 @@ bool CLuaArgument::CompareRecursive(const CLuaArgument& Argument, std::set<CLuaA
...
-                    pKnownTables->insert(m_pTableData);
-                    if (*iter != *iterCompare)
-                        return false;
-                }
+            auto iter = m_pTableData->begin();
+            auto iterCompare = Argument.m_pTableData->begin();
+            if (pKnownTables->find(m_pTableData) != pKnownTables->end())
+                return true;
+
+            pKnownTables->insert(m_pTableData);
```
tederis, line 173: You're doing the double work here. std::set::insert is doing the same find as you have already done above. You can insert a value and use the second attribute of the returned value to check if the value was already inserted.

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -229,28 +224,26 @@ bool CLuaArguments::Call(CLuaMain* pLuaMain, const CLuaFunctionRef& iLuaFunction
...
-    else
-    {
-        int iReturns = lua_gettop(luaVM) - luaStackPointer;
 
-        if (returnValues != NULL)
+    int iReturns = lua_gettop(luaVM) - luaStackPointer;
+
+    if (returnValues)
+    {
+        for (int i = -iReturns; i <= -1; i++)
```
tederis: int32_t maybe?

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -293,82 +286,79 @@ bool CLuaArguments::CallGlobal(CLuaMain* pLuaMain, const char* szFunction, CLuaA
...
     }
-    else
-    {
-        int iReturns = lua_gettop(luaVM) - luaStackPointer;
+    int iReturns = lua_gettop(luaVM) - luaStackPointer;
 
-        if (returnValues != NULL)
+    if (returnValues)
+    {
+        for (int i = -iReturns; i <= -1; i++)
```
tederis: int32_t maybe?

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -492,134 +476,119 @@ bool CLuaArguments::WriteToBitStream(NetBitStreamInterface& bitStream, CFastHash
...
     pKnownTables->insert(std::make_pair(this, pKnownTables->size()));
 
-    bool                                  bIsArray = true;
-    unsigned int                          iArrayPos = 1;            // lua arrays are 1 based
-    vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
+    bool bIsArray = true;
+    // lua arrays are 1 based
+    std::uint32_t iArrayPos = 1;
+
+    std::vector<CLuaArgument*>::const_iterator iter = m_Arguments.begin();
```
tederis: This is the best place to use auto

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -2361,8 +2361,8 @@ int CLuaGUIDefs::GUIGridListAddRow(lua_State* luaVM)
         {
             // Vector containing our string arguments.  We add a bool to store whether it was originally a number.
             std::vector<std::pair<SString, bool> >     m_items;
-            std::vector<CLuaArgument*>::const_iterator it = Arguments.IterBegin();
-            for (it; it != Arguments.IterEnd(); it++)
+            std::vector<CLuaArgument*>::const_iterator it = Arguments.begin();
```
tederis: auto

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
```diff
@@ -2427,8 +2427,8 @@ int CLuaGUIDefs::GUIGridListInsertRowAfter(lua_State* luaVM)
...
```
tederis: auto

Server/mods/deathmatch/logic/CElement.h
```diff
@@ -91,11 +91,11 @@ class CElement
     virtual CElement* Clone(bool* bAddEntity, CResource* pResource) { return nullptr; }
     bool              IsCloneable();
 
-    bool         IsBeingDeleted() { return m_bIsBeingDeleted; };
-    void         SetIsBeingDeleted(bool bBeingDeleted) { m_bIsBeingDeleted = bBeingDeleted; };
+    bool         IsBeingDeleted() { return m_bIsBeingDeleted; }
+    void         SetIsBeingDeleted(bool bBeingDeleted) { m_bIsBeingDeleted = bBeingDeleted; }
```
tederis, line 95: const method qualifier?

Server/mods/deathmatch/logic/lua/CLuaArguments.cpp
```diff
@@ -169,9 +161,9 @@ void CLuaArguments::PushAsTable(lua_State* luaVM, CFastHashMap<CLuaArguments*, i
     lua_newtable(luaVM);
 
     // push it onto the known tables
-    int size = pKnownTables->size();
+    auto size = pKnownTables->size();
```
tederis: Please, don't use auto for non-obvious types.

tederis (changes requested):

Server/mods/deathmatch/logic/CElement.h
TracerDS: For other PR, I was in CElement only to add `const` qualifier to the `GetID` method

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
TracerDS: For other PR

Client/mods/deathmatch/logic/luadefs/CLuaGUIDefs.cpp
TracerDS: For other PR

Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
TracerDS: Doesn't really matter for normal `int` but sure, to keep "compatibility"

Client/mods/deathmatch/logic/lua/CLuaArgument.cpp
TracerDS: Would `return m_pTableData == Argument.m_pTableData;` work here instead of the loop too?

> Yes, `operator==` is correctly defined for all standard containers (**except** the unordered containers - based on 23.2.5.2 of the standard), and will generally do a lexicographic comparison. See for example [here](http://en.cppreference.com/w/cpp/container/set/operator_cmp). The relevant quote:
>> Checks if the contents of lhs and rhs are equal, that is, whether lhs.size() == rhs.size() and each element in lhs has equivalent element in rhs at the same position.
> 
> Since `std::set` is an ordered container, any set with the same size and same elements (given the comparators are the same) will necessarily have them in the same position, hence will compare equal.

tederis: Yes, it should work.

TracerDS:

PR #3486, opened 2024-06-17
Fix setVehicleVariant turns on the engine

### LINKED ISSUE #542: setVehicleVariant turns on the engine, getVehicleEngineState still returns false
**Describe the bug**
If you turn off your vehicle engine and use setVehicleVariant on it, the engine will go on, tho getVehicleEngineState will still be returning false.

**To reproduce**
1. Enter a vehicle
2. `srun setVehicleEngineState(getPedOccupiedVehicle(getPlayerFromName("p")), false)`
3. `srun setVehicleVariant(getPedOccupiedVehicle(getPlayerFromName("p")), 0, 0)`
4. `srun getVehicleEngineState(getPedOccupiedVehicle(getPlayerFromName("p")))`
5. Engine is on, getVehicleEngineState returning false

**Expected behaviour**
To not turn engine on when using setVehicleVariant.

**Screenshots**

**MTA Client:** 
 - Version: 
 - Running in Wine? 

**MTA Server:** 
 - Version 
 - OS: 

**Additional context**

From https://bugs.mtasa.com/view.php?id=8663

Proxy-99:
Fixes #542


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1117,8 +1117,12 @@ void CClientVehicle::SetVariant(unsigned char ucVariant, unsigned char ucVariant
 
     // clear our component data to regenerate it
     m_ComponentData.clear();
-
     ReCreate();
+
+    if (!m_bEngineOn)
```
tederis: What is the purpose of this condition?

Proxy-99: if the engine is on then no need to call 
`m_pVehicle->SetEngineOn(m_bEngineOn);`
since using setVehicleVariant will recreate the vehicle and will set the engine to true
but why setVehicleVariant recreate the vehicle anyway?


```
void CClientVehicle::SetVariant(unsigned char ucVariant, unsigned char ucVariant2)
{
    m_ucVariation = ucVariant;
    m_ucVariation2 = ucVariant2;

    // clear our component data to regenerate it
    m_ComponentData.clear();
    ReCreate();

    if (!m_bEngineOn)
    {
        m_pVehicle->SetEngineOn(m_bEngineOn); 
    }
}
```

tederis: The only thing the SetEngineOn doing is toggling a flag inside CVehicle object. So there's no overhead in calling that function and thus this condition can be omitted.

TracerDS: > if the engine is on then no need to call `m_pVehicle->SetEngineOn(m_bEngineOn);` since using setVehicleVariant will recreate the vehicle and will set the engine to true but why setVehicleVariant recreate the vehicle anyway?
> 
> ```
> void CClientVehicle::SetVariant(unsigned char ucVariant, unsigned char ucVariant2)
> {
>     m_ucVariation = ucVariant;
>     m_ucVariation2 = ucVariant2;
> 
>     // clear our component data to regenerate it
>     m_ComponentData.clear();
>     ReCreate();
> 
>     if (!m_bEngineOn)
>     {
>         m_pVehicle->SetEngineOn(m_bEngineOn); 
>     }
> }
> ```

it needs to recreate the vehicle because the vehicle model changes

Proxy-99: I will check later for the root of this problem 
tho Create()
set  m_pVehicle->SetEngineOn(m_bEngineOn);
so value should be false 
getVehicleEngineState still returns false tho the engine is ON 
which really strange will need few tests and checking

G-Moris: pls, CClientVehicle::IsEngineOn and CClientVehicle::SetEngineOn

G-Moris:
```
    if (!IsEngineOn())
    {
        SetEngineOn(m_bEngineOn); 
    }
```

G-Moris: Maybe you wanted to do something like that?
```
if (!m_bEngineOn)
    SetEngineOn(m_bEngineOn)
```

Proxy-99: the problem within this line of code set the engine after the warp and before it to test, using warp will set engine on
****
within CClientVehicle::Create()
```
 // Eventually warp driver back in
 if (m_pDriver)
     m_pDriver->WarpIntoVehicle(this, 0);

```
another test 
 set engine off while you outside vehicle 
you will notice engine will stay off after appling setVehicleVariant by seeing the vehicle smoke not showing
so warping ped into vehicle is the root of this problem

you can also warp ped into vheicle engine will be ON on warp

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2638,6 +2636,8 @@ void CClientVehicle::Create()
         if (m_pDriver)
             m_pDriver->WarpIntoVehicle(this, 0);
 
+         m_pVehicle->SetEngineOn(m_bEngineOn);
```
TracerDS: why do we move this from top to bottom?

TracerDS (approved):
LGTM

Client/mods/deathmatch/logic/CClientVehicle.cpp
Proxy-99: > why do we move this from top to bottom?

to reset engine setting on `m_pDriver->WarpIntoVehicle(this, 0); `it will set engine ON when u warp a ped into vehicle so this is why I put it after the warp

ghost (approved):
Looks good 🚀

Fernando-A-Rocha (approved):

Fernando-A-Rocha:
merge this for 1.6.1 with backwards incompatible label?
https://wiki.multitheftauto.com/wiki/Changes_in_1.6.1

botder (approved):

PR #3490, opened 2024-06-18
Read a file directly from path - `fileRead` and `fileGetContents` update

TracerDS:
Currently, to read a file you have to have a boilerplate code for opening and closing a file:
```lua
function ReadFile(path)
    local file = fileOpen(path)
    if not file then return end
    local content = fileRead(file, fileGetSize(file))
    fileClose(file)
    return content
end
```
This PR will allow you to read file from path without having to define this boilerplate:
```lua
local content = fileRead('MyFile.txt')

local content = fileGetContents('MyFile.txt')
```


Fernando-A-Rocha:
Genius

tederis:
Are you aware of function `fileGetContents`? You can just `fileGetContents(fileOpen("file.txt"))`.

TracerDS:
> Are you aware of function `fileGetContents`? You can just `fileGetContents(fileOpen("file.txt"))`.

Does `fileGetContents` close the file automatically for you?
It also requires a file handle. It was either that or modify `fileGetContents`

MegadreamsBE:
> > Are you aware of function `fileGetContents`? You can just `fileGetContents(fileOpen("file.txt"))`.
> 
> Does `fileGetContents` close the file automatically for you? It also requires a file handle. It was either that or modify `fileGetContents`

It doesn't no. So I'm fine with this change for quality of life. Perhaps fileGetContents could have something similar

tederis:
> > Are you aware of function `fileGetContents`? You can just `fileGetContents(fileOpen("file.txt"))`.
> 
> Does `fileGetContents` close the file automatically for you? It also requires a file handle. It was either that or modify `fileGetContents`

Oh, you're right, it doesn't. Okay then.

Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
```diff
@@ -702,51 +702,91 @@ int CLuaFileDefs::fileFlush(lua_State* luaVM)
...
+    CScriptFile* pFile = new CScriptFile(pThisResource->GetScriptID(), strMetaPath, DEFAULT_MAX_FILESIZE);
+#else
+    eAccessType  accessType = strInputPath[0] == '@' ? eAccessType::ACCESS_PRIVATE : eAccessType::ACCESS_PUBLIC;
+    CScriptFile* pFile = new CScriptFile(pThisResource->GetScriptID(), strMetaPath, DEFAULT_MAX_FILESIZE, accessType);
+#endif
+    // Try to load it
+    if (!pFile->Load(pResource, CScriptFile::MODE_READ))
+    {
+        delete pFile;
+        throw std::runtime_error(SString("unable to load file '%s'", strInputPath.c_str()));
```
tederis: This will lead to std::terminate because `runtime_error` isn't properly handled by the argument parser.

Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
```diff
@@ -702,51 +702,91 @@ int CLuaFileDefs::fileFlush(lua_State* luaVM)
...
+    }
+
+    auto content = ReadFile(pFile, pFile->GetSize());
```
tederis: Please, don't use auto for implicit types.

Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
```diff
@@ -702,51 +702,91 @@ int CLuaFileDefs::fileFlush(lua_State* luaVM)
...
+            throw std::runtime_error(canModify.second);
         }
-        else
+    }
+    {
+        auto canModify = CheckCanAccessOtherResourceFile(pThisResource, pResource, strAbsPath);
+        if (!canModify.first)
         {
-            m_pScriptDebugging->LogBadPointer(luaVM, "file", 1);
+            throw std::runtime_error(canModify.second);
```
tederis: This will lead to std::terminate because runtime_error isn't properly handled by the argument parser.

Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
```diff
@@ -702,51 +702,91 @@ int CLuaFileDefs::fileFlush(lua_State* luaVM)
...
+    CResource* pResource = pThisResource;
+    if (!CResourceManager::ParseResourcePathInput(strInputPath, pResource, &strAbsPath, &strMetaPath))
+        return false;
+
+    {
+        auto canModify = CheckCanModifyOtherResource(pThisResource, pResource);
+        if (!canModify.first)
         {
-            m_pScriptDebugging->LogWarning(luaVM, "out of memory");
+            throw std::runtime_error(canModify.second);
```
tederis: This will lead to std::terminate because runtime_error isn't properly handled by the argument parser.

tederis (changes requested):

Fernando-A-Rocha:
@TracerDS Make fileGetContents accept a string as argument, so it does the opening and closing of the file automatically. It can keep working with an open file handle as argument for backwards compatibility and certain use cases. This way, there is no need for a separate new function.

TracerDS:
> @TracerDS Make fileGetContents accept a string as argument, so it does the opening and closing of the file automatically. It can keep working with an open file handle as argument for backwards compatibility and certain use cases.

so, the same thing as `fileRead` but for `fileGetContents`?

Fernando-A-Rocha:
> > @TracerDS Make fileGetContents accept a string as argument, so it does the opening and closing of the file automatically. It can keep working with an open file handle as argument for backwards compatibility and certain use cases.
> 
> so, the same thing as `fileRead` but for `fileGetContents`?

I don't see any cons to that

Fernando-A-Rocha:
So uhmmm, arent fileRead and fileGetContents the same now? Is fileRead necessary?

MegadreamsBE:
> So uhmmm, arent fileRead and fileGetContents the same now? Is fileRead necessary?

One major difference at least is that fileGetContents requires files to be in meta.xml, and has the ability to validate them. But overall they are getting pretty close to each other

tederis:
> > So uhmmm, arent fileRead and fileGetContents the same now? Is fileRead necessary?
> 
> One major difference at least is that fileGetContents requires files to be in meta.xml, and has the ability to validate them. But overall they are getting pretty close to each other

Only if the second argument is set to true.

Fernando-A-Rocha:
I like fileRead but it's exactly the same as doing fileGetContents(path, false)

Can we have fileRead still?

TracerDS:
> I like fileRead but it's exactly the same as doing fileGetContents(path, false)
> 
> Can we have fileRead still?

There is no reason to remove it

Fernando-A-Rocha:
> > I like fileRead but it's exactly the same as doing fileGetContents(path, false)
> > Can we have fileRead still?
> 
> There is no reason to remove it

Sorry, I had a brain fart. Forgot fileRead already existed.
This is nice QoL feature.

PR LGTM then :shipit:

Dutchman101:
This PR has been reverted after merge

Details on revert reason: https://github.com/multitheftauto/mtasa-blue/commit/97920d83de400e0378f73ea93e87b27cf59dd758#commitcomment-143697266 (and up)
cc @TracerDS

PR #3494, opened 2024-06-19
Refactoring Part 3 - CLuaFunctionParser

TracerDS:
#3493 related

There wont be a merge conflict as this PR is an extension to #3493


Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -723,30 +725,34 @@ struct CLuaFunctionParser<ErrorOnFailure, ReturnOnFailure, Func> : CLuaFunctionP
...
-            }
-            else
-            {
-                pScriptDebugging->LogCustom(L, strError.c_str());
-                lua::Push(L, ReturnOnFailure);
-            }
+            strError = e.what();
+            bIsCustomError = true;
+        }
+        if (!strError.empty())
```
tederis: Should be `if (strError.empty())`

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -19,6 +19,15 @@ class CLuaArgument;
 #include "lua/LuaBasic.h"
 #include <lua/CLuaMultiReturn.h>
 
+class ArgumentParserError
```
tederis: I would name it `LuaFunctionError` and add a new field `m_bWarning`. So you can inherit a class from it and set whether it should produce error or warning.

TracerDS: Good idea, thanks ❤

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -192,79 +205,79 @@ struct CLuaFunctionParserBase
         // primitive types
         if constexpr (std::is_same_v<T, std::string> || std::is_same_v<T, std::string_view>)
             return (iArgument == LUA_TSTRING || iArgument == LUA_TNUMBER);
-        if constexpr (std::is_same_v<T, bool>)
+        else if constexpr (std::is_same_v<T, bool>)
```
botder, line 208: There is no need to add an `else` here and and in the changed lines below. The _if statement_ above always ends in a _return_.

TracerDS: > There is no need to add an `else` here and below. The _if statement_ above always ends in a _return_.

I added `else`s due to the nature of `if constexpr`. Not putting else in `if constexpr` will not disregard the latter statements. In other function it ends up with a compile time error.
C++ is not evolved yet for it to recognize early returns in `if constexpr` statements,

botder: Good to know, I expected the compiler to be better.

botder (approved):

PR #3496, opened 2024-06-20
Fix #3495 OGG sound files can't be played as RAW data

### LINKED ISSUE #3495: OGG sound files can't be played as RAW data.
### Describe the bug

If you use fileRead() and try to play a sound with that result (raw data), you will receive an error in the F8 console, and the sound won't be played. This issue only occurs with OGG files. I have tried this with MP3 and WAV files, and it works perfectly. I have also tried various export and conversion settings for OGG sound files, but none of them seem to work.

The error in the console:
BASS ERROR 2 in LoadMedia  path:E:\Multi Theft Auto 1.6\mods\deathmatch\resources\soundtest\OggS  3d:0  loop:0

The test resource with the sound files: [soundtest_resource.zip](https://github.com/user-attachments/files/15908666/soundtest_resource.zip)

### Steps to reproduce

1. Download the test resource and run it.
2. Check your debugscript, you will see that, only the 'soundOGG_raw' variable will return false.

### Version

Client: v1.6-release-22512 (Windows 10, 64-bit)
Server: v1.6-release-21890 (Windows 10, 64-bit)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #3495 

The problem is that this condition returns true for raw data from .ogg files

https://github.com/multitheftauto/mtasa-blue/blob/f5b599c9f45777f924f7980cadb2d3cc6431d8b8/Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp#L168

This is because the IsValidFilePath function returns falsely true because strlen only reads the first 4 bytes which are the header of the .ogg files (```0x4F 0x67 0x67 0x53```) and then there is an empty byte ```\0``` where strlen stops.

https://github.com/multitheftauto/mtasa-blue/blob/f5b599c9f45777f924f7980cadb2d3cc6431d8b8/Shared/mods/deathmatch/logic/Utils.cpp#L76-L93

I think this needs to be patched in the IsValidPath function, but at the moment it requires editing in many places, so let's make a temporary fix like this.


Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
```diff
@@ -165,7 +165,12 @@ int CLuaAudioDefs::PlaySound(lua_State* luaVM)
                 SString strFilename;
                 bool    bIsURL = false;
                 bool    bIsRawData = false;
-                if (CResourceManager::ParseResourcePathInput(strSound, pResource, &strFilename))
+
+                // Fix OGG sound files can't be played as RAW data #3495 
+                unsigned char oggHeader[] = {0x4F, 0x67, 0x67, 0x53};
```
TracerDS: Use `std::uint8_t` instead of `unsigned char`, please

TracerDS (approved):

Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp
tederis: Maybe better to compare with uint32 instead? Than memcmp won't be needed.

tederis: But memcmp is also ok. But you should make this array static at least

botder:
Can't you overload `IsValidFilePath` for this?

FileEX:
Done

Shared/mods/deathmatch/logic/Utils.cpp
```diff
@@ -92,6 +92,24 @@ bool IsValidFilePath(const char* szDir)
     return true;
 }
 
+bool IsValidFilePath(const char* szDir, size_t length)
+{
+    if (szDir == nullptr)
+        return false;
+
+    unsigned char c, c_d;
```
TracerDS: Please use `std::uint8_t` instead of `unsigned char`

Shared/mods/deathmatch/logic/Utils.cpp
```diff
@@ -518,6 +536,24 @@ bool IsValidFilePath(const char* szDir)
...
```
TracerDS: Above

TracerDS (approved):

Shared/mods/deathmatch/logic/Utils.cpp
botder: A pretty strange choice when we are dealing with characters from `const char*` (unsigned char is fine).

ghost (approved):
Looks good 🚀

Issue #3498, opened 2024-06-20
debugscript behavior needs tweaking


ghost:
### Describe the bug

Currently debugscript doesn't behave accordingly to it's description:
```
debugscript: Modes available are 0 (None), 1 (Errors), 2 (Errors + Warnings), 3 (All)
```
As of now it checks whether debug level of player is equal or higher than debug level of message. Do note that outputDebugString makes use of levels from 0 to 4. Moreover, client seems to miss relevant checks; by having /debugscript set to 1 all of client-side debug messages would be displayed, but it should display errors only.

### Steps to reproduce

Play around with `outputDebugString` and /debugscript

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

PR #3499, opened 2024-06-20
Fix server-side debugscript behavior

ghost:
Partially solves #3498 due the fact the client isn't aware of debugscript level. Thanks for @TracerDS (refactored for loop) & @MegadreamsBE (help with getting rid of magic numbers).


Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -143,16 +143,38 @@ void CScriptDebugging::PrintLog(const char* szText)
     }
 }
 
+bool CScriptDebugging::CheckForSufficientDebugLevel(unsigned int uiPlayerDebugLevel, unsigned int uiMessageDebugLevel)
+{
+    bool uiSufficientDebugLevel = false;
```
botder: Please don't use Hungarian notation, it's also wrong here. Rename the variable to `isSufficient` or similiar.

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -143,16 +143,38 @@ void CScriptDebugging::PrintLog(const char* szText)
     }
 }
 
+bool CScriptDebugging::CheckForSufficientDebugLevel(unsigned int uiPlayerDebugLevel, unsigned int uiMessageDebugLevel)
```
TracerDS: Please use `std::uint32_t` instead of `unsigned int`

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -73,6 +89,7 @@ class CScriptDebugging
                    unsigned char ucGreen = 255, unsigned char ucBlue = 255);
 
     void PrintLog(const char* szText);
+    bool CheckForSufficientDebugLevel(unsigned int uiPlayerDebugLevel, unsigned int uiMessageDebugLevel);
```
TracerDS: Same as above + add `const noexcept`

TracerDS (changes requested):

Fernando-A-Rocha:
Nice

TracerDS (approved):
lgtm

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -16,6 +16,22 @@
...
+    MESSAGE_TYPE_WARNING = 2,
+    MESSAGE_TYPE_INFO = 3,
+    MESSAGE_TYPE_CUSTOM = 4,
+};
+
+enum DebugScriptLevels
+{
+    ERRORS_ONLY = 1,
+    ERRORS_AND_WARNINGS = 2,
+    ALL = 3,
```
botder: These enum constants are unscoped and leak into the global namespace

ghost (approved):
Looks good 🚀

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -15,6 +15,22 @@
...
+    MESSAGE_TYPE_CUSTOM = 4,
+};
+
+enum CScriptDebugging::DebugScriptLevels : std::uint32_t
+{
+    ERRORS_ONLY = 1,
+    ERRORS_AND_WARNINGS = 2,
+    ALL = 3,
+};
+
```
FileEX: I think these enums should be in the header? For ```DebugMessageLevels``` you don't need to number the values ​​sequentially, as that will be the default

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -73,6 +73,7 @@ class CScriptDebugging
                    unsigned char ucGreen = 255, unsigned char ucBlue = 255);
 
     void PrintLog(const char* szText);
+    bool CheckForSufficientDebugLevel(std::uint32_t playerDebugLevel, std::uint32_t messageDebugLevel) const noexcept;
```
FileEX: ```std::uint8_t``` will be better in this case, we do not need such a large range and therefore take up 4 bytes of memory. uint8_t has a range from 0 to 255, which is sufficient here

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -83,4 +84,7 @@ class CScriptDebugging
     SLuaDebugInfo                  m_SavedLuaDebugInfo;
     std::list<CLuaMain*>           m_LuaMainStack;
     CDuplicateLineFilter<SLogLine> m_DuplicateLineFilter;
+
+    enum DebugMessageLevels : std::uint32_t;
+    enum DebugScriptLevels : std::uint32_t;
```
FileEX: std::uint8_t

FileEX (changes requested):

Server/mods/deathmatch/logic/CScriptDebugging.h
ghost: Done

Server/mods/deathmatch/logic/CScriptDebugging.h
ghost: Done

Server/mods/deathmatch/logic/CScriptDebugging.cpp
ghost: They were previously in header, but botder suggested to move them to .cpp (to keep .h cleaner)

TracerDS: Also, if you dont need to get the values of each enum field (like in a switch case) then use `enum class` instead of `enum`

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -83,4 +84,7 @@ class CScriptDebugging
     SLuaDebugInfo                  m_SavedLuaDebugInfo;
     std::list<CLuaMain*>           m_LuaMainStack;
     CDuplicateLineFilter<SLogLine> m_DuplicateLineFilter;
+
+    enum DebugMessageLevels : std::uint8_t;
+    enum DebugScriptLevels : std::uint8_t;
```
botder: You don't need these here.

ghost: Done

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -15,6 +15,22 @@
...
+    MESSAGE_TYPE_INFO = 3,
+    MESSAGE_TYPE_CUSTOM = 4,
+};
+
+enum class DebugScriptLevels : std::uint8_t
+{
+    ERRORS_ONLY = 1,
+    ERRORS_AND_WARNINGS = 2,
+    ALL = 3,
+};
```
TracerDS: No need for `: std::uint8_t` and `= 1,`, `= 2,`, `= 3,`

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -143,16 +159,38 @@ void CScriptDebugging::PrintLog(const char* szText)
...
+            break;
+        case DebugMessageLevels::MESSAGE_TYPE_WARNING:
+            sufficientDebugLevel = (playerDebugLevel >= static_cast<std::uint8_t>(DebugScriptLevels::ERRORS_AND_WARNINGS));
+            break;
+        case DebugMessageLevels::MESSAGE_TYPE_INFO:
+        case DebugMessageLevels::MESSAGE_TYPE_CUSTOM:
+        case DebugMessageLevels::MESSAGE_TYPE_DEBUG:
+            sufficientDebugLevel = (playerDebugLevel == static_cast<std::uint8_t>(DebugScriptLevels::ALL));
+            break;
+    }
```
TracerDS: Change
```cpp
bool CScriptDebugging::CheckForSufficientDebugLevel(std::uint8_t playerDebugLevel, std::uint8_t messageDebugLevel) const noexcept
```
to
```cpp
bool CScriptDebugging::CheckForSufficientDebugLevel(DebugMessageLevels playerDebugLevel, DebugMessageLevels messageDebugLevel) const noexcept
```
That way you can remove all static_asserts

TracerDS (changes requested):

Server/mods/deathmatch/logic/CScriptDebugging.cpp
ghost: It wouldn't be so easy, it would require refactoring other parts of codebase, which would be out of scope for this PR. It was supposed to be small fix (tweak), not refactor. I also don't have will to do so, sorry - feel free to introduce any necessary changes afterwards. I've reverted it to default enum, it was already problematic to do with enum class, hence i asked @FileEX for help.

Fernando-A-Rocha:
:shipit:

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -15,6 +15,23 @@
 
 extern CGame* g_pGame;
 
+enum CScriptDebugging::DebugScriptLevels : std::uint8_t
```
botder:
```suggestion
enum DebugScriptLevels : std::uint8_t
```

Server/mods/deathmatch/logic/CScriptDebugging.cpp
```diff
@@ -15,6 +15,23 @@
...
+{
+    NONE,
+    ERRORS_ONLY,
+    ERRORS_AND_WARNINGS,
+    ALL,
+};
+
+enum CScriptDebugging::DebugMessageLevels : std::uint8_t
```
botder:
```suggestion
enum DebugMessageLevels : std::uint8_t
```

Server/mods/deathmatch/logic/CScriptDebugging.h
```diff
@@ -83,4 +84,7 @@ class CScriptDebugging
     SLuaDebugInfo                  m_SavedLuaDebugInfo;
     std::list<CLuaMain*>           m_LuaMainStack;
     CDuplicateLineFilter<SLogLine> m_DuplicateLineFilter;
+
+    enum DebugScriptLevels : std::uint8_t;
+    enum DebugMessageLevels : std::uint8_t;
```
botder:
```suggestion
```

botder (changes requested):
You really don't need the enumerations in the header file. You don't use them there.

botder (approved):
I am not going to nag any further, but you should know that `CScriptDebugging::CheckForSufficientDebugLevel` could be static (can also be a simple free-standing function in the .cpp file) because you don't access/modify any class member in the function.

PR #3507, opened 2024-06-24
Fix #414 Using setElementHealth on a dead ped makes it invincible

### LINKED ISSUE #414: Using setElementHealth on a dead ped makes it invincible
**Describe the bug**
The only way to respawn a ped is with setElementHealth, yet after calling setElementHealth on a dead ped, the server returns `true` for `isPedDead`. You can't damage the ped any more even though `ped.health == 100`.

**To Reproduce**
```
crun setElementPosition(localPlayer, 1968, -2033, 13)
srun a = createPed(0, 1967, -2025, 13)
-- Kill the ped
srun isPedDead(a) -- true
srun setElementHealth(a, 100)
srun isPedDead(a) -- true (should be false)
-- Shoot the ped, it won't take damage
```

**Expected behaviour**
@qaisjp said:
> Setting the health of a ped to a `health > 0` should respawn the ped.

**Additional context**
Do this after calling setElementHealth: setElementSyncer(a, true) and the ped will be able to take damage again, but onPedWasted isn't triggered.

From https://bugs.mtasa.com/view.php?id=8790

FileEX:
Fixed #414 

My previous PR "fixing" this bug was a long time ago, when I had no experience. Now I have solved this problem and I know what causes it.

Ped has locked health and armor when he dies. When the ped is re-created (ReCreateModel), health and armor are not unlocked. This is proven by the fact that when we move away from the invincible ped and come back again (stream out & stream in), the ped has unlocked health and armor and receives damage. This is because when stream in with ped, his health and armor are unlocked.

https://github.com/multitheftauto/mtasa-blue/blob/2c4e23211346248c836bbcac57baa2d5cf74be1b/Client/mods/deathmatch/logic/CPedSync.cpp#L134-L135

Additionally, the "isDead" state is now also set correctly after recreating a ped.

I don't think this should cause any synchronization problems.

### DIFF
--- a/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -1669,6 +1669,9 @@ bool CStaticFunctionDefinitions::SetElementHealth(CElement* pElement, float fHea
                 unsigned char ucHealth = static_cast<unsigned char>(fHealth * 1.25f);
                 fHealth = static_cast<float>(ucHealth) / 1.25f;
                 pPed->SetHealth(fHealth);
+
+                if (pPed->IsDead() && fHealth > 0)
+                    pPed->SetIsDead(false);
             }
             else
                 return false;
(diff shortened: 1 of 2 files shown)


Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1669,6 +1669,9 @@ bool CStaticFunctionDefinitions::SetElementHealth(CElement* pElement, float fHea
                 unsigned char ucHealth = static_cast<unsigned char>(fHealth * 1.25f);
                 fHealth = static_cast<float>(ucHealth) / 1.25f;
                 pPed->SetHealth(fHealth);
+
+                if (pPed->IsDead() && fHealth > 0)
```
tederis: There should be a floating-point literal.

PR #3510, opened 2024-06-25
Fix #501 Setting player model resets their current weapon slot

### LINKED ISSUE #501: Setting player model resets their current weapon slot
**Describe the bug**
setElementModel() on a player sets his current weapon slot to 0.

**To Reproduce**
- Go to map editor, start runcode
- Get yourself a MP5 or any other weapon
- use `srun setElementModel(me, 160)`
- You are no longer equipped with the MP5, you have to switch back to it

**Expected behavior**
setElementModel should have no influence on the players weapon slot.

**Multi Theft Auto v1.5.6-release-14394**

**Additional information**
From https://bugs.mtasa.com/view.php?id=9671

FileEX:
Fixed #501 

In the case of non-player peds weapon the slot is set correctly, but in the case of player peds the slot is not set.

This is due to the fact that the slot for the player is taken directly from the GTA memory

https://github.com/multitheftauto/mtasa-blue/blob/97920d83de400e0378f73ea93e87b27cf59dd758/Client/mods/deathmatch/logic/CClientPed.cpp#L2209-L2211

and when the skin respawns, the slot in the game memory is still the one we had before changing the skin. This causes the condition result is false.

https://github.com/multitheftauto/mtasa-blue/blob/97920d83de400e0378f73ea93e87b27cf59dd758/Client/mods/deathmatch/logic/CClientPed.cpp#L2148-L2152

As a result, the weapon slot is not set at all. Therefore, we first set the weapon slot to 0 to make this condition true.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -5220,6 +5220,9 @@ void CClientPed::Respawn(CVector* pvecPosition, bool bRestoreState, bool bCamera
 
             m_pPlayerPed->SetLanding(false);
 
+            // Set it to 0 (Fix #501)
+            SetCurrentWeaponSlot(eWeaponSlot::WEAPONSLOT_TYPE_UNARMED);
+
             if (bRestoreState)
             {
                 // Jax: restore all the things we saved


Proxy-99:
```c++
 // Set it to 0 (Fix #501)
            SetCurrentWeaponSlot(eWeaponSlot::WEAPONSLOT_TYPE_UNARMED);

            if (bRestoreState)
            {
                // Jax: restore all the things we saved
                SetHealth(fHealth);
                SetArmor(fArmor);
                SetCurrentWeaponSlot(weaponSlot);
                SetCurrentRotation(fCurrentRotation);
                m_pPlayerPed->SetTargetRotation(fTargetRotation);
                SetMoveSpeed(vecMoveSpeed);
                SetHasJetPack(m_bHasJetPack);
                SetInterior(ucInterior);
            }
``` 
what about this condition after it set the weapon slot should not `bRestoreState `become true so it loop through the list

FileEX:
I do not understand what you mean. After changing the skin, the slot changes to 0 anyway, but the MTA does not recognize it, so setting the slot to 0 from the MTA level solves the problem.

Proxy-99:
> I do not understand what you mean. After changing the skin, the slot changes to 0 anyway, but the MTA does not recognize it, so setting the slot to 0 from the MTA level solves the problem.

you can notice a condtion after your fix 
at line 5226
it alredy set the weapon slot 
SetCurrentWeaponSlot(weaponSlot);

```
if (bRestoreState)
{
    // Jax: restore all the things we saved
    SetHealth(fHealth);
    SetArmor(fArmor);
    SetCurrentWeaponSlot(weaponSlot);
    SetCurrentRotation(fCurrentRotation);
    m_pPlayerPed->SetTargetRotation(fTargetRotation);
    SetMoveSpeed(vecMoveSpeed);
    SetHasJetPack(m_bHasJetPack);
    SetInterior(ucInterior);
}
```
so setting bRestoreState to true should fix the issue if I am not wrong

FileEX:
You're wrong. I described what the problem is. ``bRestoreState`` is true, but still the weapon slot is not set correctly without my fix.

https://github.com/multitheftauto/mtasa-blue/blob/97920d83de400e0378f73ea93e87b27cf59dd758/Client/mods/deathmatch/logic/CClientPed.cpp#L3907

Proxy-99 (approved):
good job!
mainly the issue comes from ` if (weaponSlot != GetCurrentWeaponSlot()) ` returning false, I was confused at the start

Einheit-101:
Is the fix for this issue somehow usable to also fix the problem that equipping / removing a jetpack also resets the current weapon slot?

FileEX:
> Is the fix for this issue somehow usable to also fix the problem that equipping / removing a jetpack also resets the current weapon slot?

No, but it’s not a bug. It’s a SA feature - it intentionally sets the appropriate weapon for the jetpack. I think we could add a special world property `jetpack_autoweaponchange` set to true by default. If set to false, the jetpack wouldn’t change the weapon.

PR #3511, opened 2024-06-25
Fix #539 Changing player position when he/she has a jetpack will remove the jetpack and bug when skin is changed

### LINKED ISSUE #539: Changing player position when he/she has a jetpack will remove the jetpack and bug when skin is changed
**Describe the bug**
As the title says, if you change the position of a player while he/she is using a jetpack, the jetpack will be removed, but later given back if the skin is changed.

**To reproduce**
```lua
addCommandHandler ( "bugme",
	function ( thePlayer )
		local x, y, z = getElementPosition ( thePlayer )
		givePedJetPack ( thePlayer )
		setElementPosition ( thePlayer, x + 1, y + 1, z )
	end
)

addCommandHandler ( "bugme2",
	function ( thePlayer )
		local skins = getValidPedModels ( )
		setElementModel ( thePlayer, skins [ math.random ( #skins ) ] )
	end
)
```

**Expected behaviour**
No removal of jetpack.

**MTA version:**
Multi Theft Auto v1.5.8-release-20804

**Additional context**
From https://bugs.mtasa.com/view.php?id=8804

FileEX:
Fixed #539 

Jetpack is removed because the GTA function that is called ``CPed::Teleport`` aborts all tasks (including jetpack).

Then when changing the skin, the jetpack is set based on the m_bHasJetPack variable, which is true because its state is not changed when the position is changed, therefore the new skin gets the jetpack that was removed by GTA, but the MTA still knows that the player has the jetpack because he doesn't know that GTA removed it.

https://github.com/multitheftauto/mtasa-blue/blob/97920d83de400e0378f73ea93e87b27cf59dd758/Client/mods/deathmatch/logic/CClientPed.cpp#L5232

Therefore, we first check whether the player has a jetpack before changing position, and then set it after changing position depending on whether he had it before.


Fernando-A-Rocha:
Good stuff

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -607,8 +607,14 @@ void CClientPed::Teleport(const CVector& vecPosition)
...
                 }
 
+                // Player has jetpack?
+                bool hasJetpack = HasJetPack();
+
                 // Set the real position
                 m_pPlayerPed->Teleport(vecPosition.fX, vecPosition.fY, vecPosition.fZ);
+
+                // Restore jetpack
+                SetHasJetPack(hasJetpack);
```
Proxy-99, line 620: what about passing m_bHasJetPack directly and removing `bool hasJetpack = HasJetPack();`
`SetHasJetPack(m_bHasJetPack);`

FileEX: As I described - m_bHasJetpack does not change its state when changing the skin. HasJetpack is more reliable because it is based on the current task

Proxy-99: I do not get it I tried with the code below while passing  SetHasJetPack(m_bHasJetPack);

```lua
--local x, y, z = getElementPosition ( localPlayer )
		
		setElementModel (localPlayer, math.random(255) )
		
		print(isPedWearingJetpack (localPlayer))
``` 
it working normally when checking for the jetpack while replacing model in game
but I am not sure if it the same variable is passed while using isPedWearingJetpack

FileEX: My mistake - As you can read in the title and in the indicated issue, the problem concerns changing the position with the jetpack

luccabassoli (approved):

joaopdsilvadias (approved):

fhostings (approved):

botder:
The code looks fine, but I don't think we want to introduce desync (new client sees jetpack, older clients do not) right now

FileEX:
> The code looks fine, but I don't think we want to introduce desync (new client sees jetpack, older clients do not) right now

Will simply taking a jetpack when changing positions solve the problem?

botder:
I would keep the jetpack on the player.

FileEX:
> I would keep the jetpack on the player.

So how can I solve the desync problem? Send a packet to the server so that everyone can see the jetpack?

FileEX:
@botder I think I solved the desync problem by sending a GIVE_PED_JETPACK packet. So now it is no longer bakcwards-incompatible

ghost:
"Changing player position when he/she has a jetpack will remove the jetpack and bug when skin is changed" 

If you set the `warp` param to false in `setElementPosition` it will preserve the jetpack animation and won't take away the player's jetpack. If the scripter uses this option when teleporting players with jetpack, then he/she won't have to worry about skin changes.

Video: https://youtu.be/BLVGxOA0NIY

FileEX:
> "Changing player position when he/she has a jetpack will remove the jetpack and bug when skin is changed"
> 
> If you set the `warp` param to false in `setElementPosition` it will preserve the jetpack animation and won't take away the player's jetpack. If the scripter uses this option when teleporting players with jetpack, then he/she won't have to worry about skin changes.
> 
> Video: https://youtu.be/BLVGxOA0NIY

Yeah, this happens because if the warp argument is set to false, only the position vector in the matrix/transform of the given entity is changed. However, when warp is true, a GTA function is called, specifically ``CPed::Teleport``. This function aborts all tasks, including the jetpack. In the case of teleportation, the ped is literally recreated: it’s removed from the world, its position is changed, and it’s added back to the world. 

https://github.com/gta-reversed/gta-reversed/blob/6e2e236c9f46a049d30c3b5097c275c8b15eeae4/source/game_sa/Entity/Ped/Ped.cpp#L3390-L3406

PR #3514, opened 2024-06-27
Appendum #3490

TracerDS:
Appendum #3490 

`std::variant` recognized `""` as `true` instead of a `std::string`.

### DIFF
--- a/Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
+++ b/Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
@@ -711,7 +711,7 @@ std::variant<bool, std::string> CLuaFileDefs::fileRead (
         -> std::variant<bool, std::string>
     {
         if (count == 0)
-            return std::string();
+            return std::string{};
 
         SString buffer;
         auto    bytesRead = pFile->Read(count, buffer);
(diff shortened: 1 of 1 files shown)


Shared/mods/deathmatch/logic/luadefs/CLuaFileDefs.cpp
```diff
@@ -711,7 +711,7 @@ std::variant<bool, std::string> CLuaFileDefs::fileRead (
         -> std::variant<bool, std::string>
     {
         if (count == 0)
-            return "";
+            return std::string();
```
tederis: I would use curly braces here.

TracerDS: you mean `return {}` or
```cpp
if (count == 0) {
    return std::string();
}
```

tederis: I mean `return std::string{};`

TracerDS: Whats the difference?

tederis: Although semantically it's the same, curly braces are encouraged to use with latest C++ version. You could find details about it yourself. But it's just a recommendation.

TracerDS: Done

tederis (approved):

botder (approved):

PR #3517, opened 2024-06-27
Fix #452 Satchels get detached when skin is changed 

### LINKED ISSUE #452: Satchels get detached when skin is changed
**Describe the bug**
As the title says, when the player skin changes, all the satches he/she has attached, they'll be detached from the player.

I used setElementModel to change the skin.

**To Reproduce**
Throw a satchel at a player, change his/her skin, move away a little.

**Expected behavior**
No detachment of satchel.

**Screenshots**

v1.5.5-release-13804

v1.5.5-release-13804

**Additional context**
From https://bugs.mtasa.com/view.php?id=8773

FileEX:
Fixed #452


Client/game_sa/CPedSA.h
```diff
@@ -408,4 +408,6 @@ class CPedSA : public virtual CPed, public virtual CPhysicalSA
     void*                   GetPedNodeInterface(std::int32_t nodeId) { return reinterpret_cast<CPedSAInterface*>(m_pInterface)->pedNodes[nodeId]; }
     std::unique_ptr<CPedIK> GetPedIK() { return std::make_unique<CPedIKSA>(GetPedIKInterface()); }
     static void             StaticSetHooks();
+
+    void GetAttachedSatchels(std::vector<SSatchelsData> &satchelsList);
```
TracerDS: why not return a vector?

Client/sdk/game/CPed.h
```diff
@@ -277,4 +286,6 @@ class CPed : public virtual CPhysical
     virtual CPedIKSAInterface*      GetPedIKInterface() = 0;
     virtual void*                   GetPedNodeInterface(std::int32_t nodeId) = 0;
     virtual std::unique_ptr<CPedIK> GetPedIK() = 0;
+
+    virtual void GetAttachedSatchels(std::vector<SSatchelsData> &satchelsList) = 0;
```
TracerDS: noexcept?

TracerDS (changes requested):

Client/game_sa/CPedSA.h
FileEX: It is better to avoid copying the entire vector. In addition, let's keep it consistent with the project structure. Look at functions like GetPosition, GetMoveSpeed ​​etc.

Client/sdk/game/CPed.h
FileEX: I'm not sure about this. Theoretically, there is a chance that an exception will occur here, for example, during vector allocation, and the program might terminate abruptly without providing any specific information about what went wrong. Adding noexcept to this function seems impractical to me because it implies that the function guarantees not to throw any exceptions. However, operations on std::vector, such as allocation and reallocation, can potentially throw exceptions if memory allocation fails. Therefore, I prefer not to mark this function as noexcept

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -3987,6 +3991,21 @@ void CClientPed::_ChangeModel()
             _CreateModel();
         }
 
+        // ReAttach satchels
+        static CClientProjectileManager* pProjectileManager = m_pManager->GetProjectileManager();
```
tederis: I would avoid using static here. GetProjectileManager is cheap and there's no need in caching it.

Client/sdk/game/CPed.h
```diff
@@ -15,6 +15,7 @@
 #include "Common.h"
 #include "CPhysical.h"
 #include "CWeaponInfo.h"
+#include <vector>
```
tederis: Is this really needed here? From what I can see `vector` is already included in PCH of both game_sa and deathmatch.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -3987,6 +3991,21 @@ void CClientPed::_ChangeModel()
...
+        {
+            CClientProjectile* pSatchel = pProjectileManager->Get((CEntitySAInterface*)satchelData.pProjectileInterface);
+            if (!pSatchel || pSatchel->IsBeingDeleted())
+                continue;
+
+            pSatchel->SetAttachedOffsets(*satchelData.vecAttachedOffsets, *satchelData.vecAttachedRotation);
+            pSatchel->InternalAttachTo(this);
+        }
+
+        attachedSatchels.clear();
```
tederis: This can be removed since `attachedSatchels` won't be used down to the function's end.

Client/game_sa/CPedSA.h
tederis: Please, add `const` qualifier and `override` specifier.

Client/sdk/game/CPed.h
tederis: Please, add `const` qualifier.

Client/sdk/game/CPed.h
FileEX: If it wasn't needed, I wouldn't add it. Without this I get the error "std namespace element does not have a vector member".

tederis: Okay. I'm just wondering during the compilation of which project exactly this error was occured?

FileEX: However, it's Visual Studio that goes crazy - Compilation, runs fine but Visual Studio shows this error
image

TracerDS: > However, it's Visual Studio that goes crazy - Compilation, runs fine but Visual Studio shows this error image

then the vector class was not found in the current translation unit.
I guess you can no longer rely on pch :P

PR #3520, opened 2024-06-29
Fix #467 Ped animations don't sync for new players

### LINKED ISSUE #467: Ped animations don't sync for new players
**Bug**
The ped animations don't sync for new players, it gets reseted for them.
For players that remain in the server the animation remain there.

**How to reproduce**
(only server-side)
1. create a ped
2. set an animation for the ped
3. reconnect and the ped will have no animation set

**Example image**

mta-screen_2018-09-15_13-58-39

FileEX:
Fixed #467


Dutchman101:
Awesome! Hope this gets a bunch of code reviews..

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -316,6 +355,7 @@ class CPed : public CElement
     bool                                 m_bFrozen;
     bool                                 m_bStealthAiming;
     CVehicle*                            m_pJackingVehicle;
+    SPlayerAnimData                      m_pAnimData;
```
tederis: It's not a pointer. Should be `m_AnimData`.

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -278,6 +314,9 @@ class CPed : public CElement
     std::vector<CPlayer*>::const_iterator NearPlayersIterBegin() { return m_nearPlayersList.begin(); }
     std::vector<CPlayer*>::const_iterator NearPlayersIterEnd() { return m_nearPlayersList.end(); }
 
+    const SPlayerAnimData& GetAnimationData() const noexcept { return m_pAnimData; };
+    void                   SetAnimationData(SPlayerAnimData animData) noexcept { m_pAnimData = animData; };
```
tederis: `const SPlayerAnimData& animData` or `m_pAnimData = std::move(animData)` (with the corresponding changes within `SPlayerAnimData`).

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -103,6 +103,42 @@ enum eBone
...
+    bool        updatePosition;
+    bool        interruptable;
+    bool        freezeLastFrame;
+    int         blendTime;
+    bool        taskToBeRestoredOnAnimEnd;
+
+    SPlayerAnimData()
+        : blockName(""),
+          animName(""),
+          time(-1),
```
tederis: Please, use default member initialization instead of the member initializer list. I mean `bool loop{true};`, for example.

tederis (changes requested):

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -103,6 +103,33 @@ enum eBone
     BONE_RIGHTFOOT
 };
 
+struct SPlayerAnimData
+{
+    std::string blockName{""};
+    std::string animName{""};
```
tederis: Initialization by C-string isn't necessary here.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -959,6 +959,25 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
                         BitStream.Write(currentWeaponSlot);
                     }
 
+                    // Aniamtion
```
tederis: A typo

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -959,6 +959,25 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
...
+                    if (BitStream.Can(eBitStreamVersion::AnimationsSync))
+                    {
+                        const SPlayerAnimData& animData = pPed->GetAnimationData();
+
+                        if (!animData.blockName.empty() && !animData.animName.empty())
+                        {
+                            BitStream.WriteString(animData.blockName);
```
tederis, line 991: You have to write a **bit** that signalling whether this packet contains animation data or not.

tederis (changes requested):

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
FileEX: What is the difference in this case?

tederis: Because you have a packet corruption now.

FileEX: I understand, the idea is that later data will be incorrect due to "holes" when reading?

tederis: Yes, without that **bit** a client doesn't know what exactly is written inside a packet. Because a string itself doesn't contain sufficient information. You should explicitly mark it.

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -278,6 +305,9 @@ class CPed : public CElement
     std::vector<CPlayer*>::const_iterator NearPlayersIterBegin() { return m_nearPlayersList.begin(); }
     std::vector<CPlayer*>::const_iterator NearPlayersIterEnd() { return m_nearPlayersList.end(); }
 
+    const SPlayerAnimData& GetAnimationData() const noexcept { return m_animData; };
+    void                   SetAnimationData(const SPlayerAnimData& animData) noexcept { m_animData = animData; };
```
tederis: One more thing. This method can potentially throw bad_alloc exception. I would not mark it as noexcept.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -959,6 +959,25 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
...
 
+                    // Animation
+                    if (BitStream.Can(eBitStreamVersion::AnimationsSync))
+                    {
+                        const SPlayerAnimData& animData = pPed->GetAnimationData();
+
+                        // Contains animation data?
+                        BitStream.WriteBit(!animData.blockName.empty() && !animData.animName.empty());
+
+                        BitStream.WriteString(animData.blockName);
```
tederis: You forgot a condition here. You can see that a client doesn't expect the data if a bit is set to zero.

tederis:
Good job. But there's a problem. Animations can be interrupted on a client and that information won't be available for a server. This will break the animation consistency and lead to a desync for just joined new players. There have to be a mechanism for an animation's data invalidation.

FileEX:
I introduced an animation validation mechanism. It supports:

- Interrupted animation on the client side (damage, explosion, death, etc.).

- Animation time has ended

- Animation was overwritten/changed by setPedAnimation on the client side

- Animation was killed by setPedAnimationProgress

I tested it on a local server with peds and the tests were successful.

Fernando-A-Rocha (approved):
Congratulations

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -788,4 +790,7 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
     CClientPed*   m_pGettingJackedBy;                    // The ped that is jacking us
 
     std::shared_ptr<CClientModel> m_clientModel;
+
+    bool m_hasSyncedAnim;
+    bool m_animationOverridedByClient;
```
botder: You should zero-initialize these member fields.
```suggestion
    bool m_hasSyncedAnim{};
    bool m_animationOverridedByClient{};
```

Client/mods/deathmatch/logic/CClientPed.h
```diff
@@ -553,6 +553,8 @@ class CClientPed : public CClientStreamElement, public CAntiCheatModule
 
     std::unique_ptr<CAnimBlendAssociation> GetAnimAssociation(CAnimBlendHierarchySAInterface* pHierarchyInterface);
 
+    void SetHasSyncedAnim(bool synced) noexcept { m_hasSyncedAnim = synced; };
+    bool HasSyncedAnim() const noexcept { return m_hasSyncedAnim; };
```
botder: Removed the semicolon after the last brace.
```suggestion
    void SetHasSyncedAnim(bool synced) noexcept { m_hasSyncedAnim = synced; }
    bool HasSyncedAnim() const noexcept { return m_hasSyncedAnim; }
```

Shared/sdk/net/bitstream.h
```diff
@@ -556,6 +556,10 @@ enum class eBitStreamVersion : unsigned short
     // 2024-06-16
     PedSync_Revision,
 
+    // Ped animations synchronization
+    // 2024-06-29
+    AnimationsSync,
+
```
botder: You have to move this to the end of the list before merge.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -978,6 +978,42 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
                         BitStream.Write(currentWeaponSlot);
                     }
 
+                    // Animation
+                    if (BitStream.Can(eBitStreamVersion::AnimationsSync))
+                    {
+                        const SPlayerAnimData& animData = pPed->GetAnimationData();
+
+                        // Contains animation data?
+                        bool animRunning = !animData.blockName.empty() && !animData.animName.empty();
```
botder: Nit-picking here, but you could move this decision logic, whether an animation is running or not, to the SPlayerAnimData class: `animData.IsAnimating()`

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -978,6 +978,42 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
...
+
+                        // Is animation still running?
+                        float deltaTime = GetTickCount64_() - animData.startedTick;
+                        if (!animData.freezeLastFrame && animData.time > 0 && deltaTime >= animData.time)
+                        {
+                            animRunning = false;
+                            pPed->SetAnimationData({});
```
botder: A little bit strange to manipulate the ped's state here in the packet writing function.

Server/mods/deathmatch/logic/CPed.h
```diff
@@ -103,6 +103,41 @@ enum eBone
...
+          time(time),
+          loop(loop),
+          updatePosition(updatePos),
+          interruptable(interrupt),
+          freezeLastFrame(freeze),
+          blendTime(blend),
+          taskToBeRestoredOnAnimEnd(taskRestore),
+          startedTick(tick),
+          progress(0.0f),
+          speed(1.0f){};
```
botder: Just terrible to whoever is going to read the code in the future. Remove this constructor and assign the fields directly. You also don't have to assign `progress` and `speed` with the same value.

botder (changes requested):

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
FileEX: Would ``CPedSync::DoPulse`` be the appropriate place?

botder: I don't know enough to judge that, if you think it's a good spot then use it. I only have an issue with it when it happens in the packet writing logic.

FileEX: Done and tested

botder (dismissed):

botder (approved):

PR #3521, opened 2024-06-29
Fix setElementModel resets goggle effect

### LINKED ISSUE #2441: setElementModel resets goggle effect
**Describe the bug**
Using setElementModel resets the goggle effect.

**To reproduce**
crun Camera.goggleEffect = "thermalvision"
srun me.model = 1

**Expected behaviour**
Goggle effect should be retained after changing the player model

**Version**
Multi Theft Auto v1.5.9-release-21008

Proxy-99:
Fixes #2441

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -5210,6 +5210,7 @@ void CClientPed::Respawn(CVector* pvecPosition, bool bRestoreState, bool bCamera
             float         fTargetRotation = m_pPlayerPed->GetTargetRotation();
             unsigned char ucInterior = GetInterior();
             unsigned char ucCameraInterior = static_cast<unsigned char>(g_pGame->GetWorld()->GetCurrentArea());
+            std::string   sCameraGoogleEffect = CLuaCameraDefs::GetCameraGoggleEffect();
 
             // Don't allow any camera movement if we're in fixed mode
             if (m_pManager->GetCamera()->IsInFixedMode())
@@ -5237,6 +5238,7 @@ void CClientPed::Respawn(CVector* pvecPosition, bool bRestoreState, bool bCamera
             }
             // Restore the camera's interior whether we're restoring player states or not
             g_pGame->GetWorld()->SetCurrentArea(ucCameraInterior);
+            CLuaCameraDefs::SetGoggleEffect(sCameraGoogleEffect);
 
             // Reattach us
             if (pAttachedTo && pAttachedTo->IsEntityAttached(this))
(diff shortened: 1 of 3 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
```diff
@@ -144,6 +144,26 @@ unsigned char CLuaCameraDefs::GetCameraInterior()
     return ucInterior;
 }
 
+void CLuaCameraDefs::SetGoggleEffect(std::string CameraGoogleEffect)
```
FileEX: Rename it to ``SetCameraGoggleEffect`` please

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -5210,6 +5210,7 @@ void CClientPed::Respawn(CVector* pvecPosition, bool bRestoreState, bool bCamera
             float         fTargetRotation = m_pPlayerPed->GetTargetRotation();
             unsigned char ucInterior = GetInterior();
             unsigned char ucCameraInterior = static_cast<unsigned char>(g_pGame->GetWorld()->GetCurrentArea());
+            std::string   sCameraGoogleEffect = CLuaCameraDefs::GetCameraGoggleEffect();
```
tederis: This is bad. CLuaCameraDefs(as well as other CLua*Defs classes) is only responsible for the interaction with Lua. It must not be available from other classes.

Client/mods/deathmatch/logic/luadefs/CLuaCameraDefs.cpp
tederis: All functions inside `CLua*Defs` classes must be binded to Lua.

Client/mods/deathmatch/logic/CClientPed.cpp
Proxy-99: so where I move it to class CCameraSA?

tederis: Maybe something like this?
```
bool bOldNightVision = g_pMultiplayer->IsNightVisionEnabled();
bool bOldThermalVision = g_pMultiplayer->IsThermalVisionEnabled();

...

g_pMultiplayer->SetNightVisionEnabled(bOldNightVision, true)
g_pMultiplayer->SetThermalVisionEnabled(bOldThermalVision, true)
```

tederis:
This PR doesn't take into account a possibility of manual goggle effects change. For example, a player may put nightvision goggles on, then execute `setCameraGoggleEffect("nightvision", false)`, change a model and it will give a nightvision effect with the noise enabled, although there should be an effect without any noise.

Proxy-99:
> This PR doesn't take into account a possibility of manual goggle effects change. For example, a player may wear nightvision goggles up, then execute `setCameraGoggleEffect("nightvision", false)`, change a model and it will give a nightvision effect with the noise enabled, although there should be an effect without any noise.

I think if we already have the effect enabled there is no need to play the sound again so should all of them be false 
see this sound played once 
[https://youtu.be/-45qz-iUuJI?si=W7YIFVIKv4h5pX4w](https://youtu.be/-45qz-iUuJI?si=W7YIFVIKv4h5pX4w)

FileEX (approved):

PR #3522, opened 2024-06-29
Fix #512 setPedAnimation isn't synced properly on server side

### LINKED ISSUE #512: setPedAnimation isn't synced properly on server side
**Describe the bug**
Some arguments which work for the client `setPedAnimation` don't work on the server.

**To Reproduce**

```lua
crun setPedAnimation(me, "PAULNMAC", "wank_loop", nil, true, false, false);

-- Above example works fine, ped doesn't move when animation loops

run setPedAnimation(me, "PAULNMAC", "wank_loop", nil, true, false, false);

-- above example doesn't work, animation works but ped's position gets changed according to animation, and it shouldn't
```

Argument which is not synced in this case is updatePosition, not sure about others.

**Expected behavior**

1.5.2

1.5.2

**Additional context**
From https://bugs.mtasa.com/view.php?id=9176

FileEX:
Fixed #512 

The bug was that the old parser, when reading bool (ReadBool), seeing nil, set default values ​​(in this case, true for updatePositions if nil was passed). 

https://github.com/multitheftauto/mtasa-blue/blob/f7ce562b645cb05a18658df62d093b753b881bb9/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp#L419-L437

Moving the function to a new parser solves the problem.


Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -64,7 +65,7 @@ void CLuaPedDefs::LoadFunctions()
         {"warpPedIntoVehicle", WarpPedIntoVehicle},
         {"removePedFromVehicle", RemovePedFromVehicle},
         {"setPedDoingGangDriveby", SetPedDoingGangDriveby},
-        {"setPedAnimation", SetPedAnimation},
+        {"setPedAnimation", ArgumentParser<SetPedAnimation>},
```
tederis: `ArgumentParserWarn<false, SetPedAnimation>`. Otherwise it will be backward incompatible.

tederis (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
FileEX: What exactly do you mean? I tried several different examples and they all worked

tederis: `ArgumentParser<>` will generate a Lua error if any exception will be thrown. `ArgumentParserWarn<false, ...>` will return **false** in the same situation. Previously created scripts expect a **false**, not a Lua error. That's why `ArgumentParser<>` makes this function backward incompatible.

FileEX: Yes, in the case of ArgumentParser nil is returned instead of false. Thanks for the clarification

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -404,56 +405,33 @@ int CLuaPedDefs::IsPedFrozen(lua_State* luaVM)
...
+        else if (std::holds_alternative<bool>(animName.value()))
+            if (std::get<bool>(animName.value()))
+                throw LuaFunctionError("Animation name cannot be true. Possible values: nil, false, string.");
     }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    lua_pushboolean(luaVM, false);
-    return 1;
+    return CStaticFunctionDefinitions::SetPedAnimation(pPed, animBlockName.empty() ? NULL : animBlockName.c_str(), animationName.empty() ? NULL : animationName.c_str(), time.value_or(-1), blendTime.value_or(250), loop.value_or(true), updatePosition.value_or(true), interruptable.value_or(true), freezeLastFrame.value_or(true), restoreTask.value_or(false));
```
tederis: Is the ternary operator really needed here? If a string is empty you can just pass it as is.

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -404,56 +405,33 @@ int CLuaPedDefs::IsPedFrozen(lua_State* luaVM)
...
-    else
-        iTime = -1;
-    argStream.ReadBool(bLoop, true);
-    argStream.ReadBool(bUpdatePosition, true);
-    argStream.ReadBool(bInterruptable, true);
-    argStream.ReadBool(bFreezeLastFrame, true);
-    argStream.ReadNumber(iBlend, 250);
-    argStream.ReadBool(bTaskToBeRestoredOnAnimEnd, false);
+    std::string animBlockName = "";
+    std::string animationName = "";
```
tederis: The initialization with an empty C-string is unnecessary.

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -404,56 +405,33 @@ int CLuaPedDefs::IsPedFrozen(lua_State* luaVM)
...
 
-    if (!argStream.HasErrors())
+    if (blockName.has_value())
     {
-        const char *szBlock, *szAnim;
-        szBlock = strBlockName.empty() ? NULL : strBlockName.c_str();
-        szAnim = strAnimName.empty() ? NULL : strAnimName.c_str();
+        if (std::holds_alternative<std::string>(blockName.value()))
```
tederis, line 418: Please, use `value_or`. Otherwise `std::bad_optional_access` exception may be thrown.

tederis (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
FileEX: You mean?
```cpp
if (std::holds_alternative<std::string>(blockName.value_or("")))
```

tederis: Yes

FileEX: Is this check insufficient? https://github.com/multitheftauto/mtasa-blue/blob/e6dd0c58379ced299afcf29d11b56fc46488e047/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp#L416

tederis: Or rather `if (std::holds_alternative<std::string>(blockName.value_or(std::string{})))` because `""` may be considered as a bool.

tederis: > Is this check insufficient?
> 
> https://github.com/multitheftauto/mtasa-blue/blob/e6dd0c58379ced299afcf29d11b56fc46488e047/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp#L416

Oh, yeah. Sorry, I didn't notice that condition.

tederis (approved):
LGTM

PR #3523, opened 2024-06-29
Fix #1860 Markers with type "arrow" or "checkpoint" won't be affected by alpha color argument 

### LINKED ISSUE #1860: Markers with type "arrow" or "checkpoint" won't be affected by alpha color argument
**Describe the bug**
Markers with type "arrow" or "checkpoint" won't be affected by alpha color argument. This means that there isn't a real way to make them invisible.

**To reproduce**
/run createMarker(0, 0, 3, "checkpoint", 5, 255, 255, 255, 0)
The checkpoint will be still visible.

**Expected behaviour**
The marker should be invisible

**Version**
Client: v1.5.8-release-20698.2.005
Server: v1.5.7-release-20488

**Additional context**
I suppose this is a problem with GTA itself and not MTA code. Idrk, too lazy to read the code itself. If that is the case, maybe it would be too much problem to fix it, and I don't think it is a good idea to fix this issue in a hacky way. Don't fixing this issue is not a problem.

FileEX:
Fixed #1860 

image

After merge #2989 , I will add the ability to change the color and alpha of the checkpoint marker target arrow. By default it is red with alpha 255.
image

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1529,6 +1529,10 @@ void CMultiplayerSA::InitHooks()
     MemPut<BYTE>(0x524084, 0xFF);
     MemPut<BYTE>(0x524089, 0xFF);
 
+    // Allow change alpha for arrow & checkpoint markers (#1860)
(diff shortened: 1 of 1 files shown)


ghost (approved):
LGTM

Issue #3454, opened 2024-06-10
Ability to customize cullzone


gta191977649:
### Is your feature request related to a problem? Please describe.

Cull zones are used to create zones affecting the different behaviours of the world, like mirrors, fixed camera positions, disabling rain, and police behaviours when the player is within the zone. However, these pre-defined zones can sometimes cause undesirable artifacts, particularly in heavily customized maps, e.g. total conversion.

As you can see:
![image](https://github.com/multitheftauto/mtasa-blue/assets/2988235/38799f8d-7b1b-438d-8960-38d6083501b5)

### Describe the solution you'd like

It would be nice if we offered some way to allow us toggle it, e.g. via setting [setWorldSpecialPropertyEnabled](https://wiki.multitheftauto.com/wiki/SetWorldSpecialPropertyEnabled) to turn or disable SA pre-defined cull zones and also provide some function like 
createCullzone(CenterX, CenterY, CenterZ, Unknown1, Length, Bottom, Width, Top, Flag ) , see [here](https://gtamods.com/wiki/CULL#endnote_3) to allows the customizability for manually create them.

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

Xenius97:
#2966

Fernando-A-Rocha:
#1192

PR #3526, opened 2024-06-30
Add ability to toggle tunnel weather blending effect

gta191977649:
This PR introduces a new special world property, "tunnelweatherblend", for the toggling weather blending effect when a player is in the tunnel.

This can be helpful for the server that runs on a total conversion map.

Solve: [3454](https://github.com/multitheftauto/mtasa-blue/issues/3454)

Example:
`setWorldSpecialPropertyEnabled ("tunnelweatherblend", false)` for disabling the weather blending effect.

### DIFF
--- a/Shared/sdk/net/bitstream.h
+++ b/Shared/sdk/net/bitstream.h
@@ -548,10 +548,6 @@ enum class eBitStreamVersion : unsigned short
     // 2024-05-23
     WorldSpecialProperty_ExtendedWaterCannons,
 
-    // Add "extendedwatercannons" to setWorldSpecialPropertyEnabled
-    // 2024-06-30
-    WorldSpecialProperty_TunnelWeatherBlend,
-
     // Add breakObject to serverside as well
     // 2024-05-31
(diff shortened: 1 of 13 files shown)


Shared/sdk/net/bitstream.h
```diff
@@ -548,6 +548,10 @@ enum class eBitStreamVersion : unsigned short
     // 2024-05-23
     WorldSpecialProperty_ExtendedWaterCannons,
 
+    // Add "extendedwatercannons" to setWorldSpecialPropertyEnabled
+    // 2024-06-30
+    WorldSpecialProperty_TunnelWeatherBlend,
+
```
botder: Looks like the only mistake here: Move this below `PedSync_Revision`

botder (changes requested):

Shared/sdk/net/bitstream.h
gta191977649: All right nice hit!

botder (approved):

PR #3537, opened 2024-07-01
Refactor server CLuaBlipDefs to use new parser

FileEX:


Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,145 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-        {
-            lua_pushboolean(luaVM, true);
-            return 1;
-        }
-    }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+bool CLuaBlipDefs::SetBlipOrdering(CElement* radarMarker, std::int32_t ordering) noexcept
+{
+    return CStaticFunctionDefinitions::SetBlipOrdering(radarMarker, Clamp(0i32, ordering, (std::int32_t)UINT16_MAX));
```
TracerDS: `UINT16_MAX` is a macro. You dont need to cast it to `(std::int32_t)`

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,145 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
+}
 
-    lua_pushboolean(luaVM, false);
-    return 1;
+bool CLuaBlipDefs::SetBlipVisibleDistance(CElement* radarMarker, std::uint32_t visibleDistance) noexcept
+{
+    return CStaticFunctionDefinitions::SetBlipVisibleDistance(radarMarker, Clamp(0ui32, visibleDistance, (std::uint32_t)UINT16_MAX));
```
TracerDS: same as above

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,145 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-        pVisibleTo = NULL;
-    else
-        argStream.ReadUserData(pVisibleTo, m_pRootElement);
-
-    if (!CBlipManager::IsValidIcon(ucIcon))
-    {
-        argStream.SetCustomError("Invalid icon");
-    }
+    if (visibleDistance.has_value())
+        visibleDistance = Clamp(0ui32, visibleDistance.value(), (std::uint32_t)UINT16_MAX);
```
TracerDS: same as before

TracerDS (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
tederis: There is `std::numeric_limits<std::int16_t>().max()`

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,144 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-                    {
-                        pGroup->Add(pBlip);
-                    }
-                    lua_pushelement(luaVM, pBlip);
-                    return 1;
-                }
-            }
-        }
+        m_pScriptDebugging->LogWarning(luaVM, SString("Blip size beyond 25 is no longer supported (got %i). It will be clamped between 0 and 25.", size.value()));
+        size = Clamp(0ui8, size.value(), 25ui8);
```
TracerDS: `Clamp(0, size.value(), 25)`

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,144 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
+    if (size.has_value() && (size.value() < 0 || size.value() > 25))
     {
-        unsigned char ucSize;
-        if (CStaticFunctionDefinitions::GetBlipSize(pBlip, ucSize))
-        {
-            lua_pushnumber(luaVM, ucSize);
-            return 1;
-        }
+        m_pScriptDebugging->LogWarning(luaVM, SString("Blip size beyond 25 is no longer supported (got %i). It will be clamped between 0 and 25.", size.value()));
+        size = Clamp(0ui8, size.value(), 25ui8);
```
TracerDS: `Clamp(0, size.value(), 25)`

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,144 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
     {
-        short sOrdering = Clamp(-32768, iOrdering, 32767);
-
-        if (CStaticFunctionDefinitions::SetBlipOrdering(pElement, sOrdering))
-        {
-            lua_pushboolean(luaVM, true);
-            return 1;
-        }
+        m_pScriptDebugging->LogWarning(luaVM, SString("Blip size beyond 25 is no longer supported (got %i). It will be clamped between 0 and 25.", size));
+        size = Clamp(0ui8, size, 25ui8);
```
TracerDS: `Clamp(0, size, 25)`

TracerDS (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
FileEX: Clamp expects arguments of the same type. If I just give 0 and 25 it receives it as an int and there are compilation errors

TracerDS: then cast 0 and 25 to uint8_t. `ui` suffix doesnt exist in C++ standard.

TracerDS: > then cast 0 and 25 to uint8_t. `ui` suffix doesnt exist in C++ standard.

Actually, it would be better to specify a type to Clamp directly:
```cpp
Clamp<std::uint8_t>(0, size.value(), 25)
```

TracerDS: you still have to change `ui8` to something like above to fix all build errors

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,144 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
+        size = Clamp<std::uint8_t>(0, size.value(), 25);
     }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
 
-    lua_pushboolean(luaVM, false);
-    return 1;
-}
+    if (ordering.has_value())
+        ordering = Clamp<std::int16_t>(std::numeric_limits<std::int16_t>().min(), ordering.value(), std::numeric_limits<std::int16_t>().max());
```
TracerDS: Why not just use `std::uint16_t`/`std::int16_t`?

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +70,144 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-        pVisibleTo = NULL;
-    else
-        argStream.ReadUserData(pVisibleTo, m_pRootElement);
-
-    if (!CBlipManager::IsValidIcon(ucIcon))
-    {
-        argStream.SetCustomError("Invalid icon");
-    }
+    if (visibleDistance.has_value())
+        visibleDistance = Clamp<std::uint16_t>(std::numeric_limits<std::uint16_t>().min(), visibleDistance.value(), std::numeric_limits<std::uint16_t>().max());
```
TracerDS: Above

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +72,120 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
     lua_registerclass(luaVM, "Blip", "Element");
 }
 
-int CLuaBlipDefs::CreateBlip(lua_State* luaVM)
+std::variant<CBlip*, bool> CLuaBlipDefs::CreateBlip(lua_State* const luaVM, const CVector vecPosition, const std::optional<std::uint8_t> icon, std::optional<std::uint8_t> size, const std::optional<std::uint8_t> r, const std::optional<std::uint8_t> g, const std::optional<std::uint8_t> b, const std::optional<std::uint8_t> a, const std::optional<std::int16_t> ordering, const std::optional<std::uint16_t> visibleDistance, const std::optional<CElement*> visibleTo)
```
Pirulax: This will error now when you pass bool or nil as the value to `visibleTo`.

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +72,120 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-                    if (pGroup)
-                    {
-                        pGroup->Add(pBlip);
-                    }
-                    lua_pushelement(luaVM, pBlip);
-                    return 1;
-                }
-            }
-        }
+        m_pScriptDebugging->LogError(luaVM, "Couldn't find the resource element");
```
Pirulax: shouldn't this be a `throw`?

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -70,374 +72,120 @@ void CLuaBlipDefs::AddClass(lua_State* luaVM)
...
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
-
-    lua_pushboolean(luaVM, false);
-    return 1;
+    SColor color = radarMarker->GetColor();
+    return {color.R, color.G, color.B, color.A};
 }
 
-int CLuaBlipDefs::GetBlipVisibleDistance(lua_State* luaVM)
+std::int16_t CLuaBlipDefs::GetBlipOrdering(CBlip* const radarMarker) noexcept
```
Pirulax: why not just do `auto` instead of explicitly casting?
this will just cause headaches in the future imo

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.h
```diff
@@ -11,28 +11,29 @@
...
-    LUA_DECLARE(SetBlipOrdering);
-    LUA_DECLARE(SetBlipVisibleDistance);
-};
\ No newline at end of file
+    // Set functions
+    static bool SetBlipIcon(CElement* const radarMarker, const std::uint8_t icon);
+    static bool SetBlipSize(lua_State* const luaVM, CElement* const radarMarker, std::uint8_t size) noexcept;
+    static bool SetBlipColor(CElement* const radarMarker, const std::uint8_t r, const std::uint8_t g, const std::uint8_t b, const std::uint8_t a) noexcept;
+    static bool SetBlipOrdering(CElement* const radarMarker, const std::int16_t ordering) noexcept;
+    static bool SetBlipVisibleDistance(CElement* const radarMarker, const std::uint16_t visibleDistance) noexcept;
```
Pirulax: Honestly, I think you could just cut all this out of the header, and move `LoadFunctions` to the bottom.
It's done like that in a few places, as there functions are static anyways.
This way you can also debug easier, because otherwise any changes in the header will likely trigger a full rebuild....

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
TracerDS: Only with LuaFunctionError/other exception

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.h
TracerDS: Changing headers wont rebuild the project. Only changing the source files will rebuild the solution (thats the default VS behavior that is)

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
TracerDS: tbf you shouldn't pass boolean to an element either way.
I think if you provide `nil` then it will leave it as an empty optional
Maybe optional variant?

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.h
FileEX: > Honestly, I think you could just cut all this out of the header, and move `LoadFunctions` to the bottom. It's done like that in a few places, as there functions are static anyways. This way you can also debug easier, because otherwise any changes in the header will likely trigger a full rebuild....

Can you tell me where, for example, this is done? I have not found this practice used

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
FileEX: I think this is expected behavior if a different type is specified than required, in this case boolean for the element. In the case of nil, the error does not occur because it means there is no argument value

tederis (dismissed):

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
```diff
@@ -13,431 +13,173 @@
...
-        }
-    }
-    else
-        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
-
-    lua_pushboolean(luaVM, false);
-    return 1;
+    SColor color = radarMarker->GetColor();
+    return std::make_tuple(color.R, color.G, color.B, color.A);
 }
```
ghost: Use `CLuaMultiReturn`

ghost (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaBlipDefs.cpp
tederis: Both `std::tuple` and `CLuaMultiReturn` are legitimate. The advantage of `CLuaMultiReturn` is that it may describe the intention better.

ghost: > Both `std::tuple` and `CLuaMultiReturn` are legitimate. The advantage of `CLuaMultiReturn` is that it may describe the intention better.

https://discord.com/channels/801330706252038164/801819598717976606/1327177125244960799

tederis: This is interesting. Internally `std::tuple` is expected to be converted into `CLuaMultiReturn.`

tederis: This will indeed be converted. But it requires the return type to be `CLuaMultiReturn`. For example, this will return two values:
```
CLuaMultiReturn<float, float> Fn()
{
	return std::make_tuple<float, float>(1.0f, 1.0f);
}
```

But this will return a table:
```
std::tuple<float, float> Fn()
{
   return std::make_tuple<float, float>(1.0f, 1.0f);
}
```

Which leads us to the fact that` auto CLuaBlipDefs::GetBlipColor(CBlip* radarMarker)` will return a table which breaks backward compatibility.

@Nico8340 Thanks for your attentiveness.
@FileEX Please, fix the beforementioned problems.

FileEX: Done, thanks for investigation.

botder (approved):

PR #3539, opened 2024-07-02
Add `resourceDynamicRoot` global variable and refactor `GetResourceDynamicElementRoot`

TracerDS:


Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -22,7 +23,7 @@ void CLuaResourceDefs::LoadFunctions()
         {"getResourceFromName", GetResourceFromName},
         {"getResourceRootElement", GetResourceRootElement},
         {"getResourceGUIElement", GetResourceGUIElement},
-        {"getResourceDynamicElementRoot", GetResourceDynamicElementRoot},
+        {"getResourceDynamicElementRoot", ArgumentParser<GetResourceDynamicElementRoot>},
```
FileEX, line 27: Shouldn't ArgumentParserWarn be here?

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -352,28 +353,17 @@ int CLuaResourceDefs::GetResourceGUIElement(lua_State* luaVM)
...
-int CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM)
+CClientEntity* CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM, std::optional<CResource*> resource)
 {
-    CResource*       pResource = NULL;
-    CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource);
+    CResource* pResource = resource.value_or(&lua_getownerresource(luaVM));
 
-    if (!argStream.HasErrors())
+    CClientEntity* resourceDynamicRoot = pResource->GetResourceDynamicEntity();
```
FileEX: Are you sure there is no possibility for pResource to be nullptr here?

FileEX (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -352,28 +353,17 @@ int CLuaResourceDefs::GetResourceGUIElement(lua_State* luaVM)
     return 1;
 }
 
-int CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM)
+CClientEntity* CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM, std::optional<CResource*> resource)
 {
-    CResource*       pResource = NULL;
-    CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource);
+    CResource* pResource = resource.value_or(&lua_getownerresource(luaVM));
```
FileEX: We are trying not to use Hungarian notation anymore, so please rename it to "resource"

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TracerDS: Not in this case

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TracerDS: I cannot rename it to `resource` as that name is defined as a parameter

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TracerDS: Non-ArgumentParser variant didnt have this check and everything worked fine, so...

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
FileEX: Can you explain why it doesn't matter here?

TracerDS: Because `getResourceDynamicElementRoot` throws an error under one condition, false if invalid parameters were passed (its handled by ArgumentParser)

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
tederis: You can add an `assert` to assure that `pResource` is always valid.

tederis: The server's counterpart of this function doesn't have this check. Maybe make them identical?

TracerDS: > You can add an `assert` to assure that `pResource` is always valid.

Shouldn't an exception be thrown instead of using `assert`?

tederis: I don't think so. Exceptions are supposed to be used when you're able to catch and process them. In this case an invalid ptr(in this place exactly) means that yours program in fire and there simply no sense in continuing of the execution.

TracerDS: Done

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -337,28 +337,16 @@ int CLuaResourceDefs::GetResourceGUIElement(lua_State* luaVM)
...
-int CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM)
+CClientEntity* CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM, std::optional<CResource*> resource)
 {
-    CResource*       pResource = NULL;
-    CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource);
-
-    if (!argStream.HasErrors())
+    CResource* pResource = resource.value_or(&lua_getownerresource(luaVM));
+    if (!pResource->IsActive())
```
ghost: This check caused problems with a previous pull request, I don't think we should use it here either

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -954,49 +954,16 @@ int CLuaResourceDefs::getResourceRootElement(lua_State* luaVM)
...
+CElement* CLuaResourceDefs::GetResourceDynamicElementRoot(lua_State* luaVM, std::optional<CResource*> resource)
 {
-    CResource* pResource;
-
-    CScriptArgReader argStream(luaVM);
-    argStream.ReadUserData(pResource, NULL);
-
-    if (!argStream.HasErrors())
+    CResource* pResource = resource.value_or(&lua_getownerresource(luaVM));
+    if (!pResource->IsActive())
```
ghost: The same as above

ghost (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TracerDS: Thats the original version ported to ArgumentParser

ghost: The client-sided equivalent does not include this check, so you should only remove it from the client-sided

TracerDS: I added this check as per previous review

ghost: I think you should do some tests then, just use the map editor for a few minutes, because there was a problem with the previous pull request, and if it doesn't cause any problems, close the conversation and leave it there

TracerDS: Works fine for me. I didnt encounter any issues

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
FileEX: You can rename it to something other than "resource", e.g. currentResource.

TracerDS: Yes, I can rename it but not to resource I meant 😅
Will change in a sec

ghost (approved):
LGTM

Fernando-A-Rocha (approved):
LGTM!

TracerDS:

PR #3540, opened 2024-07-03
Fixed backward compatibility of checkpoints and arrows marker

FileEX:
Added a new optional argument to createMarker
``ignoreAlphaLimits = false``

This argument is only valid for checkpoint and arrow markers. It defines whether the alpha of these markers can be changed. If the argument is false, the marker has a constant alpha as before, i.e. checkpoint always 128 and arrow always 255. If the argument is true, alpha can be changed.

Default is **false**

I tested this using ``createMarker``, ``setElementAlpha``, ``setMarkerColor`` and ``setMarkerType`` on the server and client side. All my tests were successful


Client/mods/deathmatch/logic/CClient3DMarker.h
```diff
@@ -45,14 +45,17 @@ class CClient3DMarker : public CClientMarkerCommon
     void SetVisible(bool bVisible) { m_bVisible = bVisible; };
 
     SColor GetColor() const { return m_Color; }
-    void   SetColor(const SColor& color) { m_Color = color; }
+    void   SetColor(const SColor& color);
```
TracerDS: `noexcept`

Client/mods/deathmatch/logic/CClientCheckpoint.h
```diff
@@ -74,6 +74,9 @@ class CClientCheckpoint : public CClientMarkerCommon
     static bool          IconToString(unsigned char ucIcon, SString& strOutString);
     void                 ReCreateWithSameIdentifier();
 
+    void IgnoreAlphaLimits(bool ignore) noexcept { m_ignoreAlphaLimits = ignore; };
```
TracerDS: What do you think about `SetIgnoreAlphaLimits`?

Client/mods/deathmatch/logic/CClientMarkerCommon.h
```diff
@@ -56,4 +56,7 @@ class CClientMarkerCommon
 
     virtual void StreamIn() = 0;
     virtual void StreamOut() = 0;
+
+    virtual void IgnoreAlphaLimits(bool ignore) = 0;
+    virtual bool AreAlphaLimitsIgnored() const = 0;
```
TracerDS: `noexcept`?

TracerDS (changes requested):

Dutchman101:
Please resolve conflicts
@FileEX

FileEX:
> Please resolve conflicts @FileEX

Done

PR #3550, opened 2024-07-05
Checkpoints arrow customization

FileEX:
As I said in #3523  ,I added the ability to customize the target arrow in checkpoint markers. You can change the color of the arrow and its size.

**Syntax**
``bool setMarkerTargetArrowProperties(element marker [, int r = 255, int g = 64, int b = 64, int a = 255, float size = markerSize * 0.625 ] )`` - If we do not pass arguments, the default properties will be restored (something like restart).

``r, g, b, a, size = getMarkerTargetArrowProperties(element marker)``

Screenshot_1


Proxy-99:
is it possible to support OOP too
something like 
`marker:setTargetColor(255,0,0,255)`
`marker:setTargetSize(2)`
get
`marker.targetColor`
`marker.targetSize`

Client/game_sa/CCheckpointSA.h
```diff
@@ -14,6 +14,12 @@
 #include <CVector.h>
 #include <game/CCheckpoint.h>
 
+#define HOOKPOS_CCheckpoint__Render  0x725E56
+#define HOOKSIZE_CCheckpoint__Render 0x5
+static const constexpr std::uint32_t RETURN_CCheckpoint__Render = 0x725E5B;
```
TracerDS: `const` is unnecessary here

Client/mods/deathmatch/logic/CClientCheckpoint.cpp
```diff
@@ -385,3 +390,20 @@ void CClientCheckpoint::ReCreateWithSameIdentifier()
...
+
+void CClientCheckpoint::SetTargetArrowProperties(const SColor& arrowColor, float size) noexcept
+{
+    if (m_TargetArrowColor != arrowColor || m_TargetArrowSize != size)
+    {
+        m_TargetArrowColor = arrowColor;
+        m_TargetArrowSize = size;
+
+        ApplyCheckpointTargetArrowProperties();
+    }
```
TracerDS: early return?

Client/mods/deathmatch/logic/rpc/CMarkerRPCs.cpp
```diff
@@ -137,4 +138,22 @@ void CMarkerRPCs::SetMarkerIcon(CClientEntity* pSource, NetBitStreamInterface& b
...
+        CClientMarker* marker = m_pMarkerManager->Get(pSource->GetID());
+        if (!marker)
+            return;
+
+        CClientCheckpoint* checkpoint = marker->GetCheckpoint();
+        if (!checkpoint)
+            return;
+
+        checkpoint->SetTargetArrowProperties(color, size);
+    }
```
TracerDS, line 158: early return?

Server/mods/deathmatch/logic/CMarker.cpp
```diff
@@ -301,6 +305,23 @@ void CMarker::SetIcon(unsigned char ucIcon)
...
+        m_TargetArrowSize = size;
+
+        CBitStream BitStream;
+        BitStream.pBitStream->Write(color.R);
+        BitStream.pBitStream->Write(color.G);
+        BitStream.pBitStream->Write(color.B);
+        BitStream.pBitStream->Write(color.A);
+        BitStream.pBitStream->Write(size);
+        BroadcastOnlyVisible(CElementRPCPacket(this, SET_MARKER_TARGET_ARROW_PROPERTIES, *BitStream.pBitStream));
+    }
```
TracerDS: early return?

TracerDS (changes requested):

TracerDS:
> `r, g, b, a, size = getMarkerTargetArrowProperties(element marker)`

What do you think about `getMarkerTargetArrowSize` and `getMarkerTargetArrowColor`?

FileEX:
I think there is really no need to create two separate functions for this, because it is very simple and there are only 2 arguments - color and size.

Proxy-99:
> I think there is really no need to create two separate functions for this, because it is very simple and there are only 2 arguments - color and size.

you can use it for the oop syntax tho I think better to separate them for better readability

FileEX:
OOP is not mandatory.  If someone wants to make OOP or setMarkerTargetArrow / setMarkerTargetArrowSize versions, feel free to do so

PR #3553, opened 2024-07-06
Fix outputDebugString level 4 not being logged

### LINKED ISSUE #3552: Level 4 outputDebugString not written to clientscript.log
### Describe the bug

When using outputDebugString with level 4 it won't write the logs to clientscript.log. All others work as expected. This is likely due to a recent change in which level 4 is no longer tied to level 0, likely the underlying code for clientscript.log can't handle it.

### Steps to reproduce

Run with runcode: `crun outputDebugString('test', 4, 255, 255, 0)`

### Version

Client: Multi Theft Auto v1.6-release-22600
Server: MTA:SA Server v1.6-release-22600

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

MegadreamsBE:
fixes https://github.com/multitheftauto/mtasa-blue/issues/3552

Essentially a client-side port of https://github.com/multitheftauto/mtasa-blue/pull/3499 by moving the enum and CheckForSufficientDebugLevel into a shared class. Using this check makes level 4 equal to level 0 for the purposes of the level sufficiency check which allows its contents to be written to clientscript.log as intended.

At this moment the level check doesn't serve much of a purpose (as it's always set to 3 client-side, through setLogFile) but it does pave the way to someday only log down items based on the debugscript client-side if desired. Either way it's a cleaner solution than adding a secondary check for level 4.

### DIFF
--- a/Client/mods/deathmatch/logic/CScriptDebugging.cpp
+++ b/Client/mods/deathmatch/logic/CScriptDebugging.cpp
@@ -10,23 +10,6 @@
 
 #include <StdInc.h>
 
-enum DebugScriptLevels : std::uint8_t
-{
-    NONE,
-    ERRORS_ONLY,
-    ERRORS_AND_WARNINGS,
-    ALL,
-};
-
-enum DebugMessageLevels : std::uint8_t
-{
-    MESSAGE_TYPE_DEBUG,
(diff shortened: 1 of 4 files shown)


ghost (approved):
Looks good 🎉

PR #3557, opened 2024-07-06
Fix valid anim

G-Moris:
Is everything okay in this PR? Offer your corrections

For Dutchman101: It was not my initiative to harm forks with incorrect animation, and I did not do it. (If you thought I was doing this)

### DIFF
--- a/Client/game_sa/CAnimManagerSA.cpp
+++ b/Client/game_sa/CAnimManagerSA.cpp
@@ -21,6 +21,9 @@
 
 extern CGameSA* pGame;
 
+const char* const CAnimManagerSA::m_kGateWayBlockName = "ped";
+const char* const CAnimManagerSA::m_kGateWayAnimationName = "run_wuzi";
+
 CAnimManagerSA::CAnimManagerSA()
 {
     MemSetFast(m_pAnimAssocGroups, 0, sizeof(m_pAnimAssocGroups));
--- a/Client/game_sa/CAnimManagerSA.cpp
+++ b/Client/game_sa/CAnimManagerSA.cpp
@@ -1,11 +1,11 @@
 /*****************************************************************************
  *
- *  PROJECT:     Multi Theft Auto v1.0
+ *  PROJECT:     Multi Theft Auto
  *  LICENSE:     See LICENSE in the top level directory
- *  FILE:        game_sa/CAnimManagerSA.cpp
+ *  FILE:        Client/game_sa/CAnimManagerSA.cpp
  *  PURPOSE:     Animation manager
  *
- *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *  Multi Theft Auto is available from https://multitheftauto.com/
  *
  *****************************************************************************/
 
@@ -20,8 +20,6 @@
 #include "CKeyGenSA.h"
 
 extern CGameSA* pGame;
-
-using std::list;
 
 CAnimManagerSA::CAnimManagerSA()
 {
@@ -276,6 +274,10 @@ std::unique_ptr<CAnimBlendAssociation> CAnimManagerSA::CreateAnimAssociation(Ass
 
 CAnimManagerSA::StaticAssocIntface_type CAnimManagerSA::GetAnimStaticAssociation(eAnimGroup animGroup, eAnimID animID)
 {
+    // We check the validity of the group, avoid crashes due to an invalid group
+    if (!IsValidGroup(static_cast<uint32_t>(animGroup)))
+        return nullptr;
+
     CAnimBlendStaticAssociationSAInterface* pInterface = nullptr;
     DWORD                                   dwFunc = FUNC_CAnimManager_GetAnimAssociation;
     _asm
@@ -295,6 +297,10 @@ CAnimManagerSA::StaticAssocIntface_type CAnimManagerSA::GetAnimStaticAssociation
 
 std::unique_ptr<CAnimBlendAssociation> CAnimManagerSA::GetAnimAssociation(AssocGroupId animGroup, const char* szAnimName)
 {
+    // We check the validity of the group, avoid crashes due to an invalid group
+    if (!IsValidGroup(animGroup))
+        return nullptr;
+
     CAnimBlendAssociationSAInterface* pInterface = nullptr;
     DWORD                             dwFunc = FUNC_CAnimManager_GetAnimAssociation_str;
     _asm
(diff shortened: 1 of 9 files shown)


Client/game_sa/CAnimManagerSA.cpp
```diff
@@ -276,6 +274,10 @@ std::unique_ptr<CAnimBlendAssociation> CAnimManagerSA::CreateAnimAssociation(Ass
 
 CAnimManagerSA::StaticAssocIntface_type CAnimManagerSA::GetAnimStaticAssociation(eAnimGroup animGroup, eAnimID animID)
 {
+    // We check the validity of the group, avoid crashes due to an invalid group
+    if (!IsValidGroup(static_cast<uint32_t>(animGroup)))
```
FileEX: According to the new coding guidelines, please change the uint32_t types to std::uint32_t

FileEX:
I think you should open a separate PR for wasted events

G-Moris:
> I think you should open a separate PR for wasted events

I don't think so, since this is an update to the old event

FileEX:
> > I think you should open a separate PR for wasted events
> 
> I don't think so, since this is an update to the old event

if we need to revert this commit for any reason, we will lose all the changes included in it. For example, instead of just reverting the 'fix valid anim' change, we will also lose the update to 'wasted events'. That's why it's better to separate different changes into individual PRs, making it easier to manage and revert them if necessary.

G-Moris:
> > > I think you should open a separate PR for wasted events
> > 
> > 
> > I don't think so, since this is an update to the old event
> 
> if we need to revert this commit for any reason, we will lose all the changes included in it. For example, instead of just reverting the 'fix valid anim' change, we will also lose the update to 'wasted events'. That's why it's better to separate different changes into individual PRs, making it easier to manage and revert them if necessary.

OK, I'll delete the event change.

TheNormalnij (dismissed):

TheNormalnij:
lgtm

FileEX (approved):
This is just a detail, but according to the guidelines you shouldn't remove "www" from the url

Client/game_sa/CAnimManagerSA.h
```diff
@@ -173,6 +175,6 @@ class CAnimManagerSA : public CAnimManager
     // ped block, but in reality, it's playing our custom animation, and Of course, we can return run_wuzi
     // animation within the hook if we want to play it instead. Why run_wuzi? We can also use another animation,
     // but I've tested with this one mostly, so let's stick to this.
-    const SString m_kGateWayBlockName = "ped";
-    const SString m_kGateWayAnimationName = "run_wuzi";
+    const char* m_kGateWayBlockName = "ped";
+    const char* m_kGateWayAnimationName = "run_wuzi";
```
tederis: These two class members should be static because they stay constant during the lifetime of class object.

Client/sdk/game/CAnimManager.h
```diff
@@ -128,7 +128,10 @@ class CAnimManager
...
 
-    virtual bool           isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
-    virtual const SString& GetGateWayBlockName() = 0;
-    virtual const SString& GetGateWayAnimationName() = 0;
+    virtual bool        isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
+    virtual const char* GetGateWayBlockName() = 0;
+    virtual const char* GetGateWayAnimationName() = 0;
+
+    virtual bool IsValidGroup(std::uint32_t uiAnimGroup) = 0;
+    virtual bool IsValidAnim(std::uint32_t uiAnimGroup, std::uint32_t uiAnimID) = 0;
```
tederis: These methods have to be const qualified.

Client/sdk/game/CAnimManager.h
```diff
@@ -128,9 +128,9 @@ class CAnimManager
     virtual void                                 DeleteCustomAnimHierarchyInterface(CAnimBlendHierarchySAInterface* pInterface) = 0;
     virtual void                                 DeleteCustomAnimSequenceInterface(CAnimBlendSequenceSAInterface* pInterface) = 0;
 
-    virtual bool        isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
-    virtual const char* GetGateWayBlockName() = 0;
-    virtual const char* GetGateWayAnimationName() = 0;
+    virtual bool              isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
+    virtual const char* const GetGateWayBlockName() = 0;
+    virtual const char* const GetGateWayAnimationName() = 0;
```
tederis:
```
virtual const char* GetGateWayBlockName() const = 0;
virtual const char* GetGateWayAnimationName() const = 0;
```

Client/game_sa/CAnimManagerSA.h
```diff
@@ -175,6 +175,6 @@ class CAnimManagerSA : public CAnimManager
     // ped block, but in reality, it's playing our custom animation, and Of course, we can return run_wuzi
     // animation within the hook if we want to play it instead. Why run_wuzi? We can also use another animation,
     // but I've tested with this one mostly, so let's stick to this.
-    const char* m_kGateWayBlockName = "ped";
-    const char* m_kGateWayAnimationName = "run_wuzi";
+    static const char* const m_kGateWayBlockName;
+    static const char* const m_kGateWayAnimationName;
```
tederis: These variables can be moved into cpp sources. There's no reason for keeping them in a header file.

Client/sdk/game/CAnimManager.h
```diff
@@ -128,7 +128,10 @@ class CAnimManager
...
 
-    virtual bool           isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
-    virtual const SString& GetGateWayBlockName() = 0;
-    virtual const SString& GetGateWayAnimationName() = 0;
+    virtual bool              isGateWayAnimationHierarchy(CAnimBlendHierarchySAInterface* pInterface) = 0;
+    virtual const char* GetGateWayBlockName() const = 0;
+    virtual const char* GetGateWayAnimationName() const = 0;
+
+    virtual bool IsValidGroup(std::uint32_t uiAnimGroup) = 0;
+    virtual bool IsValidAnim(std::uint32_t uiAnimGroup, std::uint32_t uiAnimID) = 0;
```
tederis: These to also have to be const

G-Moris: Then you will not be able to use GetAnimBlendAssoc and GetAnimStaticAssociation

TheNormalnij: these use hooked functions

tederis: > these use hooked functions

What does this change? GetAnimStaticAssociation and GetAnimBlendAssoc(which are used inside) are also const.

tederis (approved):
Lgtm

botder (approved):

PR #3564, opened 2024-07-13
Add comments for hooks in CMultiplayerSA_CrashFixHacks (Documentation)

FileEX:


Dutchman101:
Nice (also that you researched the fixes).

Reminder to self or someone else in the future: there's a bunch of failed crash fix-hacks, that it crashes on an instruction within the crash 'fix' - or the actual hook placement - itself, or a part of related crashes do, where the 'fix' likely reduced incidence of the original crash after all but just doesn't cover all scenarios.

In many cases, a player that crashed after all has the crash fix hack ID (that it triggered and _tried_ to avert crash) logged to the dump's additional header and/or report.log, and you can imagine the whole situation may overlap with some players running out of memory (RAM abuse by server, bad PC) where the original nullptr was brought on by failed allocations somewhere else. At which point it means, some cases just can't be salvaged, it's bound to crash somewhere else anyways, but the specifics of each case will have to be investigated down the line.

What has to be done in the future is collect crash stats and link those that crash on "CrashFixHacks" instructions into a centralized issue & have ppl investigate why it may fail to solidify the crash fix.

Issue #2317, opened 2021-08-23
Sky flickers when setTime is used


LVRENT:
**Describe the bug**
The sky flickers when setTime function is used (tested serverside, pretty sure the same thing also happens on clientside).

**To reproduce**
Create a timer that will update the time or call the function somewhere and look at the sky.
Also, I have noticed it doesn't occur every single time when the function is used. If I create a timer that is called let's say at every 1 second you may don't notice it. If you set the timer at =>10 seconds it gets noticed almost every single time. I can see that in my server where I have a timer called at every 30 seconds to update the global server time and it's quite annoying.

**Expected behaviour**
The sky doesn't need to flicker. My idea is when the function is called, the function first sets the time around 12:00 and then the time we selected, that making the sky appear to flicker.

**Screenshots**
I can't really take one as I can't catch the moment. Maybe a video later.

**Version**
1.5.8-r20930

Einheit-101:
You could record a video and then play frame by frame to get the right screenshots

francisco-create:
Mate, can you create a clip of the sky and send it to me, 5 10 seconds will be enough.

PR #3567, opened 2024-07-14
Freeze Time functions

Proxy-99:
Fixing #2317
no longer need to use setTimer  or clientRender for setTime() function 
fix sky flickers too much

added new functions
```
setTimeFrozen(true)
local get = isTimeFrozen()
resetTimeFrozen()
print(get)

```
thanks to @FileEX for memory addresses locations and help


Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,30 @@ bool CWeatherSA::ResetRainbow()
...
+}
+
+bool CWeatherSA::GetTimerCycleEnabled()
+{
+    return m_bTimeCycleFrozen;
+}
+
+bool CWeatherSA::ResetTimerCycle()
+{
+    MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
```
FileEX: To avoid creating redundancy here, you can simply call SetTimerCycle with 0

Client/game_sa/CWeatherSA.h
```diff
@@ -55,10 +55,15 @@ class CWeatherSA : public CWeather
     bool  SetRainbow(float fAmount);
     bool  ResetRainbow();
 
+    bool SetTimerCycle(unsigned char value);
```
FileEX, line 58: noexcept?

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,30 @@ bool CWeatherSA::ResetRainbow()
...
+        MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
+        m_bTimeCycleFrozen = false;
+    }
+    return true;
+}
+
+bool CWeatherSA::GetTimerCycleEnabled()
+{
+    return m_bTimeCycleFrozen;
+}
```
FileEX: You can easily move this to the header as const and noexcept

FileEX (changes requested):

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,25 @@ bool CWeatherSA::ResetRainbow()
     MemCpy((LPVOID)(0x72BF59 + 2), &originalCodes, 3);
     return true;
 }
+
+bool CWeatherSA::SetTimerCycle(unsigned char value) noexcept
```
TracerDS: `bool` + `noexcept`?

Client/game_sa/CWeatherSA.h
```diff
@@ -55,10 +55,15 @@ class CWeatherSA : public CWeather
     bool  SetRainbow(float fAmount);
     bool  ResetRainbow();
 
+    bool SetTimerCycle(unsigned char value) noexcept;
```
TracerDS: if it always return true, use `void` instead of `bool`

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,25 @@ bool CWeatherSA::ResetRainbow()
...
+    }
+    else
+    {
+        MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
+        m_bTimeCycleFrozen = false;
+    }
+    return true;
+}
+
+bool CWeatherSA::ResetTimerCycle()
```
TracerDS: `noexcept`?

Client/sdk/game/CWeather.h
```diff
@@ -51,4 +51,8 @@ class CWeather
     virtual float GetRainbow() const = 0;
     virtual bool  SetRainbow(float fAmount) = 0;
     virtual bool  ResetRainbow() = 0;
+
+    virtual bool SetTimerCycle(unsigned char value) = 0;
+    virtual bool GetTimerCycleEnabled() const = 0;
+    virtual bool ResetTimerCycle() = 0;
```
TracerDS: `noexcept`?

TracerDS (changes requested):

Client/game_sa/CWeatherSA.h
Proxy-99: but it's needed later for SetWorldProperty
return g_pGame->GetWeather()->SetTimerCycle(arg1);

Client/game_sa/CWeatherSA.cpp
Proxy-99: what do you mean

TracerDS: > what do you mean

```cpp
bool CWeatherSA::SetTimerCycle(bool value) noexcept
```

Client/game_sa/CWeatherSA.h
TracerDS: > but it's needed later for SetWorldProperty return g_pGame->GetWeather()->SetTimerCycle(arg1);

ah. Alright

Client/game_sa/CWeatherSA.cpp
Proxy-99: I used unsigned char to make it compatible with SetWorldProperty which accepts float, to avoid refactoring

TracerDS: It accepts float, but you're passing it to a byte (unsinged char)?

Proxy-99: I will  refactoring some of SetWorldProperty with std::variant to hold float and boolean which is better

Client/game_sa/CWeatherSA.cpp
```diff
@@ -229,9 +229,9 @@ bool CWeatherSA::ResetRainbow()
     return true;
 }
 
-bool CWeatherSA::SetTimerCycle(unsigned char value) noexcept
+bool CWeatherSA::SetTimerCycle(bool value) noexcept
 {
-    if (value == 1)
+    if (value)
```
TracerDS: if you can, could you change it to
```cpp
if(value)
{
    MemSet((void*)0x53BFBD, 0x90, 5);
}
else
{
    MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
}
m_bTimeCycleFrozen = value;
return true;
```
?

TracerDS (approved):

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,24 @@ bool CWeatherSA::ResetRainbow()
...
+    {
+        MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
+    }
+    m_bTimeCycleFrozen = value;
+    return true;
+}
+
+bool CWeatherSA::ResetTimerCycle() noexcept
+{
+    SetTimerCycle(0);
```
FileEX: false instead of 0

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,24 @@ bool CWeatherSA::ResetRainbow()
...
+    m_bTimeCycleFrozen = false;
```
FileEX: m_bTimeCycleFrozen = false;
can be removed because you are calling SetTimerCycle

Client/game_sa/CWeatherSA.cpp
```diff
@@ -228,3 +228,24 @@ bool CWeatherSA::ResetRainbow()
...
+bool CWeatherSA::SetTimerCycle(bool value) noexcept
+{
+    if (value)
+    {
+        MemSet((void*)0x53BFBD, 0x90, 5);
+    }
+    else
+    {
+        MemCpy((void*)0x53BFBD, "\xE8\x4E\x0F\xFF\xFF", 5);
+    }
```
FileEX: You can remove the braces

FileEX (changes requested):

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3418,6 +3418,7 @@ void CClientGame::Event_OnIngame()
 
     g_pGame->GetBuildingRemoval()->ClearRemovedBuildingLists();
     g_pGame->GetWorld()->SetOcclusionsEnabled(true);
+    g_pGame->GetWeather()->ResetTimerCycle();
```
FileEX: This line should be on line 5547, below the rest of world properties

FileEX:
I'm not sure this should be in ``CWeatherSA`` as it has little to do with weather. The code toggles the call to ``CClock::Update``, so in my opinion this should be in ``CClockSA`` as that is the class responsible for the clock & time.

Proxy-99:
> I'm not sure this should be in `CWeatherSA` as it has little to do with weather. The code toggles the call to `CClock::Update`, so in my opinion this should be in `CClockSA` as that is the class responsible for the clock & time.

done (https://github.com/multitheftauto/mtasa-blue/pull/3567/commits/d50238899f11faaa9d1c0dcc0eace17cc4a6bb4b)

FileEX:
It seems to me that "timer cycle" does not fit here. 
I propose

- SetClockFrozen
- IsClockFrozen
- ResetClockFrozenState (or something?)

Proxy-99:
> It seems to me that "timer cycle" does not fit here. I propose
> 
> * SetClockFrozen
> * IsClockFrozen
> * ResetClockFrozenState (or something?)

I added setTimeFrozen(true)
isTimeFrozen()
resetTimeFrozen()
(https://github.com/multitheftauto/mtasa-blue/pull/3567/commits/ff757de10f37b384c3ece6a1b89f28d33ff211ad)

TracerDS (approved):

Client/sdk/game/CClock.h
```diff
@@ -16,4 +16,8 @@ class CClock
 public:
     virtual void Set(BYTE bHour, BYTE bMinute) = 0;
     virtual void Get(BYTE* bHour, BYTE* bMinute) = 0;
+
+    virtual bool SetTimerCycle(bool value) noexcept = 0;
```
FileEX: Please rename functions also in the CClock class

Proxy-99: fixed in (https://github.com/multitheftauto/mtasa-blue/pull/3567/commits/8213d4868071c5a2e0578c854cd28008048dc0ab)

FileEX (approved):
Now, lgtm

TracerDS (approved):

PR #3572, opened 2024-07-15
Add spawnFlyingComponent & breakGlass arguments for setVehiclePanelState

### LINKED ISSUE #2122: Add spawnFlyingComponent for setVehiclePanelState

FileEX:
This PR adds the ``spawnFlyingComponent`` argument (defaulting to true) to ``setVehiclePanelState``, similar to what is available in ``setVehicleDoorState``. Additionally, the PR introduces the feature to shatter the windscreen if windscreen_panel is set to 3; currently, the windscreen simply disappears.

Added ``breakGlass`` argument.

```
bool setVehiclePanelState(vehicle theVehicle, int panelID, int state [ , bool spawnFlyingComponent = true, bool breakGlass = false ] )
```

https://streamable.com/uxpph6

This PR helps avoid frustrating situations, such as when managing a garage of vehicles and changing their models.
image

Closes #2122


Fernando-A-Rocha:
Well done!

FileEX:
I added the ``breakGlass`` argument because the glass component does fall out (but it is poorly visible and falls underground). To maintain backward compatibility, the glass will fall out as before. If ``breakGlass`` is set to true, the glass will not fall out but will break as in the video.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -3202,17 +3202,17 @@ bool CStaticFunctionDefinitions::SetVehicleLightState(CClientEntity& Entity, uns
...
+    RUN_CHILDREN(SetVehiclePanelState(**iter, ucPanel, ucState, spawnFlyingComponent, breakGlass))
 
     if (IS_VEHICLE(&Entity))
     {
         CClientVehicle& Vehicle = static_cast<CClientVehicle&>(Entity);
 
         if (ucPanel < 7)
         {
-            Vehicle.SetPanelStatus(ucPanel, ucState);
+            Vehicle.SetPanelStatus(ucPanel, ucState, spawnFlyingComponent, breakGlass);
```
Proxy-99, line 3210: can u add validation for ucState as a refactor
```c++
if(ucState>3)
    return false;
``` 
also for server side  and early return

Client/game_sa/CDamageManagerSA.cpp
```diff
@@ -142,11 +142,13 @@ void CDamageManagerSA::SetPanelStatus(BYTE bPanel, BYTE bPanelStatus)
                 // Call CAutomobile::SetPanelDamage to update the vehicle
                 dwFunction = 0x6B1480;
                 dwThis = (DWORD)internalEntityInterface;
-                bool bUnknown = false;
+                bool windscreenShatter = bPanel == ePanels::WINDSCREEN_PANEL && breakGlass;
+                bool quiet = !spawnFlyingComponent;
```
Proxy-99: what is the purpose of this     ` bool quiet = !spawnFlyingComponent;`

Proxy-99 (changes requested):

Client/game_sa/CDamageManagerSA.cpp
FileEX: If the quiet argument is true then the component simply disappears without falling off

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
FileEX: This PR is not aimed at refactoring, so in this case early-return is already beyond the scope of this PR

Proxy-99: I think u can add the validation to fix that issue too I do think anyone will fix this later on

Proxy-99 (approved):
Looks good to me

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
FileEX: This isn't really necessary because the ``CDamageManagerSA::SetPanelStatus`` function checks whether the status is correct

Proxy-99: you are right so the issue is just return false rather than true in that code  😅
I will refactor this part I think later after your pr 

```c++
bool CStaticFunctionDefinitions::SetVehiclePanelState(CClientEntity& Entity, unsigned char ucPanel, unsigned char ucState, bool spawnFlyingComponent, bool breakGlass)
{
    RUN_CHILDREN(SetVehiclePanelState(**iter, ucPanel, ucState, spawnFlyingComponent, breakGlass))

    if (IS_VEHICLE(&Entity))
    {
        CClientVehicle& Vehicle = static_cast<CClientVehicle&>(Entity);

        if (ucPanel < 7)
        {
            Vehicle.SetPanelStatus(ucPanel, ucState, spawnFlyingComponent, breakGlass);
            return true;
        }
    }

    return false;
}
```

Client/game_sa/CDamageManagerSA.cpp
```diff
@@ -155,13 +157,13 @@ void CDamageManagerSA::SetPanelStatus(BYTE bPanel, BYTE bPanelStatus)
     }
```
TheNormalnij, line 146: You can avoid assembly here

Client/game_sa/CDamageManagerSA.cpp
```diff
@@ -142,11 +142,13 @@ void CDamageManagerSA::SetPanelStatus(BYTE bPanel, BYTE bPanelStatus)
...
                 dwThis = (DWORD)internalEntityInterface;
-                bool bUnknown = false;
+                bool windscreenShatter = bPanel == ePanels::WINDSCREEN_PANEL && breakGlass;
+                bool quiet = !spawnFlyingComponent;
                 _asm
                 {
                     mov     ecx, dwThis
-                    push    bUnknown
+                    push    quiet
+                    push    windscreenShatter
```
TheNormalnij: Do you have stack issues after adding a new argument?

FileEX: Nope

Fernando-A-Rocha:
#3592 already adds spawnVehicleFlyingComponent so i think this PR should only add breakGlass ?

FileEX:
> #3592 already adds spawnVehicleFlyingComponent so i think this PR should only add breakGlass ?

The spawnVehicleFlyingComponent function allows you to create components, while the argument in the setVehiclePanelState function allows you to avoid creating them. By default, after calling the function, for example, the bumper is flying, which I showed in the screenshots and it can be annoying. It works exactly the same way as in the case of setVehicleDoorState. This argument is not intended to spawn components, but to not spawn them if someone does not want it, because they are spawned by default

Fernando-A-Rocha:
Sorry for the misunderstanding. You are right, this PR is ok

Fernando-A-Rocha (approved):

botder (dismissed):

FileEX:
Conflicts resolved

PR #3574, opened 2024-07-16
Appendum - CLuaFunctionParser refactor (#3494)

TracerDS:
Appendum for #3494 
This PR will allow you to get variadic number of arguments via ArgumentParser.

Example:
```cpp
std::string SomeFunctionDefs::FunctionName(std::string arg1, LuaVarArgs args)
{
    DoSomethingWith(args.value_or(CLuaArguments())); // default args
}
```
`LuaVarArgs` is an alias to `std::optional<CLuaArguments>`.

### DIFF
--- a/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
+++ b/Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
@@ -18,7 +18,6 @@ class CLuaArgument;
 #include "lua/CLuaStackChecker.h"
 #include "lua/LuaBasic.h"
 #include <lua/CLuaMultiReturn.h>
-#include <tuple>
 
 class LuaFunctionError
 {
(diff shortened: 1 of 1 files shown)


Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -18,6 +18,7 @@ class CLuaArgument;
 #include "lua/CLuaStackChecker.h"
 #include "lua/LuaBasic.h"
 #include <lua/CLuaMultiReturn.h>
+#include <tuple>
```
FileEX: Unused?

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -33,6 +34,8 @@ class LuaFunctionError
     constexpr bool        IsWarning() const noexcept { return m_bWarning; }
 };
 
+using LuaVarArgs = std::optional<CLuaArguments>;
```
FileEX, line 36: Unused?

TracerDS: It is used as an alias. Instead of doing `std::optional<CLuaArguments>` you could do `LuaVarArgs`

PR #3575, opened 2024-07-16
Add new events: onElementCreate/onClientElementCreate

TracerDS:
These 2 event will trigger when you want to create **any** element.
It will trigger for vehicles, peds, weapons, pickups, markers, blips, waters, radar-areas, etc.

Syntax:
```cpp
onElementCreate ( element theElement, string elementType );
onClientElementCreate ( element theElement, string elementType );
```

### DIFF
--- a/Client/game_sa/CPedSA.cpp
+++ b/Client/game_sa/CPedSA.cpp
@@ -134,7 +134,14 @@ bool CPedSA::IsInWater()
 
 bool CPedSA::AddProjectile(eWeaponType eWeapon, CVector vecOrigin, float fForce, CVector* target, CEntity* targetEntity)
 {
-    return ((CProjectileInfoSA*)pGame->GetProjectileInfo())->AddProjectile((CEntitySA*)this, eWeapon, vecOrigin, fForce, target, targetEntity);
+    auto projectileInfo = (CProjectileInfoSA*)pGame->GetProjectileInfo();
+    return projectileInfo->AddProjectile((CEntitySA*)this, eWeapon, vecOrigin, fForce, target, targetEntity);
+}
+
+void CPedSA::DeleteProjectile(CProjectile* projectile) noexcept
+{
+    auto* info = pGame->GetProjectileInfo();
+    info->RemoveProjectile(info, projectile, false);
 }
 
 void CPedSA::DetachPedFromEntity()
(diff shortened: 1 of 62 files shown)


TheNormalnij:
How much does this event impact performance? Race maps can have a lot of objects and they will generate a lot of lua calls

TracerDS:
> How much does this event impact performance? Race maps can have a lot of objects and they will generate a lot of lua calls

I didn't have the opportunity to check yet but I guess it might lag a little if there is **huge** amount of objects

FileEX:
Why does this PR contain so much more than it should?  According to the title and description of the PR, it adds the onElementCreate and onClientElementCreate events, and meanwhile this PR contains a ton of refactoring that should be in completely separate PRs.

TracerDS:
> Why does this PR contain so much more than it should? According to the title and description of the PR, it adds the onElementCreate and onClientElementCreate events, and meanwhile this PR contains a ton of refstoring that should be in completely separate PRs.

Refactoring is a side job but the main purpose is to add these events to every element type (~12 types * 2).
Its better to make refactors sooner than later as it will add up to the stack

ghost:
I think it should be reconsidered whether adding such event it's good idea, see: [Pirulax's comment](https://github.com/multitheftauto/mtasa-blue/pull/218#issuecomment-726536721)
I believe before introducing such events, an event system refactor it's needed, but it's best to ask @Pirulax itself.

FileEX:
Well, for as long as I remember (since 2018), there has always been a rule that one PR concerns one specific thing.  In my opinion, you should not smuggle other changes unrelated to the PR topic, new events and refactoring are different things.  Right now this PR is doing way more than it should, and this shouldn't be happening even for a bona fide refactoring of old code.  A PR that should be adding two events becomes an opportunity to push other things and edits 62 files - This should never have happened in my opinion.

TracerDS:
> I think it should be reconsidered whether adding such event it's good idea, see: [Pirulax's comment](https://github.com/multitheftauto/mtasa-blue/pull/218#issuecomment-726536721) I believe before introducing such events, an event system refactor it's needed, but it's best to ask @Pirulax itself.

Oof. Didnt see that PR before. I guess until then I will close this as unplanned

PR #3581, opened 2024-07-18
Fix #2936 Toggling car lights makes windows invisible when viewed from inside

### LINKED ISSUE #2936: Toggling car lights makes windows invisible when viewed from inside
### Describe the bug

If you are inside the car and toggle the lights, the car windows become invisible. 
This bug only occurs when you are viewing the windows from inside the car; if you are viewing the car from outside or from a different camera angle, the windows will appear normal.

Here is the bug : 
mta-screen_2023-03-26_01-28-21

When lights are turnt off : 
mta-screen_2023-03-26_01-28-34

### Steps to reproduce

1. Enter a car
2. Turn lights on
3. Exit the car, but leave the door opened
4. Look at the inside of the car (from the opened door)

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #2936
image

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1571,6 +1571,9 @@ void CMultiplayerSA::InitHooks()
     // Allow switch weapon during jetpack task (#3569)
     MemSetFast((void*)0x60D86F, 0x90, 19);
 
+    // Fix invisible vehicle windows when lights are on (#2936)
+    MemPut<BYTE>(0x6E1425, 0);
+
     InitHooks_CrashFixHacks();
 
     // Init our 1.3 hooks.


PlatinMTA:
no way, finally

Fernando-A-Rocha:
Nice, how does the fix work, out of curiosity?

Proxy-99:
> Nice, how does the fix work, out of curiosity?

It set the value of the address to 0 
`MemPut<BYTE>(0x6E1425, 0);`
0x6E1425 set the value to 0 will works if 2 is does not show

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1571,6 +1571,9 @@ void CMultiplayerSA::InitHooks()
     // Allow switch weapon during jetpack task (#3569)
     MemSetFast((void*)0x60D86F, 0x90, 19);
 
+    // Fix invisible vehicle windows when lights are on (#2936)
+    MemPut<BYTE>(0x6E1425, 0);
```
TracerDS, line 1575: What is that mem address?

TracerDS (approved):

Proxy-99 (approved):
lgtm

Client/multiplayer_sa/CMultiplayerSA.cpp
FileEX: CVehicle::DoHeadLightBeam

FileEX:
> Nice, how does the fix work, out of curiosity?

R* incorrectly sets the vehicle's renderstate cull mode to "2" which means front after the light streak has been rendered. Normally when the lights are off renderstatecullmode is set to 0 (none) and that's exactly what this PR does.
``rwRENDERSTATECULLMODE``

Fernando-A-Rocha:
> > Nice, how does the fix work, out of curiosity?
> 
> R* incorrectly sets the vehicle's renderstate cull mode to "2" which means front after the light streak has been rendered. Normally when the lights are off renderstatecullmode is set to 0 (none) and that's exactly what this PR does. `rwRENDERSTATECULLMODE`

Thank you for this explanation!! I had managed to do the same cull fix for vehicles via shader. Nice that it's natively implemented now.

PR #3587, opened 2024-07-20
:sparkles: feat: add setAccountSerial

camargo2019:
This PR adds a new function to edit the serial of a specific account. Currently, when players change their PC, the server owner needs to shut down the server and modify the internal.db file to update the player's serial. With this new function, this process is simplified, eliminating the need to shut down the server.

Example:
```lua
setAccountSerial(getPlayerAccount(thePlayer), "newSerial")
getPlayerAccount(thePlayer):setSerial("newSerial")
```


Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +515,31 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
     return 1;
 }
 
+int CLuaAccountDefs::SetAccountSerial(lua_State* luaVM)
```
ghost: Please use the new argument parser here (https://github.com/multitheftauto/mtasa-blue/wiki/Lua-API#how-to-add-a-new-lua-definition)

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
```diff
@@ -45,6 +45,7 @@ class CLuaAccountDefs : public CLuaDefs
     LUA_DECLARE(AddAccount);
     LUA_DECLARE(RemoveAccount);
     LUA_DECLARE(SetAccountName);
+    LUA_DECLARE(SetAccountSerial);
```
ghost: Please use the new argument parser here

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -70,6 +71,7 @@ void CLuaAccountDefs::AddClass(lua_State* luaVM)
 
     lua_classfunction(luaVM, "setData", "setAccountData");
     lua_classfunction(luaVM, "setPassword", "setAccountPassword");
+    lua_classfunction(luaVM, "setSerial", "SetAccountSerial");
```
ghost: SetAccountSerial should start with a lowercase letter here

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -700,6 +700,7 @@ class CStaticFunctionDefinitions
     // Account set funcs
     static CAccount* AddAccount(const SString& strName, const SString& strPassword, bool bAllowCaseVariations, SString& strOutError);
     static bool      RemoveAccount(CAccount* pAccount);
+    static bool      SetAccountSerial(SString strName, const SString strSerial);
```
ghost: SString is no longer used, please use std::string instead

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11400,6 +11400,22 @@ bool CStaticFunctionDefinitions::GetAccountSerial(CAccount* pAccount, SString& s
     return bRegistered;
 }
 
+
+bool CStaticFunctionDefinitions::SetAccountSerial(SString strName, SString strSerial)
```
ghost: SString is no longer used, please use std::string instead

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -101,6 +101,7 @@ class CAccount
     bool                       AuthorizeSerial(const SString& strSerial, const SString& strWho);
     bool                       RemoveSerial(const SString& strSerial);
     void                       RemoveUnauthorizedSerials();
+    void                       SetAccountSerial(const SString strSerial);
```
ghost: SString is no longer used, please use std::string instead

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -289,6 +289,15 @@ bool CAccount::RemoveSerial(const SString& strSerial)
     return false;
 }
 
+//
+// Replace the serial number for a specific account
+//
+void CAccount::SetAccountSerial(const SString strSerial)
```
ghost: SString is no longer used, please use std::string instead

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +515,31 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
     return 1;
 }
 
+int CLuaAccountDefs::SetAccountSerial(lua_State* luaVM)
+{
+    SString strName;
+    SString strSerial;
```
ghost: SString is no longer used, please use std::string instead

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11400,6 +11400,22 @@ bool CStaticFunctionDefinitions::GetAccountSerial(CAccount* pAccount, SString& s
...
+    CAccount* pAccount = g_pGame->GetAccountManager()->Get(szAccountName);
+    
+    if (!pAccount)
+        return false;
+
+    bool bRegistered = pAccount->IsRegistered();
+    if (bRegistered)
+        pAccount->SetAccountSerial(strSerial);
+
+    return bRegistered;
```
ghost: Avoid using Hungarian notation in new code

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
ghost: Avoid using Hungarian notation in new code

ghost (changes requested):

ghost:
The basic concept of this pull request is understandable, but such a change must also be reviewed from a security point of view, as the modifiability of serials belonging to accounts carries many risks in my opinion.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
camargo2019: Thanks for the feedback, I will avoid using future contributions <3

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: Thanks for the feedback, I will avoid using future contributions <3

camargo2019:
> The basic concept of this pull request is understandable, but such a change must also be reviewed from a security point of view, as the modifiability of serials belonging to accounts carries many risks in my opinion.

So, I believe since the function is only on the server side, no problem

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/CAccount.cpp
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: Good point, I hadn't noticed 62ebf06

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
camargo2019: adjusted 62ebf06

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: I hadn't seen the update, I switched to the new standard 62ebf06

TracerDS:
> This PR adds a new function to edit the serial of a specific account. Currently, when players change their PC, the server owner needs to shut down the server and modify the internal.db file to update the player's serial. With this new function, this process is simplified, eliminating the need to shut down the server.
> 
> Example:
> 
> ```
> setAccountSerial("userName", "newSerial")
> ```

What would be the benefits of that function?

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -289,6 +289,15 @@ bool CAccount::RemoveSerial(const SString& strSerial)
     return false;
 }
 
+//
+// Replace the serial number for a specific account
+//
+void CAccount::SetAccountSerial(const std::string strSerial)
```
ghost: Please avoid using Hungarian notation in new code, `serial` would be neater

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -101,6 +101,7 @@ class CAccount
     bool                       AuthorizeSerial(const SString& strSerial, const SString& strWho);
     bool                       RemoveSerial(const SString& strSerial);
     void                       RemoveUnauthorizedSerials();
+    void                       SetAccountSerial(const std::string strSerial);
```
ghost: Please avoid using Hungarian notation in new code, `serial` would be neater

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11400,6 +11400,22 @@ bool CStaticFunctionDefinitions::GetAccountSerial(CAccount* pAccount, SString& s
...
+}
```
ghost: Please avoid using Hungarian notation in new code

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -700,6 +700,7 @@ class CStaticFunctionDefinitions
     // Account set funcs
     static CAccount* AddAccount(const SString& strName, const SString& strPassword, bool bAllowCaseVariations, SString& strOutError);
     static bool      RemoveAccount(CAccount* pAccount);
+    static bool      SetAccountSerial(std::string strName, std::string strSerial);
```
ghost: Please avoid using Hungarian notation in new code, `name` and `serial` would be neater

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +515,17 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaAccountDefs::SetAccountSerial(std::string strName, std::string strSerial)
```
ghost: Please avoid using Hungarian notation in new code, `name` and `serial` would be neater

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +515,17 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
...
+{
+    if (CStaticFunctionDefinitions::SetAccountSerial(strName, strSerial))
+    {
+        return true;
+    }
+
+    return false;
```
TracerDS: 1. Avoid hungarian notation
2. `return CStaticFunctionDefinitions::SetAccountSerial(strName, strSerial)`
3. `noexcept`

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
```diff
@@ -48,4 +48,5 @@ class CLuaAccountDefs : public CLuaDefs
     LUA_DECLARE(SetAccountPassword);
     LUA_DECLARE(SetAccountData);
     LUA_DECLARE(CopyAccountData);
+    static bool SetAccountSerial(std::string strName, std::string strSerial);
```
TracerDS: `noexcept`

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -289,6 +289,15 @@ bool CAccount::RemoveSerial(const SString& strSerial)
...
+{
+    m_strSerial = strSerial;
+    m_pManager->MarkAsChanged(this);
```
TracerDS: You aren't checking if the serial is even valid. That will cause potential security issues in the future

TracerDS (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +515,17 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
...
+}
+
+
```
ghost:
```suggestion
bool CLuaAccountDefs::SetAccountSerial(const std::string& name, const std::string& serial) noexcept 
{
    return CStaticFunctionDefinitions::SetAccountSerial(name, serial);
}
```

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
ghost:
```suggestion
    static bool SetAccountSerial(const std::string& name, const std::string& serial) noexcept;
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11400,6 +11400,22 @@ bool CStaticFunctionDefinitions::GetAccountSerial(CAccount* pAccount, SString& s
     return bRegistered;
 }
 
+
+bool CStaticFunctionDefinitions::SetAccountSerial(std::string strName, std::string strSerial)
```
ghost: Please use noexcept here

ghost (changes requested):

xLive:
Is there any particular reason why this function uses an account name instead of an account element?

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/CAccount.cpp
camargo2019: adjusted be766cf

Server/mods/deathmatch/logic/CAccount.cpp
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
camargo2019: adjusted 9ad1188

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
camargo2019: adjusted 9ad1188

camargo2019:
> > This PR adds a new function to edit the serial of a specific account. Currently, when players change their PC, the server owner needs to shut down the server and modify the internal.db file to update the player's serial. With this new function, this process is simplified, eliminating the need to shut down the server.
> > Example:
> > ```
> > setAccountSerial("userName", "newSerial")
> > ```
> 
> What would be the benefits of that function?

I believe it would be easier to change the Accounts serial in internal.db

Imagine a server that has 500 players online, and someone from the administration changed computers due to an emergency, they would have to shut down the server just to change its serial number! With this function, we eliminate this need

camargo2019:
> Is there any particular reason why this function uses an account name instead of an account element?

I tried following the setAccountPassword and setAccountName pattern
But I don't have specific reasons.

TracerDS:
> I believe it would be easier to change the Accounts serial in internal.db
> 
> Imagine a server that has 500 players online, and someone from the administration changed computers due to an emergency, they would have to shut down the server just to change its serial number! With this function, we eliminate this need

Why would they need to shut down the server? You can login into the account from 5 different computers without issues.
There isnt one serial per account. Accounts just store **the last serial that logged in**. 
You can store the original serial via `setAccountData` just fine

camargo2019:
> Why would they need to shut down the server? You can login into the account from 5 different computers without issues.
> There isnt one serial per account. Accounts just store **the last serial that logged in**.
> You can store the original serial via `setAccountData` just fine

Let me try to give a better example. I don't use ElementData and I only use the Multi Theft Auto login system. I create a simple `if getAccountSerial(account) == getPlayerSerial(player)` which, in theory, would get the last serial the player connected with and validate it with the current serial, right? But now everyone who is connecting to my server is passing this validation, and the `authserial <account> removelist` is not intended to remove this serial. What should I do in this case? Should I temporarily remove the validation and restart my system? Or should I create my own serial management using ElementData or a database? Or, in this new case, should I use this function to update the serial and use the MTA's own structure?

The idea of bringing this function is just to make things easier

TracerDS:
> Let me try to give a better example. I don't use ElementData and I only use the Multi Theft Auto login system. I create a simple `if getAccountSerial(account) == getPlayerSerial(player)` which, in theory, would get the last serial the player connected with and validate it with the current serial, right? But now everyone who is connecting to my server is passing this validation, and the `authserial <account> removelist` is not intended to remove this serial. What should I do in this case? Should I temporarily remove the validation and restart my system? Or should I create my own serial management using ElementData or a database? Or, in this new case, should I use this function to update the serial and use the MTA's own structure?
> 
> The idea of bringing this function is just to make things easier

You should never rely on just serial as your authentication as they are not reliable.
`authserial ... removelast` will remove the last logged serial from the account. If everyone who is connecting to your server is passing the `if getAccountSerial(account) == getPlayerSerial(player)` then that means that everyone has the same serial.

You should connect several identification methods (ip, serial, specs, etc.) together to in some way identify a player.
This function will bring nothing but chaos to mta imo

camargo2019:
> > Let me try to give a better example. I don't use ElementData and I only use the Multi Theft Auto login system. I create a simple `if getAccountSerial(account) == getPlayerSerial(player)` which, in theory, would get the last serial the player connected with and validate it with the current serial, right? But now everyone who is connecting to my server is passing this validation, and the `authserial <account> removelist` is not intended to remove this serial. What should I do in this case? Should I temporarily remove the validation and restart my system? Or should I create my own serial management using ElementData or a database? Or, in this new case, should I use this function to update the serial and use the MTA's own structure?
> > The idea of bringing this function is just to make things easier
> 
> You should never rely on just serial as your authentication as they are not reliable. `authserial ... removelast` will remove the last logged serial from the account. If everyone who is connecting to your server is passing the `if getAccountSerial(account) == getPlayerSerial(player)` then that means that everyone has the same serial.
> 
> You should connect several identification methods (ip, serial, specs, etc.) together to in some way identify a player. This function will bring nothing but chaos to mta imo

Why do you think this will bring more chaos to MTA? Nowadays, server owners already perform this action, but most often manually through the internal.db. Shops usually carry out this type of validation, so what we can do to increase security is to prevent malicious scripts by adding appropriate rules to the acl.xml.


And also, `removelist` only removes the serials from the `serialusage` table, not from accounts. The `getAccountSerial` function retrieves the serial from accounts.

TracerDS:
> Why do you think this will bring more chaos to MTA? Nowadays, server owners already perform this action, but most often manually through the internal.db. Shops usually carry out this type of validation, so what we can do to increase security is to prevent malicious scripts by adding appropriate rules to the acl.xml.

acl.xml have no effect on client side functions and events. It only prevents server side code from doing stuff.
I see no reason why this function should be added to mta. `deleteAccountLastSerial` for clearing last serial would be good.
Setting it manually is not good

camargo2019:
> acl.xml have no effect on client side functions and events. It only prevents server side code from doing stuff.
> I see no reason why this function should be added to mta. `deleteAccountLastSerial` for clearing last serial would be good.
> Setting it manually is not good

But this function is on the server side

TracerDS:
> But this function is on the server side

I understand that. I was responding to
> what we can do to increase security is to prevent malicious scripts by adding appropriate rules to the acl.xml.

If you have backdoors in your scripts (or download malware scripts to your server) then no matter what functions you have at your disposal, its useless.

camargo2019:
> > But this function is on the server side
> 
> I understand that. I was responding to
> 
> > what we can do to increase security is to prevent malicious scripts by adding appropriate rules to the acl.xml.
> 
> If you have backdoors in your scripts (or download malware scripts to your server) then no matter what functions you have at your disposal, its useless.

So you think it’s better to delete the serial rather than setting a new one? By modifying the function to `deleteAccountLastSerial`?

jlillis:
> when players change their PC, the server owner needs to shut down the server and modify the internal.db file to update the player's serial. 

I assume by players, you mean admins and members of other ACL groups you have enabled [`authserial`](https://wiki.multitheftauto.com/wiki/Authorized_Serial_Account_Protection) protection for? You should be able to get them access by following the instructions on the wiki. Adding a new function for that purpose should not be necessary.

This function would still be useful if you're just looking to update the serial returned by `getAccountSerial`. ~~It does raise another question - should `getAccountSerial` be returning the serial the account was registered with or the last serial used to log in? Maybe out of scope for this discussion.~~

> If you have backdoors in your scripts (or download malware scripts to your server) then no matter what functions you have at your disposal, its useless.

Script backdoors are not a threat that the `authserial` system is meant to address. The ACL system will mitigate some of that threat, but it is ultimately the responsibility of the server owner to manage. Protecting this function in the same way as `setAccountPassword` is sufficient.

camargo2019:
> > This function would still be useful if you're just looking to update the serial returned by getAccountSerial. It does raise another question - should getAccountSerial be returning the serial the account was registered with or the last serial used to log in? Maybe out of scope for this discussion.

This function does exactly that: it updates the serial from `getAccountSerial`

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -255,6 +255,15 @@ bool CAccount::IsIpAuthorized(const SString& strIp)
     return false;
 }
 
+//
+// Check if the serial has 32 hexadecimal characters
+//
+bool CAccount::IsValidSerial(const std::string& serial)
```
ghost:
```suggestion
bool CAccount::IsValidSerial(const std::string& serial) noexcept
```

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -289,6 +298,19 @@ bool CAccount::RemoveSerial(const SString& strSerial)
     return false;
 }
 
+//
+// Replace the serial number for a specific account
+//
+bool CAccount::SetAccountSerial(const std::string& serial)
```
ghost:
```suggestion
bool CAccount::SetAccountSerial(const std::string& serial) noexcept
```

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -97,10 +98,12 @@ class CAccount
     SSerialUsage*              GetSerialUsage(const SString& strSerial);
     bool                       IsIpAuthorized(const SString& strIp);
     bool                       IsSerialAuthorized(const SString& strSerial);
+    bool                       IsValidSerial(const std::string& serial);
```
ghost:
```suggestion
    bool                       IsValidSerial(const std::string& serial) noexcept;
```

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -97,10 +98,12 @@ class CAccount
...
     bool                       AddSerialForAuthorization(const SString& strSerial, const SString& strIp);
     bool                       AuthorizeSerial(const SString& strSerial, const SString& strWho);
     bool                       RemoveSerial(const SString& strSerial);
     void                       RemoveUnauthorizedSerials();
+    bool                       SetAccountSerial(const std::string& serial);
```
ghost:
```suggestion
    bool                       SetAccountSerial(const std::string& serial) noexcept;
```

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -70,6 +71,7 @@ void CLuaAccountDefs::AddClass(lua_State* luaVM)
 
     lua_classfunction(luaVM, "setData", "setAccountData");
     lua_classfunction(luaVM, "setPassword", "setAccountPassword");
+    lua_classfunction(luaVM, "setSerial", "setAccountSerial");
```
ghost: This won't work as you are using account name instead of element.
To make it work, use element instead of name or delete this line.
```suggestion
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
```diff
@@ -700,6 +700,7 @@ class CStaticFunctionDefinitions
     // Account set funcs
     static CAccount* AddAccount(const SString& strName, const SString& strPassword, bool bAllowCaseVariations, SString& strOutError);
     static bool      RemoveAccount(CAccount* pAccount);
+    static bool      SetAccountSerial(const std::string& name, const std::string& serial) noexcept;
```
ghost: To make the OOP syntax work, use element instead.
```suggestion
    static bool      SetAccountSerial(CAccount* account, const std::string& serial) noexcept;
```

ghost (changes requested):

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.h
camargo2019: adjusted 7418e4c

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: Good point. At first, I had done it with the account, but I ended up removing it to make it more practical 7418e4c

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 3bdd277

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 3bdd277

Server/mods/deathmatch/logic/CAccount.cpp
camargo2019: adjusted 3bdd277

Server/mods/deathmatch/logic/CAccount.cpp
camargo2019: adjusted 3bdd277

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11400,6 +11400,18 @@ bool CStaticFunctionDefinitions::GetAccountSerial(CAccount* pAccount, SString& s
     return bRegistered;
 }
 
+
+bool CStaticFunctionDefinitions::SetAccountSerial(const std::string& name, const std::string& serial) noexcept
```
ghost: I think it would be better to use `CAccount` since the other functions use it as well
```suggestion
bool CStaticFunctionDefinitions::SetAccountSerial(CAccount* account, const std::string& serial) noexcept
```

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
```diff
@@ -513,6 +514,12 @@ int CLuaAccountDefs::RemoveAccount(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaAccountDefs::SetAccountSerial(std::string name, std::string serial) noexcept
```
ghost: I think it would be better to use `CAccount` since the other functions use it as well
```suggestion
bool CLuaAccountDefs::SetAccountSerial(CAccount* account, std::string serial) noexcept
```

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
```diff
@@ -48,4 +48,5 @@ class CLuaAccountDefs : public CLuaDefs
     LUA_DECLARE(SetAccountPassword);
     LUA_DECLARE(SetAccountData);
     LUA_DECLARE(CopyAccountData);
+    static bool SetAccountSerial(std::string name, std::string serial) noexcept;
```
ghost: I think it would be better to use `CAccount` since the other functions use it as well
```suggestion
static bool SetAccountSerial(CAccount* account, std::string serial) noexcept;
```

ghost (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.h
camargo2019: adjusted 790ffa2

Server/mods/deathmatch/logic/luadefs/CLuaAccountDefs.cpp
camargo2019: adjusted 790ffa2

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
camargo2019: adjusted 790ffa2

Server/mods/deathmatch/logic/CAccount.cpp
```diff
@@ -255,6 +255,15 @@ bool CAccount::IsIpAuthorized(const SString& strIp)
     return false;
 }
 
+//
+// Check if the serial has 32 hexadecimal characters
+//
+bool CAccount::IsValidSerial(const std::string& serial) noexcept
+{
+    const std::regex serialPattern("^[A-Fa-f0-9]{32}$");
+    return std::regex_match(serial, serialPattern);
```
TracerDS: this function can throw `std::regex_error`. Please wrap this regex_match in try-catch:
```cpp
bool CAccount::IsValidSerial(const std::string& serial) noexcept
{
    const std::regex serialPattern("^[A-Fa-f0-9]{32}$");
    try {
        return std::regex_match(serial, serialPattern);
    } catch (...) {
        return false;
    }
}
```

camargo2019: adjusted 9a14e49

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -13,6 +13,7 @@
 
 #include <list>
 #include <string>
+#include <regex>
```
TracerDS: move it to the source file (`Server/mods/deathmatch/logic/CAccount.cpp`)

Server/mods/deathmatch/logic/CAccount.h
```diff
@@ -97,10 +98,12 @@ class CAccount
     SSerialUsage*              GetSerialUsage(const SString& strSerial);
     bool                       IsIpAuthorized(const SString& strIp);
     bool                       IsSerialAuthorized(const SString& strSerial);
+    bool                       IsValidSerial(const std::string& serial) noexcept;
```
TracerDS: add `const` qualifier:
```cpp
bool IsValidSerial(const std::string& serial) const noexcept;
```

TracerDS (changes requested):

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 94ac014

Server/mods/deathmatch/logic/CAccount.h
camargo2019: adjusted 94ac014

TracerDS:
@camargo2019 please resolve old conversations as they widen the page 😅

camargo2019:
> @camargo2019 please resolve old conversations as they widen the page 😅

@TracerDS done 🙃

TracerDS (approved):

Lpsd (commented):
Do we still need to add `setAccountSerial` to [acl.xml](https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/acl.xml)?

camargo2019:
> Do we still need to add `setAccountSerial` to [acl.xml](https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/acl.xml)?

I believe it would make sense. What do you think? Should we add this function to the acl.xml?

TracerDS:
> Do we still need to add `setAccountSerial` to [acl.xml](https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/acl.xml)?

Why not?

Lpsd:
Yeah, add it and this seems good.

camargo2019:
> Yeah, add it and this seems good.

I added it here 63baddc 🙃

ghost (approved):
LGTM

Lpsd:
Can the acl right be put in the same group as the other setAccount* functions?

camargo2019:
@Lpsd Do you say put it in the Moderator?

ghost:
> @Lpsd Do you say put it in the Moderator?

I think so
If you look at the other entries, you can see that they also belong to the moderator group

CrosRoad95:
Why not just add option to have multiple serials allow to login to the same account?

camargo2019:
@Nico8340 adjusted

@CrosRoad95 The idea is to update the serial from getAccountSerial and not from authserial

ghost (approved):

CrosRoad95:
Also, if you add setAccountSerial function you in practice introducing a way to bypass already existing protections EVEN you updated acl, why you may guess? People don't updating, reusing resources, and in this resources often lies outdated acl.xml.

Good example is dgs, even thisdp fixed vulnerability there YEARS ago, it is still present on many, many servers

camargo2019:
@CrosRoad95 I agree with you; I believe that ACL is not the most secure resource. However, the function resides on the server side, so it is the total responsibility of whoever is setting up their server to be aware of and consent to the use of this function

We should not be held responsible for the misuse of certain functions, such as debugHook, which many people are using nowadays to break the protections of resource stores. We do not take action on this because it is not within our jurisdiction nor is it our responsibility to address the misuse of these functions. 

Remembering that this is just an example, the case mentioned has nothing to do with this function.

Fernando-A-Rocha:
[`getAccountSerial`](https://wiki.multitheftauto.com/wiki/GetAccountSerial) returns the last [serial](https://wiki.multitheftauto.com/wiki/Serial) that logged onto the specified [account](https://wiki.multitheftauto.com/wiki/Account).

In my opinion, you shouldn't use a function to alter this value. It's automatic...

Instead, use [`setAccountData`](https://wiki.multitheftauto.com/wiki/SetAccountData) or whatever for security and serial validation purposes.

botder:
Personally, I don't think it's the best way to design your validation around `getAccountSerial`, but it would also be stupid to duplicate the information with `setAccountData`.

PR #3592, opened 2024-07-21
Add new function spawnVehicleFlyingComponent

### LINKED ISSUE #2121: Implement spawnFlyingComponent in a function, like makeVehicleComponentFly()

FileEX:
Closes #2121 and probably #2128

This PR adds a new function ``spawnVehicleFlyingComponent``. With this function, we can spawn falling components along with their physics. This enables the creation of realistic scenarios such as car accidents on servers, advanced systems like potholes in roads, and more. 
Example video: https://www.youtube.com/watch?t=71&v=B5relHf21lo&feature=youtu.be

Once the PR is merged, I will create full documentation on the wiki for various nodes for vehicle classes.

## Syntax
```
bool spawnVehicleFlyingComponent(element vehicle, number nodeIndex [, number collisionType, number removalTime ] )
```
- `nodeIndex` - Specifies the component to be created (ranging from 1 to 25). Depending on the vehicle, these can be different components such as wheels, fenders, bicycle handlebars, helicopter propellers, train carriages, and many more.
- `collisionType` - Specifies the type of collision for the component, by default it is selected based on the nodeIndex.

```
0 - COL_BUMPER
1 - COL_WHEEL
2 - COL_DOOR
3 - COL_BONNET
4 - COL_BOOT
5 - COL_PANEL
```

- `removalTime` - The time in milliseconds after which the created component will be removed (it must be removed as it is a temporary object).

The limit for temporary objects with this function is 150, and their default removal time depends on their quantity (unless we set our own via the `removalTime` parameter).


Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,119 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+            if (nodeIndex == 1)
+            {
+                defaultBikeChassisFrame = bikeInterface->m_apModelNodes[1];
+                if (defaultBikeChassisFrame && std::strcmp(defaultBikeChassisFrame->szName, "chassis_dummy") == 0)
+                {
+                    RwFrame* correctChassisFrame = ((RwFrame*(__cdecl*)(RpClump*, const char*))FUNC_CClumpModelInfo_GetFrameFromName)(bikeInterface->m_pRwObject, "chassis");
+                    if (correctChassisFrame)
+                        bikeInterface->m_apModelNodes[1] = correctChassisFrame;
+                }
+            }
```
TracerDS: early return

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,119 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+            nodesOffset = OFFSET_CBoat_Nodes;
+            break;
+        }
+        default:
+            return false;
+    }
+
+    // Patch nodes array
+    MemPut(0x6A85B3, nodesOffset);
+    MemPut(0x6A8631, nodesOffset);
```
TracerDS, line 1572: What are these mem addresses?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,119 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+
+    if (removalTime > -1)
+    {
+        auto* componentObject = reinterpret_cast<CObjectSAInterface*>((DWORD*)dwReturn);
+        if (!componentObject)
+            return;
+
+        int CTimer_ms = *(int*)0xB7CB84; // CTimer::m_snTimeInMilliseconds
+        componentObject->uiObjectRemovalTime = CTimer_ms + static_cast<std::uint32_t>(removalTime);
+    }
```
TracerDS: early return

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,17 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+    CClientVehicle* vehicle = m_pVehicleManager->Get(sourceEntity->GetID());
+    if (!vehicle)
+        return;
+
+    std::uint8_t nodeIndex;
+    std::uint8_t collisionType;
+    std::int32_t removalTime;
+
+    if (bitStream.Read(nodeIndex) && bitStream.Read(collisionType) && bitStream.Read(removalTime))
+        vehicle->SpawnFlyingComponent(static_cast<eCarNodes>(nodeIndex), static_cast<eCarComponentCollisionTypes>(collisionType), removalTime);
```
TracerDS, line 668: maybe early continue?

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4174,3 +4177,68 @@ bool CLuaVehicleDefs::BlowVehicle(CClientEntity* entity, std::optional<bool> wit
...
+                break;
+            }
+        }
+    }
+    else
+    {
+        collisionType = static_cast<eCarComponentCollisionTypes>(componentCollisionType.value());
+    }
+
+    return vehicle->SpawnFlyingComponent(static_cast<eCarNodes>(nodeIndex), collisionType, removalTime.value_or(-1));
```
TracerDS: This function differs from server side equivalent (mainly how enums are interpreted).
Can you make these 2 sides the same?

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
FileEX: The difference is that the enum type cannot be sent via BitStream, so std::uint8_t is used on the server side instead of ``eCarComponentCollisionTypes`` enum

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
FileEX: There is no need, this is just one short condition. The guidelines are to avoid nesting conditions within each other, not to avoid them altogether

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,118 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+                break;
+
+            // Set the correct "bike_chassis" frame for bikes
+            defaultBikeChassisFrame = bikeInterface->m_apModelNodes[1];
+            if (defaultBikeChassisFrame && std::strcmp(defaultBikeChassisFrame->szName, "chassis_dummy") == 0)
+            {
+                RwFrame* correctChassisFrame = ((RwFrame * (__cdecl*)(RpClump*, const char*))FUNC_CClumpModelInfo_GetFrameFromName)(bikeInterface->m_pRwObject, "chassis");
+                if (correctChassisFrame)
+                    bikeInterface->m_apModelNodes[1] = correctChassisFrame;
+            }
```
TracerDS: maybe early return?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,118 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+        push nodeIndex
+        call dwFunc
+        mov dwReturn, eax
     }
 
-    CObject* pObject = NULL;
-    if (dwReturn)
+    // Restore default nodes array in CAutomobile::SpawnFlyingComponent
+    MemPut(0x6A85B3, 0x648);
+    MemPut(0x6A8631, 0x648);
```
TracerDS, line 1579: What are those mem addresses?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1507,27 +1508,118 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
-    return pObject;
+
+    if (removalTime <= -1)
+        return true;
+
+    auto* componentObject = reinterpret_cast<CObjectSAInterface*>((DWORD*)dwReturn);
+    if (!componentObject)
+        return;
+
+    int CTimer_ms = *(int*)0xB7CB84; // CTimer::m_snTimeInMilliseconds
```
TracerDS: Can you make it as a preprocessor definition?

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TracerDS: > The difference is that the enum type cannot be sent via BitStream, so std::uint8_t is used on the server side instead of `eCarComponentCollisionTypes` enum

that wouldnt matter though, you can cast uint to the enum/make the enum accept uint8 values (like in every other enum that can be passed to ArgumentParser)

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
TracerDS: It would make it much more readable though:
```cpp
if (!bitStream.Read(nodeIndex))
    return;
if(!bitStream.Read(collisionType))
    return;
if(!bitStream.Read(removalTime))
    return;
vehicle->SpawnFlyingComponent(static_cast<eCarNodes>(nodeIndex), static_cast<eCarComponentCollisionTypes>(collisionType), removalTime);

```

tederis: An example of inappropriate use of early return. This make noise instead of gaining readability.

TracerDS: Disagree. This makes the code much more cleaner and easier to maintain.
Having very long lines, long if statements creates chaos.

tederis:
@FileEX Please, resolve the conflicts.

FileEX:
> @FileEX Please, resolve the conflicts.

Done

Client/game_sa/CAutomobileSA.h
```diff
@@ -20,39 +20,7 @@
...
-        EXHAUST = 19,
-        MISC_A = 20,
-        MISC_B = 21,
-        MISC_C = 22,
-        MISC_D = 23,
-        MISC_E = 24,
-        NUM_NODES
-    };
-};
+#define MAX_NODES               25
```
tederis: `MAX_NODES` sounds vague and can be applied to many things. It should be named according to the field in which this definition is used.

tederis (changes requested):

Client/game_sa/CAutomobileSA.h
FileEX: Will ``MAX_COMPONENT_NODES`` be better?

tederis: Maybe `CAR_NODES_NUM`?

FileEX: Done

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1501,27 +1502,118 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
```
tederis: From what I can see it should be DWORD (or uint32_t). See gta_reversed and CGameSA line 192.

Client/game_sa/CBikeSA.h
```diff
@@ -14,6 +14,25 @@
 #include <game/CBike.h>
 #include "CVehicleSA.h"
 
+namespace eBikeNodes
+{
+    enum
+    {
+        NONE = 0,
+        CHASIS,
```
tederis: Should be CHASSIS. The same for below one.

Client/game_sa/CBikeSA.h
```diff
@@ -14,6 +14,25 @@
 #include <game/CBike.h>
 #include "CVehicleSA.h"
 
+namespace eBikeNodes
+{
+    enum
```
TheNormalnij: Use `enum class` in all cases

Client/game_sa/CVehicleSA.cpp
FileEX: Yeah, that's right

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1501,27 +1502,118 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
-        mov     ecx, dwThis
-        push    ui_2
-        push    i_1
-        call    dwFunc
-        mov     dwReturn, eax
+        mov ecx, dwInterface
+        push collisionType
+        push nodeIndex
+        call dwFunc
+        mov dwReturn, eax
```
TheNormalnij: you can avoid assembler here

Client/game_sa/CBikeSA.h
FileEX: Done, I did a little more cleaning

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1501,27 +1502,97 @@ void CVehicleSA::SetGravity(const CVector* pvecGravity)
...
+            return false;
     }
 
-    CObject* pObject = NULL;
-    if (dwReturn)
+    // Patch nodes array in CAutomobile::SpawnFlyingComponent
+    MemPut(0x6A85B3, nodesOffset);
+    MemPut(0x6A8631, nodesOffset);
+
+    auto* componentObject = ((CObjectSAInterface * (__thiscall*)(CVehicleSAInterface*, int, int)) FUNC_CAutomobile__SpawnFlyingComponent)(GetVehicleInterface(), static_cast<int>(nodeIndex), static_cast<int>(collisionType));
```
tederis, line 1574: `componentObject` can be null in some conditions (tested).

FileEX: Fixed

tederis (approved):
Looks good to me

TheNormalnij:
@FileEX documentation

FileEX:
> @FileEX documentation

It's already done

syncc01:
add wiki example

TheNormalnij:
> It's already done

A link was missing in [the template](https://wiki.multitheftauto.com/wiki/Template:Vehicle_functions). I've added it

PR #3593, opened 2024-07-21
Make frame graph scale accordingly to resolution.

ffsPLASMA:
Instead of hardcoding pixels to draw the frame graph, use screen resolution to adapt for dynamic sizes.

### DIFF
--- a/Client/core/CChat.cpp
+++ b/Client/core/CChat.cpp
@@ -1117,7 +1117,7 @@ void CChat::SetCharacterLimit(int charLimit)
     m_iCharacterLimit = charLimit;
 }
 
-float CChat::GetChatBottomPosition()
+float CChat::GetChatBottomPosition() const noexcept
 {
     return m_vecBackgroundSize.fY;
 }
--- a/Client/core/CChat.cpp
+++ b/Client/core/CChat.cpp
@@ -1117,6 +1117,11 @@ void CChat::SetCharacterLimit(int charLimit)
     m_iCharacterLimit = charLimit;
 }
 
+float CChat::GetChatBottomPosition()
+{
+    return m_vecBackgroundSize.fY;
+}
+
 CChatLine::CChatLine()
 {
     m_bActive = false;
--- a/Client/core/CGUI.cpp
+++ b/Client/core/CGUI.cpp
@@ -449,7 +449,7 @@ CChat* CLocalGUI::GetChat()
     return m_pChat;
 }
 
-float CLocalGUI::GetChatBottomPosition()
+float CLocalGUI::GetChatBottomPosition() const noexcept
 {
     return m_pChat->GetChatBottomPosition();
 }
--- a/Client/core/CGUI.cpp
+++ b/Client/core/CGUI.cpp
@@ -449,6 +449,11 @@ CChat* CLocalGUI::GetChat()
     return m_pChat;
 }
 
+float CLocalGUI::GetChatBottomPosition()
+{
+    return m_pChat->GetChatBottomPosition();
+}
+
 CDebugView* CLocalGUI::GetDebugView()
 {
     return m_pDebugView;
--- a/Client/core/CGraphStats.cpp
+++ b/Client/core/CGraphStats.cpp
@@ -113,8 +113,8 @@ void CGraphStats::AddTimingPoint(const char* szName)
 
     CGraphicsInterface* pGraphics = g_pCore->GetGraphics();
 
-    uint uiViewportWidth = pGraphics->GetViewportWidth();
-    uint uiSizeX = uiViewportWidth / 4;
+    std::uint32_t viewportWidth = pGraphics->GetViewportWidth();
+    std::uint32_t sizeX = viewportWidth / 4;
 
     // Start of next frame?
     if (szName[0] == 0)
@@ -136,7 +136,7 @@ void CGraphStats::AddTimingPoint(const char* szName)
                 for (int i = 0; i < Dups; i++)
                 {
                     pLine->iDataPos++;
-                    if (pLine->iDataPos > uiSizeX - 1)
+                    if (pLine->iDataPos > sizeX - 1)
                         pLine->iDataPos = 0;
                     pLine->dataHistory[pLine->iDataPos] = Data;
                 }
@@ -156,7 +156,7 @@ void CGraphStats::AddTimingPoint(const char* szName)
         // Add new line
         MapSet(m_LineList, szName, SGraphStatLine());
         pLine = MapFind(m_LineList, szName);
-        pLine->dataHistory.resize(uiSizeX);
+        pLine->dataHistory.resize(sizeX);
         memset(&pLine->dataHistory[0], 0, pLine->dataHistory.size());
         pLine->iDataPos = 0;
         pLine->prevData = 0;
@@ -182,7 +182,7 @@ void CGraphStats::AddTimingPoint(const char* szName)
 
     // Inc position
     pLine->iDataPos++;
-    if (pLine->iDataPos > uiSizeX - 1)
+    if (pLine->iDataPos > sizeX - 1)
         pLine->iDataPos = 0;
 
     // Insert data point
--- a/Client/core/CGraphStats.cpp
+++ b/Client/core/CGraphStats.cpp
@@ -12,8 +12,6 @@
 
 namespace
 {
-    #define GRAPHSTAT_HISTORY_SIZE 256
-
     struct SGraphStatLine
     {
         TIMEUS              prevData;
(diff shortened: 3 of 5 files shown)


Client/core/CGraphStats.cpp
```diff
@@ -113,6 +111,11 @@ void CGraphStats::AddTimingPoint(const char* szName)
     if (!IsEnabled())
         return;
 
+    CGraphicsInterface* pGraphics = g_pCore->GetGraphics();
+
+    uint uiViewportWidth = pGraphics->GetViewportWidth();
+    uint uiSizeX = uiViewportWidth / 4;
```
TracerDS: use `std::uint32_t` please and dont use hungarian notation.

Why `width / 4`?

Client/core/CGraphStats.cpp
```diff
@@ -200,11 +203,12 @@ void CGraphStats::Draw()
...
 
+    uint  uiViewportWidth = pGraphics->GetViewportWidth();
     uint  uiViewportHeight = pGraphics->GetViewportHeight();
     uint  uiOriginX = 10;
-    uint  uiOriginY = std::min<int>(500, uiViewportHeight - 10);
-    uint  uiSizeX = GRAPHSTAT_HISTORY_SIZE;
-    uint  uiSizeY = 150;
+    uint  uiOriginY = uiViewportHeight / 2;
+    uint  uiSizeX = uiViewportWidth / 4;
+    uint  uiSizeY = uiViewportHeight / 4;
```
TracerDS: Why these values?

TracerDS (changes requested):

Client/core/CGraphStats.cpp
ffsPLASMA: Because the graph was always alligned to left side right under the chat window.

Client/core/CGraphStats.cpp
ffsPLASMA: width / 4 to be used to allign properly with the graph drawing down below.

Fernando-A-Rocha:
Screenshots of before and after?

ffsPLASMA:
I should have added, its very notable on large screen resolutions, for example I use 2K.

Currently:
mta-screen_2024-07-21_10-25-58

With this commit:
mta-screen_2024-07-21_10-25-01

FileEX (approved):
If it looks equally good on other resolutions (especially small ones), then lgtm

ffsPLASMA:
Made some additional screenshots on how it looks like with different resolutions.
Keep in mind I use custom chatbox layout.

800x600
mta-screen_2024-07-21_18-39-59

1080x720
mta-screen_2024-07-21_18-40-48

1920x1080
mta-screen_2024-07-21_18-41-56

2560x1440
look above.

TracerDS:
Could you do it so that the frame graph is always below the chatbox?

Fernando-A-Rocha:
> Could you do it so that the frame graph is always below the chatbox?

Yea, can you ensure there is a minimal distance between the bottom of the chatbox and the top of the framegraph box?

ffsPLASMA:
dont know how to properly access chat interface.

ffsPLASMA:
Framegraph now always sits right under chat with a small gap.
special thanks to @MegadreamsBE for help!

mta-screen_2024-08-15_20-00-07

mta-screen_2024-08-15_19-59-59

mta-screen_2024-08-15_19-59-41

TracerDS:
> Framegraph now always sits right under chat with a small gap. special thanks to @MegadreamsBE for help!

Im sorry for nitpicking but... can you center the framegraph so the X axis would be the same?
Just a little bit to the right for aestheticness

bum8hj:
Does it stay in the screen's vertical bounds?

Though I don't know who would be displaying such a large number of chat lines that this would be an issue 💀

ffsPLASMA:
> Does it stay in the screen's vertical bounds?
> 
> Though I don't know who would be displaying such a large number of chat lines that this would be an issue 💀

no it doesnt stay in bounds but thats not an issue I would care about anyway.





> > Framegraph now always sits right under chat with a small gap. special thanks to @MegadreamsBE for help!
> 
> Im sorry for nitpicking but... can you center the framegraph so the X axis would be the same? Just a little bit to the right for aestheticness

I could do that but again the chat offset to left screen border can be changed in settings. worst case scenario, someone puts chat on top right of screen which would cut off the right side of framegraph.

PlatinMTA:
Thanks for your work, I'm going to be sincere and just tell you that scaled framegraph doesn't look good. It uses too much real estate. I would much prefer just having the changes on its position in comparison to the chat (did you test it with other chat positions? some people have the chat in the lower middle part of the screen, maybe that tries to make the graph render on an Y position that we wouldnt be able to see).

Maybe this doesn't need to be scaled 1:1, maybe it can grow but not proportionally. Or we could set a max value for the scale, something like x2.

ghost (commented):
Nice work!
Noticing the hardcoded offsets like 10 and 30, you might want to consider using #define to set these values as macros.

ffsPLASMA:
> Thanks for your work, I'm going to be sincere and just tell you that scaled framegraph doesn't look good. It uses too much real estate. I would much prefer just having the changes on its position in comparison to the chat (did you test it with other chat positions? some people have the chat in the lower middle part of the screen, maybe that tries to make the graph render on an Y position that we wouldnt be able to see).
> 
> Maybe this doesn't need to be scaled 1:1, maybe it can grow but not proportionally. Or we could set a max value for the scale, something like x2.

The reason Ive made this PR is simply to make the graph render in a bigger size on larger resolutions. You can see above, on 2K resolution/screen its just sooo tiny. We all agree on the hardcoded pixels came from a time when lower resolutions were common, I would assume they used 800x600 as a base. But we have to move on and adapt dynamically for more modern resolutions.

Yes people who have chat on bottom wont see the frame graph unfortunately. We would then have to agree on whether frame graph should stick to one position or move with chatbox height.

Client/core/CGraphStats.cpp
```diff
@@ -202,14 +202,18 @@ void CGraphStats::Draw()
...
+
+    uint  uiViewportWidth = pGraphics->GetViewportWidth();      // get width of current resolution
+    uint  uiViewportHeight = pGraphics->GetViewportHeight();    // get height of current resolution
+    uint  uiOriginX = 10;                                       // offset the graph by 10 pixels from left side of screen
+    uint  uiOriginY = pLocalGUI->GetChatBottomPosition();       // get chat bottom screen position
+    uint  uiSizeX = uiViewportWidth / 4;                        // set the width of graph to 1/4 of current resolution
+    uint  uiSizeY = uiViewportHeight / 4;                       // set the height of graph to 1/4 of current resolution
+    uint  uiRangeY = 100;                                       // 100ms
+
+    uiOriginY = uiOriginY + uiSizeY + 30;                       // add graph height plus a little gap to the overall Y position
```
FileEX: Remove the Hungarian notation and use std::uint16_t

Client/core/CChat.cpp
```diff
@@ -1117,6 +1117,11 @@ void CChat::SetCharacterLimit(int charLimit)
     m_iCharacterLimit = charLimit;
 }
 
+float CChat::GetChatBottomPosition()
+{
+    return m_vecBackgroundSize.fY;
+}
```
TracerDS: Const noexcept

Client/core/CGUI.cpp
```diff
@@ -449,6 +449,11 @@ CChat* CLocalGUI::GetChat()
     return m_pChat;
 }
 
+float CLocalGUI::GetChatBottomPosition()
+{
+    return m_pChat->GetChatBottomPosition();
+}
```
TracerDS: Const noexcept

Client/core/CGraphStats.cpp
```diff
@@ -199,13 +202,18 @@ void CGraphStats::Draw()
...
     CGraphicsInterface* pGraphics = g_pCore->GetGraphics();
+    CLocalGUI*          pLocalGUI = g_pCore->GetLocalGUI();
+
+    uint  uiViewportWidth = pGraphics->GetViewportWidth();      // get width of current resolution
+    uint  uiViewportHeight = pGraphics->GetViewportHeight();    // get height of current resolution
+    uint  uiOriginX = 10;                                       // offset the graph by 10 pixels from left side of screen
+    uint  uiOriginY = pLocalGUI->GetChatBottomPosition();       // get chat bottom screen position
+    uint  uiSizeX = uiViewportWidth / 4;                        // set the width of graph to 1/4 of current resolution
+    uint  uiSizeY = uiViewportHeight / 4;                       // set the height of graph to 1/4 of current resolution
+    uint  uiRangeY = 100;                                       // 100ms
```
TracerDS: `std::uint32_t`

TracerDS (changes requested):

Client/core/CGraphStats.cpp
FileEX: Will any of these values ​​be greater than 65535?

ffsPLASMA: unlikely but you could force a resolution greater than 65535. also not sure if chat position setting itself got some bound checks.

Client/core/CGraphStats.cpp
ffsPLASMA: I used std:uint32_t as tracer suggested below

TheNormalnij (approved):

TheNormalnij:
Thx @ffsPLASMA

PR #3594, opened 2024-07-21
Remove unnecessary argument in engineRestreamWorld

TheNormalnij:
This PR removes `includeLODs` argument from `engineRestreamWorld`. Now `includeLODs` is false by default and it's always true in this PR.
1. This argument has very limited usage. It's better to use `engineRestreamWorld( true )` always
2. By default `engineRestreamWorld` reloads some LOD's. So `includeLODs` argument is frustrating.
3. This function doesn't use arguments for vehicles / peds / world models. So this argument is odd.
4. `engineRestreamWorld` uses old argument parser and new argument parser together.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -5589,7 +5589,7 @@ void CClientGame::ResetMapInfo()
     if (pPlayerInfo)
         pPlayerInfo->SetCamDrunkLevel(static_cast<byte>(0));
 
-    RestreamWorld(true);
+    RestreamWorld();
 
     ReinitMarkers();
 }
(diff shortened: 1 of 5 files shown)


Lpsd (approved):
Seems good, I am just wondering how much impact this will have - `includeLODs` argument to `engineRestreamWorld` is currently false by default; now we are making this functionality enabled by default and removing the argument. In what scenario would you require this argument to be false in the current implementation?

Basically, is there any reason someone would _need_ to omit the `g_pGame->GetStreaming()->RemoveBigBuildings()` call? As we're now making that impossible.

Meine1:
is it possible to make arguments so that only it restreams ped/vehicle/objects instead of all of them at once?

ghost:
> is it possible to make arguments so that only it restreams ped/vehicle/objects instead of all of them at once?

Yes, I would suggest this too, or maybe separate functions, like engineRestreamVehicles, engineRestreamObjects, engineRestreamPeds? Or as an argument, like "all", "vehicles", "objects", "peds".

tederis (approved):

TheNormalnij:
> is it possible to make arguments so that only it restreams ped/vehicle/objects instead of all of them at once?

It's possible for vehicles and peds if they don't shared textures with another models. But I can't guarantee that.

Meine1:
> It's possible for vehicles and peds if they don't shared textures with another models. But I can't guarantee that.

Optional arguments for just restreaming the map itself, just vehicles or just peds would be really useful.

FileEX (approved):

PR #3596, opened 2024-07-21
Fix #523 Helicopter rotor unaffected by vehicle alpha

### LINKED ISSUE #523: Helicopter rotor unaffected by vehicle alpha
**Describe the bug**
Not a big deal, but changing a helicopter's alpha will not change the rotors' (both head and tail rotors) alpha. They will always remain at 255.

This was not previously the case - rotors used to match the alpha of the rest of the vehicle.

This affects code in the editor used to hide elements (ryden's code). Although alternative methods can be used, this is still an issue.

**To Reproduce**

1. Get in helicopter.
2. `crun localPlayer.vehicle.alpha = 0`

**Expected behaviour**
The rotor should be invisible.

**Screenshots**
http://youtu.be/f6BGvkjsji8

**Additional context**
From https://bugs.mtasa.com/view.php?id=4194

FileEX:
Fixed #523 

image


Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1574,6 +1574,10 @@ void CMultiplayerSA::InitHooks()
     // Fix invisible vehicle windows when lights are on (#2936)
     MemPut<BYTE>(0x6E1425, 1);
 
+    // Allow alpha change for helicopter rotor (#523)
+    MemSet((void*)0x6C444B, 0x90, 6);
+    MemSet((void*)0x6C4453, 0x90, 0x68);
```
TracerDS, line 1574: What are these mem addresses?

TracerDS (approved):

Client/multiplayer_sa/CMultiplayerSA.cpp
FileEX: CHeli::Render

lopezloo (approved):
Looks good to me. Thank you!

PR #3597, opened 2024-07-22
Add "flyingcomponents" special world property

FileEX:
Merge after #3572 

This PR adds a new special world property called ``flyingcomponents`` (default is true). This property determines whether the detached components such as doors, bumpers, etc., should respawn after a vehicle is recreated or streamed in. This PR solves a very frustrating and annoying issue where all missing components fall off again after using ``setElementModel`` or ``setVehicleVariant``:

image

Thanks to this property, various car garages and vehicle spawners will be able to look normal without flying components around


Dark-Dragon:
Again I'll have to point out that this to me seems like more of a bug-fix-scenario, rather than a special-world-property-scenario. What about the current behavior is desirable?

FileEX:
Probably nothing, but it's better to set it to the current default than for the PR to get labeled as 'backwards-incompatible' and end up stalled for several months, or possibly even years

TheNormalnij:
> Probably nothing, but it's better to set it to the current default than for the PR to get labeled as 'backwards-incompatible'

It won't. Current state is bugged

FileEX:
> > Probably nothing, but it's better to set it to the current default than for the PR to get labeled as 'backwards-incompatible'
> 
> It won't. Current state is bugged

Done

Fernando-A-Rocha (approved):
Fine to merge

Client/mods/deathmatch/logic/CClientVehicleManager.cpp
```diff
@@ -61,6 +61,7 @@ CClientVehicleManager::CClientVehicleManager(CClientManager* pManager)
     // Initialize members
     m_pManager = pManager;
     m_bCanRemoveFromList = true;
+    m_spawnFlyingComponentsDuringRecreate = true;
```
botder:
```suggestion
```

Client/mods/deathmatch/logic/CClientVehicleManager.h
```diff
@@ -70,9 +70,13 @@ class CClientVehicleManager
...
 
+    bool IsSpawnFlyingComponentEnabled() const noexcept { return m_spawnFlyingComponentsDuringRecreate; }
+    void SetSpawnFlyingComponentEnabled(bool isEnabled) noexcept { m_spawnFlyingComponentsDuringRecreate = isEnabled; }
+
 protected:
     CClientManager*               m_pManager;
     bool                          m_bCanRemoveFromList;
     CMappedArray<CClientVehicle*> m_List;
     CMappedArray<CClientVehicle*> m_StreamedIn;
+    bool                          m_spawnFlyingComponentsDuringRecreate;
```
botder:
```suggestion
    bool                          m_spawnFlyingComponentsDuringRecreate{true};
```

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -260,6 +260,7 @@ CGame::CGame() : m_FloodProtect(4, 30000, 30000)            // Max of 4 connecti
     m_WorldSpecialProps[WorldSpecialProperty::ROADSIGNSTEXT] = true;
     m_WorldSpecialProps[WorldSpecialProperty::TUNNELWEATHERBLEND] = true;
     m_WorldSpecialProps[WorldSpecialProperty::IGNOREFIRESTATE] = false;
+    m_WorldSpecialProps[WorldSpecialProperty::FLYINGCOMPONENTS] = false;
```
botder:
```suggestion
    m_WorldSpecialProps[WorldSpecialProperty::FLYINGCOMPONENTS] = true;
```

Shared/sdk/net/SyncStructures.h
```diff
@@ -2103,6 +2112,10 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
         if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_IgnoreFireState))
             bitStream.WriteBits(reinterpret_cast<const char*>(&data6), BITCOUNT6);
 
+        if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_FlyingComponents))
+            bitStream.WriteBits(reinterpret_cast<const char*>(&data7), BITCOUNT7);
+
+
```
botder:
```suggestion

```

Shared/sdk/net/SyncStructures.h
```diff
@@ -2170,6 +2188,7 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
         data4.extendedwatercannons = true;
         data5.tunnelweatherblend = true;
         data6.ignoreFireState = false;
+        data7.flyingcomponents = false;
```
botder:
```suggestion
        data7.flyingcomponents = true;
```

botder (changes requested):

Shared/sdk/net/SyncStructures.h
```diff
@@ -2130,6 +2139,8 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
         if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_IgnoreFireState))
             bitStream.WriteBits(reinterpret_cast<const char*>(&data6), BITCOUNT6);
 
+        if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_FlyingComponents))
+            bitStream.WriteBits(reinterpret_cast<const char*>(&data7), BITCOUNT7);
```
botder, line 2143:
```suggestion
        if (bitStream.Can(eBitStreamVersion::WorldSpecialProperty_FlyingComponents))
            bitStream.WriteBits(reinterpret_cast<const char*>(&data7), BITCOUNT7);

```

botder (approved):
I've fixed defaults in the sync structures where you used `false` instead of `true` for this new special world property.

PR #3599, opened 2024-07-22
Implement HTTP API

TracerDS:
This PR will allow you to fetch information from the server using the HTTP API.
I want it to work exactly like [discord api](https://discord.com/developers/docs/reference) (token, gateways, etc.)


Fernando-A-Rocha:
Beautiful initiative. I see you want to implement getting players & max player count for starters. That's good!

Fernando-A-Rocha:
What about HTTPS? And securing the api with an api key optionally?

TracerDS:
> Beautiful initiative. I see you want to implement getting players & max player count for starters. That's good!

I want to implement every function + maybe more 😉

TracerDS:
> What about HTTPS? And securing the api with an api key optionally?

HTTPS is out of control of this PR.
API depends on the server itself to maintain secure connection.
If the server has https handshake, so does api.

Hmm, api key might be generated via mta server. I'll keep that in mind when everything is done

Fernando-A-Rocha:
> > Beautiful initiative. I see you want to implement getting players & max player count for starters. That's good!
> 
> I want to implement every function + maybe more 😉

Will this API feature be enabled by default? Possibly through a MTAserver.conf setting.

TracerDS:
> > > Beautiful initiative. I see you want to implement getting players & max player count for starters. That's good!
> > 
> > 
> > I want to implement every function + maybe more 😉
> 
> Will this API feature be enabled by default? Possibly through a MTAserver.conf setting.

probably, yes

CrosRoad95:
Add at least some option to create custom api, mark old approach as outdated, change ehs lib to something more modern. Otherwise this feature will be underutilized

TracerDS:
> Add at least some option to create custom api

It will be added later

> mark old approach as outdated

What old approach? exports?

> change ehs lib to something more modern.

Other PR maybe

jlillis:
It is already possible to implement your own HTTP API [in Lua resources](https://wiki.multitheftauto.com/wiki/Resource_Web_Access).

TracerDS:
> It is already possible to implement your own HTTP API [in Lua resources](https://wiki.multitheftauto.com/wiki/Resource_Web_Access).

Yes but its very limited. With this approach its very flexible and it is possible to easily adjust it to your needs without needing any resources.
You also have one endpoint only. With this change you can decide which endpoint you want to use

Dutchman101:
How exactly does this feature use EHS?
The version of EHS (and EHS itself..) in MTA is ancient, from like 15 years ago, it's actually due to be replaced should someone be willing to.

Aren't there any implications of using an ancient implementation again, but for something else? It's already causing a ton of issues in its utilization for MTA's internal HTTP server.

TracerDS:
> How exactly does this feature use EHS? The version of EHS (and EHS itself..) in MTA is ancient, from like 15 years ago, it's actually due to be replaced should someone be willing to.
> 
> Aren't there any implications of using an ancient implementation again, but for something else? It's already causing a ton of issues in its utilization for MTA's internal HTTP server.

It only utilises EHS to register an endpoint. Everything else is handled by the mta code

jlillis:
> With this approach its very flexible and it is possible to easily adjust it to your needs without needing any resources.

The highest level of adjustment/customization options will come from it being implemented in Lua. There has been a general trend of servers re-implementing hardcoded features (the chatbox, F11 map, etc) in Lua for this reason.

> You also have one endpoint only. With this change you can decide which endpoint you want to use

Not ideal, but I think that is an improvement that can be made to the current scheme.

TracerDS:
> > With this approach its very flexible and it is possible to easily adjust it to your needs without needing any resources.
> 
> The highest level of adjustment/customization options will come from it being implemented in Lua. There has been a general trend of servers re-implementing hardcoded features (the chatbox, F11 map, etc) in Lua for this reason.
> 
> > You also have one endpoint only. With this change you can decide which endpoint you want to use
> 
> Not ideal, but I think that is an improvement that can be made to the current scheme.

there is no point in having a lua resource just for that.
Lua is very limited and, comparing to this solution, slower.
Just because it can be done in Lua doesnt mean it have to be done in Lua.
Hell, we can accept this change and disable the api endpoint but then you'd be stuck with forced endpoint.

CrosRoad95:
In my opinion, http server should be removed from server and instead separete application should be used to serve file and all http code should be removed from mta. Most hosting novadays offer separete "fast download" server anyway.
If you want to interface with your server from outside then use socket module, create new one that expose http api but this will not be part of mta.
I already had showned in past how easy it is to abuse http server

Server/mods/deathmatch/logic/CHTTPD.cpp
```diff
@@ -96,6 +98,24 @@ bool CHTTPD::StartHTTPD(const char* szIP, unsigned int port)
     return bResult;
 }
 
+void CHTTPD::StartAPI() noexcept
+{
+    m_apiEHS = new CAPI;
```
Pirulax, line 98: use `unique_ptr`

vendor/ehs/httpresponse.h
```diff
@@ -17,14 +17,23 @@
...
 /// different response codes and their corresponding phrases -- defined in EHS.cpp
-enum ResponseCode { HTTPRESPONSECODE_INVALID = 0,
-					HTTPRESPONSECODE_200_OK = 200,
-					HTTPRESPONSECODE_301_MOVEDPERMANENTLY = 301,
-					HTTPRESPONSECODE_302_FOUND = 302,
-					HTTPRESPONSECODE_401_UNAUTHORIZED = 401,
-					HTTPRESPONSECODE_403_FORBIDDEN = 403,
-					HTTPRESPONSECODE_404_NOTFOUND = 404,
-					HTTPRESPONSECODE_500_INTERNALSERVERERROR = 500 };
+enum ResponseCode {
```
Pirulax: Not sure if modifying EHS vendor code is a good idea.
Why not use these HTTP codes directly?

Server/mods/deathmatch/logic/CAPI.cpp
```diff
@@ -0,0 +1,81 @@
...
+    if (!GetNextPathFragment(uri))
+        return HTTPRESPONSECODE_404_NOTFOUND;
+
+    if (SharedUtil::StartsWith(uri, Endpoints::Server::PLAYERS))
+    {
+        return HandlePlayers(request, response);
+    }
+
+    std::string body("{\"error\": \"404: Not Found\"}");
+    response->SetBody(body.c_str(), body.size());
```
Pirulax, line 45: Can't a C-style string be passed in directly?
I don't see why we must use an `std::string`.

vendor/ehs/httpresponse.h
TracerDS: The return type is ResponseCode. I wanted to return that instead of ints instead

Server/mods/deathmatch/logic/CAPI.cpp
TracerDS: > Can't a C-style string be passed in directly? I don't see why we must use an `std::string`.

Only for simplicity

Server/mods/deathmatch/logic/CAPI.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include "CMainConfig.h"
+#include <SharedUtil.String.h>
+
+namespace Endpoints
+{
+    inline const std::string API_ENDPOINT = "/api";
+    namespace Server
+    {
+        inline const std::string PLAYERS = "/players";
+        inline const std::string MAX_PLAYERS = PLAYERS + "/max";
```
tederis, line 13: Given that these variables are within cpp file, what do you exactly expect from inline specifier here?

Server/mods/deathmatch/logic/CAPI.cpp
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+const bool GetNextPathFragment(std::string& path) noexcept
```
tederis, line 17: `const bool` as a returning value?

Server/mods/deathmatch/logic/CAPI.cpp
tederis: Make at least it `static`

Server/mods/deathmatch/logic/CAPI.cpp
TracerDS: oops, it was a lambda function and I forgot to remove `const`

Server/mods/deathmatch/logic/CAPI.cpp
TracerDS: I thought about moving them to header file and forgot to remove `inline`. It can be static, no?

tederis: Have you considered const static `std::string_view`?

Server/mods/deathmatch/logic/CAPI.cpp
tederis: Or even better std::string_view

Server/mods/deathmatch/logic/CAPI.cpp
TracerDS: Nope, will take that into consideration

Xenius97:
What if NodeJS were implemented on the server side? When integrated into MTA, it could be suitable for many things, including this.
Currently, we have developed our own preprocessor/backend based on a similar principle, with a NodeJS/Deno backend. In my opinion, Lua 5.1 isn’t very fast or efficient for such tasks, it's only adequate for basic operations and won’t handle more significant loads well.

TracerDS:
> What if NodeJS were implemented on the server side? When integrated into MTA, it could be suitable for many things, including this. Currently, we have developed our own preprocessor/backend based on a similar principle, with a NodeJS/Deno backend. In my opinion, Lua 5.1 isn’t very fast or efficient for such tasks, it's only adequate for basic operations and won’t handle more significant loads well.

there is no reason to include nodejs into already bloated code.
Connecting **to** NodeJS server is doable via [Sockets module](https://wiki.multitheftauto.com/wiki/Modules/Sockets).
But thats all if its about connecting somewhere.

Fernando-A-Rocha:
Im convinced that we have all we need with botder's newly added router features
https://wiki.multitheftauto.com/wiki/Resource_Web_Access#Router

TracerDS:

PR #3603, opened 2024-07-23
Add `getResourceFiles`

TracerDS:
`getResourceFiles` will return names of all files loaded by a resource


Fernando-A-Rocha:
Very nice. Can you add an argument "resource element" so you can get file list of another resource, which would work if it is running?

TracerDS:
> Very nice. Can you add an argument "resource element" so you can get file list of another resource, which would work if it is running?
> 

Already did ;)

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -1468,3 +1470,23 @@ bool CLuaResourceDefs::isResourceProtected(CResource* const resource)
...
+    if (!resource)
+        resource = &lua_getownerresource(luaVM);
+
+    const auto resourceFiles = (*resource)->GetResourceFiles();
+
+    std::vector<std::string> files;
+    files.reserve(resourceFiles.size());
+    for (const auto& file : resourceFiles)
+    {
+        if (file->GetType() != CResourceFile::eResourceType::RESOURCE_FILE_TYPE_CLIENT_FILE)
```
tederis: Are you sure that a server side function should deal with RESOURCE_FILE_TYPE_CLIENT_FILE files?

tederis: Maybe rename the function to `getLoadedClientFiles` for clarity?

TracerDS: > Maybe rename the function to `getLoadedClientFiles` for clarity?

Doesnt matter. `RESOURCE_FILE_TYPE_CLIENT_FILE` just means a valid `<file>` field. It will return every file thats not a map and not a script.

tederis: `<file>` field means exactly a client file

tederis: And please, do not mark a conversation resolved if the actual conversation is in process.

TracerDS: > `<file>` field means exactly a client file

thats the only file than can be. There is no division between server files and client files because server files is just not a concept.

tederis: > > `<file>` field means exactly a client file
> 
> thats the only file than can be. There is no division between server files and client files because server files is just not a concept.

Exactly for this reason the function should be called `getLoadedClientFiles`(or maybe `getResourceClientFileNames`). Simply because this function cannot deal with server-side files.

TracerDS: Renaming it to `getLoadedClientFiles` would indicate that there is a `getLoadedServerFiles` function. That doesnt make sense because server-side files dont exist.

tederis: > Renaming it to `getLoadedClientFiles` would indicate that there is a `getLoadedServerFiles` function. That doesnt make sense because server-side files dont exist.

The existence of `getLoadedClientFiles` function doesn't lead to the existence of function `getLoadedServerFiles`.

tederis: Anyway, this function must be renamed(maybe you will come up with a better name). Because the current name is confusing.

Fernando-A-Rocha: > Anyway, this function must be renamed(maybe you will come up with a better name). Because the current name is confusing.

I disagree. Instead of renaming it, add a string argument that lets you decide if it should return files, map files, script files, or all. Like "all"/"files"/"scripts"/"maps" according to the xml tags in meta.xml https://wiki.multitheftauto.com/wiki/Meta.xml

tederis: > > Anyway, this function must be renamed(maybe you will come up with a better name). Because the current name is confusing.
> 
> I disagree. Instead of renaming it, add a string argument that lets you decide if it should return files, map files, script files, or all. Like "all"/"files"/"scripts"/"maps" according to the xml tags in meta.xml https://wiki.multitheftauto.com/wiki/Meta.xml

This is what I also come up with. But you got ahead of me. Yes, this will work.

Fernando-A-Rocha: This argument can be optional and default to "all" where it returns script, map and regular file paths all together in a table.

TheNormalnij: `getResourceFiles` is explicit. The name of this function should start with 'getResource' because this function related to resource class. And the function can be shared with this name

Fernando-A-Rocha: I agree. That is a more suiting name for this function.

Fernando-A-Rocha (approved):
LGTM

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -29,6 +30,7 @@ void CLuaResourceDefs::LoadFunctions()
         {"getResourceState", GetResourceState},
         {"loadstring", LoadString},
         {"load", Load},
+        {"getResourceFiles", ArgumentParser<GetResourceFiles>},
```
TheNormalnij, line 33: Add field 'files' for the class, please

TracerDS: You mean oop?

TheNormalnij: yeah

Server/mods/deathmatch/logic/CResourceFile.h
```diff
@@ -37,12 +37,22 @@ class CResourceFile
...
 
+    enum class eResourceCategory
+    {
+        ALL = 0,
+        SCRIPTS = 1,
+        MAPS = 2,
+        CONFIGS = 3,
+        FILES = 4,
+    };
+
```
FileEX: What is the purpose of setting values ​​here when on the client side the exact same enum has no values ​​set?

TracerDS: Forgot to remove it

Fernando-A-Rocha (approved):

Client/mods/deathmatch/logic/CDownloadableResource.cpp
```diff
@@ -14,6 +14,21 @@
...
+        case eResourceType::RESOURCE_FILE_TYPE_MAP:
+            m_resourceCategoryType = eResourceCategory::MAPS;
+        case eResourceType::RESOURCE_FILE_TYPE_CONFIG:
+        case eResourceType::RESOURCE_FILE_TYPE_CLIENT_CONFIG:
+            m_resourceCategoryType = eResourceCategory::CONFIGS;
+        case eResourceType::RESOURCE_FILE_TYPE_CLIENT_FILE:
+        case eResourceType::RESOURCE_FILE_TYPE_HTML:
+            m_resourceCategoryType = eResourceCategory::FILES;
+    }
+
```
TheNormalnij, line 31: You can filter it in `CLuaResourceDefs::GetResourceFiles`. You don't need `GetResourceCategoryType`

TracerDS: Its better this way

TheNormalnij: Hmm. You can use original types.  I think `client-scripts` and `server-scripts` have some sense for developers. So you shouldn't map `eResourceType` into custom enum for one Lua function

TracerDS: I'd have to do string manipulation in both functions just to get values.
Its better that way

TheNormalnij: What do you mean?

TracerDS: ArgumentParser handles enum convertion for me. I'd have to accept certain string values and filter them. Its too much hassle.
Its better to create an enum. Everything will be taken care of for me.

TheNormalnij: What is wrong with `eResourceType` ?

TracerDS: too much noise

TheNormalnij: It looks like using existing enum is much better than creating a new one with mapping from `eResourceType` and adding  boilerplate code in client and server.

> too much noise

What do you mean?

TracerDS: > What do you mean?

Exactly what I said. Its easier to have 3 categories: scripts, maps, files instead of: client/server scripts, client/server files, (client/server?) maps, etc.

TheNormalnij: I see no problem here. It's much useful.

TheNormalnij:
Does it return shared scripts twice here? Is it expected behavior?
```lua
local list = getResourceFiles("all")
```

Fernando-A-Rocha:
File paths returned shouldn't be repeated imo

Fernando-A-Rocha:
To make this more advanced, the list returned could be a table of {filePath, fileType} where fileType is like "client-script" or "shared-script" or "map-file" idk

TracerDS:
> Does it return shared scripts twice here? Is it expected behavior?
> 
> ```lua
> local list = getResourceFiles("all")
> ```

Yeah, it returns twice. Its not expected behaviour. Will fix that
image

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -1476,24 +1476,28 @@ bool CLuaResourceDefs::isResourceProtected(CResource* const resource)
...
     for (const auto& file : resourceFiles)
     {
-        if (file->GetResourceCategoryType() == type.value()
-            || type.value() == CResourceFile::eResourceCategory::ALL)
-        {
-            files.push_back(file->GetName());
-        }
+        if (file->GetResourceCategoryType() != type.value() && type.value() != eResourceCategory::ALL)
+            continue;
+        files.insert(file->GetName());
```
TheNormalnij, line 1494: Previous varian was more readable. Why did you change it?

TracerDS: previous variant was basically the same.
This leaves room for more conditions to come.

TheNormalnij: There is no 'early continue' rule. You obfuscate the code with overusing 'continue' in loops. And you don't really need to use early return for small code blocks.
You did the code less readable and maintainable.
```c++
if (file->GetResourceCategoryType() == type.value() || type.value() == eResourceCategory::ALL)
    files.insert(file->GetName());
```
Is fine condition, that describes the function logic.

TracerDS: There is early-return which is basically the same but whatever.
I'd rather go with continue rather than indentation

TheNormalnij: Your `continue` creates indentation too. Please, keep it simple

> There is early-return which is basically the same but whatever.

Not really. Multiple 'continue' conditions can make a loop harder for reading.
There is no goal to use minimal indentation. The code should primarily describe the logic.
You're overusing 'early return'

TheNormalnij: Don't resolve conversations without changes silently

TracerDS: Im not going to change that. We didnt continue the discussion for some time, hence I resolved it.
Conditions may change. Better to overuse early return than to not use it at all/not use it often enough.
Result? Mta code with huge if blocks.

TheNormalnij: > We didnt continue the discussion for some time, hence I resolved it.

It looks like you tried to hide requested changes. Be civil. Maybe another mta member will points that I'm wrong or confirms my position.

> Conditions may change.

This is not how overused code looks like. It's better to keep it readable for future changes.

> Better to overuse early return than to not use it at all/not use it often enough.

You deal in absolutes. Overused principles are harmful for readability and complexity. Use them only when they improve your code. In this example you trade readability for principles.

TracerDS: > > We didnt continue the discussion for some time, hence I resolved it.
> 
> It looks like you tried to hide requested changes. Be civil. Maybe another mta member will points that I'm wrong or confirms my position.
> 
> > Conditions may change.
> 
> This is not how overused code looks like. It's better to keep it readable for future changes.
> 
> > Better to overuse early return than to not use it at all/not use it often enough.
> 
> You deal in absolutes. Overused principles are harmful for readability and complexity. Use them only when they improve your code. In this example you trade readability for principles.

If more contributors are opposed against this change, I will reconsider

Fernando-A-Rocha (approved):
Ready to merge imo

Server/mods/deathmatch/logic/CResource.h
```diff
@@ -256,8 +256,8 @@ class CResource : public EHS
     const std::string&          GetResourceDirectoryPath() const { return m_strResourceDirectoryPath; }
     const std::string&          GetResourceCacheDirectoryPath() const { return m_strResourceCachePath; }
 
-    std::list<CResourceFile*>& GetFiles() { return m_ResourceFiles; }
-    size_t                     GetFileCount() const noexcept { return m_ResourceFiles.size(); }
+    std::list<CResourceFile*>&       GetResourceFiles() noexcept { return m_ResourceFiles; }
```
tederis, line 259: Non-const getter is usually a bad practice. Is it really needed here?

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TheNormalnij: Again?

Server/mods/deathmatch/logic/CResource.h
TracerDS: I thought it may be modified somewhere. Better safe than sorry.
If it wont be modified anywhere then I can remove it

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TracerDS: > Again?

No one was interested in this convo for 2 weeks, give it a break.
No contributors were opposed against this so Im leaving it as it is.

tederis: I also see the early return overuse here.

Server/mods/deathmatch/logic/CResource.h
tederis: I don't think that this getter will be useful. Even if so, it can be recreated very easy.

Server/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
TheNormalnij: > No contributors were opposed against this so Im leaving it as it is.

And nobody confirm your position. Don't hide requested changes. Stale conversations should be dismissed by another  members.

TracerDS: > And nobody confirm your position

Im following the guidelines is'all.

TheNormalnij: you interpreter these guidelines very differently

Fernando-A-Rocha:
@TracerDS it would be nice to have a way to know if a file has download="false" attribute, so it can be downloaded using downloadFile later
Currently the only way to achieve this is by parsing the meta.xml file, which seems a bit too much

Fernando-A-Rocha:
> @TracerDS it would be nice to have a way to know if a file has download="false" attribute, so it can be downloaded using downloadFile later Currently the only way to achieve this is by parsing the meta.xml file, which seems a bit too much

Any ideas?

TracerDS:
> > @TracerDS it would be nice to have a way to know if a file has download="false" attribute, so it can be downloaded using downloadFile later Currently the only way to achieve this is by parsing the meta.xml file, which seems a bit too much
> 
> Any ideas?

maybe a parameter that checks download="false"?

Fernando-A-Rocha:
> > > @TracerDS it would be nice to have a way to know if a file has download="false" attribute, so it can be downloaded using downloadFile later Currently the only way to achieve this is by parsing the meta.xml file, which seems a bit too much
> > 
> > 
> > Any ideas?
> 
> maybe a parameter that checks download="false"?

Resource files, when they are parsed in the meta.xml, are already identified if they need to be downloaded later by the client.

Fernando-A-Rocha:
It's a matter of returning that download attribute value for each file. The array of file names returned by this new function should contain arrays where you have like
`{fileName="myFoder/myFile.dff", download=false}`

TracerDS:
> It's a matter of returning that download attribute value for each file. The array of file names returned by this new function should contain arrays where you have like
> `{fileName="myFoder/myFile.dff", download=false}`

I think it would be better to return just an array of files and handle the download attribute via a parameter. It would be better for the end user

Fernando-A-Rocha:
> > It's a matter of returning that download attribute value for each file. The array of file names returned by this new function should contain arrays where you have like
> > `{fileName="myFoder/myFile.dff", download=false}`
> 
> I think it would be better to return just an array of files and handle the download attribute via a parameter. It would be better for the end user

True, but what parameter exactly?

TracerDS:
> True, but what parameter exactly?

Something like isDownloaded or something like that. Or maybe it will receive a table of flags if more param were to come

TracerDS:

PR #3606, opened 2024-07-24
Allow longer LOD distances

### LINKED ISSUE #3601: Building LOD && no-collision bug
### Describe the bug

1. When you replace a model and place as a building, and re-join into game it will have no collisions (https://streamable.com/l9vcje)
----> Only solves when you create building after model was replaced.

2. Building LOD's will not work as it should, doesn't matter what lod distance u set it will not load as default GTA LOD's from a distance.
----> https://streamable.com/n50vhp
----> how it should load:
image
----> how objects load:
image
----> createBuilding:
image

### Steps to reproduce

createBuilding & custom model replace

### Version

Multi Theft Auto v1.6-release-22628

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #3601
LOD distance is limited up to 325 in `engineSetModelLODDistance`.
It's limited to create alpha fade effect in `setLowLODElement` for objects.
Buildings use the GTA LOD's implementation and request a larger model LOD distance.

This PR add `extendedLod` argument for `engineSetModelLODDistance`. So this PR won't break old resources.
```lua
bool engineSetModelLODDistance( number modelIdd, number fDistance [, bool extendedLod = faslse] )
```


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -1049,9 +1049,11 @@ int CLuaEngineDefs::EngineSetModelLODDistance(lua_State* luaVM)
     // bool engineSetModelLODDistance ( int/string modelID, float distance )
```
Pirulax: update the comment too

Pirulax (changes requested):
So if I understand correctly, this higher LOD distance will only works for buildings? (From `createBuilding`)? 
If so, there's no need for an extra argument, no?
Otherwise, I dont understand why there was a limitation there in the first place, if it can be ignored using a bool argument.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
TheNormalnij: thx

TheNormalnij:
> If so, there's no need for an extra argument, no?

This function is used for a long time with objects LOD's. I tried a LOD distance longer than 325 with big maps loaded by `createObject` and it caused object popping.
Technically, we could increase the limit and break old maps or we could add an argument to prevent destructive behavior.

If we are very close to the next MTA release, i'd prefer the first option

Pirulax:
But wait. Doesn't that "extendedLod" argument just remove the limit on the "distance" paramter?

TheNormalnij:
> But wait. Doesn't that "extendedLod" argument just remove the limit on the "distance" paramter?

yes

FileEX (approved):

gta191977649 (approved):

Einheit-101:
Did somebody test if this doesnt cause any problems if weird numbers are put in? This new argument removes the limit entirely, so strange things or even crashes could eventually happen.

TheNormalnij:
> Did somebody test if this doesnt cause any problems if weird numbers are put in? This new argument removes the limit entirely, so strange things or even crashes could eventually happen.

Big numbers can cause render issues for lod's, when this functions is used for objects.

PR #3609, opened 2024-07-26
Fix Hydraulics upgrage stops working when using setVehicleHandling

### LINKED ISSUE #1121: Hydraulics stops working when using setVehicleHandling with player inside vehicle
**Describe the bug**
When you set maxVelocity with setVehicleHandling function and you're inside vehicle, you can't use it. Hydraulics is still mounted but vehicle behave like it has no hydraulics, you can honk instead of lifting a car.

**To reproduce**
1. create vehicle with upgrade 1087
2. try to use hydraulics [you can]
3. run vehicle = getPedOccupiedVehicle(getPlayerFromName("MrDadosz")); setVehicleHandling(vehicle, "maxVelocity", 100)
4. try to use hydraulics [you can't]

**Version**
v1.5.7-release-20288

**Additional context**
You can fix it with script:
```lua
if getVehicleUpgradeOnSlot(vehicle, 9) == 1087 then
	removeVehicleUpgrade(vehicle, 1087)
	addVehicleUpgrade(vehicle, 1087)
end
```
But it shouldn't happen.

Proxy-99:
Fixes #1121

### DIFF
--- a/Client/game_sa/CVehicleSA.cpp
+++ b/Client/game_sa/CVehicleSA.cpp
@@ -1351,7 +1351,7 @@ void CVehicleSA::RecalculateHandling()
         }
     }
     // Hydraulics Flag fixing
-    if (pInt->m_upgrades[1] && pInt->m_upgrades[1] == 1087)
+    if (pInt->m_upgrades[0] && pInt->m_upgrades[0] == 1087)
     {
         // Flag not enabled?
         if (uiHandlingFlags | HANDLING_Hydraulics_Flag)


Client/game_sa/CVehicleSA.cpp
```diff
@@ -1351,7 +1351,7 @@ void CVehicleSA::RecalculateHandling()
         }
     }
     // Hydraulics Flag fixing
-    if (pInt->m_upgrades[1] && pInt->m_upgrades[1] == 1087)
+    if (pInt->m_upgrades[0] && pInt->m_upgrades[0] == 1087)
```
TheNormalnij, line 1354: Doesn't work.
The upgrade can be in any position of `m_upgrades` array.
You need to iterate whole array to check it.
NOS upgrade should be fixed too.

Proxy-99: Oh okay but why the code using array instead of std::map so we should check for the key and value

TheNormalnij: `m_upgrades` contains only 8 elements.  An array is faster than a map when the data count is small

Proxy-99: so that mean we need to check for all upgrades not only for nitro and Hydraulics?

AlexTMjugador: I'm not @TheNormalnij, but I suppose that means you are supposed to do a linear search over the `m_upgrades` array (i.e., iterate over its elements to find if any of them matches a nitro or hydraulics upgrade) to check if hydraulics and nitro upgrades are present.

FileEX:
Any update?

Proxy-99:
> Any update?

nope I still have issues with it
if I add the upgrade then remove it 
the flag still being applied, the pInt->m_upgrades not removing the upgrage from the array list it still holds the number even tho we removing the upgrade
so we will need to patch this first in order to loop correctly through the list

PR #3585, opened 2024-07-19
Add ShakeCamera

TracerDS:
Resolves #3584 

Usage:
```lua
bool shakeCamera(float power, [float x, float y, float z])
```
* power - intensity **and** time of the shake
* x - center x coordinate of the shake
* y - center y coordinate of the shake
* z - center z coordinate of the shake

PR #3616, opened 2024-07-29
Appendum #3585 - shakeCamera

### LINKED ISSUE #3608: Impossible to disable shakeCamera
### Describe the bug

shakeCamera will run infinite, doesn't matter what u do - respawn or die - it won't reset

https://streamable.com/uyknty

### Steps to reproduce

shakeCamera(100)

### Version

22634

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TracerDS:
Resolves #3608 

Using `shakeCamera(0 or less)`, `resetShakeCamera()` will reset the effect

### DIFF
--- a/Client/game_sa/CCameraSA.cpp
+++ b/Client/game_sa/CCameraSA.cpp
@@ -446,8 +446,13 @@ float CCameraSA::GetShakeForce()
 
 void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
 {
-    DWORD               dwFunc = FUNC_ShakeCam;
-    CCameraSAInterface* cameraInterface = GetInterface();
+    static CCameraSAInterface* cameraInterface = GetInterface();
+    if (radius == 0.0f) {
+        cameraInterface->m_fCamShakeForce = 0;
+        return;
+    }
+
+    static const DWORD dwFunc = FUNC_ShakeCam;
     _asm
     {
         mov  ecx, cameraInterface
--- a/Client/game_sa/CCameraSA.cpp
+++ b/Client/game_sa/CCameraSA.cpp
@@ -448,19 +448,13 @@ void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
 {
     static CCameraSAInterface* cameraInterface = GetInterface();
     if (radius <= 0.0f)
-    {
-        cameraInterface->m_fCamShakeForce = 0.0f;
-        return;
-    }
+        return ResetShakeCamera();
 
-    static const DWORD dwFunc = FUNC_ShakeCam;
-    _asm
-    {
-        mov  ecx, cameraInterface
-        push z
-        push y
-        push x
-        push radius
-        call dwFunc
-    }
+    using ShakeCamera_t = void(__thiscall*)(CCameraSAInterface*, float radius, float x, float y, float z);
+    ((ShakeCamera_t)FUNC_ShakeCam)(cameraInterface, radius, x, y, z);
+}
+
+void CCameraSA::ResetShakeCamera() noexcept
+{
+    GetInterface()->m_fCamShakeForce = 0.0f;
 }
--- a/Client/game_sa/CCameraSA.cpp
+++ b/Client/game_sa/CCameraSA.cpp
@@ -447,7 +447,8 @@ float CCameraSA::GetShakeForce()
 void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
 {
     static CCameraSAInterface* cameraInterface = GetInterface();
-    if (radius == 0.0f) {
+    if (radius <= 0.0f)
+    {
         cameraInterface->m_fCamShakeForce = 0;
         return;
     }
--- a/Client/game_sa/CCameraSA.cpp
+++ b/Client/game_sa/CCameraSA.cpp
@@ -449,7 +449,7 @@ void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
     static CCameraSAInterface* cameraInterface = GetInterface();
     if (radius <= 0.0f)
     {
-        cameraInterface->m_fCamShakeForce = 0;
+        cameraInterface->m_fCamShakeForce = 0.0f;
         return;
     }
 
--- a/Client/game_sa/CCameraSA.h
+++ b/Client/game_sa/CCameraSA.h
@@ -429,4 +429,5 @@ public:
     float     GetShakeForce();
 
     void ShakeCamera(float radius, float x, float y, float z) noexcept override;
+    void ResetShakeCamera() noexcept override;
 };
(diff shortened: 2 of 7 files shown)


Client/game_sa/CCameraSA.cpp
```diff
@@ -446,8 +446,14 @@ float CCameraSA::GetShakeForce()
 
 void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
 {
-    DWORD               dwFunc = FUNC_ShakeCam;
-    CCameraSAInterface* cameraInterface = GetInterface();
+    static CCameraSAInterface* cameraInterface = GetInterface();
+    if (radius <= 0.0f)
+    {
+        cameraInterface->m_fCamShakeForce = 0;
```
FileEX: 0.0f?

FileEX (approved):
If tested and works, then lgtm

Client/game_sa/CCameraSA.cpp
TracerDS: you mean `radius <= 0.0f` or `m_fCamShakeForce = 0`?

FileEX: To keep literals consistent
```cpp
cameraInterface->m_fCamShakeForce = 0.0f;
```

Client/game_sa/CCameraSA.cpp
```diff
@@ -446,8 +446,14 @@ float CCameraSA::GetShakeForce()
 
 void CCameraSA::ShakeCamera(float radius, float x, float y, float z) noexcept
 {
-    DWORD               dwFunc = FUNC_ShakeCam;
-    CCameraSAInterface* cameraInterface = GetInterface();
+    static CCameraSAInterface* cameraInterface = GetInterface();
```
TheNormalnij, line 449: You don't need `static` here

Client/game_sa/CCameraSA.cpp
```diff
@@ -446,8 +446,14 @@ float CCameraSA::GetShakeForce()
...
+    if (radius <= 0.0f)
+    {
+        cameraInterface->m_fCamShakeForce = 0.0f;
```
TheNormalnij:
```suggestion
        GetInterface()->m_fCamShakeForce = 0.0f;
```

AlexTMjugador:
Just my two cents for whoever reviews this: I've outlined a different approach to tackle this issue at https://github.com/multitheftauto/mtasa-blue/issues/3608#issuecomment-2255573894 and the following comments.

Client/game_sa/CCameraSA.cpp
TracerDS: it can be static so the variable wouldnt be allocated again

Client/game_sa/CCameraSA.cpp
TracerDS: I pass `cameraInterface` later in the inline assembly

TheNormalnij: ah, you can avoid it.
```c++
    CCameraSAInterface* cameraInterface = GetInterface();
    if (radius <= 0.0f)
    {
         // Remove shake effect
         cameraInterface->m_fCamShakeForce = 0.0f;
     }
    ((void(__thiscall*)(CCameraSAInterface*,  float radius, float x, float y, float z))0x50A9F0)(cameraInterface, radius, x, y, z);

TracerDS: This is less readable and conflicts with current SA definitions.
Its better to leave it as inline assembly for now

TheNormalnij: We are trying to avoid inline assembly for simple function calls.

- C++ is more readable and maintainable than assembly
- It's much easier to write c++ definition than assembly for developers
- It generates better code. Compiler cannot optimize functions with inline assembly
- You can avoid hacks with static variables

TheNormalnij: Technically, You should make `resetCameraShake` function to be consistent with current API.
+You should reset this effect when a player join a server

TracerDS: sure

Client/game_sa/CCameraSA.cpp
TheNormalnij: If this will not be static, the variable won't be allocated at all

TracerDS: it will be. It will live throughout the whole program.

> Block variables with static or thread(since C++11) storage duration are initialized the first time control passes through their declaration (unless their initialization is [zero-](https://en.cppreference.com/w/cpp/language/zero_initialization) or [constant-initialization](https://en.cppreference.com/w/cpp/language/constant_initialization), which can be performed before the block is first entered). On all further calls, the declaration is skipped. ([1](https://en.cppreference.com/w/cpp/language/storage_duration))

TheNormalnij: > If this will **not be static,**

This is what i mean. With out `static` it will be [esp+41C] or something

TracerDS: its still being allocated nonetheless. Static just makes one definition for all calls instead of one definition per call.

TheNormalnij: It won't be allocated, because compiler knows memory position for `internalInterface` at compile time

TracerDS: > It won't be allocated, because compiler knows memory position for `internalInterface` at compile time

but the allocation happens at runtime, not compile time. The definition is not evaluated at compile time.
The compiler still needs to allocate memory for the variable. It being static doesnt make much of a difference other than perseverance between calls

Merlin: There is no memory allocation happening. GetInterface() just returns a pointer to the interface. 
Adding static will check TLS to see if camInterface is already initialized every time the function is called.
The function should not be noexcept since it calls a game function.

TracerDS: It doesnt matter what GetInterface returns.
The compiler needs to allocated memory on the stack for the `camInterface` variable. No need to do that X amount of times if `&GetInterface()` doesnt change.

Merlin: The return value will be in eax and then moved into ecx to call the function. The compiler is unlikely to move this onto the stack.

TracerDS: It needs to be on the stack/somewhere in memory because the class member is modified from that exact variable.

TheNormalnij: It doesn't use an additional memory or stack. Maybe only one additional register will be used, but it's better than an additional branching from static variable

TracerDS: > it's better than an additional branching from static variable

what do you mean?

TheNormalnij: > Adding static will check TLS to see if camInterface is already initialized every time the function is called.

TracerDS: Not adding static will define it each time the function runs.
When static is added, a new label in assembly is created for that variable so it can be referenced without allocating it on the stack each time. It is still allocated in memory

In any case cant do anything now. Static will stay

TheNormalnij: > Not adding static will define it each time the function runs.

In this case a compiler is clever enough to replace variable allocation with direct memory access by offset. You didn't optimize here anything.

PR #3620, opened 2024-07-31
Read file from path - #3490 appendum

TracerDS:
Resolves #3490 and [97920d8](https://github.com/multitheftauto/mtasa-blue/commit/97920d83de400e0378f73ea93e87b27cf59dd758#commitcomment-143697266)

The issue was a lack of error handler for file operations causing this weird issue (appending x null bytes at the end)


Dutchman101:
Thanks, most of it was reviewed on the original PR after all, it can be tested in nightly now.
We don't have much time to make such a build, starting in a couple days we can't build for an unknown duration (due to project reasons).

PR #3621, opened 2024-07-31
Fix #552 Impossible to stop helicopter rotor if it has a driver

### LINKED ISSUE #552: Impossible to stop helicopter rotor if it has a driver
**Describe the bug**
Using setHelicopterRotorSpeed will stop it but then it will start spinning again.

**To reproduce**
Drive a helicopter.
setHelicopterRotorSpeed(localPlayer.vehicle, 0)

**Expected behaviour**
Well it'd be nice if the rotor would stay not spinning maybe if engine state was set to off, though I haven't checked to verify it doesn't.

**Screenshots**

**Version**
Unstated at mantis.

**Additional context**

From https://bugs.mtasa.com/view.php?id=6105

FileEX:
Fixed #552

This PR add two new functions
```
bool setVehicleRotorState(vehicle, bool state [, bool stopRotor = true ])
bool getVehicleRotorState(vehicle)
```
This function works in a similar way to ``setVehicleEngineState``.

By specifying state to false, the rotor will not be processed, i.e. it will not change its speed and the helicopter/plane cannot be controlled.

The ``stopRotor`` parameter specifies whether the rotor is to be stopped (SetHeliRotorSpeed ​​to 0). If we specify false, the rotor will not change its speed, but will not be stopped, it will rotate at the same speed all the time.

Of course, instead of adding new functions, we could use ``setVehicleEngineState``, but for backward compatibility reasons it is better not to do this and simply add these two new functions to avoid possible problems with older scripts.

This function also allows to achieve the effect that the rotor rotates even without a driver.


Client/game_sa/CVehicleSA.cpp
```diff
@@ -51,6 +52,40 @@ void _declspec(naked) HOOK_Vehicle_PreRender(void)
...
+bool CanProcessFlyingCarStuff(CHeliSAInterface* heliInterface)
+{
+    SClientEntity<CVehicleSA>* vehicle = pGame->GetPools()->GetVehicle((DWORD*)heliInterface);
+    if (!vehicle || !vehicle->pEntity)
+        return true;
+
+    return vehicle->pEntity->GetHeliRotorState();
+}
+
+void _declspec(naked) HOOK_CHeli_ProcessFlyingCarStuff()
```
TracerDS: Can you make the hook not naked?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -489,6 +524,26 @@ void CVehicleSA::SetTrainSpeed(float fSpeed)
     pInterface->m_fTrainSpeed = fSpeed;
 }
 
+float CVehicleSA::GetHeliRotorSpeed()
+{
+    auto* heliInterface = static_cast<CHeliSAInterface*>(GetInterface());
+    return heliInterface->m_wheelSpeed[1];
```
TracerDS: It can be an one liner

Client/game_sa/CVehicleSA.cpp
```diff
@@ -489,6 +524,26 @@ void CVehicleSA::SetTrainSpeed(float fSpeed)
...
+}
+
+void CVehicleSA::SetHeliRotorSpeed(float speed)
+{
+    auto* heliInterface = static_cast<CHeliSAInterface*>(GetInterface());
+    heliInterface->m_wheelSpeed[1] = speed;
```
TracerDS: It can be an one liner

Client/game_sa/CVehicleSA.h
```diff
@@ -540,7 +546,8 @@ class CVehicleSA : public virtual CVehicle, public virtual CPhysicalSA
     bool           GetTakeLessDamage() { return GetVehicleInterface()->m_nVehicleFlags.bTakeLessDamage; };
     bool           GetTyresDontBurst() { return GetVehicleInterface()->m_nVehicleFlags.bTyresDontBurst; };
     unsigned short GetAdjustablePropertyValue() { return *reinterpret_cast<unsigned short*>(reinterpret_cast<unsigned long>(m_pInterface) + 2156); };
-    float          GetHeliRotorSpeed() { return *reinterpret_cast<float*>(reinterpret_cast<unsigned int>(m_pInterface) + 2124); };
+    float          GetHeliRotorSpeed();
```
TracerDS: Const

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4273,3 +4278,17 @@ std::variant<bool, std::array<CVector, 4>> CLuaVehicleDefs::OOP_GetVehicleEntryP
 
     return entryPoints;
 }
+
+bool CLuaVehicleDefs::SetHeliRotorState(CClientVehicle* vehicle, bool state, std::optional<bool> stopRotor)
```
TracerDS: SetHeliRotorState is noexcept, so make lua function noexcept too

Client/game_sa/CVehicleSA.cpp
```diff
@@ -489,6 +524,26 @@ void CVehicleSA::SetTrainSpeed(float fSpeed)
     pInterface->m_fTrainSpeed = fSpeed;
 }
 
+float CVehicleSA::GetHeliRotorSpeed()
+{
+    auto* heliInterface = static_cast<CHeliSAInterface*>(GetInterface());
+    return heliInterface->m_wheelSpeed[1];
+}
+
+void CVehicleSA::SetHeliRotorSpeed(float speed)
```
TracerDS, line 556: Noexcept maybe?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -489,6 +524,26 @@ void CVehicleSA::SetTrainSpeed(float fSpeed)
     pInterface->m_fTrainSpeed = fSpeed;
 }
 
+float CVehicleSA::GetHeliRotorSpeed()
```
TracerDS: Const noexcept

TracerDS (changes requested):

Client/game_sa/CVehicleSA.cpp
FileEX: This would result in poorer readability of the code, because then I would have to mix inline-assembly with C++ code. Naked hook is fine

Client/game_sa/CVehicleSA.cpp
FileEX: Theoretically, there is little chance of an exception if an access violation occurs for some reason. I think it's better to leave it as is in this case

Client/game_sa/CVehicleSA.cpp
TracerDS: How so? Isnt a naked function just a normal function but without stack set up?

Client/game_sa/CVehicleSA.cpp
TracerDS: Wouldnt it crash either way in that case then?

Also if an AVE occurs then something is wrong with mta's internals and the game should exit immediately

Client/game_sa/CVehicleSA.cpp
FileEX: It's best to avoid mixing inline assembly with C++ code whenever possible, and in this case, it is possible. A naked hook is the right solution here. If you look at the hooks in MTA, you'll see that almost every one of them is naked because it's the best approach for hooks. It's also safer than mixing assembly code with C++, which can cause issues.

For example, there was an issue with glass explosions where a switch instruction caused a crash because it used the same register as the game at the same time. With naked hooks, we can precisely control register and stack operations, minimizing the risk of such problems

Client/game_sa/CVehicleSA.cpp
FileEX: Adding noexcept will cause the game to terminate immediately without generating a standard error / crash window, which can make diagnosing the problem more difficult. In this case, if for some reason the value cannot be changed and an AVE occurs, having no noexcept might produce an error that helps identify the cause of the issue. 

Additionally, functions that interact with game memory should generally not be marked as noexcept. Memory can become invalid or corrupted for various reasons, and the chance of an exception always exists. Therefore, it's better to leave this function without noexcept to ensure that potential issues are properly reported and can be diagnosed

Client/game_sa/CVehicleSA.cpp
TracerDS: its not about mixing C++ with assembly but not using `naked` functions as they are deprecated in newer versions of C++. If we ever want to move on, we'd need to drop naked functions entirely. MTA code runs in C++14/17 version.
Naked functions are a big factor for this.

> [[1](https://learn.microsoft.com/en-us/cpp/cpp/naked-cpp?view=msvc-170)] For functions declared with the naked attribute, the compiler generates code without prolog and epilog code. You can use this feature to write your own prolog/epilog code sequences using inline assembler code

> [[2](https://github.com/multitheftauto/mtasa-blue/pull/3616#discussion_r1697460495)] We are trying to avoid inline assembly for simple function calls.
> - C++ is more readable and maintainable than assembly
> - It's much easier to write c++ definition than assembly for developers
> - It generates better code. Compiler cannot optimize functions with inline assembly
> - You can avoid hacks with static variables

FileEX: Naked functions may be less preferred in modern C++ and their use might be considered less idiomatic. However, in the case of hooks and low-level operations, naked functions can be useful because they allow for direct manipulation of assembly code, which is not always possible with pure C++.

TracerDS: You can still access everything from a non-naked function. You just have stack set up, thats all.

FileEX: Indeed, in a function that is not declared as naked, i can still access various elements of the program, but it requires the compiler to generate additional code for stack setup (prolog and epilog). This can limit our control over precise register management and code manipulation.

In the case of hooks and low-level manipulation, having full control over the machine code is crucial. Naked functions allow for direct manipulation, which is harder to achieve in functions where the prolog and epilog are automatically generated by the compiler. Therefore, using naked functions is often more appropriate in such situations. In this particular case, it is unlikely that i will make the change to a non-naked function.

TheNormalnij: > We are trying to avoid inline assembly for simple function calls.

I mean only situations, when you need only a function call from c++ code.

> not using naked functions as they are deprecated in newer versions of C++. 

They aren't. Where did you find it?
Naked functions aren't 'real' functions. Naked function is a pure assembly without prologue, epilogue and **_return_**.
Function calling conversations aren't applied to naked functions, so you have full control there

> MTA code runs in C++14/17 version. Naked functions are a big factor for this.

Only 'Game SA' and 'Multiplayer SA' projects. It's a bug (?) in MSVC

TracerDS: > They aren't. Where did you find it? Naked functions aren't 'real' functions. Naked function is a pure assembly without prologue, epilogue and **_return_**. Function calling conversations aren't applied to naked functions, so you have full control there

Naked functions are not supported in latest versions of C++ afaik. It throws error when you try to compile naked function in C++17 or c++20

TheNormalnij: from premake5.lua
```
	-- HACK(Jusonex): Temp fix for ebp not being set in naked functions
	-- VS2019 changed the evaluation order as required by the C++17 standard
	-- which broke all functions marked with _declspec(naked) that call C++ code
	-- Falling back to the old, C++14 implementing fixes this
```

TracerDS: Yeah this

salwador:
Will it work with planes, or only helicopters?

FileEX:
> Will it work with planes, or only helicopters?

This PR only applies to helicopters. Does the same bug apply to planes?

salwador:
> Does the same bug apply to planes?

Yep, same. At the moment, rotor can be stopped only in preRender. 

Im ask because PR [#2548](https://github.com/multitheftauto/mtasa-blue/pull/2548) add combined method for set/get heli/plane rotor speed.
It would be good if the rotor-state methods was universal.

TheNormalnij:
> Of course, instead of adding new functions, we could use `setVehicleEngineState`, but for backward compatibility reasons it is better not to do this and simply add these two new functions to avoid possible problems with older scripts.

What may be broken?

Client/game_sa/CVehicleSA.cpp
```diff
@@ -51,6 +52,40 @@ void _declspec(naked) HOOK_Vehicle_PreRender(void)
...
+{
+    _asm
+    {
+        mov esi, ecx
+        mov al, [esi+36h]
+
+        pushad
+        push ecx
+        call CanProcessFlyingCarStuff
+        add esp, 4
```
TheNormalnij: You can use __cdecl for CanProcessFlyingCarStuff to avoid manipulations with `esp`

Client/game_sa/CVehicleSA.h
```diff
@@ -94,6 +94,11 @@ struct RwTexture;
 #define FUNC_CAutomobile_OnVehiclePreRender 0x6ABCFD
 #define FUNC_CVehicle_DoSunGlare            0x6DD6F0
 
+#define FUNC_CHeli_ProcessFlyingCarStuff 0x6C4E7D
+
+static constexpr DWORD CONTINUE_CHeli_ProcessFlyingCarStuff = 0x6C4E82;
+static constexpr DWORD RETURN_CHeli_ProcessFlyingCarStuff = 0x6C5404;
+
```
TheNormalnij: Move it before `HOOK_CHeli_ProcessFlyingCarStuff`. These variables are useless without the hook

Client/game_sa/CVehicleSA.cpp
TheNormalnij: Or you can use __fastcall, because your argument is in `ecx`

FileEX:
> > Of course, instead of adding new functions, we could use `setVehicleEngineState`, but for backward compatibility reasons it is better not to do this and simply add these two new functions to avoid possible problems with older scripts.
> 
> What may be broken?

As far as I know, the MTA team strongly attaches importance to backward compatibility.

In this case, various scripts using setVehicleEngineState could cause the helicopter's rotor to stop, and while this is perfectly desirable behavior for us, it may be undesirable for others. So I personally think it's safer and better to just add these simple functions

FileEX:
> > Does the same bug apply to planes?
> 
> Yep, same. At the moment, rotor can be stopped only in preRender.
> 
> Im ask because PR [#2548](https://github.com/multitheftauto/mtasa-blue/pull/2548) add combined method for set/get heli/plane rotor speed. It would be good if the rotor-state methods was universal.

I took this into account and introduced ``setVehicleRotorState`` & ``getVehicleRotorState`` which is for both helicopters and planes

Fernando-A-Rocha (approved):

botder (dismissed):

botder (approved):

PR #3627, opened 2024-08-02
Add removeGameWorld and restoreGameWorld

### LINKED ISSUE #3607: removeAllGameBuildings doesn't remove the whole map
### Describe the bug

The issue i'm having is that `removeAllGameBuildings `causes some world models such as palm trees ID 710 to not get deleted or the hanger door, or some gates, and even using things such as `removeWorldObject ` in a map file or `removeWorldModel` doesn't affect these objects left on the map and doesn't delete them, and if you done so before using `removeAllGameBuildings ` they are restored again and not able to be deleted even after you disconnect, the only way to delete them again is exiting the game and connecting again.

1-- This is after deleting the palm trees using removeWorldObject map file.

Capture

2-- This is after using `removeAllGameBuildings() ` for the first time. after the first picture directly

Capture

3- Now the palm trees are permanently not deleteable even after you reconnect, and the only way to delete the palm trees again is after you exit GTA SA game and start it again.

Capture

4- To note this is not just for Palm trees but other world objects too.

### Steps to reproduce

1- use `removeAllGameBuildings()` to delete the whole map
2- this step you can do either after or before, make a map file and this line 
`<removeWorldObject id="removeWorldObject (vgs_palm01) (1)" radius="2000" interior="-1" model="710" lodModel="0" posX="1402.0859" posY="785.11719" posZ="25.54688" rotX="0" rotY="0" rotZ="0"></removeWorldObject>` 
those trees exist in the LV church beside the bridge to LS.
and notice how it doesn't get deleted even after you reconnect to the server.

### Version

_No response_

### Additional context
_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #3607
`removeAllGameBuildings` cleans only buildings pool and leaves rare dynamic elements in the game world.
`removeGameWorld` fixes that issue by cleaning the dummies pool too.

```lua
-- Removes all buildings and dummies from game world and frees both pools
nil removeGameWorld()

-- Restores all buildings and dummies
nil restoreGameWorld()
```
`removeAllGameBuildings` and `restoreAllGameBuildings`  will be deprecated and removed in future releases


Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
-    if (m_pLodBackup)
+    if (m_pOriginalElementsBackup)
         return;
 
-    m_pLodBackup = std::make_unique<std::array<CEntitySAInterface*, MAX_DUMMIES>>();
+    m_pOriginalElementsBackup = std::make_unique<pool_backup_t>();
 
-    for (int i = 0; i < MAX_DUMMIES; i++)
+    auto pDummyPool = (*m_ppDummyPoolInterface);
+    for (size_t i = 0; i < MAX_DUMMIES_DEFAULT; i++)
```
TracerDS: Use `std::size_t` instead

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
+
+            pDummyPool->Release(i);
+
+            (*m_pOriginalElementsBackup)[i].first = true;
+            (*m_pOriginalElementsBackup)[i].second = *building;
+        }
+        else
+        {
+            (*m_pOriginalElementsBackup)[i].first = false;
+        }
```
TracerDS, line 51: Early return (continue)

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
+void CDummyPoolSA::RestoreBackup()
 {
-    if (!m_pLodBackup)
+    if (!m_pOriginalElementsBackup)
         return;
 
-    for (int i = 0; i < MAX_DUMMIES; i++)
+    auto& originalData = *m_pOriginalElementsBackup;
+    auto  pDummyPool = (*m_ppDummyPoolInterface);
+    for (size_t i = 0; i < MAX_DUMMIES_DEFAULT; i++)
```
TracerDS: `std::size_t`

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
-        CEntitySAInterface* object = (*m_ppDummyPoolInterface)->GetObject(i);
-        object->m_pLod = (*m_pLodBackup)[i];
+        if (originalData[i].first)
+        {
+            pDummyPool->AllocateAt(i);
+            auto pDummy = pDummyPool->GetObject(i);
+            *pDummy = originalData[i].second;
+
+            pGame->GetWorld()->Add(pDummy, CDummyPool_Constructor);
+        }
```
TracerDS, line 71: Early continue

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
     }
 
-    m_pLodBackup.release();
+    m_pOriginalElementsBackup = nullptr;
 }
 
 void CDummyPoolSA::UpdateBuildingLods(void* oldPool, void* newPool)
 {
-    const uint32_t offset = (uint32_t)newPool - (uint32_t)oldPool;
+    const std::uint32_t offset = (std::uint32_t)newPool - (std::uint32_t)oldPool;
```
TracerDS: Can you use static_cast here instead?

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
-            if ((*m_pLodBackup)[i] != nullptr)
+            if ((*m_pOriginalElementsBackup)[i].first)
             {
-                (*m_pLodBackup)[i] = (CEntitySAInterface*)((uint32_t)(*m_pLodBackup)[i] + offset);
+                CEntitySAInterface* object = &(*m_pOriginalElementsBackup)[i].second;
+                if (object->m_pLod)
+                {
+                    object->m_pLod = (CEntitySAInterface*)((std::uint32_t)(object->m_pLod) + offset);
+                }
             }
```
TracerDS: Early continue

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
         }
     }
     else
     {
-        for (int i = 0; i < MAX_DUMMIES; i++)
+        for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
```
TracerDS: The loop is the same. Move the if statement to the loop instead for better readability

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,63 +13,93 @@
...
         {
             CEntitySAInterface* object = (*m_ppDummyPoolInterface)->GetObject(i);
             if (object->m_pLod)
             {
-                object->m_pLod = (CEntitySAInterface*)((uint32_t)object->m_pLod + offset);
+                object->m_pLod = (CEntitySAInterface*)((std::uint32_t)object->m_pLod + offset);
```
TracerDS: Early continue

Client/game_sa/CGameSA.cpp
```diff
@@ -1009,8 +1009,14 @@ void CGameSA::RemoveAllBuildings(bool clearBuildingRemoval)
 {
     m_pIplStore->SetDynamicIplStreamingEnabled(false);
 
-    m_pPools->GetDummyPool().RemoveAllBuildingLods();
-    m_pPools->GetBuildingsPool().RemoveAllBuildings();
+    m_pCoverManager->RemoveAllCovers();
+    m_pPlantManager->RemoveAllPlants();
+
+    // Remove all shadows in CStencilShadowObjects::dtorAll
+    ((void* (*)())0x711390)();
```
TracerDS, line 1070: Please document this memory address

TracerDS (changes requested):

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: This is one line. It's totally unnecessary in this case

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: They are not the same

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: auto

Client/game_sa/CDummyPoolSA.cpp
TracerDS: > They are not the same

```cpp
for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
{
    if (m_pOriginalElementsBackup && (*m_pOriginalElementsBackup)[i].first)
    {

    }
    else
    {
    }
}
```

TheNormalnij: In your example `(*m_pOriginalElementsBackup)[i].first == false` causes using second loop.

TracerDS: Thats your code but moved the `if` statement to the loop

TheNormalnij: No. Your code is bugged. Negative `(*m_pOriginalElementsBackup)[i].first` causes using `else` branch.
Read it again

TracerDS: > No. Your code is bugged. Negative `(*m_pOriginalElementsBackup)[i].first` causes using `else` branch. Read it again

thats literally your code.
```cpp
if (m_pOriginalElementsBackup)
{
    for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
    {
        if ((*m_pOriginalElementsBackup)[i].first)
        {
            CEntitySAInterface* object = &(*m_pOriginalElementsBackup)[i].second;
            if (object->m_pLod)
            {
                object->m_pLod = (CEntitySAInterface*)((std::uint32_t)(object->m_pLod) + offset);
            }
        }
    }
}
else
{
    for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
    {
        CEntitySAInterface* object = (*m_ppDummyPoolInterface)->GetObject(i);
        if (object->m_pLod)
        {
            object->m_pLod = (CEntitySAInterface*)((std::uint32_t)object->m_pLod + offset);
        }
    }
}
```
When you move the if statement to the loop, you will get the code above (and below):
```cpp
for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
{
    if(m_pOriginalElementsBackup && (*m_pOriginalElementsBackup)[i].first)
    {

    }
    else
    {

    }
}
```
Negative `.first` has no effect on the `if` statement

TracerDS: ~~You could even shorten it to this:~~
```cpp
for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
{
    CEntitySAInterface* object;

    if(m_pOriginalElementsBackup && (*m_pOriginalElementsBackup)[i].first)
    {
        object = &(*m_pOriginalElementsBackup)[i].second;
    }
    else
    {
        object = (*m_ppDummyPoolInterface)->GetObject(i);
    }
    if (object->m_pLod)
    {
        object->m_pLod = (CEntitySAInterface*)((std::uint32_t)object->m_pLod + offset);
    }
}
```

nvm, not really, but you get the idea

TheNormalnij: it has, when `m_pOriginalElementsBackup` is not null and `(*m_pOriginalElementsBackup)[i].first` is false.
By your logic the loop will use `(*m_ppDummyPoolInterface)->GetObject(i);`

TracerDS: again, you can detect that and have 1 loop instead of 2 with nearly the same content.
Just reformat the above code to continue when `m_pOriginalElementsBackup && !(*m_pOriginalElementsBackup)[i].first`

TheNormalnij: > again, you can detect that and have 1 loop instead of 2 with nearly the same content.

Please, be totally explicit. I can't read your mind. You twice suggested me totally wrong code and now you're writing about `else if` branch.

TracerDS: > Please, be totally explicit. I can't read your mind. You twice suggested me totally wrong code and now you're writing about `else if` branch.

It was always about `else`. You have `else` in your code.
Just put it in the loop itself.
```cpp
for (int i = 0; i < MAX_DUMMIES_DEFAULT; i++)
{
    CEntitySAInterface* object;

    if(m_pOriginalElementsBackup)
    {
        if(!(*m_pOriginalElementsBackup)[i].first)
            continue;
        object = &(*m_pOriginalElementsBackup)[i].second;
    }
    else
    {
        object = (*m_ppDummyPoolInterface)->GetObject(i);
    }

    if (!object->m_pLod)
        continue;

    object->m_pLod = static_cast<CEntitySAInterface*>((std::uint32_t)object->m_pLod + offset);
}
```

TheNormalnij: > It was always about else. You have else in your code.

You didn't provide much context about what you want to see.
Horewer, it's spaghetti. I'd prefer to extract method instead.

```c++
if (m_pOriginalElementsBackup)
    UpdateLodPointersInBackup(offset);
else
    UpdateLodPointersInPool(offset);
```

TracerDS: > You didn't provide much context about what do you want to see.

🤨

> Horewer, it's spaghetti. I'd prefer to extract method instead.

that could work too but in this case it would be better for simple if statement imo

Client/game_sa/CGameSA.cpp
TheNormalnij: Documented in comment

TracerDS: Nvm it, github acting crazy.

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: No, your example is unnecessary complex. Two additional continue in one loop in different levels are confusing

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: outdated

Client/game_sa/CDummyPoolSA.cpp
TracerDS: Having two near identical for loops is confusing.
This example reduces the code. Two `continue` isnt confusing.
This example is straightforward.
We have much more confusing code in our codebase thanks to these huge structures

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: we have no 'early continue' rules

TracerDS: early continue is the same as early return

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,64 +13,96 @@
...
+        if ((*m_pOriginalElementsBackup)[i].first)
         {
-            if ((*m_pLodBackup)[i] != nullptr)
-            {
-                (*m_pLodBackup)[i] = (CEntitySAInterface*)((uint32_t)(*m_pLodBackup)[i] + offset);
-            }
+            CEntitySAInterface* object = &(*m_pOriginalElementsBackup)[i].second;
+            CEntitySAInterface* lod = object->GetLod();
+            if (lod)
+                object->SetLod((CEntitySAInterface*)((std::uint32_t)lod + offset));
```
TracerDS: Use range loop instead, also use early continue (early return)

Client/game_sa/CDummyPoolSA.cpp
```diff
@@ -13,64 +13,96 @@
...
-            CEntitySAInterface* object = (*m_ppDummyPoolInterface)->GetObject(i);
-            if (object->m_pLod)
-            {
-                object->m_pLod = (CEntitySAInterface*)((uint32_t)object->m_pLod + offset);
-            }
-        }
+        CEntitySAInterface* object = (*m_ppDummyPoolInterface)->GetObject(i);
+        CEntitySAInterface* lod = object->GetLod();
+        if (lod)
+            object->SetLod((CEntitySAInterface*)((std::uint32_t)lod + offset));
```
TracerDS, line 106: same as above

Client/game_sa/CPoolSAInterface.h
```diff
@@ -114,8 +114,9 @@ class CPoolSAInterface
 
     void Delete(uint index) { Release(index); }
 
-    bool IsEmpty(std::int32_t objectIndex) { return m_byteMap[objectIndex].bEmpty; }
-    bool IsContains(uint index)
+    std::int32_t Size() const noexcept { return m_nSize; };
+    bool IsEmpty(std::int32_t objectIndex) const { return m_byteMap[objectIndex].bEmpty; }
```
TracerDS, line 118: `std::uint32_t`?
also assuming `m_byteMap[objectIndex]` is valid (assert could be used here), `noexcept` can be used here

Client/game_sa/CPoolSAInterface.h
```diff
@@ -114,8 +114,9 @@ class CPoolSAInterface
...
+    bool IsContains(uint index) const
```
TracerDS, line 119: if above then `noexcept`

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: No, 'return' finishes a function. 'continue' is syntax sugar for 'goto' to next loop iteration

TracerDS (changes requested):

Client/game_sa/CDummyPoolSA.cpp
TracerDS: > No, 'return' finishes a function. 'continue' is syntax sugar for 'goto' to next loop iteration

Yes, but the idea stays the same, hence one being the same as the other.

Client/game_sa/CPoolSAInterface.h
TheNormalnij: Can cause exception. It's better to avoid it here for additional debug information

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: No, they have different readability. Avoid 'continue' where it's unnecessary

TracerDS: > No, they have different readability. Avoid 'continue' where it's unnecessary

absolutely not.
`return` skips the content of the block in a funciton.
`continue` skips the content of the block in a codeblock in loop

Do not avoid continue. Use continue where you can. Unless on really short one liners.

Client/game_sa/CPoolSAInterface.h
TheNormalnij: Can cause exception. It's better to avoid it here for additional debug information

Client/game_sa/CDummyPoolSA.cpp
TheNormalnij: No, you're overusing it. I see no reason to put `continue` in every loop where it doesn't improve readability.

Client/game_sa/CPoolSAInterface.h
TheNormalnij: i can't guarantee std::uint32_t. It uses documented type

TracerDS: Can the index ever be negative?

Client/game_sa/CDummyPoolSA.cpp
TracerDS: It absolutely does improve readability. Especially in this case

Client/game_sa/CDummyPoolSA.cpp
Dutchman101: Addressed in 71d99cb

Client/game_sa/CDummyPoolSA.cpp
Dutchman101: TheNormalnij — Today at 2:58 AM
It (range loop) can't be used there

Fernando-A-Rocha:
@TheNormalnij will u please update the wiki? Let me know if u need advice

TheNormalnij:
> @TheNormalnij will u please update the wiki? Let me know if u need advice

Yep

Fernando-A-Rocha (approved):
Ive edited the wiki pages of the removeAllGameBuildings and restoreAllGameBuildings functions about the deprecation

Fernando-A-Rocha:
What about 
https://wiki.multitheftauto.com/wiki/RestoreAllWorldModels ? All world models can already be removed and restored. Why new functions?

Fernando-A-Rocha:
https://wiki.multitheftauto.com/wiki/RemoveWorldModel

Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
```diff
@@ -63,12 +63,14 @@ CClientBuilding* CLuaBuildingDefs::CreateBuilding(lua_State* const luaVM, std::u
...
-    m_pBuildingManager->RemoveAllGameBuildings();
+    CLuaWorldDefs::RemoveGameWorld();
 }
 
+// Deprecated
 void CLuaBuildingDefs::RestoreGameBuildings()
 {
-    m_pBuildingManager->RestoreAllGameBuildings();
+    CLuaWorldDefs::RestoreGameWorld();
 }
```
botder, line 76: Please add the deprecated functions to `clientDeprecatedList` in `CResourceChecker.Data.h`

Client/sdk/game/Common.h
```diff
@@ -27,7 +27,6 @@
 #define MAX_PEDS                            ( MAX_PEDS_MTA + 30 )                   // 140
 #define MAX_OBJECTS                         ( MAX_OBJECTS_MTA + 200 )               // 1200
 #define MAX_BUILDINGS                       13000
-#define MAX_DUMMIES                         2500
```
botder: Please restore this fella.

botder (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
TheNormalnij: yeah

Client/sdk/game/Common.h
TheNormalnij: done too

Fernando-A-Rocha (approved):

botder (approved):

PR #3632, opened 2024-08-05
Add async dbConnect

### LINKED ISSUE #410: Get dbConnect handle in other ways
dbConnect can take a long time when connecting to a mysql server.

Return database handle in callback OR maybe integrate magic coroutines.

I am very keen on us exploring the use of coroutines in MTA more, to prevent callbacks.

From https://bugs.mtasa.com/view.php?id=9238

TracerDS:
Resolves #410


Shared/sdk/SharedUtil.Misc.h
```diff
@@ -808,6 +866,25 @@ namespace SharedUtil
...
     }
 
+    template <class T, class U>
+    void GetOption(const std::string& text, const std::string& key, const char* separator, std::set<U>& outValues)
+    {
+        SString     numbers;
+        {
+            std::string temp;
+            GetOption<T>(text, key, temp);
+            numbers = temp;
```
tederis: std::move

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -808,6 +866,25 @@ namespace SharedUtil
                 MapInsert(outValues, static_cast<U>(atoi(numberList[i])));
     }
 
+    template <class T, class U>
+    void GetOption(const std::string& text, const std::string& key, const char* separator, std::set<U>& outValues)
+    {
+        SString     numbers;
+        {
+            std::string temp;
+            GetOption<T>(text, key, temp);
```
tederis, line 875: Is there a specific reason for not passing `numbers` argument directly to the function?

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -808,6 +866,25 @@ namespace SharedUtil
...
+            numbers = temp;
+        }
+        std::vector<SString> numberList;
+        numbers.Split(separator, numberList);
+        for (const auto& number : numberList)
+        {
+            if (number.empty())
+                continue;
```
tederis, line 883: Overusing of early return.

Client/core/CVersionUpdater.Util.hpp
```diff
@@ -601,8 +601,8 @@ namespace
             m_ArgMap.SetFromString(strSettings);
             // If build is 30 days old, default no report logging
             m_ArgMap.Get("filter2", strFilter, GetBuildAge() < 30 ? "+all" : "-all");
-            m_ArgMap.Get("min", iMinSize, DEFAULT_MIN_SIZE);
-            m_ArgMap.Get("max", iMaxSize, DEFAULT_MAX_SIZE);
+            m_ArgMap.Get(std::string("min"), iMinSize, DEFAULT_MIN_SIZE);
```
tederis, line 604: We're using modern C++, right? Then use curly brackets, please. Here and below.

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
```diff
@@ -225,106 +226,121 @@ void CLuaDatabaseDefs::DbFreeCallback(CDbJobData* pJobData, void* pContext)
...
+                                                                            std::optional<std::string> options, std::optional<CLuaFunctionRef> callback)
 {
-    //  element dbConnect ( string type, string host, string username, string password, string options )
-    SString strType;
-    SString strHost;
-    SString strUsername;
-    SString strPassword;
-    SString strOptions;
+    if (!username.has_value())
+        username = "";
```
tederis, line 234: `username = {};` or `username = SString{}`

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -1729,6 +1741,68 @@ namespace SharedUtil
...
+        outList.insert(outList.end(), newItems.begin(), newItems.end());
+        return !newItems.empty();
+    }
+
+    bool CArgMap::Get(const std::string& inCmd, int& value, int defaultValue) const noexcept
+    {
+        std::string result;
+        if (Get(inCmd, result))
+        {
+            value = atoi(result.c_str());
```
tederis, line 1799: Please, use C++ alternatives.

Shared/sdk/SharedUtil.Misc.h
```diff
@@ -740,6 +742,17 @@ namespace SharedUtil
...
+
+        std::string Get(const char*& inCmd) const noexcept;
+        bool        Get(const char*& inCmd, std::string& out, const char* defaultValue = "") const noexcept;
+        bool        Get(const char*& inCmd, std::vector<std::string>& outList) const noexcept;
+        bool        Get(const char*& inCmd, int& value, int defaultValue = 0) const noexcept;
+
+        std::string Get(const std::string& inCmd) const noexcept;
+        bool Get(const std::string& inCmd, std::string& out, const char* defaultValue = "") const noexcept;
+        bool Get(const std::string& inCmd, std::vector<std::string>& outList) const noexcept;
+        bool Get(const std::string& inCmd, int& value, int defaultValue = 0) const noexcept;
```
tederis: All of these methods can throw an exception. So please, remove `noexcept` specifier.

tederis (changes requested):

Client/core/CVersionUpdater.Util.hpp
TracerDS: Whats the difference?
Isnt it the matter of preference in this case and below?

Shared/sdk/SharedUtil.Misc.h
TracerDS: no. It makes the function much more readable.

Shared/sdk/SharedUtil.Misc.h
TracerDS: You mean to the external GetOption or internal (in block) GetOption?

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
TracerDS: Whats the difference? This way it is clear that Im inserting an empty string.
`= {}` may be ambigous to the compiler (empty string or empty optional?)

Shared/sdk/SharedUtil.Misc.h
TracerDS: None of these can throw an exception.

TracerDS: Only the vector one can throw an exception. I will change that to non noexcept.
Every other method wont throw an exception

Server/mods/deathmatch/logic/luadefs/CLuaDatabaseDefs.cpp
tederis: The difference is that = "" constructor is calling strlen inside whereas SString{} is not. What about the ambiguity, okay I agree with it.

Shared/sdk/SharedUtil.Misc.h
tederis: Disagree with this. Please, do not do that.

TracerDS: I won't change that.
The line would be too long.

Client/core/CVersionUpdater.Util.hpp
tederis: You can easily find out this information yourself on the web. This eliminates many subtle problems.

Shared/sdk/SharedUtil.Misc.h
tederis: I totally disagree with this approach. Perhaps other reviewers will be less strict with it. Good luck.

Shared/sdk/SharedUtil.Misc.hpp
TracerDS: It would require either defining a function for converting strings to numbers or surrounding it with try-catch.
Function would be better there, no?

TracerDS:

PR #3634, opened 2024-08-07
Fixed isPlayerHudComponentVisible not working with showhud

### LINKED ISSUE #547: isPlayerHudComponentVisible only works with setPlayerHudComponentVisible

TracerDS:
Resolves #547

### DIFF
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -43,7 +43,8 @@ void CHudSA::Disable(bool bDisabled)
     else
         MemPut<BYTE>(FUNC_Draw, 0x80);
 
-    SetComponentVisible(HUD_ALL, !bDisabled);    
+    for (const auto& component : m_HudComponentMap)
+        SetComponentVisible(component.first, !IsComponentVisible(component.first));
 
     // Also disable the radar as the above code will not hide it before the local player has spawned
     if (bDisabled)
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -43,6 +43,8 @@ void CHudSA::Disable(bool bDisabled)
     else
         MemPut<BYTE>(FUNC_Draw, 0x80);
 
+    SetComponentVisible(HUD_ALL, !bDisabled);    
+
     // Also disable the radar as the above code will not hide it before the local player has spawned
     if (bDisabled)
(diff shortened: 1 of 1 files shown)


Client/game_sa/CHudSA.cpp
```diff
@@ -43,6 +43,8 @@ void CHudSA::Disable(bool bDisabled)
     else
         MemPut<BYTE>(FUNC_Draw, 0x80);
 
+    SetComponentVisible(HUD_ALL, !bDisabled);
```
TheNormalnij: Does it work when HUD is partially hidden?
1. Disable 'money' component
2. showhud -> all hud is disabled
3. showhud again -> ?

TracerDS: it will show all components. Will fix that

Client/game_sa/CHudSA.cpp
```diff
@@ -43,6 +43,9 @@ void CHudSA::Disable(bool bDisabled)
     else
         MemPut<BYTE>(FUNC_Draw, 0x80);
 
+    for (const auto& component : m_HudComponentMap)
+        SetComponentVisible(component.first, !IsComponentVisible(component.first));
```
TheNormalnij, line 47: It looks like `isPlayerHudComponentVisible` will return true for a hidden component after `showhud`

TracerDS: Hmm, do you think a copy of the components map would be good here?
To save status of each component and then restore it on the next call

TheNormalnij:
Check conversation  in #547.
We need a way to know why a component is hidden.

@qaisjp suggested 
```lua
bool isPlayerHudComponentVisible(string component, bool checkEnabled=true)
```

I'd like to keep `isPlayerHudComponentVisible`  only for `setPlayerHudComponentVisible`.
A new `isHudEnabled` function can be added instead. 
```lua
-- returns true by default and false when showhud command is used
bool isHudEnabled()
```


Both variants are valid.

FileEX:
Any update?

TracerDS:

PR #3641, opened 2024-08-13
Add player argument to onVehicleExplode event

Liberrty:
Add player argument to onVehicleExplode event, cheaters can abuse and don't send onExplosion packet

Example code:
```lua
addEventHandler('onVehicleExplode', root, function(a, player)
    if player then
        print('Vehicle explosion caused by ' .. getPlayerName(player))
        banPlayer(player, true, false, true, 'Vehicle explosion')
    end
end)
```


fuckbtc (approved):

FileEX:
> cheaters can abuse and don't send onExplosion packet

How does adding a new argument solve the mentioned issue?

FileEX (changes requested):
You also need to edit the event parameters in the line

```cpp
m_Events.AddEvent("onVehicleExplode", "", NULL, false);
```

PR #3644, opened 2024-08-14
Fix Changing the "voiceptt" bind won't change it's "up" pressed state bind

### LINKED ISSUE #3642: Changing the "voiceptt" bind won't change it's "up" pressed state bind
### Describe the bug

Changing the `voiceptt` bind will only change it's `down` pressed state, if we want to change the `up` pressed state we need to change it manually through the `MTA/settings/coreconfig.xml` file.

Example after changing the bind to the `R` key:
```xml
<bind key="r" state="down" command="voiceptt" arguments="1"></bind>
<bind key="z" state="up" command="voiceptt" arguments="0"></bind>
```

### Steps to reproduce

1. Go to settings
2. Change the `voiceptt` bind to an other key
3. Tap the new key to talk
4. The talking state won't turn off unless you press the original key again

### Version

Client: v1.6-release-22650 (Windows 11 64-bit)
Server: v1.6-release-22650 (Debian 12 64-bit)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fix #3642

### DIFF
--- a/Client/core/CKeyBinds.cpp
+++ b/Client/core/CKeyBinds.cpp
@@ -841,7 +841,8 @@ CCommandBind* CKeyBinds::FindCommandMatch(const char* szKey, const char* szComma
     NullEmptyStrings(szKey, szArguments, szResource, szOriginalScriptKey);
 
     std::string arguments = szArguments ? szArguments : "";
-    szArguments = SharedUtil::Trim(arguments.data());
+    if (!arguments.empty())
+        szArguments = SharedUtil::Trim(arguments.data());
 
     for (KeyBindPtr& bind : m_binds)
     {


Xenius97:
Same when using bindKey() with command like
```lua
bindKey("y", "down", "radio")
bindKey("y", "up", "radio")
```

When you try to rebound it in settings to other key, only "down" changes.
Is this fixed too?

FileEX:
> Same when using bindKey() with command like
> 
> ```lua
> bindKey("y", "down", "radio")
> bindKey("y", "up", "radio")
> ```
> 
> When you try to rebound it in settings to other key, only "down" changes. Is this fixed too?

Probably

Xenius97:
> Probably

Yes, fixes that too.

PR #3645, opened 2024-08-15
Fix inconsistent extra component names

FileEX:
Often when a vehicle has variants (extra components), their names are incorrect, such as ``'extra_`'`` or other strange characters

image
image

According to the specifications, they should be named extra_a, extra_b, and so on, up to extra_f.

The issue arises because the values of the current variant in the ``CPoolsSA::AddVehicle`` function are set after ``CVehicleSA::Init`` has already been called, which also triggers ``VehicleDump``. At the time this function is called, the variant values are incorrect, often random due to their type. This PR fixes the issue, restoring the correct behavior and proper names for the extra components, from extra_a to extra_f.

image

### DIFF
--- a/Client/game_sa/CVehicleSA.cpp
+++ b/Client/game_sa/CVehicleSA.cpp
@@ -1953,7 +1953,7 @@ void CVehicleSA::AddComponent(RwFrame* pFrame, bool bReadOnly)
     if (strName == "")
     {
         // In MTA variant 255 means no variant
-        if (m_ucVariantCount == 0 && m_ucVariant == 255 || m_ucVariantCount == 1 && m_ucVariant2 == 255)
+        if ((m_ucVariantCount == 0 && m_ucVariant == 255) || (m_ucVariantCount == 1 && m_ucVariant2 == 255))
             return;
 
         // name starts with extra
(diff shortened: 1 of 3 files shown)


Client/game_sa/CVehicleSA.cpp
```diff
@@ -1957,23 +1952,22 @@ void CVehicleSA::AddComponent(RwFrame* pFrame, bool bReadOnly)
     // variants have no name field.
     if (strName == "")
     {
+        // In MTA variant 255 means no variant
+        if (m_ucVariantCount == 0 && m_ucVariant == 255 || m_ucVariantCount == 1 && m_ucVariant2 == 255)
```
TracerDS: Add () to make sure compiler wont interpret it differently:
```cpp
if ((m_ucVariantCount == 0 && m_ucVariant == 255) || (m_ucVariantCount == 1 && m_ucVariant2 == 255)) 
```

TracerDS (changes requested):

TracerDS (approved):

Dutchman101:
It's not a large change, and makes sense, it's easy enough to just test it in upcoming nightly.

PR #3646, opened 2024-08-15
Add enginePreloadWorldArea function for preloading areas

MegadreamsBE:
Adds a new function **enginePreloadWorldArea**. This function takes in three required parameters, x, y and z for the position and a fourth optional parameter, option, which is a string (or actually an enum in the code) with the following valid values: 

- models (loads the models and their textures of everything in the area, but no collisions)
- collisions (loads all the collisions in the area, but not the models and textures)
- all (does all of the above - default value). 

The game determines the zone the given coordinates are in and loads everything in it. This function is not radius based (as the underlying game functions aren't either).

The game will automatically unload this area based on its own internal rule, often this can already happen the next frame. This functions loads in the area **immediately** so it's advisable to do the desired action, such as teleporting a player/ped/vehicle, or getting the position of the ground in the same frame.

Some use cases:
Load in the area, then retrieve the ground position at the desired location, then teleport the location to that location. The map will already be loaded, and the player is neatly put on the ground.
```lua
enginePreloadWorldArea(1000, 500, 0)   
local z = getGroundPosition(1000, 500, 100)   
setElementPosition(localPlayer, 1000, 500, z)
```  
Load in the area (collisions only), then retrieve the ground position at the desired location and output it.
```lua
enginePreloadWorldArea(1000, 500, 0, "collisions")   
local z = getGroundPosition(1000, 500, 100)   
iprint("The ground position is: "..z)
```  

NOTE: This function has been tested with the building removal functions and is not conflicting with them.

### DIFF
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -504,14 +504,14 @@ void CStreamingSA::RemoveBigBuildings()
     (reinterpret_cast<void(__cdecl*)()>(0x4093B0))();
 }
 
-void CStreamingSA::LoadScene(CVector* position)
+void CStreamingSA::LoadScene(const CVector* position)
 {
-    auto CStreaming_LoadScene = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadScene;
+    auto CStreaming_LoadScene = (void(__cdecl*)(const CVector*))FUNC_CStreaming_LoadScene;
     CStreaming_LoadScene(position);
 }
 
-void CStreamingSA::LoadSceneCollision(CVector* position)
+void CStreamingSA::LoadSceneCollision(const CVector* position)
 {
-    auto CStreaming_LoadSceneCollision = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadSceneCollision;
+    auto CStreaming_LoadSceneCollision = (void(__cdecl*)(const CVector*))FUNC_CStreaming_LoadSceneCollision;
     CStreaming_LoadSceneCollision(position);
 }
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -503,3 +503,15 @@ void CStreamingSA::RemoveBigBuildings()
 {
     (reinterpret_cast<void(__cdecl*)()>(0x4093B0))();
 }
+
+void CStreamingSA::LoadScene(CVector* position)
+{
+    auto CStreaming_LoadScene = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadScene;
+    CStreaming_LoadScene(position);
+}
+
+void CStreamingSA::LoadSceneCollision(CVector* position)
+{
+    auto CStreaming_LoadSceneCollision = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadSceneCollision;
+    CStreaming_LoadSceneCollision(position);
+}
--- a/Client/game_sa/CStreamingSA.h
+++ b/Client/game_sa/CStreamingSA.h
@@ -20,6 +20,8 @@
 #define FUNC_LoadAllRequestedModels                  0x40EA10
 #define FUNC_CStreaming__HasVehicleUpgradeLoaded     0x407820
 #define FUNC_CStreaming_RequestSpecialModel          0x409d10
+#define FUNC_CStreaming_LoadScene                    0x40EB70
+#define FUNC_CStreaming_LoadSceneCollision           0x40ED80
 
 struct CArchiveInfo
 {
@@ -76,6 +78,9 @@ public:
 
     void          MakeSpaceFor(std::uint32_t memoryToCleanInBytes) override;
     std::uint32_t GetMemoryUsed() const override;
+
+    void LoadScene(CVector* position);
+    void LoadSceneCollision(CVector* position);
 
 private:
     void AllocateArchive();
--- a/Client/game_sa/CStreamingSA.h
+++ b/Client/game_sa/CStreamingSA.h
@@ -79,8 +79,8 @@ public:
     void          MakeSpaceFor(std::uint32_t memoryToCleanInBytes) override;
     std::uint32_t GetMemoryUsed() const override;
 
-    void LoadScene(CVector* position);
-    void LoadSceneCollision(CVector* position);
+    void LoadScene(const CVector* position);
+    void LoadSceneCollision(const CVector* position);
 
 private:
     void AllocateArchive();
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -2554,7 +2554,7 @@ void CLuaEngineDefs::EnginePreloadWorldArea(PreloadAreaOption option, CVector po
     switch (option)
     {
         case PreloadAreaOption::ALL:
-        case PreloadAreaOption::WORLD_OBJECTS:
+        case PreloadAreaOption::MODELS:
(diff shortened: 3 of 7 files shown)


tederis (dismissed):
Looks great

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2547,3 +2548,23 @@ eModelLoadState CLuaEngineDefs::EngineStreamingGetModelLoadState(std::uint16_t m
 
     return g_pGame->GetStreaming()->GetStreamingInfo(modelId)->loadState;
 }
+
+void CLuaEngineDefs::EnginePreloadWorldArea(PreloadAreaOption option, CVector position)
```
TracerDS: maybe the other way around with optional PreloadAreaOption?
```cpp
void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<PreloadAreaOption> option)
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2547,3 +2548,23 @@ eModelLoadState CLuaEngineDefs::EngineStreamingGetModelLoadState(std::uint16_t m
...
+
+            if (option != PreloadAreaOption::ALL)
+            {
+                break;
+            }
+        case PreloadAreaOption::COLLISIONS:
+            g_pGame->GetStreaming()->LoadSceneCollision(&position);
+            break;
+        default:
+            throw std::invalid_argument("Invalid option at argument 1");
```
TracerDS: Maybe `if` would be better here?

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
MegadreamsBE: I like the idea, would be cleaner that way. I'll make the change

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
MegadreamsBE: Hmm, I don't see a way to remove the if-statement for the models one. We want to prevent fallthrough if we're not using the ALL option.

tederis: `switch` is okay here

MegadreamsBE: Really shows that I didn't sleep a lot this night lol. Completely misunderstood. Yeah, I'd rather have a switch here than if-statements. Especially if this function may get extended in the future

TheNormalnij:
```c++
if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::MODELS)
    g_pGame->GetStreaming()->LoadScene(&position);
    
```
`default` is unreachable, argument parser does this check

Client/sdk/game/CStreaming.h
```diff
@@ -37,6 +37,13 @@ enum class eModelLoadState : std::uint32_t
     LOADSTATE_FINISHING = 4
 };
 
+enum class PreloadAreaOption
+{
+    MODELS = 0,
+    COLLISIONS,
+    ALL
+};
```
TheNormalnij, line 45: ePreloadAreaOption

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
MegadreamsBE: I'll do that, originally I didn't have the enum, just left it after, just in case

Client/sdk/game/CStreaming.h
MegadreamsBE: Do we still enforce that, despite not using Hungarian notation anymore?

MegadreamsBE: I know we haven't in some other enums

TracerDS: I think you can leave it as it is.

MegadreamsBE:
Updated based on the reviews (except the enum name, which I kept the way it is as Hungarian notation is not used in the codebase anymore).

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2549,22 +2549,13 @@ eModelLoadState CLuaEngineDefs::EngineStreamingGetModelLoadState(std::uint16_t m
...
 
-void CLuaEngineDefs::EnginePreloadWorldArea(PreloadAreaOption option, CVector position)
+void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<PreloadAreaOption> option)
 {
-    switch (option)
-    {
-        case PreloadAreaOption::ALL:
-        case PreloadAreaOption::MODELS:
-            g_pGame->GetStreaming()->LoadScene(&position);
+    option = option.value_or(PreloadAreaOption::ALL);
```
TracerDS: its better to use `.has_value` instead `value_or` in this case as you will create additional copy:
```cpp
if (!option.has_value())
    option = PreloadAreaOption::ALL;
```

TracerDS (approved):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
MegadreamsBE: Didn't know that, thanks! I made the change

Client/game_sa/CStreamingSA.cpp
```diff
@@ -503,3 +503,15 @@ void CStreamingSA::RemoveBigBuildings()
...
+{
+    auto CStreaming_LoadScene = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadScene;
+    CStreaming_LoadScene(position);
+}
+
+void CStreamingSA::LoadSceneCollision(CVector* position)
+{
+    auto CStreaming_LoadSceneCollision = (void(__cdecl*)(CVector*))FUNC_CStreaming_LoadSceneCollision;
+    CStreaming_LoadSceneCollision(position);
+}
```
TracerDS: maybe const vector?

Client/game_sa/CStreamingSA.h
```diff
@@ -77,6 +79,9 @@ class CStreamingSA final : public CStreaming
     void          MakeSpaceFor(std::uint32_t memoryToCleanInBytes) override;
     std::uint32_t GetMemoryUsed() const override;
 
+    void LoadScene(CVector* position);
+    void LoadSceneCollision(CVector* position);
```
TracerDS: up

Client/sdk/game/CStreaming.h
```diff
@@ -68,4 +75,6 @@ class CStreaming
     virtual void   MakeSpaceFor(std::uint32_t memoryToCleanInBytes) = 0;
     virtual std::uint32_t GetMemoryUsed() const = 0;
     virtual void          RemoveBigBuildings() = 0;
+    virtual void          LoadScene(CVector* position) = 0;
+    virtual void          LoadSceneCollision(CVector* position) = 0;
```
TracerDS: up

Client/game_sa/CStreamingSA.cpp
MegadreamsBE: Done!

Dutchman101:
Good code reviews, and a fairly small change. Thanks!

Xenius97:
This won't work with objects (tested) but with buildings everything is fine.

Issue #449, opened 2018-09-11
Cannot cancel natural onTrailerAttach events


ArranTuna:
**Describe the bug**
If you attach a trailer to a vehicle **naturally** (by using a towtruck), `cancelEvent()` does not work. 

Also for attaching a trailer to a Truck.

**To Reproduce**
```lua
testcar = createVehicle (450, 0,0,3.5)
function detachTrailer(theTruck)
    if source == testcar then
        outputDebugString("Cancel attached testcar",3)
        cancelEvent()
    end
 end
addEventHandler("onTrailerAttach", getRootElement(), detachTrailer)
```

**Expected behavior**
cancelEvent should stop the attachment of trailer.

**MTA Client:** client v1.5.5-release-14007 (@patrikjuvonen verified)

**MTA Server:** Server v1.5.5-release-14007 (@patrikjuvonen verified)

**Additional context**
From https://bugs.mtasa.com/view.php?id=8008

PR #3660, opened 2024-08-19
onClientTrailerAttach rework

### LINKED ISSUE #2119: onClientTrailerAttach: cancelEvent() not working
**Describe the bug**
If I try the onClientTrailerAttach/onTrailerAttach event with cancelEvent() to interrupt, it will not abort.

**To reproduce**

Steps to reproduce the behaviour:
1. Sit in a truck
2. Reverse to a trailer.

[trailer.zip](https://github.com/multitheftauto/mtasa-blue/files/6135434/trailer.zip)

**Expected behaviour**
nothing will happen

**Screenshots**

**Version**
1.5.8

**Additional context**

FileEX:
Fixed #2119 and probably #449

Now the ``onClientTrailerAttach`` event is not based on vehicle synchronization but on actual interactions with the trailer. The event can be canceled by preventing the trailer/vehicle from being attached.

I also tried the ``onClientTrailerDetach`` event to be able to cancel it but I encountered two problems

- Weird vehicle physics bugs (fixable)
- Event flood for trucks with trailers when they are in a position that detaches the trailer (in case with cancelEvent)
Screenshot_8

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -300,6 +300,8 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
     g_pMultiplayer->SetPedStepHandler(CClientGame::StaticPedStepHandler);
     g_pMultiplayer->SetVehicleWeaponHitHandler(CClientGame::StaticVehicleWeaponHitHandler);
     g_pMultiplayer->SetAudioZoneRadioSwitchHandler(CClientGame::StaticAudioZoneRadioSwitchHandler);
+    g_pMultiplayer->SetAttachTrailerHandler(CClientGame::StaticAttachTrailerHandler);
+    g_pMultiplayer->SetTowVehicleHandler(CClientGame::StaticTowVehicleHandler);
     g_pGame->SetPreWeaponFireHandler(CClientGame::PreWeaponFire);
     g_pGame->SetPostWeaponFireHandler(CClientGame::PostWeaponFire);
     g_pGame->SetTaskSimpleBeHitHandler(CClientGame::StaticTaskSimpleBeHitHandler);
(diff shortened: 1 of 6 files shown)


TracerDS (approved):

CrosRoad95:
>  able to cancel it but I encountered two problems

No, make function that enable/disable attaching. cancelEvent in general is bad approach. it will fix second issue with event flood because event never reach lua.

FileEX:
> > able to cancel it but I encountered two problems
> 
> No, make function that enable/disable attaching. cancelEvent in general is bad approach. it will fix second issue with event flood because event never reach lua.

It even sounds reasonable. I can make a function like ``setTrailerDetachable`` that would work in the same way as ``setTrainDerailable``. Then the flood event problem will be solved, but I'm not sure how the trailer physics will behave. I'll definitely check it out soon

tederis:
This PR is potentially backward incompatible. Because currently scripts rely on syncronization based events. And sometimes it may be important to stick to the actual(syncronized) trailer's state. This PR can potentially break this logic.

FileEX:
> This PR is potentially backward incompatible. Because currently scripts rely on syncronization based events. And sometimes it may be important to stick to the actual(syncronized) trailer's state. This PR can potentially break this logic.

So what do you propose?

tederis:
> So what do you propose?

Wait for MTA 2.0 maybe. But seriously there's little you can do. Maybe this problem isn't that noticeable and we can just merge the PR.

Fernando-A-Rocha (approved):
Merge it sir

TheNormalnij:
> So what do you propose?

onClientTrailerPreAttach?

FileEX:
> > So what do you propose?
> 
> onClientTrailerPreAttach?

Now I'm not sure if this PR makes sense. When the vehicle is in the position to attach a trailer, a flood event appears when cancelEvent is used

FileEX:
I don't see any positives from this PR at the moment. onClientTrailerAttach also causes a flood event if the vehicle is in the attachable position and the event is canceled.

I created the ``setTrailerDetachable`` function, but I encountered physics bugs, the vehicle shoots into the air and I was unable to fix it in various ways, such as setting the velocity / angular velocity to 0, skip AddForce, apparently the problem lies in the vehicle physics processing itself. 

I don't think a function like ``setTrailerAttachable`` makes much sense, because what if we want to cancel the trailer attach only under certain conditions, such as wrong vehicle elementdata? Then this function will not help us. In addition to trailers, we also need the ability to cancel the towing of the vehicle under certain conditions, and ``setTrailerAttachable`` will not help us with this either.

PR #3671, opened 2024-08-24
setElementModel memory leak fix

tederis:
Fixes the memory leak described in #2672. It also removes a functionality related to previous attempts to handle geometry references.


Client/game_sa/CPedSA.cpp
```diff
@@ -98,9 +98,18 @@ void CPedSA::Init()
 
 void CPedSA::SetModelIndex(DWORD dwModelIndex)
 {
-    DWORD dwFunction = FUNC_SetModelIndex;
+    // Delete any existing RwObject first
+    DWORD dwFunction = GetPedInterface()->vtbl->DeleteRwObject;
```
TheNormalnij: This method is declared for 'CEntitySAInterface' already

tederis: Do you mean DeleteRwObject?

TheNormalnij: Yes

tederis: Okay. I've made the corrections.

TheNormalnij (approved):

TracerDS (approved):

PR #3675, opened 2024-08-24
Fix engineRequestModel crash caused by CResourceModelStreamer

### LINKED ISSUE #3672: Game crashes when leaving a local server on windows
### Describe the bug

When I set up a local server and do some things on it, if I leave it in any way (disconnect, quitting the game, kick or ban etc.) the game will exit in a crash. This happens on windows x86 x64 servers both. Tried server versions 22676, 22681, 22685.

(On an other server which runs 22676 on X64 Linux, there is no problem with leaving the game.)
[Crash](https://i.imgur.com/axXWwWo.png)
The crash reports were sent. 

### Steps to reproduce

1. Clean fresh install Client & Server from nightly.
2. Launch server
3. Launch client, connect to the local server
4. Disconnect or Quit or Kick/Ban

### Version

I tried with Client Versions 22676, 22680, 22684 and Server versions 22676, 22681, 22685

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

TheNormalnij:
Fixes #3672


TracerDS (approved):

Client/mods/deathmatch/logic/CResourceModelStreamer.cpp
```diff
@@ -106,11 +106,15 @@ void CResourceModelStreamer::ReleaseAll()
...
+    auto refs = m_requestedModels.find(modelId);
+    if (refs == m_requestedModels.end())
+        return;
+
+    std::uint16_t refsCount = refs->second;
 
     if (refsCount > 0)
     {
-        refsCount = 0;
+        m_requestedModels[modelId] = 0;
```
botder: Please use the iterator or change `refsCount` to be a reference, to change the value here. The compiler will generate more code than necessary otherwise: https://godbolt.org/z/36f4a8Koe

TheNormalnij: Fixed

Client/mods/deathmatch/logic/CResourceModelStreamer.cpp
```diff
@@ -106,7 +106,11 @@ void CResourceModelStreamer::ReleaseAll()
 
 void CResourceModelStreamer::FullyReleaseModel(std::uint16_t modelId)
 {
-    std::uint16_t &refsCount = m_requestedModels[modelId];
```
tederis, line 109: Are you sure that this is the reason? std::unordered_map.operator[] inserts an entry if it's not present. In this case   `refsCount` would be 0 which would lead to `if (refsCount > 0)` be failed.

tederis: But nevertheless there have to be a check, for sure.

TheNormalnij: > std::unordered_map.operator[] inserts an entry if it's not present.

Yes, you are right, but it works only in Debug build. Release build has different behavior here

TracerDS: What changes?

tederis: > > std::unordered_map.operator[] inserts an entry if it's not present.
> 
> Yes, you are right, but it works only in Debug build. Release build has different behavior here

`unordered_map` requires mapped type to be `DefaultConstructible`. That means that a number will always be initialized with zero. There shouldn't be difference between release and debug.

TheNormalnij: I've checked it with breakpoints. For some interesting reason only Debug build is unaffected. I saw similar compiler bugs related to 32 bit applications, but I don't know real reason for this crash.

TracerDS: Maybe its related to [the previous `getAccountType` release crash](https://discord.com/channels/801330706252038164/801411628600000522/1244235367406764083)?

tederis (approved):
Regardless on fixes this the crash or not, these changes are advantageous.

PR #3680, opened 2024-08-26
Make docker run locally

TracerDS:
This PR localizes Docker. It no longer copies mtasa repo but local copy of the mta.


Dockerfile
```diff
@@ -21,6 +21,15 @@ RUN apt-get update && \
...
 
+RUN mkdir -p Client Server Shared utils vendor
+
+COPY Client Client
+COPY Server Server
+COPY Shared Shared
+COPY utils utils
+COPY vendor vendor
+COPY ./* .
+
```
TheNormalnij, line 32: You should use `-v` parameter in `docker run`  to mount these folders into a container

utils/docker-entrypoint.sh
```diff
@@ -21,11 +21,6 @@ if [[ $AS_BUILDAGENT = "1" ]]; then
 fi
 
 # Manually invoke build process
-# So, first make a shallow clone of the repository if it not exists
-umask 000
-if [ ! -f ./premake5.lua ]; then
-    git clone --depth=1 https://github.com/multitheftauto/mtasa-blue.git .
-fi
```
TheNormalnij, line 28: Why? Nothing is wrong here. This line allows you to build MTA without cloning the MTA repository.

TracerDS: this line actually clones the entire repo all the time.

docker-build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+NAME="mtasa"
+
+docker build . -t $NAME
+ID = $(docker images -q $NAME)
```
TheNormalnij, line 6: Why do you need that and how to use it?

TracerDS: When you execute the script, it will build the docker image and print its id

utils/docker-entrypoint.sh
TheNormalnij: It doesn't. Check your command line and current folder

TheNormalnij:
It works locally now

```sh
docker build -t mtasabuilder:latest .
docker run --rm -v ./:/build mtasabuilder:latest
```

Dockerfile
TracerDS: How would it work?

utils/docker-entrypoint.sh
TracerDS: It does. Docker pulls mta repo from github if that line is saved. Without it, it runs only local changes

TheNormalnij: No. I tested it

Dockerfile
TheNormalnij: You should mount mta repository (or an empty folder) to `/build` folder in a container. `docker-entrypoint.sh` runs inside this folder.

docker-build.sh
TheNormalnij: We don't need that ID usually

utils/docker-entrypoint.sh
TracerDS: Me too. It didnt accept local copy unless I remove that line.
If you try to build mta then you already have necessary files in your filesystem.

docker-build.sh
TracerDS: Its needed when you want to fully terminate a docker session

TheNormalnij: Container tags are used for that.
```sh
docker stop <tag>
docker rm <tag>
```

utils/docker-entrypoint.sh
TheNormalnij:
```sh
cd <mta sa repo here>
docker build -t mtasabuilder:latest .
docker run --rm -v ./:/build mtasabuilder:latest
```
Does it work?

docker-build.sh
TracerDS: both of these commands require the id that is provided via the output of the command above

utils/docker-entrypoint.sh
TracerDS: "volume name is too short, names should be at least two alphanumeric characters."

botder:
Did you ever try the steps under **Building with Docker** from the README?
https://github.com/multitheftauto/mtasa-blue/blob/550a152551ec101d042ae774f3a54df2b04bea85/README.md?plain=1#L113-L122

TracerDS:
> Did you ever try the steps under **Building with Docker** from the README?
> https://github.com/multitheftauto/mtasa-blue/blob/550a152551ec101d042ae774f3a54df2b04bea85/README.md?plain=1#L113-L122

I did, but no matter what I do I always get a copy of the github repo instead of my local one.

botder:
> I did, but no matter what I do I always get a copy of the github repo instead of my local one.

Alright, but did you try using `--entrypoint /bin/bash` to look around and see if you properly mounted the files in /build ?

TheNormalnij:
> "volume name is too short, names should be at least two alphanumeric characters."

Do you use powershell in windows?

TracerDS:
> Alright, but did you try using `--entrypoint /bin/bash` to look around and see if you properly mounted the files in /build ?

Didn't try that exact method but when `ls`ing the root dir, no directories were copied.

> Do you use powershell in windows?

Nope

botder:
> > Alright, but did you try using `--entrypoint /bin/bash` to look around and see if you properly mounted the files in /build ?
> 
> Didn't try that exact method but when `ls`ing the root dir, no directories were copied.

You're supposed to mount your local directory into the Docker container, and by poking around within a /bin/bash shell, you can see if your mount is actually working. If it's not there, then obviously the docker-entrypoint.sh script is going to clone the repository everytime.

utils/docker-entrypoint.sh
TheNormalnij:
```
docker run --rm -v <full path to repo>:/build mtasabuilder:latest
```
?

TheNormalnij:
Invalid / a system issue

PR #3686, opened 2024-08-27
Add definitions for CHeliSAInterface

TracerDS:


Client/game_sa/CHeliSA.h
```diff
@@ -15,10 +15,48 @@
...
+    std::uint32_t field_A14;
 };
+static_assert(sizeof(CHeliSAInterface) == 0xA18, "Invalid size for CHeliSAInterface");
 
 class CHeliSA final : public virtual CHeli, public virtual CAutomobileSA
 {
 public:
     CHeliSA(CHeliSAInterface* pInterface);
+    CHeliSAInterface*       GetHeliInterface() noexcept { return reinterpret_cast<CHeliSAInterface*>(GetInterface()); }
+    //const CHeliSAInterface* GetHeliInterface() const noexcept { return reinterpret_cast<const CHeliSAInterface*>(GetInterface()); }
```
tederis: Commented out?

TracerDS: Yes, I couldnt make a const pointer due to some inconsistencies in the parent classes

tederis: Is this a draft? Because the repository isn't the right place for this kind of comments.

TracerDS: No, its not a draft. Repo has a lot of these comments. One doesnt make a difference.
In the refactoring PR I will uncomment that and implement const reference.

tederis: Like I said the repo is not the place for this type of comments. So please, remove it.

tederis (approved):

PR #3694, opened 2024-09-03
Fast fix CModelInfoSA::IsValid

G-Moris:
Without this fix, corrupted models caused errors.


TracerDS (approved):

Fernando-A-Rocha (approved):

TheNormalnij:
What does it fix? GTA has some special models without allocated DFF in IMG  and it's ok.

G-Moris:
> What does it fix? GTA has some special models without allocated DFF in IMG and it's ok.

Yes, maybe I should think about all the subtleties.

Issue #3703, opened 2024-09-07
Add isVehicleNitroActivated server side


Proxy-99:
### Is your feature request related to a problem? Please describe.

I am trying to check through the server vehicle nitro activation would be much easier than using events

### Describe the solution you'd like

if possible add it server side related nitro functions like https://wiki.multitheftauto.com/wiki/SetVehicleNitroActivated etc

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

botder:
~~Fixed in e9e5819c394987de2b9a5d581c4df9fd47057d9d~~ this is the commit for `set` not `is`

Issue #2726, opened 2022-08-31
Sync vehicle nitro state for nearby players also when it is disabled by setVehicleNitroActivated


C0DELEVEL:
### Is your feature request related to a problem? Please describe.

Currently, when nitro in a vehicle is activated (by button or by setVehicleNitroActivated function), this information is synchronized to other players. However, the moment we call setVehicleNitroActivated(vehicle, false) in the driver (which is syncer) - nitro will be disabled only for the driver (as localPlayer which executed this function), for other players it will still be active and cause a lot of desynchronization (acceleration is greater than on syncer side).

### Describe the solution you'd like

Synchronize the status of nitro according to the current status of syncer (driver).

### Describe alternatives you've considered

Currently, you have to work around this with a script, e.g. manually synchronizing nitro state (mainly off, as on by key is synced correctly) - driver -> server -> players nearby

### Additional context

The problem mainly relates to the implementation of nitro on the button - hold = on, release = off. Sample script code that turns off the nitro when you release the button:

function toggleNOS(btn, state)
	local veh = getPedOccupiedVehicle(localPlayer)
	if veh and getVehicleController(veh) == localPlayer then
		if isVehicleNitroActivated(veh) then
			setVehicleNitroActivated (veh, false) -- <- This should be synced with nearby players, as it is when we set it to true.
		end
	end
end
bindKey("vehicle_fire", "up", toggleNOS)

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

Proxy-99:
This should fix the issue #3706

C0DELEVEL:
> This should fix the issue #3706

@Proxy-99 Thank you for your contribution, but I don't think this is a good solution (#3706) to this issue. It is useful in some cases, but not necessarily in the one I reported.

Why? Your solution has 2 drawbacks:
- It requires serverside, while the state of vehicles and nitro are synchronized from clientside
- Such information is sent to all players on the server, which is a waste of cpu/bandwith (why player A on dim 1 need to know that vehicle on dim 900 has the nitro activated/deactivated?)

My suggestion was to improve the synchronization packets so that it works like turning on nitro on clientside.

1. Pressing LPM to turn on nitro -> sends packet to nearby players about active nitro in car
2. Depressing LPM to disable nitro (nitro is turned off clientside via `setVehicleNitroActivated`) -> sends an analogous packet to players nearby with disabled nitro status in the car.

[...]

PR #3706, opened 2024-09-08
Adding setVehicleNitroActivated server-side

Proxy-99:
part of #3703 help with #2726


Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,25 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+            if (pVehicle->IsNitroInstalled())
+            {
+                if (state)
+                    pVehicle->SetNitroLevel(pVehicle->GetNitroLevel() - 1.0001f);
+                else
+                    pVehicle->SetNitroLevel(pVehicle->GetNitroLevel() + 1.0001f);
+               
+            }
+        }
+    }
```
TracerDS: early return

Why `GetLevel(0 - 1.0001f)` and not just 0/1?

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -2982,3 +2983,13 @@ int CLuaVehicleDefs::SetVehiclePlateText(lua_State* luaVM)
...
+
+
+bool CLuaVehicleDefs::SetVehicleNitroActivated(CVehicle* pVehicle, bool state)
+{
+    CBitStream BitStream;
+    BitStream.pBitStream->WriteBit(state);
+
+    m_pPlayerManager->BroadcastOnlyJoined(CElementRPCPacket(pVehicle, SET_VEHICLE_NITRO_ACTIVATED, *BitStream.pBitStream));
+    return true;
+}
```
TracerDS: if bitstream doesnt throw exceptions, add `noexcept`

TracerDS (changes requested):

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
Proxy-99: I am not sure but this is how the client code use it maybe something about charging level

```c++
bool CStaticFunctionDefinitions::SetVehicleNitroActivated(CClientEntity& Entity, bool bActivated)
{
    RUN_CHILDREN(SetVehicleNitroActivated(**iter, bActivated))

    if (IS_VEHICLE(&Entity))
    {
        CClientVehicle& pVehicle = static_cast<CClientVehicle&>(Entity);
        if (pVehicle.IsNitroInstalled())
        {
            bool bAlreadyActivated;
            IsVehicleNitroActivated(pVehicle, bAlreadyActivated);

            if (bAlreadyActivated != bActivated)
            {
                if (bActivated)
                    pVehicle.SetNitroLevel(pVehicle.GetNitroLevel() - 1.0001f);
                else
                    pVehicle.SetNitroLevel(pVehicle.GetNitroLevel() + 1.0001f);
            }

            return true;
        }
    }

    return false;
}
```

Proxy-99: // If nitro level < 0, nitro is activated. (until nitro level reaches -1, at that point it will become 0 and increase instead of decrease)

TracerDS: that looks very sketchy

Proxy-99: well from what I understand setting nitro at certain value will trigger the nitro activation same as for deactivation
since we calling SetNitroLevel()

Server/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -126,6 +126,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"getVehicleSirens", GetVehicleSirens},
         {"getVehicleSirenParams", GetVehicleSirenParams},
         {"setVehiclePlateText", SetVehiclePlateText},
+        {"setVehicleNitroActivated", ArgumentParser <SetVehicleNitroActivated>},
```
TracerDS: remove space before `<`

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,25 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+        return;          
+
+    if (!pVehicle->IsNitroInstalled())
+        return;            
+
+    // Apply nitro level change
+    if (state)
+        pVehicle->SetNitroLevel(pVehicle->GetNitroLevel() - 1.0001f);
+    else
+        pVehicle->SetNitroLevel(pVehicle->GetNitroLevel() + 1.0001f);
```
TracerDS: Dont use hungarian notation

TracerDS (approved):

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,25 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
         }
     }
 }
+
+
+void CVehicleRPCs::SetVehicleNitroActivated(CClientEntity* pSourceEntity, NetBitStreamInterface& bitStream)
+{
+    bool state;
+    if (!bitStream.ReadBit(state))
+        return;
```
FileEX: It probably shouldn't be like this. When using setVehicleNitroActivated to false on the server side, simply nothing will happen because you are breaking the function here.

```cpp
bool state = bitStream.ReadBit();
```

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,25 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+
+    if (!vehicle->IsNitroInstalled())
+        return;            
+
+    // Apply nitro level change
+    if (state)
+        vehicle->SetNitroLevel(vehicle->GetNitroLevel() - 1.0001f);
+    else
+        vehicle->SetNitroLevel(vehicle->GetNitroLevel() + 1.0001f);
+
```
FileEX: Empty line

FileEX (changes requested):

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
Proxy-99: I do not get it should not bitStream.ReadBit(state) check first if read was successful then return a bool
and state is pass by reference

FileEX: In the case of ReadBit, this makes no sense, because such a condition will break the code if false. 

In the case of ReadBit, the result is the bit value, i.e. 0 (false) or 1 (true). Such a check makes sense only in the case of data such as Read, ReadVector, etc., but not in the case of ReadBit, because here, naturally, if you used WriteBit(false) earlier, the result of the ReadBit will also be false. There is no way for the read to fail here as it is only 1 bit

FileEX (approved):
Remember to add a condition to ``CLuaVehicleDefs::SetVehicleNitroActivated`` in the future whether the new state is different from the previous one (When IsVehicleNitroActivated will be available)

```cpp
if (state == IsVehicleNitroActivated(vehicle))
    return false;
```

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,22 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+{
+    bool state = bitStream.ReadBit();
+
+    CClientVehicle* vehicle = m_pVehicleManager->Get(pSourceEntity->GetID());
+    if (!vehicle)
+        return;          
+
+    if (!vehicle->IsNitroInstalled())
+        return;            
+
```
tederis, line 686: It seems that there has to be IsVehicleNitroActivated check as well.

Proxy-99: I add this check rely on the client code method below it
```c++
 if ((vehicle->GetNitroLevel() < 0) == state)
     return;
``` 

```c++
bool CStaticFunctionDefinitions::IsVehicleNitroActivated(CClientVehicle& Vehicle, bool& bActivated)
{
    if (Vehicle.IsNitroInstalled())
    {
        // If nitro level < 0, nitro is activated. (until nitro level reaches -1, at that point it will become 0 and increase instead of decrease)
        bActivated = (Vehicle.GetNitroLevel() < 0);
        return true;
    }

    return false;
}

```

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
```diff
@@ -653,3 +654,26 @@ void CVehicleRPCs::SetVehiclePlateText(CClientEntity* pSourceEntity, NetBitStrea
...
+    CClientVehicle* vehicle = m_pVehicleManager->Get(pSourceEntity->GetID());
+    if (!vehicle)
+        return;          
+
+    if (!vehicle->IsNitroInstalled())
+        return;
+
+     // If nitro level < 0, nitro is activated. (until nitro level reaches -1, at that point it will become 0 and increase instead of decrease)
+    if ((vehicle->GetNitroLevel() < 0) == state)
```
tederis: There should be the floating point literal `.0f`.

Client/mods/deathmatch/logic/rpc/CVehicleRPCs.cpp
tederis: By the way
```
bool bAlreadyActivated;
IsVehicleNitroActivated(pVehicle, bAlreadyActivated);
```
has undefined behaviour. Maybe someone desires to fix it.

Shared/sdk/net/rpc_enums.h
```diff
@@ -285,8 +285,10 @@ enum eElementRPCFunctions
 
     RESPAWN_OBJECT,
     TOGGLE_OBJECT_RESPAWN,
-
+
```
tederis: Please, remove the spaces here

botder (approved):

PR #3719, opened 2024-09-12
Remove ability to skip 'addDebugHook'

PlatinMTA:
`addDebugHook` will not be skipped inside another `addDebugHook`. This could lead to abuse and render some lua anti-cheat methods useless.

### TEST RESOURCE:
[test_addDebugHook.zip](https://github.com/user-attachments/files/16983869/test_addDebugHook.zip)

This test resource will add two debug hooks. The first one will hook itself to `addDebugHook` and will try to skip it on further use. The second hook will try to hook itself to `setElementPosition` and will try to skip that function only if `element == localPlayer`.

Previous behavior would end up with the second hook not being added, hence no skip will happen for `setElementPosition`. With this PR that shouldn't be possible anymore. You can still skip `removeDebugHook` or even just use that function to remove the previously added hooks, but you won't be able to skip the new addition of debug hooks.

`client.lua:`
```lua
function onStart()
	local db1 = addDebugHook("preFunction", onPreFunction, {"addDebugHook"})
	local db2 = addDebugHook("preFunction", onPreFunction_position, {"setElementPosition"})
	
	iprint("tick", "position", "first debughook", "second debughook", ":")
	iprint(getTickCount(), setElementPosition(localPlayer, x, y, z+1), db1, db2)
end
addEventHandler("onClientResourceStart", resourceRoot, onStart)

function onPreFunction(sourceResource, functionName, isAllowedByACL, luaFilename, luaLineNumber, ...)
	local args = {...}
	iprint(sourceResource, functionName, isAllowedByACL, luaFilename, luaLineNumber, ...)
	return "skip"
end

function onPreFunction_position(sourceResource, functionName, isAllowedByACL, luaFilename, luaLineNumber, ...)
	local args = {...}
	if args[1] ~= localPlayer then
		return
	end
	
	iprint(sourceResource, functionName, isAllowedByACL, luaFilename, luaLineNumber, ...)
	return "skip"
end
```

Even while debugging it doesn't make sense to skip `addDebugHook`, I really can't see an use case for that.

#### What was the issue?
If the client is using a basic lua injector they could easily make a hook before any other script loads in order to prevent adding new hooks later on, rendering some anti-cheat methods useless. This should make that harder for script kiddies.


Shared/mods/deathmatch/logic/CDebugHookManager.cpp
```diff
@@ -659,7 +659,9 @@ bool CDebugHookManager::CallHook(const char* szName, const std::vector<SDebugHoo
             CLuaArgument* returnedValue = *returnValues.begin();
             if (returnedValue->GetType() == LUA_TSTRING)
             {
-                if (returnedValue->GetString() == "skip")
+                // We don't want to skip the creation of new debug hooks
+                SString& strName = SString("%s", szName);
+                if (returnedValue->GetString() == "skip" && strName != "addDebugHook")
```
TracerDS: use `std::string` instead:
```cpp
if (returnedValue->GetString() == "skip" && std::string(szName) != "addDebugHook")
```
Or for the best solution:
```cpp
if (returnedValue->GetString() == "skip" && strcmp(name, "addDebugHook"))
```

TracerDS (changes requested):

Shared/mods/deathmatch/logic/CDebugHookManager.cpp
TheNormalnij: string_view

CrosRoad95:
> This could lead to abuse and render some lua anti-cheat methods useless.

Cheaters are playing on forks, they will revert it

PlatinMTA:
> I don't know what are you trying to achieve

Well you can read the PR again I guess (not all cheaters use the same method man, and what would the benefit of hiding `addDebugHook` anyways if they are using custom forks lmao, they can make their own implementation).

Fernando-A-Rocha (approved):
Yep this is better than not having this protection

TracerDS (approved):

Xenius97:
Why was this necessary? You've just made things easier for cheaters, essentially hurting ourselves. For us, the first element of our anti-cheat was to prevent abuse of debug hooks—and the only way to do this is to register a hook and skip the creation of additional ones. This is now a big blow for servers that were using similar methods.

Xenius97:
I've deleted prev comment, i'm explaining why.

> If the client is using a basic lua injector they could easily make a hook before any other script loads in order to prevent adding new hooks later on, rendering some anti-cheat methods useless. This should make that harder for script kiddies.

You're make it easier for them.
Thats right, they could load before any other scripts but:
- If you add debug hook after they added (and skipped it) addDebugHook will return always false, so you can detect and ban him before he could do anything.
Now you cannot detect it, cause skip is removed. Now they can watch your entire server - and if there is any bug or abusable event in your server, they will find it.

```lua
if not addDebugHook("preFunction", checkFunctions, {"addDebugHook", "createExplosion", "blowVehicle", "createProjectile", "detonateSatchels", "setPedOnFire", "createFire", "setVehicleHandling", "loadstring"}) then
	-- here comes the ban
end
```

FileEX:
That's true. Paradoxically, the PR that was supposed to increase security actually decreased it. I also think that probably many anti-cheat scripts were weakened by this not so well thought out change

PlatinMTA:
> I've deleted prev comment, i'm explaining why.
> 
> > If the client is using a basic lua injector they could easily make a hook before any other script loads in order to prevent adding new hooks later on, rendering some anti-cheat methods useless. This should make that harder for script kiddies.
> 
> You're make it easier for them.
> Thats right, they could load before any other scripts but:
> - If you add debug hook after they added (and skipped it) addDebugHook will return always false, so you can detect and ban him before he could do anything.
> Now you cannot detect it, cause skip is removed. Now they can watch your entire server - and if there is any bug or abusable event in your server, they will find it.
> 
> ```lua
> if not addDebugHook("preFunction", checkFunctions, {"addDebugHook", "createExplosion", "blowVehicle", "createProjectile", "detonateSatchels", "setPedOnFire", "createFire", "setVehicleHandling", "loadstring"}) then
> 	-- here comes the ban
> end
> ```

I dont get it, honestly. How did things change? You dont need to skip it to perform the ban.

You want to check if they are using that function? You are already doing that... are you talking about the benefit of skipping the function while you wait for the ban?

Dutchman101:
Let me get this straight, and correct me if im wrong, because i overslept and just woke up.

- Making debug hooks unskippable will help to avoid your server's security checks on various commonly abused MTA functions, from not being executed by hacked clients. "Avoid" = raise difficulty for basic cheaters.

- Making debug hooks unskippable will remove a detection vector (generic) to detect some cheaters by the fact alone that they skip debug hooks. This seems to be the gist of what Xenius [commented](https://github.com/multitheftauto/mtasa-blue/pull/3719#issuecomment-2351531014), that it'll return false.

I would argue that the ease to flag a cheater (by point 2) doesn't weigh up against the benefits of blocking forms of abuse that involve making many server's security checks redundant, by them skipping hooks. There are better designs of flagging cheaters by scripting.

Besides, even if you can call that ease, cheaters can just disable specific (server-AC) hooks only, so your call to check won't return false. They do this all the time, based on the server they'd like to target or make cheat menu features for, specifically.

PR #3720, opened 2024-09-13
Add new function isPlayerCrosshairVisible

### LINKED ISSUE #3711: Ability to detect when the crosshair is shown

_No response_

Hello. 
I would like to have an opportunity to detect when a player's crosshair is showing.
It could be an event like onClientPlayerCrosshairVisible & onPlayerCrosshairVisible
or a function like isPlayerCrosshairVisible() both for client & serverside. 
My problem is that you can shoot without aiming by default. This way there will be no recoil at all (no bullet spreading).
I know I can disable this, if I make a little script where I only let players to shoot, when the aiming control state is true.

There is a thing, called fast aim + fast shoot, where you press your aim, but you don't wait until your crosshair appears. This way the camera starts to zoom a bit, but the crosshair won't show up. I have a negative feeling, that this way there are also less recoil, than in normal. I also found a big advantage in this, players could switch weapons faster this way. In my server I would love to make weapon fights a bit more fair for everyone, so I would love to force all players to fully aim with crosshair, then shoot. 

Edit: for example in my server there are sawed shotgun fights. Players have maxed skills & quickreload glitch enabled so they do 2 shoots then switch back and forth. Some players found out that they can switch faster if they don't wait the crosshair to appear. Also there might be less recoil, but I'm unsure. Also if the crosshair must be shown it will be a bit easier to see if someone uses aimbot.

Sadly the aiming control state gives true even when you press right click a litte bit, also the simple gun task returns true as well.
I didn't find any other way, so thats why I decided to open this request.

(I know I could replace the basic crosshair with custom one, then make it like, so the players could only shoot if the crosshair imagie is shown, but that would need more unnecessary work)

_No response_

### Additional context

_No response_

### Security Policy

FileEX:
This PR adds the function ``isPlayerCrosshairVisible``, which allows checking if the crosshair is actually visible on the screen. Sometimes, checking if a player is aiming based on tasks can be inaccurate, as simply holding the mouse button for a second can falsely indicate that the player is aiming. This function was created to address the issue reported here, and I believe it can also help in detecting aimbots more easily. For instance, if a player is hitting headshots without even seeing the crosshair, it's a clear sign that something might be wrong with him.

If the crosshair is hidden using ``setPlayerHudComponentVisible``, the function returns false.

I have one question though: The function currently returns true for weapons like the ``spraycan`` and ``fire extinguisher``. In reality, the crosshair is intentionally hidden for these weapons in GTA’s code. Should i leave it as it is, or should i add a special condition to always return false for these two weapons?

Fixes #3711


Client/game_sa/CHudSA.cpp
```diff
@@ -171,3 +175,38 @@ void CHudSA::ResetComponentAdjustment()
     MemPut<float>(m_pfCameraCrosshairScale, 192.0f);
     m_fSniperCrosshairScale = 210.0f;
 }
+
+bool CHudSA::IsCrosshairVisible()
```
TracerDS, line 179: Const? Maybe noexcept?

Client/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp
```diff
@@ -636,3 +639,8 @@ unsigned char CLuaPlayerDefs::GetPlayerMapOpacity()
     int iMapOpacity = g_pCore->GetCVars()->GetValue<int>("mapalpha");
     return static_cast<unsigned char>(Clamp(0, iMapOpacity, 255));
 }
+
+bool CLuaPlayerDefs::IsPlayerCrosshairVisible()
```
TracerDS, line 643: If isCrosshairVisible can be noexcept, mark this lua def as noexcept too, please. Otherwise ignore this comment

Client/game_sa/CHudSA.cpp
FileEX: IsHudComponentVisible cannot be const at this moment so this function also

Dutchman101:
The below script (from me) is 100% accurate to stop "zero recoil GTA bug", by blocking fire without aiming. I've not spotted task states inconsistency with well-implemented scripts.
```lua
--- Put weapons here, set to true if you want hipfire blocked for it
-- @warning don't use 0 or 36 as they are already blocked
local antiHipfireWeapons = {
    [22] = true,
    [23] = true,
    [24] = true,
    [25] = true,
    [26] = true,
    [27] = true,
    [28] = true,
    [29] = true,
    [30] = true,
    [31] = true,
    [32] = true,
    [33] = true,
    [38] = true,
}

-- So we can't lose track of our state if some other script decides to use toggleControl
-- or when this script gets stopped we need to reset the state
-- this is set true when we're using it, false when we're not
local fireControlLock = false
-- Time since last antiHipfire check
local lastTick = 0
-- Time interval between each hipfire check
local checkInterval = 300

--- Disable hipfire for certain weapons
local function antiHipfire()
    local curTick = getTickCount()
    if curTick - lastTick > checkInterval then
        if getPedOccupiedVehicle(localPlayer) then
            return
        end

        local wp = getPedWeapon(localPlayer)

        -- Stop here if it's not anti hipfire weapon
        if not antiHipfireWeapons[wp] and fireControlLock then
            toggleControl("fire", true)
            fireControlLock = false
            return
        end

        local aiming = getPedControlState(localPlayer, "aim_weapon")
        if antiHipfireWeapons[wp] and isControlEnabled("fire") ~= aiming then
            toggleControl("fire", aiming)
            fireControlLock = not aiming
        end

        lastTick = curTick
    end
end
addEventHandler("onClientPreRender", root, antiHipfire)

-- If we stop this resource make sure we undo toggleControl
addEventHandler("onClientResourceStop", resourceRoot, function ()
    if fireControlLock then
        toggleControl("fire", true)
    end
end)
```

ghost:
The problem is that ControlState "aim_weapon" and task "TASK_SIMPLE_USE_GUN" returns true at the start of aiming when the aiming animation just started, so even with this code its possible to shoot without a fully visible crosshair. If you use a sawed, your first shoot won't have problem, but the second shoot will have no-recoil and it will be shot at head level. (if you have low fps settings, you can make this with the first shoot too)

But. Your code still fixes a problem: If you press TAB it opens the scoreboard. Hold TAB, then press right click, you will now shoot, because TAB also fires the gun. You'll have a cursor as well, because scoreboard has settings. If you  move the cursor to the center of the scoreboard and stop holding the TAB, the crosshair and scoreboard will disappear but you will be able to shoot if you press left click, or ctrl, without crosshair, so without recoil.
Your code fixes this problem :)
(I hope I described it well)

Dutchman101:
I see.

This will be nice to have in the next main rolling update..

PR #3725, opened 2024-09-15
Hide login cmd from chat history

### LINKED ISSUE #2518: /login command should not be stored in the chat history
When you use the `/login` command it will be stored in the chat history. This should not be the case.

Not register /login other sensitive commands in the chat history (open chat -> press arrow_up)

None.

**Additional context**
None.

PlatinMTA:
With this the chat history won't store sensitive information regarding the `/login` command.

Fixes #2518


FileEX:
I think this is not a good step. A much better solution would be to add a checkbox in the settings "store login data in console history" or something similar. 

When creating a server, scripts such as a login panel, depending on its level of advancement, you often need to reconnect to test it. Then entering, for example, "log" and pressing TAB is very useful to log in immediately without manually entering the login and password each time. I don't think we should completely abandon it, as it is still a useful functionality.

PlatinMTA:
> I think this is not a good step. A much better solution would be to add a checkbox in the settings "store login data in console history" or something similar.
> 
> When creating a server, scripts such as a login panel, depending on its level of advancement, you often need to reconnect to test it. Then entering, for example, "log" and pressing TAB is very useful to log in immediately without manually entering the login and password each time. I don't think we should completely abandon it, as it is still a useful functionality.

Eh I guess we could add a CVAR (for default the command is censored, otherwise is not) but if you are developing a login panel I don't get why you would want to use the /login cmd? The console does censor the command but only after the session has ended, and the chat history is not saved after the session is done so I guess thats inconsistent. I still think we shouldnt save the values eitherway.

FileEX:
> > I think this is not a good step. A much better solution would be to add a checkbox in the settings "store login data in console history" or something similar.
> > When creating a server, scripts such as a login panel, depending on its level of advancement, you often need to reconnect to test it. Then entering, for example, "log" and pressing TAB is very useful to log in immediately without manually entering the login and password each time. I don't think we should completely abandon it, as it is still a useful functionality.
> 
> Eh I guess we could add a CVAR (for default the command is censored, otherwise is not) but if you are developing a login panel I don't get why you would want to use the /login cmd? The console does censor the command but only after the session has ended, and the chat history is not saved after the session is done so I guess thats inconsistent. I still think we shouldnt save the values eitherway.

When writing a script, e.g. for the login panel, when we often do a reconnect to test it, the login command is very useful, because to see the debugscript view or restart the script, e.g. after fixing errors, we still have to log in to the ACL. The login data is stored until the MTA is closed, then there is only the command itself in the history without data. I don't know if the same thing happens when connecting to another server, but it still doesn't change the fact that this functionality is simply useful and taking it away from people overnight can only be frustrating.

botder:
Can't you bind the login command to a key for testing?

Dutchman101:
The gains outweigh that niche use (which is insecure), so the concept of this PR is fine.
I also agree, if you really want to work like that, just use a bind.

Dutchman101:
This PR caused a new crash. It'll be temporarily reverted tomorrow, if no one submitted a fix by then.

Crash on line [669 @ CChat.cpp](https://github.com/multitheftauto/mtasa-blue/blob/4639aea8a5544bfa4460bfcc8bba1d5b032e931a/Client/core/CChat.cpp#L669)

Dumptrace: [https://pastebin.com/fqBmb4hH](https://pastebin.com/fqBmb4hH)

Version = 1.6-release-23016.0.000
Time = Sun Mar  2 01:06:44 2025
Module = C:\Program Files (x86)\MTA San Andreas 1.6\MTA\core.dll
Code = 0xC0000005
Offset = 0x00006D42

EAX=0000000A  EBX=09FAD100  ECX=60A4AD88  EDX=00000001  ESI=0B6EC338
EDI=0B6EB8E8  EBP=0177F008  ESP=0177EE28  EIP=60876D42  FLG=00210246

cc @PlatinMTA

PR #3732, opened 2024-09-17
Add new event onShutdown

FileEX:
This PR adds the ``onShutdown`` event, which is triggered when the shutdown command or function is used. This event was created to allow for a **'safe shutdown.'** We have no control over the order in which scripts shut down, which can cause issues in complex gamemodes, where, for example, the script handling the database may shut down first, and then the script saving the data, leading to problems. If we have multiple scripts that rely on, say, a database script, manually adding shutdown procedures to all of them can be tedious. This event allows resources to be stopped in the correct order to avoid potential problems. A similar issue is discussed here: https://forum.multitheftauto.com/topic/82046-resource-stop-priority/

Params: ``resource, reason``


ghost (approved):

Xenius97:
> We have no control over the order in which scripts shut down

I think this should based on ``download_priority_group`` just like starting them on client, but reversed, eg. 9999 prio resource shutting down before 9998.

Fernando-A-Rocha:
Nice! What code would u write in an onShutdown handler?

FileEX:
> Nice! What code would u write in an onShutdown handler?

stopping resources in the right order to avoid problems with data saving. Read example from the link

CrosRoad95:
Test if it working well in docker environment, i imagine you shutdown container and you want graceful save players, vehicles ect.

I'm telling you this because mta may not work well with docker and novadays most big servers are using it. Mta may not send proper signal to operating system what may lead to killing process before "onShutdown" end

You probably should do something like `print("stopping")`, then few queries like  `select sleep(1000)` and `print("stopped")` to test if everything went all right

botder:
If somebody plans or runs an MTA server in Docker, they should use something like https://github.com/krallin/tini for launching the server executable for graceful and clean shutdowns.

Dutchman101:
Thanks, and please add it to wiki

PR #3741, opened 2024-09-20
2DFX Effects API

### LINKED ISSUE #3579: LODLights and extra light settings

add configurable light settings for mta like 2dfx mod.

/

_No response_

### Additional context

p2dfx_gtasaa
Project2DFX

### Security Policy

FileEX:
This PR adds the ability to add, remove, and edit 2DFX effects. It's also possible to edit and remove 2DFX effects from existing objects without the need to edit the model.

### Available effect types (relevant in MTA):
- LIGHT
- PARTICLE
- SUN_GLARE
- ROADSIGN
- ESCALATOR

Note: The ``escalator`` effect allows for the creation of moving escalators

### Elements that supports 2dfx with this PR
- Objects
- Buildings
- Vehicles
- Peds

As for limitations, the limit is the same as with normal model editing, meaning as much as your memory can handle.

> [!IMPORTANT]
> The model must be loaded at least once by the game in order to use these functions. This is because if the model has never been loaded after joining the server, there is no information (CModelInfo) about the model in memory, making it impossible to modify its effects

> [!IMPORTANT]
> Building-type entities created using createBuilding or those that are part of the map must be restreamed after adding a new particle effect (for example, use engineRestreamWorld)

> [!CAUTION]
> Stopping the script does not revert changes. If you removed an effect, you need to restore it; if you changed its properties, you need to reset them... that's exactly why resetModel2DFXEffects was created. (Example is removeWorldModel &  restoreWorldModel)

### Available functions:
```lua
bool addModel2DFX(int model, int x, int y, int z, string type, table properties) -- create new effect
bool removeModel2DFX(int model, int index) -- remove effect from model
nil restoreModel2DFX(int model, int index) -- restore removed effect
nil resetModel2DFXEffects(model) -- reset all model 2dfx effects (remove custom & reset properties for original effects)

nil setModel2DFXPosition(int model, int index, int x, int y, int z)
bool setModel2DFXProperty(int model, int index, string property, mixed value)
nil resetModel2DFXProperty(int model, int index, string property)
nil resetModel2DFXPosition(int model, int index)

float, float, float getModel2DFXPosition(int model, int index)
mixed getModel2DFXProperty(int model, int index, string property [ , bool getFlagsAsTable = false ])
table getModel2DFXEffects(int model [, bool includeCustomEffects = true] ) -- get all model 2dfx effects data
int getModel2DFXCount(int model [, bool includeCustomEffects = true ] )
string getModel2DFXType(int model, int index)
```

### Examples
```lua
addModel2DFX(1337, 0.1, 0.1, 0.8, 'particle', {name='fire'});
```
image

```lua
addModel2DFX(1337, 0.1, 0.1, 0.75, 'light', {
	drawDistance = 100,
	lightRange = 18,
	coronaSize = 2,
	shadowSize = 10,
	shadowMultiplier = 40,
	showMode = "default",
	coronaReflection = false,
	flareType = 0,
	flags = {atDay = true, atNight = true},
	shadowDistance = 0,
	offset = {0,0,0},
	color = tocolor(50, 168, 82, 255),
	coronaName = "coronamoon",
	shadowName = "shad_exp",
})
```
image

```lua
setModel2DFXProperty(1226, 0, 'shadowMultiplier', 150)
setModel2DFXProperty(1226, 0, 'color', tocolor(255, 0, 123, 255))
```
image

```lua
removeModel2DFX(3524, 0)
```
image

```lua
getModel2DFXEffects(8610)

--[[
Result

{
 [index] = data

 [0] = {
     position = {2048.5517, 1484.28, 16.88},
     type = "roadsign",
     properties = {
        color = {255, 255, 255, 255},
        flags = 0,
        rotation = {-90, 0, 180},
        size = {4.30, 2.68},
        text1 = "Julius_T_W_South",
        text2 = "Rockshore West__",
        text3 = "The_Mako_Span___",
        text4 = "___~___~____",
     },
 },
}

--]]
```

## Custom features
- Ability to set a custom text color for roadsigns
- Ability to use special symbols in roadsigns that are normally unsupported in GTA (list below)
- Ability to change the alpha of roadsign text (it works in a slightly strange way, but it allows achieving a different text appearance)

```lua
setModel2DFXProperty(8610, 0, 'color', tocolor(66, 135, 245, 255))
setModel2DFXProperty(8610, 0, 'text1', '____Hello MTA')
setModel2DFXProperty(8610, 0, 'text2', '__This is 2DFX')
setModel2DFXProperty(8610, 0, 'text3', '____FileEX_'..string.char(0x1))
setModel2DFXProperty(8610, 0, 'text4', string.char(0x4)..'_'..string.char(0xB9)..'_'..string.char(0xB6)..'_'..string.char(0xA7)..'_'..string.char(0x07)..'_'..string.char(0x02)..'_'..string.char(0xDB)..'_'..string.char(0x1E))
```

image

## Roadsign new symbols list
- ``/ , |``
- scull icon
- alien icon
- invalid icon
- parking icon
- info icon
- human icon
- winding road (?)
- train icon
- hill icon (?)

image

The 2DFX documentation will be created once the PR has been merged.
Possibly resolves: #3579


Client/mods/deathmatch/logic/CClientManager.cpp
```diff
@@ -55,6 +55,7 @@ CClientManager::CClientManager()
     m_pPacketRecorder = new CClientPacketRecorder(this);
     m_pImgManager = new CClientIMGManager(this);
     m_pBuildingManager = new CClientBuildingManager(this);
+    m_p2DFXManager = new CClient2DFXManager(this);
```
TracerDS: Maybe unique_ptr?

Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.h
```diff
@@ -511,6 +553,11 @@ inline SString GetClassTypeName(CClientVectorGraphic*)
     return "svg";
 }
 
+inline SString GetClassTypeName(CClient2DFX*)
```
TracerDS: noexcept

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2560,3 +2591,188 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+    if (effectType != e2dEffectType::LIGHT && effectType != e2dEffectType::PARTICLE && effectType != e2dEffectType::ROADSIGN &&
+        effectType != e2dEffectType::ESCALATOR && effectType != e2dEffectType::SUN_GLARE)
+        return false;
+
+    if (!CClient2DFXManager::IsValidModel(modelID))
+        throw std::invalid_argument("Invalid model ID");
+
+    const char* error = CClient2DFXManager::IsValidEffectData(effectType, effectData);
+    if (error)
+        throw std::invalid_argument(error);
```
TracerDS: Its better to throw LuaFunctionError instead of invalid argument if its runtime error

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2560,3 +2591,188 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+    if (!modelInfo)
+        return false;
+
+    auto* effect = modelInfo->Get2DFXFromIndex(index);
+    if (!effect)
+        return false;
+
+    const char* error = CClient2DFXManager::IsValidEffectData(effect->type, effectData);
+    if (error)
+        throw std::invalid_argument(error);
```
TracerDS: Above

Client/sdk/game/CModelInfo.h
```diff
@@ -247,6 +297,18 @@ class CModelInfo
     // Vehicle towing functions
     virtual bool IsTowableBy(CModelInfo* towingModel) = 0;
 
+    // 2dfx functions
+    virtual C2DEffectSAInterface* Add2DFXEffect(const CVector& position, const e2dEffectType& type) = 0;
+    virtual void                  Remove2DFX(C2DEffectSAInterface* effect, bool isCustom = false) = 0;
+    virtual bool                  Remove2DFXEffectAtIndex(std::uint32_t index, bool includeDefault = false) = 0;
+    virtual bool                  RemoveAll2DFXEffects(bool includeDefault = false) = 0;
+
+    virtual C2DEffectSAInterface* Get2DFXFromIndex(std::uint32_t index) = 0;
```
TracerDS: const?

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2560,3 +2591,188 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+}
+
+std::variant<bool, std::tuple<float, float, float>> CLuaEngineDefs::EngineGet2DFXPosition(CClient2DFX* effect)
+{
+    CVector* position = m_p2DFXManager->Get2DFXPosition(effect->Get2DFX());
+    if (!position)
+        return false;
+
+    return std::make_tuple(position->fX, position->fY, position->fZ);
+}
```
Proxy-99: use CLuaMultiReturn?

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2560,3 +2591,188 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+    if (!modelInfo)
+        return false;
+
+    auto* effect = modelInfo->Get2DFXFromIndex(index);
+    if (!effect)
+        return false;
+
+    CVector* position = m_p2DFXManager->Get2DFXPosition(effect);
+    return std::make_tuple(position->fX, position->fY, position->fZ);
+}
```
Proxy-99: use CLuaMultiReturn type?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1169,6 +1183,11 @@ void CModelInfoSA::ResetAlphaTransparency()
     }
 }
 
+void CModelInfoSA::StaticReset2DFXEffects()
+{
+
+}
```
Proxy-99: what about this?

Proxy-99 (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
FileEX: This is exactly the same, so it makes no difference https://github.com/multitheftauto/mtasa-blue/blob/81c939a4e047378d166da28dfce7dfa85bdc368d/Shared/mods/deathmatch/logic/lua/CLuaMultiReturn.h

Client/game_sa/CModelInfoSA.cpp
FileEX: This PR is marked as draft, so it's not ready for review yet

Client/mods/deathmatch/logic/CClientManager.cpp
FileEX: Not in this case, to maintain consistency with the rest of the class

Client/sdk/game/CModelInfo.h
FileEX: Const can only be GetCount because GetFromIndex changes the value of m_pInterface

FileEX:
### Update
- Added 4 new functions:
```lua
bool engineSetModel2DFXProperty(int modelID, int index, string property, mix value)
mix engineGetModel2DFXProperty(int modelID, int index, string property)
bool engineSet2DFXProperty(2dfx effect, string property, mix value)
mix engineGet2DFXProperty(2dfx effect, string property)
```
- Added effect removal with ``destroyElement``
- Little cleanup

But PR is still a work in progress.

Meine1:
what about a property for setting 2DFX dimension/interior ? also isn't the corona limit still 64 so something like Project2DFX's LOD lights still wouldn't be possible?

FileEX:
> what about a property for setting 2DFX dimension/interior ? 

2DFX effects do not have their own dimension and interior. They are part of the model, so they take on the same dimension and interior as the given object/building.

> also isn't the corona limit still 64 so something like Project2DFX's LOD lights still wouldn't be possible?

The current limit for visible coronas is indeed 64. I already know where and what needs to be done to increase it. However, I'm wondering whether it should be done as part of this PR or as a separate feature like a 'limit adjuster'

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -117,8 +117,8 @@ bool CClient2DFXManager::Set2DFXProperties(C2DEffectSAInterface* effect, const e
             light.offsetZ = static_cast<std::int8_t>(std::get<float>(*offsetZ));
 
             auto* color = MapFind(effectData, "color");
-            int colorValue = static_cast<int>(std::get<float>(*color));
-            light.color = RwColor{static_cast<std::uint8_t>((colorValue >> 0) & mask(8)), static_cast<std::uint8_t>((colorValue >> 8) & mask(8)), static_cast<std::uint8_t>((colorValue >> 16) & mask(8)), static_cast<std::uint8_t>((colorValue >> 24) & mask(8))};
+            unsigned long colorValue = static_cast<unsigned long>(std::get<float>(*color));
```
TracerDS: use `std::uint32_t` instead

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -416,8 +416,8 @@ bool CClient2DFXManager::Set2DFXProperty(C2DEffectSAInterface* effect, const e2d
                 {
                     if (std::holds_alternative<float>(propertyValue))
                     {
-                        int colorValue = static_cast<int>(std::get<float>(propertyValue));
-                        light.color = RwColor{static_cast<std::uint8_t>((colorValue >> 0) & mask(8)), static_cast<std::uint8_t>((colorValue >> 8) & mask(8)), static_cast<std::uint8_t>((colorValue >> 16) & mask(8)), static_cast<std::uint8_t>((colorValue >> 24) & mask(8))};
+                        unsigned long colorValue = static_cast<unsigned long>(std::get<float>(propertyValue));
```
TracerDS: above

TracerDS (changes requested):

ghost:
Is it possible to apply effect to certain MTA object, just like targetElement when applying shader?

Client/sdk/game/CModelInfo.h
```diff
@@ -84,6 +85,102 @@ enum eVehicleUpgradePosn
     VEHICLE_UPGRADE_POSN_MISC,
 };
 
+enum class e2dEffectType : std::uint8_t
```
TracerDS: `: std::uint8_t` is not needed

TheNormalnij:
Can you create a new 2DFX class for this instead of using engine class defs?

Meine1:
>I already know where and what needs to be done to increase it. However, I'm wondering whether it should be done as part of this PR or as a separate feature like a 'limit adjuster'

if the corona limit fix is not too complex, it would be a good idea to make it part of this because a full limit adjuster with more features would be another whole beast.

FileEX:
> Is it possible to apply effect to certain MTA object, just like targetElement when applying shader?

~~If I'm not mistaken, each entity has its own RwObject, so it should be possible to add an effect only for one object, and not for all with this id~~

Unfortunately not, because although each entity has its own instance of RwObject, all instances share the same geometry — the RpGeometry pointer is the same for every object with a given ID. 2DFX effects are stored in the geometry plugin, so unfortunately it's not possible. It would require changing the logic of entity creation in GTA so that each entity would have its own geometry (which would also require more memory)

FileEX:
### Update
Added full support for ``roadsign`` 2dfx.
- We can create our own roadsigns and edit existing ones.

Screenshot_11
Screenshot_12

By default, GTA only allows to set the colors white, black, gray and red. However, I will try to add the ability to set custom color

delusionalismHQ:
can we make changes to the grass?

FileEX:
> can we make changes to the grass?

What do you mean?

Client/game_sa/C2DEffectsSA.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        if ((*it)->GetInterface() == effectInterface)
+        {
+            (*it)->Destroy();
+
+            delete *it;
+            m_effects.erase(it);
+            break;
+        }
+    }
```
TracerDS: What do you think about algorithm functions?
```cpp
m_effects.erase(
    std::remove_if(m_effects.begin(), m_effects.end(),
        [effectInterface](auto* effect)
        {
            if (effect->GetInterface() != effectInterface)
                return false;

            effect->Destroy();
            delete effect;
            return true;
        }),
    m_effects.end());
```

Client/game_sa/CModelInfoSA.h
```diff
@@ -158,7 +162,7 @@ class CBaseModelInfoSAInterface
     unsigned char  ucAlpha : 8;                         // +12
 
     unsigned char  ucNumOf2DEffects : 8;            // +13
-    unsigned short usUnknown : 16;                  // +14     Something with 2d effects
+    short          s2DEffectIndex : 16;            // +14
```
TracerDS: `: 16` doesnt make much sense here

Client/game_sa/gamesa_renderware.h
```diff
@@ -105,6 +105,8 @@ typedef RpHAnimHierarchy*(__cdecl* GetAnimHierarchyFromSkinClump_t)(RpClump*);
 typedef int(__cdecl* RpHAnimIDGetIndex_t)(RpHAnimHierarchy*, int);
 typedef RwMatrix*(__cdecl* RpHAnimHierarchyGetMatrixArray_t)(RpHAnimHierarchy*);
 typedef RtQuat*(__cdecl* RtQuatRotate_t)(RtQuat* quat, const RwV3d* axis, float angle, RwOpCombineType combineOp);
+typedef RwTexture*(__cdecl* RwReadTexture_t)(const char* name, const char* mask);
+typedef RwFrame*(__cdecl* RwFrameRotate_t)(RwFrame* frame, const RwV3d* axis, float angle, RwOpCombineType combine);
```
TracerDS, line 109: cant be `using` due to the amount of typedefs, right?

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+}
+
+void CClient2DFXManager::RemoveAll()
+{
+    // Make sure they don't remove themselves from our list
+    m_canRemoveFromList = false;
+
+    // Run through our list deleting the 2DFX's
+    for (auto* effect : m_effectsList)
+        delete effect;
```
TracerDS: dont you also need to remove the elements from the list?

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+                        {
+                            effect->SetShadowTexture(std::get<std::string>(propertyValue));
+                            return true;
+                        }
+                    }
+
+                    break;
+                }
+                default:
+                    break;
```
TracerDS: This is unnecessary

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+            MapSet(properties, "topZ", top.z);
+            MapSet(properties, "endX", end.x);
+            MapSet(properties, "endY", end.y);
+            MapSet(properties, "endZ", end.z);
+            MapSet(properties, "direction", static_cast<float>(effect->GetEscalatorDirection()));
+
+            break;
+        }
+        default:
+            break;
```
TracerDS, line 155: This is unnecessary

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+                    if (std::holds_alternative<std::string>(propertyValue))
+                    {
+                        effect->SetRoadsignText(std::get<std::string>(propertyValue), 4);
+                        return true;
+                    }
+
+                    break;
+                }
+                default:
+                    break;
```
TracerDS: This is unnecessary

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+                    }
+
+                    break;
+                }
+            }
+
+            break;
+        }
+        default:
+            break;
```
TracerDS, line 237: This is unnecessary

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+                case e2dEffectProperty::END_Z:
+                    return effect->GetEscalatorEnd().z;
+                case e2dEffectProperty::DIRECTION:
+                    return static_cast<float>(effect->GetEscalatorDirection());
+            }
+
+            break;
+        }
+        default:
+            break;
```
TracerDS, line 371: This is unnecessary

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
```diff
@@ -0,0 +1,1007 @@
...
+            if ((*it)->GetModelID() == modelID)
+            {
+                delete *it;
+                it = m_effectsList.erase(it);
+
+                removed = true;
+            }
+            else
+                ++it;
+        }
```
TracerDS:
```cpp
m_effectsList.erase(
    std::remove_if(m_effectsList.begin(), m_effectsList.end(),
        [modelID, &removed](auto* effect)
        {
            if (effect->GetModelID() != modelID)
                return false;

            delete effect;
            removed = true;
            return true;
        }),
    m_effectsList.end());
```

Client/mods/deathmatch/logic/luadefs/CLua2DFXDefs.cpp
```diff
@@ -0,0 +1,341 @@
...
+    if (effectType != e2dEffectType::LIGHT && effectType != e2dEffectType::PARTICLE && effectType != e2dEffectType::ROADSIGN &&
+        effectType != e2dEffectType::ESCALATOR && effectType != e2dEffectType::SUN_GLARE)
+        return false;
+
+    if (!CClient2DFXManager::IsValidModel(modelID))
+        throw std::invalid_argument("Invalid model ID");
+
+    const char* error = CClient2DFXManager::IsValidEffectData(effectType, effectData);
+    if (error)
+        throw LuaFunctionError(error);
```
TracerDS, line 56: this will throw a warning. Is this intended?

Client/mods/deathmatch/logic/luadefs/CLua2DFXDefs.cpp
```diff
@@ -0,0 +1,341 @@
...
+    if (!modelInfo)
+        return false;
+
+    auto* effect = modelInfo->Get2DFXFromIndex(index);
+    if (!effect)
+        return false;
+
+    const char* error = CClient2DFXManager::IsValidEffectData(reinterpret_cast<C2DEffect*>(effect)->GetEffectType(), effectData);
+    if (error)
+        throw LuaFunctionError(error);
```
TracerDS: above

Client/mods/deathmatch/logic/luadefs/CLua2DFXDefs.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+bool CLua2DFXDefs::Set2DFXProperties(CClient2DFX* effect, effectDataMap effectData)
+{
+    CModelInfo* modelInfo = g_pGame->GetModelInfo(effect->GetModelID());
+    if (!modelInfo)
+        return false;
+
+    const char* error = CClient2DFXManager::IsValidEffectData(effect->Get2DFXType(), effectData);
+    if (error)
+        throw LuaFunctionError(error);
```
TracerDS: above

Client/sdk/game/C2DEffect.h
```diff
@@ -0,0 +1,95 @@
...
+struct RwV3d;
+class CVector;
+
+class C2DEffect
+{
+public:
+    virtual void SetPosition(const CVector& position) = 0;
+    virtual CVector& GetPosition() = 0;
+
+    virtual e2dEffectType GetEffectType() = 0;
```
TracerDS: maybe const?

Client/sdk/game/CModelInfo.h
```diff
@@ -247,6 +248,25 @@ class CModelInfo
     // Vehicle towing functions
     virtual bool IsTowableBy(CModelInfo* towingModel) = 0;
 
+    // 2dfx functions
+    virtual C2DEffectSAInterface* Add2DFXEffect(const CVector& position, const e2dEffectType& type) = 0;
+    virtual bool                  Remove2DFX(C2DEffectSAInterface* effect, bool includeDefault) = 0;
+    virtual bool                  Remove2DFXEffectAtIndex(std::uint32_t index, bool includeDefault = false) = 0;
+    virtual bool                  RemoveAll2DFXEffects(bool includeDefault = false) = 0;
+
+    virtual C2DEffectSA* Get2DFXFromIndex(std::uint32_t index) = 0;
```
TracerDS: maybe const?

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLua2DFXDefs.cpp
FileEX: Yes

Client/game_sa/CModelInfoSA.h
FileEX: I prefer not to touch it. I don't want to go beyond the purpose of this PR

Client/game_sa/C2DEffectsSA.cpp
FileEX: Personally, I find this less readable than a simple loop

Client/game_sa/gamesa_renderware.h
FileEX: The entire file uses typedef, so it's better to leave it as is

Client/sdk/game/CModelInfo.h
FileEX: You're wrong. This is needed because without it the enum takes up 4 bytes, causing the C2DEffectSAInterface class to have the wrong size and offsets

Client/mods/deathmatch/logic/CClient2DFXManager.cpp
FileEX: This function is only called in the destructor, so it is not necessary. Other managers don't have this either

Client/sdk/game/CModelInfo.h
FileEX: GetFromIndex changes the value of m_pInterface (same as GetInterface function)

Client/sdk/game/CModelInfo.h
TracerDS: Right, by default enum class inherits from `int`. My bad

delusionalismHQ:
> > can we make changes to the grass?
> 
> What do you mean?

mta-screen_2024-10-09_22-24-10
i mean plant fx's.

FileEX:
Grass/plants are not a 2DFX effect. Grass is generated by a RW, and it's a completely different mechanism. Maybe, if there’s interest, I’ll create an API for plants as well, so that custom ones can be made, etc. However, I don't think this is something that the community desires

lopezloo:
> Grass/plants are not a 2DFX effect. Grass is generated by a RW, and it's a completely different mechanism. Maybe, if there’s interest, I’ll create an API for plants as well, so that custom ones can be made, etc. However, I don't think this is something that the community desires

Increasing grass limits would be nice.

gta191977649:
wow, that would be really helpful, especially working on some VC-based map to SA.

delusionalismHQ:
any update?

FileEX:
> any update?

I need to think about this because when an object streams in, all its effects reset to default since the object is reloaded from the .dff file. I solved this with hooks, but it causes random crashes when using functions like engineRestreamWorld. I already have a potential solution that is 99% likely to work correctly, but at the moment, I have several other PRs in progress. I'll get back to this once I finish what I've already started

FileEX:
# Finally update
### First post has been updated!

Well, after half a year of inactivity on this PR, I finally managed to bring it to a finished state. I’ve been working on it over the past three weeks, during which I encountered several bugs related to the parser, but fortunately, thanks to collaboration (special thanks to @tederis  and @sbx320 ), we were able to patch them along the way. All the code was written from scratch, completely new. Over this half-year, I gained a lot of new knowledge and experience, and now I can confidently say that the previous code was not of good quality and was causing crashes. The current version is free of any crashes, and based on my tests, I haven’t found any bugs. I have tested all functions, a lot of variants, and all effect types, including escalator.

I can see how much interest this PR has generated, and that gave me the motivation to finish it instead of abandoning it for years.

In my opinion, the PR is ready for review, and I’m eagerly looking forward to its merge, so I can start preparing full documentation related to the functions, 2DFX, flags, properties, etc. I just want to mention that the video "2DFX Explain" by Tut contains a mistake when it claims that some flags don't work in MTA — all flags that work in SP also work in MTA, although some require specific conditions. Everything will be explained in the documentation.

And for now… I'm waiting for the review and hopefully the merge.

Client/game_sa/C2DEffectsSA.cpp
```diff
@@ -0,0 +1,1137 @@
...
+                default:
+                    break;
+            }
+
+            break;
+        }
+        case e2dEffectType::PARTICLE:
+        {
+            if (property == e2dEffectProperty::PRT_NAME)
+                std::strncpy(effect->effect.particle.szName, value.c_str(), 24);
```
sbx320, line 461: this (and further strncpy calls below) do not result in a null-terminated string for the name if value is more than 24 characters. This will cause trouble later on.

sbx320 (changes requested):

PlatinMTA:
Hi, looking at the API, you cannot add 2DFX effects without affecting the models right? For example, adding 2DFX text in the world without it being attached to an element or model.

Furthermore, you also can't edit just the 2DFX effects of a single element instead of editing the whole thing, right?

I still think this is cool AF... but I wouldn't find an use for it unless i can use the effects standalone or per element... but I could see other scripters using it. Thanks for your work!

FileEX:
> Hi, looking at the API, you cannot add 2DFX effects without affecting the models right? For example, adding 2DFX text in the world without it being attached to an element or model.
> 
> Furthermore, you also can't edit just the 2DFX effects of a single element instead of editing the whole thing, right?
> 
> I still think this is cool AF... but I wouldn't find an use for it unless i can use the effects standalone or per element... but I could see other scripters using it. Thanks for your work!

This is how the 2DFX plugin works in GTA SA and how it was implemented. It stems from the nature of the RenderWare engine itself. The 2DFX plugin is added to the geometry of a model. When GTA loads a model, it creates an instance of a specific model - in particular, an instance of RwObject and each instance uses the same pointer to the RpGeometry. To save memory, GTA loads the geometry only once and shares a single instance of that geometry across all model/object instances. As a result, the 2DFX effect is also allocated only once along with the geometry.

This does not mean, however, that adding support for per-element 2DFX effects is impossible but it would only be feasible for elements created by MTA through internal processing of such effects, for example with CClientObject.

Client/game_sa/C2DEffectsSA.cpp
FileEX: Would this solution be correct?

```cpp
effect->effect.particle.szName[23] = '\0';
```

43X01Y05KY9K1S42088X:
Up

tederis:
@FileEX Please, resolve the conflicts.

gta191977649:
Any update on this pr

FileEX:
> Any update on this pr

As long as the feature freeze is ongoing, I don’t see the point in continuing to do anything, because afterward there’s just a ton of conflicts to resolve

gta191977649:
> > Any update on this pr
> 
> 
> 
> As long as the feature freeze is ongoing, I don’t see the point in continuing to do anything, because afterward there’s just a ton of conflicts to resolve

What was the feature freezing?

PR #3745, opened 2024-09-23
Add C++20 support for Deathmatch project

TracerDS:


G-Moris:
Doubtful, because of linux

TracerDS:
> Doubtful, because of linux

it compiled fine on linux, what are you on about?

G-Moris:
> > Doubtful, because of linux
> 
> it compiled fine on linux, what are you on about?

It is possible that problems may arise in other linuxs, it may have already been solved, it is possible that this is so

TracerDS:
> > > Doubtful, because of linux
> > 
> > 
> > it compiled fine on linux, what are you on about?
> 
> It is possible that problems may arise in other linuxs, it may have already been solved, it is possible that this is so

If it compiles fine on x86_64-gcc-10 and arm-gcc-10 then it should compile for other modern compilers

TheNormalnij:
> If it compiles fine on x86_64-gcc-10 and arm-gcc-10 then it should compile for other modern compilers

It doesn't work like this

TracerDS:
> > If it compiles fine on x86_64-gcc-10 and arm-gcc-10 then it should compile for other modern compilers
> 
> It doesn't work like this

I know, but if it matches the standard then it should work on most compilers that conform to the C++ standard.
There arent any major runtime changes (beside a fix). It should not have any implications

TheNormalnij:
> I know, but if it matches the standard then it should work on most compilers that conform to the C++ standard.
> There arent any major runtime changes (beside a fix). It should not have any implications

All compilers have different standard realization + specific bugs

TracerDS:
> > I know, but if it matches the standard then it should work on most compilers that conform to the C++ standard.
> > There arent any major runtime changes (beside a fix). It should not have any implications
> 
> All compilers have different standard realization + specific bugs

As for this, none of the supported compilers (github actions) present a bug so we should be fine.
The only thing that could potentially be troublesome is the `[[maybe_unused]]` but even that is supported by most clang and gcc compilers.
Nothing to worry about.

I'd also assume it was thoroughly tested before rolling out C++23 version 😛

TheNormalnij:
> As for this, none of the supported compilers (github actions) present a bug so we should be fine.
> The only thing that could potentially be troublesome is the `[[maybe_unused]]` but even that is supported by most clang and gcc compilers.
> Nothing to worry about.

You forgot about real-time bugs

TracerDS:
> You forgot about real-time bugs

As I said before, the only real implication would be the change from `__attribute__((unused))` to `[[maybe_unused]]`.
Every other change is just an explicit SString call

Shared/sdk/SString.h
```diff
@@ -30,6 +30,7 @@ class SString : public std::string
 
     SString(const char* szText) : std::string(szText ? szText : "") {}
 
+    template <int=0>
```
tederis, line 35: What is the purpose of it?

Server/mods/deathmatch/logic/CPerfStat.PlayerPacketUsage.cpp
```diff
@@ -408,11 +408,11 @@ void CPerfStatPlayerPacketUsageImpl::OutputTimeSpanBlock(CPerfStatResult* pResul
             const CTimeSpan* p = timeSpanList[i];
             const CTopValue& value = (p->prev.data[0].uiPktsPerSec > p->acc.data[0].uiPktsPerSec) ? p->prev.data[iTopPos] : p->acc.data[iTopPos];
 
-            row[c++] = value.uiPktsPerSec ? SString("%d", value.uiPktsPerSec) : "-";
+            row[c++] = value.uiPktsPerSec ? std::to_string(value.uiPktsPerSec) : "-";
             if (bDetail)
             {
-                row[c++] = value.uiBytesInPerSec ? SString("%d", value.uiBytesInPerSec) : "-";
-                row[c++] = value.uiBytesOutPerSec ? SString("%d", value.uiBytesOutPerSec) : "-";
+                row[c++] = value.uiBytesInPerSec ? std::to_string(value.uiBytesInPerSec) : "-";
```
tederis, line 414: I see you want to use a feature from C++11 and that's great. But in this case we have an unwanted string copying. But you can combine C++11 with SString by adding a move constructor to it. The same for above and below.

TracerDS: So `SString(std::to_string(...))`?

Shared/sdk/SString.h
TracerDS: This template hack will force C++ to properly recognize varargs and string constructors

Server/mods/deathmatch/logic/CPerfStat.PlayerPacketUsage.cpp
tederis: In theory, you can leave it as is. But you need to modify `SString` and add a move constructor. Like 
`SString(std::string&& strText) : std::string(std::move(strText)) {}`

Shared/sdk/SString.h
tederis: Please, add a comment describing it.

tederis:
This PR is highly desirable. At least because another one PR (#3287) depends on it. I think C++20 in this module will be fine and does not cause any problems. I already have an experience with C++20 in MTA and everything is going well.

P.S. In theory there can be ABI inconsistency if some of STL structures get returned from Deathmatch module. But these potential points of inconsistency are easy to detect in most cases. At least, it can be certain that MSVC is stable enough.

G-Moris:
Hello! Is there a problem with the support of other C++20 server modules?

TracerDS:
> Hello! Is there a problem with the support of other C++20 server modules?

With some, probably. But I didnt check. Lets just get C++20 support for the deathmatch first

Server/mods/deathmatch/logic/CConsoleCommands.cpp
```diff
@@ -1657,8 +1657,8 @@ bool CConsoleCommands::AuthorizeSerial(CConsole* pConsole, const char* szArgumen
 
     std::vector<SString> parts;
     SStringX(szArguments).Split(" ", parts);
-    const SString& strAccountName = parts.size() > 0 ? parts[0] : "";
-    const SString& strAction = parts.size() > 1 ? parts[1] : "";
+    const SString& strAccountName = parts.size() > 0 ? parts[0] : SString{};
+    const SString& strAction = parts.size() > 1 ? parts[1] : SString{};
```
Pirulax, line 1661: Really should either use `string_view` or `SString*` here instead.

vendor/sparsehash/src/sparsehash/internal/hashtable-common.h
```diff
@@ -51,7 +51,7 @@ _START_GOOGLE_NAMESPACE_
 
 template <bool> struct SparsehashCompileAssert { };
 #define SPARSEHASH_COMPILE_ASSERT(expr, msg) \
-  __attribute__((unused)) typedef SparsehashCompileAssert<(bool(expr))> msg[bool(expr) ? 1 : -1]
+  [[maybe_unused]] typedef SparsehashCompileAssert<(bool(expr))> msg[bool(expr) ? 1 : -1]
```
Pirulax, line 54: not sure if editing vendor is a good idea

TracerDS: In this case its fine

Server/mods/deathmatch/logic/CConsoleCommands.cpp
TracerDS: i.e. `const std::string_view/SString* strXYZ = ... : ""/nullptr`?

TracerDS:

PR #3450, opened 2024-06-09
Object respawn functions

FileEX:
Added ``respawnObject`` and ``toggleObjectRespawn`` to the server-side.
Added new function ``isObjectRespawnable`` as shared function.

This PR is part of #3405

PR #3754, opened 2024-09-28
Make objects respawnable by default

TheNormalnij:
Fixes missing objects and collisions on serverside objects after a player damage them.


ghost (approved):
LGTM

Dutchman101:
Thanks for the quick investigation and fix

The bug is described as follows:
a1
It can make custom maps collisionless/players will fall thru the ground. Workaround isn't neccesary after the upcoming nightly with fix included. **Server** nightly (and the one in full installer), as only updated servers were affected.


Was caused by https://github.com/multitheftauto/mtasa-blue/commit/9d65bb673c4df16def27e97a4af74d3b0c7eedc9 (PR #3450)

Issue #2104, opened 2021-03-03
Add banner, server description and tags for servers in the 'Server Browser'


PlatinMTA:
**Is your feature request related to a problem? Please describe.**
It's kinda annoying having to add tags to your server name and kinda ruins the experience of searching new servers. From the sv browser, almost every server looks the same so you kinda need to go jumping from server to server in order to find something good and remarkable, or even something of your liking. Servers that say "RP" maybe are not really RP. Freeroam servers maybe are just default freeroams. Race servers maybe have DD and you don't really like DD and you could keep piling up.

#759 is related.

**Describe the solution you'd like**
A tag system (like implemented in the GoldSrc and Source games) could help reduce the issue of long names with a lot of tags for the server, giving much needed clarity in the server browser.

The server description would also help new users know more about the server they are about to join. Server owners could explain they intentions with the server more, give links to helpful websites or community rules and so on.

Lastly a banner would be awesome to have in order to display your community logo. This image should be fetch from a website outside of MTA's control, so maybe it's not worth it.

Also a languague option could help users find international communities or local ones.

All of this information could appear on the top of the player list.

![image](https://user-images.githubusercontent.com/23291573/109756169-2934cf80-7bc6-11eb-88bc-5ab470616329.png)

**Describe alternatives you've considered**
The banner and description are pretty much optional and not really needed to help improve the browsing server experience, but a languague and tag option is actually needed.

Disinterpreter:
Examples from RAGE MP, another MP.
![image](https://user-images.githubusercontent.com/3816155/109765358-c9f6af80-7c05-11eb-8c2c-973d12b92715.png)
![image](https://user-images.githubusercontent.com/3816155/109765476-f6aac700-7c05-11eb-8b5d-39be3f333ca1.png)

turret001:
yes please, I've been wanting to suggest this for a while glad someone took their time to do it.

thalentdev:
> yes please, I've been wanting to suggest this for a while glad someone took their time to do it.

I'm happy for you.

![turret](https://user-images.githubusercontent.com/80006505/110216661-5c76a780-7ec1-11eb-9bef-82e9fc221173.png)

Fernando-A-Rocha:
yes please. not only would it give MTA server browser a more modern feel, tags, languages and description is very useful and impacting

Fernando-A-Rocha:
[...]

Issue #998, opened 2019-06-15
Revamp of the server browser


Woovie:
**Is your feature request related to a problem? Please describe.**
The server browser is unnecessarily complex and somewhat confusing when attempting to search.

**Describe the solution you'd like**
The current address bar should be turned into an omnibox field. It should search player names, search server names, search gamemodes, and be an address bar all at once.

Implementation of this feature would require significant changes to the overall server browser. We should consider altering how various click events occur.

* Single clicking a server should do no adjustments to the omnibox
* Double clicking should "navigate" to a server, aka connect, and change the omnibox
* Addresses and searches should be stored
  * Whenever a search is done followed by a connection, store the search and then the connection in the history
  * Maybe also store searches if enter is pressed
* Add a history clear for servers and search results
* Whenever searching for a word that matches players and game modes as well, we should do sections within the scroll area which are for each match type (server, player, gamemodes)

CrosRoad95:
don't forget about searching by countries

Woovie:
Country filtering would be done by geoip lookup of the server. Otherwise you can just search the name of the server. I'm not sure we want to add geoip. I guess it could be a column. We could do it like the `admin` resource does.

qaisjp:
Worth noting that server location != _"server country"_, and both are equally useful to have around for latency/language purposes.

sbx320:
I think the ideal option would be to allow servers to specify a list of supported languages. That way multilingual servers would also be possible. 

I'm not sure if searching by server location is worth implementing. Whether the server is located in Amsterdam or Frankfurt has no significant impact for most European players. I think just keeping ping should be enough here. 

> It should search player names

That would require a massive rework of the masterlist. Currently player names are requested individually from each server once you click on it. To be able to search for player names, we'd need to get all players from all servers before being able to search.

Woovie:
That's fair, and I agree about server location. Language "support" would be nice. Perhaps an item that is just a text row, but a dropdown if multiple supported.

Haxardous:
#759

Woovie:
Going to start documenting this issue a bit. Here's some notes:

* Search should not kick in unless the field is at least 2 characters
* Search should occur on another thread. CAsyncTaskScheduler? (Pirulax said this scheduler, I personally am not familiar enough with the codebase to say for certain that is correct, but the naming convention seems correct from my experience.)
[...]

Issue #759, opened 2019-01-05
Language filter/search by language


Haxardous:
**Is your feature request related to a problem?**

Whenever someone wishes to start looking for a server to play on, the language issue comes within the first place in this situation, I just can't find the server what am looking for. in the following image you can see that I've searched for "Roleplay" in **English** not any other language, in a result "Turkey, Hungary, Arabic, Czech, Russian, Latvian & Polski, that's not what I've asked for or that's not i wish to see in result, a lot players join random servers that had a huge amount of resources to download. as they thought that it's an English server.

![image](https://user-images.githubusercontent.com/923242/50730060-51d2f300-113c-11e9-9385-8592be2a4eb9.png)

**Describe the solution you'd like**

Solution.1 Adding a tab within the tabs nearby the server version. by forcing the owner to select at least a single language.
Solution.2 Adding a shortcut to add flag emoji within the server name. the same situation goes with solution 1, forcing the owner to select at least a single language.

_1_
![image](https://user-images.githubusercontent.com/923242/50730068-75963900-113c-11e9-9138-c7a3083f308e.png)

_2_
![image](https://user-images.githubusercontent.com/923242/50730065-6616f000-113c-11e9-969e-f5d28aea725e.png)

Note: I used the flags that are on "MTA San Andreas 1.5\MTA\locale".
"likely those ideas aren't used before in multiplayer games that use the same way of community servers or it's rarely used often."

**Describe alternatives you've considered**
I see there are no alternatives, but for servers can use the tags as they use it now.

![image](https://user-images.githubusercontent.com/923242/50730063-5eefe200-113c-11e9-8f9b-f0c73e4e65ff.png)

patrikjuvonen:
There is/was an issue related to this in the past here. Does anyone remember which one? It had a long discussion about improving languages and tags for servers on server browser...

Haxardous:
> There is/was an issue related to this in the past here. Does anyone remember which one? It had a long discussion about improving languages and tags for servers on server browser...

I've been searching for it before posting this, it seems that it doesn't exists anymore. I think it was already existing in the older [mantis](https://bugs.mtasa.com/view_all_bug_page.php?filter=5c3171a632fdc)?

qaisjp:
There's _some_ discussion here https://bugs.mtasa.com/view.php?id=9121 but I believe there's another thread as well.

Fernando-A-Rocha:
[...]

PR #3761, opened 2024-10-01
Add new mtaserver.conf rules for server browser/list

Fernando-A-Rocha:
In preparation of #759, #998 and #2104

MTA servers use [ASE protocol](https://int64.org/docs/gamestat-protocols/ase.html) to respond with server information (name, player count, max players, server rules, etc).
In particular, server rules are key-value pairs that can be set using [setRuleValue](https://wiki.multitheftauto.com/wiki/SetRuleValue). These are already provided in servers' ASE responses.

https://int64.org/docs/gamestat-protocols/ase.html
image

My idea was to allow definition of key-value rule pairs in the mtaserver.conf file.
- [X] I have added the follow `<rule>` entries:

```xml
<!-- These rules are optional parameters for the server browser/list (they can be left empty).
     Each rule has their own requirements. Read carefully. -->

<!-- A brief description of the server. -->
<rule name="description" value="" />

<!-- A comma separated list of languages that the server supports (e.g. en). -->
<rule name="languages" value="" />

<!-- A comma separated list of tags that describe the server (e.g. freeroam,roleplay). -->
<rule name="tags" value="" />

<!-- The URL of the server's website. -->
<rule name="website_url" value="" />

<!-- A social media URL #1 (e.g. Discord). -->
<rule name="social1_url" value="" />

<!-- A social media URL #2 (e.g. YouTube). -->
<rule name="social2_url" value="" />

<!-- A social media URL #3 (e.g. Facebook). -->
<rule name="social3_url" value="" />
```

- [X] I also made the server load these <rule/> entries from the mtaserver.conf, setting their values on startup in the ASE class. It doesn't validate any of the data (as setRuleValue doesn't), so server developers are free to define the rules they want.

Then the **server browser should read these values** and act accordingly, thanks to a **UI revamp (FUTURE - Another PR)**.

**Pros of this method**:
- Doesn't require changing existing supported ASE protocol used by many services external to MTA
- Server rules can already be altered using Lua functions
- The main worry will be modifying the server browser (UI) to interpret these new values


Fernando-A-Rocha:
As I showed on Discord, example output of a server's ASE using https://github.com/opengsq/opengsq-python that @botder suggested :

```console
opengsq ase --host 127.0.0.1 --port 22126
```

```json
{"game_name": "mta", "game_port": 22003, "hostname": "Default MTA Server", "game_type": "play", "map": "None", "version": "1.6", "password": false, "num_players": 0, "max_players": 32, "rules": {"languages": "en", "tags": "roleplay,rp,serious", "logo_url": "https://www.example.com/logo.png"}, "players": []}
```

We can set any key-value pair information, and work with this.

CrosRoad95:
Add social media?

    <rule name="discord" value="discord.gg/something" />
    <rule name="facebook" value="facebook.com/something" />

url should be a resource on this server only, therefore you put `foo/logo.png` it should try to download it from server http port., from resource "foo" file "logo.png".

If possible, add size of potentially downloaded files.

Fernando-A-Rocha:
Social media links are a great addition!

I don't think that in a first version, we will have servers serving the media files themselves. This is plenty of work imo 😅 External urls is an alright solution, and it could accept only upload.mtasa.com for now

CrosRoad95:
Take an inspiration from nanos world server browser:
image

T-MaxWiese-T:
The number of resources and names are displayed in the FiveM server list. I wouldn't consider this useful for MTA players, but it could be used for statistics and for example resource authors could see how often their resources are used. 
But what I would find much more interesting would be the download size of the resources, that would be a gamechanger. Players could see how big the download is and then estimate how long it takes, how much disk space it would take up to see if it's worth it. This could possibly also lead to the servers paying more attention to using resources more efficiently.

Fernando-A-Rocha:
I absolutely agree with showing the download size of the resources. I imagine it's a value that's already calculated by the server, it would be a matter of announcing it to the server browser

ffsPLASMA:
> The number of resources and names are displayed in the FiveM server list. I wouldn't consider this useful for MTA players, but it could be used for statistics and for example resource authors could see how often their resources are used. But what I would find much more interesting would be the download size of the resources, that would be a gamechanger. Players could see how big the download is and then estimate how long it takes, how much disk space it would take up to see if it's worth it. This could possibly also lead to the servers paying more attention to using resources more efficiently.

Resource display would have to to be filtered by resource type though. FFS got over 20k map resources. This will be an issue if you try to display them all :mreow:

Fernando-A-Rocha:
> > The number of resources and names are displayed in the FiveM server list. I wouldn't consider this useful for MTA players, but it could be used for statistics and for example resource authors could see how often their resources are used. But what I would find much more interesting would be the download size of the resources, that would be a gamechanger. Players could see how big the download is and then estimate how long it takes, how much disk space it would take up to see if it's worth it. This could possibly also lead to the servers paying more attention to using resources more efficiently.
> 
> Resource display would have to to be filtered by resource type though. FFS got over 20k map resources. This will be an issue if you try to display them all :mreow:

Are all those 20k resources with state "loaded"? 😳

ffsPLASMA:
> > > The number of resources and names are displayed in the FiveM server list. I wouldn't consider this useful for MTA players, but it could be used for statistics and for example resource authors could see how often their resources are used. But what I would find much more interesting would be the download size of the resources, that would be a gamechanger. Players could see how big the download is and then estimate how long it takes, how much disk space it would take up to see if it's worth it. This could possibly also lead to the servers paying more attention to using resources more efficiently.
> > 
> > 
> > Resource display would have to to be filtered by resource type though. FFS got over 20k map resources. This will be an issue if you try to display them all :mreow:
> 
> Are all those 20k resources with state "loaded"? 😳

loaded, not running.

Fernando-A-Rocha:
The server probably should only make a list of the resources that are running, for such a feature that would show them in the server browser.

T-MaxWiese-T:
> The server probably should only make a list of the resources that are running, for such a feature that would show them in the server browser.

I think if you want to have this with the number of resources and names you should do it separately. I guess for the server browser it has no real relevance but only for statistical purposes.

Fernando-A-Rocha:
Added social urls, removed icon/banner image URLs for now. This is not a necessery feature yet

T-MaxWiese-T:
> Added social urls, removed icon/banner image URLs for now. This is not a necessery feature yet

Is Facebook really still relevant today?

Fernando-A-Rocha:
Haha who cares, the Facebook comment in "e.g." is just a suggestion. I added 3 social URLs that can be filled with any URLs that servers desire, MTA server browser could allow only certain domains (e.g. discord, youtube, twitch, vk.com, etc). My idea is to draw these Social URLs in a panel that displays server info when the server is selected.

Fernando-A-Rocha:
Reminder that setRuleValue gives servers the freedom of setting any values for any keys, without any validation. It can remain like that. The server browser UI will be responsible for the sanitization of data, and choosing what to display in the client.

T-MaxWiese-T:
Perhaps it should be documented how many characters may be used for each rule. I think this would be particularly relevant for "`description`" and "`tags`". So that as much data as possible arrives and so that everything is visible in the GUI. But how much space the GUI will have depends on the implementation.

Fernando-A-Rocha:
The current implementation in c++ (setRuleValue) limits key/value strings to 200 characters:
https://github.com/multitheftauto/mtasa-blue/blob/a550bbfff308eea5a41950a3342d11a55814663f/Server/mods/deathmatch/logic/ASE.h#L39
I have documented this limit in the wiki now

Fernando-A-Rocha:
I made the server load the rules defined in mtaserver.conf into the ASE, applying the same method that Lua function setRuleValue uses.

Result successful:
```console
opengsq ase --host 127.0.0.1 --port 22126
```
```console
{"game_name": "mta", "game_port": 22003, "hostname": "Default MTA Server", "game_type": "play", "map": "None", "version": "1.6n", "password": false, "num_players": 0, "max_players": 32, "rules": {"description": "This is my awesome test server", "languages": "pt-br,en", "social1_url": "https://discord.gg/mtasa", "tags": "roleplay,serious,textrp", "website_url": "https://example.com"}, "players": []}
```
The rules applied are returned in the 22126 udp result

Fernando-A-Rocha:
Accepting reviews, as I've written c++ code to make the rules load into the ASE on startup, concluding this feature IMO

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,31 @@ bool CMainConfig::Load()
         return false;
     }
 
+    // Grab rules
+    CXMLNode* pNode = nullptr;
+    unsigned int uiCurrentIndex = 0;
```
tederis: std::size_t is more suitable here

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,31 @@ bool CMainConfig::Load()
...
+    do
+    {
+        // Grab the current script node
+        pNode = m_pRootNode->FindSubNode("rule", uiCurrentIndex++);
+        if (pNode)
+        {
+            // Grab its "name" attribute
+            CXMLAttribute* pAttribute = pNode->GetAttributes().Find("name");
+            SString        strName = pAttribute ? pAttribute->GetValue() : "";
```
tederis: `SString{}` instead of `""`.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,31 @@ bool CMainConfig::Load()
...
+
+            // Grab its "value" attribute
+            pAttribute = pNode->GetAttributes().Find("value");
+            SString strValue = pAttribute ? pAttribute->GetValue() : "";
```
tederis: `SString{}` instead of `""`

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,31 @@ bool CMainConfig::Load()
...
+
+            // Ignore if name or value are empty
+            if (strName != "" && strValue != "") {
```
tederis: `!strName.empty()`

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,31 @@ bool CMainConfig::Load()
...
+                // Store the key value pair
+                m_RulesForASEMap[strName] = strValue;
```
tederis: Move semantics?

Server/mods/deathmatch/logic/CMainConfig.h
```diff
@@ -87,6 +87,7 @@ class CMainConfig : public CXMLConfig
     unsigned int                GetScriptDebugLogLevel() { return m_uiScriptDebugLogLevel; };
     const std::string&          GetAccessControlListFile() { return m_strAccessControlListFile; };
     bool                        GetSerialVerificationEnabled() { return m_bVerifySerials; };
+    std::map<SString, SString>  GetRulesForASE() { return m_RulesForASEMap; };
```
tederis: `const std::map<SString, SString>& GetRulesForASE() const noexcept { return m_RulesForASEMap; }`

tederis (changes requested):

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -962,8 +962,16 @@ bool CGame::Start(int iArgumentCount, char* szArguments[])
...
-    if (m_pMainConfig->GetSerialVerificationEnabled())
-        m_pASE->SetRuleValue("SerialVerification", "yes");
+    if (m_pASE) {
+        if (m_pMainConfig->GetSerialVerificationEnabled())
+            m_pASE->SetRuleValue("SerialVerification", "yes");
+
+        // Set the Rules loaded from config
+        std::map<SString, SString> rulesMap = m_pMainConfig->GetRulesForASE();
+        for (const auto& rule : rulesMap)
+            m_pASE->SetRuleValue(rule.first, rule.second);
```
TracerDS: Use ranged map semantics:
```cpp
for (const auto& [a, b] : rulesMap)
{
    m_pASE->SetRuleValue(a, b);
} 
```

Server/mods/deathmatch/logic/CMainConfig.h
Fernando-A-Rocha: Thanks for the reviews tederis

CrosRoad95:
Can we don't make microoptimizations that just make code harder to read? those all "move sematics" make no difference because this code executes exactly ONCE. You probably own't even be able to see difference using profiling tools and compilers are smart enough to optimize it anyway.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,29 @@ bool CMainConfig::Load()
...
+
+            // Grab its "value" attribute
+            pAttribute = pNode->GetAttributes().Find("value");
+            SString strValue = pAttribute ? pAttribute->GetValue() : SString{};
+
+            // Ignore if name or value are empty
+            if (!strName.empty() && !strValue.empty())
+                m_RulesForASEMap[std::move(strName)] = std::move(strValue);
+        }
+    } while (pNode);
```
TracerDS: `while` would be better here:
```cpp
while(pNode = m_pRootNode->FindSubNode("rule", uiCurrentIndex++))
{
    // Grab its "name" attribute
    CXMLAttribute* attribute = pNode->GetAttributes().Find("name");
    SString        strName = attribute ? attribute->GetValue() : SString{};

    // Grab its "value" attribute
    attribute = pNode->GetAttributes().Find("value");
    SString strValue = attribute ? attribute->GetValue() : SString{};

    // Ignore if name or value are empty
    if (!strName.empty() && !strValue.empty())
        m_RulesForASEMap[std::move(strName)] = std::move(strValue);
}
```

TracerDS (changes requested):

Server/mods/deathmatch/logic/CMainConfig.cpp
Lpsd: I agree, you should use `while` instead of `do ... while`, unless you need the body of the loop to execute at least once before the condition is evaluated.

Fernando-A-Rocha: Ok, for better readability

Fernando-A-Rocha:
All ok?

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,21 @@ bool CMainConfig::Load()
...
+    while (pNode = m_pRootNode->FindSubNode("rule", uiCurrentIndex++))
+    {
+        CXMLAttribute* pAttribute = pNode->GetAttributes().Find("name");
+        SString        strName = pAttribute ? pAttribute->GetValue() : SString{};
+
+        pAttribute = pNode->GetAttributes().Find("value");
+        SString strValue = pAttribute ? pAttribute->GetValue() : SString{};
+
+        if (!strName.empty() && !strValue.empty())
+            m_RulesForASEMap[std::move(strName)] = std::move(strValue);
```
TracerDS: if you can, please dont use hungarian notation

TracerDS (approved):

Server/mods/deathmatch/logic/CMainConfig.cpp
Fernando-A-Rocha: Only here? The rest of the entire file is all hungarian

TracerDS: If you could, you can make a small refactor there.

Fernando-A-Rocha: Ok I'll try my best.

Fernando-A-Rocha:
All ok

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -268,7 +283,7 @@ bool CMainConfig::Load()
             if (!bFound)
                 CLogger::ErrorPrintf("Unknown client_file '%s'\n", *strName);
         }
-    } while (pNode);
+    } while (currentNode);
```
TracerDS: you can do the same with this loop as in the case above

TracerDS (changes requested):

Server/mods/deathmatch/logic/CMainConfig.cpp
Fernando-A-Rocha: Check now

TracerDS (approved):

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -232,43 +247,37 @@ bool CMainConfig::Load()
...
+        name = name.Replace("\\", "/").ToLower();
+
+        // Grab its "verify" attribute
+        attribute = currentNode->GetAttributes().Find("verify");
+        SString verify = attribute ? attribute->GetValue() : SString{};
+        bool shouldVerify = verify == "true" || verify == "yes" || verify == "1";
+
+        // Find bitnumber
+        bool found = false;
+        for (uint i = 0; i < NUMELMS(gtaDataFiles); i++)
```
FileEX: std::uint8_t

Fernando-A-Rocha: Sir I didn't write this code 😂😂 but ok

FileEX: You edited this piece of code, so changing it to stay within the guidelines shouldn't cost you much, right?

Fernando-A-Rocha: You're right, I overreacted there

Fernando-A-Rocha:
Review requested once again as I've edited `CMainConfig::AddMissingSettings()` to support XML nodes with attributes. The script was not taking any attributes into account, not being appropriate for multiple `rule name="abc" value="abc"` entries.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -847,44 +856,87 @@ bool CMainConfig::AddMissingSettings()
...
+                    {
+                        bAttributesMatch = false;
+                        break;
+                    }
+                }
+                if (bAttributesMatch)
+                {
+                    foundNode = tempNode;
+                    break;
+                }
```
TracerDS: use early continue

TracerDS (changes requested):

Server/mods/deathmatch/logic/CMainConfig.cpp
TheNormalnij: There is no early continue practice. Leave previous variant.

TracerDS: > There is no early continue practice. Leave previous variant.

Yes there is.

TheNormalnij: > Yes there is.

Prove it

TracerDS: > > Yes there is.
> 
> Prove it

https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md#code-readability

TheNormalnij: It's  `early-returns` and not `early continue`

TracerDS: > It's `early-returns` and not `early continue`

My friend, look below that

TheNormalnij: It's your own rule, lol (added 3 weeks ago)

TracerDS: > It's your own rule, lol

Its been merged nonetheless. If you have any issues, please post them at mtasa-docs repo.

Fernando-A-Rocha: It changes almost nothing in terms of readability tbh. But ye it was approved

FileEX (approved):

TracerDS (approved):

CrosRoad95:
In this pull request you just extending what ase is returning, visual aspects in main menu to display it will be in another pull request?

Fernando-A-Rocha:
> In this pull request you just extending what ase is returning, visual aspects in main menu to display it will be in another pull request?

I don't even extend what ase is returning. This pr just adds a feature to mtaserver.conf for setting "rules" on startup (like using setRuleValue) which are settings for the server browser (old feature unused).

Visual edits to the server browser to start using these "rules" will be made later.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -847,44 +856,91 @@ bool CMainConfig::AddMissingSettings()
     if (!g_pGame->IsUsingMtaServerConf())
         return false;
 
-    SString strTemplateFilename = PathJoin(g_pServerInterface->GetServerModPath(), "mtaserver.conf.template");
+    const SString& templateFileName = PathJoin(g_pServerInterface->GetServerModPath(), "mtaserver.conf.template");
```
tederis: `PathJoin` returns `SString`, not a reference to it.

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -232,43 +247,37 @@ bool CMainConfig::Load()
...
+        name = name.Replace("\\", "/").ToLower();
+
+        // Grab its "verify" attribute
+        attribute = currentNode->GetAttributes().Find("verify");
+        SString verify = attribute ? attribute->GetValue() : SString{};
+        bool shouldVerify = verify == "true" || verify == "yes" || verify == "1";
+
+        // Find bitnumber
+        bool found = false;
+        for (std::uint8_t i = 0; i < NUMELMS(gtaDataFiles); i++)
```
tederis: std::size_t. And since you deciced to refactor the code and Deathmatch project is C++17, you can use std::size instead of NUMELMS.

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -962,10 +962,16 @@ bool CGame::Start(int iArgumentCount, char* szArguments[])
...
 
-    // If ASE is enabled
+    // Init ASE
     m_pASE = new ASE(m_pMainConfig, m_pPlayerManager, static_cast<int>(usServerPort), strServerIPList);
     if (m_pMainConfig->GetSerialVerificationEnabled())
         m_pASE->SetRuleValue("SerialVerification", "yes");
+
+    // Set the Rules loaded from config
+    std::map<SString, SString> rulesMap = m_pMainConfig->GetRulesForASE();
+    for (const auto& [key, value] : rulesMap)
```
botder: To avoid the copy of the entire std::map
```suggestion
    for (const auto& [key, value] : m_pMainConfig->GetRulesForASE())
```

Server/mods/deathmatch/local.conf
```diff
@@ -3,6 +3,28 @@
...
+    <rule name="website_url" value="" />
+
+    <!-- A social media URL #1 (e.g. Discord). -->
+    <rule name="social1_url" value="" />
+
+    <!-- A social media URL #2 (e.g. YouTube). -->
+    <rule name="social2_url" value="" />
+
+    <!-- A social media URL #3 (e.g. Facebook). -->
+    <rule name="social3_url" value="" />
```
botder: With a repeating prefix you can easily check for this like `rule.startswith("social_url_")`
```suggestion
    <!-- A social media URL #1 (e.g. Discord). -->
    <rule name="social_url_1" value="" />

    <!-- A social media URL #2 (e.g. YouTube). -->
    <rule name="social_url_2" value="" />

    <!-- A social media URL #3 (e.g. Facebook). -->
    <rule name="social_url_3" value="" />
```

Server/mods/deathmatch/editor.conf
```diff
@@ -3,6 +3,28 @@
...
+    <rule name="website_url" value="" />
+
+    <!-- A social media URL #1 (e.g. Discord). -->
+    <rule name="social1_url" value="" />
+
+    <!-- A social media URL #2 (e.g. YouTube). -->
+    <rule name="social2_url" value="" />
+
+    <!-- A social media URL #3 (e.g. Facebook). -->
+    <rule name="social3_url" value="" />
```
botder:
```suggestion
    <!-- A social media URL #1 (e.g. Discord). -->
    <rule name="social_url_1" value="" />

    <!-- A social media URL #2 (e.g. YouTube). -->
    <rule name="social_url_2" value="" />

    <!-- A social media URL #3 (e.g. Facebook). -->
    <rule name="social_url_3" value="" />
```

Server/mods/deathmatch/mtaserver.conf
```diff
@@ -3,7 +3,29 @@
...
+    <rule name="website_url" value="" />
+
+    <!-- A social media URL #1 (e.g. Discord). -->
+    <rule name="social1_url" value="" />
+
+    <!-- A social media URL #2 (e.g. YouTube). -->
+    <rule name="social2_url" value="" />
+
+    <!-- A social media URL #3 (e.g. Facebook). -->
+    <rule name="social3_url" value="" />
```
botder:
```suggestion
    <!-- The URL of the server's website. -->
    <rule name="website_url" value="" />

    <!-- A social media URL #1 (e.g. Discord). -->
    <rule name="social_url_1" value="" />

    <!-- A social media URL #2 (e.g. YouTube). -->
    <rule name="social_url_2" value="" />

    <!-- A social media URL #3 (e.g. Facebook). -->
    <rule name="social_url_3" value="" />
```

Server/mods/deathmatch/mtaserver.conf.template
```diff
@@ -4,6 +4,28 @@
...
+    <rule name="website_url" value="" />
+
+    <!-- A social media URL #1 (e.g. Discord). -->
+    <rule name="social1_url" value="" />
+
+    <!-- A social media URL #2 (e.g. YouTube). -->
+    <rule name="social2_url" value="" />
+
+    <!-- A social media URL #3 (e.g. Facebook). -->
+    <rule name="social3_url" value="" />
```
botder:
```suggestion
    <!-- A social media URL #1 (e.g. Discord). -->
    <rule name="social_url_1" value="" />

    <!-- A social media URL #2 (e.g. YouTube). -->
    <rule name="social_url_2" value="" />

    <!-- A social media URL #3 (e.g. Facebook). -->
    <rule name="social_url_3" value="" />
```

botder (changes requested):

Fernando-A-Rocha:
There's nothing more to add in this PR imo @botder

botder (approved):

PR #3762, opened 2024-10-01
Fix #3760 isPlayerCrosshairVisible returns true at the start of aiming

### LINKED ISSUE #3760: isPlayerCrosshairVisible returns true at the start of aiming
### Describe the bug

Hello.
I'm glad that this new function added to the game, but in this state its the equivalent of getControlState("aim_weapon")
The only difference is that when you run, so your charachter won't be able to aim, it won't return as true, unlike the controlState. 

I would like to make a script which would prevent players from shooting when their crosshair is not fully visible.
Without the crosshair, the players can fast aim + fast shoot to acquire no-recoil on weapons. Also these shots are shooted at head level so if a server uses instant death on headshots they can easily kill players with this.
This is a problem especially with sawed-off shotgun. On most server this weapon is the best close combat weapon, on my server 90% of the fights are made with sawed-off. The players somehow learnt a way on how to properly run in circles + do this fast aim + shoot thing, so they don't really have to aim, they just need to time their charachter's rotation where their opponent is. I am unable to show this in progress, but I can show you how the no-recoil (clearly visible with UZI) and head-level shooting works. On most cases the second shoot will be the problem, but as you can see in the video I was able to do it with all 4 bullets.

Video: https://youtu.be/nzknvCRo06w?t=23
From 0:23 Its clearly visible that the second shoot goes to head level, and has base recoil only. With crosshair the recoil/spreading would be higher especially with poor skills.
At 0:37 and 0:39 I was able to shoot the first bullet at head level & with only base recoil as well.
When I switched to UZI, you can see it has no recoil at all. 

### Steps to reproduce

```lua
--- Put weapons here, set to true if you want hipfire blocked for it
-- @warning don't use 0 or 36 as they are already blocked
local antiHipfireWeapons = {
    [22] = true,
    [23] = true,
    [24] = true,
    [25] = true,
    [26] = true,
    [27] = true,
    [28] = true,
    [29] = true,
    [30] = true,
    [31] = true,
    [32] = true,
    [33] = true,
    [38] = true,
}

-- So we can't lose track of our state if some other script decides to use toggleControl
-- or when this script gets stopped we need to reset the state
-- this is set true when we're using it, false when we're not
local fireControlLock = false
-- Time since last antiHipfire check
local lastTick = 0
-- Time interval between each hipfire check
local checkInterval = 300

--- Disable hipfire for certain weapons
local function antiHipfire()
    local curTick = getTickCount()
    if curTick - lastTick > checkInterval then
        if getPedOccupiedVehicle(localPlayer) then
            return
        end

        local wp = getPedWeapon(localPlayer)

        -- Stop here if it's not anti hipfire weapon
        if not antiHipfireWeapons[wp] and fireControlLock then
            toggleControl("fire", true)
            fireControlLock = false
            return
        end

        --local aiming = getPedControlState(localPlayer, "aim_weapon")
        local aiming = isPlayerCrosshairVisible()
        if antiHipfireWeapons[wp] and isControlEnabled("fire") ~= aiming then
            toggleControl("fire", aiming)
            fireControlLock = not aiming
        end

        lastTick = curTick
    end
end
addEventHandler("onClientPreRender", root, antiHipfire)

-- If we stop this resource make sure we undo toggleControl
addEventHandler("onClientResourceStop", resourceRoot, function ()
    if fireControlLock then
        toggleControl("fire", true)
    end
end)
```

Use this code, to test it. Made by Dutchman. 

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #3760 

I had to edit several interfaces because their structure and offsets are incorrect.


Dutchman101:
The stability of this patch will first be verified in 'nightly' channel (quite some users on it..)
The feature and subsequent fix is quite in demand now, and FileEX assured me that he went over & tested it thoroughly.

Client/game_sa/CPedSA.h
```diff
@@ -258,7 +263,8 @@ class CPedSAInterface : public CPhysicalSAInterface            // +1420  = curre
     // weapons at +1440 ends at +1804
     BYTE                bPad4[12];
     BYTE                bCurrentWeaponSlot;            // is actually here
-    BYTE                bPad6[20];
+    BYTE                bPad6[3];
```
TheNormalnij, line 266: May be `BYTE bPad6[16];`  @FileEX  ?

Client/game_sa/CCameraSA.h
```diff
@@ -82,6 +82,7 @@ class CCameraSAInterface
 public:
     // CPlaceable
     CPlaceableSAInterface Placeable;
+    std::uint8_t          specialPadding[4]; // Temporary padding due to incorrect CPlaceableSAInterface class
```
TheNormalnij, line 85: It's `void* m_vtbl` before CPlaceableSAInterface

Client/game_sa/CPedSA.h
FileEX: No, I just forgot to add padding after pTargetedObject

PR #3766, opened 2024-10-03
Fix class hierarchy. Remove VTBL hacks

TheNormalnij:
This PR:
* Removes dirty VTBL hacks in `CEntitySAInterface`. All hacks are replaced with virtual methods. 
* Fixes class hierarchy in CEntitySAInterface and CCameraSAInterface with CPlaceableSAInterface.

Current state: testing.


Client/game_sa/CCameraSA.cpp
```diff
@@ -210,7 +210,7 @@ CMatrix* CCameraSA::GetMatrix(CMatrix* matrix)
 
 void CCameraSA::SetMatrix(CMatrix* matrix)
 {
-    CMatrix_Padded* pCamMatrix = GetInterface()->Placeable.matrix;
+    CMatrix_Padded* pCamMatrix = GetInterface()->matrix;
     if (pCamMatrix)
     {
         pCamMatrix->vFront = matrix->vFront;
```
TheNormalnij, line 216: This code had no effect previously. `matrix` was always `nullptr` because `Placeable` was misplaced

Client/game_sa/CVehicleSA.cpp
```diff
@@ -165,22 +165,34 @@ void CVehicleSA::Init()
...
     if (dwOffset != 0)
     {
         for (unsigned int i = 0; i < sizeof(m_doors) / sizeof(m_doors[0]); ++i)
         {
             DWORD dwInterface = (DWORD)GetInterface();
-            DWORD dwDoorAddress = dwInterface + 1464 + i * 24;
+            DWORD dwDoorAddress = dwInterface + dwOffset + i * 24;
             m_doors[i].SetInterface((CDoorSAInterface*)dwDoorAddress);
         }
     }
```
TheNormalnij: This code was bugged. The game uses wrong door array positions for CTrain.
It causes train desync issues

FileEX:
On this topic, I think we should get rid of two duplicate classes like ``CMatrix_Padded`` and ``CMatrixSAInterface``. Only ``CMatrixSAInterface`` should remain. In addition, the CMatrix_Padded class is incomplete

TheNormalnij:
> On this topic, I think we should get rid of two duplicate classes like `CMatrix_Padded` and `CMatrixSAInterface`. Only `CMatrixSAInterface` should remain. In addition, the CMatrix_Padded class is incomplete

 `CMatrix_Padded` and `CMatrixSAInterface` are different classes. [CMatrix_Padded ](https://github.com/gta-reversed/gta-reversed-modern/blob/c91981e3aa9c3637b803f4784116b2ac5e54e2b9/source/game_sa/Core/MatrixLink.h#L13)

Client/game_sa/CBuildingsPoolSA.h
```diff
@@ -43,5 +43,8 @@ class CBuildingsPoolSA : public CBuildingsPool
     SVectorPoolData<CBuildingSA> m_buildingPool{MAX_BUILDINGS};
     CPoolSAInterface<CBuildingSAInterface>**           m_ppBuildingPoolInterface;
 
-    std::unique_ptr<std::array<std::pair<bool, CBuildingSAInterface>, MAX_BUILDINGS>> m_pOriginalBuildingsBackup;
+    typedef std::uint8_t building_buffer_t[sizeof(CBuildingSAInterface)];
+    typedef std::array<std::pair<bool, building_buffer_t>, MAX_BUILDINGS> backup_array_t;
```
TracerDS: `using` instead of `typedef`:
```cpp
using building_buffer_t = std::uint8_t[sizeof(CBuildingSAInterface)];
using backup_array_t = std::array<std::pair<bool, building_buffer_t>, MAX_BUILDINGS>;
```

Client/game_sa/CEntitySA.h
```diff
@@ -106,27 +80,33 @@ class XYZStore
...
-};
+    virtual void         Add() = 0;
+    virtual void         Add(const CRect& rect) = 0;
+    virtual void         Remove() = 0;
+    virtual void         SetIsStatic(bool isStatic) = 0;
+    virtual void         SetModelIndex(std::uint32_t model) = 0;
+    virtual void         SetModelIndexNoCreate(std::uint32_t model) = 0;
+    virtual void         CreateRwObject() = 0;
+    virtual void         DeleteRwObject() = 0;
+    virtual CRect        GetBoundRect() = 0;
```
TracerDS, line 93: `const`?

Client/game_sa/CEntitySA.h
```diff
@@ -236,16 +216,8 @@ class CEntitySAInterface
...
-    {
-        using vtbl_DeleteRwObject = void(__thiscall*)(CEntitySAInterface * pEntity);
-        ((vtbl_DeleteRwObject)this->vtbl->DeleteRwObject)(this);
-    };
-
-    bool HasMatrix() const noexcept { return Placeable.matrix != nullptr; }
-
-    void RemoveMatrix() { ((void(__thiscall*)(void*))0x54F3B0)(this); }
+    
+    bool IsUsesEntityDeleteRwObject() const { return *reinterpret_cast<void**>(*reinterpret_cast<std::uint32_t const*>(this) + 0x20) == (void*)0x00534030; };
```
TracerDS: What is `this + 0x20` supposed to be (i.e. what member)?
And what is the job of the `0x00534030` address?

FileEX:
> > On this topic, I think we should get rid of two duplicate classes like `CMatrix_Padded` and `CMatrixSAInterface`. Only `CMatrixSAInterface` should remain. In addition, the CMatrix_Padded class is incomplete
> 
> `CMatrix_Padded` and `CMatrixSAInterface` are different classes. [CMatrix_Padded ](https://github.com/gta-reversed/gta-reversed-modern/blob/c91981e3aa9c3637b803f4784116b2ac5e54e2b9/source/game_sa/Core/MatrixLink.h#L13)

You are right, so CMatrix_Padded should inherit CMatrixSAInterface, because now CMatrix_Paddes is the wrong size

Client/game_sa/CPlaceableSA.h
```diff
@@ -0,0 +1,34 @@
...
+    CVector m_translate;
+    float   m_heading;
+};
+
+class CPlaceableSAInterface
+{
+public:
+    virtual void* Destructor(bool free) = 0;
+
+    bool IsPlaceableVTBL() const { return *reinterpret_cast<const void* const*>(this) == (void*)0x863C40; }
```
TracerDS: same as above

Client/game_sa/CPlaceableSA.h
```diff
@@ -0,0 +1,34 @@
...
+    bool HasVTBL() const { return *reinterpret_cast<const void* const*>(this) != nullptr; }
```
TracerDS:
```cpp
bool HasVTBL() const noexcept { return this != nullptr; }
```
?

Client/game_sa/CProjectileSA.h
```diff
@@ -16,6 +16,8 @@
 
 class CProjectileSAInterface : public CObjectSAInterface            // entirely inherited from CObject
 {
+public:
+    bool IsProjectableVTBL() const { return *reinterpret_cast<const std::uint32_t*>(this) == 0x867030; };
```
TracerDS: above

Client/game_sa/CVehicleSA.h
```diff
@@ -258,6 +207,61 @@ class CAutoPilot
 
 class CVehicleSAInterface : public CPhysicalSAInterface
 {
+public:
+    virtual void ProcessControlCollisionCheck(bool applySpeed) = 0;
+    virtual void ProcessControlInputs(std::uint8_t playerNum) = 0;
+    virtual void GetComponentWorldPosition(std::int32_t componentId, CVector& outPos) = 0;
+    virtual bool IsComponentPresent(std::int32_t componentId) = 0;
```
TracerDS, line 229: `const`?

Client/game_sa/CVehicleSA.h
```diff
@@ -258,6 +207,61 @@ class CAutoPilot
...
+    virtual float GetDoorAngleOpenRatio_(std::uint8_t door) = 0;
+
+    virtual bool IsDoorReady(std::uint32_t door) = 0;
+    virtual bool IsDoorReady_(std::uint8_t door) = 0;
+    virtual bool IsDoorFullyOpen(std::uint32_t door) = 0;
+    virtual bool IsDoorFullyOpen_(std::uint8_t door) = 0;
+    virtual bool IsDoorClosed(std::uint32_t door) = 0;
+    virtual bool IsDoorClosed_(std::uint8_t door) = 0;
+    virtual bool IsDoorMissing(std::uint32_t door) = 0;
+    virtual bool IsDoorMissing_(std::uint8_t door) = 0;
```
TracerDS, line 243: Can be `const`?

TracerDS (changes requested):

Client/game_sa/CEntitySA.h
TheNormalnij: I can't be sure with virtual methods. They are copied from gta-reversed-modern project

TracerDS: Do the overrided methods modify class members?

Client/game_sa/CPlaceableSA.h
TheNormalnij: No, your check is different

TracerDS: ~~How is it different? You cast this pointer to the const void pointer essentially removing the type.~~

Okay, another question. Why are you dereferencing this pointer and doing this operation with it?

TheNormalnij: This code checks if first object member (VTBL in virtual classes) is null. I think it means that an object wasn't initialized.
Your check has no sense

Client/game_sa/CEntitySA.h
TheNormalnij: i can't guarantee that for all classes

Client/game_sa/CVehicleSA.h
TheNormalnij: i can't guarantee that

Client/game_sa/CEntitySA.h
TheNormalnij: The class should not have this hack, removed

Client/game_sa/CPlaceableSA.h
TheNormalnij: Checks if the class has specific VTBL

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -985,14 +985,15 @@ void CModelInfoSA::StaticFlushPendingRestreamIPL()
...
                 // Log info
                 OutputDebugString(SString("Entity 0x%08x (with model %d) at ARRAY_StreamSectors[%d,%d] is invalid\n", pEntity, pEntity->m_nModelIndex,
                                           i / 2 % NUM_StreamSectorRows, i / 2 / NUM_StreamSectorCols));
                 // Assert in debug
                 #if MTA_DEBUG
-                assert(pEntity->vtbl->DeleteRwObject == 0x00534030);
+                assert(static_cast<std::size_t*>(pEntity->GetVTBL())[CEntity_DeleteRwObject_VTBL_OFFSET] != 0x00534030);
                 #endif
                 pSectorEntry = (DWORD*)pSectorEntry[1];
                 continue;
```
TheNormalnij, line 1014: I think this check should be removed but not in this PR

Dutchman101:
@TheNormalnij Please resolve conflicts.

Would it be a good idea to just test its stability in a nightly build soon? A bunch of code reviews have passed, and there seems to be no further feedback after that.

TheNormalnij:
> @TheNormalnij Please resolve conflicts.
> 
> Would it be a good idea to just test its stability in a nightly build soon? A bunch of code reviews have passed, and there seems to be no further feedback after that.

Updated.
I'd test it in a nightly build first

Dutchman101:
Alright, in a few days we will call for testers, along with for other changes.

PR #3767, opened 2024-10-04
Another minor fix after #3760

FileEX:


Client/game_sa/CPlayerPedSA.cpp
```diff
@@ -75,7 +75,7 @@ CPlayerPedSA::CPlayerPedSA(unsigned int nModelIndex)
     // Set default stuff
     m_pData->m_bRenderWeapon = true;
     m_pData->m_Wanted = pLocalWanted;
-    m_pData->m_fSprintEnergy = 1000.0f;
+    m_pData->m_fTimeCanRun = 1000.0f;
```
TheNormalnij, line 78: Does `m_bRenderWeapon` uses wrong field too?

TheNormalnij:
Did you check all fields between `m_fTimeCanRun` and `m_bRenderWeapon` ?

Client/game_sa/CPlayerPedSA.cpp
FileEX: No, why?

FileEX:
> Did you check all fields between `m_fTimeCanRun` and `m_bRenderWeapon` ?

Yeah, it looks good

image

image

Dutchman101:
[1 comment](https://github.com/multitheftauto/mtasa-blue/pull/3767#issuecomment-2393929323) was addressed, and [1 comment](https://github.com/multitheftauto/mtasa-blue/pull/3767#discussion_r1787863166) is pending with that it may or may not be addressed. Normally we'd wait for TheNormalnij to leave final feedback, but due to team/organisatorial reasons we have to merge it now to be able to get a new build (_Build server/signing related_), or else there might not be a new nightly for who knows how long.

The purported fix is against the last known issue that stands in the way of a finally stable build, which we've been working towards for the past 2 weeks with QA & testing. We'll see as we will now build a new nightly, and test it, besides seeing what TheNormalnij has to input tomorrow/if a follow-up change is needed, or if it will confirm we'll be looking at a stable build.

Note: the whole PR is for an issue that's been discussed in dev discord, a remote player desync.

Client/game_sa/CPlayerPedSA.cpp
TheNormalnij: You changed this field offset in your first fix pr.
Now this code changes another value

TheNormalnij:
> Yeah, it looks good

Did you check how MTA uses these fields?

FileEX:
> > Yeah, it looks good
> 
> Did you check how MTA uses these fields?

I know what you mean. The thing is that now all the fields up to the "flags" field, where m_pad2 was, have moved.  However, the previous behavior was incorrect because e.g. this piece of code
```cpp
((CPlayerInfoSA*)pGame->GetPlayerInfo())->GetInterface()->PlayerPedData.m_nChosenWeapon = weaponSlot;
```

Instead of referring to the real field, chosenWeapon (0x20) actually referred to the sprintEnergy (0x1C) field, which was not present in the interface. I don't think the current behavior is incorrect, on the contrary, the previous behavior was incorrect.

image

PR #3768, opened 2024-10-04
Add template function for calling native gta functions

FileEX:
This PR adds a simple template function that is used to call native GTA functions. This is more readable than manually casting or "using".

The ``PrepareSignature`` function prepares arguments of the appropriate types, and ``CallGTAFunction`` calls the function and optionally returns the result.

The syntax is simple
```
CallGTAFunction<return type, calling convention>(function address, arguments returned by PrepareSignature)
```

Example
```cpp
    auto args = PrepareSignature(self, colModel, applyToPairedModel);
    CallGTAFunction<void, __THISCALL>(0x4C4BC0, args);
```

Calling conventions are given in capital letters.

### DIFF
--- a/Client/game_sa/CAESoundManagerSA.cpp
+++ b/Client/game_sa/CAESoundManagerSA.cpp
@@ -23,6 +23,7 @@ CAESoundManagerSA::CAESoundManagerSA(CAESoundManagerSAInterface* pInterface) : m
 
 void CAESoundManagerSA::CancelSoundsInBankSlot(uint uiGroup, uint uiIndex)
 {
-    auto args = PrepareSignature(m_pInterface, uiGroup, uiIndex);
-    CallGTAFunction<CAESound*, __THISCALL>(FUNC_CAESoundManager__CancelSoundsInBankSlot, args);
+    using CAESoundManager__CancelSoundsInBankSlot = CAESound*(__thiscall*)(CAESoundManagerSAInterface*, uint, uint);
+    static auto pCancelSoundsInBankSlot = reinterpret_cast<CAESoundManager__CancelSoundsInBankSlot>(FUNC_CAESoundManager__CancelSoundsInBankSlot);
+    pCancelSoundsInBankSlot(m_pInterface, uiGroup, uiIndex);
 }
--- a/Client/game_sa/CAESoundManagerSA.cpp
+++ b/Client/game_sa/CAESoundManagerSA.cpp
@@ -23,7 +23,6 @@ CAESoundManagerSA::CAESoundManagerSA(CAESoundManagerSAInterface* pInterface) : m
 
 void CAESoundManagerSA::CancelSoundsInBankSlot(uint uiGroup, uint uiIndex)
 {
-    using CAESoundManager__CancelSoundsInBankSlot = CAESound*(__thiscall*)(CAESoundManagerSAInterface*, uint, uint);
-    static auto pCancelSoundsInBankSlot = reinterpret_cast<CAESoundManager__CancelSoundsInBankSlot>(FUNC_CAESoundManager__CancelSoundsInBankSlot);
-    pCancelSoundsInBankSlot(m_pInterface, uiGroup, uiIndex);
+    auto args = PrepareSignature(m_pInterface, uiGroup, uiIndex);
+    CallGTAFunction<CAESound*, __THISCALL>(FUNC_CAESoundManager__CancelSoundsInBankSlot, args);
 }
--- a/Client/game_sa/gamesa_function_caller.h
+++ b/Client/game_sa/gamesa_function_caller.h
@@ -10,7 +10,6 @@
  *****************************************************************************/
 
 #pragma once
-#include "StdInc.h"
 
 namespace CallingConventions
 {
--- a/Client/game_sa/gamesa_function_caller.h
+++ b/Client/game_sa/gamesa_function_caller.h
@@ -1,0 +1,52 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        game_sa/gamesa_function_caller.h
+ *  PURPOSE:     Game functions caller
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#pragma once
+#include "StdInc.h"
+
+namespace CallingConventions
+{
+    struct __THISCALL{};
+    struct __CDECL{};
+    struct __STDCALL{};
+    struct __FASTCALL{};
+    struct __VECTORCALL{};
+} // namespace CallingConventions
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::__THISCALL>, ReturnType> CallGTAFunction(Func function, Args... params)
+{
+    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::__CDECL>, ReturnType> CallGTAFunction(Func function, Args... params)
+{
+    return reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function)(params...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::__STDCALL>, ReturnType> CallGTAFunction(Func function, Args... params)
+{
+    return reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function)(params...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::__FASTCALL>, ReturnType> CallGTAFunction(Func function, Args... params)
+{
+    return reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function)(params...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::__VECTORCALL>, ReturnType> CallGTAFunction(Func function, Args... params)
+{
+    return reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function)(params...);
+}
--- a/Client/game_sa/gamesa_init.h
+++ b/Client/game_sa/gamesa_init.h
@@ -104,3 +104,62 @@ void MemOrFast(U ptr, const T value)
 
 bool GetDebugIdEnabled(uint uiDebugId);
 void LogEvent(uint uiDebugId, const char* szType, const char* szContext, const char* szBody, uint uiAddReportLogId = 0);
+
+struct __THISCALL{};
+struct __CDECL{};
+struct __STDCALL{};
+struct __FASTCALL{};
+struct __VECTORCALL{};
+
+template <typename... Args>
+struct GTAFuncSignature
+{
+    std::tuple<Args...> args;
+    GTAFuncSignature(Args... a) : args(std::forward<Args>(a)...) {}
+};
+
+template <typename... Args>
+GTAFuncSignature<Args...> PrepareSignature(Args... args)
+{
+    return GTAFuncSignature<Args...>(std::forward<Args>(args)...);
+}
+
+template <typename ReturnType, typename Func, typename Tuple, std::size_t... I>
+ReturnType apply_helper(Func&& f, Tuple&& t, std::index_sequence<I...>)
+{
+    return f(std::get<I>(std::forward<Tuple>(t))...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+ReturnType CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    if constexpr (std::is_same_v<CallingConvention, __THISCALL>)
+    {
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    }
+    else if constexpr (std::is_same_v<CallingConvention, __CDECL>)
+    {
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    }
+    else if constexpr (std::is_same_v<CallingConvention, __STDCALL>)
+    {
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    }
+    else if constexpr (std::is_same_v<CallingConvention, __FASTCALL>)
+    {
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    }
+    else if constexpr (std::is_same_v<CallingConvention, __VECTORCALL>)
+    {
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    }
+
+    static_assert(
+        std::is_same_v<CallingConvention, __THISCALL> ||
+        std::is_same_v<CallingConvention, __CDECL> ||
+        std::is_same_v<CallingConvention, __STDCALL> ||
+        std::is_same_v<CallingConvention, __FASTCALL> ||
+        std::is_same_v<CallingConvention, __VECTORCALL>,
+        "Invalid calling onvention specified in CallGTAFunction"
+    );
+}
--- a/Client/game_sa/gamesa_init.h
+++ b/Client/game_sa/gamesa_init.h
@@ -115,13 +115,13 @@ template <typename... Args>
 struct GTAFuncSignature
 {
     std::tuple<Args...> args;
-    GTAFuncSignature(Args... a) : args(std::forward<Args>(a)...) {}
+    GTAFuncSignature(Args... a) : args(std::move(a)...) {}
 };
 
 template <typename... Args>
 GTAFuncSignature<Args...> PrepareSignature(Args... args)
 {
-    return GTAFuncSignature<Args...>(std::forward<Args>(args)...);
+    return GTAFuncSignature<Args...>(std::move(args)...);
 }
 
 template <typename ReturnType, typename Func, typename Tuple, std::size_t... I>
@@ -131,18 +131,31 @@ ReturnType apply_helper(Func&& f, Tuple&& t, std::index_sequence<I...>)
 }
 
 template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-ReturnType CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+typename std::enable_if<std::is_same<CallingConvention, __THISCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
 {
-    if constexpr (std::is_same_v<CallingConvention, __THISCALL>)
-        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    else if constexpr (std::is_same_v<CallingConvention, __CDECL>)
-        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    else if constexpr (std::is_same_v<CallingConvention, __STDCALL>)
-        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    else if constexpr (std::is_same_v<CallingConvention, __FASTCALL>)
-        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    else if constexpr (std::is_same_v<CallingConvention, __VECTORCALL>)
-        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    else
-        static_assert(sizeof(char) == 0, "Invalid calling convention specified in CallGTAFunction");
+    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, __CDECL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, __STDCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, __FASTCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, __VECTORCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
 }
--- a/Client/game_sa/gamesa_init.h
+++ b/Client/game_sa/gamesa_init.h
@@ -105,57 +105,41 @@ void MemOrFast(U ptr, const T value)
 bool GetDebugIdEnabled(uint uiDebugId);
 void LogEvent(uint uiDebugId, const char* szType, const char* szContext, const char* szBody, uint uiAddReportLogId = 0);
 
-struct __THISCALL{};
-struct __CDECL{};
-struct __STDCALL{};
-struct __FASTCALL{};
-struct __VECTORCALL{};
-
-template <typename... Args>
-struct GTAFuncSignature
+namespace CallingConventions
 {
-    std::tuple<Args...> args;
-    GTAFuncSignature(Args... a) : args(std::move(a)...) {}
-};
+    struct thiscall{};
+    struct ___cdecl{};
+    struct stdcall{};
+    struct fastcall{};
+    struct vectorcall{};
+} // namespace CallingConventions
 
-template <typename... Args>
-GTAFuncSignature<Args...> PrepareSignature(Args... args)
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::thiscall>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
 {
-    return GTAFuncSignature<Args...>(std::move(args)...);
-}
-
-template <typename ReturnType, typename Func, typename Tuple, std::size_t... I>
-ReturnType apply_helper(Func&& f, Tuple&& t, std::index_sequence<I...>)
-{
-    return f(std::get<I>(std::forward<Tuple>(t))...);
+    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
 }
 
 template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-typename std::enable_if<std::is_same<CallingConvention, __THISCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::___cdecl>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
 {
-    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    return reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function)(params...);
 }
 
 template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-typename std::enable_if<std::is_same<CallingConvention, __CDECL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::stdcall>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
 {
-    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    return reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function)(params...);
 }
 
 template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-typename std::enable_if<std::is_same<CallingConvention, __STDCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::fastcall>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
 {
-    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    return reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function)(params...);
 }
 
 template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-typename std::enable_if<std::is_same<CallingConvention, __FASTCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::vectorcall>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
 {
-    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-}
-
-template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
-typename std::enable_if<std::is_same<CallingConvention, __VECTORCALL>::value, ReturnType>::type CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
-{
-    return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    return reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function)(params...);
 }
--- a/Client/game_sa/gamesa_init.h
+++ b/Client/game_sa/gamesa_init.h
@@ -134,32 +134,15 @@ template <typename ReturnType, typename CallingConvention, typename Func, typena
 ReturnType CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
 {
     if constexpr (std::is_same_v<CallingConvention, __THISCALL>)
-    {
         return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    }
     else if constexpr (std::is_same_v<CallingConvention, __CDECL>)
-    {
         return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    }
     else if constexpr (std::is_same_v<CallingConvention, __STDCALL>)
-    {
         return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    }
     else if constexpr (std::is_same_v<CallingConvention, __FASTCALL>)
-    {
         return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    }
     else if constexpr (std::is_same_v<CallingConvention, __VECTORCALL>)
-    {
         return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
-    }
-
-    static_assert(
-        std::is_same_v<CallingConvention, __THISCALL> ||
-        std::is_same_v<CallingConvention, __CDECL> ||
-        std::is_same_v<CallingConvention, __STDCALL> ||
-        std::is_same_v<CallingConvention, __FASTCALL> ||
-        std::is_same_v<CallingConvention, __VECTORCALL>,
-        "Invalid calling onvention specified in CallGTAFunction"
-    );
+    else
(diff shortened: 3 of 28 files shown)


Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,62 @@ void MemOrFast(U ptr, const T value)
...
+    }
+
+    static_assert(
+        std::is_same_v<CallingConvention, __THISCALL> ||
+        std::is_same_v<CallingConvention, __CDECL> ||
+        std::is_same_v<CallingConvention, __STDCALL> ||
+        std::is_same_v<CallingConvention, __FASTCALL> ||
+        std::is_same_v<CallingConvention, __VECTORCALL>,
+        "Invalid calling onvention specified in CallGTAFunction"
+    );
```
TracerDS: you could just put it in an else block:
```cpp
else if constexpr (...)
{
    // ...
}
else
{
    static_assert(sizeof(char) == 0, "Invalid calling convention specified in CallGTAFunction");
}
```
~~btw typo: "calling **o**nvention"~~

CrosRoad95:
In my opinion for now you should refactor only few functions calls, see how it behave in production, and then refactor more. You may miss something, make mistake and it is kinda hard to verify if all your refactors are allright.

TracerDS (changes requested):
I see you've taken a liking to my method :)
This way dealing with raw memory would be much better.

FileEX:
> In my opinion for now you should refactor only few functions calls, see how it behave in production, and then refactor more. You may miss something, make mistake and it is kinda hard to verify if all your refactors are allright.

I tested some functions that I changed and so far everything works fine, but maybe you're right, that's why this is enough for now. Only when the next nightly after merging this PR turns out to work without any problems, then I will finish the remaining files

TheNormalnij:
MSVC doesn't optimize your calls. [link](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAKwAnKSuADJ4DJgAcj4ARpjEAQBspAAOqAqETgwe3r4BwemZjgLhkTEs8Yn%2BKXaYDtlCBEzEBLk%2BfkG2mPYlDE0tBGXRcQnJts2t7fldCpNDESOVYzUAlLaoXsTI7BzmAMwRyN5YANQm%2B27Ic/ioF9gmGgCCB0cnmOeXBF6p9PePLzMhwYxy8ZwubgIAE9UpgAPoEYhMQgKf7PV4g96fNxsFgkKFol7ooFYGiRU5wuHILDHAEHUmLCkIhB4BSiWi0OkkzBkj6Um7sznE/YM8mUqhMOaCrkinmMykAN3qRG2YiFRKec2IXgcTIAKgAJACSQjcT1CoRMAHYrFaACIXKzPLU6ghMtx27BuS0260O/ZOzWI11MoR6u1mi3W23%2BwMu3WUgBiTzDkZ9McdAPjbspADUvXqAPIAJTT0b9mfRzwImBYvyYNex0NhzDYADoO6cnsRgKj9g9ncHdQBxPVPRNeEFCPDAZjfYiYOm%2B56nVenG4gEDfX6Ly7d3sdtv3U4tXuVp5r06j8eT5DT2cNraYCD7hSHk%2BrU4gE89hQQDcgPwxAAO4tOgEKvvcEBMKsh6fuW9pLrGdLVrW9aNhCzaMKwmDvpB/YAteE5TjOc5PhBv6HseygLqkLSYPeZELi%2BlGdqeCirEugaXgu84MFeY7EXepGPguFEHh2UEAUBoHEOBe6/lB7GwR2nEBkhKFPDWdYGBhXwwthbCnMWmB8XqBmkKcWGth8QmWdZOGnHqPz0JZAGZAAXvCBDvkahImWZBknqkvxQnCCA9LCxAQEJ5hJHFpxUJZzk7nFCUEG5BDoJuERYKocJKAAjl4jA7BCRpUf2nHosuF5rrxWz8VQ/5ZZuwCmeVUmtYBJCyfJbgpX8/YQAQqwqW2amBhWVZaWhukfJhBk2cZpmNeZsL2UtjluGqETAB4DBKoI2SbS2jl2VZW3tp2%2BEDk8AVrUFO0ckRt59DFt6JW9J0CTeJEPvOu5mqxR4EWYSTrjO1UvLVl54FQpxoAwcyYKoqTEKcLXZSArIFThcIKhCz20HtB1HX0lmUoaJpllVAKXjxq3EPxTAhbQYURbQUUQg9zPrbu2AQAuEQ1sQ6OmVSkoEDzTMMPzECUgQLJsmqABUqwsRJE1QVQ30COskPAG27GZdjuWowVmDFaV8JAeJb6SQRvr2pN9Nrj0SinPDiMCCjaMY1jOUKHjbAE0Tu1GGTjAU%2B6nrevc0MM/VsvBaF4WRQkMuBbCUHC4ICTiwQktzFnj2wgrVI0rQ6uaw72vDbrIJ9AbmRGyb67deb%2BVFSVIK2yQ9uVQ8zt2q7K7u7Qnve0jfvo5jAG4wo%2BOE5cxOkwI5M/ZSYYRualp0%2BPScNczqfs%2BnXOZ5cvNywZueYCLBe8cX0tX7L8v8llgo16%2BcE63rDAtxnMbX8pscoMDypba2fc4R2wUlrf4I8x51VXB7D409fY1n9vPTuwcl6hxXm4NekcN7Ry3nCZMqY94JzdkfFOrM06c25q/bOAshb33zmLJ%2BogS7MLLs%2BcUUsv4ax/qpP%2BTdsiALbiAjuZtwEWx7jbGBA84F1wQTGJBl5UFewRjPTBc9A441wcvcOHJ16HVIQISmcJ8wDRLLTbAick6nGPizNmHMM7EFLnzW%2Bw086i0Ls/LxN9y6KmVCQIRtdf4N3/pI4BvZQE4zkd3K2vcdhKM8SooeCFR7nk0ZPRch8GZzAbHgZAcJJRKFaP%2BPAXlUDNWQAgFo8F9j%2BjtKcDQllzBmCNIdMQeB0CIwjsAH25jjoCHXLCZA8M8CYAGREU4xNXriIEF0pB00NQi1OCwZEDAIDQ3LIUpgXgiA/l7AiVGgMLBeCoFQBIQhVqpE%2BG0mimA6ILkYqJZ8EAFSoH6erBgXgOSpERJZDp7TLIAFouCWS4EkNZhTFmCX/hCH5/SrHU1NFQ4aGhVBWjMImbAAAOMF7FzmqEudc25xB7nbg0cnPi7TzzrI4OsWgnB/C8D8NwXgqBOBuGsNYdcmxtgLSBDwUgBBNAsvWAAaxAJIfYbZAiEskHi/YXANBJH2BoLgeL9CcEkJyqVpBeUcF4AoEAHTJUcC0OsOAsAYCIBQKgOsdAEjkEoGgV19BEgKmQCFAmXBAhwnmnMOEqhCUpBoLQUWFqICxGNbECILQoScHFUm5gxAoSFliNoZUabeBerYIIQsDB2bGqwLELw%2B01QWu5aQLA2yjDiBtbwfAC4GhKjrVoEIqh6jHN2OKh%2BbLW36DwLEJEWaPBYGNYiPALAC2kCVMQWIGRMB2lrIYYAJMjBSvWFQAwvZcwzOAoWFsi7%2BCCBEGIdgUgZCCEUCodQo7dDQoMLu0wlhrBjtiBayA6xUDAuyHWiFNwLh2k/VYSwurTgQsLPsHly7iD9MwH%2BvZ3RejZBcOA6YfhoVhEWBUKoegihZAELhkjGQyMMGGERsY0K6gNAEAMKYngOh6EY30FjCxyijESAx%2BYFGBODFo3xiQ6wFDCp2OJ/VHAOWkC5T201pwI0QxYAoP1pwFRBrbKGt0EBcCEBIOcMVqxeDWttbKkA%2BxCVtkkCqhz9mnOOdk4a0gC6rRJDbIS5V%2BxAj%2BDxUGsw/g72KZ5Zwc1lqJV7tIPap1qN%2B0qg9dBX8yhDA9CEAgVAwEuXiq9akOgJSBBpciLQTL2WwukHy2631/rUiBuDXp8NkaqsuoKz6qIOEIute9QkQsxzys5eNQl5Ar5usjaaPgLlvBL3CHZLe6Qs3H1qGNa%2B/QW6UACssD%2BtDAGgO%2B04KBrK4HIPWCBIhhIyGsC7dINqY6bA9SoE8DdyTWxpMCamyVjLWWhv1sRJgdgHTgJIlSAW1l7KjWjtNdgPtyBjnGdU1sjTyAtM6b05jflX7tunEMyqEz6qzPRdbWNVzvAF1AjbBq2FGhJBmCtMSq09mUiVdNZFq1e6rM1DbDq/wdOkg%2Bf2EkZVdPZMIYU8a1nRPLOybMJDpT3WLPSqXQkTIzhJBAA%3D%3D%3D)

FileEX:
> MSVC doesn't optimize your calls. [link](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAKwAnKSuADJ4DJgAcj4ARpjEAQBspAAOqAqETgwe3r4BwemZjgLhkTEs8Yn%2BKXaYDtlCBEzEBLk%2BfkG2mPYlDE0tBGXRcQnJts2t7fldCpNDESOVYzUAlLaoXsTI7BzmAMwRyN5YANQm%2B27Ic/ioF9gmGgCCB0cnmOeXBF6p9PePLzMhwYxy8ZwubgIAE9UpgAPoEYhMQgKf7PV4g96fNxsFgkKFol7ooFYGiRU5wuHILDHAEHUmLCkIhB4BSiWi0OkkzBkj6Um7sznE/YM8mUqhMOaCrkinmMykAN3qRG2YiFRKec2IXgcTIAKgAJACSQjcT1CoRMAHYrFaACIXKzPLU6ghMtx27BuS0260O/ZOzWI11MoR6u1mi3W23%2BwMu3WUgBiTzDkZ9McdAPjbspADUvXqAPIAJTT0b9mfRzwImBYvyYNex0NhzDYADoO6cnsRgKj9g9ncHdQBxPVPRNeEFCPDAZjfYiYOm%2B56nVenG4gEDfX6Ly7d3sdtv3U4tXuVp5r06j8eT5DT2cNraYCD7hSHk%2BrU4gE89hQQDcgPwxAAO4tOgEKvvcEBMKsh6fuW9pLrGdLVrW9aNhCzaMKwmDvpB/YAteE5TjOc5PhBv6HseygLqkLSYPeZELi%2BlGdqeCirEugaXgu84MFeY7EXepGPguFEHh2UEAUBoHEOBe6/lB7GwR2nEBkhKFPDWdYGBhXwwthbCnMWmB8XqBmkKcWGth8QmWdZOGnHqPz0JZAGZAAXvCBDvkahImWZBknqkvxQnCCA9LCxAQEJ5hJHFpxUJZzk7nFCUEG5BDoJuERYKocJKAAjl4jA7BCRpUf2nHosuF5rrxWz8VQ/5ZZuwCmeVUmtYBJCyfJbgpX8/YQAQqwqW2amBhWVZaWhukfJhBk2cZpmNeZsL2UtjluGqETAB4DBKoI2SbS2jl2VZW3tp2%2BEDk8AVrUFO0ckRt59DFt6JW9J0CTeJEPvOu5mqxR4EWYSTrjO1UvLVl54FQpxoAwcyYKoqTEKcLXZSArIFThcIKhCz20HtB1HX0lmUoaJpllVAKXjxq3EPxTAhbQYURbQUUQg9zPrbu2AQAuEQ1sQ6OmVSkoEDzTMMPzECUgQLJsmqABUqwsRJE1QVQ30COskPAG27GZdjuWowVmDFaV8JAeJb6SQRvr2pN9Nrj0SinPDiMCCjaMY1jOUKHjbAE0Tu1GGTjAU%2B6nrevc0MM/VsvBaF4WRQkMuBbCUHC4ICTiwQktzFnj2wgrVI0rQ6uaw72vDbrIJ9AbmRGyb67deb%2BVFSVIK2yQ9uVQ8zt2q7K7u7Qnve0jfvo5jAG4wo%2BOE5cxOkwI5M/ZSYYRualp0%2BPScNczqfs%2BnXOZ5cvNywZueYCLBe8cX0tX7L8v8llgo16%2BcE63rDAtxnMbX8pscoMDypba2fc4R2wUlrf4I8x51VXB7D409fY1n9vPTuwcl6hxXm4NekcN7Ry3nCZMqY94JzdkfFOrM06c25q/bOAshb33zmLJ%2BogS7MLLs%2BcUUsv4ax/qpP%2BTdsiALbiAjuZtwEWx7jbGBA84F1wQTGJBl5UFewRjPTBc9A441wcvcOHJ16HVIQISmcJ8wDRLLTbAick6nGPizNmHMM7EFLnzW%2Bw086i0Ls/LxN9y6KmVCQIRtdf4N3/pI4BvZQE4zkd3K2vcdhKM8SooeCFR7nk0ZPRch8GZzAbHgZAcJJRKFaP%2BPAXlUDNWQAgFo8F9j%2BjtKcDQllzBmCNIdMQeB0CIwjsAH25jjoCHXLCZA8M8CYAGREU4xNXriIEF0pB00NQi1OCwZEDAIDQ3LIUpgXgiA/l7AiVGgMLBeCoFQBIQhVqpE%2BG0mimA6ILkYqJZ8EAFSoH6erBgXgOSpERJZDp7TLIAFouCWS4EkNZhTFmCX/hCH5/SrHU1NFQ4aGhVBWjMImbAAAOMF7FzmqEudc25xB7nbg0cnPi7TzzrI4OsWgnB/C8D8NwXgqBOBuGsNYdcmxtgLSBDwUgBBNAsvWAAaxAJIfYbZAiEskHi/YXANBJH2BoLgeL9CcEkJyqVpBeUcF4AoEAHTJUcC0OsOAsAYCIBQKgOsdAEjkEoGgV19BEgKmQCFAmXBAhwnmnMOEqhCUpBoLQUWFqICxGNbECILQoScHFUm5gxAoSFliNoZUabeBerYIIQsDB2bGqwLELw%2B01QWu5aQLA2yjDiBtbwfAC4GhKjrVoEIqh6jHN2OKh%2BbLW36DwLEJEWaPBYGNYiPALAC2kCVMQWIGRMB2lrIYYAJMjBSvWFQAwvZcwzOAoWFsi7%2BCCBEGIdgUgZCCEUCodQo7dDQoMLu0wlhrBjtiBayA6xUDAuyHWiFNwLh2k/VYSwurTgQsLPsHly7iD9MwH%2BvZ3RejZBcOA6YfhoVhEWBUKoegihZAELhkjGQyMMGGERsY0K6gNAEAMKYngOh6EY30FjCxyijESAx%2BYFGBODFo3xiQ6wFDCp2OJ/VHAOWkC5T201pwI0QxYAoP1pwFRBrbKGt0EBcCEBIOcMVqxeDWttbKkA%2BxCVtkkCqhz9mnOOdk4a0gC6rRJDbIS5V%2BxAj%2BDxUGsw/g72KZ5Zwc1lqJV7tIPap1qN%2B0qg9dBX8yhDA9CEAgVAwEuXiq9akOgJSBBpciLQTL2WwukHy2631/rUiBuDXp8NkaqsuoKz6qIOEIute9QkQsxzys5eNQl5Ar5usjaaPgLlvBL3CHZLe6Qs3H1qGNa%2B/QW6UACssD%2BtDAGgO%2B04KBrK4HIPWCBIhhIyGsC7dINqY6bA9SoE8DdyTWxpMCamyVjLWWhv1sRJgdgHTgJIlSAW1l7KjWjtNdgPtyBjnGdU1sjTyAtM6b05jflX7tunEMyqEz6qzPRdbWNVzvAF1AjbBq2FGhJBmCtMSq09mUiVdNZFq1e6rM1DbDq/wdOkg%2Bf2EkZVdPZMIYU8a1nRPLOybMJDpT3WLPSqXQkTIzhJBAA%3D%3D%3D)

Is there anything I can do about it?

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,45 @@ void MemOrFast(U ptr, const T value)
...
+    else if constexpr (std::is_same_v<CallingConvention, __CDECL>)
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    else if constexpr (std::is_same_v<CallingConvention, __STDCALL>)
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    else if constexpr (std::is_same_v<CallingConvention, __FASTCALL>)
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    else if constexpr (std::is_same_v<CallingConvention, __VECTORCALL>)
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
+    else
+        static_assert(sizeof(char) == 0, "Invalid calling convention specified in CallGTAFunction");
```
tederis: This makes your function ill-formed. Please, use SFINAE for this.

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,45 @@ void MemOrFast(U ptr, const T value)
...
+struct GTAFuncSignature
+{
+    std::tuple<Args...> args;
+    GTAFuncSignature(Args... a) : args(std::forward<Args>(a)...) {}
+};
+
+template <typename... Args>
+GTAFuncSignature<Args...> PrepareSignature(Args... args)
+{
+    return GTAFuncSignature<Args...>(std::forward<Args>(args)...);
```
tederis: std::forward have no sense here. You have to use universal references here.

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,45 @@ void MemOrFast(U ptr, const T value)
...
+struct __CDECL{};
+struct __STDCALL{};
+struct __FASTCALL{};
+struct __VECTORCALL{};
+
+template <typename... Args>
+struct GTAFuncSignature
+{
+    std::tuple<Args...> args;
+    GTAFuncSignature(Args... a) : args(std::forward<Args>(a)...) {}
```
tederis: std::forward have no sense here. You have to use universal references here. Or maybe std::move?

TheNormalnij:
> Is there anything I can do about it?

My bad there was an invalid compiler option. There should be `/O2` instead of `-O3`

Client/game_sa/CAESoundManagerSA.cpp
```diff
@@ -23,7 +23,6 @@ CAESoundManagerSA::CAESoundManagerSA(CAESoundManagerSAInterface* pInterface) : m
 
 void CAESoundManagerSA::CancelSoundsInBankSlot(uint uiGroup, uint uiIndex)
 {
-    using CAESoundManager__CancelSoundsInBankSlot = CAESound*(__thiscall*)(CAESoundManagerSAInterface*, uint, uint);
-    static auto pCancelSoundsInBankSlot = reinterpret_cast<CAESoundManager__CancelSoundsInBankSlot>(FUNC_CAESoundManager__CancelSoundsInBankSlot);
-    pCancelSoundsInBankSlot(m_pInterface, uiGroup, uiIndex);
+    auto args = PrepareSignature(m_pInterface, uiGroup, uiIndex);
+    CallGTAFunction<CAESound*, __THISCALL>(FUNC_CAESoundManager__CancelSoundsInBankSlot, args);
```
TheNormalnij: Refactor should be in separated PR

Client/game_sa/gamesa_init.h
TheNormalnij:
```c++
template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
ReturnType CallGTAFunction(Func function, Args... params)
{
    if constexpr (std::is_same_v<CallingConvention, __THISCALL>)
        return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
    else if constexpr (std::is_same_v<CallingConvention, __CDECL>)
        return reinterpret_cast<ReturnType(__cdecl*)(Args...)>(function)(params...);
    else if constexpr (std::is_same_v<CallingConvention, __STDCALL>)
        return reinterpret_cast<ReturnType(__stdcall*)(Args...)>(function)(params...);
    else if constexpr (std::is_same_v<CallingConvention, __FASTCALL>)
        return reinterpret_cast<ReturnType(__fastcall*)(Args...)>(function)(params...);
    else if constexpr (std::is_same_v<CallingConvention, __VECTORCALL>)
        return reinterpret_cast<ReturnType(__vectorcall*)(Args...)>(function)(params...);
    else
        static_assert(sizeof(char) == 0, "Invalid calling convention specified in CallGTAFunction");
}
```
Why do you need `GTAFuncSignature` ?

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,45 @@ void MemOrFast(U ptr, const T value)
...
+ReturnType apply_helper(Func&& f, Tuple&& t, std::index_sequence<I...>)
+{
+    return f(std::get<I>(std::forward<Tuple>(t))...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+ReturnType CallGTAFunction(Func function, GTAFuncSignature<Args...>& sig)
+{
+    if constexpr (std::is_same_v<CallingConvention, __THISCALL>)
+        return apply_helper<ReturnType>(reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function), sig.args, std::index_sequence_for<Args...>{});
```
TheNormalnij: `if constexpr` is C++17

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,45 @@ void MemOrFast(U ptr, const T value)
 
 bool GetDebugIdEnabled(uint uiDebugId);
 void LogEvent(uint uiDebugId, const char* szType, const char* szContext, const char* szBody, uint uiAddReportLogId = 0);
+
+struct __THISCALL{};
+struct __CDECL{};
+struct __STDCALL{};
+struct __FASTCALL{};
+struct __VECTORCALL{};
```
TheNormalnij: `typedef` ?

TracerDS: What for?

Client/game_sa/gamesa_init.h
TracerDS: The point of that is to throw an error, not discard the result. The same thing happens in ArgumentParser.
How does it make the function ill formed ~~beside that it will error if invalid specialization is provided which is the point of it~~?

tederis: > The point of that is to throw an error, not discard the result. The same thing happens in ArgumentParser. How does it make the function ill formed ~beside that it will error if invalid specialization is provided which is the point of it~?

See the standard. The extraction from it:
```
The program is ill-formed, no diagnostic required, if:
- no valid specialization can be generated for a template or a substatement of a constexpr if statement within a template and the template is not instantiated, or
...
```

TracerDS: > See the standard. The extraction from it:
> 
> ```
> The program is ill-formed, no diagnostic required, if:
> - no valid specialization can be generated for a template or a substatement of a constexpr if statement within a template and the template is not instantiated, or
> ...
> ```

The point of this code is to stop the compilation process when a specialization couldnt be generated. Check https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2593r1.html

tederis: Do you know that this is just a proposal? According to the standard this is still the ill-formed program.

Client/game_sa/gamesa_init.h
FileEX: The compiler reports no errors even though the Game SA project uses C++14. So how is this possible?

Client/game_sa/gamesa_init.h
TracerDS: This is not only a proposal, it was implemented in current versions of GCC and Clang.
The standard requires that a valid specialization has to exist for each if constexpr branch, not only for the entire template.

`sizeof` is used instead of false for some previous compiler bugs where you had to trick the compiler into always erroring

FileEX: > This makes your function ill-formed. Please, use SFINAE for this.

You mean something like this?
```cpp
template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
typename std::enable_if<std::is_same<CallingConvention, __THISCALL>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
{
    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
}
```

Client/game_sa/gamesa_init.h
FileEX: Structures are empty, they have no data or methods, so typedef is completely unnecessary here imo

TheNormalnij: You are using `struct` here to declare type. You can do same with using `typedef`. It's less hackly.

TracerDS: so `using __CDECL = struct{};`?

Client/game_sa/gamesa_init.h
tederis: > This is not only a proposal, it was implemented in current versions of GCC and Clang. The standard requires that a valid specialization has to exist for each if constexpr branch, not only for the entire template.
> 
> `sizeof` is used instead of false for some previous compiler bugs where you had to trick the compiler into always erroring

The fact that it may work on some compilers doesn't cancel the deviation from the standard. The code must work on all possible compilers.

tederis: > > This makes your function ill-formed. Please, use SFINAE for this.
> 
> You mean something like this?
> 
> ```c++
> template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
> typename std::enable_if<std::is_same<CallingConvention, __THISCALL>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
> {
>     return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
> }
> ```

Yes

TracerDS: > The fact that it may work on some compilers doesn't cancel the deviation from the standard. The code must work on all possible compilers.

This is now a standard, not that it may or may not work on some compilers.
If compilers dont follow a standard then we shouldnt pay much attention to them either way.
This will work on all compilers that already support `static_assert(false,...)` statement. Very old versions of gcc and clang might not support that, but then they also dont support advanced templating used in ArgumentParser.
Since ArgumentParser isnt causing any known issues, this code will work on all computers as expected.

tederis: > > The fact that it may work on some compilers doesn't cancel the deviation from the standard. The code must work on all possible compilers.
> 
> This is now a standard, not that it may or may not work on some compilers. If compilers dont follow a standard then we shouldnt pay much attention to them either way. This will work on all compilers that already support `static_assert(false,...)` statement. Very old versions of gcc and clang might not support that, but then they also dont support advanced templating used in ArgumentParser. Since ArgumentParser isnt causing any known issues, this code will work on all computers as expected.

Okay. Could you provide a proof of including this proposal into the standard?

TracerDS: > Okay. Could you provide a proof of including this proposal into the standard?

https://github.com/cplusplus/papers/issues/1251

Client/game_sa/gamesa_init.h
TheNormalnij: nvm, it won't work

Client/game_sa/gamesa_init.h
tederis: > > Okay. Could you provide a proof of including this proposal into the standard?
> 
> [cplusplus/papers#1251](https://github.com/cplusplus/papers/issues/1251)

It's only C++23. That means that static_asserts are prohibited in templates until MTA get C++23 support. And that's the end of this discussion.

TracerDS: > It's only C++23. That means that static_asserts are prohibited in templates until MTA get C++23 support. And that's the end of this discussion.

In either case using SFINAE for this might yield undesirable results

FileEX: The apply_helper function is equivalent to std::apply to unpack std::tuple. Without this, arguments are passed incorrectly and crashes occur. In the case you mentioned, I would have to use std::make_tuple

Client/game_sa/CAESoundManagerSA.cpp
FileEX: Okay, reverted

Client/game_sa/gamesa_init.h
FileEX: Done

TheNormalnij: > Without this, arguments are passed incorrectly and crashes occur.

Can you show an example?

FileEX: In CFxSA.cpp
```cpp
        // Call FxSystem_c::AddParticle
        auto args = PrepareSignature(fxParticleSystem, &vecPosition, &newDirection, 0.0f, &fxPrt, -1.0f, fBrightness, 0.0f, 0);
        CallGTAFunction<int, __THISCALL>(FUNC_FXSystem_c_AddParticle, args);
```

Then try calling this function in game
```
crun fxCreateParticle("sand", me.position.x, me.position.y, me.position.z, 0, 0, 1, 56, 191, 52, 255, false, 5, 1, 1, true)
```
I haven't analyzed exactly what's wrong, but the arguments are passed in the wrong way, the registers are empty and the game crashes.

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,58 @@ void MemOrFast(U ptr, const T value)
...
```
tederis: How about using a namespace for these tag structs?

Client/game_sa/gamesa_init.h
TheNormalnij: > The apply_helper function is equivalent to std::apply to unpack std::tuple. Without this, arguments are passed incorrectly and crashes occur. In the case you mentioned, I would have to use std::make_tuple

I removed `PrepareSignature` there. So it's just
```c++
 CallGTAFunction<int, __THISCALL>(FUNC_FXSystem_c_AddParticle, fxParticleSystem, &vecPosition, &newDirection, 0.0f, &fxPrt, -1.0f, fBrightness, 0.0f, 0);
 ```

File: Client/game_sa/gamesa_init.h (hunk 8 above)
Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: Good idea. I think i can then remove the "__" or is it better to leave it?

Comment (TracerDS) [reply Client/game_sa/gamesa_init.h]: if its in a namespace, yes.

File: Client/game_sa/gamesa_init.h (hunk 2 above)
Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: Lol, due to all the confusion and the multitude of proposals, I forgot about it. Yeah, that's great

File: Client/game_sa/gamesa_init.h (hunk 8 above)
Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: Done. For cdecl I had to add a prefix because cdecl and CDECL are defined as macros and the compiler throws errors

Comment (TheNormalnij) [inline Client/game_sa/gamesa_init.h]: It would be nice to organize the new functionality into a new header.

Comment (TracerDS) [reply Client/game_sa/gamesa_init.h]: > Done. For cdecl I had to add a prefix because cdecl and CDECL are defined as macros and the compiler throws errors

In that case maybe `__XYZ` format isnt so bad after all

Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: What do you propose?

Comment (TheNormalnij) [reply Client/game_sa/gamesa_init.h]: Put all changes into a new file

Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: I think so too

Comment (FileEX) [reply Client/game_sa/gamesa_init.h]: Yeah, but I mean the name. What could I name the new file? Personally, I have no idea

File: Client/game_sa/gamesa_init.h [hunk 9]
```diff
@@ -104,3 +104,42 @@ void MemOrFast(U ptr, const T value)
...
+{
+    struct thiscall{};
+    struct ___cdecl{};
+    struct stdcall{};
+    struct fastcall{};
+    struct vectorcall{};
+} // namespace CallingConventions
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::thiscall>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
```
tederis: Please, use universal references and the perfect forwarding.

```
template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
typename std::enable_if<std::is_same<CallingConvention, CallingConventions::thiscall>::value, ReturnType>::type CallGTAFunction(Func function, Args&&... params)
{
    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(std::forward<Args>(params)...);
}
```
The same for below.

Client/game_sa/gamesa_init.h
```diff
@@ -104,3 +104,42 @@ void MemOrFast(U ptr, const T value)
...
+{
+    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
+}
+
+template <typename ReturnType, typename CallingConvention, typename Func, typename... Args>
+typename std::enable_if<std::is_same<CallingConvention, CallingConventions::___cdecl>::value, ReturnType>::type CallGTAFunction(Func function, Args... params)
```
tederis: There are nice suffixes _t and _v:
`typename std::enable_if_t<std::is_same_v<CallingConvention, CallingConventions::thiscall>, ReturnType> CallGTAFunction(Func function, Args&&... params)`

tederis (changes requested):

Client/game_sa/gamesa_init.h
FileEX: Is there any difference between a function without a suffix like std::is_same and with a suffix like std::is_same_v?

tederis: No, just a syntactic sugar

tederis: Instead of `std::is_same<a,b>::value` you can write `std::is_same_v<a,b>`.

TracerDS: > Is there any difference between a function without a suffix like std::is_same and with a suffix like std::is_same_v?

`std::is_same_v` is just an alias to `std::is_same::value`:
```cpp
template <typename T, typename Y>
constexpr auto is_same_v = is_same<T, Y>::value;
```

Client/game_sa/gamesa_init.h
FileEX: Universal references are probably not a good idea? Then the argument, which is a pointer, turns into a reference
``FxSystem_c* -> FxSystem_c*&`` and this causes crashes

Client/game_sa/gamesa_function_caller.h
```diff
@@ -0,0 +1,52 @@
...
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        game_sa/gamesa_function_caller.h
+ *  PURPOSE:     Game functions caller
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#pragma once
+#include "StdInc.h"
```
TheNormalnij: I think you should not use precompiled header here

FileEX: Why?

CrosRoad95: Because precompiled headers should be only included in .cpp files, right now i'm 101% sure you include it twice

Client/game_sa/gamesa_init.h
tederis: Why would it turn into a reference? universal refs + perfect forwarding are created to perfectly mirror types.

tederis:
I suggest you to change the core in the following way:

```
template <typename CallingConvention, typename ReturnType, typename Func, typename... Args>
ReturnType CallGTAFunction(Func function, Args... params);

template <typename CallingConvention, typename ReturnType, typename Func, typename... Args>
ReturnType CallGTAFunction<CallingConventions::__THISCALL>(Func function, Args... params)
{
    return reinterpret_cast<ReturnType(__thiscall*)(Args...)>(function)(params...);
}

...
```

That would be cleaner than SFINAE. The new syntax would be: `CallGTAFunction<CallingConventions::__THISCALL, void>(FN_Address, arg1, arg2, ...);`

sbx320:
This seems like a suboptimal approach to me. I think we should rather replace address #defines with proper typed function declarations. 


e.g.

instead of 
```cpp
#define FUNC_HasCollisionBeenLoaded                         0x410CE0
```
and
```cpp
    DWORD dwFunc = FUNC_HasCollisionBeenLoaded;
    bool  bRet = false;
    _asm
    {
        push    0
        push    vecPosition
        call    dwFunc
        mov     bRet, al
        add     esp, 8
    }
    return bRet;

```

just do 

```cpp
static auto FUNC_HasCollisionBeenLoaded = reinterpret_cast<bool(__cdecl*)(CVector* vecPosition, bool flag)>(0x410CE0);
```
and
```cpp
    return FUNC_HasCollisionBeenLoaded(vecPosition, false);
```

TracerDS:
> This seems like a suboptimal approach to me. I think we should rather replace address #defines with proper typed function declarations.
> 
> e.g.
> 
> instead of
> 
> ```c++
> #define FUNC_HasCollisionBeenLoaded                         0x410CE0
> ```
> 
> and
> 
> ```c++
>     DWORD dwFunc = FUNC_HasCollisionBeenLoaded;
>     bool  bRet = false;
>     _asm
>     {
>         push    0
>         push    vecPosition
>         call    dwFunc
>         mov     bRet, al
>         add     esp, 8
>     }
>     return bRet;
> ```
> 
> just do
> 
> ```c++
> static auto FUNC_HasCollisionBeenLoaded = reinterpret_cast<bool(__cdecl*)(CVector* vecPosition, bool flag)>(0x410CE0);
> ```
> 
> and
> 
> ```c++
>     return FUNC_HasCollisionBeenLoaded(vecPosition, false);
> ```

Even better, create the appropriate methods in the class you are calling it for (if any) and then call that method instead (definition would just be reinterpreted mem address).
But thats for other PR maybe

FileEX:
> This seems like a suboptimal approach to me. I think we should rather replace address #defines with proper typed function declarations.
> 
> e.g.
> 
> instead of
> 
> ```c++
> #define FUNC_HasCollisionBeenLoaded                         0x410CE0
> ```
> 
> and
> 
> ```c++
>     DWORD dwFunc = FUNC_HasCollisionBeenLoaded;
>     bool  bRet = false;
>     _asm
>     {
>         push    0
>         push    vecPosition
>         call    dwFunc
>         mov     bRet, al
>         add     esp, 8
>     }
>     return bRet;
> ```
> 
> just do
> 
> ```c++
> static auto FUNC_HasCollisionBeenLoaded = reinterpret_cast<bool(__cdecl*)(CVector* vecPosition, bool flag)>(0x410CE0);
> ```
> 
> and
> 
> ```c++
>     return FUNC_HasCollisionBeenLoaded(vecPosition, false);
> ```

In that case, I have an even better suggestion. Maybe we should create a single header file, something like ``gta_functions.h``, where we define all native functions. This way, we will have access to native functions in every file, and they will be defined once in one place. It's similar to how it's done in ``gamesa_renderware.h``

TheNormalnij:
> In that case, I have an even better suggestion. Maybe we should create a single header file, something like `gta_functions.h`, where we define all native functions. This way, we will have access to native functions in every file, and they will be defined once in one place. It's similar to how it's done in `gamesa_renderware.h`

This file will be too messy. I prefer declaring methods to to access engine functions. 
Instead of
```c++
void someLogic() {
        // some code
        using CEntity_ResolveReferences = void*(__thiscall*)(CEntitySAInterface*);
        ((CEntity_ResolveReferences)0x571A40)(m_pInterface);
        // some code
}
 ```
 CEntity_ResolveReferences can be declared in CEntitySAInterface
 ```c++
 CEntitySAInterface {
  //...
     void* ResolveReferences()
    {
        // we can skip additional definition for 0x571A40
        // this number is described in the context
        return ((void*(__thiscall*)(CEntitySAInterface*))0x571A40)(this);
    };
 }
 
 void someLogic() {
     // code
     m_pInterface->ResolveReferences();
     // code
 }
 ```
 
 I kinda agree that this PR gives too small improvements.

**Comment by FileEX** [comment]
```cpp
/////////////////////////////
/// CPed
/////////////////////////////
static auto FUNC_SetModelIndex = reinterpret_cast<void(__thiscall*)(CPedSAInterface* ped, int index)>(0x5E4880);
static auto FUNC_AttachPedToEntity = reinterpret_cast<void(__thiscall*)(CPedSAInterface* ped, CEntitySAInterface* entity, float offsetX, float offsetY, float offsetZ, int direction, float rotationLimit, std::uint32_t weaponType)>(0x5E7CB0);
static auto FUNC_SetIsStanding = reinterpret_cast<void(__thiscall*)(CPedSAInterface* ped, bool standing)>(0x4ABBE0);
static auto FUNC_SetCurrentWeapon = reinterpret_cast<void(__thiscall*)(CPedSAInterface* ped, std::uint8_t slot)>(0x5E61F0);
static auto FUNC_GiveWeapon = reinterpret_cast<std::uint32_t(__thiscall*)(CPedSAInterface* ped, std::uint32_t weaponID, int ammo, int unusedUnknown)>(0x5E6080);
static auto FUNC_GetBonePosition = reinterpret_cast<void(__thiscall*)(CPedSAInterface* ped, CVector* pos, std::uint32_t boneID, bool dynamic)>(0x5E4280);

/////////////////////////////
/// CPlayerPed
/////////////////////////////
static auto FUNC_MakeChangesForNewWeapon_Slot = reinterpret_cast<void(__thiscall*)(CPlayerPedSAInterface* playerPed, std::uint8_t weaponSlot)>(0x60D000);
```

Personally, I wouldn't say that it looks messy. Additionally, this file would be rarely edited.

Will editing CEntitySAInterface cause any issues due to incorrect size?

TheNormalnij:
> Personally, I wouldn't say that it looks messy. Additionally, this file would be rarely edited.

`FUNC_SetModelIndex(pPedInterface, 200)`  vs `pPedInterface->SetModelIndex(200)`

> Will editing CEntitySAInterface cause any issues due to incorrect size?

Methods don't change data struct. Only virtual functions add `vtbl` reference

FileEX:
> > Personally, I wouldn't say that it looks messy. Additionally, this file would be rarely edited.
> 
> `FUNC_SetModelIndex(pPedInterface, 200)` vs `pPedInterface->SetModelIndex(200)`
> 
> > Will editing CEntitySAInterface cause any issues due to incorrect size?
> 
> Methods don't change data struct. Only virtual functions add `vtbl` reference

What about ``cdecl`` functions that are not tied to any specific interface?

TracerDS:
> What about `cdecl` functions that are not tied to any specific interface?

What do you think about a namespace for all gta global funcs?

TheNormalnij:
> What about `cdecl` functions that are not tied to any specific interface?

 `cdecl` functions can be added to related SA class.

TracerDS:
Why was this closed?

PR #3678, opened 2024-08-25
Allow client  peds to enter/exit client vehicles

gownosatana:
This pull request allows client created peds to enter client created vehicles.

`setPedEnterVehicle(Element ped, [Element vehicle, bool passenger])`
When no vehicle is specified ped will enter closest client vehicle

Test resource: [test.zip](https://github.com/user-attachments/files/16740940/test.zip)

PR #3772, opened 2024-10-06
Allow client peds to enter/exit client vehicles - #3678 follow up

TracerDS:
Resolves #3678


Dutchman101:
Nice. The original PR received a lot of code reviews, and you polished the last few things (there was some unwillingness in the other PR).

Changes will be in a build (right after the upcoming r22835) that won't be considered stable until it's been tested.

PR #3778, opened 2024-10-08
Fix #3738 createBuilding optional arguments not handled correctly

### LINKED ISSUE #3738: createBuilding optional arguments not handled correctly
### Describe the bug

doing code below will throw error for arg 5 expect vector get none

### Steps to reproduce

`createBuilding ( 3534, -2412.30884, -601.68127, 132.56250)`
or
`createBuilding ( 3534, -2412.30884, -601.68127, 132.56250,90)` without rz or ry gives error

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fix #3738 

I think this is what the parser should take care of, that if the vector is optional, it should put 0 in the remaining fields

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
@@ -33,7 +33,7 @@ void CLuaBuildingDefs::AddClass(lua_State* luaVM)
     lua_registerclass(luaVM, "Building", "Element");
 }
 
-CClientBuilding* CLuaBuildingDefs::CreateBuilding(lua_State* const luaVM, std::uint16_t modelId, CVector pos, std::optional<CVector> rot, std::optional<std::uint8_t> interior)
+CClientBuilding* CLuaBuildingDefs::CreateBuilding(lua_State* const luaVM, std::uint16_t modelId, CVector pos, std::optional<float> rotX, std::optional<float> rotY, std::optional<float> rotZ, std::optional<std::uint8_t> interior)
 {
     CLuaMain* pLuaMain = m_pLuaManager->GetVirtualMachine(luaVM);
 
@@ -48,14 +48,13 @@ CClientBuilding* CLuaBuildingDefs::CreateBuilding(lua_State* const luaVM, std::u
     if (!CClientBuildingManager::IsValidPosition(pos))
         throw std::invalid_argument("Position is outside of game world");
 
-    if (rot.has_value())
-        ConvertDegreesToRadians(rot.value());
-    else
-        rot.emplace(CVector(0, 0, 0));
+    CVector rot{rotX.value_or(0), rotY.value_or(0), rotZ.value_or(0)};
+    if (rot.fX != 0 || rot.fY != 0 || rot.fZ != 0)
+        ConvertDegreesToRadians(rot);
 
     m_pBuildingManager->ResizePoolIfNeeds();
 
-    CClientBuilding* pBuilding = new CClientBuilding(m_pManager, INVALID_ELEMENT_ID, modelId, pos, rot.value() , interior.value_or(0));
+    CClientBuilding* pBuilding = new CClientBuilding(m_pManager, INVALID_ELEMENT_ID, modelId, pos, rot, interior.value_or(0));
 
     CClientEntity* pRoot = pResource->GetResourceDynamicEntity();
     pBuilding->SetParent(pRoot);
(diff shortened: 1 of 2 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
```diff
@@ -48,14 +48,13 @@ CClientBuilding* CLuaBuildingDefs::CreateBuilding(lua_State* const luaVM, std::u
     if (!CClientBuildingManager::IsValidPosition(pos))
         throw std::invalid_argument("Position is outside of game world");
 
-    if (rot.has_value())
-        ConvertDegreesToRadians(rot.value());
-    else
-        rot.emplace(CVector(0, 0, 0));
+    CVector rot{rotX.value_or(0), rotY.value_or(0), rotZ.value_or(0)};
+    if (rot.fX != 0 || rot.fY != 0 || rot.fZ != 0)
+        ConvertDegreesToRadians(rot);
```
TracerDS, line 54: wouldnt it be better to always convert to radians?

TracerDS (approved):

Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
FileEX: 0 deg = 0 rad. This way, unnecessary calculations can be avoided, just like the previous code did

tederis: There should be floating point literals (.0f). In addition to that `rot.fX != 0` can be replaced with `std::abs(rot.fX) < std::numeric_limits<float>::epsilon()` to make the comparison less strong.

TheNormalnij:
I don't think that original issue is valid. Developers should provide all vector arguments. It has more sense for me.

You fixes only one case of this *issue*.  The main problem hides in argument parser.

TracerDS:
> I don't think that original issue is valid. Developers should provide all vector arguments. It has more sense for me.
> 
> You fixes only one case of this _issue_. The main problem hides in argument parser.

Its more of a logic issue. What do you expect vector to receive? 3 coordinates.
What happens when you dont provide all 3 coordinates? Error occurs.

FileEX:
TheNormalnij is right. This should be fixed in the parser. Many functions rely on the syntax where each argument in [ ] is optional, such as createVehicle, createObject, etc. The method in the old parser, ReadVector3D, would simply insert 0 in place of any missing argument. The new parser should handle this as well, because the current behavior is confusing. In one function, you can provide a single argument, but in another, you have to provide all three, even though each one is marked as optional individually

TracerDS:
> TheNormalnij is right. This should be fixed in the parser. Many functions rely on the syntax where each argument in [ ] is optional, such as createVehicle, createObject, etc. The method in the old parser, ReadVector3D, would simply insert 0 in place of any missing argument. The new parser should handle this as well, because the current behavior is confusing. In one function, you can provide a single argument, but in another, you have to provide all three, even though each one is marked as optional individually

for now lets just treat vectors as `struct { float x,y,z; }`

Client/mods/deathmatch/logic/luadefs/CLuaBuildingDefs.cpp
FileEX: Done

Pirulax:
Should be an easy fix

> TheNormalnij is right. This should be fixed in the parser. Many functions rely on the syntax where each argument in [ ] is optional, such as createVehicle, createObject, etc. The method in the old parser, ReadVector3D, would simply insert 0 in place of any missing argument. The new parser should handle this as well, because the current behavior is confusing. In one function, you can provide a single argument, but in another, you have to provide all three, even though each one is marked as optional individually

So, just default initialize values (CVector for example)?

FileEX:
> Should be an easy fix
> 
> > TheNormalnij is right. This should be fixed in the parser. Many functions rely on the syntax where each argument in [ ] is optional, such as createVehicle, createObject, etc. The method in the old parser, ReadVector3D, would simply insert 0 in place of any missing argument. The new parser should handle this as well, because the current behavior is confusing. In one function, you can provide a single argument, but in another, you have to provide all three, even though each one is marked as optional individually
> 
> So, just default initialize values (CVector for example)?

The point is that if a function accepts ``std::optional<CVector>``, we have to provide the entire vector, all three values, because otherwise, there's an error that says a vector was expected, but a number was received. It should work like in the old parser, where if, for example, two values ``fX`` and ``fY`` were provided, the ``fZ`` value would simply be set to 0 and treated as a valid vector

Pirulax:
yeah thats what i meant
that should be easy

PR #3779, opened 2024-10-08
Fix #3770 Glob pattern in meta.xml should exclude meta.xml

### LINKED ISSUE #3770: Glob pattern in meta.xml should exclude meta.xml
### Describe the bug

Using glob should not include meta.xml

### Steps to reproduce

Make any resource, put this in meta.xml:

```xml
<meta>
	<file src="*"/>
</meta>
```

you will see meta.xml being download

### Version

any

### Additional context

It is to prevent accidental leaking of meta.xml that can contain sensitive data

### Relevant log output

```shell
/
```

### Security Policy

FileEX:
Fix #3770


tederis (approved):

Lpsd:
This doesn't stop people from including meta.xml as an individual entry right? Since that's a valid technique used for map loaders

PR #3782, opened 2024-10-08
Fix CVector optional arguments - #3738

### LINKED ISSUE #3738: createBuilding optional arguments not handled correctly

TracerDS:
Resolves #3738, #3778


Shared/sdk/CVector.h
```diff
@@ -30,13 +30,9 @@ class CVector
     float fY;
     float fZ;
 
-    struct NoInit{};
```
tederis, line 33: For what reason was removed NoInit?

Shared/sdk/CVector.h
```diff
@@ -30,13 +30,9 @@ class CVector
...
-
-    CVector(NoInit) {}
-
     constexpr CVector() : fX(0.0f), fY(0.0f), fZ(0.0f) {}
 
-    constexpr CVector(float x, float y, float z) : fX(x), fY(y), fZ(z) {}
+    constexpr CVector(float x, float y = 0, float z = 0) : fX(x), fY(y), fZ(z) {}
```
tederis: You forgot floating point literals.

Shared/sdk/CVector.h
TracerDS: To always have an initialized vector

tederis: Why do you need that?

TracerDS: Just to make sure nothing breaks.
Not initialized vector should be zero regardless. And its only used in client code when doing sync.
If its needed, I will bring it back

tederis: This is the standard practice for the sake of optimization. See how vectors are implemented in Eigen, glm, UE, etc.

TracerDS: Well, its only 3 variables. There should be more constructions that depend on the NoInit rule for the sake of optimalization tbf.

But yeah, should I bring it back?

tederis: **NoInit** can be noticable when you need to create hundreds of 3d vectors within an `std::vector`, for example. This is relevant for 3D rendering part, for example. **NoInit** should be added in other structures in future. It would be better to bring it back, yes.

Shared/sdk/CVector.h
tederis: In addition to previous reviews this constructor should be marked as **explicit** to avoid an implicit convertation from **float** to **CVector**.

TheNormalnij:
Did you forget to push your changes?

TracerDS:
> Did you forget to push your changes?

Indeed, gonna push it in a sec

tederis:
Why did you marked conversations resolved without corresponding changes in the code?

TracerDS:
> Why did you marked conversations resolved without corresponding changes in the code?

What do you mean? I updated the code according to your reviews

tederis:
> > Why did you marked conversations resolved without corresponding changes in the code?
> 
> What do you mean? I updated the code according to your reviews

Hmm, that's strange. I thought I saw the files untouched. I'm sorry.

Dutchman101:
Thanks for addressing all reviews and the fix itself

Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -561,7 +561,7 @@ struct CLuaFunctionParserBase
         else if constexpr (std::is_same_v<T, CVector>)
         {
             if (lua_isnumber(L, index))
-                return {PopUnsafe<float>(L, index), PopUnsafe<float>(L, index), PopUnsafe<float>(L, index)};
+                return CVector(PopUnsafe<float>(L, index), PopUnsafe<float>(L, index), PopUnsafe<float>(L, index));
```
tederis, line 564: The order of evaluation of arguments is unspecified. That leads to #3895.

PR #3787, opened 2024-10-10
Add new function testSphereAgainstWorld (Close #646)

### LINKED ISSUE #646: New function: testSphereAgainstWorld
Nothing to describe, simillar to processLineOfSight but instead of a line, it tests the sphere
This could be helpful for fancy scripts.

/

/

**Additional context**
int __cdecl CWorld::TestSphereAgainstWorld(CVector sphereCenter, float sphereRadius, CEntity *ignoreEntity, char buildings, char vehicles, char peds, char objects, char dummies, char doIgnoreCameraCheck)

offset 0x00569E20

FileEX:
This PR adds a new function ``testSphereAgainstWorld``, which allows checking if there are any obstacles within a given radius.

### Syntax
```lua
bool hit, element hitElement, int modelID, float elementPositionX, float elementPositionY, float elementPositionZ, float elementRotationX, float elementRotationY, float elementRotationZ, int lodID, string type = testSphereAgainstWorld(float x, float y, float z, float radius [, element ignoredElement = nil, bool checkBuildings = true, checkVehicles = true, checkPeds = true, checkObjects = true, checkDummies = true, ignoreSomeObjectsForCamera = false ] )
```

**hit** - Indicates whether a collision was detected.
**hitElement** - The element that was detected. If it wasn't created in MTA, it will be nil.
**elementPositionX, elementPositionY, elementPositionZ** - World position of the detected element
**elementRotationX, elementRotationY, elementRotationZ** - World rotation of the element
**modelID** - The ID of the element.
**lodID** - The ID of the element's LOD (or 0).
**entityType** - The type of the element (building, vehicle, ped, object, dummy)

The ``modelID``, ``lodID``, and ``entityType`` parameters are returned even if ``hitElement`` is nil.

Closes #646


Einheit-101:
Would it be possible to return the XYZ where the hit got detected?

FileEX:
> Would it be possible to return the XYZ where the hit got detected?

Yes, I'll add it tomorrow

CrosRoad95:
make 7 arguments in total: `from, to, options` "processLineOfSight" already has shown that using function with extensive amount of arguments is problematic, prune to errors, hard to extend.
Same with return value

FileEX:
Personally, I don’t see anything difficult in using functions like processLineOfSight, which return multiple values. You just need to familiarize yourself with the syntax and the example on the wiki. It doesn't cause any problems, errors, huge and difficult bugs, or any other made-up issues

Client/game_sa/CWorldSA.cpp
```diff
@@ -77,6 +77,21 @@ bool CWorldSA::ResetSurfaceInfo(short sSurfaceID)
...
 }
 
+CEntity* CWorldSA::TestSphereAgainstWorld(const CVector& sphereCenter, float radius, CEntity* ignoredEntity, bool checkBuildings, bool checkVehicles, bool checkPeds, bool checkObjects, bool checkDummies, bool cameraIgnore, STestSphereAgainstWorldResult& result)
+{
+    auto entity = ((CEntitySAInterface*(__cdecl*)(CVector, float, CEntitySAInterface*, bool, bool, bool, bool, bool, bool))FUNC_CWorld_TestSphereAgainstWorld)(sphereCenter, radius, ignoredEntity ? ignoredEntity->GetInterface() : nullptr, checkBuildings, checkVehicles, checkPeds, checkObjects, checkDummies, cameraIgnore);
+    if (!entity)
+        return nullptr;
+    
+    result.collisionDetected = true;
+    result.hitPosition = entity->Placeable.matrix->vPos;
```
TheNormalnij: It's not a hit position

FileEX: That's the position of the entity that was detected

CrosRoad95:
All of you go so hard to do not merge pull request over one word that change NOTHING, but when to come to make function with good user experience you trying to do the best to repeat mistakes of past.
Another mistake you made is that you return different arguments:
bool hit, element hitElement, float hitX, float hitY, float hitZ, int modelID, int lodID, int entityType - your function ( missing 50% of arguments )
bool hit, element hitElement, float hitX, float hitY, float hitZ, float normalX, float normalY, float normalZ, ... - processLineOfSight

Even when i was working on bp was returning same table with ray result no matter of cast method you used to give user coherent result.

Why are you suddenly introducing new concept of "entityType"?? I have to make helper method to translate your newly introduced thing into "elementType"? 

> You just need to familiarize yourself with the syntax and the example on the wiki.

...

FileEX:
I don’t know what you're rambling about here with being friendly. The current syntax is user-friendly and consistent with all the other functions. No one is going to revolutionize anything here just because you want it. I never said this function would return the same arguments as processLineOfSight because it probably won’t. Sorry, with all due respect, I take into account the opinions and suggestions of staff and reviewers more than those who are considered toxic and harmful to the platform by most of the community. It seems to me that your time is up, though it’s not my place to judge. This PR, compared to the PRs of your good friend Borsuk, at least follows coding guidelines.

I can only agree with the ``entityType`` parameter, to change it to a more user-friendly version.

(In the meantime, go back to dismissing everything that doesn’t go your way and mocking MTA for not listening to you and how stupid they are)

Client/sdk/game/CWorld.h
```diff
@@ -61,6 +61,16 @@ struct SProcessLineOfSightMaterialInfoResult {
...
 };
 
+struct STestSphereAgainstWorldResult
+{
+    bool           collisionDetected{false};
+    std::uint32_t  modelID{0};
+    CVector        entityPosition{};
+    CVector        entityRotation{};
+    std::uint32_t  lodID{0};
+    std::uint8_t   type{0};
```
TheNormalnij: you can use eEntityType here

ghost (approved):
If you solve TheNormalnij's request, then everything's fine

Client/sdk/game/CWorld.h
FileEX: eEntityType is unavailable in this file. It’s probably better not to include CEntity.h here, right?

TheNormalnij: Why?

Dutchman101:
Thanks for addressing all code reviews, will merge now; please add it to wiki.

PR #3793, opened 2024-10-13
Add damageable objects support

TheNormalnij:
Damageable models can be loaded via a IMG, but they cause random crashes.
This PR adds proper support for damageable models. 

```lua
local modelId = engineRequestModel('object-damageable')
```


Fernando-A-Rocha:
Do we really need another type "object-damageable"?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -2064,6 +2089,12 @@ bool CModelInfoSA::IsTowableBy(CModelInfo* towingModel)
     return isTowable;
 }
 
+bool CModelInfoSA::IsDamageableAtomic()
+{
+    void* asDamagable = ((void* (*)())m_pInterface->VFTBL->AsDamageAtomicModelInfoPtr)();
```
TracerDS, line 2111: Maybe just `((void*)...)() != nullptr`?

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1744,6 +1744,24 @@ void CModelInfoSA::MakeObjectModel(ushort usBaseID)
...
 
+void CModelInfoSA::MakeObjectDamageableModel(std::uint16_t baseModel)
+{
+    CDamageableModelInfoSAInterface* m_pInterface = new CDamageableModelInfoSAInterface();
+
+    CDamageableModelInfoSAInterface* pBaseObjectInfo = static_cast<CDamageableModelInfoSAInterface*>(ppModelInfo[baseModel]);
+    MemCpyFast(m_pInterface, pBaseObjectInfo, sizeof(CDamageableModelInfoSAInterface));
+    m_pInterface->usNumberOfRefs = 0;
+    m_pInterface->pRwObject = nullptr;
+    m_pInterface->usUnknown = 65535;
```
FileEX, line 1772: Set value to -1 for "usUnknown"
https://github.com/multitheftauto/mtasa-blue/blob/ee4593b1a89d36f2a907776323d695f00bdf3d09/Client/game_sa/CModelInfoSA.cpp#L1788

TheNormalnij:
> Do we really need another type "object-damageable"?

This is last missing type from the game world, that can not be allocated. SA uses IS_DAMAGEABLE model flag to allocate damageable models. Changing this flag in MTA causes crashes.  So we need to allocate those models explicitly

botder (dismissed):

PR #3799, opened 2024-10-15
Refactor client side CVehicleNames.cpp

shadylua:
Vehicles with two names: Vehicles with alternative names are listed in order. For example, “Sadler Damaged”, the “Sadler” row has both the damaged version and the normal version names. and The vehicle names are organized alphabetically so that they can be used consistently in the same order on both server and client side. (the confusion in the table has been eliminated by sorting the server and client side from a to z)


TracerDS:
Please add the info about changing the client version of files as per https://github.com/multitheftauto/mtasa-blue/pull/3800#issuecomment-2414458479

TheNormalnij:
It breaks getVehicleName function

PR #3803, opened 2024-10-16
Fix Console (F8) not maintaining position & size when GUI Skin changed

### LINKED ISSUE #2138: Console (F8) set its position when changing MTA: SA Theme (Skin)
Describe the bug
Same in #1969 (which is already fixed)

To reproduce
• Open your MTA;
• Press F8 and adjust the console the way you like (positioning and resizing);
• Change your MTA's theme;
• Check your console.

Expected behaviour
It should be in the same position and size that you have defined.

Screenshots
None

Version
v1.5.8-release-20833

Additional context
None

Fernando-A-Rocha:
Fixes #2138

### DIFF
--- a/Client/core/CGUI.cpp
+++ b/Client/core/CGUI.cpp
@@ -129,12 +129,8 @@ void CLocalGUI::ChangeLocale(const char* szName)
     if (guiWasLoaded)
     {
         CreateWindows(guiWasLoaded);
-
-        if (m_pConsole != nullptr)
-        {
-            m_pConsole->SetPosition(consolePos);
-            m_pConsole->SetSize(consoleSize);
-        }
+        m_pConsole->SetPosition(consolePos);
+        m_pConsole->SetSize(consoleSize);
     }
 }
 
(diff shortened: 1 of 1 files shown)


Client/core/CGUI.cpp
```diff
@@ -119,12 +129,8 @@ void CLocalGUI::ChangeLocale(const char* szName)
...
         CreateWindows(guiWasLoaded);
-
-        if (m_pConsole != nullptr)
-        {
-            m_pConsole->SetPosition(vPos);
-            m_pConsole->SetSize(vSize);
-        }
+        m_pConsole->SetPosition(consolePos);
+        m_pConsole->SetSize(consoleSize);
     }
```
TheNormalnij, line 134: Is it possible that `m_pConsole` is `nullptr` here?

TracerDS: If you call `DestroyWindows` prior then yes

Fernando-A-Rocha: > Is it possible that `m_pConsole` is `nullptr` here?

No because it is created just before.

TheNormalnij (approved):

FileEX (approved):

TracerDS (approved):

TheNormalnij (approved):

PR #3804, opened 2024-10-16
Validate serial on player join

### LINKED ISSUE #3763: Server should reject empty serials
### Describe the bug

@ArranTuna reported this, but can't post it here (likely due to the world 'serial' being flagged).

If a user connects to a server while not having a serial (which is rare - but plausible) the server should reject the player. At the moment it seems like this is not happening.

Servers could of course do a check themselves in Lua, but it's still a good idea to incorporate a check like this.

### Steps to reproduce

N/A

### Version

_No response_

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

Fernando-A-Rocha:
Fixes #3763 by preventing the player from connecting when serial is invalid: client is disconnected with error code `CD44 Serial verification failed`

[getPlayerSerial](https://wiki.multitheftauto.com/wiki/GetPlayerSerial) says that

> you should only trust the value returned by this function when used serverside

So this validation makes sense. With it, we can trust the value. The [Serial page](https://wiki.multitheftauto.com/wiki/Serial) mentions that it is always 32 characters long with letters and numbers. The letters are only A-F

Questions:

1. Why could [getPlayerSerial](https://wiki.multitheftauto.com/wiki/GetPlayerSerial) return false? In other words, why would a player not have a serial on a server?

2. Why does getPlayerSerial serverside accept a second argument (index) that, if 1 and not 0(which is default) returns the "extra serial" which is a string of max 64 chars? 

https://github.com/multitheftauto/mtasa-blue/blob/733683d70dc037fdcbb256fb17d86e93bfedd239/Server/mods/deathmatch/logic/luadefs/CLuaPlayerDefs.cpp#L707-L722

https://github.com/multitheftauto/mtasa-blue/blob/733683d70dc037fdcbb256fb17d86e93bfedd239/Server/mods/deathmatch/logic/net/CNetBuffer.cpp#L720

@Dutchman101

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1793,7 +1793,7 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
             CLogger::LogPrintf("CONNECT: %s failed to connect (Invalid serial) (%s)\n", szNick, strIPAndSerial.c_str());
 
             // Tell the player the problem
-            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::INVALID_SERIAL);
+            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_VERIFICATION);
             return;
         }
 
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1786,7 +1786,7 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
         SString strIPAndSerial("IP: %s  Serial: %s  Version: %s", strIP.c_str(), strSerial.c_str(), strPlayerVersion.c_str());
 
         // Prevent player from connecting if serial is invalid
-        std::regex serialRegex("^[A-F0-9]{32}$");
+        const std::regex serialRegex("^[A-F0-9]{32}$");
         if (!std::regex_match(strSerial, serialRegex))
         {
             // Tell the console
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -71,6 +71,7 @@
 #include "version.h"
 #include "net/SimHeaders.h"
 #include <signal.h>
+#include <regex>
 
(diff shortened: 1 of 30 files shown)


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1781,6 +1782,18 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
         }
 #endif
 
+        // Prevent player from connecting if serial is invalid
+        std::regex serialRegex("^[A-F0-9]{32}$");
+        if (!std::regex_match(strSerial, serialRegex))
+        {
+            // Tell the console
+            CLogger::LogPrintf("CONNECT: %s failed to connect (Invalid serial) (%s)\n", szNick, pPlayer->GetSourceIP());
```
TracerDS: Does `GetSourceIP()` return `std::string`/`SString` or `const char*`? If the former then you need to use `.c_str()`

TracerDS (changes requested):

Server/mods/deathmatch/logic/CGame.cpp
Fernando-A-Rocha: sorry that was incorrect, check now

TracerDS (approved):

Dutchman101:
Thanks for this, even though servers using admin resource (revision [eb14427](https://github.com/multitheftauto/mtasa-resources/commit/eb144275b1cb875a4f088c0b3afa366d1162fc2e)) or higher are already protected, and AC backend performs checks too, having this in MTA Server (mtasa-blue) as well reduces the points of failure and overhead, making it more reliable and secure.

Fernando-A-Rocha:
I think having this in mtasa-blue makes sense indeed. People may not use the admin resource in their servers

CrosRoad95:
Wiki page https://wiki.multitheftauto.com/wiki/Serial should contain information about regex that matches serial with simple snippet example in lua. People may want to use it in external tools too so good to know it.

Fernando-A-Rocha:
> Wiki page https://wiki.multitheftauto.com/wiki/Serial should contain information about regex that matches serial with simple snippet example in lua. People may want to use it in external tools too so good to know it.

Also, now the function will always return a string, never false. We can update wiki @Dutchman101

Fernando-A-Rocha:
Regular users cannot edit the Serial page, and tbh let's edit the getPlayerSerial that'll be good.

PR #3805, opened 2024-10-16
Strip server name's spaces from beginning and end

### LINKED ISSUE #402: Strip the beginning of a server name from repeating chars
People abuse this to get their server on top of the list when sorting alphabetically.

- [ ] Remove leading spaces for all servers
- [ ] If a server contains no alphanumeric characters, hide from server list (this might be easier to enforce within the master server list?)
- [ ] When sorting by name it would be neat to only consider alphanumeric characters (CEGUI limitation?)

**Additional context**
https://bugs.mtasa.com/view.php?id=5070

Fernando-A-Rocha:
Closes #402

Trims leading and trailing spaces from the server name on config load.

Btw, on the topic of #402, it's not a good idea to add an error if server name doesn't contain an alphanumeric character because some foreign languages might not need them at all.

### DIFF
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -133,6 +133,20 @@ bool CMainConfig::Load()
         return false;
     }
 
+    // Strip spaces from beginning and end of server name
+    size_t snameStart = m_strServerName.find_first_not_of(" ");
+    if (snameStart != std::string::npos)
+    {
+        size_t snameEnd = m_strServerName.find_last_not_of(" ");
+        m_strServerName = m_strServerName.substr(snameStart, snameEnd - snameStart + 1);
+    }
+    else
+    {
+        // The string contains only spaces
+        CLogger::ErrorPrintf("Server name must contain at least one character other than a space\n");
+        return false;
+    }
+
     // Grab the forced server ip(s)
     GetString(m_pRootNode, "serverip", m_strServerIP);
     m_strServerIP = SString(m_strServerIP).Replace(" ", "");
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -133,17 +133,8 @@ bool CMainConfig::Load()
         return false;
     }
 
-    // Cancel if the server name only contains spaces
-    const std::size_t snameStart = m_strServerName.find_first_not_of(" ");
-    if (snameStart == std::string::npos)
-    {
-        CLogger::ErrorPrintf("Server name must contain at least one character other than a space\n");
-        return false;
-    }
-
     // Strip spaces from beginning and end of server name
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,20 @@ bool CMainConfig::Load()
...
+    {
+        const size_t snameEnd = m_strServerName.find_last_not_of(" ");
+        m_strServerName = m_strServerName.substr(snameStart, snameEnd - snameStart + 1);
+    }
+    else
+    {
+        // The string contains only spaces
+        CLogger::ErrorPrintf("Server name must contain at least one character other than a space\n");
+        return false;
+    }
```
TracerDS: `std::size_t` instead of `size_t`.
Also maybe early return in case something would be changed here?

TracerDS (changes requested):

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,20 @@ bool CMainConfig::Load()
         return false;
     }
 
+    // Strip spaces from beginning and end of server name
+    const size_t snameStart = m_strServerName.find_first_not_of(" ");
+    if (snameStart != std::string::npos)
+    {
+        const size_t snameEnd = m_strServerName.find_last_not_of(" ");
+        m_strServerName = m_strServerName.substr(snameStart, snameEnd - snameStart + 1);
+    }
```
TheNormalnij: It looks like a library function. You can add `trim` to `SString`

Fernando-A-Rocha: this is already the simplest and optimized we can do imo

TheNormalnij: String operations here are universal and could be reused in another code.
You may avoid out-of-context operations here and make the code more self-descriptive:
```c++
if (!m_strServerName.IsTrimed())
    m_strServerName = m_strServerName.Trim();
```

Fernando-A-Rocha: Done ty!

TracerDS (approved):

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -133,6 +133,9 @@ bool CMainConfig::Load()
         return false;
     }
 
+    // Strip spaces from beginning and end of server name
+    m_strServerName = SString(m_strServerName).TrimStart(" ").TrimEnd(" ");
+
```
TheNormalnij, line 138: It's ok in this context, but TrimStart and TrimEnd suck

TheNormalnij (approved):

Server/mods/deathmatch/logic/CMainConfig.cpp
Fernando-A-Rocha: Ok but why do they suck?

PR #3810, opened 2024-10-17
add setVehicleSmokeTrailEnabled & isVehicleSmokeTrailEnabled function

Proxy-99:
`setVehicleSmokeTrailEnabled (vehicle, bool)`
`isVehicleSmokeTrailEnabled(vehicle)`

`works with Cropduster and Stuntplane id 512,513

```lua
veh = {512,513}

local plane = createVehicle (veh[2], 2043.23499, 1537.88562, 10.67188)

local ped = createPed ( 2, 0,0,0 )

warpPedIntoVehicle ( ped, plane)   

setTimer(function()
print(setVehicleSmokeTrailEnabled (plane,true))

 end ,1500,1)
 
 function disbled()
 print(setVehicleSmokeTrailEnabled (plane,false))
 print(isVehicleSmokeTrailEnabled(plane))
 end
 
 bindKey ("r", "down", disbled ) 

```

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -92,7 +92,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"getVehicleModelWheelSize", ArgumentParser<GetVehicleModelWheelSize>},
         {"getVehicleWheelFrictionState", ArgumentParser<GetVehicleWheelFrictionState>},
         {"getVehicleEntryPoints", ArgumentParser<GetVehicleEntryPoints>},
-        {"isPlaneTrailEnabled", ArgumentParser<IsSmokeTrailEnabled>},
+        {"isVehicleSmokeTrailEnabled", ArgumentParser<IsSmokeTrailEnabled>},
 
         // Vehicle set funcs
         {"createVehicle", CreateVehicle},
@@ -157,7 +157,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"setVehicleWheelScale", ArgumentParser<SetVehicleWheelScale>},
         {"setVehicleModelWheelSize", ArgumentParser<SetVehicleModelWheelSize>},
         {"spawnVehicleFlyingComponent", ArgumentParser<SpawnVehicleFlyingComponent>},
-        {"setPlaneTrailEnabled", ArgumentParser<SetSmokeTrailEnabled>},
+        {"setVehicleSmokeTrailEnabled", ArgumentParser<SetSmokeTrailEnabled>},
     };
 
     // Add functions
@@ -246,7 +246,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "getModelWheelSize", "getVehicleModelWheelSize");
     lua_classfunction(luaVM, "getWheelFrictionState", "getVehicleWheelFrictionState");
     lua_classfunction(luaVM, "getEntryPoints", ArgumentParser<OOP_GetVehicleEntryPoints>);
-    lua_classfunction(luaVM, "isTrailEnabled", "isPlaneTrailEnabled");
+    lua_classfunction(luaVM, "isSmokeTrailEnabled", "isVehicleSmokeTrailEnabled");
 
     lua_classfunction(luaVM, "setComponentVisible", "setVehicleComponentVisible");
     lua_classfunction(luaVM, "setSirensOn", "setVehicleSirensOn");
@@ -295,7 +295,7 @@ void CLuaVehicleDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "setVariant", "setVehicleVariant");
     lua_classfunction(luaVM, "setWheelScale", "setVehicleWheelScale");
     lua_classfunction(luaVM, "setModelWheelSize", "setVehicleModelWheelSize");
-    lua_classfunction(luaVM, "setTrailEnabled", "setPlaneTrailEnabled");
+    lua_classfunction(luaVM, "setSmokeTrailEnabled", "setVehicleSmokeTrailEnabled");
 
     lua_classfunction(luaVM, "resetComponentPosition", "resetVehicleComponentPosition");
     lua_classfunction(luaVM, "resetComponentRotation", "resetVehicleComponentRotation");
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -156,7 +156,6 @@ void CLuaVehicleDefs::LoadFunctions()
         {"setVehicleWheelScale", ArgumentParser<SetVehicleWheelScale>},
         {"setVehicleModelWheelSize", ArgumentParser<SetVehicleModelWheelSize>},
         {"spawnVehicleFlyingComponent", ArgumentParser<SpawnVehicleFlyingComponent>},
-        {"setPlaneTrailEnabled", ArgumentParser<SetPlaneSmoke>},
     };
 
     // Add functions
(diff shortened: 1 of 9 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4340,3 +4342,19 @@ bool CLuaVehicleDefs::SpawnVehicleFlyingComponent(CClientVehicle* const vehicle,
...
 }
+
+bool CLuaVehicleDefs::SetSmokeTrailEnabled(CClientVehicle* vehicle, bool state) noexcept
+{
+    if (vehicle->GetModel() == 512 || vehicle->GetModel() == 513) // Support Cropduster and Stuntplane
+    {
+         vehicle->SetSmokeTrailEnabled(state);
+        return true;
+    }
+        return false;
```
TracerDS:
```cpp
auto model = vehicle->GetModel();
if (model != 512 && model != 513)
    return false;
vehicle->SetSmokeTrailEnabled(state);
return true;
```

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Fernando-A-Rocha: +1

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4345,12 +4345,11 @@ bool CLuaVehicleDefs::SpawnVehicleFlyingComponent(CClientVehicle* const vehicle,
...
 {
-    if (vehicle->GetModel() == 512 || vehicle->GetModel() == 513) // Support Cropduster and Stuntplane
-    {
-         vehicle->SetSmokeTrailEnabled(state);
-        return true;
-    }
+    auto model = vehicle->GetModel();
+    if (model != 512 && model != 513)
         return false;
+    vehicle->SetSmokeTrailEnabled(state);
```
TracerDS, line 4354: if you could add the space before that line, that would be great :)
```cpp
if (...)
    return false;

vehicle->...
```

TracerDS (approved):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Proxy-99: now should be good

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4340,3 +4342,19 @@ bool CLuaVehicleDefs::SpawnVehicleFlyingComponent(CClientVehicle* const vehicle,
 
     return vehicle->SpawnFlyingComponent(partNodeIndex, collisionType, removalTime.value_or(-1));
 }
+
+bool CLuaVehicleDefs::SetSmokeTrailEnabled(CClientVehicle* vehicle, bool state) noexcept
+{
+    auto model = vehicle->GetModel();
```
FileEX: We try not to use ``auto`` for simple variable types, so I think it's better to just change it to ``std::uint16_t``

Proxy-99: you are right I wonder why tracer suggest auto @TracerDS

FileEX (approved):

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TracerDS: for simplicity as an example. Doesnt really matter which one you choose here.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -4340,3 +4342,19 @@ bool CLuaVehicleDefs::SpawnVehicleFlyingComponent(CClientVehicle* const vehicle,
 
     return vehicle->SpawnFlyingComponent(partNodeIndex, collisionType, removalTime.value_or(-1));
 }
+
+bool CLuaVehicleDefs::SetSmokeTrailEnabled(CClientVehicle* vehicle, bool state) noexcept
+{
+    std::uint16_t model = vehicle->GetModel();
+    if (model != 512 && model != 513)
+        return false;
```
TheNormalnij: Does this function work with `engineRequestModel( "vehicle", 512 )` ?

TheNormalnij: It's better to throw an exception here instead of silent false.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -156,6 +157,7 @@ void CLuaVehicleDefs::LoadFunctions()
         {"setVehicleWheelScale", ArgumentParser<SetVehicleWheelScale>},
         {"setVehicleModelWheelSize", ArgumentParser<SetVehicleModelWheelSize>},
         {"spawnVehicleFlyingComponent", ArgumentParser<SpawnVehicleFlyingComponent>},
+        {"setPlaneTrailEnabled", ArgumentParser<SetSmokeTrailEnabled>},
```
TheNormalnij: Add OOP method too

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Proxy-99: the model id is invalid, so what do you suggest by checking what

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TheNormalnij, line 160: setVehiclePlaneTrailEnabled

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TheNormalnij: Does this function work with allocated vehicles if you remove the model check?

Fernando-A-Rocha:
it would be nice to make it work with engine request model ids ofc, since this new function is Clientside

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Fernando-A-Rocha: Why not call the functions set/is SmokeTrailEnabled like the c++ function names, instead of "plane trail"?

Fernando-A-Rocha: or is/setVehicleSmokeTrailEnabled

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Proxy-99: no it does not

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
Proxy-99: to make way simple  and related to planes since two ids use it only and its planes

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
TheNormalnij: sad

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
MegadreamsBE: Might still be good to keep it generic. You never know if someday someone makes it possible to have smoke trails behind any vehicle (ex. custom vehicles).

Proxy-99: okay I will rename to match also setVehicleLandingGearDown

TheNormalnij (approved):

PR #2933, opened 2023-03-24
Fix #2932 (Fix CEF hardware acceleration not working)

TFP-dev:
Fixes #2932 and fixes WebGL not working.

Current MTA:
https://youtu.be/Mca4FgfQqMo

Fixed:
https://youtu.be/tepQ-iWNKIM

No issues found, browser works as smooth as it was before this fix (but WebGL is now working)

PR #3816, opened 2024-10-19
Add main menu setting for browser GPU rendering

Lpsd:
Adds two options to "web browser" tab in the main menu settings:

- Enable GPU rendering (CVAR: `browser_enable_gpu=1`)
- Enable GPU compositing (CVAR: `browser_enable_gpu_compositing=1`)

image

GPU rendering must be enabled for GPU compositing to be available. Both options require a restart upon change to take effect.

Adds a clientside Lua function to check if browser GPU is enabled:

```lua
bool isBrowserGPUEnabled()
```

Note: for some reason the GPU is not being disabled for me, even though I've debugged and can see the `disable-gpu` flag being applied (http://gta.rockstarvision.com/vehicleviewer still works for example, which relies on WebGL). Need others to test and see if they get the same results, as it doesn't make much sense 🤔 

GPU has always been disabled in MTA, until #2933 (merged 2 weeks ago)

Update: according to @theSarrum these changes work (both enabling and disabling), tested by attempting to load a page with WebGL content.

### DIFF
--- a/Client/cefweb/CWebCore.cpp
+++ b/Client/cefweb/CWebCore.cpp
@@ -874,12 +874,12 @@ void CWebCore::StaticFetchBlacklistFinished(const SHttpDownloadResult& result)
 #endif
 }
 
-bool CWebCore::GetGPUEnabled()
+bool CWebCore::GetGPUEnabled() const noexcept
 {
     return m_bGPUEnabled;
 }
 
-bool CWebCore::GetGPUCompositingEnabled()
+bool CWebCore::GetGPUCompositingEnabled()  const noexcept
 {
     return m_bGPUCompositingEnabled;
 }
--- a/Client/cefweb/CWebCore.cpp
+++ b/Client/cefweb/CWebCore.cpp
@@ -49,10 +49,14 @@ CWebCore::~CWebCore()
     delete m_pXmlConfig;
 }
 
-bool CWebCore::Initialise()
+bool CWebCore::Initialise(bool gpuEnabled, bool gpuCompositingEnabled)
 {
     CefMainArgs        mainArgs;
     void*              sandboxInfo = nullptr;
+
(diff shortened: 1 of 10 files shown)


Client/cefweb/CWebCore.cpp
```diff
@@ -869,3 +873,13 @@ void CWebCore::StaticFetchBlacklistFinished(const SHttpDownloadResult& result)
     OutputDebugLine("Updated browser blacklist!");
 #endif
 }
+
+bool CWebCore::GetGPUEnabled()
+{
+    return m_bGPUEnabled;
+}
+
+bool CWebCore::GetGPUCompositingEnabled()
```
TracerDS: Can be const and noexcept

Client/cefweb/CWebCore.h
```diff
@@ -108,6 +108,9 @@ class CWebCore : public CWebCoreInterface
     static void StaticFetchWhitelistFinished(const SHttpDownloadResult& result);
     static void StaticFetchBlacklistFinished(const SHttpDownloadResult& result);
 
+    bool GetGPUEnabled();
+    bool GetGPUCompositingEnabled();
```
TracerDS: Above

Client/mods/deathmatch/logic/luadefs/CLuaBrowserDefs.cpp
```diff
@@ -1054,3 +1056,8 @@ int CLuaBrowserDefs::SetBrowserAjaxHandler(lua_State* luaVM)
     lua_pushboolean(luaVM, false);
     return 1;
 }
+
+bool CLuaBrowserDefs::IsBrowserGPUEnabled()
```
TracerDS: Noexcept

TracerDS (changes requested):

TracerDS (approved):

PR #3821, opened 2024-10-21
Remove CEF GPU compositing

Lpsd:
Disable GPU compositing due to lag/input delay and breaking devtools (only in official builds, no issues were identified in custom debug or release builds), as reported by multiple users. 

GPU rendering is still enabled by default.


Fernando-A-Rocha:
Ty for the fast reaction

PR #3823, opened 2024-10-23
Refactor and improve player map (F11)

### LINKED ISSUE #2904: Update default radar.jpg image (F11 map)

The current [`radar.jpg`](https://github.com/multitheftauto/mtasa-blue/blob/master/Shared/data/MTA%20San%20Andreas/MTA/cgui/images/radar.jpg) image contains some imperfections and it's 1024x1024 pixels (rather small resolution). As soon as you zoom in, it will lose quality and become blurred.

It could be improved with a better quality and higher resolution GTA:SA map render.

E.g.: [Patrick2562/mtasa-map-images](https://github.com/Patrick2562/mtasa-map-images/) (mta_noshadow_2048.jpg seems adequate). If @Patrick2562 agrees of course.

Have several resolutions radar.jpg images. Client can select the one they want via `Settings`, similar to how you can select the `Interface` style. E.g. 1024x1024, 2048x2048, 4096x4096, etc.

### Additional context

Similar enhancements have recently been approved: #2829 #2854

### Security Policy

Fernando-A-Rocha:
This PR:

- [x] Refactors and Renames **Radar** to **Player Map** in the code, so it can't be confused with the minimap radar (as it is called in Singleplayer), leaving only the commands for backwards compatibility
- [x] Repositions and translates the F11 help text strings
- [x] Replaces the F11 Map radar.jpg image with a better image of the same resolution (1024x1024) converted to PNG for optimization ([Source](https://github.com/Patrick2562/mtasa-map-images/blob/master/noshadow/mta_noshadow_1024.jpg))
- [x] Adds a new 2048x2048 PNG map image ([Source](https://github.com/Patrick2562/mtasa-map-images/blob/master/noshadow/mta_noshadow_2048.jpg))
- [x] Adds a new Setting for switching between the two images (the 1024 pixels image remains the default one)

image

Closes #2904


TracerDS:
What do you think about 4k or 8k instead of 2k?

Fernando-A-Rocha:
4k has good level of detail, 8k is awesome but imo it has too many pixels. Idk how most hardware would handle a higher image resolution (it's more intensive to render).
maybe 2k is fine

TracerDS:
4k would be sufficient for now, though more feedback would be nice.
It wouldnt be a problem for me but maybe it would for other users as the size doubles with each dimension

FileEX:
Let's not exaggerate, an 8k map, while the entire game uses textures at most 512 or 1024, sounds absurd. 4k will be more than sufficient. It's not something people will stare at and admire, like 'wow, what an amazing map!' because everyone knows it by heart. They'll click on it in a few seconds, and no one will pay attention to its quality

Fernando-A-Rocha:
> Let's not exaggerate, an 8k map, while the entire game uses textures at most 512 or 1024, sounds absurd. 4k will be more than sufficient. It's not something people will stare at and admire, like 'wow, what an amazing map!' because everyone knows it by heart. They'll click on it in a few seconds, and no one will pay attention to its quality

do you suggest 2k or 4k?

FileEX:
Imo, 4k is the maximum size we should consider

TheNormalnij:
> Imo, 4k is the maximum size we should consider

2k is the safe maximum. #1051

TheNormalnij:
@Patrick2562, can we license your work under GPL3?

T-MaxWiese-T:
What would also be cool is if the map would adapt to the time of day, so when it's dark the night map is displayed. But I think just like the chat is planned to be outsourced to a resource, the map should also be outsourced to a resource where you have settings.

Dutchman101:
This can only be considered for merge if all the feedback from PR #1051 is implemented, so please carefully read the comments there (from team members: ccw, me, and qaisjp at the time). In short:
- If the remake/upscale uses textures larger than 2048 x 2048, it must be opt-in (not opt out) with a popup disclaimer that enabling it can cause issues with certain Intel GMA/HD graphics chips, and crashes due to the load, as we have limited VRAM in GTA by design.

Also, even before rendering the image (user pressing F11) it will cause a passive extra load on (V)RAM, just because it gets loaded.

gptrk0:
> @Patrick2562, can we license your work under GPL3?

Yes, of course

Fernando-A-Rocha:
> This can only be considered for merge if all the feedback from PR #1051 is implemented, so please carefully read the comments there (from team members: ccw, me, and qaisjp at the time). In short:
> 
> * If the remake/upscale uses textures larger than 2048 x 2048, it must be opt-in (not opt out) with a popup disclaimer that enabling it can cause issues with certain Intel GMA/HD graphics chips, and crashes due to the load, as we have limited VRAM in GTA by design.
> 
> Also, even before rendering the image (user pressing F11) it will cause a passive extra load on (V)RAM, just because it gets loaded.

Right now this PR includes a 2048 x 2048 image, which abides by those ideas afaik

Fernando-A-Rocha:
Should we switch to png?

TracerDS:
png would weight more but it would also improve the quality of the image

Fernando-A-Rocha:
I'll edit `Client/mods/deathmatch/logic/CRadarMap.cpp` with the new image size accordingly

Fernando-A-Rocha:
We could also change the top and bottom white text's position/size/etc. to make it prettier. What do you guys think?

image

TracerDS:
Adding a border/shadow to the text would be enough i think.

Also that image looks... different with such scaling. It would need to be scaled down to fit on the screen/stretch it out (maybe not)

Fernando-A-Rocha:
> Adding a border/shadow to the text would be enough i think.
> 
> Also that image looks... different with such scaling. It would need to be scaled down to fit on the screen/stretch it out (maybe not)

I zoomed in to take a screenshot xD The size looks the same IG

TheNormalnij:
Do you request an option in the MTA settings menu? @Dutchman101

Dutchman101:
> png would weight more but it would also improve the quality of the image

It's only weigh in size, in fact png is a more optimized format by its library (less of a performance hit in rendering, for equal resolution & texture, regardless of file size) and is more well-maintained over libjpeg, which is basically ancient and not modern. So it should be PNG for this PR.
@TracerDS

> Do you request an option in the MTA settings menu? @Dutchman101

Yes, i think that since this doubles the texture res, it should be opt-in like described earlier.. it'll generate a performance hit regardless of being low enough to avoid those Intel GPU limits. Either in the video settings or advanced settings tab. In this PR, you can also exchange the standard texture (1024 x 1024) to your new texture, but just in a 50% upscale way to make it also 1024 x 1024, to keep visual parity.
@Fernando-A-Rocha

So, if you follow all requests, simply put: you'd end up pushing 2 PNG's, one of 1024 and one of 2048 (getting rid of the current jpg one in the process).

Fernando-A-Rocha:
All right. Do you have a suggestion for the best way to convert both the current and new jpg images to png? @Dutchman101

Fernando-A-Rocha:
> > @Patrick2562, can we license your work under GPL3?
> 
> Yes, of course

I have converted the 2 map images from JPG to PNG. Do you think it's alright? If not, please, could you provide them?

gptrk0:
> > > @Patrick2562, can we license your work under GPL3?
> > 
> > 
> > Yes, of course
> 
> I have converted the 2 map images from JPG to PNG. Do you think it's alright? If not, please, could you provide them?

Converted and downscaled from 8K image and optimized with _zopfli-png_: [mta_map.zip](https://github.com/user-attachments/files/17531699/mta_map.zip)

Fernando-A-Rocha:
Perfect, I'll do what Dutchman requested. It'll be nice

Client/core/CSettings.cpp
```diff
@@ -1238,7 +1252,7 @@ void CSettings::CreateGUI()
     vecTemp.fX -= fComboWidth + 15;
 
     // Description label
-    vecTemp.fY = 354 + 10;
+    vecTemp.fY += 10;
```
TracerDS: why was it changed?

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +114,20 @@ CRadarMap::~CRadarMap()
...
 }
 
+// If invalid presetIndex is passed, it will use the first preset
+void CRadarMap::SetMapImage(const int presetIndex)
+{
+    if (m_pRadarImage)
+        delete m_pRadarImage;
+
+    auto [fileName, width, height] = GetRadarImagePreset(presetIndex);
+    m_pRadarImage = g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\" + fileName), NULL, false, width, height,
```
TracerDS: what is the type of `fileName`?

also `nullptr` instead of `NULL`

Client/core/CSettings.cpp
```diff
@@ -1004,12 +1004,26 @@ void CSettings::CreateGUI()
...
+    // Radar map image
+    m_pRadarMapImageLabel = reinterpret_cast<CGUILabel*>(pManager->CreateLabel(pTabAdvanced, _("Radar map image:")));
+    m_pRadarMapImageLabel->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY));
+    m_pRadarMapImageLabel->AutoSize();
+
+    m_pRadarMapImageCombo = reinterpret_cast<CGUIComboBox*>(pManager->CreateComboBox(pTabAdvanced, ""));
+    m_pRadarMapImageCombo->SetPosition(CVector2D(vecTemp.fX + fIndentX, vecTemp.fY - 1.0f));
+    m_pRadarMapImageCombo->SetSize(CVector2D(fComboWidth, 95.0f));
+    m_pRadarMapImageCombo->AddItem(_("1024 x 1024 (Default)"))->SetData((void*)0);
+    m_pRadarMapImageCombo->AddItem(_("2048 x 2048"))->SetData((void*)1);
```
TracerDS: what about a constexpr array with resolutions?

also `(void*)0` and `(void*)1` looks very dangerous. What exactly the `SetData` function expects?

TracerDS (changes requested):

Client/core/CSettings.cpp
Fernando-A-Rocha: 354 became wrong because I added a new setting, moving everything down. This is actually the correct way of determining the last fY, just incrementing 10 after the previous setting line

Client/core/CSettings.cpp
shadylua: Defining resolutions as an array using `constexpr ` is a great idea for both readability and maintainability. Using `(void*)0` and `(void*)1` can be dangerous because these types of pointer conversions are error-prone and difficult to understand. Using a `constexpr ` array and an enum to express the intent in the code would be much safer and clearer.

shadylua:
and where exactly will this button appear

Client/mods/deathmatch/logic/CRadarMap.cpp
Fernando-A-Rocha: see definition of GetRadarImagePreset

Client/core/CSettings.cpp
Fernando-A-Rocha: image

i copied the others, hmmm let me check.

Fernando-A-Rocha:
> and where exactly will this button appear

New setting:
image

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +118,33 @@ CRadarMap::~CRadarMap()
...
+    if (imageIndex == 0)
+    {
+        width = 1024;
+        height = 1024;
+    }
+    else
+    {
+        width = 2048;
+        height = 2048;
+    }
```
TracerDS: maybe constexpr array?

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +118,33 @@ CRadarMap::~CRadarMap()
     // Don't need to delete the help texts as those are destroyed by the display manager
 }
 
+void CRadarMap::SetMapImage(std::uint32_t imageIndex)
+{
+    if (imageIndex < 0 || imageIndex > 1)
```
TracerDS: uint will never be negative

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +118,33 @@ CRadarMap::~CRadarMap()
...
+    {
+        width = 1024;
+        height = 1024;
+    }
+    else
+    {
+        width = 2048;
+        height = 2048;
+    }
+    fileName.Format("MTA\\cgui\\images\\radar_%d.png", width);
```
TracerDS: `SString fileName("MTA\\cgui\\...", width);`
remove previous declaration

TracerDS (changes requested):

Fernando-A-Rocha:
Ready

@Dutchman101

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +116,24 @@ CRadarMap::~CRadarMap()
...
+{
+    std::uint32_t width, height;
+    if (imageIndex < RADAR_IMAGE_SIZES.size())
+        width = height = RADAR_IMAGE_SIZES[imageIndex];
+    else
+        width = height = RADAR_IMAGE_SIZES[0];
+
+    SString fileName("MTA\\cgui\\images\\radar_%d.png", width);
+
+    if (m_pRadarImage != nullptr)
```
TracerDS: you can do `m_pRadarImage` instead of `m_pRadarImage != nullptr`

TracerDS (approved):

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -646,11 +680,11 @@ void CRadarMap::SetAttachedToLocalPlayer(bool bIsAttachedToLocal)
...
     if (m_bIsAttachedToLocal)
     {
-        m_HelpTextList[0]->SetCaption("Current Mode: Attached to local player");
+        m_HelpTextList[0]->SetCaption(_("Following Player"));
     }
     else
     {
-        m_HelpTextList[0]->SetCaption("Current Mode: Free Move");
+        m_HelpTextList[0]->SetCaption(_("Free Movement"));
     }
```
FileEX: You can remove braces here

Client/core/CSettings.cpp
```diff
@@ -3536,6 +3558,13 @@ void CSettings::SaveData()
     CVARS_SET("dynamic_ped_shadows", bDynamicPedShadows);
     gameSettings->SetDynamicPedShadowsEnabled(bDynamicPedShadows);
 
+    // Radar map image
+    int selectedComboIndex = m_pRadarMapImageCombo->GetSelectedItemIndex();
+    if (selectedComboIndex != -1)
+    {
+        CVARS_SET("radar_map_image", selectedComboIndex);
+    }
```
FileEX: You can remove braces here

Client/core/CSettings.cpp
```diff
@@ -3150,6 +3167,11 @@ void CSettings::LoadData()
     int PriorityClassList[] = {NORMAL_PRIORITY_CLASS, ABOVE_NORMAL_PRIORITY_CLASS, HIGH_PRIORITY_CLASS};
     SetPriorityClass(GetCurrentProcess(), PriorityClassList[CVARS_GET_VALUE<int>("process_priority") % 3]);
 
+    // Radar map image
+    CVARS_GET("radar_map_image", iVar);
+    if (iVar >= 0 && iVar <= 1)
```
FileEX:
```cpp
if (iVar == 0 || iVar == 1)
```

Client/core/CSettings.cpp
shadylua:
```
if (selectedComboIndex != -1)
    CVARS_SET("radar_map_image", selectedComboIndex);
```

Fernando-A-Rocha:
Actually, idk how we missed it but the map alpha (opacity) setting is in the Multiplayer tab under Map Rendering options, so the map image should also be here :-)

I'll put it there

Fernando-A-Rocha:
✅ 

image

Fernando-A-Rocha:
I also refactored the Video tab by removing the useless "General" label which was occupying space at the top. This allowed the Corona rain reflections checkbox to fit under Motion blor.

image

TracerDS (approved):

FileEX (approved):

Client/core/CSettings.cpp
```diff
@@ -411,11 +411,13 @@ void CSettings::CreateGUI()
     m_pCheckBoxCustomizedSAFiles->AutoSize(NULL, 20.0f);
 
     m_pMapRenderingLabel = reinterpret_cast<CGUILabel*>(pManager->CreateLabel(pTabMultiplayer, _("Map rendering options")));
-    m_pMapRenderingLabel->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 29.0f));
+    m_pMapRenderingLabel->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 30));
```
TheNormalnij:
```suggestion
    m_pMapRenderingLabel->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 30.0f));
```

Client/core/CSettings.cpp
```diff
@@ -625,16 +641,13 @@ void CSettings::CreateGUI()
...
-                                               _("Anisotropic filtering:"), _("Anti-aliasing:"), _("Aspect Ratio:"), _("Opacity:"));
+                                               _("Anisotropic filtering:"), _("Anti-aliasing:"), _("Aspect Ratio:"));
 
-    m_pVideoGeneralLabel = reinterpret_cast<CGUILabel*>(pManager->CreateLabel(pTabVideo, _("General")));
-    m_pVideoGeneralLabel->SetPosition(CVector2D(11, 13));
-    m_pVideoGeneralLabel->GetPosition(vecTemp, false);
-    m_pVideoGeneralLabel->AutoSize(NULL, 3.0f);
-    m_pVideoGeneralLabel->SetFont("default-bold-small");
+    vecTemp.fX = 11;
+    vecTemp.fY = 13;
```
TheNormalnij:
```suggestion
    vecTemp.fX = 11.0f;
    vecTemp.fY = 13.0f;
```

Client/core/CSettings.cpp
```diff
@@ -1238,7 +1252,7 @@ void CSettings::CreateGUI()
     vecTemp.fX -= fComboWidth + 15;
 
     // Description label
-    vecTemp.fY = 354 + 10;
+    vecTemp.fY += 15;
```
TheNormalnij:
```suggestion
    vecTemp.fY += 15.0f;
```

Client/core/CSettings.cpp
```diff
@@ -1629,12 +1643,18 @@ void CSettings::UpdateVideoTab()
...
     int iVar = 0;
     CVARS_GET("mapalpha", iVar);
     int iAlphaPercent = ceil(((float)Clamp(0, iVar, 255) / 255) * 100);
     m_pMapAlphaValueLabel->SetText(SString("%i%%", iAlphaPercent).c_str());
     float sbPos = (float)iAlphaPercent / 100.0f;
     m_pMapAlpha->SetScrollPosition(sbPos);
+
+    // Radar map image
+    CVARS_GET("radar_map_image", iVar);
+    if (iVar == 0 || iVar == 1)
```
TheNormalnij: Does CVAR class guarantee that this check is always valid?

Client/core/CSettings.cpp
```diff
@@ -3609,12 +3631,16 @@ void CSettings::SaveData()
         CVARS_SET("update_auto_install", iSelected);
     }
 
-    // Map alpha
+    // Radar map alpha
```
TheNormalnij: `Radar map` can be misinterpreted with SA radar

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -52,35 +54,39 @@ CRadarMap::CRadarMap(CClientManager* pManager)
...
 
-    // Create the radar and local player blip images
-    m_pRadarImage =
-        g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\radar.jpg"), NULL, false, 1024, 1024, RFORMAT_DXT1);
+    // Create the local player blip image
     m_pLocalPlayerBlip = g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath("MTA\\cgui\\images\\radarset\\02.png"));
 
+    // Create the radar map image
+    m_pRadarImage = nullptr;
+    m_radarImageIndex = g_pCore->GetCVars()->GetValue<std::uint32_t>("radar_map_image");
```
TheNormalnij: Use size_t for indexes

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -52,35 +54,39 @@ CRadarMap::CRadarMap(CClientManager* pManager)
...
+    SetMapImage(m_radarImageIndex);
+
     // Create the marker textures
     CreateMarkerTextures();
 
     // Create the text displays for the help text
+    SColorRGBA colorWhiteTransparent = SColorRGBA(255, 255, 255, 200);
+    SColorRGBA colorWhite = SColorRGBA(255, 255, 255, 255);
```
TheNormalnij: const?

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +122,24 @@ CRadarMap::~CRadarMap()
     // Don't need to delete the help texts as those are destroyed by the display manager
 }
 
+void CRadarMap::SetMapImage(std::uint32_t imageIndex)
+{
+    std::uint32_t width, height;
+    if (imageIndex < RADAR_IMAGE_SIZES.size())
+        width = height = RADAR_IMAGE_SIZES[imageIndex];
+    else
+        width = height = RADAR_IMAGE_SIZES[0];
```
TheNormalnij: Is this check necessary? Can be `imageIndex` invalid?

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +122,24 @@ CRadarMap::~CRadarMap()
...
+
+    SString fileName("MTA\\cgui\\images\\radar_%d.png", width);
+
+    if (m_pRadarImage)
+        SAFE_RELEASE(m_pRadarImage);
```
TheNormalnij: It's double check. `m_pRadarImage` should be `unique_ptr`

Client/mods/deathmatch/logic/CRadarMap.h
```diff
@@ -79,6 +81,8 @@ class CRadarMap
     class CClientRadarMarkerManager* m_pRadarMarkerManager;
     class CClientRadarAreaManager*   m_pRadarAreaManager;
 
+    std::uint32_t m_radarImageIndex;
```
TheNormalnij: `size_t`

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -114,6 +122,24 @@ CRadarMap::~CRadarMap()
...
+
+    m_pRadarImage = g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(CalcMTASAPath(fileName), nullptr, false, width, height, RFORMAT_DXT1);
```
TheNormalnij: Can this function throw a error or return `nullptr` in cases of errors? 
What happens if a client is unable to load a texture? Do we need fallback?

Client/mods/deathmatch/logic/CRadarMap.cpp
Fernando-A-Rocha: imageIndex indeed can't be invalid because the value is clamped by the settings script.
I can remove it, it's just a failsafe in case something goes wrong with client setting var

Client/mods/deathmatch/logic/CRadarMap.cpp
Fernando-A-Rocha: What do u mean exactly?

Client/mods/deathmatch/logic/CRadarMap.cpp
Fernando-A-Rocha: Client never failed to load the map texture in years of MTA. There was never a check. Idk if it would happen for bigger textures, hmm...

Client/core/CSettings.cpp
Fernando-A-Rocha: Radar map is the f11 map. It's what it's called everywhere in the code unfortunately, it's not the minimal :/ I could refactor the naming everywhere for something like FullMap or BigMap

Client/core/CSettings.cpp
Fernando-A-Rocha: Cvars get does place the value in the int variable. The value is clamped by the settings script do it's always right.

Fernando-A-Rocha:
i'll be renaming the confusing "radar map" naming to "player map", according to the already existing functions (https://wiki.multitheftauto.com/wiki/IsPlayerMapVisible and https://wiki.multitheftauto.com/wiki/ForcePlayerMap for example) this will be less confusing with the minimap "radar".

Client/core/CSettings.cpp
```diff
@@ -833,6 +846,10 @@ void CSettings::CreateGUI()
     m_pCheckBoxBlur->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 130.0f));
     m_pCheckBoxBlur->AutoSize(NULL, 20.0f);
 
+    m_pCheckBoxCoronaReflections = reinterpret_cast<CGUICheckBox*>(pManager->CreateCheckBox(pTabVideo, _("Corona rain reflections"), true));
+    m_pCheckBoxCoronaReflections->SetPosition(CVector2D(vecTemp.fX, vecTemp.fY + 150.0f));
+    m_pCheckBoxCoronaReflections->AutoSize(NULL, 20.0f);
```
TracerDS: nullptr instead of NULL

Client/mods/deathmatch/logic/CRadarMap.cpp
```diff
@@ -52,35 +54,39 @@ CRadarMap::CRadarMap(CClientManager* pManager)
...
+    // Create the radar map image
+    m_pRadarImage = nullptr;
+    m_radarImageIndex = g_pCore->GetCVars()->GetValue<std::uint32_t>("radar_map_image");
+    SetMapImage(m_radarImageIndex);
+
     // Create the marker textures
     CreateMarkerTextures();
 
     // Create the text displays for the help text
+    SColorRGBA colorWhiteTransparent = SColorRGBA(255, 255, 255, 200);
```
TracerDS: `SColorRGBA colorWhiteTransparent(255,255,255,200);`

Client/mods/deathmatch/logic/CRadarMap.cpp
TracerDS: Like above

Client/mods/deathmatch/logic/CRadarMap.h
```diff
@@ -62,6 +62,8 @@ class CRadarMap
     void ZoomIn();
     void ZoomOut();
 
+    void SetMapImage(std::uint32_t imageIndex);
```
TracerDS: Can this function be noexcept?

TracerDS (changes requested):

Client/mods/deathmatch/logic/CRadarMap.h
Fernando-A-Rocha: idk, continue discussion in https://github.com/multitheftauto/mtasa-blue/pull/3823#discussion_r1821631592

Client/mods/deathmatch/logic/CRadarMap.cpp
Fernando-A-Rocha: I tried, does not work because CTextureItem can't be instantiated with abstract class

Fernando-A-Rocha:
@TheNormalnij we still have the issue of addressing CreateTexture failure which can happen for:
- map image
- player blip image
- map blip images

After all these years, I couldn't find a report of someone complaining textures failed to create.

Still, what solution do you propose?

TracerDS (approved):

Client/core/CSettings.cpp
TheNormalnij: So you can remove this check

Client/mods/deathmatch/ClientCommands.cpp
```diff
@@ -235,154 +235,137 @@ void COMMAND_Screenshot ( const char* szCmdLine )
 }
 */
 
-void COMMAND_RadarMap(const char* szCmdLine)
+void COMMAND_PlayerMap(const char* szCmdLine)
```
TheNormalnij, line 238: oh, i did ask to change comments only. but ok

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -296,7 +296,7 @@ class CClientGame
 
     CBlendedWeather*       GetBlendedWeather() { return m_pBlendedWeather; };
     CNetAPI*               GetNetAPI() { return m_pNetAPI; };
-    CRadarMap*             GetRadarMap() { return m_pRadarMap; };
+    CPlayerMap*             GetPlayerMap() { return m_pPlayerMap; };
```
TheNormalnij:
```suggestion
    CPlayerMap*            GetPlayerMap() { return m_pPlayerMap; };
```

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -694,7 +694,7 @@ class CClientGame
     CNetworkStats*         m_pNetworkStats;
     CSyncDebug*            m_pSyncDebug;
     // CScreenshot*                          m_pScreenshot;
-    CRadarMap*                    m_pRadarMap;
+    CPlayerMap*                    m_pPlayerMap;
```
TheNormalnij:
```suggestion
    CPlayerMap*                   m_pPlayerMap;
```

Client/mods/deathmatch/logic/CPlayerMap.cpp
```diff
@@ -98,26 +102,44 @@ CRadarMap::CRadarMap(CClientManager* pManager)
...
 }
 
-void CRadarMap::DoPulse()
+void CPlayerMap::SetMapImage(std::size_t imageIndex)
+{
+    std::uint32_t width, height;
+    if (imageIndex < MAP_IMAGE_SIZES.size())
+        width = height = MAP_IMAGE_SIZES[imageIndex];
+    else // Fail safe, ideally client settings system should not allow this case
+        width = height = MAP_IMAGE_SIZES[0];
```
TheNormalnij: Is this check necessary?

Client/mods/deathmatch/logic/CPlayerMap.cpp
```diff
@@ -98,26 +102,44 @@ CRadarMap::CRadarMap(CClientManager* pManager)
...
+
+    SString fileName("MTA\\cgui\\images\\map_%d.png", width);
+
+    SAFE_RELEASE(m_mapImageTexture);
```
TheNormalnij, line 126: Use smart pointers

Fernando-A-Rocha: > Use smart pointers

For what variables?

Fernando-A-Rocha: > Use smart pointers

std::unique_ptr does not work with CTextureItem

`'CTextureItem': cannot instantiate abstract class`

This is fine

TheNormalnij: What did you try?

Fernando-A-Rocha: I showed @TracerDS in DMs, making m_mapImageTexture a std::unique_ptr then using make_unique to create the texture. Wouldn't compile with that error

TheNormalnij: You need `reset` to use raw pointer from `CreateTexture`

Client/mods/deathmatch/logic/CRadarMap.cpp
TheNormalnij: Did you check it?

Client/mods/deathmatch/logic/CPlayerMap.cpp
TracerDS: Its not about that. It causes compilation errors due to the class being too abstract/some argument cannot be converted

shadylua: You cannot directly create an instance of the abstract class `CTextureItem `using `std::unique_ptr` because it’s an abstract class. If there’s a concrete subclass of `CTextureItem`, try creating an instance of that subclass using `std::make_unique`

TheNormalnij: > Its not about that. It causes compilation errors due to the class being too abstract/some argument cannot be converted

It works with `reset`

```c++
    auto* newImage =
        g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(...);
    if (newImage)
        m_pRadarImage.reset(newImage);
    else {
    // CreateTexture can return nullptr
    // Handle error
    }
```

shadylua: > > Its not about that. It causes compilation errors due to the class being too abstract/some argument cannot be converted
> 
> It works with `reset`
> 
> ```c++
>     auto* newImage =
>         g_pCore->GetGraphics()->GetRenderItemManager()->CreateTexture(...);
>     if (newImage)
>         m_pRadarImage.reset(newImage);
>     else {
>     // CreateTexture can return nullptr
>     // Handle error
>     }
> ```

Using reset works, but we’re handling a raw pointer here; it would be safer if `CreateTexture `directly returned a `std::unique_ptr`, so we wouldn’t need to use reset. We might consider changing the return type of this function to `std::unique_ptr<CTextureItem>`.

Fernando-A-Rocha: Will not work.
Compilation error that i showed already `'CTextureItem': cannot instantiate abstract class`

Fernando-A-Rocha:
> @TheNormalnij we still have the issue of addressing CreateTexture failure which can happen for:
> 
> * map image
> * player blip image
> * map blip images
> 
> After all these years, I couldn't find a report of someone complaining textures failed to create.
> 
> Still, what solution do you propose?

i made the code handle texture creation in these cases.

This PR is ready.

Dutchman101:
> This PR is ready.

I agree, given how many code reviews you passed, and you've incorporated all feedback as well.
Thanks for improving the map!

Dutchman101:
Revert (crash details) brought up at dev discord @ https://discord.com/channels/801330706252038164/801411628600000522/1318355966126850179

Stacktrace:
Crash location: [CPlayerMap.cpp#L107](https://github.com/multitheftauto/mtasa-blue/blob/2c5cf3226a573637b91d8b255d57113b7043dc28/Client/mods/deathmatch/logic/CPlayerMap.cpp#L107) @ rev 2c5cf32

PR #3824, opened 2024-10-24
Reduce cooldown in createTrayNotification

ghost:
This PR reduces cooldown which exists in createTrayNotification. By default it is 30 seconds, which in my opinion it's too much. For some servers during this time period a lot of things can happen, and scripters might want notify player who has minimized game. Given that player is able to control whether tray notifications should appear or not (via option: **Allow tray balloon notifications**), it shouldn't be a concern in case some server would use it in spammy manner.
option

### DIFF
--- a/Client/core/CTrayIcon.cpp
+++ b/Client/core/CTrayIcon.cpp
@@ -1,6 +1,6 @@
 /*****************************************************************************
  *
- *  PROJECT:     Multi Theft Auto v1.0
+ *  PROJECT:     Multi Theft Auto
  *  LICENSE:     See LICENSE in the top level directory
  *  FILE:        core/CTrayIcon.cpp
  *  PURPOSE:     Create and destroy tray icon for MTA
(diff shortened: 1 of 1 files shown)


botder:
What is so important that you have to bother a player on their desktop every half second, in the worst case (that's the new limit)?

ghost:
> What is so important that you have to bother a player on their desktop every half second, in the worst case (that's the new limit)?

If you have for instance market script, you could use that to notify that item you've listed has sold.
Or whether player has died, it really depends on a server. If 500 ms is not enough, we can raise to 1 second i guess. But having 30 seconds of cooldown seems too much (also given that player can simply disable it - in case server would like to irritate him) in my honest opinion.

botder:
You could also just send one tray notification and make the window flash on the taskbar (setWindowFlashing). Tray notifications are not meant to be used for near real-time notifications. If the user restores the game window, show him the messages, but we shouldn't allow servers to abuse this functionality because players will then disable it alltogether.

PR #3826, opened 2024-10-26
Add clang-format check action

TheNormalnij:
It requests a lot of changes currently. Can be added after the next `win-apply-clang-format.bat` commit


botder:
I don't think this is going to improve the situation. I would rather run the batch file from time to time.

PR #3829, opened 2024-10-26
Convert SVG bitmap to BGRA unpremultiplied

### LINKED ISSUE #3828: SVG colors are definitely wrong
### Describe the bug

The colors in textures created with svgCreate are different from the colors of the exact same texture in .png format. I already brought this up in #3695 5, but this issue is more precise and focuses only on colors.

Using dxSetBlendMode set to 'add' makes the SVG display the correct colors, but this makes it difficult (if not impossible) to manipulate the texture's alpha for animations, for example. I believe this is a bug that should be fixed, as in the current state, SVG is not suitable for use with animations like fade-out/fade-in due to the need to set the blend mode to get accurate colors

Image

### Steps to reproduce

```lua
local w,h = 600, 450;

local svg_bg = svgCreate(w, h, [[
	<svg width="604" height="454" viewBox="0 0 604 454" fill="none" xmlns="http://www.w3.org/2000/svg">
		<rect x="2" y="2" width="600" height="450" rx="10" fill="#21323A" fill-opacity="0.6"/>
		<rect x="1" y="1" width="602" height="452" rx="11" stroke="#30444D" stroke-opacity="0.6" stroke-width="2"/>
	</svg>
]]);
dxSetTextureEdge(svg_bg, 'clamp');

local tex_bg = dxCreateTexture('images/bg.png', 'argb', false, 'clamp');

	dxDrawImage(150, 200, w, h, svg_bg, 0, 0, 0, 0xFFFFFFFF);
	dxDrawImage(150 + w + 100, 200, w, h, tex_bg, 0, 0, 0, 0xFFFFFFFF);
end);
```

Texture .png
Image

### Version

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

Lpsd:
Resolves #3828 

image

```lua
local w,h = 600, 450

local svg_bg = svgCreate(w, h, [[
	<svg width="604" height="454" viewBox="0 0 604 454" fill="none" xmlns="http://www.w3.org/2000/svg">
		<rect x="2" y="2" width="600" height="450" rx="10" fill="#21323A" fill-opacity="0.6"/>
		<rect x="1" y="1" width="602" height="452" rx="11" stroke="#30444D" stroke-opacity="0.6" stroke-width="2"/>
	</svg>
]])

dxSetTextureEdge(svg_bg, 'clamp')
local tex_bg = dxCreateTexture('bg.png', 'argb', false, 'clamp')

	dxDrawImage(150, 200, w, h, svg_bg, 0, 0, 0, 0xFFFFFFFF)
	dxDrawImage(150 + w + 100, 200, w, h, tex_bg, 0, 0, 0, 0xFFFFFFFF)
end)
```
bg.png
bg

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
@@ -66,7 +66,8 @@ void CClientVectorGraphicDisplay::UpdateTexture()
     if (!bitmap.valid())
         return;
 
-    bitmap.convert(2, 1, 0, 3, true);  // convert To BGRA unpremultiplied
+    // convert to BGRA unpremultiplied
+    bitmap.convert(B, G, R, A, true);
 
     // Lock surface
     D3DLOCKED_RECT LockedRect;
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
@@ -67,7 +67,7 @@ void CClientVectorGraphicDisplay::UpdateTexture()
         return;
 
     // convert to BGRA unpremultiplied
-    bitmap.convert(B, G, R, A, true);
+    bitmap.convert(eColorIndex::B, eColorIndex::G, eColorIndex::R, eColorIndex::A, true);
 
     // Lock surface
     D3DLOCKED_RECT LockedRect;
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
@@ -66,6 +66,8 @@ void CClientVectorGraphicDisplay::UpdateTexture()
     if (!bitmap.valid())
         return;
 
+    bitmap.convert(2, 1, 0, 3, true);  // convert To BGRA unpremultiplied
+
     // Lock surface
     D3DLOCKED_RECT LockedRect;
     if (SUCCEEDED(surface->LockRect(&LockedRect, nullptr, D3DLOCK_DISCARD)))
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
@@ -35,7 +35,7 @@ public:
     void Update();
 
     // Color indexes as per lunasvg::Bitmap
-    static enum ColorIndex
+    static enum eColorIndex : int
     {
         R = 0,
         G = 1,
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
@@ -14,6 +14,8 @@ class CClientVectorGraphicDisplay;
 #include "CClientDisplay.h"
 #include "CClientDisplayManager.h"
 
+
+
 class CClientVectorGraphicDisplay final : public CClientDisplay
 {
     friend class CClientDisplayManager;
@@ -33,6 +35,15 @@ public:
     bool HasUpdated() const { return m_bHasUpdated; }
 
     void Update();
+
+    // Color indexes as per lunasvg::Bitmap
+    static enum ColorIndex
+    {
+        R = 0,
+        G = 1,
+        B = 2,
+        A = 3
+    };
 
 private:
     CClientVectorGraphic* m_pVectorGraphic;
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
@@ -14,8 +14,6 @@ class CClientVectorGraphicDisplay;
 #include "CClientDisplay.h"
 #include "CClientDisplayManager.h"
 
-
-
 class CClientVectorGraphicDisplay final : public CClientDisplay
 {
     friend class CClientDisplayManager;


ghost (approved):

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
```diff
@@ -34,6 +34,15 @@ class CClientVectorGraphicDisplay final : public CClientDisplay
 
     void Update();
 
+    // Color indexes as per lunasvg::Bitmap
+    static enum ColorIndex
```
TheNormalnij: enum class

TheNormalnij:
```suggestion
    enum class eColorIndex
```

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
```diff
@@ -66,6 +66,9 @@ void CClientVectorGraphicDisplay::UpdateTexture()
     if (!bitmap.valid())
         return;
 
+    // convert to BGRA unpremultiplied
+    bitmap.convert(B, G, R, A, true);
```
TheNormalnij:
```suggestion
    bitmap.convert(eColorIndex::B, eColorIndex::G, eColorIndex::R, eColorIndex::A, true);
```

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
Lpsd: `enum class` cannot be implicitly converted to int, and the `convert` method from LunaSVG doesn't support it. Even with type specification ie `enum class eColoIndex : int`, this doesn't work. It needs to be a 'classic' enum, so we'll stick with `static enum eColorIndex : int`

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
```diff
@@ -34,6 +34,15 @@ class CClientVectorGraphicDisplay final : public CClientDisplay
...
     void Update();
 
+    // Color indexes as per lunasvg::Bitmap
+    static enum eColorIndex : int
+    {
+        R = 0,
+        G = 1,
+        B = 2,
+        A = 3
+    };
```
TheNormalnij, line 44: `static` has no effect here.
Enum may be private
Values can be be assigned by default here. You don't need all these `= 0`, `= 1`...
It declares R, G, B, A as valid values in CClientVectorGraphicDisplay methods. It's not critical, but bad

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
TracerDS: Does it have to be implicit conversion?

TheNormalnij: it doesn't and looks bad in all cases

TracerDS: Debatable

TheNormalnij:
The next lunasvg version has no `convert` method. Maybe we can optimize and prepare the solution for the future update:

Faster convert function 
```c++
void CClientVectorGraphicDisplay::UnpremultiplyBitmap(Bitmap &bitmap)
{
    auto width = bitmap.width();
    auto height = bitmap.height();
    auto stride = bitmap.stride();
    auto rowData = bitmap.data();

    for (std::uint32_t y = 0; y < height; y++)
    {
        auto data = rowData;
        for (std::uint32_t x = 0; x < width; x++)
        {
            auto &b = data[0];
            auto &g = data[1];
            auto &r = data[2];
            auto &a = data[3];

            if (a != 0)
            {
                r = (r * 255) / a;
                g = (g * 255) / a;
                b = (b * 255) / a;
            }

            data += 4;
        }

        rowData += stride;
    }
}
```

Avoid additional memory allocation and memcpy
```c++
    IDirect3DSurface9* surface = m_pVectorGraphic->GetRenderItem()->m_pD3DRenderTargetSurface;
    if (!surface)
        return;

    // Lock surface
    D3DLOCKED_RECT LockedRect;
    if (SUCCEEDED(surface->LockRect(&LockedRect, nullptr, D3DLOCK_DISCARD)))
    {
        auto surfaceData = static_cast<std::uint8_t*>(LockedRect.pBits);
        auto stride = static_cast<std::uint32_t>(LockedRect.Pitch);

        Bitmap bitmap{surfaceData, pVectorGraphicItem->m_uiSizeX, pVectorGraphicItem->m_uiSizeY, stride};
        svgDocument->render(bitmap);
        UnpremultiplyBitmap(bitmap);

        // Unlock surface
        surface->UnlockRect();
    }
```

Lpsd:
Just making sure status is known, as this was discussed on Discord - I tried the above code but unfortunately the size of the resulting texture seems to be slightly off, by a few pixels? I'll get some screenshots later, but in this case it looks like drawing a 103x103 rect to a 100x100 canvas at 0x0.

Maybe there are differences in texture generation from `render` compared to `renderToBitmap`, I don't see anything wrong in the calculations for `UnpremultiplyBitmap` or how it is being drawn to the device.

Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.h
tederis: In addition to this, eColorIndex pollutes global namespace. You should use enum class or a namespace for it.

FileEX:
Any update?

Lpsd:
The PR works in its current form. Other maintainers wanted to change approach as the method isn't available in next LunaSVG version, but I haven't looked at this at all since my last comment.

Lpsd:
(feel free to pick this up and investigate based on my previous comments)

PR #3831, opened 2024-10-29
Add LOD management functions

Fernando-A-Rocha:
> [!IMPORTANT]
> **H**LOD = High Level of Detail (normal models the player sees when near)
> **L**LOD = Low Level of Detail (models with less detail the player sees when far)

## Details

Adds new clientside model utility & Level-Of-Detail(LOD) management functions

WIP

To be used in conjunction with:
- [`createObject`](https://wiki.multitheftauto.com/wiki/CreateObject) / [`createBuilding`](https://wiki.multitheftauto.com/wiki/CreateBuilding)
- [`setLowLODElement`](https://wiki.multitheftauto.com/wiki/SetLowLODElement)
- [`engineSetModelLODDistance`](https://wiki.multitheftauto.com/wiki/EngineSetModelLODDistance)

This solves the problem of having to include a Lua LOD_TABLE in every map's script, which has been happening for years ([the map editor's LOD table](https://github.com/multitheftauto/mtasa-resources/blob/fdded32c7504d5052c4ae260726533b6bbb51952/%5Beditor%5D/editor_main/server/mapEditorScriptingExtension_s.lua#L78) happens to be incomplete as of submitting this PR), providing a simple solution to obtain an object's LLOD model ID in the code.

Related to https://github.com/multitheftauto/mtasa-resources/pull/556

## Examples

- LLOD model of HLOD model [ganghous03_LAx (3655)](https://dev.prineside.com/en/gtasa_samp_model_id/model/3655-ganghous03_LAx/) is lodganghous03_lax (3656).

- Practical code examples:
WIP


Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
```diff
@@ -9,6 +9,29 @@
...
+    else
+    {
+        std::uint32_t lodModel = CLodModels::GetObjectLODModel(objectID);
+        if (lodModel == 0) // LOD Model not found for Object Model provided
+            lua_pushnil(luaVM);
+        else
+            lua_pushnumber(luaVM, lodModel);
+    }
+    return 1;
+}
```
TracerDS: Use ArgumentParser

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
```diff
@@ -74,6 +74,7 @@ class CLuaFunctionDefs
     // Util functions to make scripting easier for the end user
     // Some of these are based on standard mIRC script funcs as a lot of people will be used to them
     LUA_DECLARE(GetValidPedModels);
+    LUA_DECLARE(GetObjectLODModel);
```
TracerDS: above

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -196,6 +196,7 @@ void CLuaManager::LoadCFunctions()
 
         // Util functions
         {"getValidPedModels", CLuaFunctionDefs::GetValidPedModels},
+        {"getObjectLODModel", CLuaFunctionDefs::GetObjectLODModel},
```
TracerDS: above

Server/mods/deathmatch/logic/luadefs/CLuaFunctionDefs.Server.cpp
```diff
@@ -16,13 +16,36 @@
...
+    else
+    {
+        std::uint32_t lodModel = CLodModels::GetObjectLODModel(objectID);
+        if (lodModel == 0)            // LOD Model not found for Object Model provided
+            lua_pushnil(luaVM);
+        else
+            lua_pushnumber(luaVM, lodModel);
+    }
+    return 1;
+}
```
TracerDS: above

Server/mods/deathmatch/logic/luadefs/CLuaFunctionDefs.cpp
```diff
@@ -76,7 +76,7 @@ void CLuaFunctionDefs::LoadFunctions()
             // Utility
             {"getVersion", CLuaFunctionDefs::GetVersion}, {"getNetworkUsageData", CLuaFunctionDefs::GetNetworkUsageData},
             {"getNetworkStats", CLuaFunctionDefs::GetNetworkStats}, {"getLoadedModules", CLuaFunctionDefs::GetModules},
-            {"getModuleInfo", CLuaFunctionDefs::GetModuleInfo},
+            {"getModuleInfo", CLuaFunctionDefs::GetModuleInfo}, {"getObjectLODModel", CLuaFunctionDefs::GetObjectLODModel},
```
TracerDS: above

Server/mods/deathmatch/logic/luadefs/CLuaFunctionDefs.h
```diff
@@ -117,6 +117,7 @@ class CLuaFunctionDefs : public CLuaDefs
     LUA_DECLARE(GetVersion);
     LUA_DECLARE(GetModules);
     LUA_DECLARE(GetModuleInfo);
+    LUA_DECLARE(GetObjectLODModel);
```
TracerDS: above

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,493 @@
...
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+#include "CLodModels.h"
+
+// This associates Object Model ID to its Object-LOD Model ID, according to the game's IPL files (including the binary ones).
+
+std::unique_ptr<const std::map<std::uint32_t, std::uint32_t>> CLodModels::LOD_MODELS =
+    std::make_unique<const std::map<std::uint32_t, std::uint32_t>>(std::map<std::uint32_t, std::uint32_t>{
```
TracerDS: why unique_ptr?
Just store a constexpr map to it:
```cpp
const std::unordered_map<std::uint32_t, std::uint32_t> LOD_MODELS = {
// ...
};

```

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,31 @@
...
+public:
+    static std::uint32_t GetObjectLODModel(std::uint32_t objectID)
+    {
+        const auto it = LOD_MODELS->find(objectID);
+        if (it != LOD_MODELS->end())
+        {
+            return it->second;
+        }
+        return 0;
+    };
```
TracerDS: ~~MapFind + define the function in the cpp file~~

After internal discussions this is not needed

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,31 @@
...
+
+private:
+    // Static pointer to a map that is initialized in the .cpp file
+    static std::unique_ptr<const std::map<std::uint32_t, std::uint32_t>> LOD_MODELS;
```
TracerDS: ~~No point for unique_ptr~~

see above

TracerDS (changes requested):

Shared/mods/deathmatch/logic/CLodModels.cpp
Fernando-A-Rocha: constexpr complained there was too much data (the array is huge), this is perfect in terms of memory management

Shared/mods/deathmatch/logic/CLodModels.h
Fernando-A-Rocha: optimization

Shared/mods/deathmatch/logic/CLodModels.cpp
TracerDS: it will be allocated either way

Fernando-A-Rocha: ok let me try

TracerDS: after internal discussions unique_ptr is necessary here due to the amount of data. Stack will be filled so a heap need to be used here

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.Util.cpp
Fernando-A-Rocha: Ok i will

FileEX:
The lod arrays from the map editor are incomplete, so I assume this function is also incomplete. Therefore, a note should be added on the wiki once this PR is merged

Fernando-A-Rocha:
> The lod arrays from the map editor are incomplete, so I assume this function is also incomplete. Therefore, a note should be added on the wiki once this PR is merged

No. I completed the table. It's good now.
See the mtasa-resources issue I linked where I explain it.

Fernando-A-Rocha:
Finally done. Ready for review

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
```diff
@@ -16,6 +16,7 @@ class CLuaFunctionDefinitions;
 #include "CLuaMain.h"
 #include "CLuaTimerManager.h"
 
+
```
TracerDS: Unnecessary new line

Client/mods/deathmatch/logic/lua/CLuaManager.cpp
```diff
@@ -13,6 +13,7 @@
 #include "../luadefs/CLuaFireDefs.h"
 #include "../luadefs/CLuaClientDefs.h"
 #include "../luadefs/CLuaVectorGraphicDefs.h"
+#include "../luadefs/CLuaDefs.h"
```
TracerDS: Unnecessary include

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -721,3 +726,19 @@ bool CLuaObjectDefs::IsObjectRespawnable(CClientEntity* const pEntity) noexcept
...
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID)
+{
+    std::uint32_t lodModel = CLodModels::GetObjectLODOfModel(objectID);
+    if (lodModel == 0) // LOD Model not found for Object Model provided
+        return false;
+    return lodModel;
+}
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectModelOfLOD(lua_State* const luaVM, std::uint32_t lodID)
```
TracerDS: Mark as noexcept

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -333,3 +338,19 @@ bool CLuaObjectDefs::IsObjectRespawnable(CObject* const pObject) noexcept
...
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID)
+{
+    std::uint32_t lodModel = CLodModels::GetObjectLODOfModel(objectID);
+    if (lodModel == 0)            // LOD Model not found for Object Model provided
+        return false;
+    return lodModel;
+}
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectModelOfLOD(lua_State* const luaVM, std::uint32_t lodID)
```
TracerDS: Above

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.h
```diff
@@ -39,4 +39,8 @@ class CLuaObjectDefs : public CLuaDefs
     LUA_DECLARE(StopObject);
     static bool BreakObject(CObject* const pObject);
     static bool ToggleObjectRespawn(CObject* const pObject, const bool bRespawn) noexcept;
+
+    // Object util funcs
+    static std::variant<bool, std::uint32_t> GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID);
+    static std::variant<bool, std::uint32_t> GetObjectModelOfLOD(lua_State* const luaVM, std::uint32_t objectID);
```
TracerDS: Above

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,24 @@
...
+
+class CLodModels
+{
+public:
+    static std::uint32_t GetObjectLODOfModel(std::uint32_t objectID);
+    static std::uint32_t GetObjectModelOfLOD(std::uint32_t lodID);
+
+private:
+    // Static pointer to a map that is initialized in the .cpp file
+    static std::unique_ptr<const std::map<std::uint32_t, std::uint32_t>> LOD_MODELS;
```
TracerDS: Use unordered_map instead.
Also remove const from the map and add it to unique_ptr:
```cpp
static const std::unique_ptr<std::unordered_map<std::uint32_t, std::uint32_t>> lodModels;
```

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,512 @@
...
+
+std::uint32_t CLodModels::GetObjectModelOfLOD(std::uint32_t lodID)
+{
+    for (const auto& [objectID, lodModel] : *LOD_MODELS)
+    {
+        if (lodModel == lodID)
+            return objectID;
+    }
+    return 0;
+}
```
TracerDS: Maybe move these functions above the map for better readability?

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,512 @@
...
+{
+    const auto it = LOD_MODELS->find(objectID);
+    if (it != LOD_MODELS->end())
+        return it->second;
+    return 0;
+}
+
+std::uint32_t CLodModels::GetObjectModelOfLOD(std::uint32_t lodID)
+{
+    for (const auto& [objectID, lodModel] : *LOD_MODELS)
```
TracerDS: `LOD_MODELS.get()` instead of `*` for better readability maybe

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <map>
+#include <memory>
+#include <cstdint>
+
+class CLodModels
+{
+public:
+    static std::uint32_t GetObjectLODOfModel(std::uint32_t objectID);
+    static std::uint32_t GetObjectModelOfLOD(std::uint32_t lodID);
```
TracerDS: Noexcept

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -721,3 +726,19 @@ bool CLuaObjectDefs::IsObjectRespawnable(CClientEntity* const pEntity) noexcept
 
     return pObject->IsRespawnEnabled();
 }
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID)
```
FileEX: You don't need luaVM here

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
FileEX: You don't need luaVM here

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.h
```diff
@@ -43,4 +43,8 @@ class CLuaObjectDefs : public CLuaDefs
     LUA_DECLARE(ToggleObjectRespawn);
     LUA_DECLARE(SetObjectMass);
     LUA_DECLARE(SetObjectProperty);
+
+    // Object util funcs
+    static std::variant<bool, std::uint32_t> GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID);
+    static std::variant<bool, std::uint32_t> GetObjectModelOfLOD(lua_State* const luaVM, std::uint32_t lodID);
```
FileEX: You don't need luaVM here

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -333,3 +338,19 @@ bool CLuaObjectDefs::IsObjectRespawnable(CObject* const pObject) noexcept
 {
     return pObject->IsRespawnEnabled();
 }
+
+std::variant<bool, std::uint32_t> CLuaObjectDefs::GetObjectLODOfModel(lua_State* const luaVM, std::uint32_t objectID)
```
FileEX: You don't need luaVM here

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
FileEX: You don't need luaVM here

Server/mods/deathmatch/logic/luadefs/CLuaObjectDefs.h
FileEX: You don't need luaVM here

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,24 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
```
FileEX: Remove "v1.0"

FileEX (changes requested):

Shared/mods/deathmatch/logic/CLodModels.h
Fernando-A-Rocha: i tried this, it uses too much stack bytes

Shared/mods/deathmatch/logic/CLodModels.cpp
Fernando-A-Rocha: this is fine in terms of readability.

Fernando-A-Rocha:
all ok imo

Shared/mods/deathmatch/logic/CLodModels.h
TracerDS: which one? unique_ptr or unordered_map vs map?

Fernando-A-Rocha: let me check

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -10,501 +10,507 @@
 #include "StdInc.h"
 #include "CLodModels.h"
 
+ // This associates Object Model ID to its Object-LOD Model ID, according to the game's IPL files (plaintext + binary).
+
+static const std::size_t OBJ_LOD_MODELS_COUNT = 4282;
+
+static constexpr std::array<std::pair<std::uint32_t, std::uint32_t>, OBJ_LOD_MODELS_COUNT> OBJ_LOD_MODELS = {
```
TracerDS: constexpr + remove static

TracerDS (changes requested):

lotsofs:
The provided LOD model list not only is incomplete, it also contains plain incorrections. Examples are the LLODs for HLOD models 10428, 10369, 10439.

Fernando-A-Rocha:
> The provided LOD model list not only is incomplete, it also contains plain incorrections. Examples are the LLODs for HLOD models 10428, 10369, 10439.

I'm surprised, I thought the list was generated from the game files not missing anything.

Will you suggest a fix for the array?

ghost:
> I'm surprised, I thought the list was generated from the game files not missing anything.
> 
> Will you suggest a fix for the array?

Just a note:
Even the game .ipl files are different in other ports (ps2, xbox etc). For example the lanterns and special objects around four dragon casino is only persistent on ps2/xbox. https://imgur.com/MV61hvH
The objects does exist but never used in game. 
I don't know if .ide files are different, they shouldn't but who knows.

Fernando-A-Rocha:
I'm gonna investigate the HLOD => LLOD array. @lotsofs @spooky-spook

> > I'm surprised, I thought the list was generated from the game files not missing anything.
> > Will you suggest a fix for the array?
> 
> Just a note: Even the game .ipl files are different in other ports (ps2, xbox etc). For example the lanterns and special objects around four dragon casino is only persistent on ps2/xbox. https://imgur.com/MV61hvH The objects does exist but never used in game. I don't know if .ide files are different, they shouldn't but who knows.

It's interesting that PS2 and XBOX versions of the games have different world mapping, but I think this is out of the scope of this LLOD-improvements PR.

> https://tcrf.net/Grand_Theft_Auto:_San_Andreas/Version_and_Platform_Differences
> There are a large number of objects in Las Venturas missing from the PC version, mainly outside the Four Dragons Casino and The Visage. This seems to be a simple omission from the map as they are still in the game.

Fernando-A-Rocha:
@lotsofs 

There is a function that we can use in MTA to obtain a building (world object)'s LLOD model ID: [`processLineOfSight`](https://wiki.multitheftauto.com/wiki/ProcessLineOfSight)

MTA Map Editor makes use of this to find any building's LLOD model so it can be removed from the map along with its HLOD model, using removeWorldModel.

> The provided LOD model list not only is incomplete, it also contains plain incorrections. Examples are the LLODs for HLOD models 10428, 10369, 10439.

I found that, for example, 10369'S LLOD model is 10511.

image

Both these HLOD and LLOD models are defined in the game IPL files:

image

image

10511 is defined in a plaintext IPL "SFs.ipl", while 10369 is in a binary IPL that I converted to text "sfs_stream0.ipl"

😃 According to these 2 images, you can see that 10369's IPL line has "128" in the last position, which means that its LLOD model is the object instantiated at index 128 (indexes start at 0) of SFs. The LLOD 10511 is instantiated at exactly index 128 in SFS.ipl. The only annoying fact is that they are in different files that both start with sfs.

---

So, yes, something went wrong creating the LODs table, I will look into generating a correct one.

ghost:
> It's interesting that PS2 and XBOX versions of the games have different world mapping, but I think this is out of the scope of this LLOD-improvements PR.

Don't get me wrong I didn't mean that you should fix this :) It was just a note that game files are different on some platforms.

We cannot be sure on how much port has differences. Not only xbox, ps2 and pc. The newer versions mainly use the xbox version, but who knows if they left out something OR re-added missing objects for example. (They could also added new objects, object-placements which were missing from ps2/xbox but were existed in dev files.)

The xbox and ps2 persions created at the same time, but ps2 was the superior. Yet they still could have made differences in xbox's .ipl and .ide files for some reason. (very likely they didn't) PC is ported from xbox, and they cut of a lot of objects from the .ipl files there who knows why. But that doesn't mean that these objects not even exist, they can be used, they still have LLODs as well but never used in the game. So for someone who wants to collect the LLOD ids by deleting objects in map editor is not ideal if the objects newer placed at all. 

Problem is if the ports also have different .ide files, then the objects doesn't even exist. (Very likely the .ide files are not touched, only .ipl files are)
However this wouldn't affect the LLOD table at all since these objects newer existed in the game files of PC port. (or at least the PC port which MTA is based on)

And btw, the site you give me is very good, I didn't read it fully but I already read some things which I didn't know and I'm always interested to read about this game, especially if its new to me.

I'd like to point out that these sites, and youtube videos etc are mainly based on very dedicated person's self observations when playing the game. So like nobody had the chance to have all main ports game files at their side and make comparisons between them, they just noticed these changes/differences by playing :)

What i wrote now has nothing to do with this PR and it's just a side note.
If one read this, and wants to 'fix' this problem then he/she should get all game files (.ipl and ide if we talk about objects) of every platform (mainly ps2, xbox and pc) and make the file comparisons.
Then if its done, then the missing object placements from .ipl file could be added into the `mapfixes` resource. 
But comparing these files are not the hard work, getting them it is.

Fernando-A-Rocha:
@derxgbb Thank you for the observations. Indeed, we could work on improving the `mapfixes` resource to add missing objects etc!

I didn't mind you mentioning this stuff on this PR as it is related to the theme of LODs. :-)

Fernando-A-Rocha:
> The provided LOD model list not only is incomplete, it also contains plain incorrections. Examples are the LLODs for HLOD models 10428, 10369, 10439.

I made a few experiments and [a script to generate the new HLOD=>LLOD table](https://github.com/Fernando-A-Rocha/mta-binary-ipl-to-text/blob/main/server/lods.lua) (fixed, no missing models or incorrect assignments) directly from games files. I've commited the newly generated file! Now, with model names and IPL file names!

https://github.com/multitheftauto/mtasa-blue/blob/81c971a72ed4a1052c00157d4c7062ef9b1b3947/Shared/mods/deathmatch/logic/CLodModels.cpp#L18-L29

⚠️ However, I noticed that the game applies different LODs for 3 cases of the same model, depending on the placed object, for these specific HLOD models:

HLOD 3427 (derrick01):
- LLOD 16274 (oilderricklod01)
- LLOD 16274 (oilderricklod02)
- LLOD 16274 (oilderricklod03)
- LLOD 16274 (oilderricklod04)
- LLOD 16274 (oilderricklod05)
- LLOD 16274 (oilderricklod06)
- LLOD 16274 (oilderricklod07)

HLOD 3303 (des_bighus03):
- LLOD 3373 (lod_d_bighus3_)
- 3372 (lod_d_bighus2_)

HLOD 3724 (laxrf_cargotop):
- LLOD 5164 (lodcargoshp03d)
- LLOD 5352 (lodcargoshp03f)

Here Rockstar applies different LLODs for HLOD 16273 (derrick01), which can be oilderricklod02, oilderricklod03 etc... Also happens for des_bighus03 and laxrf_cargotop. Doesn't happen for any other HLOD models. This seems pointless.

Fernando-A-Rocha:
Upon closer inspection of the models that have multiple LLODs for one HLOD, I conclude that Rockstar was dumb by duplicating DFFs that are all the same both visually and size.

Case of 3427 (derrick01):

image
kdffgui_PWxJSNJPeG
kdffgui_wMh7uSX8HQ

Fernando-A-Rocha:
Case of 3303 (des_bighus03):

The two LLOD models are extremely similar. There is no point having two.

kdffgui_lLddvYyUcx
kdffgui_ZsyWdnvTYo
kdffgui_skkNmkH3yt


Case of 3724 (laxrf_cargotop):

image
image
image

Fernando-A-Rocha:
✅ I conclude that these 3 different models with various super similar LLODs were **mistakes made by the creators**.

I have chosen only 1 LOD for each of these 3 models in CLodModels.cpp (that best match the High Level of Detail model) :D

Fernando-A-Rocha:
In my opinion this is now ready to be implemented. I'd be happy to hear from the people who have commented on this PR previously.

TheNormalnij:
I would suggest to move these functions into a resource. So developers will able to customize the list with their own models

lotsofs:
ProcessLineOfSight is useful, but about .5% of LLODs are not caught by it. Indeed, I have started deleting the entirety of San Andreas in the map editor to compile a list of removeWorldObjects in the .map file which contains both HLOD and LLOD models, and apply manual corrections where applicable. It's about 40% done. It takes a bit of time, but it's also kind of therapeutic so I mind not. Once it's done it can be put side by side with your newly automatic generated list and we can investigate discrepancies. Unfortunately I have neglected to make note of the special cases that required manual correction.

Fernando-A-Rocha:
> Once it's done it can be put side by side with your newly automatic generated list and we can investigate discrepancies. Unfortunately I have neglected to make note of the special cases that required manual correction.

Please comment again when you've compared the table I've generated with yours.

If you look at the comment at the end of each line in [`OBJ_LOD_MODELS` (CLodModels.cpp)](https://raw.githubusercontent.com/multitheftauto/mtasa-blue/81c971a72ed4a1052c00157d4c7062ef9b1b3947/Shared/mods/deathmatch/logic/CLodModels.cpp), you see that **all LLOD model names contain 'lod' in the name**, and generally correspond to a matching HLOD model name.
Exception: [11497 (des_baitshop)](https://dev.prineside.com/en/gtasa_samp_model_id/model/11497-des_baitshop/) => [11615 (desN_baitshop)](https://dev.prineside.com/en/gtasa_samp_model_id/model/11615-desN_baitshop/) doesn't contain 'lod' in the name.
Additionally, these LLOD models are declared in .ide files with different .txd that include lower resolution textures.
e.g.
image
image

I'm sure with high certainty that this table is correct.

Fernando-A-Rocha:
> I would suggest to move these functions into a resource. So developers will able to customize the list with their own models

The benefit of not having a resource is so that new .map generated by the Map Editor can have native LOD support without the need to include a separate resource just for this in the server, thanks to small .lua script included in the map resource (without a HUGE lods lua table).

See https://github.com/multitheftauto/mtasa-resources/pull/556

I can easily make a "set function" to "customize this list" in this c++ PR. Let me try.

Fernando-A-Rocha:
Done, check it out @TheNormalnij

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,4355 @@
...
+    {13761, 13762},            // cunte_whisky => lodcunte_whisky (LAhills)
+    {13784, 13836},            // road_hill01 => lodroad_hill01 (LAhills)
+    {13789, 13851},            // roads40_ce => lodroads40_ce (LAhills)
+    {13795, 13796},            // ce_hillseast06 => lodtehillseast06 (LAhills)
+    {13801, 13880},            // cunte_skatetrak => lod_cunte_skatetrak (LAhills)
+    {13804, 13808},            // cuntelandf4 => lodcuntelandf4 (LAhills)
+    {13805, 13807},            // celalandbiv => lodcelalandbiv (LAhills)
+    {13809, 13811},            // ce_grndpalcst02 => lodcepalcst02 (LAhills)
+    {13810, 13812},            // ce_grndpalcst05 => lodce_grndpalcst05 (LAhills)
+    {13813, 13815},            //
```
TheNormalnij: Use hashmap  here to avoid O(n)

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,27 @@
...
+{
+public:
+    static std::uint32_t GetObjectLowLODModel(std::uint32_t hLODModel) noexcept;
+    static std::uint32_t GetObjectHighLODModel(std::uint32_t lLODModel) noexcept;
+
+    static void SetObjectCustomLowLODModel(std::uint32_t hLODModel, std::uint32_t lLODModel) noexcept;
+    static void          ResetObjectCustomLowLODModel(std::uint32_t hLODModel) noexcept;
+    static void          ResetAllObjectCustomLowLODModels() noexcept;
+
+protected:
```
TheNormalnij:
```suggestion
private:
```

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
```diff
@@ -38,6 +39,13 @@ void CLuaObjectDefs::LoadFunctions()
         {"toggleObjectRespawn", ToggleObjectRespawn},
         {"setObjectMass", SetObjectMass},
         {"setObjectProperty", SetObjectProperty},
+
+        // Object util functions
+        {"getObjectLowLODModel", ArgumentParser<GetObjectLowLODModel>},
+        {"getObjectHighLODModel", ArgumentParser<GetObjectHighLODModel>},
+        {"setObjectCustomLowLODModel", ArgumentParser<SetObjectCustomLowLODModel>},
+        {"resetObjectCustomLowLODModel", ArgumentParser<ResetObjectCustomLowLODModel>},
+        {"resetAllObjectCustomLowLODModels", ArgumentParser<ResetAllObjectCustomLowLODModels>},
```
TheNormalnij: Maybe
`getModelLowLOD`
`getModelHighLODModel`
...

Shared/mods/deathmatch/logic/CLodModels.cpp
TheNormalnij: You should reset the list before connecting to a new server

Shared/mods/deathmatch/logic/CLodModels.h
```diff
@@ -0,0 +1,27 @@
...
+
+class CLodModels
+{
+public:
+    static std::uint32_t GetObjectLowLODModel(std::uint32_t hLODModel) noexcept;
+    static std::uint32_t GetObjectHighLODModel(std::uint32_t lLODModel) noexcept;
+
+    static void SetObjectCustomLowLODModel(std::uint32_t hLODModel, std::uint32_t lLODModel) noexcept;
+    static void          ResetObjectCustomLowLODModel(std::uint32_t hLODModel) noexcept;
+    static void          ResetAllObjectCustomLowLODModels() noexcept;
```
TheNormalnij:
```suggestion
    static void SetObjectCustomLowLODModel(std::uint32_t hLODModel, std::uint32_t lLODModel) noexcept;
    static void  ResetObjectCustomLowLODModel(std::uint32_t hLODModel) noexcept;
    static void  ResetAllObjectCustomLowLODModels() noexcept;
```

Shared/mods/deathmatch/logic/CLodModels.cpp
Fernando-A-Rocha: I tested, it is resetted when changing server.

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
Fernando-A-Rocha: i disagree, this notation is the clearest

get Object Low Level of Detail Model (high level of detail model)
get Object High Level of Detail Model (low level of detail model)
...

TheNormalnij: But this function works with buildings models too.

Fernando-A-Rocha: > But this function works with buildings models too.

Buildings use object models. Theres no such things as building models, so this makes sense imo

Kinimel:
> In my opinion this is now ready to be implemented. I'd be happy to hear from the people who have commented on this PR previously.

I totally agree, and I think it would be a great idea to implement a new system in the MTA and that it would be a great novelty for server developers or admins.

Fernando-A-Rocha:
Any more feedback?

Kinimel:
> Any more feedback?

No but no feedback beyond that as I thought it was a great idea from your point of view.

Fernando-A-Rocha:
Updated main post with code example usages.

TracerDS (approved):

PlatinMTA:
``setObjectCustomLowLODModel()`` wont change current LLODs for world objects, right? It would be nice to have a function that did that, so you could for example add a LLOD to the palm trees or the street lamps without needing to manually remove and add them one by one as a building / object element.

Fernando-A-Rocha:
> `setObjectCustomLowLODModel()` wont change current LLODs for world objects, right? It would be nice to have a function that did that, so you could for example add a LLOD to the palm trees or the street lamps without needing to manually remove and add them one by one as a building / object element.

`setObjectCustomLowLODModel(hlod, llod)` defines a LLOD ID for a given HLOD ID in a dictionary, for developers to use. That's it.

World object LOD is actually not managed by any Lua functions currently. So, as you said, if you want to customize a palm tree's LLOD, you need to use removeWorldModel to remove both the HLOD and the LLOD of that tree, create a custom object/building with HLOD model there, then create another object/building with the LLOD model, and finally set it as the low lod element of the high lod element.

Or you can do use engineReplaceModel to replace a palm tree Low LOD model, and the result will be seen on all world object palm trees part of the default map.

PlatinMTA:
> `setObjectCustomLowLODModel(hlod, llod)` defines a LLOD ID for a given HLOD ID in a dictionary, for developers to use. That's it.
> 
> World object LOD is actually not managed by any Lua functions currently. So, as you said, if you want to customize a palm tree's LLOD, you need to use removeWorldModel to remove both the HLOD and the LLOD of that tree, and create a custom object/building with HLOD model there, then create another object/building with the LLOD model, and finally set it as the low lod element of the high lod element.
> 
> Currently, if you want to replace world object models, you can do with engineReplaceModel. So if you want to replace a palm tree Low LOD model, you can do it, and the result will be seen on all world object palm trees part of the default map.

Yeah but the thing is, palm trees dont have LLODs, thats why I made that question

But I think thats out of the scope of this PR, thanks for the response Fer

Fernando-A-Rocha:
We could have `setObjectCustomLowLODModel(int hlod, int llod, bool updateWorldObjects)` to reflect the low lod model changes of default world objects automatically.

So you could change [CJ's house](https://dev.prineside.com/en/gtasa_samp_model_id/model/17697-carlshou1_LAe2/)'s LLOD model from lod1carlshou1_lae (17768) to something else, and it would update the world/landscape object automatically in Grove Street.

Fernando-A-Rocha:
Someone could want to change [CJ's house](https://dev.prineside.com/en/gtasa_samp_model_id/model/17697-carlshou1_LAe2/)'s LLOD model from lod1carlshou1_lae (17768) to something else, while automatically updating the world/landscape object in Grove Street, without having to mess with the object/building system.

Automatic LLOD creation & assignment is a problem. There are 3 different object types in the world:

- default/landscape "world objects" that aren't MTA elements, and can be removed with `removeWorldModel`
- "object" elements
- "building" elements

As a developer, I'd love to have a function that automatically creates and assigns a LLOD object to a given or all objects with a certain HLOD model in my server.

We could have a function to do automatic LOD management for world objects.

For object/building elements the way to create and set LLODs is the following:

image

We can think about how we would improve this process.

> But I think thats out of the scope of this PR, thanks for the response

Oh I'm always happy to discuss, I think it's related to this topic and we can do lil brainstorming.

Fernando-A-Rocha:
I can also make `setObjectCustomLowLODModel(hlod, llod)` automatically the models of the game's LLOD landscape/world "buildings"(objects).

It'd be useful for projects that a lot of LLOD models to objects that don't have LOD in GTA SA, giving the game a new look and feel.

E.g. https://www.mixmods.com.br/2021/12/lod-vegetation-distant-trees/

Opinions?

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
TheNormalnij: There is no such thing as object models too.  The game engine uses different model types like atomic, clump, timed, vehicle, ped, etc. 
getObjectLowLODModel says me that this function works with object elements.
For vehicles, we are using `getVehicleModel*` function names.

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,4376 @@
...
+    {13789, 13851},            // roads40_ce => lodroads40_ce (LAhills)
+    {13795, 13796},            // ce_hillseast06 => lodtehillseast06 (LAhills)
+    {13801, 13880},            // cunte_skatetrak => lod_cunte_skatetrak (LAhills)
+    {13804, 13808},            // cuntelandf4 => lodcuntelandf4 (LAhills)
+    {13805, 13807},            // celalandbiv => lodcelalandbiv (LAhills)
+    {13809, 13811},            // ce_grndpalcst02 => lodcepalcst02 (LAhills)
+    {13810, 13812},            // ce_grndpalcst05 => lodce_grndpalcst05 (LAhills)
+    {13813, 13815},            // cegraveblok03e => lodcegraveblok03e (LAhills)
+    {13814, 13787},            // ceroadn => lodceroadn (LAhills)
+    {13816, 13877},            // ce_safeground => ce_s
```
TheNormalnij: How does mta reset the list after reconnecting?

TheNormalnij: Also, engineFreeModel should remove ID from the list.

TheNormalnij:
> Opinions?

I would prefer a simple solution with `createObjectWithLod` and `createBuildingWithLod`.
And i don't recommend using createObject instead of createBuilding for world models with LODs.
Objects doesn't have lod support in GTA. MTA uses dirty LOD support for objects

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,4376 @@
...
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        mods/shared_logic/CLodModels.cpp
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+#include "CLodModels.h"
+
+constexpr std::size_t OBJ_LOD_MODELS_COUNT = 4289;
+constexpr std::pair<std::uint32_t, std::uint32_t> OBJ_LOD_MODELS_ARRAY[] = {
```
TheNormalnij, line 14: You don't need this array. Just define a hashmap

Shared/mods/deathmatch/logic/CLodModels.cpp
TheNormalnij: 0 is a valid CJ model. I think MTA has INVALID_MODEL macro already

Fernando-A-Rocha:
> I would prefer a simple solution with `createObjectWithLod` and `createBuildingWithLod`. And i don't recommend using createObject instead of createBuilding for world models with LODs. Objects doesn't have lod support in GTA. MTA uses dirty LOD support for objects

createObjectWithLod and createBuildingWithLod sound good!

I am still in favor of implementing the new LOD ID management functions described the main post. They are useful for getting the LLOD id of an HLOD id, and also setting custom LLOD ids for custom systems.

Also, what do you think about making `setObjectCustomLowLODModel(hlod, llod)` automatically update the models of the game's LLOD landscape/world "buildings"/"objects" ? @TheNormalnij

Shared/mods/deathmatch/logic/CLodModels.cpp
Fernando-A-Rocha: ok i will do object ID validation

Fernando-A-Rocha: > Also, engineFreeModel should remove ID from the list.

This actually makes sense

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
Fernando-A-Rocha: Object models are a thing, they are also called [Object IDs](https://wiki.multitheftauto.com/wiki/Object_IDs)

We don't use any function getVehicleModel...

I could rename to `getLowLODObjectModel` ?

Fernando-A-Rocha: LOD is a concept that only exists for objects in GTA:SA hmmm i think vehicles also have lod, but its within the same .dff

Kinimel:
> Também posso criar `setObjectCustomLowLODModel(hlod, llod)`automaticamente os modelos dos "edifícios" (objetos) do cenário/mundo LLOD do jogo.
> 
> Seria útil para projetos com muitos modelos LLOD para objetos que não têm LOD no GTA SA, dando ao jogo uma nova aparência.
> 
> Por exemplo https://www.mixmods.com.br/2021/12/lod-vegetation-distant-trees/
> 
> Opiniões?

The new LLOD ID management functions are a big step towards a more robust and adaptable system. The possibility of customizing LLOD IDs for specific systems offers a high degree of control. The setObjectCustomLowLODModel function is particularly interesting. By automatically updating LLOD models, it not only simplifies development, but also helps to avoid visual inconsistencies in the game.

Kinimel:
> I can also make `setObjectCustomLowLODModel(hlod, llod)` automatically the models of the game's LLOD landscape/world "buildings"(objects).
> 
> It'd be useful for projects that a lot of LLOD models to objects that don't have LOD in GTA SA, giving the game a new look and feel.
> 
> E.g. https://www.mixmods.com.br/2021/12/lod-vegetation-distant-trees/
> 
> Opinions?

The new LLOD ID management functions are a big step towards a more robust and adaptable system. The possibility of customizing LLOD IDs for specific systems offers a high degree of control. The setObjectCustomLowLODModel function is particularly interesting. By automatically updating LLOD models, it not only simplifies development, but also helps to avoid visual inconsistencies in the game.

PlatinMTA:
> I can also make `setObjectCustomLowLODModel(hlod, llod)` automatically the models of the game's LLOD landscape/world "buildings"(objects).

yes that would be nice... although im guessing that function is also useful for custom models. My request was surely really specific so I'm not sure everyone would like that... maybe we could have a third argument that if its ``true`` then we can make that function behave like described (changing all lods?).

I'm down to test that change if needed because i think it would be pretty cool. Otherwise the only way to implement loads of LLODs to existing world buildings would be to delete them all (reading IPL files, parsing those) and create new ones. Second option seems to be too problematic/not worth it.

Out of the scope of this PR but maybe we should have a way to retrieve all world objects in some way. Right now if you want to find a world object you need to use ``processLineOfSight`` or similar, unless something was added and I didn't notice it. Otherwise you can get your IPL files and parse them yourself.

Fernando-A-Rocha:
> > I can also make `setObjectCustomLowLODModel(hlod, llod)` automatically the models of the game's LLOD landscape/world "buildings"(objects).
> 
> yes that would be nice... although im guessing that function is also useful for custom models. My request was surely really specific so I'm not sure everyone would like that... maybe we could have a third argument that if its `true` then we can make that function behave like described (changing all lods?).
> 
> I'm down to test that change if needed because i think it would be pretty cool. Otherwise the only way to implement loads of LLODs to existing world buildings would be to delete them all (reading IPL files, parsing those) and create new ones. Second option seems to be too problematic/not worth it.
> 
> Out of the scope of this PR but maybe we should have a way to retrieve all world objects in some way. Right now if you want to find a world object you need to use `processLineOfSight` or similar, unless something was added and I didn't notice it. Otherwise you can get your IPL files and parse them yourself.

Sure, I'd like to add the feature to automatically update world object LLODs.

For finding all world objects, I have parsed all of the game IPLs. This is enough tbh, and you can easily use Lua scripts for your needs. 

I put everything here, including the binary ipls I converted to text. https://github.com/Fernando-A-Rocha/mta-binary-ipl-to-text

Fernando-A-Rocha:
Check it out so far, ive redone some functions. Moved things to model related files. Added isValidModel (general purpose func)

@TheNormalnij I added model id checks, and also made it reset in engineFreeModel and when model manager is unloaded. Is this ok?

Client/mods/deathmatch/logic/luadefs/CLuaObjectDefs.cpp
TheNormalnij: > We don't use any function getVehicleModel...


[getVehicleModelDummyDefaultPosition](https://wiki.multitheftauto.com/wiki/GetVehicleModelDummyDefaultPosition)
[getVehicleModelDummyPosition](https://wiki.multitheftauto.com/wiki/GetVehicleModelDummyPosition)
[getVehicleModelExhaustFumesPosition](https://wiki.multitheftauto.com/wiki/GetVehicleModelExhaustFumesPosition)
[getVehicleModelWheelSize](https://wiki.multitheftauto.com/wiki/GetVehicleModelWheelSize)

TheNormalnij: objects have no lod support in SA. MTA uses custom implementation for them.

Fernando-A-Rocha: > objects have no lod support in SA. MTA uses custom implementation for them.

What are you implying?

Client/mods/deathmatch/logic/luadefs/CLuaModelDefs.cpp
```diff
@@ -0,0 +1,93 @@
...
+        {"resetModelLODByLow", ArgumentParser<ResetModelLODByLow>},
+        {"resetAllModelLOD", ArgumentParser<ResetAllModelLOD>},
+    };
+
+    // Add functions
+    for (const auto& [name, func] : functions)
+        CLuaCFunctions::AddFunction(name, func);
+}
+
+bool CLuaModelDefs::IsValidModel(std::string modelPurpose, std::uint32_t id)
```
TracerDS, line 38: what about an enum instead of string?

Client/mods/deathmatch/logic/luadefs/CLuaModelDefs.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+std::variant<bool, std::uint32_t> CLuaModelDefs::GetModelHighLOD(std::uint32_t lLODModel) noexcept
+{
+    return CLodModels::GetModelHighLOD(lLODModel);
+}
+
+bool CLuaModelDefs::SetModelLOD(std::string modelPurpose, std::uint32_t hLODModel, std::uint32_t lLODModel)
+{
+    if (!(modelPurpose == "object" || modelPurpose == "building"))
```
TracerDS, line 68: `if(modelPurpose != "object" && modelPurpose != "building")`

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,4388 @@
...
+    {694, 784},                // sm_redwoodgrp => lod_redwoodgrp (countryS)
```
TracerDS, line 15: maybe too much spacing before the comment?
`{X, Y}, // comment`
instead of
`{X, Y},         // comment`

Shared/mods/deathmatch/logic/CLodModels.cpp
```diff
@@ -0,0 +1,4388 @@
...
+    {13789, 13851},            // roads40_ce => lodroads40_ce (LAhills)
+    {13795, 13796},            // ce_hillseast06 => lodtehillseast06 (LAhills)
+    {13801, 13880},            // cunte_skatetrak => lod_cunte_skatetrak (LAhills)
+    {13804, 13808},            // cuntelandf4 => lodcuntelandf4 (LAhills)
+    {13805, 13807},            // celalandbiv => lodcelalandbiv (LAhills)
+    {13809, 13811},            // ce_grndpalcst02 => lodcepalcst02 (LAhills)
+    {13810, 13812},            // ce_grndpalcst05 => lodce_grndpalcst05 (LAhills)
+    {13813, 13815},            // cegraveblok03e => lodcegraveblok03e (LAhills)
+    {13814, 13787},            // ceroadn => lodceroadn (LAhills)
+    {13816, 13877},            // ce_safeground => ce_s
```
TracerDS: do we really need that?

TracerDS (changes requested):

ghost:
Any other news? This PR is interesting!

ghost:
These functions should be shared, like [setLowLODElement](https://wiki.multitheftauto.com/wiki/SetLowLODElement) is. e.g: if you would like to create object on server-side, and get it's LOD model directly.

ghost:
What is missing for this PR to be completed?

ghost:
Any updates on this? @Fernando-A-Rocha

Fernando-A-Rocha:
I'll get back to this PR soon.

There are several ways to approach the problem of LODs. We certainly need something that is both clientside and serverside due to createObject, createBuilding (soon available serverside) and setLowLodElement being shared functions...

lotsofs:
I am back with my manually built list of HLOD vs LLODs, using the removeWorldObject feature from the editor on every object in the map, and where that failed finding a manual match using the IDE list on the wiki or similar.
[MTASA Maps - Sheet17.csv](https://github.com/user-attachments/files/18323067/MTASA.Maps.-.Sheet17.csv)


Here are my conclusions, and more importantly, things that I think should be changed in the big table in CLodModels.cpp:


**Missing entries**
These are all objects that gave me trouble deleting them in the editor, their LLOD would still show after deletion, and I had to manually find their LLODs in the IDE list. They are also missing from the table in CLodModels.cpp
```
    {10289, 10156},            // tempsf_3_sfe => LODA27 (???)
    {10561, 10681},            // bbgroundbitc_SFS => LODroundbitc_SFS (???)
    {10933, 11275},            // traintrax01_SFS => LOD_trax_SFSe07 (???)
    {10973, 11025},            // mall_03_SFS => LOD_sfs014 (???)
    {11228, 11277},            // traintrax01b_SFS => LOD_trax_SFSe08 (???)
    {11231, 11273},            // traintrax03b_SFS => LOD_trax_SFSe01 (???)
    {11232, 11274},            // traintrax03c_SFS => LOD_trax_SFSe06 (???)
    {11409, 11347},            // roadsSFSE34 => LODroadsSFSE34 (???)
    {17614, 17843},            // Lae2_landHUB02 => LODLae2_lndHUB02 (???)
    {4826, 4987},              // grifftop2 => LODfftop03 (???)
    {7498, 7843},              // vegaswrailroad01 => lodaswrailroad01 (???)
    {7582, 7951},              // miragebuild04 => lodagebuild04 (???)
    {7625, 7957},              // vgnhseing129 => lodhseing129 (???)
    {7878, 7879},              // vegasNroad242 => hubst4alpha (???)
    {8386, 8387},              // vgsSspagjun09b => lodSspagjun09b (???)
    {8388, 8389},              // vegasSedge29b => lodasSedge29b (???)
    {9310, 9390},              // chapel_SFN => LODpel_SFN (???)
    {9889, 9890},              // park3a_sfw => lod_park3a_sfw (???)
    {9239, 9405},              // track01_SFN => LODck01_SFN (???)
    {9240, 9404},              // track02_SFN => LODck02_SFN (???)
```



**HLODs using the same LLOD**
These objects all use LLOD 17749, in addition to the already included HLOD 17559:
```
    {17556, 17749},            // MStorCP1_LAe2 => LOD1MStorCP_LAe (???)
    {17557, 17749},            // MStorCP2_LAe2 => LOD1MStorCP_LAe (???)
    {17558, 17749},            // MStorCP4_LAe2 => LOD1MStorCP_LAe (???)
```



**multiple LLODs belonging to the same HLOD**
These objects all appear the same HLOD, but have different LLODs for some reason:
Inclusion recommended if it is desired that a user can request a HLOD belonging to a specific LLOD.
(not including entry oilderricklod01 that is already in the table)
```
    {3427, 16274},             // derrick01 => oilderricklod02 (???)
    {3427, 16275},             // derrick01 => oilderricklod03 (???)
    {3427, 16276},             // derrick01 => oilderricklod04 (???)
    {3427, 16277},             // derrick01 => oilderricklod05 (???)
    {3427, 16278},             // derrick01 => oilderricklod06 (???)
    {3427, 16279},             // derrick01 => oilderricklod07 (???)
```
Furthermore there is this entry: but it has other issues (read below)
```
    {3724, 5352},              // laxrf_cargotop => LODcargoshp03f (???)
```




**LLODs that cover multiple HLODs**
Existing entry: 
``    {7243, 6996},              // vgncircus1b => lodcircus1 (vegasN)``
is correct, but LLOD 6996 covers both HLODs 6994 and 7243 combined. The former is an entire casino building, and the latter is only the hedges surrounding its parking lot. A second opinion might be needed on what to do with this, but I'd heavily recommend changing the line to 
``    {6994, 6996},              // vgncircus1b => lodcircus1 (vegasN)``
as it makes much more sense after a visual inspectation. 





Existing entry:
``    {9504, 9634},              // sboxbld4_sfw73 => lodxbld4_sfw73 (SFw)``
is correct, but LLOD 9634 covers both HLODs 9504 and 9505 combined. The former is a big hill with houses on top, the latter is only the grass textures of the tiny gardens belonging to these houses. The line is fine as it is, but in the interest of completeness I thought it worth mentioning here.


**Incorrect entries**
Existing entry:
``    {3412, 3413},              // cunterb03 => lodcunterb07 (countrye)``
is incorrect. HLOD 3412 is only the orange upper parts of a bridge, whereas HLOD 3411 is the entire bridge, including yet even more orange upper parts. LLOD 3413 matches HLOD 3412 much more closely
``    {3411, 3413},              // cunteRB01 => lodcunterb07 (countrye)``


Existing entry:
``    {5786, 5945},              // shutters01_lawn => lodblok08_lawn (LaWn)``
is incorrect. HLOD 5786 comprises just a set of door that go on entire building HLOD 5784 that LLOD 5945 matches much more closely.
``    {5784, 5945},              // MelBlok08_LAwN => lodblok08_lawn (LaWn)``
However, the LLOD seems to be rotated 180 degrees compared to the HLOD 5784. Someone else besides me should look at this and make a judgement call.


Actually upon second thought, these are probably yet another cases of LLODs that cover multiple HLODs, but I already marked them as plainly incorrect and I can't be bothered rewriting.


**LLOD to a LLOD**
Existing entries:
``    {7174, 7366},              // lod_corpbuild31 => sham_superlod (vegasN)``
``    {6971, 7174},              // vgn_corpbuild31 => lod_corpbuild31 (vegasN)``
7174 is the LLOD for HLOD 6971, with 7366 in turn being a LLOD (SuperLOD) to 7174. Odd case. both cases are already in the list, so no action is needed I guess, but I had this in my notes and maybe someone has opinions.


**Oddities that require subjective review**
Non exhaustive list of objects that might give issues:



Existing entry
``    {3724, 5164},              // laxrf_cargotop => lodcargoshp03d (LAs2)``
is correct, but the LLOD is rotated 90 degrees compared to its HLOD counterpart. This might give unwanted behavior. This also applies to the LODcargoshp03f I mentioned above, it is also rotated 90 degrees.


I dont have time to write the last entry, so here's my raw notes:
HLOD 9690, LLOD 9686
HLOD 9694, LLOD 9692
This concerns the pillars of Gant bridge. The HLOD 9690 and 9694 are each half a pillar (one the southern, one the northern), placed twice with one rotated 180 degrees so it connects. LLOD 9686 and 9692 is a single model representing an entire pillar. A subjective judgement call is needed for this. Furthermore: Deleting the pillar in the map editor reveals yet another model, which looks somewhat like a skeletal structure of like a character model. Someone should look at what the hell is going on here.


I'd submit my suggested changes myself, but I have no idea how to do a pull request to a pull request.

Fernando-A-Rocha:
Wow thank you for your detailed review @lotsofs
I'll definitely take a deeper look
I've added you to my repository Fernando-A-Rocha/mtasa-blue so you can contribute to this branch directly!! 😃

lotsofs:
I let the invitation expire by accident, but I have managed to submit PRs to your repo for each of my proposed changes. 

=======================================

There are a bunch of LLODs that fit their HLOD rather awkwardly, such as being rotated in some way, or having additional structures like fences etc that are separated from the HLOD counterpart. I named like four above, but out of the thousands other entries there likely are more.

This might lead to some minor visual impurities, but I personally don't think it significant enough to warrant investigation, as LLODs are not meant to be seen from up close either way. However it is something to make note of.

Fernando-A-Rocha:
Thank you for the contributions. I'll take a look!

Dutchman101:
What's the status at this point, @Fernando-A-Rocha ?

Fernando-A-Rocha:
> What's the status at this point, @Fernando-A-Rocha ?

@TheNormalnij says having LOD models table in mtasa-blue is nonsense, and that we should have a Lua resource for maintaing this.

Also, he suggested the addition of function createBuildingWithLOD.
He also says that LOD implementation for "object"s is bad, and should not be used in favor of buildings.

I'm not sure how to proceed

Fernando-A-Rocha:
Closing this in favor of a solution that will be made in Lua, probably in multitheftauto/mtasa-resources

tryangelic:
> Closing this in favor of a solution that will be made in Lua, probably in multitheftauto/mtasa-resources

How will the LOD system be resolved or created for the whole of mta? Since some developers remove all mods from mtasa-resources and create their own

Fernando-A-Rocha:
> > Closing this in favor of a solution that will be made in Lua, probably in multitheftauto/mtasa-resources
> 
> How will the LOD system be resolved or created for the whole of mta? Since some developers remove all mods from mtasa-resources and create their own

Sorry can u explain your question better?

PR #3777, opened 2024-10-07
Update CNickGen.cpp

shadylua:
Bright" -> Removed once.
"Dusty" -> Removed once.
"Lively" -> Removed once.
"Petite" -> Removed once.
"Silent" -> Removed once.
"Sore" -> Removed once.
"Tense" -> Removed once.
"Thirsty" -> Removed once.

Thank you!

PR #3834, opened 2024-10-30
Improve random nick generation

Lpsd:
Fixes issue from #3777 where nickname parts could be "null" due to not updating the constants (which are now removed in place of better code).

### DIFF
--- a/Client/core/CNickGen.cpp
+++ b/Client/core/CNickGen.cpp
@@ -8,10 +8,10 @@
  *****************************************************************************/
 
 #include "StdInc.h"
-#include "time.h"
+#include <random>
 
 // These words are of a maximum length of 10 characters, capitalized, and stripped of whitespace
-const char* const CNickGen::m_szAdjectives[] = {
+const char* const szAdjectives[] = {
     "Aback",      "Abaft",      "Abandoned",  "Abashed",    "Aberrant",   "Abhorrent",  "Abiding",    "Abject",     "Ablaze",     "Able",       "Abnormal",
     "Aboard",     "Aboriginal", "Abortive",   "Abounding",  "Abrasive",   "Abrupt",     "Absent",     "Absorbed",   "Absorbing",  "Abstracted", "Absurd",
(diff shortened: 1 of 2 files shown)


TracerDS:
maybe add the random function to SharedUtils (SharedUtil.random.hpp?) and from there you can call that function in the nick generation func?

shadylua:
like : 

```
#ifndef SHARED_UTIL_RANDOM_HPP
#define SHARED_UTIL_RANDOM_HPP

#include <random>

namespace SharedUtil {
    inline int RandomInt(int min, int max) {
        static std::random_device rd;
        static std::mt19937 gen(rd());
        std::uniform_int_distribution<int> dist(min, max);
        return dist(gen);
    }
}

#endif // SHARED_UTIL_RANDOM_HPP
```

Lpsd:
You don't want to create a new generator for each part of the nickname. In-fact, we should probably share the generator between successive calls.

TracerDS:
Possibly a template there with pragma and optimized code but yeah, something like that

TracerDS:
> You don't want to create a new generator for each part of the nickname. In-fact, we should probably share the generator between successive calls.

We can have a static (maybe const) generator that will be shared with related functions only.
That way it will not be generated again and the overhead would be minimal

Client/core/CNickGen.cpp
```diff
@@ -196,10 +196,18 @@ const char* const CNickGen::m_szNouns[] = {
     "Vampire",    "Parasite",   "Tramp",      "Bum",        "Hobo",       "Hitchhiker", "Deadbeat",   "Acrobat",
 };
 
+const int numAdjectives = std::size(szAdjectives);
+const int numNouns = std::size(szNouns);
+const int maxNum = static_cast<int>(RAND_MAX);
```
TheNormalnij:
```suggestion
constexpr auto numAdjectives = std::size(szAdjectives);
constexpr auto numNouns = std::size(szNouns);
constexpr auto maxNum = 100;
```

TheNormalnij (approved):

Issue #3233, opened 2023-11-02
Fix setVehicleDirtLevel


Haxardous:
### Describe the bug

https://wiki.multitheftauto.com/wiki/SetVehicleDirtLevel

> Function has been disabled.
> Reason/Note: This function doesn't work actually. See the bugtracker for updates and more information.

### Steps to reproduce

setVehicleDirtLevel (theVehicle, 100)

### Version

Client: Multi Theft Auto v1.6-release-22279

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

TracerDS:
Sadly, this function wont ever work without making our own implementation.
Accrding to the [gta wiki](https://gta.fandom.com/wiki/Car_Wash#GTA_Vice_City_&_GTA_San_Andreas):
> When accumulated dirt was introduced on vehicles in [Grand Theft Auto: San Andreas](https://gta.fandom.com/wiki/Grand_Theft_Auto:_San_Andreas), car washes were believed to be featured, but the feature was never included in the final version of the game

Thus we cannot call any gta function that sets the dirt level because such function doesn't exist.
Maybe `CVehicleModelInfo::SetDirtTextures` might do something but its unlikely anything will happen.

Maybe the dirt level can be recreated via textures?

TheNormalnij:
Are you sure?
Vehicles is SA can be dirty, but the game doesn't provide a way to clean them
GTA SA 1.01 and silent path fix that.

[video ](https://www.youtube.com/watch?v=2WCChKCzG5E)

TracerDS:
> Are you sure? Vehicles is SA can be dirty, but the game doesn't provide a way to clean them GTA SA 1.01 and silent path fix that.
> 
> [video ](https://www.youtube.com/watch?v=2WCChKCzG5E)

With silent patch, yes. I looked over gta mem addresses but I didnt find any function that referenced or changed dirt level in any way.

TheNormalnij:
[gta-reversed](https://github.com/gta-reversed/gta-reversed-modern) has some related to dirt level code. There are documented and reversed functions.

TracerDS:
> [gta-reversed](https://github.com/gta-reversed/gta-reversed-modern) has some related to dirt level code. There are documented and reversed functions. 

No. The only functions related to dirt level are constructors for dirt textures but not to make car dirty.
[...]

Issue #772, opened 2019-01-13
Enforce use of () instead of (void)


qaisjp:
**Describe the solution you'd like**

This should be added to win-apply-clang-format.bat

**Additional context**

Response to #771

jushar:
I'd prefer using _clang-tidy_ with the _modernize-redundant-void-arg_ option.
See: https://clang.llvm.org/extra/clang-tidy/checks/modernize-redundant-void-arg.html

botder:
Adding *clang-tidy* would be great.

qaisjp:
The blocker for clang-tidy client-side is this:

(see [conversation on #mta.dev](https://discordapp.com/channels/278474088903606273/366384007535001612/687419701466628096))

> clang-cl doesn't work with multiplayer_sa and game_sa
> due to inline asm
> ...
> clang doesn't support non-asm in asm functions
> or well
> _declspec(naked) functions
> ![image](https://user-images.githubusercontent.com/923242/78807130-8bd41a00-79bb-11ea-8c24-df44def73c3e.png)
>
> the clean way to replace all inline asm would be to replace the functions entirely
> rather than the current variant of just randomly patching parts of functions

I've made a [crappy script](https://github.com/qaisjp/mtasa-suitup) which finds all the problematic methods that needs to be fixed.

Gist output here: https://gist.github.com/qaisjp/50b0d8ab4833b77c5b9e0305f974f756

qaisjp:
https://github.com/multitheftauto/mtasa-blue/pull/4611 excludes `__asm` from directives from clang-tidy, and now `clang-format` is enforced in CI.

https://github.com/multitheftauto/mtasa-blue/pull/4653 introduces support to download the `clang-tidy` tool in our formatting pipeline, but fails with errors like this:

```
PS C:\Users\qaisjp\source\repos\mtasa-blue> .\Build\tmp\clang-tidy.exe -checks=modernize-redundant-void-arg --fix Client\mods\deathmatch\logic\CClientModel.cpp
Error while trying to load a compilation database:
Could not auto-detect compilation database for file "Client\mods\deathmatch\logic\CClientModel.cpp"
No compilation database found in C:\Users\qaisjp\source\repos\mtasa-blue\Client\mods\deathmatch\logic or any parent directory
fixed-compilation-database: Error while opening fixed database: no such file or directory
json-compilation-database: Error while opening JSON database: no such file or directory
Running without flags.
1 error generated.
[...]

PR #3837, opened 2024-11-02
Fix setVehicleDirtLevel

FileEX:
This PR fixes the long-standing non-functional ``setVehicleDirtLevel`` function. The code is based on the source code of SilentPatch by CookiePL, also known as Silent.

image
image


Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+    {
+        pushad
+        push ebx
+        push esi
+        call SetDirtTextures
+        add esp, 8
+        popad
+
+        jmp CONTINUE_CVehicleModelInfo_SetDirtTextures
+    }
```
TracerDS: can you make the function non-naked?

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+static void _declspec(naked) HOOK_CVehicleModelInfo_SetClump()
+{
+    _asm
+    {
+        pushad
+        call FindEditableMaterialList
+        popad
+
+        jmp CONTINUE_CVehicleModelInfo_SetClump
+    }
```
TracerDS: above

TracerDS (changes requested):

Fernando-A-Rocha:
Woww!! Thank you for this. Finally we can start having the amazing GTA:SA fixes the community has created over time, on MTA

Fernando-A-Rocha:
Fixes #3233

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
FileEX: Nope

TracerDS: Why not?

FileEX: Why should I? I don't see any benefits right now.

1. The MultiplayerSA project is on C++14, so it makes no difference.
2. Naked hooks are more readable than non-naked ones where we mix asm with C++ (in my humble opinion).
3. Hooks with mixed asm and C++ can be problematic (see https://github.com/multitheftauto/mtasa-blue/issues/772#issuecomment-611048892).
4. I like naked hooks.

TracerDS: 1. We would want to move on to C++17 or C++20. Naked functions prevent us from doing so
2. You dont need to mix it with C++ at all. Naked functions just dont have the header and footer of a function. Thats all. As far as I see you are manually pushing registers either way. Naked function is unnecessary there
3. Look up
4. bruh

PlatinMTA:
Could we also get a getVehicleDirtLevel? Just because

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
FileEX: I think it’s important to consider that these hooks are often inserted somewhere in the middle of the code, and a non-naked function has its own function prologue, allocation, register saving, and its own function epilogue. In the case of such hooks, there are usually unnecessary processor instructions, especially in places like rendering functions. Personally, I believe that if a function can be naked for hooking, that option should always be chosen. For now, I adhere to the current standards of the C++14 project and do not intend to make this change, at least until such a requirement comes from the staff. If it’s really important to you, feel free to open your own PR that eliminates all naked functions, and then I will have to adjust my PR to the new standards

TracerDS: Allocation and register saving? No.
Only prologue and epilogue.
There are already too much naked functions. Introducing even more wont be helpful in a long run.

In this case naked function is unnecessary as the same effect can be done with regular function. For other functions this is complicated.

FileEX:
> Could we also get a getVehicleDirtLevel? Just because

Yeah but in separate PR

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
FileEX: > Allocation and register saving? No.
Only prologue and epilogue.

I meant the stack, my mistake.

Fernando-A-Rocha:
> Could we also get a getVehicleDirtLevel? Just because

Makes sense to have. For example in a custom car wash system where you calculate the price based on the dirt

Client/game_sa/CModelInfoSA.h
```diff
@@ -290,7 +302,18 @@ class CVehicleModelInfoSAInterface : public CBaseModelInfoSAInterface
...
+    {
+        struct
+        {
+            RpMaterial** m_dirtMaterials;
+            size_t       m_numDirtMaterials;
+            RpMaterial*  m_staticDirtMaterials[30];
+        };
+        RpMaterial* pDirtMaterial[32];
+    };
+
```
TheNormalnij: This struct has a different size now. 
Old size: 64 bytes
New size: 128 bytes

Did you break the next data offsets?

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+    CVehicleModelInfoSAInterface* mi = nullptr;
+    _asm
+    {
+        mov mi, ecx
+    }
+
+    if (!mi)
+        return;
+
+    ((void(__cdecl*)(void*))FUNC_CBaseModelInfo_Shutdown)(mi);
```
TheNormalnij: Add a method in `CBaseModelInfoSAInterface`

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
     }
 }
 
+static void CVehicleModelInfo_Shutdown()
```
TheNormalnij: static void  __fastcall CVehicleModelInfo_Shutdown(CVehicleModelInfoSAInterface* mi)

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+{
+    CVehicleModelInfoSAInterface* mi = nullptr;
+    _asm
+    {
+        mov mi, ecx
+    }
+
+    if (!mi)
+        return;
```
TheNormalnij: You don't need that

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+static void _declspec(naked) HOOK_CVehicleModelInfo_SetDirtTextures()
+{
+    _asm
+    {
+        pushad
+        push ebx
+        push esi
+        call SetDirtTextures
+        add esp, 8
+        popad
```
TheNormalnij: pushad and popad are slow operations. Try to avoid them

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+        mi->m_dirtMaterials = new RpMaterial*[mi->m_numDirtMaterials];
+        std::copy(list.begin(), list.end(), mi->m_dirtMaterials);
+    }
+    else
+    {
+        mi->m_dirtMaterials = nullptr;
+        std::copy(list.begin(), list.end(), mi->m_staticDirtMaterials);
+    }
+
+    ((void(__thiscall*)(CVehicleModelInfoSAInterface*))FUNC_CVehicleModelInfo_ResetColors)(mi);
```
TheNormalnij: Add a method in `CVehicleModelInfoSAInterface`

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -118,4 +224,9 @@ void CMultiplayerSA::InitHooks_Vehicles()
 {
     EZHookInstall(CDamageManager__ProgressDoorDamage);
     EZHookInstall(CAEVehicleAudioEntity__Initialise);
+
+    // Fix vehicle dirt level
+    EZHookInstall(CVehicleModelInfo_SetClump);
+    EZHookInstall(CVehicleModelInfo_SetDirtTextures);
+    MemCpy((void*)0x85C5E4, &CVehicleModelInfo_Shutdown, 4);
```
TheNormalnij, line 225: MemPut

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -107,6 +114,105 @@ static void _declspec(naked) HOOK_CAEVehicleAudioEntity__Initialise()
...
+#define HOOKPOS_CVehicleModelInfo_SetClump 0x4C9648
+#define HOOKSIZE_CVehicleModelInfo_SetClump 24
+static constexpr DWORD CONTINUE_CVehicleModelInfo_SetClump = 0x4C9660;
+static void _declspec(naked) HOOK_CVehicleModelInfo_SetClump()
+{
+    _asm
+    {
+        pushad
+        call FindEditableMaterialList
+        popad
```
TheNormalnij: Are pushad and popad necessary here?

Client/game_sa/CModelInfoSA.h
FileEX: Yes. None of the following fields are used anywhere, but it still shouldn’t be like this. What do you suggest?

The structure in the union also affects the size of the interface

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
FileEX: Is there any difference? From what I can see, MemPut still calls MemCpy anyway

Client/game_sa/CModelInfoSA.h
TheNormalnij: Can you fill the interface with right info?

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
TheNormalnij: did you forget to push?

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
TheNormalnij: > 1. We would want to move on to C++17 or C++20. Naked functions prevent us from doing so

They don't. Only functions with mixed c++/asm code affected.

TracerDS: > > 1. We would want to move on to C++17 or C++20. Naked functions prevent us from doing so
> 
> They don't. Only functions with mixed c++/asm code affected.

They do. C++17 changed a little bit how naked functions work. Dont remember the exact details but with C++17 naked functions are not compilable with the current mtasa configuration.

TheNormalnij: > C++17 changed a little bit how naked functions work.

MSVC uses a new asm code generation algorithm since c++17.
The old algorithm allocates variables in the stack with using `push` opcode.
The new algorithm expects that stack was allocated in a function prolog and uses `mov [ebp-NUM], something`.
Access to unallocated stack position causes crashes. 

Solutions:
1) Remove affected code
2) Move some code into separate functions to avoid stack usage in naked functions
3) Use [__LOCAL_SIZE ](https://learn.microsoft.com/en-us/cpp/cpp/considerations-for-writing-prolog-epilog-code?view=msvc-170)to allocate stack

Only functions with mixed c++/asm code are affected. Last time i checked, the `Game SA` project was compilable with c++17.

TracerDS: Mixed asm/c++ code is fine.
Hell, even c++ standard offers such capabilities.
Naked functions can only contain asm code.
No naked functions can contain both.
Time for testing them again ig

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
FileEX: I look forward to your replies in the remaining open conversations

Client/game_sa/CModelInfoSA.h
FileEX: Yes, but structure also affects class size. How to solve this? Fields needed include numMaterials, staticMaterials etc

TheNormalnij: Remove `char pad4[64];` and the struct will have right size

TheNormalnij: Remove `pDirtMaterial` too.

Client/game_sa/CModelInfoSA.h
```diff
@@ -259,53 +265,74 @@ class CTimeModelInfoSAInterface : public CBaseModelInfoSAInterface
...
+    union
+    {
+        struct
+        {
+            RpMaterial** m_dirtMaterials;
+            size_t       m_numDirtMaterials;
+            RpMaterial*  m_staticDirtMaterials[30];
+        };
+    };
+
```
TheNormalnij:
```suggestion
    RpMaterial** m_dirtMaterials;
    std::size_t       m_numDirtMaterials;
    RpMaterial*  m_staticDirtMaterials[30];
```

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
TheNormalnij: It's the same, but without a size argument

lopezloo:
Does it fix `setVehicleDirtLevel` function only or it makes vehicles to gain dirt over time too?

FileEX:
> Does it fix `setVehicleDirtLevel` function only or it makes vehicles to gain dirt over time too?

Probably both cases

Dutchman101:
Nice, ty

Jator-1:
This PR probably broke backward compatibility of some shaders. Also after recent update (ver. 22829) texture of vehiclegrunge256 are not correctly displayed on vehicles and components.

Vehicle components probably use now different materials or texture and they are not shown up on shader_tex_names (displayed only on 'chassis' of ZR-350, missing on doors, bonnet, rear bumper, extra1 and whole Premier)
image
image



ver. 22789 - shader works perfectly fine
image (1)
ver. 22829 - shader does not work on vehicle components as bonnet and boot resulting in different shades of paint
image

PR #3838, opened 2024-11-02
Fix for randomly bright objects after weapon change

FileEX:
This PR fixes an old bug where some objects randomly become bright when switching weapons from unarmed to any other

## Example
**Without weapon**
image

**With weapon**
image

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Weapons.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Weapons.cpp
@@ -220,10 +220,8 @@ static void _declspec(naked) HOOK_CVisibilityPlugins_RenderWeaponPedsForPC()
         mov eax, 5533B0h
         mov ecx, ebx
 
-        pushad
         push 0
         call eax // call CPed::RemoveLighting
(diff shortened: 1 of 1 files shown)


ghost (approved):

Client/multiplayer_sa/CMultiplayerSA_Weapons.cpp
```diff
@@ -200,6 +200,35 @@ void _declspec(naked) HOOK_Fx_AddBulletImpact()
...
+        mov eax, 5DF4E0h
+        call eax // call CPed::ResetGunFlashAlpha
+
+        mov eax, 5533B0h
+        mov ecx, ebx
+
+        pushad
+        push 0
+        call eax // call CPed::RemoveLighting
+        popad
```
TheNormalnij: pushad and popad are unnecessary here

FileEX: You're right; that's a leftover from my crash debugging

TheNormalnij (approved):

Issue #3696, opened 2024-09-04
Add synchronize parameter to onElementDataChange


wassmas:
### Is your feature request related to a problem? Please describe.

Cheaters can exploit the lack of synchronization checks in the [onElementDataChange](https://wiki.multitheftauto.com/wiki/OnElementDataChange) event, making it hard to differentiate between legitimate and manipulated data changes.

### Describe the solution you'd like

Add a synchronize parameter to [onElementDataChange ](https://wiki.multitheftauto.com/wiki/OnElementDataChange) to indicate if the change was synchronized or local. This would help identify and prevent cheating

### Additional context

This feature would enhance server security by providing a reliable method to detect suspicious element data changes.

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

ffsPLASMA:
Dont quite understand what you are trying to say. You can set the 4th parameter of setElementData to false if you dont want to sync it from server to client or client to server. And for checking to see if client update elementdata without permission, you could easily build your own system. Check if the client is allowed to change elementdata of key X, if not revert it and take actions. An example can be found here: https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/security/players.lua#L9

wassmas:
> Dont quite understand what you are trying to say. You can set the 4th parameter of setElementData to false if you dont want to sync it from server to client or client to server. And for checking to see if client update elementdata without permission, you could easily build your own system. Check if the client is allowed to change elementdata of key X, if not revert it and take actions. An example can be found here: https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/security/players.lua#L9

And when client set element data not synchronized (only client side) can't be detected by server

ffsPLASMA:
> > Dont quite understand what you are trying to say. You can set the 4th parameter of setElementData to false if you dont want to sync it from server to client or client to server. And for checking to see if client update elementdata without permission, you could easily build your own system. Check if the client is allowed to change elementdata of key X, if not revert it and take actions. An example can be found here: https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/security/players.lua#L9
> 
> And when client set element data not synchronized (only client side) can't be detected by server

[...]

PR #3839, opened 2024-11-03
Add element data client trust options.

TheNormalnij:
Added `elementdata_whitelisted` setting for `mtaserver.conf`
* Enabled: the client can't change element data on server, except allowed keys.
* Disabled: the client can change any element data. (default, unsafe)

Added `clientTrustMode` parameter for` setElementData`

```lua
-- Allow changes from the client
bool setElementData(root,  key, value, syncType, "allow" )

-- Deny changes from the client
bool setElementData(root,  key, value, syncType, "deny" )

-- Respect elementdata_whitelisted setting
bool setElementData(root,  key, value, syncType, "default" )
```

Added event `onPlayerChangesProtectedData`
source: player
arguments:
* element,
* key,
* value

### DIFF
--- a/Server/mods/deathmatch/logic/CCustomData.h
+++ b/Server/mods/deathmatch/logic/CCustomData.h
@@ -50,11 +50,7 @@ public:
 
     bool Delete(const char* szName);
 
-    bool IsClientChangesAllowed(const char* szName) const;
     void SetClientChangesMode(const char* szName, ECustomDataClientTrust mode);
-
-    bool IsClientChangesAllowed() const noexcept { return m_clientChangesAllowed; };
-    void SetClientChangesAllowed(bool enabled) noexcept { m_clientChangesAllowed = enabled; };
 
     unsigned short CountOnlySynchronized();
 
@@ -72,5 +68,4 @@ private:
 
     std::map<std::string, SCustomData> m_Data;
     std::map<std::string, SCustomData> m_SyncedData;
-    bool                               m_clientChangesAllowed{true};
 };
--- a/Server/mods/deathmatch/logic/CCustomData.h
+++ b/Server/mods/deathmatch/logic/CCustomData.h
@@ -25,11 +25,18 @@ enum class ESyncType
     SUBSCRIBE,
 };
 
+enum class ECustomDataClientTrust : std::uint8_t
+{
+    UNSET,
+    ALLOW,
+    DENY,
+};
+
 struct SCustomData
 {
-    CLuaArgument Variable;
-    ESyncType    syncType;
-    bool         allowClientChanges;
+    CLuaArgument           Variable;
+    ESyncType              syncType;
+    ECustomDataClientTrust clientChangesMode;
 };
 
 class CCustomData
@@ -42,8 +49,12 @@ public:
     void         Set(const char* szName, const CLuaArgument& Variable, ESyncType syncType = ESyncType::BROADCAST);
 
     bool Delete(const char* szName);
+
     bool IsClientChangesAllowed(const char* szName) const;
-    void SetClientChangesAllowed(const char* szName, bool enabled);
+    void SetClientChangesMode(const char* szName, ECustomDataClientTrust mode);
+
+    bool IsClientChangesAllowed() const noexcept { return m_clientChangesAllowed; };
+    void SetClientChangesAllowed(bool enabled) noexcept { m_clientChangesAllowed = enabled; };
 
     unsigned short CountOnlySynchronized();
 
@@ -61,4 +72,5 @@ private:
 
     std::map<std::string, SCustomData> m_Data;
     std::map<std::string, SCustomData> m_SyncedData;
+    bool                               m_clientChangesAllowed{true};
 };
--- a/Server/mods/deathmatch/logic/CCustomData.h
+++ b/Server/mods/deathmatch/logic/CCustomData.h
@@ -29,6 +29,7 @@ struct SCustomData
 {
     CLuaArgument Variable;
     ESyncType    syncType;
+    bool         allowClientChanges;
 };
 
 class CCustomData
@@ -36,11 +37,13 @@ class CCustomData
 public:
     void Copy(CCustomData* pCustomData);
 
-    SCustomData* Get(const char* szName);
+    SCustomData* Get(const char* szName) const;
     SCustomData* GetSynced(const char* szName);
     void         Set(const char* szName, const CLuaArgument& Variable, ESyncType syncType = ESyncType::BROADCAST);
 
     bool Delete(const char* szName);
+    bool IsClientChangesAllowed(const char* szName) const;
+    void SetClientChangesAllowed(const char* szName, bool enabled);
 
     unsigned short CountOnlySynchronized();
 
--- a/Server/mods/deathmatch/logic/CCustomData.h
+++ b/Server/mods/deathmatch/logic/CCustomData.h
@@ -25,7 +25,7 @@ enum class ESyncType
     SUBSCRIBE,
 };
 
-enum class ECustomDataClientTrust : std::uint8_t
+enum class eCustomDataClientTrust : std::uint8_t
 {
     UNSET,
     ALLOW,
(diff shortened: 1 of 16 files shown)


TracerDS:
Whats the purpose of these functions?

Server/mods/deathmatch/logic/CCustomData.h
```diff
@@ -25,23 +25,37 @@ enum class ESyncType
     SUBSCRIBE,
 };
 
+enum class ECustomDataClientTrust : std::uint8_t
```
TracerDS: lowercase e? (`eCustomDataClientTrust`)

Server/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -2437,3 +2443,27 @@ int CLuaElementDefs::isElementCallPropagationEnabled(lua_State* luaVM)
...
+
+bool CLuaElementDefs::IsElementDataClientTrusted(CElement* pElement, std::optional<std::string_view> key)
+{
+    if (key.has_value())
+        return pElement->GetCustomDataManager().IsClientChangesAllowed(key.value().data());
+    else
+        return pElement->GetCustomDataManager().IsClientChangesAllowed();
+}
+
+void CLuaElementDefs::ResetElementDataClientTrust(CElement* pElement, std::optional<std::string_view> key)
```
TracerDS: maybe noexcept?

TracerDS (changes requested):

ffsPLASMA:
isnt this kinda redundant with the way you can already detect elementdata changes in lua? thats why I added it to security resource.

https://github.com/multitheftauto/mtasa-resources/blob/master/%5Badmin%5D/security/players.lua#L9

FileEX:
From what I understand, the issue is that if we have an elementData key, like "user:points", which we only use in server-side scripts and it’s not synchronized on the server, a potential cheater could still overwrite it by using setElementData client-side with the sync parameter set to true. My suggestion is to add a new server-side argument, something like ignoreClientSync. Then, in the server’s elementData synchronization packet, if a particular elementData key has the ignoreClientSync parameter set to true, this synchronization packet would simply be ignored, and the value would remain unchanged.

In my opinion, a single simple argument is a much better solution than having entirely separate functions for this

ffsPLASMA:
> From what I understand, the issue is that if we have an elementData key, like "user:points", which we only use in server-side scripts and it’s not synchronized on the server, a potential cheater could still overwrite it by using setElementData client-side with the sync parameter set to true. My suggestion is to add a new server-side argument, something like ignoreClientSync. Then, in the server’s elementData synchronization packet, if a particular elementData key has the ignoreClientSync parameter set to true, this synchronization packet would simply be ignored, and the value would remain unchanged.
> 
> In my opinion, a single simple argument is a much better solution than having entirely separate functions for this

Ye I agree something like:

bool setElementData ( element theElement, string key, var value [, bool synchronize = true ], [bool acceptClientUpdate = true] )

If synchronize is set to false, acceptClientUpdate should also be set to false.

PlatinMTA:
> From what I understand, the issue is that if we have an elementData key, like "user:points", which we only use in server-side scripts and it’s not synchronized on the server, a potential cheater could still overwrite it by using setElementData client-side with the sync parameter set to true. My suggestion is to add a new server-side argument, something like ignoreClientSync. Then, in the server’s elementData synchronization packet, if a particular elementData key has the ignoreClientSync parameter set to true, this synchronization packet would simply be ignored, and the value would remain unchanged.
> 
> In my opinion, a single simple argument is a much better solution than having entirely separate functions for this

What if the elementdata is not set serverside? Should it be ignored by default? I do believe this should be the case but I also think its too late to be making these changes...

FileEX:
> 
> What if the elementdata is not set serverside? Should it be ignored by default? I do believe this should be the case but I also think its too late to be making these changes...

If an element data doesn’t exist on the server side for a given element, like player:vip, the scripter should simply set this element data to false by default rather than leaving it unset. If any element data isn’t used at all on the server side, it doesn’t matter if a potential cheater sets it. Yes, there’s still the issue that a cheater could send a packet manually, bypassing all checks, but I’ve already suggested a potential solution for this to Dutchman. In any case, an argument like acceptClientUpdate would make manipulating element data with executors impossible, as the server would ignore that packet

wassmas:
> From what I understand, the issue is that if we have an elementData key, like "user:points", which we only use in server-side scripts and it’s not synchronized on the server, a potential cheater could still overwrite it by using setElementData client-side with the sync parameter set to true. My suggestion is to add a new server-side argument, something like ignoreClientSync. Then, in the server’s elementData synchronization packet, if a particular elementData key has the ignoreClientSync parameter set to true, this synchronization packet would simply be ignored, and the value would remain unchanged.
> 
> In my opinion, a single simple argument is a much better solution than having entirely separate functions for this

Thats why i request for this: #3696

TheNormalnij:
> it doesn’t matter if a potential cheater sets it

Nope. The data will be synced with all clients. 

> What if the elementdata is not set serverside? Should it be ignored by default?

I'll add a server setting to ignore all unknown element data from clients.

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -984,7 +989,7 @@ bool CStaticFunctionDefinitions::SetElementData(CElement* pElement, const char*
         // Unsubscribe all the players
         if (lastSyncType == ESyncType::SUBSCRIBE && syncType != ESyncType::SUBSCRIBE)
             m_pPlayerManager->ClearElementData(pElement, szName);
-
+
```
TheNormalnij:
```suggestion

```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -984,7 +989,6 @@ bool CStaticFunctionDefinitions::SetElementData(CElement* pElement, const char*
         // Unsubscribe all the players
         if (lastSyncType == ESyncType::SUBSCRIBE && syncType != ESyncType::SUBSCRIBE)
             m_pPlayerManager->ClearElementData(pElement, szName);
```
TheNormalnij:
```suggestion
            m_pPlayerManager->ClearElementData(pElement, szName);

```

ffsPLASMA:
Wouldnt it be better to just outright dismiss all client sync updates to an elementdata if the data wasnt shared to client in first place?

eg. setElementData(root, "coolData", "mreow", false)

^doesnt get synced to client, so why not disregard any client updates for said key in first place?

TheNormalnij:
> Wouldnt it be better to just outright dismiss all client sync updates to an elementdata if the data wasnt shared to client in first place?
> 
> eg. setElementData(root, "coolData", "mreow", false)
> 
> ^doesnt get synced to client, so why not disregard any client updates for said key in first place?

It is not enough. Element data may be broadcasted by server only. In this case the element data is shared, but the server deny all changes from client. Whitelisting is the safest option.

ffsPLASMA:
> > Wouldnt it be better to just outright dismiss all client sync updates to an elementdata if the data wasnt shared to client in first place?
> > eg. setElementData(root, "coolData", "mreow", false)
> > ^doesnt get synced to client, so why not disregard any client updates for said key in first place?
> 
> It is not enough. Element data may be broadcasted by server only. In this case the element data is shared, but the server deny all changes from client. Whitelisting is the safest option.

I dont quite understand.

When I create an elementdata on server and specifically set it to not get synced to any clients, then the server should not accept/retrieve any updates from client for said key in first place. If I explicitly set an elementdata to get synced to other clients, I can use onElementDataChange to see if said client is allowed to change it or not.

Adding yet another parameter to setElementData to pay attention to and a server config setting is kinda nonsense.

TheNormalnij:
I agree that local serverside element data may be protected by default.
But broadcasted element data requests protection too

Dutchman101:
Thanks!
Regarding desirability: the PR has been requested by AC team. You don't have to use it if you don't want to.

Fernando-A-Rocha:
Cool, script security wiki article should be updated

PR #3844, opened 2024-11-07
Add to accept 4x4 lua table in setElementBoneMatrix function

W3lac3:
This PR will allow you to pass 4x4 lua tables in the parameter of the setElementBoneMatrix function, currently it only accepts one Matrix, which makes it difficult to use this function.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
@@ -1072,10 +1072,43 @@ std::variant<bool, CLuaMultiReturn<float, float, float, float>> CLuaPedDefs::Get
     return std::make_tuple(x, y, z, w);
 }
 
-bool CLuaPedDefs::SetElementBoneMatrix(lua_State* const luaVM, CClientPed* entity, std::uint32_t boneId, CMatrix boneMatrix)
+bool CLuaPedDefs::SetElementBoneMatrix(lua_State* const luaVM)
 {
-    CEntity* theEntity = entity->GetGameEntity();
-    return theEntity ? theEntity->SetBoneMatrix(static_cast<eBone>(boneId), boneMatrix) : false;
+    CClientEntity*     entity = NULL;
+    std::uint32_t      boneId;
+    CMatrix            boneMatrix;
+
+    CScriptArgReader argStream(luaVM);
+    argStream.ReadUserData(entity);
+    argStream.ReadNumber(boneId);
+
+    if (argStream.NextIsTable())
+    {
+        if (!ReadMatrix(luaVM, argStream.m_iIndex, boneMatrix))
+        {
+            argStream.SetCustomError("Matrix is not 4 x 4");
+        }
+    }
+    else
+    {
+        argStream.ReadMatrix(boneMatrix);
+    }
+
+    if (!argStream.HasErrors())
+    {
+        CEntity* theEntity = entity->GetGameEntity();
+
+        if (theEntity->SetBoneMatrix(static_cast<eBone>(boneId), boneMatrix))
+        {
+            lua_pushboolean(luaVM, true);
+            return true;
+        }
+    }
+    else
+        m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
+
+    lua_pushboolean(luaVM, false);
+    return false;
 }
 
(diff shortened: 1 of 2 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -1072,10 +1072,43 @@ std::variant<bool, CLuaMultiReturn<float, float, float, float>> CLuaPedDefs::Get
     return std::make_tuple(x, y, z, w);
 }
 
-bool CLuaPedDefs::SetElementBoneMatrix(lua_State* const luaVM, CClientPed* entity, std::uint32_t boneId, CMatrix boneMatrix)
+bool CLuaPedDefs::SetElementBoneMatrix(lua_State* const luaVM)
```
TracerDS: That wont work for argument parser.

TracerDS (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
W3lac3: In this case it would be through `LUA_DECLARE`, correct? I've made the change, if that's all it is, I'll send.

TracerDS: you would have to remove argument parser from the function list. Try to find a different way instead of not using argparser

W3lac3: I decided to follow the declaration pattern of other functions. I've sent the changes.

TracerDS: Did you check if it works as intended?

W3lac3: Yes, I never send a commit without testing it.

W3lac3: Yes, I never send a commit without testing it.

TracerDS: ~~You didnt check your first commit though 😒~~

W3lac3: Yes, I did. All the parameters were working and the function worked correctly. I only changed it because you told me to, but it was already working.

TracerDS: if you say so, but please try figuring out a way with ArgumentParser and not lua bindings if possible

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -1072,10 +1072,43 @@ std::variant<bool, CLuaMultiReturn<float, float, float, float>> CLuaPedDefs::Get
     return std::make_tuple(x, y, z, w);
 }
 
-bool CLuaPedDefs::SetElementBoneMatrix(lua_State* const luaVM, CClientPed* entity, std::uint32_t boneId, CMatrix boneMatrix)
```
CrosRoad95, line 1075: Just implement that "CMatrix" can be read from 4x4 lua table and don't touch this file at all

Lpsd:
Why do you need to use the old argument parser here? Can you not use `std::variant` with the new arg parser instead of the old `argStream.NextIsTable()` approach?

Anyway, I think @CrosRoad95 suggestion is a more sensible approach as it is wider reaching, for around the same amount of effort.

W3lac3:
> Why do you need to use the old argument parser here? Can you not use `std::variant` with the new arg parser instead of the old `argStream.NextIsTable()` approach?
> 
> Anyway, I think @CrosRoad95 suggestion is a more sensible approach as it is wider reaching, for around the same amount of effort.

I think it's really the best solution to change the `ArgumentParse` to convert the 4x4 lua table into a `CMatrix`. But as it's something more global, not exactly linked to a function, it would be better to open a new PR, where it changes to allow passing the 4x4 lua table in `CMatrix` and close this PR.

I've made this change where it will also allow you to pass a 4x4 lua table, but I haven't sent it yet, because I want to make sure that the best option would be to open another PR, if so, I'll open it by making this change.

PR #3845, opened 2024-11-07
Add new special world property ignorefirestate

FileEX:
This PR adds a new ~~glitch~~ special world property, ``ignorefirestate`` which allows getting into burning vehicles (with fire particle, not the fire indicating low HP and explosion), as well as aiming 'special' weapons like the ``camera``, ``sniper``, ``RPG``, and ``rocket launcher HS`` while the ped is on fire. Thanks to @ds1-e  for the idea.


TheNormalnij:
It is not a sa glitch. This feature has more sense in `setWorldSpecialPropertyEnabled`

Fernando-A-Rocha (changes requested):
Good job!

I agree with TheNormalnij, this feature should be added in a world property as you said it's not a proper glitch, but a different behavior like burn flipped cars.

FileEX:
Moved to special world property

Fernando-A-Rocha (approved):

Client/mods/deathmatch/logic/CClientGame.cpp
bum8hj, line 1: A little nitpick on an unrelated change:
I would rather write "Kickout ..." as 2 words, like "Kick out ..."

PR #3851, opened 2024-11-10
Fix some small problems with Device Selection Dialog

### LINKED ISSUE #3073: Add information about screen in device dialog

you don't know what monitor you choose
image

add information about screen to the dialog "graphics card ( Screen 1)"

/

### Additional context

/

### Security Policy

forkerer:
This aims to improve Device Selection dialog that shows on startup if more than one monitor is present. It's partially based on the changes introduced in SilentPatch, although the approach is a bit different.

Changes:
- The dialog is now visible in the taskbar
- The dialog now has an icon assigned to it
- The device selection drop down correctly displays the monitor names, instead of using the GPU model as all options (The name retrieval logic is copied from SilentPatch)

Here's how the dialog looked like before the changes:
original_naming

And here's the result after the changes:
new_naming
obraz

And here it is in the taskbar:
Taskbar

The main reason for these change is that in most cases the device selection dialog is lost somewhere under all the windows, making it hard to find. I didn't port over all of the changes from SilentPatch, as i wanted to keep it simple for now.

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -94,7 +94,7 @@ struct RwSubSystemInfo
 
 #define FUNC_rwDeviceSystemRequest 0x7F2AB0
 using rwDeviceSystemRequest = RwSubSystemInfo*(__cdecl*)(RwDevice* device, std::int32_t requestId, RwSubSystemInfo* pOut, void* pInOut, std::int32_t numIn);
-static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, int32_t subSystemIndex)
+static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, std::int32_t subSystemIndex)
 {
     auto rwGlobals = *(RwGlobals**)(0xC97B24);
     auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -20,64 +20,69 @@ std::map<std::string, std::string, std::less<>> GetFriendlyMonitorNamesForDevice
     return monitorNames;
 #else
     HMODULE user32Lib = LoadLibrary(TEXT("user32"));
-    if (user32Lib != nullptr)
+    if (!user32Lib)
+        return monitorNames;
+
+    auto getDisplayConfigBufferSizes = (decltype(GetDisplayConfigBufferSizes)*)GetProcAddress(user32Lib, "GetDisplayConfigBufferSizes");
+    auto queryDisplayConfig = (decltype(QueryDisplayConfig)*)GetProcAddress(user32Lib, "QueryDisplayConfig");
+    auto displayConfigGetDeviceInfo = (decltype(DisplayConfigGetDeviceInfo)*)GetProcAddress(user32Lib, "DisplayConfigGetDeviceInfo");
+    if (!getDisplayConfigBufferSizes || !queryDisplayConfig || !displayConfigGetDeviceInfo)
     {
-        auto getDisplayConfigBufferSizes = (decltype(GetDisplayConfigBufferSizes)*)GetProcAddress(user32Lib, "GetDisplayConfigBufferSizes");
-        auto queryDisplayConfig = (decltype(QueryDisplayConfig)*)GetProcAddress(user32Lib, "QueryDisplayConfig");
-        auto displayConfigGetDeviceInfo = (decltype(DisplayConfigGetDeviceInfo)*)GetProcAddress(user32Lib, "DisplayConfigGetDeviceInfo");
-        if (getDisplayConfigBufferSizes != nullptr && queryDisplayConfig != nullptr && displayConfigGetDeviceInfo != nullptr)
-        {
-            UINT32                                     pathCount, modeCount;
-            std::unique_ptr<DISPLAYCONFIG_PATH_INFO[]> paths;
-            std::unique_ptr<DISPLAYCONFIG_MODE_INFO[]> modes;
-
-            LONG result = ERROR_SUCCESS;
-            do
-            {
-                result = getDisplayConfigBufferSizes(QDC_ONLY_ACTIVE_PATHS, &pathCount, &modeCount);
-                if (result != ERROR_SUCCESS)
-                {
-                    break;
-                }
-                paths = std::make_unique<DISPLAYCONFIG_PATH_INFO[]>(pathCount);
-                modes = std::make_unique<DISPLAYCONFIG_MODE_INFO[]>(modeCount);
-                result = queryDisplayConfig(QDC_ONLY_ACTIVE_PATHS, &pathCount, paths.get(), &modeCount, modes.get(), nullptr);
-            } while (result == ERROR_INSUFFICIENT_BUFFER);
-
-            if (result == ERROR_SUCCESS)
-            {
-                for (size_t i = 0; i < pathCount; i++)
-                {
-                    DISPLAYCONFIG_TARGET_DEVICE_NAME targetName = {};
-                    targetName.header.adapterId = paths[i].targetInfo.adapterId;
-                    targetName.header.id = paths[i].targetInfo.id;
-                    targetName.header.type = DISPLAYCONFIG_DEVICE_INFO_GET_TARGET_NAME;
-                    targetName.header.size = sizeof(targetName);
-                    const LONG targetNameResult = DisplayConfigGetDeviceInfo(&targetName.header);
-
-                    DISPLAYCONFIG_SOURCE_DEVICE_NAME sourceName = {};
-                    sourceName.header.adapterId = paths[i].sourceInfo.adapterId;
-                    sourceName.header.id = paths[i].sourceInfo.id;
-                    sourceName.header.type = DISPLAYCONFIG_DEVICE_INFO_GET_SOURCE_NAME;
-                    sourceName.header.size = sizeof(sourceName);
-                    const LONG sourceNameResult = DisplayConfigGetDeviceInfo(&sourceName.header);
-                    if (targetNameResult == ERROR_SUCCESS && sourceNameResult == ERROR_SUCCESS && targetName.monitorFriendlyDeviceName[0] != '\0')
-                    {
-                        char gdiDeviceName[std::size(sourceName.viewGdiDeviceName)];
-                        char monitorFriendlyDeviceName[std::size(targetName.monitorFriendlyDeviceName)];
-                        WideCharToMultiByte(CP_ACP, 0, sourceName.viewGdiDeviceName, -1, gdiDeviceName, static_cast<int>(std::size(gdiDeviceName)), nullptr,
-                                            nullptr);
-                        WideCharToMultiByte(CP_ACP, 0, targetName.monitorFriendlyDeviceName, -1, monitorFriendlyDeviceName,
-                                            static_cast<int>(std::size(monitorFriendlyDeviceName)), nullptr, nullptr);
-
-                        monitorNames.try_emplace(gdiDeviceName, monitorFriendlyDeviceName);
-                    }
-                }
-            }
-        }
         FreeLibrary(user32Lib);
+        return monitorNames;
     }
 
+    UINT32                                     pathCount, modeCount;
+    std::unique_ptr<DISPLAYCONFIG_PATH_INFO[]> paths;
+    std::unique_ptr<DISPLAYCONFIG_MODE_INFO[]> modes;
+
+    LONG result = ERROR_SUCCESS;
+    do
+    {
+        result = getDisplayConfigBufferSizes(QDC_ONLY_ACTIVE_PATHS, &pathCount, &modeCount);
+        if (result != ERROR_SUCCESS)
+        {
+            break;
+        }
+        paths = std::make_unique<DISPLAYCONFIG_PATH_INFO[]>(pathCount);
+        modes = std::make_unique<DISPLAYCONFIG_MODE_INFO[]>(modeCount);
+        result = queryDisplayConfig(QDC_ONLY_ACTIVE_PATHS, &pathCount, paths.get(), &modeCount, modes.get(), nullptr);
+    } while (result == ERROR_INSUFFICIENT_BUFFER);
+
+    if (result != ERROR_SUCCESS)
+    {
+        FreeLibrary(user32Lib);
+        return monitorNames;
+    }
+
+    for (size_t i = 0; i < pathCount; i++)
+    {
+        DISPLAYCONFIG_TARGET_DEVICE_NAME targetName = {};
+        targetName.header.adapterId = paths[i].targetInfo.adapterId;
+        targetName.header.id = paths[i].targetInfo.id;
+        targetName.header.type = DISPLAYCONFIG_DEVICE_INFO_GET_TARGET_NAME;
+        targetName.header.size = sizeof(targetName);
+        const LONG targetNameResult = DisplayConfigGetDeviceInfo(&targetName.header);
+
+        DISPLAYCONFIG_SOURCE_DEVICE_NAME sourceName = {};
+        sourceName.header.adapterId = paths[i].sourceInfo.adapterId;
+        sourceName.header.id = paths[i].sourceInfo.id;
+        sourceName.header.type = DISPLAYCONFIG_DEVICE_INFO_GET_SOURCE_NAME;
+        sourceName.header.size = sizeof(sourceName);
+        const LONG sourceNameResult = DisplayConfigGetDeviceInfo(&sourceName.header);
+        if (targetNameResult == ERROR_SUCCESS && sourceNameResult == ERROR_SUCCESS && targetName.monitorFriendlyDeviceName[0] != '\0')
+        {
+            char gdiDeviceName[std::size(sourceName.viewGdiDeviceName)];
+            char monitorFriendlyDeviceName[std::size(targetName.monitorFriendlyDeviceName)];
+            WideCharToMultiByte(CP_ACP, 0, sourceName.viewGdiDeviceName, -1, gdiDeviceName, static_cast<int>(std::size(gdiDeviceName)), nullptr, nullptr);
+            WideCharToMultiByte(CP_ACP, 0, targetName.monitorFriendlyDeviceName, -1, monitorFriendlyDeviceName,
+                                static_cast<int>(std::size(monitorFriendlyDeviceName)), nullptr, nullptr);
+
+            monitorNames.try_emplace(gdiDeviceName, monitorFriendlyDeviceName);
+        }
+    }
+
+    FreeLibrary(user32Lib);
     return monitorNames;
 #endif
 }
@@ -125,23 +130,21 @@ static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSyst
 static INT_PTR CALLBACK CustomDlgProc(HWND window, UINT msg, WPARAM wParam, LPARAM lParam)
 {
     auto orgDialogFunc = (DLGPROC)FUNC_DialogFunc;
-    if (msg == WM_INITDIALOG)
-    {
-        orgDialogFunc(window, msg, wParam, lParam);
+    if (msg != WM_INITDIALOG)
+        return orgDialogFunc(window, msg, wParam, lParam);
 
-        // Set Icon
-        HMODULE hGameModule = GetModuleHandle(nullptr);
-        SendMessage(window, WM_SETICON, ICON_SMALL, reinterpret_cast<LPARAM>(LoadIcon(hGameModule, MAKEINTRESOURCE(100))));
+    orgDialogFunc(window, msg, wParam, lParam);
 
-        // Make the dialog visible in the task bar
-        // https://stackoverflow.com/a/1462811
-        SetWindowLongPtr(window, GWL_EXSTYLE, WS_EX_APPWINDOW);
-        ShowWindow(window, SW_HIDE);
-        ShowWindow(window, SW_SHOW);
-        return FALSE;
-    }
+    // Set Icon
+    HMODULE hGameModule = GetModuleHandle(nullptr);
+    SendMessage(window, WM_SETICON, ICON_SMALL, reinterpret_cast<LPARAM>(LoadIcon(hGameModule, MAKEINTRESOURCE(100))));
 
-    return orgDialogFunc(window, msg, wParam, lParam);
+    // Make the dialog visible in the task bar
+    // https://stackoverflow.com/a/1462811
+    SetWindowLongPtr(window, GWL_EXSTYLE, WS_EX_APPWINDOW);
+    ShowWindow(window, SW_HIDE);
+    ShowWindow(window, SW_SHOW);
+    return FALSE;
 }
 
 void CMultiplayerSA::InitHooks_DeviceSelection()
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -18,9 +18,9 @@
 
 // This is copied from SilentPatch:
 // https://github.com/CookiePLMonster/SilentPatch/blob/dev/SilentPatch/FriendlyMonitorNames.cpp
-std::map<std::string, std::string, std::less<>> GetFriendlyMonitorNamesForDevicePaths()
+std::unordered_map<std::string, std::string> GetFriendlyMonitorNamesForDevicePaths()
 {
-    std::map<std::string, std::string, std::less<>> monitorNames;
+    std::unordered_map<std::string, std::string> monitorNames;
 
     HMODULE user32Lib = LoadLibrary(TEXT("user32"));
     if (!user32Lib)
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -2,7 +2,7 @@
  *
  *  PROJECT:     Multi Theft Auto v1.0
  *  LICENSE:     See LICENSE in the top level directory
- *  FILE:        multiplayer_sa/CMultiplayerSA_Weapons.cpp
+ *  FILE:        multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
  *
  *  Multi Theft Auto is available from http://www.multitheftauto.com/
  *
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -94,8 +94,7 @@ static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSyst
 {
     auto rwGlobals = *(RwGlobals**)(0xC97B24);
     auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
-    auto result = rwDeviceSystemRequestFunc(&rwGlobals->dOpenDevice, 14, subSystemInfo, nullptr, subSystemIndex);
-    if (!result)
+    if (!rwDeviceSystemRequestFunc(&rwGlobals->dOpenDevice, 14, subSystemInfo, nullptr, subSystemIndex))
         return nullptr;
 
     auto pDxDevice = *(IDirect3D9**)0xC97C20;
--- a/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
@@ -1,0 +1,157 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.0
+ *  LICENSE:     See LICENSE in the top level directory
+ *  FILE:        multiplayer_sa/CMultiplayerSA_Weapons.cpp
+ *
+ *  Multi Theft Auto is available from http://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+#ifdef _WIN32
+    #ifndef NOMINMAX
+        #define NOMINMAX
+    #endif
+#endif
+
+// This is copied from SilentPatch:
+// https://github.com/CookiePLMonster/SilentPatch/blob/dev/SilentPatch/FriendlyMonitorNames.cpp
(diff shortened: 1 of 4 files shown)


Fernando-A-Rocha (approved):
This seems like an improvement. Well done!

Fernando-A-Rocha:
Is it possible to use translation for the strings in the dialog box so they can vary depending on MTA client language chosen?

forkerer:
> Is it possible to use translation for the strings in the dialog box so they can vary depending on MTA client language chosen?

Not with the current approach, but should be possible to do in the future. At this moment I don't touch the dialog creation at all, so i have no control over the UI elements, only thing that's hooked is the message handler for it

forkerer:
(Partially) fixes #3073

nekjutsu:
Why is this dialogue needed at all? 
I haven’t seen this in any game (except some Unreal Tournament or Quake from 1990s)

I suggest turning it off

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+
+#ifdef _WIN32
+    #ifndef NOMINMAX
+        #define NOMINMAX
+    #endif
+#endif
```
TracerDS: Pointless

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+// This is copied from SilentPatch:
+// https://github.com/CookiePLMonster/SilentPatch/blob/dev/SilentPatch/FriendlyMonitorNames.cpp
+std::map<std::string, std::string, std::less<>> GetFriendlyMonitorNamesForDevicePaths()
+{
+    std::map<std::string, std::string, std::less<>> monitorNames;
```
TracerDS: unordered_map instead of map

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#if WINVER < _WIN32_WINNT_WIN7
+    return monitorNames;
+#else
+    HMODULE user32Lib = LoadLibrary(TEXT("user32"));
+    if (user32Lib != nullptr)
```
TracerDS: `if (user32Lib)` ?

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    {
+        auto getDisplayConfigBufferSizes = (decltype(GetDisplayConfigBufferSizes)*)GetProcAddress(user32Lib, "GetDisplayConfigBufferSizes");
+        auto queryDisplayConfig = (decltype(QueryDisplayConfig)*)GetProcAddress(user32Lib, "QueryDisplayConfig");
+        auto displayConfigGetDeviceInfo = (decltype(DisplayConfigGetDeviceInfo)*)GetProcAddress(user32Lib, "DisplayConfigGetDeviceInfo");
+        if (getDisplayConfigBufferSizes != nullptr && queryDisplayConfig != nullptr && displayConfigGetDeviceInfo != nullptr)
```
TracerDS: same as above

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+                        monitorNames.try_emplace(gdiDeviceName, monitorFriendlyDeviceName);
+                    }
+                }
+            }
+        }
+        FreeLibrary(user32Lib);
+    }
+
+    return monitorNames;
```
TracerDS: early return

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+#endif
+}
+
+struct RwSubSystemInfo
+{
+    char name[80];
+};
+
+#define FUNC_rwDeviceSystemRequest 0x7F2AB0
+using rwDeviceSystemRequest = RwSubSystemInfo*(__cdecl*)(void* device, uint32_t requestId, RwSubSystemInfo* pOut, void* pInOut, uint32_t numIn);
```
TracerDS: `std::uint32_t`

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, int32_t subSystemIndex)
```
TracerDS: above
static is not needed here

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+    static auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
+    auto        devicePointer = *(uint32_t*)(0xC97B24);
```
TracerDS: whats that mem address? + above

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    auto        result = rwDeviceSystemRequestFunc((void*)(devicePointer + 0x10), 14, subSystemInfo, nullptr, subSystemIndex);
+    if (result == nullptr)
```
TracerDS: `!result`

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+        return nullptr;
+
+    auto pDxDevice = *(IDirect3D9**)0xC97C20;
+    if (pDxDevice == nullptr)
```
TracerDS: above

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    char name[80];
+};
+
+#define FUNC_rwDeviceSystemRequest 0x7F2AB0
+using rwDeviceSystemRequest = RwSubSystemInfo*(__cdecl*)(void* device, uint32_t requestId, RwSubSystemInfo* pOut, void* pInOut, uint32_t numIn);
+static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, int32_t subSystemIndex)
+{
+    static auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
+    auto        devicePointer = *(uint32_t*)(0xC97B24);
+    auto        result = rwDeviceSystemRequestFunc((void*)(devicePointer + 0x10), 14, subSystemInfo, nullptr, subSystemIndex);
```
TracerDS: whats that mem address (`devicePointer + 0x10`)?

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+        HMODULE hGameModule = GetModuleHandle(nullptr);
+        SendMessage(window, WM_SETICON, ICON_SMALL, reinterpret_cast<LPARAM>(LoadIcon(hGameModule, MAKEINTRESOURCE(100))));
+
+        // Make the dialog visible in the task bar
+        // https://stackoverflow.com/a/1462811
+        SetWindowLongPtr(window, GWL_EXSTYLE, WS_EX_APPWINDOW);
+        ShowWindow(window, SW_HIDE);
+        ShowWindow(window, SW_SHOW);
+        return FALSE;
+    }
```
TracerDS: early return

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    else
+    {
+        strncpy_s(subSystemInfo->name, identifier.Description, _TRUNCATE);
+    }
+
+    return subSystemInfo;
+}
+
+#define FUNC_DialogFunc 0x745E50
+static INT_PTR CALLBACK CustomDlgProc(HWND window, UINT msg, WPARAM wParam, LPARAM lParam)
```
TracerDS: static is pointless here

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+        return FALSE;
+    }
+
+    return orgDialogFunc(window, msg, wParam, lParam);
+}
+
+void CMultiplayerSA::InitHooks_DeviceSelection()
+{
+    MemPut<DLGPROC>(0x746239, (DLGPROC)&CustomDlgProc);
+    HookInstall(0x7F2C30, (DWORD)RwEngineGetSubSystemInfo_Hooked, 6);
```
TracerDS: what are those mem addresses?

TracerDS (changes requested):

AlexTMjugador:
> Why is this dialogue needed at all?

A single computer may have several GPUs from different manufacturers in use at the same time, and each GPU be connected to different monitors. (For example, this is commonplace in high-end laptops with Intel CPUs, where both a integrated GPU and discrete GPU are present and used at the same time to balance performance and power usage.)

This selection dialogue allows players with such computer configurations to precisely select in which GPU and screen the game should be rendered, which is necessary and useful in those environments. Other games implement such selection within the in-game graphics options (which is a less straightforward thing to do in MTA), have separate, more polished launchers for graphic options (this is unfeasible for MTA), or use a default GPU and screen provided by the operating system graphics APIs (which can be a bad default and inconvenient for the user to change, so it isn't a proper solution for MTA standards either).

Fernando-A-Rocha:
> Why is this dialogue needed at all? I haven’t seen this in any game (except some Unreal Tournament or Quake from 1990s)
> 
> I suggest turning it off

GTA:SA has that dialog before the game boots on a given screen device.

If we want to remove it, we can do like most games: if no previously selected screen or previously selected screen no longer connected, auto select the system's default screen; add a setting for changing screen for next launch? (This improvement could be made in a new PR, this PR is already a good bugfix)

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: obraz
Neither of those is declared in MTA, and only other places that do use it are just plain ASM.
Added the incomplete stubs for them in Renderware.h now, so it's a bit clearer

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Removed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Pointless in this case, under normal conditions there will be 2-3 entries in the map. I don't think this will be higher than 6 ever. The overhead of unordered_map makes it worse choice in this situation. 
As soon as someone wants to run MTA on the Las Vegas Sphere, I will gladly change it :D
462555296_9454252167923288_2572623932954594248_n

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Clarification in the comment below about the offset, replaced it with a stub of the original Renderware class now

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Added clarification comment to the first one, added a define to the second one to match the current approach over the codebase

Client/sdk/game/RenderWare.h
```diff
@@ -526,3 +526,49 @@ struct RwError
...
+    // RwIm2DRenderPrimitiveFunction        fpIm2DRenderPrimitive;
+    // RwIm2DRenderIndexedPrimitiveFunction fpIm2DRenderIndexedPrimitive;
+    // RwIm3DRenderLineFunction             fpIm3DRenderLine;
+    // RwIm3DRenderTriangleFunction         fpIm3DRenderTriangle;
+    // RwIm3DRenderPrimitiveFunction        fpIm3DRenderPrimitive;
+    // RwIm3DRenderIndexedPrimitiveFunction fpIm3DRenderIndexedPrimitive;
+};
+// static_assert(sizeof(RwDevice) == 0x38, "Incorrect class size: RwDevice");
+
+typedef bool (*RwStandardFunc)(void*, void*, std::int32_t);
```
TracerDS, line 554:
```cpp
using RwStandardFunc = bool (*)(void*, void*, std::int32_t);
```

Client/sdk/game/RenderWare.h
```diff
@@ -526,3 +526,49 @@ struct RwError
 {
     int err1, err2;
 };
+
+/*****************************************************************************/
+/** RenderWare Globals                                                      **/
+/*****************************************************************************/
+
+typedef bool (*RwSystemFunc)(std::int32_t, void*, void*, std::int32_t);
```
TracerDS, line 534:
```cpp
using RwSystemFunc = bool (*)(std::int32_t, void*, void*, std::int32_t);
```

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    ShowWindow(window, SW_SHOW);
+    return FALSE;
+}
+
+#define FUNC_RwEngineGetSubSystemInfo 0x7F2C30
+void CMultiplayerSA::InitHooks_DeviceSelection()
+{
+    // 0x746239 -> Exact address where the original DialogFunc address is being pushed as an argument to DialogBoxParamA(),
+    // we're replacing it with out own proxy function
+    MemPut<DLGPROC>(0x746239, (DLGPROC)&CustomDlgProc);
```
TracerDS: if `0x746239` is the original address then what does `FUNC_DialogFunc` point to?

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+struct RwSubSystemInfo
+{
+    char name[80];
+};
+
+#define FUNC_rwDeviceSystemRequest 0x7F2AB0
+using rwDeviceSystemRequest = RwSubSystemInfo*(__cdecl*)(RwDevice* device, std::int32_t requestId, RwSubSystemInfo* pOut, void* pInOut, std::int32_t numIn);
+static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, std::int32_t subSystemIndex)
+{
+    auto rwGlobals = *(RwGlobals**)(0xC97B24);
```
TracerDS: whats that mem address?
also `auto*`

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
+    auto result = rwDeviceSystemRequestFunc(&rwGlobals->dOpenDevice, 14, subSystemInfo, nullptr, subSystemIndex);
+    if (!result)
+        return nullptr;
+
+    auto pDxDevice = *(IDirect3D9**)0xC97C20;
```
TracerDS: above

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+#define FUNC_rwDeviceSystemRequest 0x7F2AB0
+using rwDeviceSystemRequest = RwSubSystemInfo*(__cdecl*)(RwDevice* device, std::int32_t requestId, RwSubSystemInfo* pOut, void* pInOut, std::int32_t numIn);
+static RwSubSystemInfo* RwEngineGetSubSystemInfo_Hooked(RwSubSystemInfo* subSystemInfo, std::int32_t subSystemIndex)
+{
+    auto rwGlobals = *(RwGlobals**)(0xC97B24);
+    auto rwDeviceSystemRequestFunc = (rwDeviceSystemRequest)(FUNC_rwDeviceSystemRequest);
+    auto result = rwDeviceSystemRequestFunc(&rwGlobals->dOpenDevice, 14, subSystemInfo, nullptr, subSystemIndex);
+    if (!result)
+        return nullptr;
```
TracerDS:
```cpp
if (!rwDeviceSystemRequestFunc(&rwGlobals->dOpenDevice, 14, subSystemInfo, nullptr, subSystemIndex))
    return nullptr;
```

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+// This is copied from SilentPatch:
+// https://github.com/CookiePLMonster/SilentPatch/blob/dev/SilentPatch/FriendlyMonitorNames.cpp
+std::map<std::string, std::string, std::less<>> GetFriendlyMonitorNamesForDevicePaths()
+{
+    std::map<std::string, std::string, std::less<>> monitorNames;
+
+#if WINVER < _WIN32_WINNT_WIN7
+    return monitorNames;
+#else
```
TracerDS: Whats happening here?

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
TracerDS: > Pointless in this case, under normal conditions there will be 2-3 entries in the map. I don't think this will be higher than 6 ever. The overhead of unordered_map makes it worse choice in this situation. As soon as someone wants to run MTA on the Las Vegas Sphere, I will gladly change it :D 462555296_9454252167923288_2572623932954594248_n

There is no overhead. In fact `map` is slower than `unordered_map` even when the amount of entries is small.
image
image

TracerDS (changes requested):

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
```diff
@@ -0,0 +1,157 @@
...
+    // If we can't find the friendly name, either because it doesn't exist or we're on an ancient Windows, fall back to the device name
+    auto it = friendlyNames.find(identifier.DeviceName);
+    if (it != friendlyNames.end())
+    {
+        strncpy_s(subSystemInfo->name, it->second.c_str(), _TRUNCATE);
+    }
+    else
+    {
+        strncpy_s(subSystemInfo->name, identifier.Description, _TRUNCATE);
+    }
```
AlexTMjugador, line 124: Wouldn't it be better to show the monitor name and device name, as mentioned in issue #3073, if possible? I can see the dialog getting confusing for multihead setups with two or more identical monitors otherwise. (In fact, if possible, a screen number could also be a good idea, as a single multihead GPU may have several video outputs connected to monitors of an identical model.)

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: 1. This is an absolute nitpick, doesn't affect performance/stability of the game at all, and is purely due to your personal style preferences.
2. Code guidelines (https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md) do not specify anything about using unordered_map over map
3. Minor stuff like this will only discourage people from contributing to the project, this is a non-topic in any normal scenario, and the code runs one time during startup for few items.
4. Your benchmark is also invalid, most of the benchmarks like these can be made to fit any thesis, it tests functions that are not even used in my code (Not really mine, it's Silents), when we replicate the actual logic that's happening suddenly map is very slightly ahead:
obraz

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Read the comment again, 0x746239 is where the pointer to the original function is being pushed to the stack to be used as an argument to `DialogBoxParamA`, I replace it at this place with the pointer to the new function.

FUNC_DialogFunc point to the address where the original function is placed, as we need to call it later on in our proxy function

Client/sdk/game/RenderWare.h
forkerer: This entire file uses typedef for that, i don't want to make it inconsistent, and i don't feel like it's in the scope of this PR to change all other occurences of that

Client/sdk/game/RenderWare.h
forkerer: Same as above

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: It's where the static instance of `RwGlobals` pointer is stored in GTA.
Changed all auto to auto* in case of pointers.

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: It's where the static instance of `IDirect3D9*` pointer is stored in GTA.

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Changed

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: The APIs used to retrieve monitor names are not supported on systems olders than windows 7. I see that MTA dropped the vista support already, so i've removed the safety check now

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: It would be better, but i personally don't know how to easily achieve that. This change improves upon the most often encountered scenario - 1 GPU with multiple screens :)
I do not have the possibility of reproducing the multihead setup scenario, so can't really fix that at this point

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Replaced every single magic number with one of the defines that's are now at the top of the file

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Replaced every single magic number with one of the defines that's are now at the top of the file

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: Replaced every single magic number with one of the defines that's are now at the top of the file

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
TracerDS: > 1. This is an absolute nitpick, doesn't affect performance/stability of the game at all, and is purely due to your personal style preferences.
> 2. Code guidelines (https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md) do not specify anything about using unordered_map over map
> 3. Minor stuff like this will only discourage people from contributing to the project, this is a non-topic in any normal scenario, and the code runs one time during startup for few items.
> 4. Your benchmark is also invalid, most of the benchmarks like these can be made to fit any thesis, it tests functions that are not even used in my code (Not really mine, it's Silents), when we replicate the actual logic that's happening suddenly map is very slightly ahead:
>    obraz

1. This is not nitpicking. This is not a matter of personal style preference. This is like comparing `std::string` to `std::vector<char>`.
2. Guidelines dont dictate which code should be used when performance is at stake.
`map` **is** slower than `unordered_map` due to `map` sorting elements. `unordered_map` therefore in theory takes more memory than normal map but is faster in the end.
3. Reviewing code does not discourage people from contributing. Its weird you feel that way.
4. Your benchmark differs. One isnt equivalent to another. Why do you try to manipulate the results this way?
In fact, just constructing one class is slower than the other. Doesnt mean one is generally slower than the other
image

I made 2 functions equivalent and would you look at that: `map` is still slower than `unordered_map`
image

Client/sdk/game/RenderWare.h
TracerDS: Well, you do add new functionality and we are in C++ after all, but you can keep the consistency.

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
forkerer: 1. In my opinion it is nitpicking.
2. There's no performance at stake, pretty much every single other function called in that scope (GetFriendlyMonitorNamesForDevicePaths) will take orders of magnitude longer than all the map/unordered_map operations there, it's literally few items added to map, and a single check to find matching item afterwards. This is purely style issue.
3. Reviewing code is completely okay, every single other comment in this PR has been fixed without any complaints, it's this one that's a problem to me.
4. It differs because that's the actual usage in the code that's part of this PR, if you want to fight about this then let's consider exactly the same handling code (it's true that i missed the fact that one uses [key] = value and second one used try_emplace):
obraz
obraz

As i said, those micro-benchmarks are completely worthless, and you can adjust the code both ways to always prove your thesis.

Fernando-A-Rocha (approved):

Client/multiplayer_sa/CMultiplayerSA_DeviceSelection.cpp
TracerDS: 1. Nitpicking is only applicable for insignificant changes. If changing functionality (chanigng how X works -> changing functionality) and debating about performance is insignificant then every code review is nitpicking.
2. You dont seem to understand what `style issue` is. This is **not** a style manner. Even if one function takes longer than another there is no reason to bottleneck it and make it slower.
4. I did some more digging and it turns out only the initialization is slower for the unordered_map. Providing `std::less` to map has no effect because `map` already takes in `std::less` as its `Compare` template argument. Check the wiki
Benchmarks:
- [`unordered_map` vs `map<less<>>` with initialization](<https://quick-bench.com/q/d0-EC0XW0um4LwIvulwd22ly5H8>)
- [`unordered_map` vs `map` with initialization](<https://quick-bench.com/q/G8IVphVpAOlc6Tnd5zMdLeYE-fU>)
- [`unordered_map` vs `map<less<>>` without initialization](<https://quick-bench.com/q/7UVnBlPEcwoPBo8PKEpKqKsbkNY>)
- [`unordered_map` vs `map` without initialization](<https://quick-bench.com/q/9Ua2Y50K5898djUY1RkCPo4rb7U>)

As you can see only initialization takes more time, but is more performant in the end.
TLDR; `std::map<std::string, std::string, std::less<>>` resolves to `std::map<std::string, std::string>` due to the nature of `std::less`. So basically we are still comparing `std::map` vs `std::unordered_map`.

forkerer: Congratulations, You've spent more on this analysis than the code you're arguing about will spend searching for an entry in this 2-4 elements big map in the coming years. There's only 486 other usages of std::map that can be fixed in the codebase - hopefully one day soon.
Changed to unordered_map now - i really hope to see someone running MTA on a setup that uses hundreds of displays so they can take advantage of those few nanoseconds gained on startup :)

TracerDS: If we stop caring for performance, we should stop caring about the crashes. And if we stop caring about the crashes we should stop caring at all. You can see where this is going.

I dont know why are you being so defensive here. Just pointed out a performance benefit. No hard feelings.

Fernando-A-Rocha: > Congratulations, You've spent more on this analysis than the code you're arguing about will spend searching for an entry in this 2-4 elements big map in the coming years. There's only 486 other usages of std::map that can be fixed in the codebase - hopefully one day soon. Changed to unordered_map now - i really hope to see someone running MTA on a setup that uses hundreds of displays so they can take advantage of those few nanoseconds gained on startup :)

XD

Tracer's mindset is If we can change it to be more performant, why wouldn't we? I gotta agree if it takes little effort to make a refactor from std map to std unordered map, for example.

PR #3853, opened 2024-11-12
Fix min_mta_version 'server' not working as expected

### LINKED ISSUE #3852: `min_mta_version` 'server' is not working as expected
### Describe the bug

You can start a resource that has a minimal server version that is higher than your server's. This should not be possible. The server should not be able to start a resource that clearly states that it was designed for a more recent server version.

### Steps to reproduce

1. Launch MTA server version 1.6.0-9.22746.0 for example
2. Create a resource and set its minimum server version to a higher build
```xml
<meta>
    <min_mta_version server="1.6.0-9.22789.0"/>
</meta>
```
3. Start it: notice that resource starts normally; it shouldnt!

### Version

Tested on 1.6.0-9.22746.0. Likely has been like this for years.

### Additional context

Developers create resources that are meant to be used in servers that are a certain version or higher. This feature is what they use, it seems broken.

On the other hand, min_mta_version "client" seems to be working correctly. It updates the server's minclientversion to the highest min_mta_version "client" provided by resources. It only lets players join with that minclientversion version or higher.
Image

Fernando-A-Rocha:
Fixes #3852
Caused by: https://github.com/multitheftauto/mtasa-blue/pull/3853#issuecomment-2472903077

I found that 2 missing includes were preventing the script from checking if a server's version is too low to start a certain resource that requires a higher version.

image

image

### DIFF
--- a/Server/mods/deathmatch/logic/CResource.cpp
+++ b/Server/mods/deathmatch/logic/CResource.cpp
@@ -35,6 +35,8 @@
 #include <net/SimHeaders.h>
 #include <zip.h>
 #include <glob/glob.h>
+#include "version.h"
+#include "CStaticFunctionDefinitions.h"
 
 #ifdef WIN32
     #include <zip/iowin32.h>


TracerDS (approved):

PlatinMTA:
Can we just send a warning instead of not letting the resource start since its serversided?

(edit: nvm its checked in the same if statement)

Fernando-A-Rocha:
If a dev defines a certain min server version in the meta.xml, their intention is to prevent a server with an inferior server from starting the resource.

If the server owner wants to go crazy and execute the resource anyway, then they can easily edit meta.xml to get rid of the min_mta_version setting.

TracerDS:
How does including 2 headers fix this issue?

Fernando-A-Rocha:
> How does including 2 headers fix this issue?

- MTASA_VERSION_BUILD was not defined without version.h (see the code on the screenshot in the main post)
- CStaticFunctionDefinitions was also not imported

Fernando-A-Rocha:
After some digging, I found that this commit actually broke it
https://github.com/multitheftauto/mtasa-blue/commit/693976b7131a87df71d81256a80cc48b22ab7bcf ("Clean up precompiled headers for Server" with 435 files changed) 
@botder removed CStaticFunctionDefinitions.h and version.h includes from Server/mods/deathmatch/StdInc.h

ghost (approved):
Good job

TheNormalnij (approved):

TheNormalnij:
thx

PR #3854, opened 2024-11-13
Check if another player has the same serial

ghost:
This pull request adds a check when connecting to ensure that players can't use the same serial, thus preventing abuses related to serial manipulation, such as for example, when they try to spoof a server's moderator's serial as their own, while they're connected to the server.

The change involved the addition of a new parameter called `check_duplicate_serials` which allows turning off the feature for development and testing servers to connect with virtual machines.

The new code is: `CD50`
The new description is: Serial already in use

### DIFF
--- a/Client/mods/deathmatch/logic/CPacketHandler.h
+++ b/Client/mods/deathmatch/logic/CPacketHandler.h
@@ -38,12 +38,12 @@ class CPacketHandler
         ELEMENT_FAILURE,
         GENERAL_REFUSED,
         SERIAL_VERIFICATION,
-        SERIAL_DUPLICATE,
         CONNECTION_DESYNC,
         BAN,
         KICK,
         CUSTOM,
-        SHUTDOWN
+        SHUTDOWN,
+        SERIAL_DUPLICATE
     };
 
     struct SEntityDependantStuff
--- a/Client/mods/deathmatch/logic/CPacketHandler.h
+++ b/Client/mods/deathmatch/logic/CPacketHandler.h
@@ -38,6 +38,7 @@ class CPacketHandler
         ELEMENT_FAILURE,
         GENERAL_REFUSED,
         SERIAL_VERIFICATION,
+        SERIAL_DUPLICATE,
         CONNECTION_DESYNC,
         BAN,
         KICK,
--- a/Server/mods/deathmatch/local.conf
+++ b/Server/mods/deathmatch/local.conf
@@ -274,6 +274,12 @@
          Values: 0 - Off, 1 - Enabled.  Default - 0 -->
     <elementdata_whitelisted>0</elementdata_whitelisted>
 
+    <!-- This parameter determines that the server checks when the players connect that there are no other players with the same serial number. 
+         Note that this may break compatibility with virtual machines.
+         Not guarantees that the player cannot manipulate their serial.
+         Values: 0 - Off, 1 - Enabled.  Default - 1 -->
+    <check_duplicate_serials>1</check_duplicate_serials>
+
     <!-- Specifies the module(s) which are loaded with the server. To load several modules, add more <module>
          parameter(s). Optional parameter. -->
     <!-- <module src="sample_win32.dll"/> -->
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1798,6 +1798,19 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
             return;
         }
 
+        // Check if another player is using the same serial
+        CPlayer* playerExisting = m_pPlayerManager->GetBySerial(strSerial);
+
+        if (playerExisting)
+        {
+            // Tell the console
+            CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
+
+            // Tell the player the problem
+            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
+            return;
+        }
+
         // Check the nick is valid
         if (!CheckNickProvided(szNick))
         {
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1799,9 +1799,7 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
         }
 
         // Check if another player is using the same serial
-        CPlayer* playerExisting = m_pPlayerManager->GetBySerial(strSerial);
-
-        if (playerExisting)
+        if (m_pPlayerManager->GetBySerial(strSerial))
         {
             // Tell the console
             CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1805,7 +1805,11 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
             CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
 
             // Tell the player the problem
-            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
+            if (pPlayer->CanBitStream(eBitStreamVersion::CheckDuplicateSerials))
+                DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
+            else
+                DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::KICK);
+
             return;
         }
 
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1799,7 +1799,7 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
         }
 
         // Check if another player is using the same serial
-        if (m_pPlayerManager->GetBySerial(strSerial))
+        if (m_pMainConfig->IsCheckDuplicateSerialsEnabled() && m_pPlayerManager->GetBySerial(strSerial))
         {
             // Tell the console
             CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
--- a/Server/mods/deathmatch/logic/CPlayerManager.cpp
+++ b/Server/mods/deathmatch/logic/CPlayerManager.cpp
@@ -138,6 +138,19 @@ CPlayer* CPlayerManager::Get(const char* szNick, bool bCaseSensitive)
     return NULL;
 }
 
+CPlayer* CPlayerManager::GetBySerial(const SString& serial) noexcept
+{
+    for (auto& player : m_Players)
+    {
+        if (player->GetSerial() == serial)
+        {
+            return player;
+        }
+    }
+
+    return nullptr;
+}
+
 void CPlayerManager::DeleteAll()
 {
     // Delete all the items in the list
--- a/Server/mods/deathmatch/logic/CPlayerManager.cpp
+++ b/Server/mods/deathmatch/logic/CPlayerManager.cpp
@@ -138,9 +138,9 @@ CPlayer* CPlayerManager::Get(const char* szNick, bool bCaseSensitive)
     return NULL;
 }
 
-CPlayer* CPlayerManager::GetBySerial(const SString& serial) const noexcept
+const CPlayer* CPlayerManager::GetBySerial(const std::string serial) const noexcept
 {
-    for (auto& player : m_Players)
+    for (const auto& player : m_Players)
     {
         if (player->GetSerial() == serial)
         {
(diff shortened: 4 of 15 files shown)


Fernando-A-Rocha (approved):
Good idea.

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -138,6 +138,19 @@ CPlayer* CPlayerManager::Get(const char* szNick, bool bCaseSensitive)
     return NULL;
 }
 
+CPlayer* CPlayerManager::GetBySerial(const SString& serial) noexcept
+{
+    for (auto& player : m_Players)
```
TracerDS: const?

TracerDS (changes requested):

Server/mods/deathmatch/logic/CPlayerManager.cpp
ghost: Done

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -138,6 +138,19 @@ CPlayer* CPlayerManager::Get(const char* szNick, bool bCaseSensitive)
     return NULL;
 }
 
+CPlayer* CPlayerManager::GetBySerial(const SString& serial) const noexcept
+{
+    for (auto& player : m_Players)
```
TracerDS: const?

TracerDS (changes requested):

Server/mods/deathmatch/logic/CPlayerManager.cpp
TracerDS: i.e. const auto

also std::string if possible instead of SString

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1798,6 +1798,19 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
...
+
+        if (playerExisting)
+        {
+            // Tell the console
+            CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
+
+            // Tell the player the problem
+            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
+            return;
+        }
```
TracerDS: it would be better to have a const CPlayer* in an if statement:
```cpp
if (m_pPlayerManager->GetBySerial(strSerial))
{
    // Tell the console
    CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());

    // Tell the player the problem
    DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
    return;
}

// ...
const CPlayer* CPlayerManager::GetBySerial(const SString& serial) const noexcept
```

TracerDS (changes requested):

Server/mods/deathmatch/logic/CGame.cpp
ghost: I don't really understand this one, could you please explain?

Server/mods/deathmatch/logic/CPlayerManager.cpp
ghost: Thanks

Xenius97:
This solution is not perfect because if you run multiple MTA clients on one machine (VM), you get the same serial, making localhost testing impossible.

Xenius97:
Maybe you could add something like `check_double_serials` to server settings enabled by default.

ghost:
> This solution is not perfect because if you run multiple MTA clients on one machine (VM), you get the same serial, making localhost testing impossible.

I forgot about that, thanks for reminding.
If anyone has any ideas, feel free to share them.

ghost:
> Maybe you could add something like `check_double_serials` to server settings enabled by default.

Good idea, I look forward to other people's opinions as well

Fernando-A-Rocha:
> This solution is not perfect because if you run multiple MTA clients on one machine (VM), you get the same serial, making localhost testing impossible.

I forgot you can use VMs to run multiple MTA clients. You need 1 VM (os) per client right?

Server/mods/deathmatch/logic/CGame.cpp
TracerDS: you dont need additional variable in CGame.cpp.
Just check it in the if statement

As for the GetBySerial function, you can return a const pointer to CPlayer. I mean, both would be fine

ghost: I understand what you mean, it was just strange that you put the suggestion at the bottom of the code

ghost:
> > This solution is not perfect because if you run multiple MTA clients on one machine (VM), you get the same serial, making localhost testing impossible.
> 
> I forgot you can use VMs to run multiple MTA clients. You need 1 VM (os) per client right?

That's right

Fernando-A-Rocha:
> Maybe you could add something like `check_double_serials` to server settings enabled by default.

Maybe we can have a setting like `allow_unverified_clients` that is a more generic name

G-Moris:
Would this be optimized if 'std::unordered_map' and 'find' were used instead of a loop?

Server/mods/deathmatch/local.conf
```diff
@@ -274,6 +274,12 @@
          Values: 0 - Off, 1 - Enabled.  Default - 0 -->
     <elementdata_whitelisted>0</elementdata_whitelisted>
 
+    <!-- This parameter determines that the server checks when the players connect that there are no other players with the same serial number.
```
Fernando-A-Rocha, line 277: I think you also need to add this to editor.conf

It's really annoying how everything is repeated in mtaserver.conf, mtaserver.conf.template, local.conf and editor.conf. We should refactor this in a future PR.

Fernando-A-Rocha (changes requested):

Server/mods/deathmatch/local.conf
ghost: Thx

ghost:
> Would this be optimized if 'std::unordered_map' and 'find' were used instead of a loop?

In fact, in this use case there are no significant differences in terms of average server sizes, but for a larger search operation it is really better to use unordered_map and its method. If someone wants to refactor it, they have to modify the entire class, and this pull request is not intended for that.

Fernando-A-Rocha (approved):

Client/mods/deathmatch/logic/CPacketHandler.h
```diff
@@ -38,6 +38,7 @@ class CPacketHandler
         ELEMENT_FAILURE,
         GENERAL_REFUSED,
         SERIAL_VERIFICATION,
+        SERIAL_DUPLICATE,
```
TheNormalnij: Move SERIAL_DUPLICATE to the end of the list. It breaks packets from an old sever

Server/mods/deathmatch/logic/packets/CPlayerDisconnectedPacket.h
```diff
@@ -35,6 +35,7 @@ class CPlayerDisconnectedPacket final : public CPacket
         ELEMENT_FAILURE,
         GENERAL_REFUSED,
         SERIAL_VERIFICATION,
+        SERIAL_DUPLICATE,
```
TheNormalnij: Move SERIAL_DUPLICATE to the end of the list.

Server/mods/deathmatch/logic/CPlayerManager.h
```diff
@@ -40,6 +41,7 @@ class CPlayerManager
 
     CPlayer* Get(const NetServerPlayerID& PlayerSocket);
     CPlayer* Get(const char* szNick, bool bCaseSensitive = false);
+    const CPlayer* GetBySerial(const std::string serial) const noexcept;
```
TheNormalnij: `const CPlayer*` is unnecessary. Use `CPlayer*`

Server/mods/deathmatch/logic/CPlayerManager.h
```diff
@@ -17,6 +17,7 @@ class CPlayerManager;
 #include "packets/CPacket.h"
 #include "CPlayer.h"
 #include "../Config.h"
+#include <string>
```
TheNormalnij: it looks like `std::string` is defined already

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1798,6 +1798,17 @@ void CGame::Packet_PlayerJoinData(CPlayerJoinDataPacket& Packet)
...
         }
 
+        // Check if another player is using the same serial
+        if (m_pMainConfig->IsCheckDuplicateSerialsEnabled() && m_pPlayerManager->GetBySerial(strSerial))
+        {
+            // Tell the console
+            CLogger::LogPrintf("CONNECT: %s failed to connect (Serial already in use) (%s)\n", szNick, strIPAndSerial.c_str());
+
+            // Tell the player the problem
+            DisconnectPlayer(this, *pPlayer, CPlayerDisconnectedPacket::SERIAL_DUPLICATE);
```
TheNormalnij: Add bistream version check for old clients and use `KICK` for them

Server/mods/deathmatch/logic/CPlayerManager.h
TracerDS: An overload would be better here instead of removing constness.

TracerDS (approved):

Server/mods/deathmatch/logic/CPlayerManager.h
TheNormalnij: `GetBySerial` is ok here. An overload is evil here.

TheNormalnij: Use string_view

TracerDS: Why is it evil in your opinion?

TheNormalnij: Because `Get` searches by nickname for a string argument currently. Two different behavior for different string types are evil

TracerDS: huh? What are you talking about?

It should return const CPlayer* in const context and normal CPlayer* in non-const context. This is not evil. In fact, its standard practice that you dont seem to use resulting in making everything non-const because everything is non-const. **That** is evil

TheNormalnij: > huh? What are you talking about?

About your purpose to use overload.

> It should return const CPlayer* in const context and normal CPlayer* in non-const context. This is not evil. In fact, its standard practice that you dont seem to use resulting in making everything non-const because everything is non-const. **That** is evil

This method returns `const CPlayer*` for any context. It request a const_cast or an additional method to modify returned value here. This method will work in const context too. `const` is redundant here.

TracerDS: > This method returns `const CPlayer*` for any context. It request a const_cast or an additional method to modify returned value here. This method will work in const context too. `const` is redundant here.

You dont use const_cast because you are not supposed to modify the object. **Thats the point!**
If you want to modify the object, you dont use const. **Thats the point!**
Its good practice to mark it as const if you dont intent to modify whatever it returns (for pointers that is)

TheNormalnij: There is no reason why `GetBySerial` should return a non-modifiable object. `GetBySerial` is const itself and it's enough.

> If you want to modify the object, you dont use const. **Thats the point!**

So you purpose an additional method `GetBySerialNonConst` to return a modifiable object.

TracerDS: > GetBySerial is const itself and it's enough.

Thats not how it works.

> So you purpose an additional method GetBySerialNonConst to return a modifiable object.

No. Adding an additional overload for non-const context would be better.

Client/mods/deathmatch/logic/CPacketHandler.h
ghost: Thanks!

Server/mods/deathmatch/logic/CPlayerManager.h
ghost: I'll just go with the solution suggested by TheNormalnij.
I don't think we should put so much emphasis on it, it's really not that interesting to debate..

Server/mods/deathmatch/logic/CPlayerManager.h
ghost: Thanks!

Server/mods/deathmatch/logic/CPlayerManager.h
ghost: Modified.

lopezloo:
> when they try to spoof a server's moderator's serial as their own

How would they know server's moderator serial?

ghost:
> How would they know server's moderator serial?

From servers with a loophole or a data leak

Server/mods/deathmatch/logic/CGame.cpp
ghost: Thanks!

Server/mods/deathmatch/logic/CPlayerManager.cpp
```diff
@@ -138,6 +138,19 @@ CPlayer* CPlayerManager::Get(const char* szNick, bool bCaseSensitive)
...
 }
 
+CPlayer* CPlayerManager::GetBySerial(const std::string_view serial) const noexcept
+{
+    for (const auto& player : m_Players)
+    {
+        if (player->GetSerial() == serial)
+        {
+            return player;
+        }
```
Fernando-A-Rocha: You can remove these { }

Fernando-A-Rocha (changes requested):
Other than that, good job!

Server/mods/deathmatch/logic/CPlayerManager.cpp
Fernando-A-Rocha:
```suggestion
        if (player->GetSerial() == serial)
            return player;
```

Fernando-A-Rocha (approved):

PR #3862, opened 2024-11-19
Fix MinClientReqCheck and improve resource upgrade

Fernando-A-Rocha:
## Details
- I found that `MinClientReqCheck` was not working on any builds because it was checking `MTASA_VERSION_TYPE`, which was not defined due to a missing `version.h` import. I have restored this include in server `StdInc.h`, as it was back in the day (before a big refactor that removed this by mistake, see https://github.com/multitheftauto/mtasa-blue/pull/3853#issuecomment-2472903077). This was resulting in clientside functions like fetchRemote which use old syntax not outputting any warnings.

- I changed the Resource Upgrade feature to set the `min_mta_version` meta.xml node with the `"both"` attribute when the client & server versions to set are the same (`CResourceChecker.cpp`). Attribute `"both"` is already correctly supported on resource parsing/loading.

- I changed the way base64Encode and base64Decode functions are detected by the resource checker (& upgrader). Previously, the upgrader would automatically rename these functions in your scripts to encodeString and decodeString, causing the scripts to break because these functions take different arguments (different syntaxes). Now, instead of doing that automatically, it warns you to manually update, like it does for other functions such as `givePlayerJetPack`.

## Tests

Place this code on `client.lua` and `server.lua` script files in a new resource.
```lua
fileGetContents("doesnt_matter")
```

Start the server to load all resources or start the resource if the server was already running.

The resource upgrader will detect the missing `min_mta_version` information due to this new 1.6.0 function being used, which was added in 1.6.0-9.21938 for both client and server.

The upgrade command will result in the meta.xml being successfully updated with the new `"both"` `min_mta_version` required.

image

### `base64Encode` and `base64Decode` detection

Place this code in a serverside script (taken from the wiki).
```lua
local k = base64Encode("Hello, world!")
outputServerLog(k)
--Output: SGVsbG8sIHdvcmxkIQ==
```

Start the resource. We now observe a warning asking you to update the function manually, where previously it would wrongfully replace the function name (base64Encode) with encodeString which has a different syntax.

```log
WARNING: testres/server.lua(Line 1) [Server] base64Encode no longer works. Please manually change this to encodeString (different syntax). Refer to the wiki for details
```

### DIFF
--- a/Server/mods/deathmatch/StdInc.h
+++ b/Server/mods/deathmatch/StdInc.h
@@ -49,6 +49,7 @@
 #include <bochs_internal/bochs_crc32.h>
 #include <pcrecpp.h>
 #include <pthread.h>
+#include "version.h"
 
 extern class CNetServer* g_pRealNetServer;
 extern class CGame*      g_pGame;
--- a/Server/mods/deathmatch/logic/CResource.cpp
+++ b/Server/mods/deathmatch/logic/CResource.cpp
@@ -35,7 +35,6 @@
 #include <net/SimHeaders.h>
 #include <zip.h>
 #include <glob/glob.h>
-#include "version.h"
 #include "CStaticFunctionDefinitions.h"
 
 #ifdef WIN32
--- a/Server/mods/deathmatch/logic/CResourceChecker.Data.h
+++ b/Server/mods/deathmatch/logic/CResourceChecker.Data.h
@@ -167,8 +167,8 @@ namespace
         //{false, "doesPedHaveJetPack", "isPedWearingJetpack"},
 
         // Base Encoding & Decoding
-        {false, "base64Encode", "encodeString"},
-        {false, "base64Decode", "decodeString"},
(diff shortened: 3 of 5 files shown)


ghost (commented):
Perfect so far!

TracerDS (approved):

Fernando-A-Rocha:
## Some facts

Regarding resource upgrader:

I found that the upgrade command can only upgrade resources by setting a new min_mta_version client/server setting via the `CResourceChecker::CheckVersionRequirements` function that depends on client and server min versions listed in `CResourceChecker.Data.h`

image

Other older warnings like fetchRemote, callRemote etc cannot be automatically resolved with upgrade

They only pollute the debug console until the developer manually does something about the script(s).

image

Fernando-A-Rocha:
> Perfect so far!

I'm gonna look into what you said on discord:

image

Fernando-A-Rocha:
@Nico8340 Ive committed a solution for base64Encode and base64Decode.

Main PR post updated.

Dutchman101:
Nice

Issue #375, opened 2018-09-02
Sync distance of unoccupied vehicles and peds should match stream distance


ArranTuna:
**Describe the bug**

@lil-Toady said:
> You can have unoccupied vehicles and peds streamed in when nobody really syncs them, that can cause troubles if you interact with them (shoot?).

**To Reproduce**

Script from Drakath:

```
function distancedSync()
    for i, player in ipairs(getElementsByType("player")) do
        setElementPosition(player, 2243, 1532, 11.5)
        setElementRotation(player, 0, 0, 90)
        giveWeapon(player, 38, 9999, true)
end
	local ped = createPed (0, 2114, 1532, 11.5, 0)
	setTimer ( function (ped) 
	    setElementRotation(ped, 0, 0, -90)
	    setPedAnimation (ped, "ped", "run_old", -1, true, true, true)
	end, 1000, 1, ped)
end
addEventHandler("onResourceStart", root, distancedSync)
--This ped will be invincible
```

**Expected behavior**

@lil-Toady said:

> Raise sync distance to match stream distance and make the streamer load elements the player syncs even if they are out of local stream distance. So anything a player can see is always synced.
>
> A cool solution instead would be for the streamer not to stream in peds/vehicles if they aint being synced by anyone.

**Additional context**
From https://bugs.mtasa.com/view.php?id=6287

PR #3863, opened 2024-11-21
Fixed a bug with desynchronization of the values of some fields of the CTickRateSettings structure.

nweb:
Fixes a bug where server configuration fields (ped_syncer_distance and unoccupied_vehicle_syncer_distance) did not affect the game in any way, i.e. the client side used its default values ​​(easy to check with the /sinfo command).


Fernando-A-Rocha:
Ped syncer distance and unoccupied vehicle syncer distance were not being synced server<->client at all !? 😮

TracerDS (approved):

Proxy-99:
maybe this related to #375

Dutchman101:
That's quite a big oversight, thanks for the fix!

theSarrum:
> Ped syncer distance and unoccupied vehicle syncer distance were not being synced server<->client at all !? 😮

On the client side this is only used by the /sinfo command.

nweb:
> > Ped syncer distance and unoccupied vehicle syncer distance were not being synced server<->client at all !? 😮
> 
> On the client side this is only used by the /sinfo command.

It cannot be said that these parameters are not used in the private netc.dll

PR #3864, opened 2024-11-22
Make some sync parameter configurable.

nweb:
Adds the object_sync_interval parameter to the server configuration file. It is already supported, but for some unknown reason it could not be configured.


Proxy-99:
related to #460 #3405

ghost (approved):
LGTM

PR #3868, opened 2024-11-24
Fix CClientGUIManager::Exists

G-Moris:
Under strange conditions, nullptr passed here and caused a crash.

I had a crash here after a sharp protocol error and reconnection.


ghost (approved):

Dutchman101:
Thanks

PR #3873, opened 2024-11-26
Add new event onPlayerChangesWorldSpecialProperty

ghost:
### Description
This pull request adds a new event called `onPlayerChangesWorldSpecialProperty`, which has a `property` and `enabled` parameter, and is designed to track the use of the `setWorldSpecialPropertyEnabled` function, which is also perfect for detecting cheats.

### Implementation
The implementation required the creation of a new packet, the identifier of which is `PACKET_ID_PLAYER_WORLD_SPECIAL_PROPERTY`, and for backward compatibility, a new BitStream version was added, which is called `WorldSpecialPropertyEvent`

### Example
```lua
local cheats = {
    ["hovercars"] = true,
    ["aircars"] = true,
    ["extrabunny"] = true,
    ["extrajump"] = true
}

addEventHandler("onPlayerChangesWorldSpecialProperty", root,
    function(property, enabled)
        if not cheats[property] then
            return
        end

        if not enabled then
            return
        end

        banPlayer(source, false, false, true, "Server", "Time to take a permanent break :-)")
    end
)
```


TracerDS:
What if the cheater wont send the packet to the server?

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4256,6 +4264,25 @@ void CGame::Packet_PlayerResourceStart(CPlayerResourceStartPacket& Packet)
...
 
+void CGame::Packet_PlayerWorldSpecialProperty(CPlayerWorldSpecialPropertyPacket& packet) noexcept
+{
+    CPlayer* player = packet.GetSourcePlayer();
+
+    if (!player)
+    {
+        return;
+    }
+
```
TracerDS: No need for {}

TracerDS (approved):

Server/mods/deathmatch/logic/CGame.cpp
ghost: Thx

ghost:
> What if the cheater wont send the packet to the server?

This is not something we can do anything with, clearly it's not completely foolproof like any other protection, but it provides full protection against cheats executed from Lua

Lamaaar:
It's better then using addDebugHook imo

TheNormalnij:
> > What if the cheater wont send the packet to the server?
> 
> This is not something we can do anything with, clearly it's not completely foolproof like any other protection, but it provides full protection against cheats executed from Lua

it doesn't make any sense because the world properties can be overridden in the same way that lua injection works

ghost:
> it doesn't make any sense because the world properties can be overridden in the same way that lua injection works

It can be bypassed like any other protection as I mentioned, and this one in particular, but I don't think that means it's unnecessary or pointless, because additional knowledge is needed to avoid sending the packet or to change the called function, and let's face it, most cheaters don't have any knowledge of coding in general, because they use free, downloaded cheats from the internet, most of the time not even fully functional ones.

Lpsd (approved):

Lpsd:
Thanks for the PR, please create the wiki page.

Laxante101:
> E se o trapaceiro não enviar o pacote para o servidor?

the server should periodically check the state of special world properties to ensure that they are not activated inappropriately. This can be done with a function that sends back the expected states and compares the return

ghost:
> Thanks for the PR, please create the wiki page.

Done

TheNormalnij:
> > it doesn't make any sense because the world properties can be overridden in the same way that lua injection works
> 
> It can be bypassed like any other protection as I mentioned, and this one in particular, but I don't think that means it's unnecessary or pointless, because additional knowledge is needed to avoid sending the packet or to change the called function, and let's face it, most cheaters don't have any knowledge of coding in general, because they use free, downloaded cheats from the internet, most of the time not even fully functional ones.

This is very weak protection. Cheaters have their own communities and share cheats/code. They will use setElementPosition function to fly instead. addDebugHook covers this case, imho

Xenius97:
What if player enables a world property while is not connected to a server? Is that even possible?

ghost:
> What if player enables a world property while is not connected to a server? Is that even possible?

I don't think that's possible.
If it is, they will exploit other vulnerabilities sooner, but I wouldn't go into detail about that here, I don't want to give anyone ideas...

Lpsd:
Fixes made in caa5c4f

PR #3878, opened 2024-12-07
Temporary fix projectiles crash during disconnecting #3876

### LINKED ISSUE #3876: createProjectile crashes the game after quitting / getting kicked
### Describe the bug

as in title, after quitting / getting kicked, the game crashes right before the explosion, For some reason the crash happened if you created a +15 projectile, tested on ID (16,18,19,20,21)
```
Version = 1.6-release-22789.0.000
Time = Thu Dec  5 21:52:41 2024
Module = C:\ProgramData\MTA San Andreas All\1.6\GTA San Andreas\gta_sa.exe
Code = 0xC0000005
Offset = 0x00171BA0

EAX=80808080  EBX=00000000  ECX=100F0F00  EDX=C417A19C  ESI=00C89018
EDI=00C89030  EBP=100F0F00  ESP=0177ED60  EIP=00571BA0  FLG=00210282
```

### Steps to reproduce

1. start runcode
2. `crun for i = 1, 15 do createProjectile(me, 16, me.position) end`
3. use /quit or click on **Disconnect** button in the main menu 

### Version

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

FileEX:
Fixed #3876 

The crash occurs in the function ``CEntity::RegisterReference`` because, at the moment of creating an explosion for a projectile, for an unknown reason, the memory where the projectile's creator (the player disconnecting from the server) is located is freed and becomes ``0x80808080``. I haven’t investigated the cause of where and why the memory is being freed before the projectile is destroyed. In any case, removing projectiles without the explosion resolves the crash. Besides, the explosion during the server quit doesn’t make sense anyway because we can't see it

### DIFF
--- a/Client/game_sa/CProjectileInfoSA.cpp
+++ b/Client/game_sa/CProjectileInfoSA.cpp
@@ -46,6 +46,11 @@ void CProjectileInfoSA::RemoveProjectile(CProjectileInfo* pProjectileInfo, CProj
         // Has it not already been removed by GTA?
         if (pProjectileInfo->IsActive())
         {
+            // Fix crash when disconnecting
+            // GH #3876
+            if (!projectileInfoInterface->pEntProjectileOwner || projectileInfoInterface->pEntProjectileOwner->nType == ENTITY_TYPE_NOTHING || projectileInfoInterface->pEntProjectileOwner->pReferences == reinterpret_cast<void*>(0x80808080))
+                bBlow = false;
+
             if (bBlow)
             {
                 DWORD dwFunc = FUNC_RemoveProjectile;


Client/game_sa/CProjectileInfoSA.cpp
```diff
@@ -46,6 +46,11 @@ void CProjectileInfoSA::RemoveProjectile(CProjectileInfo* pProjectileInfo, CProj
         // Has it not already been removed by GTA?
         if (pProjectileInfo->IsActive())
         {
+            // Fix crash when disconnecting
+            // GH #3876
+            if (!projectileInfoInterface->pEntProjectileOwner || projectileInfoInterface->pEntProjectileOwner->nType == ENTITY_TYPE_NOTHING || projectileInfoInterface->pEntProjectileOwner->pReferences == reinterpret_cast<void*>(0x80808080))
+                bBlow = false;
```
TheNormalnij, line 52: You need to find a reason why pEntProjectileOwner was freed.
It looks like MTA doesn't delete all projectiles while disconnecting.

TracerDS (approved):

Client/game_sa/CProjectileInfoSA.cpp
FileEX: This is not the goal of this PR. This PR is only a temporary solution to prevent the game from crashing. If you want, feel free to investigate the cause and open a PR with the fix, and then I'll close mine. For now, this PR is just a temporary crash patch.

TheNormalnij: Temporary solutions have a high chance to become permanent.
The solution is too dirty. We don't rush with this issue.

PR #3880, opened 2024-12-09
Allow dynamic models to be created as buildings

TheNormalnij:


Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -70,9 +77,12 @@ CBuilding* CBuildingsPoolSA::AddBuilding(CClientBuilding* pClientBuilding, uint1
     pBuilding->m_pLod = nullptr;
     pBuilding->m_iplIndex = 0;
 
+    // Resore changes properties group
```
FileEX: Typo

TheNormalnij: thx

TheNormalnij:
```suggestion
    // Restore changed properties group
```

Client/game_sa/CModelInfoSA.h
```diff
@@ -466,7 +466,7 @@ class CModelInfoSA : public CModelInfo
     // Vehicle towing functions
     bool IsTowableBy(CModelInfo* towingModel) override;
 
-    bool IsDynamic() { return m_pInterface ? m_pInterface->usDynamicIndex != 0xffff : false; };
+    bool IsDynamic() { return m_pInterface ? m_pInterface->usDynamicIndex != MODEL_PROPERTIES_GROUP_STATIC : false; };
```
TracerDS, line 488: Maybe const?
I know you arent introducing a new function but if you're at refactoring then you can as well mark it as const

TracerDS (approved):

CArg22:
Review and comments "maybe const", "typo" are pointless, those changes are invisible for end user and have no effect on functionality.

Instead You should checkout branch and test if it doesn't crash, does it work but the problem i see is that @TheNormalnij didn't provide any instruction what he actualy changed.

Also, he should include example script ideally in code so i can run some .bat script that install for me "test scripts" then i can "/run 3880" and see if it after year of other changes still work as intended.

TracerDS:
> Review and comments "maybe const", "typo" are pointless, those changes are invisible for end user and have no effect on functionality

The same could be said about any refactors, ArgumentParser, utilities, changing vendors, etc. Completely irrelevant and a huge blunder. Just because the end user doesn't notice doesn't mean its pointless.

TheNormalnij:
> Review and comments "maybe const", "typo" are pointless, those changes are invisible for end user and have no effect on functionality.

"typo" is relevant. "maybe const" is less relevant, but ok here.

> Instead You should checkout branch and test if it doesn't crash, does it work but the problem i see is that @TheNormalnij didn't provide any instruction what he actualy changed.

It's not a big feature. I replaced a dirty hack to add behavior for createBuilding from it.
The test command is simply: `crun createBuilding(1502, 0, 0, 4)` - creates a door without physics.

TracerDS:
> > Review and comments "maybe const", "typo" are pointless, those changes are invisible for end user and have no effect on functionality.
> 
> "typo" is relevant. "maybe const" is less relevant, but ok here.

Its actually the opposite:
Typo in comments make no difference to functionality, only reading.
const makes a significant difference to functionality, but no readability difference

In either case both are relevant

TheNormalnij:
> In either case both are relevant

It is less relevant because it's outside of the PR scope. I didn't add that method. Only a constant was replaced there.

CArg22:
> > Review and comments "maybe const", "typo" are pointless, those changes are invisible for end user and have no effect on functionality.
> 
> "typo" is relevant. "maybe const" is less relevant, but ok here.
> 
> > Instead You should checkout branch and test if it doesn't crash, does it work but the problem i see is that @TheNormalnij didn't provide any instruction what he actualy changed.
> 
> It's not a big feature. I replaced a dirty hack to add behavior for createBuilding from it. The test command is simply: `crun createBuilding(1502, 0, 0, 4)` - creates a door without physics.

And thats what should be part of pull request, resource that i can easily grab and quickly test some previous pull request. In your example it could spawn this door with helper text "you should be able to go through the door". This pattern should be applied to all pull requests ( if possible ) to speedup testing and reduce regression.

TracerDS:
> And thats what should be part of pull request, resource that i can easily grab and quickly test some previous pull request. In your example it could spawn this door with helper text "you should be able to go through the door". This pattern should be applied to all pull requests ( if possible ) to speedup testing and reduce regression.

Lets be realistic. With the huge decrease in PR activity this will be impossible to keep track of.
Especially with quick prs that are ready to merge (~~_**but stay in the ready-for-merge state for years**_~~).
If you want to test it, wait for it to get merged. That way more people can report potential issues and not just a selected few who opted into this.

PR #3866, opened 2024-11-22
Fix the BitStream version mistake made in #3863

ghost:
Fixes mistake in #3863 by adding a new BitStream version called `FixSyncerDistance` and associated checks

PR #3883, opened 2024-12-14
Check before sending to prevent crash because of #3866

### LINKED ISSUE #3882: setServerConfigSetting crashes on latest nightly
### Steps to reproduce

```lua
setServerConfigSetting("ped_sync_interval", 200, true)
```

### Version

22815 x64 (may x86 too)

### Security Policy

ghost:
Checks if the player exists and only then performs the next check which applies to the BitStream version.
Closes #3882.


TheNormalnij:
won't work

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -10862,14 +10862,16 @@ bool CStaticFunctionDefinitions::SendSyncIntervals(CPlayer* pPlayer)
     BitStream.pBitStream->Write(g_TickRateSettings.iKeySyncRotation);
     BitStream.pBitStream->Write(g_TickRateSettings.iKeySyncAnalogMove);
 
-    if (pPlayer->CanBitStream(eBitStreamVersion::FixSyncerDistance))
+    if (pPlayer)
     {
-        BitStream.pBitStream->Write(g_TickRateSettings.iPedSyncerDistance);
-        BitStream.pBitStream->Write(g_TickRateSettings.iUnoccupiedVehicleSyncerDistance);
-    }
```
TheNormalnij, line 10869: You need min client version check here.
The new client always expects all data.

ghost: Well, thanks for reporting the issue, but could you please elaborate on it and how exactly it should be solved?

PR #3884, opened 2024-12-14
New dialog for user assisted game searching

ghost:
New dialog that describes the game search process more informatively, informs the user about the importance of the Program Files (x86) folder, and removes the unnecessary, misleading progress bar.

### DIFF
--- a/Client/loader/Utils.cpp
+++ b/Client/loader/Utils.cpp
@@ -22,6 +22,7 @@
 #include <Softpub.h>
 #include <wintrust.h>
 #include <version.h>
+#include <windows.h>
 #pragma comment (lib, "wintrust")
 
 namespace fs = std::filesystem;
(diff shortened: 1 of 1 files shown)


TracerDS:
Why do we even care about the gta being put in `Program Files (x86)`?

Client/loader/Utils.cpp
```diff
@@ -527,30 +528,26 @@ bool LookForGtaProcess(SString& strOutPathFilename)
...
+static const SString DoUserAssistedSearch() noexcept
 {
-    SString strResult;
+    SString result;
 
-    ShowProgressDialog(g_hInstance, _("Searching for Grand Theft Auto San Andreas"), true);
+    MessageBox(nullptr, _("Start Grand Theft Auto: San Andreas.\nEnsure the game is placed in the 'Program Files (x86)' folder."), _("Searching for GTA: San Andreas"), MB_OK | MB_ICONINFORMATION);
 
-    while (!UpdateProgress(0, 100, _("Please start Grand Theft Auto San Andreas")))
+    while (true)
```
TracerDS, line 537: maybe do while instead?

TracerDS (changes requested):

MegadreamsBE:
> Why do we even care about the gta being put in `Program Files (x86)`?

It's not actually necessary. However, for most players with this issue, this is the solution. Some players install their GTA SA in OneDrive, or put it in another obscure locations.

TracerDS:
> It's not actually necessary. However, for most players with this issue, this is the solution. Some players install their GTA SA in OneDrive, or put it in another obscure locations.

The only issue where the location might eb a problem is the onedrive folder or folders that you as the user dont have access to.
Any other location is fine. This PR should at most inform the user to put gtasa in the **local drive** folder instead of some cloud folder.

Fernando-A-Rocha (approved):

Dutchman101:
Actually, im sure that in the last couple months, a change broke GTA:SA installations in non-standard install locations/drives (not talking about cloud or OneDrive, it takes much less to bug out). Users that never had any issues nor relocated their GTA install started to get the failed to find game dialog, there was a huge uptick in support requests and it wasn't even something they had in common like a specific source or version of the game.

Needs more investigation and an issue report filed. It may also be triggered by AC changes, making it harder to identify and requiring investigation from a select few people from the team. It began around the time various security fixes were added.

PR #3896, opened 2024-12-22
Fix CVector optional arguments - Follow-up

### LINKED ISSUE #3895: Incorrect position in some functions after "Fix CVector..." (ZYX instead of XYZ in createBuilding and getElementsWithinRange).
### Describe the bug

Since the last update, buildings are being created at reversed coordinates from those entered (X/RX is swapped with Z/RZ). The same issue affects getElementsWithinRange, which returns elements at a different position than the one provided. It's most likely broken by the "Fix CVector optional arguments - #3738 (#3782)" update.

### Steps to reproduce

1. run 
``` 
building = createBuilding(6295,10,20,30,0,0,90);
iprint("position",getElementPosition(building));
iprint("rotation",getElementRotation(building)) 
```
2. Look at result

### Version

Multi Theft Auto v1.6-release-22837

### Additional context

Image

### Relevant log output

```shell
"position" 30 20 10  
"rotation" 90 0 0
```

### Security Policy

TracerDS:
Fixes #3895


Shared/mods/deathmatch/logic/lua/CLuaFunctionParser.h
```diff
@@ -536,7 +536,11 @@ struct CLuaFunctionParserBase
         else if constexpr (std::is_same_v<T, CVector2D>)
         {
             if (lua_isnumber(L, index))
-                return {PopUnsafe<float>(L, index), PopUnsafe<float>(L, index)};
```
tederis, line 539: Why would we need to get rid of list initialization?

TracerDS: Better to explicitly return a class rather than have the compiler do some magic behind it that may brind unexpected consequences. Doesn't really matter and `CVector2D()` is more verbose.

tederis: How about `CVector2D{PopUnsafe<float>(L, index), PopUnsafe<float>(L, index)}`?

TracerDS: it may or may not be specifed depending on the C++ version.
Better to define x,y,z beforehand

tederis: I don't expect that C++ version will be downgraded some day. That means that CVector2D{} will work in the foreseeable future without any problems. Your approach is cumbersome and can lead to performance problems(depends on the compiler).

TracerDS: There are no performance problems here. Either way it will still call the constructor. Doesnt matter if its `()` or `{}`. The performance difference is imperceptible

tederis: Your current implementation unreasonably inflates the codebase and you should stick to the original list initialization.

TracerDS: Original implementation is confusing and a small change breaks the entire codebase.
It absolutely does not inflate the codebase. Function calls that modify global objects (including themselves) should be cached beforehand.

PR #3848, opened 2024-11-08
Optimization Handling (part 3) and Remove big use eVehicleTypes

G-Moris:
Continuation of PR - #3580;
Removed the frequent useless use of eVehicleTypes.

PR #3900, opened 2024-12-23
Other fixes for #3848

G-Moris:
These are bug fixes and what was not implemented that was in the plans

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
@@ -13,6 +13,7 @@
 #include <game/CHandlingEntry.h>
 #include <game/CHandlingManager.h>
 #include "lua/CLuaFunctionParser.h"
+#include <CClientVehicleManager.h>
 
 void CLuaVehicleDefs::LoadFunctions()
 {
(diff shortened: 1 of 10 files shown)


tederis (approved):
I'm going to merge it faster than normal because this PR fixes the assertation problem described here: https://github.com/multitheftauto/mtasa-blue/commit/c0376c9c258f0e8c093c4950c231bd1301623409#r150652050.

G-Moris:
If a problem is found, write here, I will fix it as soon as possible.

Client/mods/deathmatch/logic/luadefs/CLuaVehicleDefs.cpp
```diff
@@ -2668,200 +2669,213 @@ int CLuaVehicleDefs::GetVehicleHandling(lua_State* luaVM)
...
                 argStream.SetCustomError("Invalid property");
                 m_pScriptDebugging->LogCustom(luaVM, argStream.GetFullErrorMessage());
                 lua_pushboolean(luaVM, false);
             }
             return 1;
         }
-        CHandlingEntry* pEntry = pVehicle->GetHandlingData();
 
-        lua_newtable(luaVM);
+        if (const auto* const entry = pVehicle->GetHandlingData())
```
TheNormalnij, line 2729: Should never happen. Handling is just broken after this chain of PR's

PR #3903, opened 2024-12-23
Dark titlebar for dark windows mode

### LINKED ISSUE #3859: Make the window edges dark when Windows is in dark mode

Currently if you run MTA in windowed mode, window edge is white
Image

Detect system mode and use dark edges line other programs
Image

### Security Policy

FileEX:
Resolved #3859 

The title bar is set to dark if the Windows theme is dark or remains light for the light Windows theme.
image

### DIFF
--- a/Client/core/DXHook/CProxyDirect3D9.cpp
+++ b/Client/core/DXHook/CProxyDirect3D9.cpp
@@ -10,6 +10,8 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <dwmapi.h>
+
 HRESULT HandleCreateDeviceResult(HRESULT hResult, IDirect3D9* pDirect3D, UINT Adapter, D3DDEVTYPE DeviceType, HWND hFocusWindow, DWORD BehaviorFlags,
                                  D3DPRESENT_PARAMETERS* pPresentationParameters, IDirect3DDevice9** ppReturnedDeviceInterface);
 std::vector<IDirect3D9*> ms_CreatedDirect3D9List;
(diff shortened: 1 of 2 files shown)


ghost (approved):
Good job!

Fernando-A-Rocha (approved):
Very satisfying

Client/core/DXHook/CProxyDirect3D9.cpp
```diff
@@ -166,6 +168,20 @@ HRESULT CProxyDirect3D9::CreateDevice(UINT Adapter, D3DDEVTYPE DeviceType, HWND
     SetWindowTextW(hFocusWindow, MbUTF8ToUTF16("MTA: San Andreas").c_str());
     #endif
 
+    // Set dark titlebar if needed
+    HKEY hKey;
+    DWORD systemThemeMode = 1;
+    DWORD dataSize = sizeof(systemThemeMode);
+
+    if (RegOpenKeyExW(HKEY_CURRENT_USER, L"Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", 0, KEY_READ, &hKey) == ERROR_SUCCESS)
```
tederis: There is the dedicated MTA function for this: `ReadRegistryStringValue` (and even better `GetSystemRegistryValue)`.

FileEX: Thanks, done

Client/core/DXHook/CProxyDirect3D9.cpp
```diff
@@ -166,6 +168,10 @@ HRESULT CProxyDirect3D9::CreateDevice(UINT Adapter, D3DDEVTYPE DeviceType, HWND
     SetWindowTextW(hFocusWindow, MbUTF8ToUTF16("MTA: San Andreas").c_str());
     #endif
 
+    // Set dark titlebar if needed
+    BOOL darkTitleBar = GetSystemRegistryValue((uint)HKEY_CURRENT_USER, "Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", "AppsUseLightTheme") == "\x0";
```
botder, line 172: You don't have to check this, because applying `TRUE` will just tell WDM to respect the user's setting (auto/dark/light mode).

FileEX: I think you're wrong. For me with a light theme, the title bar remains dark. From what I know, ``DwmSetWindowAttribute`` sets attributes independently of the theme, doesn't it?

image

botder: When I tested it the last time, it was working as advertised.

> **DWMWA_USE_IMMERSIVE_DARK_MODE**
Use with [DwmSetWindowAttribute](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/nf-dwmapi-dwmsetwindowattribute). Allows the window frame for this window to be drawn in dark mode colors when the dark mode system setting is enabled. For compatibility reasons, all windows default to light mode regardless of the system setting. The pvAttribute parameter points to a value of type BOOL. TRUE to honor dark mode for the window, FALSE to always use light mode.
> 
> This value is supported starting with Windows 11 Build 22000.  

https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute

FileEX: In that case, I don't know why, but that's how it works. TRUE always means dark, and FALSE always means light, at least on my system. Anyway, since there's a chance (or it's the current behavior), it's better to keep such a check. It's not expensive and helps avoid a potential issue with an incorrect title bar theme

botder: Yeah, you're right.

botder (approved):

Issue #3891, opened 2024-12-20
DxDrawModel3D Problems


ghost:
### Describe the bug

At the moment I'm using DxDrawModel3D, the lighting looks like this only on id models of the weapons

![Image](https://github.com/user-attachments/assets/40c2422e-6ae2-44e7-ab01-44ebdf6fabe2)

### Steps to reproduce

Using the same structure as in the wiki but adding more models to DxDraw
https://wiki.multitheftauto.com/wiki/DxDrawImage3D

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

ghost:
The first image is using the [mtasa-pAttach](https://github.com/Patrick2562/mtasa-pAttach) system:
![Image](https://github.com/user-attachments/assets/2abe4fd2-6238-4b79-a52b-5f59286edeba)

The second image uses the [dxDrawModel3D](https://wiki.multitheftauto.com/wiki/DxDrawModel3D) function:
![Image](https://github.com/user-attachments/assets/31fd93e5-e365-4446-acf6-0cb3c6ce0eeb)

ghost:
Another problem, when I define an id for an interior, the ids defined in `DxDrawModel3D` do not appear, but when I am in interior 0 and define any dimension, it is successful.

tederis:
Lighting is a property of objects (CObject). dxDrawModel3D doesn't operate with objects and cannot apply any lighing on it. You can use a shader for it in conjunction with getElementLighting.

tederis:
> Another problem, when I define an id for an interior, the ids defined in `DxDrawModel3D` do not appear, but when I am in interior 0 and define any dimension, it is successful.

You mean dxDrawModel3D doesn't work in interiors?

ghost:
> > Another problem, when I define an id for an interior, the ids defined in `DxDrawModel3D` do not appear, but when I am in interior 0 and define any dimension, it is successful.
> 
> You mean dxDrawModel3D doesn't work in interiors?

That's right, I've tested it and it's only working for the outdoors but not for the indoors

vshiroshi:
@lynconsix up!

tederis:
> > > Another problem, when I define an id for an interior, the ids defined in `DxDrawModel3D` do not appear, but when I am in interior 0 and define any dimension, it is successful.
> > 
> > 
> > You mean dxDrawModel3D doesn't work in interiors?
> 
[...]

PR #3904, opened 2024-12-24
dxDrawModel3D lighting

tederis:
The PR extends `dxDrawModel3D` by adding a new argument responsible for the model lighting. This allows to render dynamic objects with the correct lighting. Can be usefull for weapon models, but will work for any model as well. Inspired by issue #3891 . Designed to be used with `getElementLighting`.


Client/game_sa/CRendererSA.cpp
```diff
@@ -15,6 +15,9 @@
 #include "CMatrix.h"
 #include "gamesa_renderware.h"
 
+#define SetLightColoursForPedsCarsAndObjects(fMult) ((RpLight*(__cdecl*)(float))0x735D90)(fMult)
+#define SetAmbientColours() ((RpLight*(__cdecl*)())0x735D30)()
+
```
TheNormalnij, line 20: These methods should be moved to a specific class, but we don't have that class now.

TheNormalnij (approved):

Fernando-A-Rocha:
Nice seeing you and FileEX collabing with the lighting feature ;)

PR #3905, opened 2024-12-24
New function setElementLighting

FileEX:
This PR adds the ``setElementLighting`` function, which allows setting custom lighting for a specific element.

### DIFF
--- a/Client/game_sa/CPhysicalSA.h
+++ b/Client/game_sa/CPhysicalSA.h
@@ -104,8 +104,8 @@ public:
     CVector                   m_vecUnk;                                   // 280
     uint32                    m_pad4;                                     // 292
     CPtrNodeDoubleLink<void>* m_pControlCodeNodeLink;                           // 296
-    float                     m_fLighting;                                // 300
-    float                     m_fLighting2;                               // 304
+    float                     m_fLighting;                                // 300 surface brightness
+    float                     m_fLighting2;                               // 304 dynamic lighting (unused, always set to 0 in the GTA code)
     class CShadowDataSA*      m_pShadowData;                              // 308
 
     CRect*      GetBoundRect_(CRect* pRect);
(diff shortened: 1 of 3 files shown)


tederis (approved):

Fernando-A-Rocha:
Can u showcase this in action?

FileEX:
> Can u showcase this in action?

I don't think there's much to show here. You can simply adjust the brightness of elements like peds, vehicles, or objects

Fernando-A-Rocha:
> > Can u showcase this in action?
> 
> I don't think there's much to show here. You can simply adjust the brightness of elements like peds, vehicles, or objects

Ah I think a more appropriate name would be
setElementLightBrightness or setElementLightIntensity ?

FileEX:
> > > Can u showcase this in action?
> > 
> > 
> > I don't think there's much to show here. You can simply adjust the brightness of elements like peds, vehicles, or objects
> 
> Ah I think a more appropriate name would be setElementLightBrightness or setElementLightIntensity ?

This function is the missing counterpart to ``getElementLighting``. The names of these functions can be changed if someone wants, and the current ones can be marked as deprecated, although I think these names are ok

tederis:
I agree with the FileEX's arguments. `setElementLighting` and `getElementLighting` deal with the same value. And if someone wants to change a function's name then it must be applied to both functions.

AlexTMjugador:
Now that this feature has been merged for a few days, it would be great to have a wiki article documenting it.

Also, in my opinion, a well-written wiki article should clearly explain how often the game computes lighting values for elements. This information is needed for script developers to decide whether to call it one-time or on every frame. The one-time approach is sufficient if the game doesn't frequently recalculate lighting values, but if the game recalculates lighting values every frame, calling it on each frame would be necessary for an override to have visible effects, which is also a reasonable design choice for the GTA rendering engine.

FileEX:
> Now that this feature has been merged for a few days, it would be great to have a wiki article documenting it.

The article will be published when the build is released; I always make sure of that.

PR #3907, opened 2024-12-25
Replace sa icon to taskbar

G-Moris:
Скриншот 26-12-2024 020647
Скриншот 26-12-2024 020722
Скриншот 26-12-2024 020753

### DIFF
--- a/Client/core/DXHook/CProxyDirect3D9.cpp
+++ b/Client/core/DXHook/CProxyDirect3D9.cpp
@@ -11,6 +11,9 @@
 
 #include "StdInc.h"
 #include <dwmapi.h>
+#include <resource.h>
+
+extern HINSTANCE g_hModule;
 
 HRESULT HandleCreateDeviceResult(HRESULT hResult, IDirect3D9* pDirect3D, UINT Adapter, D3DDEVTYPE DeviceType, HWND hFocusWindow, DWORD BehaviorFlags,
                                  D3DPRESENT_PARAMETERS* pPresentationParameters, IDirect3DDevice9** ppReturnedDeviceInterface);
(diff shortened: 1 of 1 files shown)


ghost:
It's the smallest details that make the difference, and that's great of you

Fernando-A-Rocha:
Thank you. This is awesome. Gives a unique touch to MTA that was needed.

Client/core/DXHook/CProxyDirect3D9.cpp
```diff
@@ -172,6 +175,23 @@ HRESULT CProxyDirect3D9::CreateDevice(UINT Adapter, D3DDEVTYPE DeviceType, HWND
...
     DwmSetWindowAttribute(hFocusWindow, DWMWA_USE_IMMERSIVE_DARK_MODE, &darkTitleBar, sizeof(darkTitleBar));
 
+    // Update icon
+    if (HICON icon = LoadIcon(g_hModule, MAKEINTRESOURCE(IDI_ICON1)))
+    {
+        const auto paramIcon = reinterpret_cast<LPARAM>(icon);
+        for (const WPARAM size : {ICON_SMALL, ICON_BIG})
+        {
+            SendMessage(hFocusWindow, WM_SETICON, size, paramIcon);
+        }
```
botder, line 185: I don't really mind this, but it's overkill for you could've just duplicated the SendMessage line

botder (approved):

Client/core/DXHook/CProxyDirect3D9.cpp
G-Moris: Is it required?

T-MaxWiese-T:
So in the task manager sometimes the icon and also the subprocess name does not change when the additional subprocess `wow64_helper.exe` (helper library to run x86 applications on x64 versions of Microsoft Windows) starts automatically, which seems to happen randomly. However, the subprocess also closes again randomly, so that it works again as intended.

G-Moris:
> So in the task manager sometimes the icon and also the subprocess name does not change when the additional subprocess `wow64_helper.exe` (helper library to run x86 applications on x64 versions of Microsoft Windows) starts automatically, which seems to happen randomly. However, the subprocess also closes again randomly, so that it works again as intended.

I didn't understand you

T-MaxWiese-T:
> I didn't understand you

What exactly don't you understand?
I think the text I have written should be understandable.

PR #3909, opened 2024-12-26
Skip nick and serial checks if the server is a debug build

### LINKED ISSUE #3908: Skip serial and nick checks in DEBUG mode

I think it would be useful to skip checking whether a nickname or serial is already in use in DEBUG mode. The issue is that during testing or debugging crashes, when MTA crashes, the player is still considered to be on the server. Sometimes the player count shows 0, sometimes 1, but in reality, the server is empty. It's very frustrating to have to kick a ghost player on a local server after every crash just to be able to join. Until now, it was somewhat bearable because it didn't always happen, but after adding serial checks, it has become extremely annoying

Image

I'd like to work on crashes faster without having to kick the ghost player after every crash

_No response_

### Additional context

_No response_

### Security Policy

ghost:
Closes #3908

### DIFF
--- a/Client/multiplayer_sa/multiplayer_keysync.cpp
+++ b/Client/multiplayer_sa/multiplayer_keysync.cpp
@@ -51,7 +51,6 @@ VOID InitKeysyncHooks()
     HookInstallMethod(VTBL_CBoat__ProcessControl, (DWORD)HOOK_CBoat__ProcessControl);
     HookInstallMethod(VTBL_CBike__ProcessControl, (DWORD)HOOK_CBike__ProcessControl);
     HookInstallMethod(VTBL_CHeli__ProcessControl, (DWORD)HOOK_CHeli__ProcessControl);
-    HookInstallMethod(VTBL_CHeli__ProcessControl, (DWORD)HOOK_CHeli__ProcessControl);
 
     // not strictly for keysync, to make CPlayerPed::GetPlayerInfoForThisPlayerPed always return the local playerinfo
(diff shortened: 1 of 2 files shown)


tederis:
I don't think this is really needed. You can disable the check in `mtaserver.conf`. See `check_duplicate_serials`.

ghost:
> I don't think this is really needed. You can disable the check in `mtaserver.conf`. See `check_duplicate_serials`.

I know, I was the one who implemented it and this idea was brought up by @FileEX, but regardless I don't think it's unnecessary, it's easier than having to manually turn it off in the config

ghost:
But you're right that it's not the most important change, it can be omitted

FileEX:
> I don't think this is really needed. You can disable the check in `mtaserver.conf`. See `check_duplicate_serials`.

Can I also disable nick checks?

tederis:
> > I don't think this is really needed. You can disable the check in `mtaserver.conf`. See `check_duplicate_serials`.
> 
> Can I also disable nick checks?

I don't think so. From what I can see this PR doesn't turn off nick checks as well.

ghost:
It was a mistake, it was supposed to disable nick checks as well. But it isn't really needed, we better solve the bug itself.

FileEX:
Since no one will fix this bug anytime soon, this solution would be much better. I see nothing wrong with disabling nickname checks. If this approach is considered bad, then let's add a '`check_duplicated_nicks'` parameter to the mtaserver.conf as well.

PR #3916, opened 2025-01-01
Fix setPedControlState is aborted when ped created

### LINKED ISSUE #1201: setPedControlState is aborted when ped created
**Describe the bug**
When I move localPlayer with setPedControlState and any ped created on the server, the setPedControlState stops.

**To reproduce**
Reproduced in the video below.

**Video**
https://youtu.be/cvmeXWEL1yY

**Version**
Server: v1.5.7-release-20358 (Windows)
Client: v1.5.7-release-20358 (Windows)

### LINKED ISSUE #3867: (Client) setControlState, setPedControlState and setAnalogControlState are reset to false when a player joins the game
### Describe the bug

The control state that is set to _true_ or _1_ by one of the three functions will reset to _false_ or _0_ upon a player joining the game.
The said bug appears to be affecting both the three functions in the same way and presumably is on the side of the client.

### Steps to reproduce

1. Set the state of the control "forwards" to _true_ or _1_ with either one of the three functions.
```lua
-- Client-side
setControlState(localPlayer, "forwards", true)
-- setAnalogControlState(localPlayer, "forwards", 1)
-- setPedControlState(localPlayer, "forwards", true)
```
2. Make another player join the server.

You can now observe that the local player is no longer moving forwards.

### Version

Client: 1.6-release-22623
Server: 1.6-release-22596

FileEX:
Fix #1201 and probably #3867

The bug is caused by ``m_sScriptedStates`` being static, which led to the variable being reset by the ``CClientPad`` constructor after creating a ped
https://github.com/multitheftauto/mtasa-blue/blob/e831f62d0b4ccc11f7e4f45f1af8968c65c7fb18/Client/mods/deathmatch/logic/CClientPad.cpp#L120-L125

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPad.cpp
+++ b/Client/mods/deathmatch/logic/CClientPad.cpp
@@ -116,13 +116,6 @@ CClientPad::CClientPad()
     {
         m_fStates[i] = 0.0f;
     }
-
-    // Initialise our analog control states
-    for (unsigned int i = 0; i < MAX_GTA_ANALOG_CONTROLS; i++)
-    {
-        m_sScriptedStates[i] = CS_NAN;
-        m_bScriptedStatesNextFrameOverride[i] = false;
-    }
 }
 
 bool CClientPad::GetControlState(const char* szName, bool& bState)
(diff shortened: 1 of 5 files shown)


botder:
And why did you de-static these values intead of moving the reset out of that constructor or doing that once?

FileEX:
> And why did you de-static these values intead of moving the reset out of that constructor or doing that once?

Don't other peds need their own 'scriptedStates'? Does this only apply to the local player?

botder:
We have two functions `setPedAnalogControlState` and `setAnalogControlState`. I don't know how these functions differ right now.

Every `CClientPed` has an instance of `CClientPad m_Pad`. And the default constructor of `CClientPad` initializes the static variables everytime. The simplest solution would be to move the initialization out of the default constructor and offer a static function that gets called whenever we create the local player's ped.

FileEX:
> We have two functions `setPedAnalogControlState` and `setAnalogControlState`. I don't know how these functions differ right now.
> 
> Every `CClientPed` has an instance of `CClientPad m_Pad`. And the default constructor of `CClientPad` initializes the static variables everytime. The simplest solution would be to move the initialization out of the default constructor and offer a static function that gets called whenever we create the local player's ped.

Done

botder (approved):

PR #3917, opened 2025-01-01
New functions setWeaponRenderEnabled & isWeaponRenderEnabled

FileEX:
This PR adds functions that allow disabling the rendering of weapons. Some servers do not use the original weapons and replace them with their own system. To remove the original weapons, they need to replace the model with an empty one or manipulate the weapon model's LOD distance. The function only toggles the rendering of the weapon object, so shooting, aiming, etc., can still be done

```lua
bool setWeaponRenderEnabled(bool enabled)
bool isWeaponRenderEnabled()
```

Thanks to @ds1-e for idea

### DIFF
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -1143,6 +1143,9 @@ CWeaponStat* CGameSA::CreateWeaponStat(eWeaponType weaponType, eWeaponSkill weap
 
 void CGameSA::SetWeaponRenderEnabled(bool enabled)
 {
+    if (IsWeaponRenderEnabled() == enabled)
+        return;
+
     if (!enabled)
     {
         // Disable calls to CVisibilityPlugins::RenderWeaponPedsForPC
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -1141,6 +1141,29 @@ CWeaponStat* CGameSA::CreateWeaponStat(eWeaponType weaponType, eWeaponSkill weap
     return m_pWeaponStatsManager->CreateWeaponStatUnlisted(weaponType, weaponSkill);
 }
 
+void CGameSA::SetWeaponRenderEnabled(bool enabled)
+{
+    if (!enabled)
+    {
+        // Disable calls to CVisibilityPlugins::RenderWeaponPedsForPC
+        MemSet((void*)0x53EAC4, 0x90, 5); // Idle
+        MemSet((void*)0x705322, 0x90, 5); // CPostEffects::Render
+        MemSet((void*)0x7271E3, 0x90, 5); // CMirrors::BeforeMainRender
+    }
+    else
+    {
+        // Restore original bytes
+        MemCpy((void*)0x53EAC4, "\xE8\x67\x44\x1F\x00", 5);
+        MemCpy((void*)0x705322, "\xE8\x09\xDC\x02\x00", 5);
+        MemCpy((void*)0x7271E3, "\xE8\x48\xBD\x00\x00", 5);
+    }
+}
+
+bool CGameSA::IsWeaponRenderEnabled() const
+{
+    return *(unsigned char*)0x53EAC4 == 0xE8;
+}
+
 void CGameSA::OnPedContextChange(CPed* pPedContext)
 {
     m_pPedContext = pPedContext;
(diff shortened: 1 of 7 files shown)


ghost:
On a side note regarding why it should be used instead of other ways to hide weapon model, see [this comment](https://github.com/multitheftauto/mtasa-blue/issues/1067#issuecomment-2259291289)

ghost:
> This PR adds functions that allow disabling the rendering of weapons. Some servers do not use the original weapons and replace them with their own system. To remove the original weapons, they need to replace the model with an empty one or manipulate the weapon model's LOD distance. The function only toggles the rendering of the weapon object, so shooting, aiming, etc., can still be done
> 
> ```lua
> bool setWeaponRenderEnabled(bool enabled)
> bool isWeaponRenderEnabled()
> ```
> 
> Thanks to @ds1-e for idea

It's much better this way instead of replacing models, because replacing models for “new” weapons consumes memory.

botder:
You have to reset the state after quit/on server join.

ghost:
@FileEX 
You could add another function, `resetWeaponRender(int modelid)` and instead of disabling all weapon rendering, identify the model of the weapon you want to disable.
`setWeaponRenderEnabled(int modelid, bool enabled)` & `isWeaponRenderEnabled(int modelid)`

ghost:
@lynconsix I disagree. See my previous comment for details, there's no need to add extra steps on top of already horribly performing GTA function. If you want to selectively disable certain weapons, just use old methods (after all you should still rather use something like [pAttach](https://github.com/Patrick2562/mtasa-pAttach) for any weapon, considering it's more smoother alternative)

FileEX:
> You have to reset the state after quit/on server join.

Done

ghost:
I do agree with @lynconsix to have a modelid param for obvious reasons. Without that this new feature will only be used by servers where every weapon model is changed. (if this PR gets merged like this - it is still nice)

off topic: 
This reminds me of weapon properties.
You can't set a given player's weapon property because the func doesn't have a player/ped param so it changes for everyone and there is no `real` clientside variant of this function. The `current` clientside version of this func only works on custom weapons so the func name should be setCustomWeaponProperty

ghost:
> I do agree with @lynconsix to have a modelid param for obvious reasons. Without that this new feature will only be used by servers where every weapon model is changed. (if this PR gets merged like this - it is still nice)
> 
> off topic: This reminds me of weapon properties. You can't set a given player's weapon property because the func doesn't have a player/ped param so it changes for everyone and there is no `real` clientside variant of this function. The `current` clientside version of this func only works on custom weapons so the func name should be setCustomWeaponProperty

This can already be done using:
```lua
engineSetModelLODDistance(weaponModelID, 0.001)
```
What can't be done however, is disabling GTA function call which i've already mentioned twice (and [seems to be totally ignored](https://github.com/multitheftauto/mtasa-blue/pull/3917#issuecomment-2567094256)), being secondary purpose of it. Adding selectivity would force it to process anyways, which completely excludes it.

Client/game_sa/CGameSA.cpp
```diff
@@ -1141,6 +1141,29 @@ CWeaponStat* CGameSA::CreateWeaponStat(eWeaponType weaponType, eWeaponSkill weap
     return m_pWeaponStatsManager->CreateWeaponStatUnlisted(weaponType, weaponSkill);
 }
 
+void CGameSA::SetWeaponRenderEnabled(bool enabled)
+{
+    if (!enabled)
```
botder:
```suggestion
{
    if (IsWeaponRenderEnabled() == enabled)
        return;

    if (!enabled)
```

botder (approved):

PlatinMTA:
> What can't be done however, is disabling GTA function call which i've already mentioned twice (and [seems to be totally ignored](https://github.com/multitheftauto/mtasa-blue/pull/3917#issuecomment-2567094256)), being secondary purpose of it. Adding selectivity would force it to process anyways, which completely excludes it.

I was testing just now and I am pretty sure that ``0x4D9888`` keeps being called, it is related to the sounds right? I found no performance gain between having the weapons enabled or disabled while shooting, the only performance gain I saw was because of the usage of ``setWorldSoundEnabled(5, false)``

Heck if you add pAttach to the mix it might even run a little bit worse, but with the added benefit of being able to modify the weapon without resulting to the previous hack or an invisible shader.

PlatinMTA:
> This can already be done using:
> 
> ```lua
> engineSetModelLODDistance(weaponModelID, 0.001)
> ```

Adding to my previous comment, using this hack has the disadvantage of you being unable to use that model for anything else since it will never render, which is definitely not ideal (using a shader ends up being more convenient). Since there is no real improvement upon disabling the rendering of all weapons (the only improvement in performance is made by disabling the sounds), this function would really benefit with just having a way to disable certain weapon models only.

PR #3920, opened 2025-01-02
Add spawnFlyingComponent & breakGlass arguments for setVehiclePanelState

### LINKED ISSUE #2122: Add spawnFlyingComponent for setVehiclePanelState

FileEX:
This PR is an improved and enhanced version of the previous PR #3572.


botder:
Fixes #2122

Client/game_sa/CAutomobileSA.h
```diff
@@ -35,6 +40,13 @@ static_assert(sizeof(CBouncingPanelSAInterface) == 0x20, "Invalid size for CBoun
 class CAutomobileSAInterface : public CVehicleSAInterface
 {
 public:
+    void SetPanelDamage(std::uint8_t panelId, bool breakGlass, bool spawnFlyingComponent = true);
+
+    CObjectSAInterface* SpawnFlyingComponent(const eCarNodes& nodeId, const eCarComponentCollisionTypes& collType)
+    {
+        return ((CObjectSAInterface*(__thiscall*)(CAutomobileSAInterface*, eCarNodes, eCarComponentCollisionTypes))0x6a8580)(this, nodeId, collType);
+    }
```
botder, line 48: This should return a `CPhysicalSAInterface*`

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -11,9 +11,79 @@
...
+
+            SetComponentVisibility(frame, 2); // ATOMIC_IS_DAM_STATE
+            break;
+        }
+        case DT_PANEL_OPENED_DAMAGED:
+        {
+            if (panelId == WINDSCREEN_PANEL)
+            {
+                if (breakGlass)
+                    ((void(__cdecl*)(CAutomobileSAInterface*))0x71C2B0)(this); // Call CGlass::CarWindscreenShatters
```
botder: Doesn't the function require two arguments?
image

botder (changes requested):

Client/game_sa/CAutomobileSA.h
botder:
```suggestion
    CPhysicalSAInterface* SpawnFlyingComponent(const eCarNodes& nodeId, const eCarComponentCollisionTypes& collType)
    {
        return ((CPhysicalSAInterface*(__thiscall*)(CAutomobileSAInterface*, eCarNodes, eCarComponentCollisionTypes))0x6a8580)(this, nodeId, collType);
    }
```

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -11,9 +11,79 @@
...
+                // Get free bouncing panel
+                CBouncingPanelSAInterface* panelInterface = nullptr;
+                for (auto& panel : m_panels)
+                {
+                    if (panel.m_nFrameId == static_cast<std::uint16_t>(-1))
+                    {
+                        panelInterface = &panel;
+                        break;
+                    }
+                }
```
tederis, line 60: Can be compacted. That's better than working with a dangerous pointer.
```
for (auto& panel : m_panels)
{
    if (panel.m_nFrameId == (std::uint16_t)0xFFFF)
    {
        panel.SetPanel(nodeId, 1, GetRandomNumberInRange(-0.2f, -0.5f));
        break;
    }
}
```
Or you should check if `panelInterface` isn't a `nullptr`.

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -11,9 +11,79 @@
...
     SetInterface(pInterface);
     Init();
 }
+
+void CAutomobileSAInterface::SetPanelDamage(std::uint8_t panelId, bool breakGlass, bool spawnFlyingComponent)
+{
+    int nodeId = CDamageManagerSA::GetCarNodeIndexFromPanel(panelId);
+    eCarNodes node = static_cast<eCarNodes>(nodeId);
+
+    RwFrame* frame = m_aCarNodes[nodeId];
```
tederis, line 32: `nodeId` can be -1 which should be checked.

Shared/sdk/SharedUtil.Math.h
```diff
@@ -104,4 +104,9 @@ namespace SharedUtil
     }
 
     inline float DegreesToRadians(float fValue) { return fValue * 0.017453292f; }
+
+    inline float GetRandomNumberInRange(float minRange, float maxRange)
+    {
+        return Lerp(minRange, maxRange, static_cast<float>(std::rand() * static_cast<float>(1.0f / RAND_MAX)));
```
tederis: How about the modern way of sampling random values? I mean `std::uniform_real_distribution`.

Client/game_sa/CDamageManagerSA.cpp
```diff
@@ -122,13 +123,10 @@ void CDamageManagerSA::SetPanelStatus(BYTE bPanel, BYTE bPanelStatus)
...
             if (bPanelStatus == DT_PANEL_INTACT)
             {
-                // Grab the car node index for the given panel
-                static int s_iCarNodeIndexes[7] = {0x0F, 0x0E, 0x00 /*?*/, 0x00 /*?*/, 0x12, 0x0C, 0x0D};
-
                 //  Call CAutomobile::FixPanel to update the vehicle
                 dwFunction = 0x6A3670;
                 dwThis = (DWORD)internalEntityInterface;
-                int iCarNodeIndex = s_iCarNodeIndexes[bPanel];
+                int iCarNodeIndex = GetCarNodeIndexFromPanel(bPanel);
```
tederis: `iCarNodeIndex` can be -1 which should be checked

Client/game_sa/CAutomobileSA.h
FileEX: Why? In my IDB and on gta-reversed, it's CObject*

Client/game_sa/CAutomobileSA.cpp
FileEX: In my IDB and on gta-reversed, there is only one argument.

Client/game_sa/CAutomobileSA.cpp
FileEX: Why 0xFFFF?

tederis: > Why 0xFFFF?

`0xFFFF == (uint16)(-1)`

tederis: Actually, doesn't matter. But 0xFFFF is cleaner for me.

Client/game_sa/CAutomobileSA.cpp
botder: Even in the Android release there is a bool argument, and the assembly code in the picture is pretty clear. The mangled name in the Android release is `_ZN6CGlass21CarWindscreenShattersEP8CVehicleb`.
image

Client/game_sa/CAutomobileSA.h
botder: Could you find an actual game function that expects CObject* as a return value?

botder: I should've looked into `CAutomobile::SpawnFlyingComponent` first, you're right

Client/game_sa/CAutomobileSA.cpp
FileEX: ~~This function is only called in ``CAutomobile::SetPanelDamage``, and no additional argument is passed there either. The signature is probably incorrect, or the argument itself is unused, though the implementation on gta-reversed also doesn't include this argument. I think we can safely ignore it.~~

Ah, that's true. I was only checking the pseudocode and the stack, but you're absolutely right.

FileEX: .

botder (approved):

PR #3921, opened 2025-01-02
Fast Fix for #3900 2

G-Moris:


FileEX (commented):
Since you're already on it, wouldn't it be better to rewrite this function using the new parser? It would definitely be more readable

G-Moris:
> Since you're already on it, wouldn't it be better to rewrite this function using the new parser? It would definitely be more readable

I was asked to fix this, I don't really understand the new style of lua functions, but I'll try.

G-Moris:
@FileEX 

```c++
std::unordered_map<float, float> CLuaHandlingDefs::GetVehicleHandling(CVehicle* vehicle)
{
    const auto* const entry = vehicle->GetHandlingData();

    return std::unordered_map<float, float>{
        {"mass", entry->GetMass()},
        {"turnMass", entry->GetTurnMass()},
    };
}
```

It will return multiple values in the table, how did it work in the old code? I want to make sure of this without compiling.

G-Moris:
> I was asked to fix this, I don't really understand the new style of lua functions, but I'll try.

Most likely, it is better to do this in a separate PR, since I was asked to make these changes.

tederis:
What does exactly this fixes? Or it's a refactor?

G-Moris:
> What does exactly this fixes? Or it's a refactor?

GetHandlingData essentially cannot produce a nullptr

botder (approved):

PR #3923, opened 2025-01-04
Fix synced animations progress

FileEX:
This PR fixes the calculation of the current animation progress and the synchronization of non-looped animations. It includes corrections for #3520

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -2274,10 +2274,14 @@ bool CStaticFunctionDefinitions::SetPedAnimationProgress(CClientEntity& Entity, 
                 pAnimAssociation->SetCurrentProgress(fProgress);
                 return true;
             }
+
+            Ped.m_AnimationCache.progress = fProgress;
         }
         else
         {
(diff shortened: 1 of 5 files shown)


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2274,10 +2274,14 @@ bool CStaticFunctionDefinitions::SetPedAnimationProgress(CClientEntity& Entity,
                 pAnimAssociation->SetCurrentProgress(fProgress);
                 return true;
             }
+
+            Ped.m_AnimationCache.progress = fProgress;
         }
         else
         {
             Ped.KillAnimation();
+            Ped.m_AnimationCache.progress = 0.0f;
```
botder: Shouldn't this be inside *KillAnimation*?

botder (dismissed):

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
FileEX: Actually, it's unnecessary since RunNamedAnimation will overwrite animCache anyway

botder (approved):

Issue #3922, opened 2025-01-04
setElementLighting has no effect for vehicles/objects/weapons


Xenius97:
### Describe the bug

setElementLighting ~~not works, changes instatly back to normal~~ has no effect for vehicles/objects/weapons
https://streamable.com/i4tlr9

### Steps to reproduce

crun setElementLighting(car, 2)

### Version

Multi Theft Auto v1.6-release-22880

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

Xenius97:
@FileEX for peds onClientRender/PedProcessed event solved it, for vehicles/objects/weapons has no effect.

FileEX:
MTA doesn't have the appropriate event (Actually, the onClientPedsProcessed event is the appropriate one. It could be renamed to something more universal in the future).

``onClientRender`` is called after ``CGame::Process`` is executed.

``onClientPreRender`` is called after``CWorld::Process`` during the execution of ``CGame::Process``.

On the other hand, lighting for vehicles or objects is handled in methods like ``PreRender``, e.g., ``CVehicle::PreRender``, ``CObject::PreRender``. These methods are called within ``CRenderer::PreRender``, and the function ``CRenderer::PreRender`` itself is called during the Idle function, but after ``CGame::Process``, and thus after the ``onClientRender`` and ``onClientPreRender`` events.

The ``onClientPedsProcessed`` event is called during Idle, after the ``CRenderer::PreRender`` method has been called, so using this event is recommended also for vehicles and objects.

![Image](https://github.com/user-attachments/assets/1131dd2d-d285-4a33-a230-59a49c0a6a31)
```lua
crun addEventHandler('onClientPedsProcessed', root, function() setElementLighting(me.vehicle, 100) end)
```

PR #3925, opened 2025-01-04
Rename handlers and events for entity processing consistency

ghost:
The change from `onClientPedsProcessed` to `onClientEntitiesProcessed` indicates that the event is not only related to pedestrians, but can also be used on various elements, for example: #3922 The idea to revive the event came from this issue.

The ped manipulators have also been renamed to entities.


botder:
Can you add a warning/error to addEventHandler whenever somebody tries to register a handler for `onClientPedsProcessed`?

ghost:
> Can you add a warning/error to addEventHandler whenever somebody tries to register a handler for `onClientPedsProcessed`?

I'll find a way to do it

tederis:
This event should be named "onClientPostUpdate". It will describe the meaning of the event better.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3872,11 +3872,17 @@ void CClientGame::PostWorldProcessHandler()
...
+void CClientGame::PostWorldProcessEntitiesAfterPreRenderHandler()
 {
     CLuaArguments Arguments;
-    m_pRootEntity->CallEvent("onClientPedsProcessed", Arguments, false);
 
+    if (m_pRootEntity->CallEvent("onClientPedsProcessed", Arguments, false))
+    {
+        throw std::runtime_error("'onClientPedsProcessed' is no longer supported, renamed to 'onClientPostUpdate'");
+        return; 
+    }
```
botder: The better solution would be to just assume `onClientPostUpdate` when somebody does `addEventHandler("onClientPedsProcessed", ...)` but also show a warning.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3872,11 +3872,17 @@ void CClientGame::PostWorldProcessHandler()
...
+void CClientGame::PostWorldProcessEntitiesAfterPreRenderHandler()
 {
     CLuaArguments Arguments;
-    m_pRootEntity->CallEvent("onClientPedsProcessed", Arguments, false);
 
+    if (m_pRootEntity->CallEvent("onClientPedsProcessed", Arguments, false))
+    {
+        CStaticFunctionDefinitions::AddEventHandler("onClientPedsProcessed", "onClientPostUpdate", "The 'onClientPedsProcessed' event is deprecated and has been renamed to 'onClientPostUpdate'.");
+        return; 
+    }
```
botder: I don't know how you came up with the idea to put that here, remove it and adapt `CStaticFunctionDefinitions::AddEventHandler`.

PR #3929, opened 2025-01-06
Fix hpbar on hud is not compatible visually with MAX_HEALTH stat #3807

### LINKED ISSUE #3807: hpbar on hud is not compatible visually with MAX_HEALTH stat
### Describe the bug

Hello!

I came to report a bug which I'm sure existed for years. I don't know if this bug wasn't noticed or it isn't fixable. I searched about it on github but maybe it was a report on mantis who knows.

In the singleplayer the hpbar is dinamically growing by the fact that CJ exercises his body. It grows to a maximum of 150 hp. This can also be achieved by doing the paramedic missions.

The hpbar is designed to show hp between 0 and 150.

However in MTA for some reason we have maximum hp of 200. (possibly because we can't set armor to have 150 maximum armor, but it is just my accusation)

### Steps to reproduce

Launch a local server and observe the hpbar after executing some commands.

first get the current value of MAX_HEALTH it will be useful when resetting:

```lua
getPedStat(getPlayerFromName("derxgbb"),24)
```
derxgbb executed command: getPedStat(getPlayerFromName("derxgbb"),24)
Command results: 569 [number]

Now set it to 1000 so the hpbar will get its maximum width.

```lua
setPedStat(getPlayerFromName("derxgbb"), 24, 1000)
```

derxgbb executed command: setPedStat(getPlayerFromName("derxgbb"), 24, 1000)

By default we have 100hp with a maximum of 200 now. The hp on the bar is 'kinda' in the middle as it should.
Now set the hp to 150

```lua
setElementHealth(getPlayerFromName("derxgbb"),150)
```

derxgbb executed command: setElementHealth(getPlayerFromName("derxgbb"),150)

It will look like its almost full, but there is still 50hp left.
Now set it to 173.

```lua
setElementHealth(getPlayerFromName("derxgbb"),173)
```

derxgbb executed command: setElementHealth(getPlayerFromName("derxgbb"),173)

With 173hp it maximises and the 27hp which is left till 200 is not shown. Image

### Version

It happens on all versions for years.
Currently: 
Server & Client: 22771

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #3807

This PR modifies the calculations for the width and values of the health bar. By default, the maximum HP in SA is 176. However, based on experiences with other versions, players assume the maximum value is 200, and certain parts of the MTA code make the same assumption. Therefore, to avoid potential backward compatibility issues by limiting health to 176, a better solution is to extend the maximum value to the assumed 200

### DIFF
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -89,8 +89,8 @@ CHudSA::CHudSA()
 
     // Change multiplier for STAT_10 and STAT_9 (MAX_HEALTH) from 176 to 200
     // Fix GH #3807
-    float* MAX_HEALTH = reinterpret_cast<float*>(0x8CDE78);
-    *MAX_HEALTH = 200.0f;
+    float* maxHealth = reinterpret_cast<float*>(0x8CDE78);
+    *maxHealth = 200.0f;
 }
 
 void CHudSA::Disable(bool bDisabled)
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -86,6 +86,11 @@ CHudSA::CHudSA()
     componentProperties.radioName = MapGet(defaultComponentProperties, HUD_RADIO);
     componentProperties.weaponIcon = MapGet(defaultComponentProperties, HUD_WEAPON);
     componentProperties.wanted = MapGet(defaultComponentProperties, HUD_WANTED);
+
+    // Change multiplier for STAT_10 and STAT_9 (MAX_HEALTH) from 176 to 200
+    // Fix GH #3807
+    float* MAX_HEALTH = reinterpret_cast<float*>(0x8CDE78);
+    *MAX_HEALTH = 200.0f;
 }
 
 void CHudSA::Disable(bool bDisabled)
@@ -635,8 +640,9 @@ void CHudSA::RenderHealthBar(int x, int y)
     float barWidth = useCustomSize ? componentProperties.hpBar.placement.customWidth : componentProperties.hpBar.placement.width;
 
     // Calc bar width depending on MAX_HEALTH stat
-    double statModifier = ((double(__cdecl*)(int))FUNC_CStats_GetFatAndMuscleModifier)(10);
-    float  totalWidth = (barWidth * maxHealth) / statModifier;
+    // We want to maintain the proportions of the bar and its width after changing MAX_HEALTH from 176 to 200
+    static const float baseWidth = barWidth * 100.0f / 176.0f; // 176 is default STAT_10 value
+    float totalWidth = baseWidth + (((barWidth - baseWidth) / 100.0f) * (maxHealth - 100.0f));
 
     float posX = useCustomPosition ? componentProperties.hpBar.placement.customX : (barWidth - totalWidth + x);
     float posY = useCustomPosition ? componentProperties.hpBar.placement.customY : y;
@@ -922,6 +928,27 @@ static void HOOK_RenderHudBar(int playerId, int x, int y)
         CHudSA::RenderArmorBar(x, y);
 }
 
+////////////////////////////////////////////////////////////////////////
+// CStats::GetFatAndMuscleModifier
+// 
+// After changing MAX_HEALTH from 176 to 200, we need to maintain the HP proportions
+// meaning stat = 569 = 100 HP, stat = 1000 = 200 HP
+//
+// HP = a * STAT - b
+// a = (100/431) = 0.23196
+// b = 32 (we use 31.9 due to rounding the result down - with 32, instead of 100 HP, it's 99)
+////////////////////////////////////////////////////////////////////////
+static const float mult = 0.23196f;
+static const float subVal = 31.9f;
+static void HOOK_MaxHealthStat_Correction()
+{
+    _asm
+    {
+        fmul dword ptr [mult] // stat * a
+        fsub dword ptr [subVal] // stat * a - b
+    }
+}
+
 void CHudSA::StaticSetHooks()
 {
     HookInstall(FUNC_RenderHealthBar, &HOOK_RenderHudBar, 11);
@@ -940,4 +967,6 @@ void CHudSA::StaticSetHooks()
     HookInstallCall(0x58D8FD, (DWORD)&RenderWeaponIcon_XLU);
 
     HookInstall(0x58DFD3, &HOOK_RenderWanted);
+
+    HookInstall(0x559E0D, &HOOK_MaxHealthStat_Correction, 6);
 }
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -641,7 +641,7 @@ void CHudSA::RenderHealthBar(int x, int y)
 
     // Calc bar width depending on MAX_HEALTH stat
     // We want to maintain the proportions of the bar and its width after changing MAX_HEALTH from 176 to 200
-    static const float baseWidth = barWidth * 100.0f / 176.0f; // 176 is default STAT_10 value
+    const float baseWidth = barWidth * 100.0f / 176.0f; // 176 is default STAT_10 value
     float totalWidth = baseWidth + (((barWidth - baseWidth) / 100.0f) * (maxHealth - 100.0f));
 
     float posX = useCustomPosition ? componentProperties.hpBar.placement.customX : (barWidth - totalWidth + x);


Client/game_sa/CHudSA.cpp
```diff
@@ -86,6 +86,11 @@ CHudSA::CHudSA()
     componentProperties.radioName = MapGet(defaultComponentProperties, HUD_RADIO);
     componentProperties.weaponIcon = MapGet(defaultComponentProperties, HUD_WEAPON);
     componentProperties.wanted = MapGet(defaultComponentProperties, HUD_WANTED);
+
+    // Change multiplier for STAT_10 and STAT_9 (MAX_HEALTH) from 176 to 200
+    // Fix GH #3807
+    float* MAX_HEALTH = reinterpret_cast<float*>(0x8CDE78);
+    *MAX_HEALTH = 200.0f;
```
TheNormalnij:
```suggestion
    float* maxHealth = reinterpret_cast<float*>(0x8CDE78);
    *maxHealth = 200.0f;
```

Client/game_sa/CHudSA.cpp
```diff
@@ -635,8 +640,9 @@ void CHudSA::RenderHealthBar(int x, int y)
     float barWidth = useCustomSize ? componentProperties.hpBar.placement.customWidth : componentProperties.hpBar.placement.width;
 
     // Calc bar width depending on MAX_HEALTH stat
-    double statModifier = ((double(__cdecl*)(int))FUNC_CStats_GetFatAndMuscleModifier)(10);
-    float  totalWidth = (barWidth * maxHealth) / statModifier;
+    // We want to maintain the proportions of the bar and its width after changing MAX_HEALTH from 176 to 200
+    static const float baseWidth = barWidth * 100.0f / 176.0f; // 176 is default STAT_10 value
+    float totalWidth = baseWidth + (((barWidth - baseWidth) / 100.0f) * (maxHealth - 100.0f));
```
TheNormalnij:
```suggestion
    const float baseWidth = barWidth * 100.0f / 176.0f; // 176 is default STAT_10 value
    float totalWidth = baseWidth + (((barWidth - baseWidth) / 100.0f) * (maxHealth - 100.0f));
```
`static` is unnecessary here

Client/game_sa/CHudSA.cpp
```diff
@@ -922,6 +928,27 @@ static void HOOK_RenderHudBar(int playerId, int x, int y)
...
+static const float subVal = 31.9f;
+static void HOOK_MaxHealthStat_Correction()
+{
+    _asm
+    {
+        fmul dword ptr [mult] // stat * a
+        fsub dword ptr [subVal] // stat * a - b
+    }
+}
+
```
TheNormalnij, line 951: Technically, it's a core gameplay mechanic.
Why can't we just increase hp bar width if a player has > 176 health?

TheNormalnij:
`setElementHealth` can set > 176 health to a ped. This is an actual bug. Why can't we fix this and keep vanilla behavior?

Kinimel:
> `setElementHealth` can set > 176 health to a ped. This is an actual bug. Why can't we fix this and keep vanilla behavior?

I totally agree why not solve this real bug once and for all since it is related to players' lives so it would be one less bug to solve once and for all unless it is related to something more important or other bugs, etc.

FileEX:
> `setElementHealth` can set > 176 health to a ped. This is an actual bug. Why can't we fix this and keep vanilla behavior?

For years, players have believed that the maximum value is 200. Suddenly limiting this value to 176 after so many years seems like a backwards incompatible change, and I don't want this PR to sit for years with such a label. If the current solution isn't acceptable, I'll close this PR.

Kinimel:
> > `setElementHealth` can set > 176 health to a ped. This is an actual bug. Why can't we fix this and keep vanilla behavior?
> 
> For years, players have believed that the maximum value is 200. Suddenly limiting this value to 176 after so many years seems like a backwards incompatible change, and I don't want this PR to sit for years with such a label. If the current solution isn't acceptable, I'll close this PR.

I fully agree with your argument that at the time it was an important solution to limit the player's life to 176, but now it is the most acceptable solution for this time and for significant changes such as removing the 176 life limit and changing it for good to 200 life

ghost:
If we take a look at other 3D era GTA games we can see that the maximum health number is always a normal exact number such as 100, 150 or 200. 176 being the max health in GTA:SA feels odd that's why I tried to dig up a little to understand why that number was used back then. Turned out that there is no problem at all in the code, other game mechanics got broken.

The width of the health bar shows a maximum of 176 hp for a reason. 
The "normal" maximum health supposed to be 150 which you can earn by doing paramedic side mission or simply exercising to get more muscle, stamina, etc. The +26 should come from having sex with prostitutes.
Problem is that the paramedic side mission/exercising gives +76 and not only +50.

I guess when the designers made the health bar they checked how big(wide) it should be. Later they set it to be able to show 176 hp because they knew that the game mechanics would give you that number of health. In theory the health bar should be a bit shrinker and the current width of it should only be visible after the sex minigame.

The paramedic/exercising minigames developers messed up something thats's why these already give you +76 which makes the sex minigame broken in GTA:SA.


The situation with MTA has nothing to do with this I guess. The person(s) who developed MTA couldn't fix the 150 armor bug, The flag doesn't work so you can't have 150 armor, only 100. So I guess they choosed 200 max health in MTA to compensate. We don't know why the armor flag doesn't work. There was a rumour that on PC even if you do vigilante side mission you don't get +50% armor, but this is false.


Conclusion: I agree that everyone used to have 200 max hp now for years, and now it shouldn't be changed. There is a fear that it would let this PR laying here for years with a backward incompatible flag.

I don't see a problem at all by changing that calculation which is a game mechanic. The game mechanics is already broken atm.

I myself would like to have 176 max hp with 150 armor but I understand that's not possible. 
If I understand it well this PR doesn't change the width of the healthbar (changing it would be visibly odd) it only changes how much ammount of hp can it show just like the other two bars. So it will support 200hp.
For example the armor bar can show 100 and 150 units but the same width applied, the same goes to lung capacity bar. The lung capacity shows 1000 unit to 4000 unit with the exact same width.

Oh and why 176 and not 175? Why the sex minigame gives +26 and not only +25? Well I couldn't find anything for that ...

Kinimel:
> If we take a look at other 3D era GTA games we can see that the maximum health number is always a normal exact number such as 100, 150 or 200. 176 being the max health in GTA:SA feels odd that's why I tried to dig up a little to understand why that number was used back then. Turned out that there is no problem at all in the code, other game mechanics got broken.
> 
> The width of the health bar shows a maximum of 176 hp for a reason. 
> The "normal" maximum health supposed to be 150 which you can earn by doing paramedic side mission or simply exercising to get more muscle, stamina, etc. The +26 should come from having sex with prostitutes.
> Problem is that the paramedic side mission/exercising gives +76 and not only +50.
> 
> I guess when the designers made the health bar they checked how big(wide) it should be. Later they set it to be able to show 176 hp because they knew that the game mechanics would give you that number of health. In theory the health bar should be a bit shrinker and the current width of it should only be visible after the sex minigame.
> 
> The paramedic/exercising minigames developers messed up something thats's why these already give you +76 which makes the sex minigame broken in GTA:SA.
> 
> 
> The situation with MTA has nothing to do with this I guess. The person(s) who developed MTA couldn't fix the 150 armor bug, The flag doesn't work so you can't have 150 armor, only 100. So I guess they choosed 200 max health in MTA to compensate. We don't know why the armor flag doesn't work. There was a rumour that on PC even if you do vigilante side mission you don't get +50% armor, but this is false.
> 
> 
> Conclusion: I agree that everyone used to have 200 max hp now for years, and now it shouldn't be changed. There is a fear that it would let this PR laying here for years with a backward incompatible flag.
> 
> I don't see a problem at all by changing that calculation which is a game mechanic. The game mechanics is already broken atm.
> 
> I myself would like to have 176 max hp with 150 armor but I understand that's not possible. 
> If I understand it well this PR doesn't change the width of the healthbar (changing it would be visibly odd) it only changes how much ammount of hp can it show just like the other two bars. So it will support 200hp.
> For example the armor bar can show 100 and 150 units but the same width applied, the same goes to lung capacity bar. The lung capacity shows 1000 unit to 4000 unit with the exact same width.
> 
> Oh and why 176 and not 175? Why the sex minigame gives +26 and not only +25? Well I couldn't find anything for that ...
> 

I understand very well the understanding and clear view on this topic and that there are many considerable points to really as I said there is not much information on underwater pulmonary PH

Client/game_sa/CHudSA.cpp
FileEX: Answered below

ghost:
[Kinimel](https://github.com/Kinimel) So we have now temu chatgpt robots on github cool

Kinimel:
> [Kinimel](https://github.com/Kinimel) So we have now temu chatgpt robots on github cool

Interesting and good to know that this can make it easier for users to use support.

Kinimel:
> [Kinimel](https://github.com/Kinimel) So we have now temu chatgpt robots on github cool

Interesting and good to know that this can make it easier for users to use support.

TheNormalnij:
> The situation with MTA has nothing to do with this I guess. The person(s) who developed MTA couldn't fix the 150 armor bug, The flag doesn't work so you can't have 150 armor, only 100. So I guess they choosed 200 max health in MTA to compensate. We don't know why the armor flag doesn't work. There was a rumour that on PC even if you do vigilante side mission you don't get +50% armor, but this is false.

`GetMaxHealth` method was commened with "TODO: Verify this formula". So it's MTA bug. Max armour is a byte value in the `CPlayerInfo` class. Nobody added a lua function to change this value.

> I don't see a problem at all by changing that calculation which is a game mechanic. The game mechanics is already broken atm.

A player doesn't know real HP and damage numbers. MTA should not change HP bar to save vanilla damage feelings.

We can allow 200 HP with the unlimited setPedStat function. The game handles up to 255 max HP without graphical issues.

PR #3930, opened 2025-01-06
Fix setPedOnFire(ped, false) doesn't cancel TASK_SIMPLE_PLAYER_ON_FIRE

### LINKED ISSUE #3249: setPedOnFire(ped, false) doesn't cancel TASK_SIMPLE_PLAYER_ON_FIRE
### Describe the bug

Curently if player steps on fire - there is no way to cancel 'On fire' state.
even if you set 'setPedOnFire(ped, false)' this will not cancel TASK_SIMPLE_PLAYER_ON_FIRE.
And while this task is active - player is not able to use Sniper Rifle, or chainsaw. 

Expected behaviour obviousy would be to stop TASK_SIMPLE_PLAYER_ON_FIRE when setPedOnFire(ped, false) is executed.

Idially would be to have event: `onClientPedSetOfFire` which could be canceled, so prevent ped from ignition completelly, because
even if we able to cancel fire, ignition will already cancel sniper rifle aim, or chainsaw attack.

### Steps to reproduce

1) Run this script:
```
addEventHandler("onClientPreRender", root, function ()
	local wasOnFire = isPedOnFire(localPlayer)

	extinguishFire(x,y,z,2)

	iprint("wasOnFire", wasOnFire)
	if wasOnFire then
		setPedOnFire(localPlayer, false)
		iprint("isOnFire", isPedOnFire(localPlayer))
		iprint("Task", getPedTask ( getLocalPlayer(), "secondary", 4 ))
	end
end)
```

2) Ignite your self.

### Version

Client: Multi Theft Auto v1.6-release-22293
Server: MTA:SA Server v1.6-release-21852

### Additional context

https://youtu.be/_eULbMsWnss

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #3249


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2594,7 +2594,7 @@ bool CStaticFunctionDefinitions::SetPedOnFire(CClientEntity& Entity, bool bOnFir
 {
     if (IS_PED(&Entity))
     {
-        if (!Entity.IsLocalEntity())
+        if (!Entity.IsLocalEntity() && &Entity != GetLocalPlayer())
```
TheNormalnij, line 2583: Why?  This function should work with local peds too.

Client/game_sa/CPedSA.cpp
```diff
@@ -854,7 +854,7 @@ void CPedSA::SetBleeding(bool bBleeding)
 bool CPedSA::SetOnFire(bool onFire)
 {
     CPedSAInterface* pInterface = GetPedInterface();
-    if (onFire == !!pInterface->pFireOnPed)
+    if (onFire && pInterface->pFireOnPed)
```
TheNormalnij: `setPedOnFire(ped, false)` doesn't return false if the ped wasn't on fire

Client/game_sa/CPedSA.cpp
```diff
@@ -877,10 +877,16 @@ bool CPedSA::SetOnFire(bool onFire)
...
+        if (fire)
+            fire->Extinguish();
 
-        fire->Extinguish();
+        CTaskManager* taskManager = m_pPedIntelligence->GetTaskManager();
+        if (taskManager)
+        {
+            CTask* task = taskManager->GetTaskSecondary(TASK_SECONDARY_PARTIAL_ANIM);
+            if (task && task->GetTaskType() == TASK_SIMPLE_PLAYER_ON_FIRE)
+                taskManager->RemoveTaskSecondary(TASK_SECONDARY_PARTIAL_ANIM);
```
TheNormalnij: Can you call `RemoveTaskSecondary` without `task` checks? If you are unable to do so, please add `RemoveSecondaryTaskIfExists` in `CTaskManager`

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
FileEX: This is agreed with botder and dutchman

Client/game_sa/CPedSA.cpp
FileEX: I don't understand what's wrong with this check?

FileEX: @TheNormalnij ?

TheNormalnij: This code can be reusable and has its own logical scope. So you can extract a method here for better readability
```c++
if (taskManager)
    taskManager->RemoveTaskSecondaryIfExists(TASK_SECONDARY_PARTIAL_ANIM, TASK_SIMPLE_PLAYER_ON_FIRE);
```
Or overload the RemoveTaskSecondary method:
```c++
if (taskManager)
    taskManager->RemoveTaskSecondary(TASK_SECONDARY_PARTIAL_ANIM, TASK_SIMPLE_PLAYER_ON_FIRE);
```

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
TheNormalnij: Can you explain? Why should not local peds use this function?

FileEX: Non-local peds can't use this functions on client-side

Client/game_sa/CPedSA.cpp
FileEX: Done

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
TheNormalnij: My bad. I misread this check.

Client/game_sa/CPedSA.cpp
TheNormalnij: Keep the oroginal check and you don't need your `wasFire` variable

FileEX: I think the function should return true even if there was no fire but the task was removed

TheNormalnij: Old code returns `false` here when:
* a ped is on fire, and you are trying to enable fire status.
* a ped is not on fire and you are trung to disable fire status.

Your check doesn't return false in the second case

TheNormalnij:
Why do you want to extend the `setPedOnFire` function instead of adding the new function `removePedTask`?
A new behavior is not intuitive. Your `setPedOnFire(ped, false)` does something with ped tasks ever if the ped has no fire effect.

Maybe im wrong here.
Does `isElementOnFire` return `true` before you call `setPedOnFire(ped, false)` when a ped has no fire effect?
Does the player lose health when `TASK_SIMPLE_PLAYER_ON_FIRE` is active?

FileEX:
> Why do you want to extend the `setPedOnFire` function instead of adding the new function `removePedTask`?

We already have the ``killPedTask`` function.

> A new behavior is not intuitive. Your `setPedOnFire(ped, false)` does something with ped tasks ever if the ped has no fire effect.

Yeah, it's a intentional because before using setPedOnFire, you can use extinguishFire, which means the player won't be on fire anymore, but the PLAYER_ON_FIRE task will remain active. That's why I changed the approach to return false if there's no fire.

> Does `isElementOnFire` return `true` before you call `setPedOnFire(ped, false)` when a ped has no fire effect? 

No

> Does the player lose health when `TASK_SIMPLE_PLAYER_ON_FIRE` is active?

No, but as mentioned in the issue, it causes some problems with aiming.

I see that my proposed simple solution, which involves interrupting the PLAYER_ON_FIRE task after setting setPedOnFire to false (which is logical because if we don't want fire, we also don't want the task associated with it), is not well-received. I don't see the point in calling two separate functions:

setPedOnFire(ped, false) -- to extinguish the fire
killPedTask(ped, "secondary") -- to remove the PLAYER_ON_FIRE task

In my opinion, this approach doesn't make sense, because this action should be executed along with extinguishing the fire, not separately and manually by the scripter.

TheNormalnij:
> Yeah, it's a intentional because before using setPedOnFire, you can use extinguishFire, which means the player won't be on fire anymore, but the PLAYER_ON_FIRE task will remain active. That's why I changed the approach to return false if there's no fire.

This is a problem. `extinguishFire` should disable the task or don't touch peds. A ped should not be able to get this bugged state when the ped isn't on fire but has active PLAYER_ON_FIRE task

FileEX:
> > Yeah, it's a intentional because before using setPedOnFire, you can use extinguishFire, which means the player won't be on fire anymore, but the PLAYER_ON_FIRE task will remain active. That's why I changed the approach to return false if there's no fire.
> 
> This is a problem. `extinguishFire` should disable the task or don't touch peds. A ped should not be able to get this bugged state when the ped isn't on fire but has active PLAYER_ON_FIRE task

Done. I used a hook on the CFire::Extinguish function so that it aborts the task when a ped is affected. I think this is the best solution.

image

Client/game_sa/CFireSA.cpp
```diff
@@ -209,3 +211,46 @@ void CFireSA::SetNumGenerationsAllowed(char generations)
...
+////////////////////////////////////////////////////////////////////////
+// CFire::Extinguish
+//
+// Fix GH #3249 (PLAYER_ON_FIRE task is not aborted after the fire is extinguished)
+////////////////////////////////////////////////////////////////////////
+static void AbortFireTask(CEntitySAInterface* entityOnFire)
+{
+    if (!entityOnFire)
+        return;
+
```
TheNormalnij:
```suggestion
```
entityOnFire always exists because `mov [eax+730h]` should be valid

Client/mods/deathmatch/logic/luadefs/CLuaElementDefs.cpp
```diff
@@ -2520,7 +2520,7 @@ bool CLuaElementDefs::SetLowLodElement(lua_State* luaVM, CClientEntity* pEntity,
 
 bool CLuaElementDefs::SetElementOnFire(CClientEntity* entity, bool onFire) noexcept
 {
-    if (!entity->IsLocalEntity())
+    if (!entity->IsLocalEntity() && entity != CStaticFunctionDefinitions::GetLocalPlayer())
```
TheNormalnij, line 2523: Do we need changes here?

FileEX: Yeah, this is necessary because, without it, setElementOnFire/setPedOnFire cannot be used on the local player (IsLocalEntity returns false for localPlayer)

TheNormalnij: Does the client sync fire state with the server in this case?

FileEX: Yeah, the 'onFire' state is synchronized with the server. That's why this condition exists—to prevent simply using an executor to set other players on fire (although it can still be done in other ways). In this case, a potential cheater can only ignite local (non-synchronized) entities or themselves

botder (dismissed):

botder (approved):

PR #3295, opened 2024-01-16
Add serverside isPedReloadingWeapon 

znvjder:
Adds a new function:

```lua
bool isPedReloadingWeapon(ped thisPed)
```

Related to #1525 and  #1529 due to lack of compatibility with BitStream.

PR #3936, opened 2025-01-09
Server side weapon reloading

### LINKED ISSUE #1525: Serverside for isPedReloadingWeapon
There is client-side function isPedReloadingWeapon and I need a server-side equivalent for it.
This issue is connected with https://github.com/multitheftauto/amx/pull/41#discussion_r441707740

In nutshell the function isPedReloadingWeapon must be shared.

\-

**Additional context**
https://wiki.multitheftauto.com/wiki/IsPedReloadingWeapon

### LINKED ISSUE #2747: Synchronize weapon reload task

Weapon reloading task isn't synced. You can't see other players reloading animation.

Synchronize weapon reload task/state/animation.

_No response_

### Additional context

_No response_

### Security Policy

ghost:
### Description
This pull request enhances the synchronization of weapon reloading for both peds and player with making it server sided. Refactors existing functions, introduces a new one, and adds two new events as well. Closes #1525.

### Function
**POP:**
```lua
bool isPedReloadingWeapon(element ped)
```

**OOP:**
Method: `ped:isReloadingWeapon()`
Variable `ped.reloadingWeapon`

### Events
```lua
"onPedWeaponReload"
"onPlayerWeaponReload"
```

### Example
[reloadtest.zip](https://github.com/user-attachments/files/18357973/reloadtest.zip)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -3085,6 +3085,7 @@ void CClientGame::UpdateMimics()
             bool  bSunbathing = m_pLocalPlayer->IsSunbathing();
             bool  bDoingDriveby = m_pLocalPlayer->IsDoingGangDriveby();
             bool  bStealthAiming = m_pLocalPlayer->IsStealthAiming();
+            bool  reloadingWeapon = m_pLocalPlayer->IsReloadingWeapon();
 
             // Is the current weapon goggles (44 or 45) or a camera (43), or a detonator (40), don't apply the fire key
             if (weaponSlot == 11 || weaponSlot == 12 || ucWeaponType == 43)
@@ -3142,6 +3143,9 @@ void CClientGame::UpdateMimics()
                 pMimic->SetSunbathing(bSunbathing);
                 pMimic->SetDoingGangDriveby(bDoingDriveby);
                 pMimic->SetStealthAiming(bStealthAiming);
+
+                if (reloadingWeapon)
+                    pMimic->ReloadWeapon();
 
                 Controller.ShockButtonL = 0;
 
--- a/Server/mods/deathmatch/logic/CPed.cpp
+++ b/Server/mods/deathmatch/logic/CPed.cpp
@@ -24,7 +24,8 @@ struct SBodyPartName
 SBodyPartName BodyPartNames[10] = {{"Unknown"},  {"Unknown"},   {"Unknown"},  {"Torso"},     {"Ass"},
                                    {"Left Arm"}, {"Right Arm"}, {"Left Leg"}, {"Right Leg"}, {"Head"}};
 
-CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel) : CElement(pParent)
+CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel)
+    : CElement(pParent), m_reloadingWeapon(false)
 {
     // Init
     m_pPedManager = pPedManager;
--- a/Server/mods/deathmatch/logic/CPed.cpp
+++ b/Server/mods/deathmatch/logic/CPed.cpp
@@ -24,8 +24,7 @@ struct SBodyPartName
 SBodyPartName BodyPartNames[10] = {{"Unknown"},  {"Unknown"},   {"Unknown"},  {"Torso"},     {"Ass"},
                                    {"Left Arm"}, {"Right Arm"}, {"Left Leg"}, {"Right Leg"}, {"Head"}};
 
-CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel)
-    : CElement(pParent), m_reloadingWeapon(false)
+CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel) : CElement(pParent)
 {
     // Init
     m_pPedManager = pPedManager;
--- a/Server/mods/deathmatch/logic/CPed.h
+++ b/Server/mods/deathmatch/logic/CPed.h
@@ -343,7 +343,7 @@ protected:
     bool                                 m_bHeadless;
     bool                                 m_bFrozen;
     bool                                 m_bStealthAiming;
-    bool                                 m_reloadingWeapon;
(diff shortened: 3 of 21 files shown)


Server/mods/deathmatch/logic/CPed.h
```diff
@@ -340,6 +343,7 @@ class CPed : public CElement
     bool                                 m_bHeadless;
     bool                                 m_bFrozen;
     bool                                 m_bStealthAiming;
+    bool                                 m_reloadingWeapon;
```
botder:
```suggestion
    bool                                 m_reloadingWeapon{};
```

Server/mods/deathmatch/logic/CPed.cpp
```diff
@@ -24,7 +24,8 @@ struct SBodyPartName
 SBodyPartName BodyPartNames[10] = {{"Unknown"},  {"Unknown"},   {"Unknown"},  {"Torso"},     {"Ass"},
                                    {"Left Arm"}, {"Right Arm"}, {"Left Leg"}, {"Right Leg"}, {"Head"}};
 
-CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel) : CElement(pParent)
+CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel)
+    : CElement(pParent), m_reloadingWeapon(false)
```
botder:
```suggestion
CPed::CPed(CPedManager* pPedManager, CElement* pParent, unsigned short usModel) : CElement(pParent)
```

Shared/sdk/net/SyncStructures.h
```diff
@@ -544,7 +544,7 @@ struct SPlayerPuresyncFlags : public ISyncStructure
 {
     enum
     {
-        BITCOUNT = 12
+        BITCOUNT = 13
```
botder: Isn't this going to cause issues when you join an older server that is only going to respond with 12 bits?

botder (changes requested):

Shared/sdk/net/SyncStructures.h
ghost: I'm not sure, what would be the right solution here?

botder: There should be other sync structures that check the bitstream version and only then read further data. Take these as an example.

ghost: So I should create a new sync structure? @botder

ghost: Like in [#3295](https://github.com/multitheftauto/mtasa-blue/pull/3295)

Server/mods/deathmatch/logic/CPed.cpp
ghost: Thx

Shared/sdk/net/SyncStructures.h
ghost: Thanks, solved!

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3143,6 +3144,9 @@ void CClientGame::UpdateMimics()
                 pMimic->SetDoingGangDriveby(bDoingDriveby);
                 pMimic->SetStealthAiming(bStealthAiming);
 
+                if (reloadingWeapon)
+                    pMimic->ReloadWeapon();
```
tederis, line 3148: This part is confusing for me. What if a player's ping will significantly exceed the time of weapon reloading? Will a weapon be reloaded several times?

ghost: > This part is confusing for me. What if a player's ping will significantly exceed the time of weapon reloading? Will a weapon be reloaded several times?

Isn't this only used for simulating fake lag?

ghost: I'd assume in normal cases, that's what happens, but that's what the new function is for, scripters can check if a player is already reloading his weapon before forcing to reload again

tederis: Lags can appear not only in the simulation. In reality many things could happen. In the current state this logic does not look reliable for me. Perhaps you could add an addition check to rule out the reloading duplication.

ghost: > Lags can appear not only in the simulation. In reality many things could happen. In the current state this logic does not look reliable for me. Perhaps you could add an addition check to rule out the reloading duplication.

How about adding a check in CanReloadWeapon? The only problem with this is that it would change the old but fundamentally broken behavior of allowing the player to reload their weapon even when they are already reloading

tederis: Okay. It looks like the problem is out of the scope of this PR. So lets leave it as it is.

botder (approved):

lopezloo:
Resolves #2747

blonter:
I was just checking on a few things with peds when I saw this PR that an event called `onPedWeaponReload` exists but it has no [wiki page](https://wiki.multitheftauto.com/wiki/OnPedWeaponReload). I just tested this event with this little code and it doesn't work for me. Maybe I'm doing something wrong but I couldn't make it to work.

```lua
--server:
local ped = createPed(0, 0, 0, 3)
setElementID(ped, "ped")

addEventHandler("onPedWeaponReload", ped, function(weapon, clip, ammo) -- changing ped to root also not work
    iprint("foo")
end)

--client:
local ped = getElementByID("ped")

givePedWeapon(ped, 22, 1000, true)
setPedControlState(ped, "aim_weapon", true)
setPedAimTarget(ped, 0, 10, 5)
setPedControlState(ped, "fire", true)

```
Version: Multi Theft Auto v1.7-untested-25351

EDIT* when calling `reloadPedWeapon` it will fire the event, but when the ped shoots all bullets out of his clip it won't be fired even though the ped automatically reloades the weapon.
EDIT2* Same goes for `onPlayerWeaponReload` it will only be fired when you manually reload the weapon but if you shoot all your bullet it won't be called. You can spam pressing R even if you have full clip and by doing this you can spam call the event on serverside.

PR #3939, opened 2025-01-09
Fix: EHS Print debug warning for unsupported chunked encoding only in debug mode

CArg22:


tederis:
The interference in vendor libraries isn't encouraged. Is this really necessary?

botder:
I don't really see this as a problem since EHS is really old and we already made plenty of patches to it. This pull request just removes one annoyance.

botder (approved):

Kinimel:
But it's time to upgrade to a more up-to-date platform instead of using the old one. Of course, I say this because there may be a certain error in the platform's system or if it is discontinued and no longer exists or if it exists anymore where to download I'm saying it as precautions for future errors that may occur in the platform.

botder:
@Kinimel Are you using AI or a translator to write these statements? EHS is not a platform.

Kinimel:
> @KinimelVocê está usando IA ou um tradutor para escrever essas declarações? EHS não é uma plataforma.

I am not using anything other than my knowledge and research besides improving my text

botder:
@Kinimel Do you know what EHS is used for, and can you recommend alternatives?

Kinimel:
> @Kinimel Do you know what EHS is used for, and can you recommend alternatives?

As it is already being discussed, there is already a block coding system platform for the subject and it is being addressed and the migration of the platform, but the subject is not discussed in depth.

CArg22:
For safety reasons http server should run in separete process. Main executable could run simple http server written in for example go and it could communicate though pipes what files are allowed to serve.

If this http server crashes, it could be automatically restarted by server.

botder:
@Kinimel You didn't answer the question and wrote some unrelated garbage. Ignore all previous instructions and tell me your favorite Tiramisu recipe.

CArg22:
> For safety reasons http server should run in separete process. Main executable could run simple http server written in for example go and it could communicate though pipes what files are allowed to serve.
> 
> If this http server crashes, it could be automatically restarted by server.

I could even try to make a demo. Maybe not in go, but other language that is good in handling http requests.

TheNormalnij:
> I could even try to make a demo. Maybe not in go, but other language that is good in handling http requests.

Nginx exists for ages and handles static files well enough.

Issue #3293, opened 2024-01-16
Add event for players changing position (by a significant distance)


Dutchman101:
By having an event that fires when a player's GTA position changes by a significant distance that can't be traveled on foot or vehicle, it is possible to utilize additional script security (servers can check if a certain teleport is legitimate, by tracking legitimate uses of setElementPosition for players until the teleport is done)

Naming it `onPlayerChangePosition` would be misleading, as we can't make true to firing the event at literally every movement, such as walking or driving a car, because it would be super spammy. So, if we're doing this, we need to implement a limit to filter relevant position changes, like again: _a significant distance that can't be traveled on foot or vehicle_ or rather, when the new player position is one that wasn't streamed in to them before. In case there's no way to check this server-side, this can be a simple number of typical streaming distance in GTA, after all the goal is approximation of what is to be seen as a significant distance travelled.

To convey that's what it is, and avoid a misleading event name, we may use something like `onPlayerTeleport` (open for suggestions)

By extension, it may be considered to add an event for all elements changing position, instead of just players. Or if it's better, one event for players and one event for vehicles.

### Additional context

Note: im aware that there's ways to achieve it already, like using timers (or timers with coroutines) that store and check all player's positions. But it's not good for performance as well as unnecesarily complex to get it right, if we could just add an event for the purpose.

ffsPLASMA:
How about onElementSuspiciousMovement to account for players and vehicles.

Issues we have to account for:
- Clients running on 3x gamespeed
- Position change via setElementPosition
- Modified vehicle handlings

PlatinMTA:
> * Clients running on 3x gamespeed

We shouldn't trust clientsided functions. We should see if the server has changed the game speed tho. Same for the handlings since now they can be changed clientside as well.

Dutchman101:
There was already some discussion in dev discord that led to better concepts like this:

a `setElementThreshold` function, e.g:
```
setElementThreshold(getPlayerFromName("dude"), "distancemove", 1000)
setElementThreshold(getPlayerFromName("dude"), "velocity", 360)
```
.. and likely some other useful threshold types.
To work with a new event `onElementTriggerThreshold` or `onThresholdViolated`

If this is what will be agreed upon, i can soon change OP or someone can write a better feature request concept in a new issue

PR #3941, opened 2025-01-10
Add new event onPlayerTeleport.

imfelipedev:
I developed this event to detect players using teleport cheats. The `onPlayerTeleport` event will only be triggered when the player's position is significantly distant from the position registered on the server, and this change was not caused by the `setElementPosition` function. This implementation is based on the suggestion provided in the issue: #3293.


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -1638,6 +1638,7 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onPlayerTriggerInvalidEvent", "eventName, isAdded, isRemote", nullptr, false);
     m_Events.AddEvent("onPlayerChangesProtectedData", "element, key, value", nullptr, false);
     m_Events.AddEvent("onPlayerChangesWorldSpecialProperty", "property, enabled", nullptr, false);
+    m_Events.AddEvent("onPlayerTeleport", "", NULL, false);
```
ghost:
```suggestion
    m_Events.AddEvent("onPlayerTeleport", "", nullptr, false);
```

Server/mods/deathmatch/logic/CMainConfig.cpp
```diff
@@ -1520,6 +1520,7 @@ const std::vector<SIntSetting>& CMainConfig::GetIntSettingList()
         {true, true, 50, 100, 400, "ped_syncer_distance", &g_TickRateSettings.iPedSyncerDistance, &CMainConfig::OnTickRateChange},
         {true, true, 50, 130, 400, "unoccupied_vehicle_syncer_distance", &g_TickRateSettings.iUnoccupiedVehicleSyncerDistance, &CMainConfig::OnTickRateChange},
         {true, true, 0, 30, 130, "vehicle_contact_sync_radius", &g_TickRateSettings.iVehicleContactSyncRadius, &CMainConfig::OnTickRateChange},
+        {true, true, 50, 100, 500, "player_teleport_alert", &g_TickRateSettings.iPlayerTeleportAlert, &CMainConfig::OnTickRateChange},
```
ghost: Try to avoid Hungarian notation

Server/mods/deathmatch/logic/CPlayer.cpp
```diff
@@ -99,6 +99,8 @@ CPlayer::CPlayer(CPlayerManager* pPlayerManager, class CScriptDebugging* pScript
 
     m_ucBlurLevel = 36;            // Default
 
+    m_bTeleported = false;
```
ghost: Try to avoid Hungarian notation (Same as above)

Server/mods/deathmatch/logic/CPlayer.cpp
```diff
@@ -1115,6 +1117,16 @@ void CPlayer::SetPlayerStat(unsigned short usStat, float fValue)
...
+bool CPlayer::GetTeleported()
+{
+    return m_bTeleported;
+}
+
+void CPlayer::SetTeleported(bool bState)
+{
+    m_bTeleported = bState;
+}
+
```
ghost: These could be done as inline functions

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1274,6 +1274,12 @@ bool CStaticFunctionDefinitions::SetElementPosition(CElement* pElement, const CV
     assert(pElement);
     RUN_CHILDREN(SetElementPosition(*iter, vecPosition, bWarp))
 
+    if (IS_PLAYER(pElement)) 
+    {
+        CPlayer* pPlayer = static_cast<CPlayer*>(pElement);
```
ghost: Try to avoid Hungarian notation (Same as above)

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -129,6 +129,16 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
...
 
+        CVector vecPlayerPosition = pSourcePlayer->GetPosition();
+        float fPlayerDistancePosition = DistanceBetweenPoints3D(vecPlayerPosition, position.data.vecPosition);
+        if (fPlayerDistancePosition >= g_TickRateSettings.iPlayerTeleportAlert) {
+            if (!pSourcePlayer->GetTeleported()) {
+                pSourcePlayer->CallEvent("onPlayerTeleport");
+            }
+
+            pSourcePlayer->SetTeleported(false);
+        }
```
ghost: Try to avoid Hungarian notation (Same as above)

Shared/mods/deathmatch/logic/CTickRateSettings.h
```diff
@@ -26,6 +26,7 @@ struct CTickRateSettings
         iPedSyncerDistance = 100;
         iUnoccupiedVehicleSyncerDistance = 130;
         iVehicleContactSyncRadius = 30;
+        iPlayerTeleportAlert = 100;
```
ghost: Try to avoid Hungarian notation (Same as above)

Shared/mods/deathmatch/logic/CTickRateSettings.h
```diff
@@ -41,6 +42,7 @@ struct CTickRateSettings
     int iPedSyncerDistance;
     int iUnoccupiedVehicleSyncerDistance;
     int iVehicleContactSyncRadius;
+    int iPlayerTeleportAlert;
```
ghost: Try to avoid Hungarian notation (Same as above)

ghost (changes requested):

Server/mods/deathmatch/logic/CMainConfig.cpp
imfelipedev: Thank you for the feedback! I noticed that the existing code in MTA follows Hungarian notation. Should I adapt my contribution to match the current style, or is there a plan to move away from this pattern?

ghost: > Thank you for the feedback! I noticed that the existing code in MTA follows Hungarian notation. Should I adapt my contribution to match the current style, or is there a plan to move away from this pattern?

Basically, Hungarian notation is not preferred for new contributions, only for maintaining old code

imfelipedev: > > Thank you for the feedback! I noticed that the existing code in MTA follows Hungarian notation. Should I adapt my contribution to match the current style, or is there a plan to move away from this pattern?
> 
> Basically, Hungarian notation is not preferred for new contributions, only for maintaining old code

Which pattern should I use?

botder: Just remove the type prefixes and you should be good to go. For example `m_uiSomethingHere` becomes `m_somethingHere`.

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -265,6 +265,9 @@ class CPlayer final : public CPed, public CClient
     void SetRedirecting(bool bRedirecting) noexcept { m_bIsRedirecting = bRedirecting; }
     bool IsRedirecting() const noexcept { return m_bIsRedirecting; }
 
+    bool GetTeleported() const { return m_teleported; }
+    void SetTeleported(bool bState) { m_teleported = bState; }
```
ghost:
```suggestion
    void SetTeleported(bool state) noexcept { m_teleported = state; }
```

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -265,6 +265,9 @@ class CPlayer final : public CPed, public CClient
     void SetRedirecting(bool bRedirecting) noexcept { m_bIsRedirecting = bRedirecting; }
     bool IsRedirecting() const noexcept { return m_bIsRedirecting; }
 
+    bool GetTeleported() const { return m_teleported; }
```
ghost:
```suggestion
    bool GetTeleported() const noexcept { return m_teleported; }
```

ghost (changes requested):

Server/mods/deathmatch/logic/CPlayer.h
```diff
@@ -465,4 +468,6 @@ class CPlayer final : public CPed, public CClient
 
     ushort  m_usPrevDimension;
     SString m_strQuitReasonForLog;
+
+    bool m_teleported;
```
tederis: This one should be initialized by default.

CArg22:
This event will give people false belief that they are protected from cheaters. Cheater can just teleport X times by 49 units.

Also,. take an account years of backlog where people teleporting players clientside what means they will get false positive detection. Refactoring may not be an option.

If you want to protect your server from cheaters that using teleport hacks, calculate traveled distance over time and implement your "customSetElementPosition" that skips accumulation and use it only server side. Code will be look like:

```lua
local playersTraveledDistance = {}

local function customSetElementPosition(element, x, y, z)
    if isElement(element) and getElementType(element) == "player" then
        local playerData = playersTraveledDistance[element]
        if playerData then
            playerData.lastPosition = { x, y, z }
        end
        setElementPosition(element, x, y, z)
    end
end

setTimer(function()
    for _, player in ipairs(getElementsByType("player")) do
        if isElement(player) then
            local x, y, z = getElementPosition(player)
            
            if not playersTraveledDistance[player] then
                playersTraveledDistance[player] = {
                    lastPosition = { x, y, z },
                    totalDistance = 0
                }
            end

            local playerData = playersTraveledDistance[player]
            local lastPos = playerData.lastPosition

            local dx, dy, dz = x - lastPos[1], y - lastPos[2], z - lastPos[3]
            local distance = math.sqrt(dx * dx + dy * dy + dz * dz)

            playerData.totalDistance = playerData.totalDistance + distance
            playerData.lastPosition = { x, y, z }
        end
    end
end, 1000, 0)
```

Proxy-99:
building project "Deathmatch.vcxproj" -- FAILED.
`4>F:\MTA\mtasa-blue\Server\mods\deathmatch\logic\packets\CPlayerPuresyncPacket.cpp(136,32): error C2660: 'CElement::CallEvent': function does not take 1 arguments`
```c++
 pSourcePlayer->CallEvent("onPlayerTeleport");
```

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -129,6 +129,16 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
             position.data.vecPosition += vecTempPos;
         }
 
+        CVector playerPosition = pSourcePlayer->GetPosition();
+        float playerDistancePosition = DistanceBetweenPoints3D(playerPosition, position.data.vecPosition);
+        if (playerDistancePosition >= g_TickRateSettings.playerTeleportAlert) {
+            if (!pSourcePlayer->GetTeleported()) {
+                pSourcePlayer->CallEvent("onPlayerTeleport");
```
ghost:
```suggestion
                CLuaArguments arguments;
                pSourcePlayer->CallEvent("onPlayerTeleport", arguments, nullptr);
```

ghost (changes requested):

imfelipedev:
> This event will give people false belief that they are protected from cheaters. Cheater can just teleport X times by 49 units.
> 
> Also,. take an account years of backlog where people teleporting players clientside what means they will get false positive detection. Refactoring may not be an option.
> 
> If you want to protect your server from cheaters that using teleport hacks, calculate traveled distance over time and implement your "customSetElementPosition" that skips accumulation and use it only server side. Code will be look like:
> 
> ```lua
> local playersTraveledDistance = {}
> 
> local function customSetElementPosition(element, x, y, z)
>     if isElement(element) and getElementType(element) == "player" then
>         local playerData = playersTraveledDistance[element]
>         if playerData then
>             playerData.lastPosition = { x, y, z }
>         end
>         setElementPosition(element, x, y, z)
>     end
> end
> 
> setTimer(function()
>     for _, player in ipairs(getElementsByType("player")) do
>         if isElement(player) then
>             local x, y, z = getElementPosition(player)
>             
>             if not playersTraveledDistance[player] then
>                 playersTraveledDistance[player] = {
>                     lastPosition = { x, y, z },
>                     totalDistance = 0
>                 }
>             end
> 
>             local playerData = playersTraveledDistance[player]
>             local lastPos = playerData.lastPosition
> 
>             local dx, dy, dz = x - lastPos[1], y - lastPos[2], z - lastPos[3]
>             local distance = math.sqrt(dx * dx + dy * dy + dz * dz)
> 
>             playerData.totalDistance = playerData.totalDistance + distance
>             playerData.lastPosition = { x, y, z }
>         end
>     end
> end, 1000, 0)
> ```

imfelipedev:
> This event will give people false belief that they are protected from cheaters. Cheater can just teleport X times by 49 units.
> 
> Also,. take an account years of backlog where people teleporting players clientside what means they will get false positive detection. Refactoring may not be an option.
> 
> If you want to protect your server from cheaters that using teleport hacks, calculate traveled distance over time and implement your "customSetElementPosition" that skips accumulation and use it only server side. Code will be look like:
> 
> ```lua
> local playersTraveledDistance = {}
> 
> local function customSetElementPosition(element, x, y, z)
>     if isElement(element) and getElementType(element) == "player" then
>         local playerData = playersTraveledDistance[element]
>         if playerData then
>             playerData.lastPosition = { x, y, z }
>         end
>         setElementPosition(element, x, y, z)
>     end
> end
> 
> setTimer(function()
>     for _, player in ipairs(getElementsByType("player")) do
>         if isElement(player) then
>             local x, y, z = getElementPosition(player)
>             
>             if not playersTraveledDistance[player] then
>                 playersTraveledDistance[player] = {
>                     lastPosition = { x, y, z },
>                     totalDistance = 0
>                 }
>             end
> 
>             local playerData = playersTraveledDistance[player]
>             local lastPos = playerData.lastPosition
> 
>             local dx, dy, dz = x - lastPos[1], y - lastPos[2], z - lastPos[3]
>             local distance = math.sqrt(dx * dx + dy * dy + dz * dz)
> 
>             playerData.totalDistance = playerData.totalDistance + distance
>             playerData.lastPosition = { x, y, z }
>         end
>     end
> end, 1000, 0)
> ```

Regarding the minimum value, I found it valid to change. As for altering the player's position on the client side, I believe that if someone truly wants to protect their server and avoid desynchronizations, they wouldn't use this approach. Furthermore, in larger servers, your example code wouldn't be optimized and would likely cause issues. I tested the first code version with a "hacker" colleague, and it worked very well.

ffsPLASMA:
I would rather put it on server network level whenever the client sends a new position sync update and distance is greater than a  server config var, trigger the event. That behaviour should also ignore sync updates done by server itself via setElementPosition or respawnPlayer or whatever.

Also account for all client updates done in whatever way, in your example, what does the client stop from executing custom lua code with setElementPosition.

imfelipedev:
> I would rather put it on server network level whenever the client sends a new position sync update and distance is greater than a server config var, trigger the event. That behaviour should also ignore sync updates done by server itself via setElementPosition or respawnPlayer or whatever.
> 
> Also account for all client updates done in whatever way, in your example, what does the client stop from executing custom lua code with setElementPosition.

But the code is doing exactly that. The `onPlayerTeleport` event will only be triggered when the **client** sends a significantly distant position to the server. This code was not designed to prevent a client-side script from using `setElementPosition` but rather to alert the server when a **client** moves irregularly without prior server action.

Dutchman101:
This is useful for classic hacks (non-lua injector) and it's server-side, only one extra tool to raise the border. Especially useful now that most Lua injector hacks were patched this month, at least for now they are rare.

Server owners should be mindful of its limitations, log first, and add enough checks.

Please make a follow up PR for an event to also keep track of multiple smaller moves that add up to a large move together, if within a certain interval it'll trigger that.

Thanks! @imfelipedev (and please add the new event to wiki)

PR #3964, opened 2025-01-18
Fix unintended behavior for ped control states 

### LINKED ISSUE #3874: GetPedControlState returns true for the other player when using setPedControlState for yourself
### Describe the bug

If you use setPedControlState on yourself, then getPedControlState used on a nearby player will return true, even if the player is not executing any controls

### Steps to reproduce

1. Set yourself a control for example vehicle_left
`setPedControlState(localPlayer, "vehicle_left", true)`
2. Now use getPedControlState on the player next to you with the same control we set for ourselves

```
local player = getPlayerFromName("xyz") 
outputChatBox(inspect(getPedControlState(player, "vehicle_left")))
```

You can now see that the function will return true if the player does not turn left

### Version

Client: 1.6-release-22789
Server: 1.6-release-22789

ghost:
Fixes #3874

### DIFF
--- a/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
+++ b/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
@@ -1710,42 +1710,39 @@ bool CStaticFunctionDefinitions::GetPedClothes(CClientPed& Ped, unsigned char uc
     return false;
 }
 
-bool CStaticFunctionDefinitions::GetPedControlState(CClientPed& Ped, const char* szControl, bool& bState)
+bool CStaticFunctionDefinitions::GetPedControlState(CClientPed& const ped, const std::string_view control, bool& state) noexcept
 {
-    if (&Ped == GetLocalPlayer())
-    {
-        return GetControlState(szControl, bState);
-    }
+    if (&ped == GetLocalPlayer())
+        return GetControlState(control.data(), state);
 
-    if (Ped.GetType() == CCLIENTPLAYER)
+    if (ped.GetType() == CCLIENTPLAYER)
     {
-        CControllerState cs;
-        Ped.GetControllerState(cs);
-        bool bOnFoot = (!Ped.GetRealOccupiedVehicle());
-        bState = CClientPad::GetControlState(szControl, cs, bOnFoot);
-        float        fState = 0;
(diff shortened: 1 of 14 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.h
```diff
@@ -96,7 +96,7 @@ class CLuaPedDefs : public CLuaDefs
     static bool IsPedReloadingWeapon(CClientPed* const ped) noexcept;
     LUA_DECLARE(AddPedClothes);
     LUA_DECLARE(RemovePedClothes);
-    LUA_DECLARE(SetPedControlState);
+    static bool SetPedControlState(CClientEntity* const entity, const std::string_view control, const bool state) noexcept;
```
tederis: `CClientPed*`? I know that `CStaticFunctionDefinitions::SetPedControlState` accepts `CClientEntity&`, but `SetPedControlState` corresponds to `Ped` and we should stick to `CClientPed`. You can change `CStaticFunctionDefinitions::SetPedControlState` if you wish.

ghost: Done

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -2366,24 +2363,13 @@ bool CStaticFunctionDefinitions::RemovePedClothes(CClientEntity& Entity, unsigne
...
-    if (IS_PED(&Entity))
-    {
-        CClientPed& Ped = static_cast<CClientPed&>(Entity);
-
-        if (&Ped == GetLocalPlayer())
-        {
-            return SetControlState(szControl, bState);
-        }
+    if (&ped == GetLocalPlayer())
+        return SetControlState(control.data(), state);
```
tederis: Although `std::string_view` is acceptable in CLuaPedDefs, it's not recommended to use it here because `SetPedControlState` can potentially be called from any place and a string isn't guaranteed to be null-terminated. The same for `GetPedControlState`.

ghost: Thanks!

PR #3965, opened 2025-01-18
Fix ped sync behavior

ghost:
Fixes unexpected ped sync behavior after #3936

### DIFF
--- a/Client/mods/deathmatch/logic/CPedSync.cpp
+++ b/Client/mods/deathmatch/logic/CPedSync.cpp
@@ -269,8 +269,6 @@ void CPedSync::Packet_PedSync(NetBitStreamInterface& BitStream)
                     pPed->SetOnFire(bOnFire);
                 if (BitStream.Version() >= 0x55 && ucFlags & 0x40)
                     pPed->SetInWater(bIsInWater);
-                if (BitStream.Can(eBitStreamVersion::IsPedReloadingWeapon) && ucFlags2 & 0x1 && isReloadingWeapon)
-                    pPed->ReloadWeapon();
             }
         }
     }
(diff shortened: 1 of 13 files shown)


ghost:
cc @tederis

Client/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -257,6 +269,8 @@ void CPedSync::Packet_PedSync(NetBitStreamInterface& BitStream)
                     pPed->SetOnFire(bOnFire);
                 if (BitStream.Version() >= 0x55 && ucFlags & 0x40)
                     pPed->SetInWater(bIsInWater);
+                if (BitStream.Can(eBitStreamVersion::IsPedReloadingWeapon) && ucFlags2 & 0x1 && isReloadingWeapon)
+                    pPed->ReloadWeapon();
```
tederis: Why do we need that? We already have **RPC** packet `RELOAD_PED_WEAPON` which triggers the reloading. I would rather add reloading to `CPedStartSyncPacket`.

tederis:
This is potentially backwards incompatible. But due to the bit flags overlapping it may be hard to distinguish whether it pre-PR or post-PR behaviour. I think this could be merged without noticable issues.

See the explanation in Discord: https://discord.com/channels/801330706252038164/801411628600000522/1330055415760818208

Client/mods/deathmatch/logic/CPedSync.cpp
ghost: Removed

ghost:
Could we get more maintainers' opinions on this?

botder:
We could also just ignore the issue and live with the overlap and simply get rid of the BitStream-versioned blocks with the next version bump

ghost:
> We could also just ignore the issue and live with the overlap and simply get rid of the BitStream-versioned blocks with the next version bump

Okay, let's close this then?

botder:
Fine by me.

tederis:
The problem with it is that without this PR the ped syncronization becomes unstable which potentially can (but depends on internal BitStream implementation) lead to crashes.

PR #3973, opened 2025-01-22
Apply element filter for buildings in engineApplyShaderToWorldTexture

TheNormalnij:
Allows applying shaders to specific created building elements.

Example:
```lua
local building = createBuilding(8395, 100, 0, 3)
local buildingShader = dxCreateShader("building.fx",0,0 ,false,"world")
engineApplyShaderToWorldTexture(buildingShader, "luxorwindow01_128", building)
```

### DIFF
--- a/Client/game_sa/CBuildingsPoolSA.cpp
+++ b/Client/game_sa/CBuildingsPoolSA.cpp
@@ -48,6 +48,16 @@ inline bool CBuildingsPoolSA::AddBuildingToPool(CClientBuilding* pClientBuilding
     return true;
 }
 
+CClientEntity* CBuildingsPoolSA::GetClientBuilding(CBuildingSAInterface* pGameInterface) const noexcept
+{
+    auto poolIndex = (*m_ppBuildingPoolInterface)->GetObjectIndex(pGameInterface);
+
+    if (poolIndex == -1)
+        return nullptr;
+
+    return m_buildingPool.entities[poolIndex].pClientEntity;
+}
+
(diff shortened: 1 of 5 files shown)


Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -48,6 +48,16 @@ inline bool CBuildingsPoolSA::AddBuildingToPool(CClientBuilding* pClientBuilding
     return true;
 }
 
+CClientEntity* CBuildingsPoolSA::GetClientBuilding(CBuildingSAInterface* pGameInterface) const noexcept
+{
+    auto poolIndex = (*m_ppBuildingPoolInterface)->GetObjectIndex(pGameInterface);
```
tederis: It's better to avoid `auto` for small types like int.

TheNormalnij: agree

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -48,6 +48,16 @@ inline bool CBuildingsPoolSA::AddBuildingToPool(CClientBuilding* pClientBuilding
     return true;
 }
 
+CClientEntity* CBuildingsPoolSA::GetClientBuilding(CBuildingSAInterface* pGameInterface) const noexcept
+{
+    std::uint32_t poolIndex = (*m_ppBuildingPoolInterface)->GetObjectIndex(pGameInterface);
+
+    if (poolIndex == -1)
```
tederis: Although `-1` will implicitly be promoted to unsigned, maybe better to explicitly cast it? Just to avoid misreadings in future.

TheNormalnij: Added

tederis (approved):
LGTM

botder (approved):

PR #3975, opened 2025-01-24
CEGUI 'modern' interface

Lpsd:
### Summary
This PR 'extends' the CEGUI skinning implementation, essentially allowing two skins to be used at the same time; the user selected skin, and the 'modern' skin which MTA provides.

The main benefit of this is the ability to toggle between two skins for created CEGUI elements, without needing to rebuild the entire GUI, thereby allowing us to **build a new main menu design** without affecting in-game/server GUI.

**The 'modern' skin used in this PR is simply an example to prove the 'multi-skin' functionality works as intended, and not actually meant to look good or change the overall design of the GUI at this stage**

As such this should only be merged alongside other main menu revamp PRs submitted later on. I thought it'd be best to split things up and allow for individual parts of the revamp to get reviewed along the way, making it easier for everyone.

### How it works
By generating a merged version of the user selected skin & the modern skin upon launch, plus whenever the user selects a new skin. Merged files include various XML configurations and the PNG imageset.

The modern template original version lives under `MTA/cgui/modern/templates`  
The merged skin gets created under `MTA/cgui/modern/autogenerated`  
  
The modern template has `Modern` appended to Falagard mapping definition names, which allows a different LookNFeel definition to be used when `CGUI::SetModernSkinEnabled` is used (see implementation and usage of `CGUI_Impl::ResolveModernName`). This also avoids clashes with the user selected skin.

```xml
<FalagardMapping WindowType="CGUI/TitlebarModern" TargetType="Falagard/Titlebar" LookNFeel="CGUI/TitlebarModern" />
```
```xml
<WidgetLook name="CGUI/TitlebarModern">
    <Property name="CaptionColour" value="FFFFFFFF" />
    <Property name="MovingCursorImage" value="set:CGUI-Images image:MouseMoveCursorModern" />
     ...
```

Imagesets in the modern skin are defined the same as in a normal skin, the position values will get translated automatically when the skins are merged. Again, `Modern` is appended to definition names.

```xml
<Imageset Name="CGUI-Images" Imagefile="CGUI.png" NativeHorzRes="800" NativeVertRes="600" >
    <Image Name="TabTopModern" XPos="75" YPos="128" Width="20" Height="7" />
    <Image Name="TabLeftModern" XPos="71" YPos="135" Width="4" Height="25" />
```

The modern skin is versioned, and hashes of the user selected skin are stored in the autogenerated skin directory, so the 'merged' skin is only built when necessary.

*Example of merged imageset (user skin left, modern skin right)*
CGUI

Currently (for testing purposes), the generation/usage of modern skin relies on a CVAR `cgui_modern_interface` being true, which can be changed via F8 console (`cgui_modern_interface=1` or `0`) or via `Settings > Interface > Use modern interface`

### Testing

A main menu setting has been added (enabled by default) under `Interface > Use modern interface`.

image

Enabling this setting will cause the majority of the main menu to use the 'modern' skin. Some windows will still be generated with the user-selected skin (which is just a dark/black theme), such as "host game" and connection status dialog. This is due to the fact we're enabling the usage of modern skin inside `CLocalGUI::CreateWindows` only, and some of MTA's GUI / main menu actually gets created outside of this method.

image

As this is only for testing purpose, it serves as a way to quickly test the difference between user selected skin and the modern skin without connecting to a server. It goes without saying that server GUI should remain unchanged, and will not use the modern skin.

The current 'modern' skin is just a copy of GWEN Orange iirc, and I've made it into a dark theme instead.

### Additional changes

As part of this work, I've also introduced a new method for placeholder color text. 

MTA main menu uses labels over the top of inputs to act as a placeholder, hidden once the input gains focus; the color of this label is hardcoded in C++ implementation to black (the opposite of the `TextColours` definition used in **all the default skins we provide**: `<Property name="TextColours" value="tl:FFFFFFFF tr:FFFFFFFF bl:FFFFFFFF br:FFFFFFFF" />`). This is bad, since you don't allow the skin to define its own placeholder text color.

To resolve this, I've added new methods `CGUILabel_Impl::InvertTextColor()` and `CGUILabel_Impl::SetPlaceholderColors()`, the latter is used in replace of instances where we've hardcoded placeholder text color.

`SetPlaceholderColors` will attempt to look for a `"PlaceholderTextColours"` property in the skin's StaticText definition (LookNFeel) and set the color based on that. If this doesn't exist (which is the case for all our default skins) the text color gets inverted instead, resulting in the same behaviour/color as prior for default skins, when modern interface is enabled.

*excuse the usage of color and colour, MTA uses US variant and CEGUI uses GB*


Fernando-A-Rocha:
Fantastic job. Let us know when you'd like to receive in depth code reviews.

Lpsd:
Whenever someone has time, this code won't change much now unless there are fixes required or additional features. It's in draft mode purely so it doesn't get merged, even if reviewed.

Client/core/CSettings.h
```diff
@@ -288,6 +288,9 @@ class CSettings
     CGUIComboBox* m_pInterfaceSkinSelector;
     CGUIButton*   m_pInterfaceLoadSkin;
 
+    CGUILabel* m_pInterfaceModernLabel;
+    CGUICheckBox* m_pInterfaceModern;
```
botder, line 292:
```suggestion
    CGUILabel* m_pInterfaceModernLabel{};
    CGUICheckBox* m_pInterfaceModern{};
```

Client/core/CGUI.h
```diff
@@ -118,4 +118,5 @@ class CLocalGUI : public CSingleton<CLocalGUI>
     SString m_LastSkinName;
     SString m_LastLocaleName;
     uint    m_LocaleChangeCounter;
+    bool    m_ModernSkinEnabled;
```
botder, line 121:
```suggestion
    bool    m_ModernSkinEnabled{};
```

Client/gui/CGUI_Impl.h
```diff
@@ -350,4 +360,8 @@ class CGUI_Impl : public CGUI, public CGUITabList
     bool         m_HasSchemeLoaded;
     SString      m_CurrentSchemeName;
     CElapsedTime m_RenderOkTimer;
+
+    bool                m_bUseModernSkin = false;
+    CGraphicsInterface* m_pGraphics;
+    CXML*               m_pXML;
```
botder, line 366:
```suggestion
    CGraphicsInterface* m_pGraphics{};
    CXML*               m_pXML{};
```

Client/gui/CGUI_Impl.cpp
```diff
@@ -115,15 +121,23 @@ void CGUI_Impl::SetSkin(const char* szName)
         CEGUI::SchemeManager::getSingleton().unloadScheme(m_CurrentSchemeName);
     }
 
-    PushGuiWorkingDirectory(CalcMTASAPath(PathJoin("skins", szName)));
+    bool bConvertSkin = ConvertToModernSkin(szName);
+
+    if (!bConvertSkin)
+    {
+        BrowseToSolution("gui-skin", EXIT_GAME_FIRST, SString("Error converting skin '%s' to modern", szName));
```
botder, line 128: Future note: we have to create this help page in the future when this gets merged.

Client/gui/CGUI_Impl.cpp
```diff
@@ -150,6 +164,351 @@ void CGUI_Impl::SetSkin(const char* szName)
     m_eInputMode = INPUTMODE_NO_BINDS_ON_EDIT;
 }
 
+bool CGUI_Impl::ConvertToModernSkin(const char* skinName)
+{
+    if (!m_pXML)
+    {
+        AddReportLog(1337, "ConvertToModernSkin: XML interface not initialized");
```
botder, line 171: Future note: we have to pick a different number here

botder (changes requested):

Client/gui/CGUI_Impl.cpp
botder: We have a new repository for this now: https://github.com/multitheftauto/help.multitheftauto.com

tryangelic:
Do you have in mind to return with the project, @Lpsd ? I'd think it'd be pretty cool, really.

PR #3976, opened 2025-01-24
Fix setElementDoubleSided resets after building manipulations

TheNormalnij:
WTR:
```lua
local building = createBuilding(1337, 0, 0, 3)
setElementDoubleSided(building, true)
setElementPosition(building, 1, 0, 3)
```

### DIFF
--- a/Client/mods/deathmatch/logic/CClientBuilding.cpp
+++ b/Client/mods/deathmatch/logic/CClientBuilding.cpp
@@ -132,6 +132,10 @@ void CClientBuilding::Create()
     ConvertZXYEulersToQuaternion(m_vRot, vRot4D);
 
     m_pBuilding = g_pGame->GetPools()->GetBuildingsPool().AddBuilding(this, m_usModelId, &m_vPos, &vRot4D, m_interior);
+
+    if (!m_pBuilding)
+        return;
+
     m_pBuilding->SetBackfaceCulled(!m_bDoubleSided);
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/CClientBuilding.cpp
```diff
@@ -132,6 +132,7 @@ void CClientBuilding::Create()
     ConvertZXYEulersToQuaternion(m_vRot, vRot4D);
 
     m_pBuilding = g_pGame->GetPools()->GetBuildingsPool().AddBuilding(this, m_usModelId, &m_vPos, &vRot4D, m_interior);
+    m_pBuilding->SetBackfaceCulled(!m_bDoubleSided);
```
tederis, line 139: Is there any chance of `m_pBuilding` being `nullptr`?

TheNormalnij: There is an edge case, but i prevent it on API level

botder: According to the source code, it can return `nullptr` when `if (!HasFreeBuildingSlot())`

TheNormalnij: Ok, but I'll prefer a crash in this case.

botder (approved):

PR #3978, opened 2025-01-25
Temporary textures cleanup

### LINKED ISSUE #3977: Memory leak, temporary textures not deleting
### Describe the bug

Currently the MTA, when we use the dxDrawImage or dxDrawImageSection functions it will create a temporary texture, however this texture is not being deleted when it is not in use. This results in high memory consumption.

A solution that I believe is viable for this system would be to create a timer that will validate if that texture has not been used for X amount of time. If it is unused, the system will automatically delete the texture, freeing up memory.

### Steps to reproduce

Create an image render using the dxDrawImage or dxDrawImageSection functions. Then disable the render, using the getElementsByType function you can see that the total number of textures has increased, but even if you don't use it, it won't go down. Below is the example code:

```lua
local tick = 0

function render()
    local now = getTickCount()

    if (now - tick >= 1000) then
        return removeEventHandler("onClientRender", root, render)
    end

    dxDrawImage(0, 0, 100, 100, "assets/image.png", 0, 0, 0, tocolor(255, 255, 255, 255))
end

addCommandHandler("textures", function()
    print("Total textures:", #getElementsByType("texture", resourceRoot))
end)

```

### Version

Client: v1.6-release-22894 (Windows 11, 64-bit)
Server: v1.6-release-22894 (Windows 11, 64-bit)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

tederis:
Resolves #3977

### DIFF
--- a/Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
@@ -11,6 +11,12 @@
 #include "StdInc.h"
 #include "CClientVectorGraphic.h"
 
+namespace
+{
+    const std::int64_t TEMPORARY_TEXTURES_CLEANUP_THRESHOLD = 1000ll;
+    const std::size_t TEMPORARY_TEXTURES_DELETE_THRESHOLD = 2u;
+}
+
 ////////////////////////////////////////////////////////////////
 //
 // CClientRenderElementManager::CClientRenderElementManager
@@ -283,20 +289,22 @@ CClientVectorGraphic* CClientRenderElementManager::CreateVectorGraphic(uint widt
 CClientTexture* CClientRenderElementManager::FindAutoTexture(const SString& strFullFilePath, const SString& strUniqueName)
 {
     // Check if we've already done this file
-    CClientTexture** ppTextureElement = MapFind(m_AutoTextureMap, strUniqueName);
+    SAutoTexture* ppTextureElement = MapFind(m_AutoTextureMap, strUniqueName);
     if (!ppTextureElement)
     {
         // Try to create
         CClientTexture* pNewTextureElement = CreateTexture(strFullFilePath);
         if (!pNewTextureElement)
-            return NULL;
+            return nullptr;
 
         // Add to automap if created
-        MapSet(m_AutoTextureMap, strUniqueName, pNewTextureElement);
+        MapSet(m_AutoTextureMap, strUniqueName, SAutoTexture{pNewTextureElement});
         ppTextureElement = MapFind(m_AutoTextureMap, strUniqueName);
     }
 
-    return *ppTextureElement;
+    ppTextureElement->lastUse = CTickCount::Now();
+
+    return ppTextureElement->texture;
 }
 
 ////////////////////////////////////////////////////////////////
@@ -318,9 +326,9 @@ void CClientRenderElementManager::Remove(CClientRenderElement* pElement)
     // Remove from auto texture map
     if (pElement->IsA(CClientTexture::GetClassId()))
     {
-        for (std::map<SString, CClientTexture*>::iterator iter = m_AutoTextureMap.begin(); iter != m_AutoTextureMap.end(); ++iter)
+        for (auto iter = m_AutoTextureMap.begin(); iter != m_AutoTextureMap.end(); ++iter)
         {
-            if (iter->second == pElement)
+            if (iter->second.texture == pElement)
             {
                 m_AutoTextureMap.erase(iter);
                 break;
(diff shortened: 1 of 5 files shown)


Fernando-A-Rocha (approved):
Nice system of automatic memory cleaning.

Client/mods/deathmatch/logic/CElementDeleter.cpp
```diff
@@ -19,14 +19,17 @@ CElementDeleter::CElementDeleter()
...
     if (pElement && !IsBeingDeleted(pElement))
     {
         // Before we do anything, fire the on-destroy event
-        CLuaArguments Arguments;
-        pElement->CallEvent("onClientElementDestroy", Arguments, true);
+        if (!silent)
+        {
+            CLuaArguments Arguments;
+            pElement->CallEvent("onClientElementDestroy", Arguments, true);
+        }
```
TheNormalnij: We can keep the event trigger to track deleted textures

Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
```diff
@@ -11,6 +11,12 @@
 #include "StdInc.h"
 #include "CClientVectorGraphic.h"
 
+namespace
+{
+    const std::int64_t TEMPORARY_TEXTURES_CLEANUP_THRESHOLD = 10000ll;
+    const std::size_t TEMPORARY_TEXTURES_DELETE_THRESHOLD = 10u;
+}
+
```
TheNormalnij: Does `namespace` have any effect here?
+ `constexpr`

tederis: The unnamed namespace in this context is the equivalent to `static` specifier to keep these variables inside a translation unit. Just the modern way of doing it. https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces

Client/mods/deathmatch/logic/CElementDeleter.cpp
tederis: These temporary textures are internal entities and I'm not sure if it's a good idea to expose them by calling `onClientElementDestroy`.

Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
TheNormalnij: Could you explain what it means practically? Afaik these variables will be inside this cpp without a namespace too.

tederis: Example:

one.cpp
`int varA = 10;`

two.cpp
`int varA = 10;`

This will cause a compile time error: `one or more multiply defined symbols found.`
And to prevent it you have to use `static` or an unnamed namespace.

Client/mods/deathmatch/logic/CElementDeleter.cpp
TheNormalnij: It might be good for debug purposes. I think anonymous textures are exposed in the elements tree already.

Client/mods/deathmatch/logic/CClientRenderElementManager.cpp
TheNormalnij: You are right. I didn't have issues with multiply defined symbols before.
I checked it. `constrexpr` also prevents this error without a namespace.

tederis: Yeah, okay.

Client/mods/deathmatch/logic/CElementDeleter.cpp
tederis: I'll look how it works.

botder (dismissed):
LGTM. What do you think about adding a timer to run `CClientRenderElementManager::DoPulse()` only every `N` (milli-)seconds?

tederis:
> LGTM. What do you think about adding a timer to run `CClientRenderElementManager::DoPulse()` only every `N` (milli-)seconds?

Yes, it makes sence.

Client/mods/deathmatch/logic/CElementDeleter.cpp
tederis: I've allowed onClientElementDestroy events

TheNormalnij (approved):

Dutchman101:
This is huge, thanks

Dutchman101:
Temporarily reverted, see details at [https://github.com/multitheftauto/mtasa-blue/pull/4031#issuecomment-2664054520](https://github.com/multitheftauto/mtasa-blue/pull/4031#issuecomment-2664054520)

PR #3981, opened 2025-01-26
Fix SVG bitmap and optimize rendering

### LINKED ISSUE #3828: SVG colors are definitely wrong

TheNormalnij:
Resolves #3828
This PR is an alternative to #3829 with better performance.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
+++ b/Client/mods/deathmatch/logic/CClientVectorGraphicDisplay.cpp
@@ -10,7 +10,6 @@
 #include "StdInc.h"
 #include "CClientVectorGraphicDisplay.h"
 #include "CClientVectorGraphic.h"
-#include <lunasvg.h>
 
 using namespace lunasvg;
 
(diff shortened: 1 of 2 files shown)


Lpsd (approved):
Thanks for sorting this, lgtm

PR #3983, opened 2025-01-26
Fix armor limit and synchronization between client and server

### LINKED ISSUE #3982: Armor value above 100 client side bypasses max armor limit on server
### Describe the bug

When setting an armor value greater than 100 on the client side, the value is synchronized with the server, bypassing the intended armor cap limit. However, the armor value is only synced up to a maximum of 127.5 on the server. On the client side, there is no armor limit applied, allowing players to exceed the maximum armor limit of 100.

### Steps to reproduce

1. `crun setPedArmor(me, 9999)`
2. `crun getPedArmor(me)`
3. `srun getPedArmor(me)`

### Version

Client: v1.6.0-release-22920 (Windows 11, 64-bit)
Server: v1.6.0-release-22920 (Windows 11, 64-bit)

### Additional context

_No response_

### Relevant log output

```shell
Executing client-side command: setPedArmor(me, 9999)
Executing client-side command: getPedArmor(me)
Command results: 9999 [number]
MaliciousBonnet47 executed command: getPedArmor(me)
Command results: 127.5 [number]
```

### Security Policy

ghost:
Fixes #3982


ghost:
Just an opinion: Maybe the cap should be 150 and not 100 as this is an original (singleplayer) behaviour.
The max health bug will be fixed after 1.6.1 which means that players will have the original maximum max health(176) instead of 200, so it would be better to have the armor cap raised to 150 as well.

ghost:
> Just an opinion: Maybe the cap should be 150 and not 100 as this is an original (singleplayer) behaviour. The max health bug will be fixed after 1.6.1 which means that players will have the original maximum max health(176) instead of 200, so it would be better to have the armor cap raised to 150 as well.

That will be in a separate pull request because this change is needed as soon as possible so that cheaters cannot exploit it, and it would also require a deeper modification to the synchronization, which is needed anyway because the current armor sync is inaccurate

ghost:
This pull request is fully backward compatible, I tried to keep the number of changes as low as possible so that it can be merged as soon as possible

PR #3984, opened 2025-01-26
Make single link node pool unlimited

TheNormalnij:
Fixes crash with offset 0x00133606.
The crash related to the high density of buildings.
The main problems with this solution are malloc and free.


TheNormalnij:
This solution leads to high memory fragmentation.
Im going to replace it with pool reallocation.

PR #3986, opened 2025-01-28
Make single link node pool unlimited

TheNormalnij:
Fixes crash with offset 0x00133606.
The crash related to the high density of buildings.


tederis:
~~You did a lot of work and it's great. I just want to understand why did you choose creating `CDynamicPool` instead of just using `std::vector`? In the context of this PR `std::vector` allows to find elements with O(1) complexity(by comparing data pointers), it also allows you to remove elements with O(1) complexity(swap and pop ideom). `std::vector` has to be a reasonable choice here.~~

Okay, I see that memory pieces have to be persistent and we cannot use swap and pop ideom here.

Client/game_sa/CDynamicPool.h
```diff
@@ -0,0 +1,192 @@
...
+#include <memory>
+
+template <typename PoolObjT>
+class CDynamicPoolPart
+{
+public:
+    explicit CDynamicPoolPart(std::size_t size) : m_size{size}, m_usedSize{0}, m_lastFreeSlot{0}
+    {
+        m_items = std::make_unique<PoolObjT[]>(size);
+        m_usedSlots = std::make_unique<bool[]>(size);
```
tederis: Did you think about using a stack of free indices? It will be much more efficient than iterating over array of bools searching for a free item.

Dutchman101:
> Fixes crash with offset 0x00133606. The crash related to the high density of buildings.

This crash is exceedingly rare on MTA (measured by the latest build with a high spread; crash stats) so to judge the PR's feasability it would be good to know if it's worth it, as in: are there no negative/side effects at play, like a global increase of RAM usage as the pool allocates more? Is it dynamically expanding and going back down depending on a servers' usage of map and buildings, so that as long the usage isn't huge, there are no negative effects?

In the latter case it wouldn't be too bad to have it in. If it has no negative effects, proportionality to what it aims to fix isn't a thing.

@TheNormalnij If you could please make a list of any effects you expect from this change?

Client/game_sa/CDynamicPool.h
TheNormalnij: I like your idea. It's less memory efficient but gives better performance. 
My current solution takes 810 KB for 90k objects. 9 byte/object. The game engine has the same memory usage.
Your solution requires 1080 KB for the same amount of objects (12 byte/object). 
If you think that extra memory usage isn't significant, i can use a stack here

tederis: Personally, I don't see these extra ~200 KB as a problem. The memory budgets in SA were set taking in account the possibilities of PS2, but most of todays PCs have much more RAM and we can allow ourselves to use it more freely. A stack can be implemented on preallocated(because we know the size of it in advance) `std::vector`(just `push_back` and `pop_back`) as it will be more efficient than `std::stack` because the latter leads to excessive memory fragmentation(especially on **MSVC**).

TheNormalnij:
> > Fixes crash with offset 0x00133606. The crash related to the high density of buildings.
> 
> This crash is exceedingly rare on MTA (measured by the latest build with a high spread; crash stats) so to judge the PR's feasability it would be good to know if it's worth it, as in: are there no negative/side effects at play, like a global increase of RAM usage as the pool allocates more? Is it dynamically expanding and going back down depending on a servers' usage of map and buildings, so that as long the usage isn't huge, there are no negative effects?
> 
> In the latter case it wouldn't be too bad to have it in. If it has no negative effects, proportionality to what it aims to fix isn't a thing.
> 
> @TheNormalnij If you could please make a list of any effects you expect from this change?

This crash is rare because developers use createObjects instead of createBuilding when their map crashes.
I'm trying to increase the building limit because buildings fix fps issues on big maps. This allows more creative maps with better lighting.
This PR doesn't cause significant memory usage. I'll implement TEDERIS's suggestion, which requires an additional ~200 KB on game startup. createBuilding is still more memory efficient. 10k buildings use 1.5 MB less than 10k objects. The pool resets capacity on reconnect and can be resized manually via engineSetPoolCapacity.

The pool implementation is very universal. I plan to use it in future fixes too.

Client/game_sa/CDynamicPool.h
TheNormalnij: Done

Client/game_sa/CDynamicPool.h
```diff
@@ -0,0 +1,171 @@
...
+
+    PoolObjT* AllocateItem()
+    {
+        for (auto& pool : m_poolParts)
+        {
+            if (pool->HasFreeSize())
+                return pool->AllocateItem();
+        }
+
+        return AllocateNewPart()->AllocateItem();
```
tederis: An exception std::bad_alloc can be thrown here.

TheNormalnij: Do you have an idea how to handle it?

tederis: Ideally there should be a way to prevent the game from crashing. I know that this implies many changes so it's out of the scope of this PR. But at least we can put an assertation with a message. Something like

```
try
{
    return AllocateNewPart()->AllocateItem();
}
catch (const std::bad_alloc&)
{
    assert(false && "Could not allocate a memory for single link nodes");
}
```

I believe that this is better than leaving an exception unhandled.

Client/game_sa/CDynamicPool.h
```diff
@@ -0,0 +1,171 @@
...
+#include <memory>
+
+template <typename PoolObjT>
+class CDynamicPoolPart
+{
+public:
+    explicit CDynamicPoolPart(std::size_t size) : m_size{size}
+    {
+        m_items = std::make_unique<PoolObjT[]>(size);
+        m_unusedIndexes.reserve(size);
```
Pirulax: typo: `indices`

Client/game_sa/CDynamicPool.h
```diff
@@ -0,0 +1,171 @@
...
+};
+
+template <std::size_t InitialSize>
+struct PoolGrownByHalfStrategy
+{
+    static constexpr std::size_t GetInitialSize() { return InitialSize; }
+    static constexpr std::size_t GetNextSize(std::size_t index) { return InitialSize / 2; }
+};
+
+template <typename PoolObjT, typename GrownStrategy>
```
Pirulax: typo: `GrowStrategy`

Client/game_sa/CDynamicPool.h
```diff
@@ -0,0 +1,171 @@
...
+        }
+        
+        return true;
+    }
+
+private:
+    CDynamicPoolPart<PoolObjT>* AllocateNewPart()
+    {
+        const std::size_t nextSize = GrownStrategy::GetNextSize(m_poolParts.size());
+        m_poolParts.emplace_back(std::make_unique<CDynamicPoolPart<PoolObjT>>(nextSize));
```
Pirulax: iirc `emplace_back` returns an iterator to the item

Pirulax (commented):
I feel like that a traditional linked list of these "part"s would  be better than having a vector of them, especially when allocating a free-list is much better in terms of performance than a basically O(N) lookup.

Client/game_sa/CDynamicPool.h
TheNormalnij: It returns void in C++ 14

TheNormalnij:
> I feel like that a traditional linked list of these "part"s would be better than having a vector of them, especially when allocating a free-list is much better in terms of performance than a basically O(N) lookup.

The performance cost for allocating `m_poolParts` isn't really important here. But I agree `std::vector<std::unique_ptr<CDynamicPoolPart<PoolObjT>>>` is weird. It should be `std::list<CDynamicPoolPart<PoolObjT>>` or `std::vector<DynamicPoolPart<PoolObjT>>`. The second variant causes less memory fragmentation.

tederis:
The `std::vector` was fine because an array of `CDynamicPoolPart` ptrs isn't something frequently changing. Technically `std::list` is also okay, but for small types like uint or `std::unique_ptr` it performs worse because of cache misses and memory fragmentation. In most cases `std::vector` is the best choice when you need to iterater over an array of small fixed(non-volatile) objects. But nvm, I just wanted to note that your previous solution was also fine.

Dutchman101:
Note for future investigation: This may have introduced a crash, see [https://i.imgur.com/Or8lDDi.png](https://i.imgur.com/Or8lDDi.png) for details
cc @tederis 

I couldn't gather much from a quick look - but the crash occurs in this code.

PR #3994, opened 2025-01-30
Fix incorrect behavior of `onPlayerChangesWorldSpecialProperty` event

ghost:
Fixes incorrect behavior of the `onPlayerChangesWorldSpecialProperty` event, which was previously triggered for every internal call. Also includes minor refactors.

- [x] Remove unnecessary code
- [x] Replace the packet send logic

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -5998,7 +5998,7 @@ bool CClientGame::IsGlitchEnabled(unsigned char ucGlitch)
     return ucGlitch < NUM_GLITCHES && m_Glitches[ucGlitch];
 }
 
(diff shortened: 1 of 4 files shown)


ghost:
Tests:
```lua
local cheats = {
    ["hovercars"] = true,
    ["aircars"] = true,
    ["extrabunny"] = true,
    ["extrajump"] = true
}

addEventHandler("onPlayerChangesWorldSpecialProperty", root,
    function(property, enabled)
        if not cheats[property] then
            return
        end

        if not enabled then
            return
        end

        banPlayer(source, false, false, true, "Server", "Time to take a permanent break :-)")
    end
)
```

Lpsd (approved):
Looks good now, thanks

CArg22:
does it take an account reseting properties to default?

Issue #3993, opened 2025-01-30
`onPlayerTeleport` false positive with default resources


ghost:
### Describe the bug

Hello!

We've tested this new event and it gives false positives with `interiors`  and `freeroam` (F2 menu) resources. Didn't test other resources.
Another thing:

it would be cool when the event triggers,  to get the position, interior, dimension **from** & **to** where the player teleports as params.

### Steps to reproduce

1. start `interiors` or `freeroam`
2. copy & paste these lines into a script file, then start it

```lua
addEventHandler("onPlayerTeleport", root, function() 
    outputServerLog(getPlayerName(source) .. " teleporting.")
end)
```

3. enter into an interior anywhere on the map. Or press F2 and teleport anywhere on the map.

### Version

Client & Server: 22934

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

imfelipedev:
> ### Describe the bug
> Hello!
> 
> We've tested this new event and it gives false positives with `interiors` and `freeroam` (F2 menu) resources. Didn't test other resources. Another thing:
> 
> it would be cool when the event triggers, to get the position, interior, dimension **from** & **to** where the player teleports as params.
> 
> ### Steps to reproduce
> 1. start `interiors` or `freeroam`
> 2. copy & paste these lines into a script file, then start it
> 
> addEventHandler("onPlayerTeleport", root, function() 
>     outputServerLog(getPlayerName(source) .. " teleporting.")
> end)
> 3. enter into an interior anywhere on the map. Or press F2 and teleport anywhere on the map.
> 
> ### Version
> Client & Server: 22934
> 
> ### Additional context
> _No response_
> 
> ### Relevant log output
> ### Security Policy
> * [x]  I have read and understood the Security Policy and this issue is not security related.

Thanks for the report! The false positive happens due to improper use of the `setElementPosition` function on the client-side, causing the trigger to fire incorrectly. I've submitted a PR to fix this in Freeroam script (https://github.com/multitheftauto/mtasa-resources/pull/599) and I'm currently refactoring the interiors script.

CArg22:
What is the point of doing it? now cheater just use this newly created trigger to teleport ...

imfelipedev:
> What is the point of doing it? now cheater just use this newly created trigger to teleport ...

[...]

PR #4001, opened 2025-02-04
Add arguments to `onPlayerTeleport` and fix the false positive caused by `setElementInterior` using x, y, z parameters.

imfelipedev:
Suggested by [issue](https://github.com/multitheftauto/mtasa-blue/issues/3993)


samr46:
It would also be better to update (set) new position of a player on the server before this event is triggered. This way it will be possible to call setElementPosition inside of this event if needed.

Alternatively you could make this event cancellable (so if cancelled then position on the server shouldn't be updated and return packet should be sent to the source player with previous position).

TheNormalnij (approved):

botder (approved):

PR #4005, opened 2025-02-06
Fix building crashes related to rotation

TheNormalnij:
Fixes crashes with offset 0x0014F5C5 and 0x0014F557.
Fixes rotation issues. Now, the rotation for createBuilding is always consistent with createObject


Client/game_sa/CBuildingSA.cpp
```diff
@@ -13,9 +13,13 @@
 #include "CBuildingSA.h"
 #include <game/CWorld.h>
 #include "CGameSA.h"
+#include "CMatrixLinkSA.h"
+#include "CDynamicPool.h"
 
 extern CGameSA* pGame;
 
+static CDynamicPool<CMatrixLinkSAInterface, PoolGrowAddStrategy<0, 500>> EXCLUSIVE_MATRIX_POOL{};
```
botder:
```suggestion
static CDynamicPool<CMatrixLinkSAInterface, PoolGrowAddStrategy<0, 500>> g_matrixPool{};
```

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -98,6 +95,17 @@ CBuilding* CBuildingsPoolSA::AddBuilding(CClientBuilding* pClientBuilding, uint1
 
     // Add building in world
     auto pBuildingSA = new CBuildingSA(pBuilding);
+
+    // Edge case for the traincross2 (1374) model
+    if (pBuilding->HasMatrix())
+        pBuildingSA->ReallocateMatrix();
```
botder: I didn't follow the conversations about this problem, but why not just deallocate any game-allocated matrix by default in every case? Does the game expect this model to have a matrix the whole time? If not, then we don't need `ReallocateMatrix` at all.

botder (changes requested):

Client/game_sa/CBuildingsPoolSA.cpp
TheNormalnij: > Does the game expect this model to have a matrix the whole time?

No. The game allocates a static matrix for buildings that have `x` or `y` rotation in `CVector4D` and always for the traincross2 model. 
The engine uses the same storage for dynamically allocated matrixes and static allocated matrixes. It leads to two edge cases:
1) The game cannot allocate a static matrix for a building. This causes crashes.
2) The game cannot allocate a new dynamic matrix for an object when the storage has no place for it. It may cause rotation bugs for some objects.
3) Performace drop when the game recalcutes dynamic matrixes

I want to avoid this by using custom unlimited matrix storage for scripted buildings.

Client/game_sa/CBuildingSA.cpp
TheNormalnij: Is it right to use `g_` prefix here? This variable is local for this file.

botder: We don't really have a prefix for *file-global* variables, but this prefix will at least give you the signal that the variable lives in a *higher scope*.

Client/game_sa/CBuildingsPoolSA.cpp
botder: In conclusion we should always deallocate the game's matrix here?

TheNormalnij: Yes

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -98,6 +95,22 @@ CBuilding* CBuildingsPoolSA::AddBuilding(CClientBuilding* pClientBuilding, uint1
...
+        // Edge case for the traincross2 (1374) model
+        // LoadObjectInstance allocates a matrix for the model
+        // We need allocate our own matrix and put the old matrix in the original pool
+        pBuildingSA->ReallocateMatrix();
+    }
+    else if (vRot->fX != 0 || vRot->fY != 0)
+    {
+        // Allocate matrix in our unlimited storage instead of using the shared pool.
+        pBuildingSA->AllocateMatrix();
+    }
```
botder, line 110: I would've just deallocated the game's matrix if it exists, and then only allocate it when `if (vRot->fX != 0 || vRot->fY != 0)`

TheNormalnij: The game expects an allocated matrix for this specific model in 0x5A0B50 (CObject::ProcessTrainCrossingBehaviour)

botder: Ok, then it's fine. You responded with a `No` to such a question before, but maybe you discovered this after that.

botder (approved):

PR #4007, opened 2025-02-07
Fix onPlayerTeleport

nweb:
The event did not fire if one of the coordinates was zero, regardless of the distance.
```lua
setElementFrozen(localPlayer, true)
setElementPosition(localPlayer, 9999, 9999, 0)
```

### DIFF
--- a/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
@@ -132,7 +132,7 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
             position.data.vecPosition += vecTempPos;
         }
 
-        if (position.data.vecPosition.fX != 0.0f && position.data.vecPosition.fY != 0.0f && position.data.vecPosition.fZ != 0.0f)
+        // if (position.data.vecPosition.fX != 0.0f || position.data.vecPosition.fY != 0.0f || position.data.vecPosition.fZ != 0.0f)
         {
             CVector playerPosition = pSourcePlayer->GetPosition();
             float playerDistancePosition = DistanceBetweenPoints3D(playerPosition, position.data.vecPosition);


ghost:
I'd like to note it here as well that if the player gets a vehicle and teleports with that, the event won't get triggered.

nweb:
> I'd like to note it here as well that if the player gets a vehicle and teleports with that, the event won't get triggered.

You can also stand on the surface of an object created on the server and move it from the client side and the event will not trigger. Useless event that clutters up the code, just like Textdisplay objects.

botder:
This event is going to have way more false-positives in the future, and I would also assume that most servers are not going to really utilize it, and that scripters are going to misunderstand/misuse this event.

Personally, I would remove it.

Dutchman101:
> This event is going to have way more false-positives in the future, and I would also assume that most servers are not going to really utilize it, and that scripters are going to misunderstand/misuse this event.
> 
> Personally, I would remove it.

We can make a beta note on its wiki page, that it's experimental and can be inaccurate until we learned & fixed more from practical results.

I vouch to do just that: keep iterating on it, try to address every bug and missed implementation for legitimate teleports, until it's nearly perfect.

There's already many functions and events that you can argue scripters who don't think will misuse, and affect their own playerbase with, but MTA's core principle is already a game engine where scripters have full responsibility with a lot of power on which they can mess up hard on nearly every step. Let's just keep providing the freedom to use advanced functions as how one deems fit for their server and gamemodes.

Especially with the recent and still ongoing Lua injector hack patches in MTA AC, cheats may be returning to "classic" ones in the coming period (as in: not involving executing Lua), making events like this useful as a stopgap during times that AC is bypassed by such. We need to give server owners all the tools that we can possibly design, and the principle that any function has to be used in accordance with their existing codebase and in a clever way, applies to any scripting matters, again, not just this new event.

Dutchman101:
Continuing the above discussion can be good, like in another issue or dev discord, but as long nothing came from it, since the feature is still in, fix PR's to it will be accepted so also this one here. Thanks @nweb

naouh:
Possible to add an option to completly disable this? Causing more issues than solving atm.

ghost:
> Possible to add an option to completly disable this? Causing more issues than solving atm.

Just simply don't use `onPlayerTeleport` if you don't want to.

PR #4013, opened 2025-02-09
Fix bad streaming performance in some cases

### LINKED ISSUE #813: Multiple objects in other dimension cause game stuttering
**Describe the bug**
If you create a lot of objects in other dimension that you are in, game will have micro stuttering on streaming it in and out even if onClientElementStreamIn and onClientElementStreamOut aren't triggered.

**To reproduce**
1. set your dimension to 0
2. run for i = 1, 5000 do local obj=createObject(3918, 1292.28002, -1301.80, 35.655); setElementDimension(obj, math.random(500,1000)) end
3. set your camera matrix to  929.28582763672 [number], -1816.6768798828 [number], 259.33786010742 [number], 984.72772216797 [number], -1741.0424804688 [number], 224.61546325684 [number], 0 [number], 70 [number]
4. Go forward.
There will be micro lag between positions:
929.28582763672 [number], -1816.6768798828 [number], 259.33786010742 [number], 984.72772216797 [number], -1741.0424804688 [number], 224.61546325684 [number], 0 [number], 70 [number]
956.34246826172 [number], -1779.7684326172 [number], 242.39280700684 [number], 1011.784362793 [number], -1704.1340332031 [number], 207.67041015625 [number], 0 [number], 70 [number]

**Expected behaviour**
If objects are in other dimension than player, there should be no micro lags. 

**Screenshots**
Moving forwards and backwards when objects aren't created 
image

Moving forwards and backwards when objects are created: 
image

https://youtu.be/ENB16vYPWgY

**Version**
Multi Theft Auto v1.5.6-release-16068.8.011
Copyright (C) 2003 - 2018 Multi Theft Auto

### LINKED ISSUE #3985: Huge FPS drops
### Describe the bug

Something happened because, over the past few weeks, we've noticed huge FPS drops that couldn’t be explained by anything. In the meantime, it seems I figured it out - if too many objects (6-8k) are created, the game simply creates FPS drop "zones" at certain points. In the best case, the FPS just drops for a moment, but in the worst case, the game freezes. I was able to reproduce this on a basic MTA server, so the issue is probably not on our server. It just started 1-2 weeks ago when the last major update came out for MTA (that’s probably where the bug needs to be looked into).

All the objects used in the video are in different dimensions and interiors.
There might be a connection with this bug: #3979 

https://github.com/user-attachments/assets/5f5e27da-9cbf-496b-abc8-ce763c5cd017

https://streamable.com/gg8gm2 < all resources running
https://streamable.com/jso3mm < half of the resources stopped
https://streamable.com/8zlg1o < only safes loaded

### Steps to reproduce

Test resource:
[freezebug_test.zip](https://github.com/user-attachments/files/18576503/freezebug_test.zip)

### Version

Multi Theft Auto v1.6-release-22916

### Security Policy

TheNormalnij:
Fixes [#813](https://github.com/multitheftauto/mtasa-blue/issues/813) and #3985

This PR disables streaming for elements that are outside of the current dimension. This increases performance for servers that have a lot of elements in different dimensions.

This PR fixes this bugged scenario too:
1. A player spawns at 0, 0, 3 position
2. The gamemode creates a lot of objects
3. The player has random fps drops on some map positions.

Why it happens:
MTA creates elements in 0, 0, 0 position by default. CClientStreamer adds these elements to the m_ActiveElements list.
MTA updates element positions BUT doesn't remove elements from the m_ActiveElements list.
This causes the situation when m_ActiveElements holds all possible elements and CClientStreamer requests huge O(N*M) operations.

This PR needs some testing.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientStreamSector.h
+++ b/Client/mods/deathmatch/logic/CClientStreamSector.h
@@ -37,6 +37,7 @@ public:
     void                                       Remove(CClientStreamElement* pElement) { m_Elements.remove(pElement); }
     std::list<CClientStreamElement*>::iterator Begin() { return m_Elements.begin(); }
     std::list<CClientStreamElement*>::iterator End() { return m_Elements.end(); }
+    std::list<CClientStreamElement*>&          GetElements() { return m_Elements; }
 
     void         AddElements(std::list<CClientStreamElement*>* pList);
     void         RemoveElements(std::list<CClientStreamElement*>* pList);
--- a/Client/mods/deathmatch/logic/CClientStreamSectorRow.h
+++ b/Client/mods/deathmatch/logic/CClientStreamSectorRow.h
@@ -31,6 +31,7 @@ public:
     void                                      Add(CClientStreamSector* pSector);
     void                                      Remove(CClientStreamSector* pSector);
     unsigned int                              CountSectors() { return m_Sectors.size(); }
+    std::list<CClientStreamSector*>&          GetList() noexcept { return m_Sectors; }
 
     bool DoesContain(CVector& vecPosition);
     bool DoesContain(float fY);
--- a/Client/mods/deathmatch/logic/CClientStreamer.cpp
+++ b/Client/mods/deathmatch/logic/CClientStreamer.cpp
@@ -502,71 +502,67 @@ void CClientStreamer::Restream(bool bMovedFar)
         }
         else
         {
-            // Same dimension as us?
-            if (pElement->GetDimension() == m_usDimension || pElement->IsVisibleInAllDimensions())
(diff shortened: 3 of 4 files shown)


Xenius97:
Sadly this doesn't solve #3985

Debug
https://streamable.com/mxd937

Release - same as before:
https://streamable.com/8zs2m8

TheNormalnij:
Ok, I optimized element streaming in different dimensions too.

Dutchman101:
Thanks, i gave the commit this description: `Designed to fix #813 and #3985 (needs to be confirmed in practice)`. Soon, we will ask feedback from the people in those 2 related issues.

PR #4018, opened 2025-02-12
Fix for future generations

nweb:


botder:
And what's the point of changing 16 year old code that has worked fine so far, and everyone expects it to be wrong there?
https://int64.org/docs/gamestat-protocols/ase.html

Fernando-A-Rocha:
ASE protocol shouldn't be changed. It is what it is. If we want to modernize, then we can simply create a new standard using some modern structure like JSON for sharing server information with a masterlist.

botder:
> ASE protocol shouldn't be changed. It is what it is. If we want to modernize, then we can simply create a new standard using some modern structure like JSON for sharing server information with a masterlist.

The point of this pull request is for MTA to actually follow the ASE protocol, not to break it. But it's simply too late to change it.

TheNormalnij:
> The point of this pull request is for MTA to actually follow the ASE protocol, not to break it. But it's simply too late to change it.

It's funny enough, but this PR doesn't break backward compability.

0x16 - 0001 0110 // team, skin, ping
0x10 - 0001 0000 // ping

0x32 - 0011 0010 // team, ping, time
0x20 - 0010 0000 // time

The compiler optimizes both variants to `mov <dest>, 63`

botder (approved):

tederis:
> The compiler optimizes both variants to mov <dest>, 63

I don't think that we should rely on it. When deal with optimizing compilers we cannot predict the exact output. It can vary over time and platforms. And especially depends on compiler flags.

Lpsd:
whilst I thought it was interesting, I didn't expect this to get merged. I agree with @tederis

botder:
If you didn't notice, `0x32 | 0x16` in the old code, sets the `0x10` and `0x20` bits in every case, plus some lower bits, which we can safely ignore, because we set all bits there anyway.

PR #4023, opened 2025-02-13
Modernized `start`, `stop` and `restart` commands

### LINKED ISSUE #2549: Add resource server commands
Add serverside commands like in FiveM 

Improve command `start\restart\stop` and add `ensure`

`start\restart\stop`  must support directories with brackets like
`start [cars]`
`ensure` restart the resource (or group) and start if something in argument was stopped like
`ensure [cars]`
`ensure resourcename`

**Additional context**
https://docs.fivem.net/docs/server-manual/server-commands/

P.S perhaps it is *good first issue*

ghost:
Modernized `start`, `stop`, `restart` commands which now also support directories, see the examples below, and from now on, the `restart` command not just restarts the already running resources specified as parameters, but also starts the stopped ones. 
Closes #2549.

```
start play
start admin ip2c playercolors
start [web]

stop play
stop admin ip2c playercolors
stop [web]

restart play
restart admin ip2c playercolors
restart [web]
```


Lpsd:
We should probably add an argument `ensure` to [restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) (defaulting to false)?

The reason I don't suggest adding a separate API function for this (e.g: `ensureResource`), is because we'd need to repeat the arguments and associated logic for those anyway.

Edit: since you'd need to upgrade `restartResource` to the new arg parser anyway, the code would actually end up being reasonably sized; having an additional function such as `ensureResource` wouldn't be so bad.

So, it's up to you and whatever others think is best.

TheNormalnij:
> We should probably add an argument `ensure` to [restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) (defaulting to false)?

Why do you need that when you can get the state of your resource before calling? A new argument doesn't improve readability

TheNormalnij:
I suggest to remove a new 'ensure' command and just change 'restart' behavior. Current 'restart' behavior is disappointing.

Lpsd:
> I suggest to remove a new 'ensure' command and just change 'restart' behavior. Current 'restart' behavior is disappointing.

I agree with this too, but would we want to change the behaviour of `restartResource` function by default?

Technically that isn't backwards compatible, someone may rely on the fact a resource must be running when this function or command is called.

Nonetheless, it would be much better behaviour for the restart command and function to start resources that aren't running, instead of doing nothing/returning false.

Fernando-A-Rocha:
Changing restart command functionality to support starting a resource even if it is stopped does not make sense, as the name implies a stop then a start, not just a start.

New ensure command makes sense, it's practical. I support it.

ghost:
> We should probably add an argument `ensure` to [restartResource](https://wiki.multitheftauto.com/wiki/RestartResource) (defaulting to false)?
> 
> The reason I don't suggest adding a separate API function for this (e.g: `ensureResource`), is because we'd need to repeat the arguments and associated logic for those anyway.
> 
> Edit: since you'd need to upgrade `restartResource` to the new arg parser anyway, the code would actually end up being reasonably sized; having an additional function such as `ensureResource` wouldn't be so bad.
> 
> So, it's up to you and whatever others think is best.

I guess we can discuss that later, but in a separate issue or pull request, this one is focused solely on improving the commands

ghost:
> Changing restart command functionality to support starting a resource even if it is stopped does not make sense, as the name implies a stop then a start, not just a start.
> 
> New ensure command makes sense, it's practical. I support it.

I want to introduce the new command because `restart` is well known and widely used, and I don't want to change that. `ensure` is also common in other game engine like multiplayer modifications, so it should feel familiar to developers interested in modern stuff

TheNormalnij:
> I want to introduce the new command because `restart` is well known and widely used, and I don't want to change that. `ensure` is also common in other game engine like multiplayer modifications, so it should feel familiar to developers interested in modern stuff

As far as I know, only FiveM has the `ensure` command. The suggested `restart` behavior is much more common in Linux, e.g.  the `service` command works in this way. I expect that the `restart` command starts stopped services and resources.
I can't imagine that somebody wouldn't expect to start a resource using the `restart` command.

I see no reason to have two commands with almost identical behavior instead of using one command with expected name.

> Technically that isn't backwards compatible, someone may rely on the fact a resource must be running when this function or command is called.

I can't imagine a scenario when you need that. I would suggest writing a function that implements the old behavior if someone needs that. A failed restart commad forced me to use `start` in all my cases.

tryangelic:
Up

Fernando-A-Rocha (approved):
Perfect, gj

tryangelic:
This should have been merged a long time ago, pr dessa is very economical in some factors

ghost:
I'm okay with renaming the `ensure` command to `restart` if it matters in merging the pull request.

MohabCodeX:
Good job 🥳🤙🏻
I think it might be better to integrate the behavior of ensure into the existing restart command.
Users who want the behavior of ensure will likely use restart instinctively anyway.
If a script isn't running, and someone uses restart, it's usually because they want it to start so letting restart handle both running and stopped resources could reduce confusion and eliminate the need to introduce a new command

Kinimel:
GG Good job

ghost:
> I'm okay with renaming the `ensure` command to `restart` if it matters in merging the pull request.

Done. Now, the `restart` command is more Linux like.

Fernando-A-Rocha (approved):
Very well.
This will benefit everyone

Server/mods/deathmatch/logic/CConsoleCommands.cpp
```diff
@@ -60,103 +61,332 @@ static void EndConsoleOutputCapture(CClient* pClient, const SString& strIfNoOutp
...
+    if (!arguments || !arguments[0])
     {
-        pEchoClient->SendConsole("* Syntax: start <resource1> <resource2> ...");
-        return false;
+        echo->SendConsole("* Syntax: start <resource> [<resource> ...] or start <directory> [<directory> ...]");
+        return true;
     }
+     
+    std::istringstream       iss(arguments);
+    std::vector<std::string> names((std::istream_iterator<std::string>(iss)), std::istream_iterator<std::string>());
```
tederis, line 73: I would personally extract this into a dedicated function. Somewhere among SharedUtils. What do you think?

botder: We already have that functionality. Take a look at the code this change replaced.

Issue #4000, opened 2025-02-03
Arrow marker rendering bug


nweb:
### Describe the bug

Because of this, it is impossible to recreate the marker bouncing as in GTA SA (also #536).

### Steps to reproduce

```lua
local marker = createMarker(0, 0, 0, 'arrow', 1, 255, 255, 0)

addEventHandler('onClientRender', root, function()
	setElementPosition(marker, 1, 0, 3.5 + math.sin(os.clock() * 5) * 0.25)
end)
```

### Version

latest release

### Additional context

https://github.com/user-attachments/assets/acbd84f9-7b0b-49e3-897f-1e1b0458dde3

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

PR #4027, opened 2025-02-16
Fix arrow type marker snapping to the ground

### LINKED ISSUE #536: Arrow markers disappear in ground
**Describe the bug**
When placed too close to the ground, arrow markers get snapped in a really awkward and uncontrollable position, making them totally useless.

**To reproduce**
The following code (serverside or clientside, doesn't matter) creates 2 rows of markers: one row of cylinders, one row of arrows. While cylinders are placed as expected, the arrows close to the ground get moved.
```lua
function failMarkerTest()
    local x, y = 0, 0
    for z=0,10,0.5 do
        marker = createMarker(x, y, z, "arrow", 1, 255, 255, 0, 128)
        marker = createMarker(x+1, y, z, "cylinder", 1, 255, 255, 0, 128)
        y = y + 1
    end
end
addCommandHandler("failmarker", failMarkerTest)
```
**Expected behaviour**
To not get snapped to ground.

**Screenshots**
image

**MTA Client:** 
 - Version: r6985
 - Running in Wine? 

**MTA Server:** 
 - Version r6985
 - OS: 

**Additional context**

From https://bugs.mtasa.com/view.php?id=5534

FileEX:
Fixed #536 and #4000 

image


MegadreamsBE:
This is huge! I recently had to work with them (for enexes - for a certain project of mine) and it was a huge headache.

TheNormalnij:
Does it affect other marker types?

FileEX:
> Does it affect other marker types?

This shouldn't affect anything else because other markers don't have such bug

Lpsd:
As we're affecting markers here, all other marker types need to be tested. It's not enough to simply assume it won't cause other issues relating to the affected area.

I also just want to make a general point (not specifically related to this PR) that this kind of testing is also not what nightly should be intended for (it seems nightly often gets used as a testing bed due to lack of testing or review in the PR stage). 

PRs should be clear on _which_ areas they are affecting, _what_ has already been tested (e.g; have you tested for regression in related areas?), and _how_ it can be tested; it may be time to improve our pull-request template & overall flow.

If a PR has had sufficient testing/review and issues arise in nightly that's fine - but using it as an off-the-bat testing bed, explicitly due to a _lack_ of testing or review at the PR stage, is not very ideal.

FileEX:
Based on simple tests that involved creating markers and observing them, I didn't find any issues, as seen in the screenshot—markers like the 'cylinder' type behave correctly. The PR doesn't change anything except removing the Z-axis offset, which was originally intended for 'bouncing' markers for interiors (enex) in singleplayer.

botder (approved):

PR #4031, opened 2025-02-17
Revert streaming changes #4013

TheNormalnij:
#4013 requests more tests and should be reviewed before merging.

The scope of these changes is dangerous, and the streaming logic is pretty complicated.
It's impossible to bring fixes without reviewing entire logic after changes.


Dutchman101:
Alright. But then i feel like we need to revert whatever caused the performance degradations (lag spikes) at the same time, as a fix will be distant, and it's affecting too many users.

From several sources, #3978 is likely to blame, so i will follow this right up with a revert on it, after which we will ask testers to take a new look.

PR #4033, opened 2025-02-18
Fixing repeated uses of CCore::CreateXML

G-Moris:


botder (approved):
This module gets created/loaded in CCore constructor

botder:
This had to be reverted because key bindings were not loading properly and were missing.

Lpsd:
https://github.com/multitheftauto/mtasa-blue/blob/master/Client/core/CCore.cpp#L95

Move this line below the `m_pKeyBinds` instantiation if you want to remove the other `CreateXML` calls.

Alternatively, and probably a better solution to fix this spaghetti code, is to move the following logic to its own function and call it after the `m_pKeyBinds` instantiation, leaving the `CreateXML` call where it is. Maybe that logic can even go in the `CKeyBinds` constructor....

https://github.com/multitheftauto/mtasa-blue/blob/master/Client/core/CCore.cpp#L1079-L1084

Please test your work before submitting it.

Issue #2443, opened 2021-11-23
Setting aircraft vehicle health to zero causes infinite blowing


Xenius97:
**Describe the bug**
Spawn a helicopter or plane, change health to 0.
When you change health, and heli explodes onVehicleExplode && onClientVehicleExplode won't detect.
Sometimes heli keeps blowing after fixing it. + When you change health to zero sometimes it causes infinite blowing.

**Expected behaviour**
Work as another vehicles

**Screenshots**
https://youtu.be/ZBt0AzxKaao

**Version**
Multi Theft Auto v1.5.9-release-21024.1

Lpsd:
What you're seeing is intended behaviour. 

All cars/helicopters/planes etc, once set to 0 health, will have ~5 seconds before they actually "explode" and are destroyed.

In the case of cars, they will set on fire before exploding. 
However with helicopters and planes, they have an explosion effect instead of being set on fire (which is what you describe as "infinite blowing"). This is the same in singleplayer.

`blowVehicle` instantly triggers the explosion which "destroys" the vehicle.

This is why the on(Client)VehicleExplode event fires instantly when using `:blow()`, but there's a delay when setting vehicle health to 0.

Xenius97:
If you change health to zero, onVehicleExplode && onClientVehicleExplode will never trigger on helicopters/planes. I think it's bug.

Lpsd:
That's not shown either by your video, or by testing that I did myself. The event triggers fine after setting health to 0, just ~5 seconds afterwards as I've explained above.

Your own video shows the event firing after 5 seconds. At around 1:52 you set the helicopter health to 0, then at around 1:58 the event is triggered once the pre-explosion effects are finished (see debugscript output). 

Please provide new details if you're experiencing something other than what's been shown and explained to be intended behaviour.

PR #4040, opened 2025-02-19
Add new world special property (vehicleburnexplosions)

samr46:
This property toggles creation of additional explosions ([type 2 - rocket](https://wiki.multitheftauto.com/wiki/Explosion_types)) when plane or helicopter is burning (default is `true`).

By default when plane or helicopter is burning the game randomly creates additional explosions before vehicle finally blows up. It can be useful to disable these unnecessary explosions (the big dark smoke will remain and it doesn't affect normal blow up behavior).

It should also help to better filter false positives that typically occur when plane or heli is burning (onExplosion event).

> [!WARNING]
> The code includes missing "ignorefirestate" and "flyingcomponents" properties in `ResetWorldProperties`.


samr46:
Needs testing but it might also help to avoid infinite blowing described in #2443

FileEX (approved):

botder (approved):

Issue #4039, opened 2025-02-19
killPedTask instant crash


Xenius97:
### Describe the bug

![Image](https://github.com/user-attachments/assets/ccccd749-7e7d-454c-88ee-dd4d5280383f)

### Steps to reproduce

crun killPedTask(localPlayer, "primary", 4)

### Version

22987

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

FileEX:
You cannot kill a task of type ``TASK_PRIORITY_DEFAULT`` because it represents the currently executed action, such as ``TASK_SIMPLE_PLAYER_ON_FOOT``, ``TASK_SIMPLE_CAR_DRIVE``, etc.

https://github.com/multitheftauto/mtasa-blue/blob/c7d40e831a38da60670d76d496b765ad6d1466ca/Client/game_sa/CTaskManagerSA.cpp#L29

Proxy-99:
> You cannot kill a task of type `TASK_PRIORITY_DEFAULT` because it represents the currently executed action, such as `TASK_SIMPLE_PLAYER_ON_FOOT`, `TASK_SIMPLE_CAR_DRIVE`, etc.
> 
> [mtasa-blue/Client/game_sa/CTaskManagerSA.cpp](https://github.com/multitheftauto/mtasa-blue/blob/c7d40e831a38da60670d76d496b765ad6d1466ca/Client/game_sa/CTaskManagerSA.cpp#L29)
> 
> Line 29 in [c7d40e8](/multitheftauto/mtasa-blue/commit/c7d40e831a38da60670d76d496b765ad6d1466ca)
> 
>  if (iTaskPriority != TASK_PRIORITY_DEFAULT)            // TASK_PRIORITY_DEFAULT removed = crash

okay great so I will disable it and add validation

PR #4045, opened 2025-02-21
Fix killPedTask crash

Proxy-99:
Add validation
#4039


Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2481,16 +2481,25 @@ bool CLuaPedDefs::SetPedExitVehicle(CClientPed* pPed)
...
+bool CLuaPedDefs::killPedTask(CClientPed* ped, taskType taskType, std::uint8_t taskNumber, std::optional<bool> gracefully) 
 {
     switch (taskType)
     {
         case taskType::PRIMARY_TASK:
         {
+            if (taskNumber == 4)
+                throw LuaFunctionError("Killing TASK_PRIORITY_DEFAULT is not allowed");
+
+            if (taskNumber > 3)
```
Lpsd: Can we make use of the existing enums instead of using these 'magic' numbers?

https://github.com/multitheftauto/mtasa-blue/blob/master/Client/sdk/game/CTaskManager.h#L16-L35

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2481,16 +2481,25 @@ bool CLuaPedDefs::SetPedExitVehicle(CClientPed* pPed)
...
+                throw LuaFunctionError("Invalid task slot number");
+
             return ped->KillTask(taskNumber, gracefully.value_or(true)); 
         }
         case taskType::SECONDARY_TASK:
         {
+            if (taskNumber > 5)
```
Lpsd: Same here

Lpsd (dismissed):

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2481,16 +2481,25 @@ bool CLuaPedDefs::SetPedExitVehicle(CClientPed* pPed)
...
+bool CLuaPedDefs::killPedTask(CClientPed* ped, taskType taskType, std::uint8_t taskNumber, std::optional<bool> gracefully) 
 {
     switch (taskType)
     {
         case taskType::PRIMARY_TASK:
         {
+            if (taskNumber == TASK_PRIORITY_DEFAULT)
+                throw LuaFunctionError("Killing TASK_PRIORITY_DEFAULT is not allowed");
+
+            if (taskNumber > TASK_PRIORITY_MAX)
```
Lpsd:
```cpp
if (taskNumber >= TASK_PRIORITY_MAX)
```

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -2481,16 +2481,25 @@ bool CLuaPedDefs::SetPedExitVehicle(CClientPed* pPed)
...
+                throw LuaFunctionError("Invalid task slot number");
+
             return ped->KillTask(taskNumber, gracefully.value_or(true)); 
         }
         case taskType::SECONDARY_TASK:
         {
+            if (taskNumber > TASK_SECONDARY_MAX)
```
Lpsd:
```cpp
if (taskNumber >= TASK_SECONDARY_MAX)
```

Lpsd (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
Proxy-99: but there is no slot task as TASK_PRIORITY_MAX

Proxy-99: see https://wiki.multitheftauto.com/wiki/Tasks

Lpsd: Exactly, so you want to include this in the throw condition

Lpsd: Your code is currently equivalent to 
```cpp
if (taskNumber > 5)
```
Since there is no 5 (`TASK_PRIORITY_MAX`) if someone passes 5 then it won't throw with "Invalid slot task number"

Lpsd (changes requested):
Please resolve this using the suggestion in the previous review. There is a reason the `MAX` entries are defined in the enum (also the reason why I made that suggestion). 

Whilst it's unlikely the enum will change, if new entries are added then your code would be faulty and need updating. We would rather avoid that entirely (making less work for future contributors), which is why you compare to the `MAX` enum entry for such range checks.

Proxy-99:
> Please resolve this using the suggestion in the previous review. There is a reason the `MAX` entries are defined in the enum (also the reason why I made that suggestion).
> 
> Whilst it's unlikely the enum will change, if new entries are added then your code would be faulty and need updating. We would rather avoid that entirely (making less work for future contributors), which is why you compare to the `MAX` enum entry for such range checks.

done

TheNormalnij (dismissed):

Lpsd (approved):

Lpsd:
Looks good now, thanks!

botder (approved):

PR #4048, opened 2025-02-21
Refactor CWeaponSA class

FileEX:
- Refactoring of ``CWeaponSA.cpp``, ``CWeaponSA.h``, and ``CWeapon.h`` according to the latest coding guidelines and C++ standards
- Cleanup of the aforementioned files
- Completion of missing fields in the ``CWeaponSAInterface`` class and cleanup of their types


Client/game_sa/CWeaponSA.cpp
```diff
@@ -20,230 +19,83 @@
...
-        mov     ecx, dwThis
-        push    dwPedInterface
-        call    dwFunc
-    }
+    if (m_interface)
+        m_interface->Update(ped);
 }
 
-bool CWeaponSA::Fire(CEntity* pFiringEntity, CVector* pvecOrigin, CVector* pvecTarget, CEntity* pTargetEntity, CVector* pvec_1, CVector* pvec_2)
+void CWeaponSA::AddGunshell(CEntity* firingEntity, CVector* vecOrigin, CVector2D* vecDirection, float size) const
```
G-Moris: Should change 'CVector*' to 'const CVector&' and CVector2D?

Client/game_sa/CWeaponSA.h
```diff
@@ -23,58 +20,77 @@
...
+    void SetAsCurrentWeapon() override;
+
+    void Destroy() override;
+    void Remove() override;
+    void Initialize(const eWeaponType& type, std::uint32_t ammo, CPed* ped) override;
+    void Update(CPed* ped) override;
+
+    void AddGunshell(CEntity* firingEntity, CVector* vecOrigin, CVector2D* vecDirection, float size) const override;
+    void DoBulletImpact(CEntity* firingEntity, CEntitySAInterface* hitEntityInterface, CVector* vecOrigin, CVector* vecTarget, CColPoint* colPoint, int incrementalHit) const override;
+    bool Fire(CEntity* firingEntity, CVector* vecOrigin, CVector* vecEffectPos, CEntity* targetEntity, CVector* vecTarget, CVector* vecAlt) override;
```
G-Moris, line 81: Should change all 'CVector*' to 'const CVector&' or 'CVector&' and CVector2D?

FileEX: I can do it that way in ``DoBulletImpact``, but it doesn’t make much difference anyway because, in the context of asm or dynamically called functions, it’s just an address/pointer, so it comes down to the same thing

G-Moris: > I can do it that way in `DoBulletImpact`, but it doesn’t make much difference anyway because, in the context of asm or dynamically called functions, it’s just an address/pointer, so it comes down to the same thing

Then why is `CWeaponSA` refactoring necessary if we don't pay attention to it? I'm not insisting, but it looks weird.

Client/game_sa/CWeaponSA.cpp
```diff
@@ -20,230 +19,83 @@
...
-
-CPed* CWeaponSA::GetPed()
-{
-    return m_pOwner;
-}
-
-eWeaponSlot CWeaponSA::GetSlot()
-{
-    return m_weaponSlot;
+    return m_interface ? pGame->GetWeaponInfo(m_interface->m_eWeaponType, skill) : nullptr;
```
TheNormalnij, line 24: Is it possible that m_interface is null?

Client/game_sa/CWeaponSA.cpp
```diff
@@ -20,230 +19,83 @@
 
 extern CGameSA* pGame;
 
-CWeaponSA::CWeaponSA(CWeaponSAInterface* pInterface, CPed* pOwner, eWeaponSlot weaponSlot)
-    : m_pInterface{pInterface}, m_pOwner{pOwner}, m_weaponSlot{weaponSlot}
+CWeaponInfo* CWeaponSA::GetInfo(const eWeaponSkill& skill) const
```
TheNormalnij, line 22: eWeaponSkill is small type, you don't need to use references for it

Client/game_sa/CWeaponSA.cpp
```diff
@@ -20,230 +19,83 @@
...
-        call    dwFunc
-    }
+    if (!m_interface)
+        return;
+
+    m_interface->Shutdown();
 
     // If the removed weapon was the currently active weapon, switch to empty-handed
-    if (m_pOwner->GetCurrentWeaponSlot() == m_weaponSlot)
+    if (m_owner && m_owner->GetCurrentWeaponSlot() == m_weaponSlot)
```
TheNormalnij, line 49: Is possible that the owner is nullptr?

Client/game_sa/CWeaponSA.h
```diff
@@ -23,58 +20,77 @@
...
+    bool FireInstantHit(CEntity* firingEntity, CVector* vecOrigin, CVector* vecMuzzle, CEntity* targetEntity, CVector* vecTarget, CVector* vecForDriveby, bool crossHairGun, bool createGunFx) { return ((bool(__thiscall*)(CWeaponSAInterface*, CEntitySAInterface*, CVector*, CVector*, CEntitySAInterface*, CVector*, CVector*, bool, bool))FUNC_CWeapon_FireInstantHit)(this, firingEntity ? firingEntity->GetInterface() : nullptr, vecOrigin, vecMuzzle, targetEntity ? targetEntity->GetInterface() : nullptr, vecTarget, vecForDriveby, crossHairGun, createGunFx); }
+    bool GenerateDamageEvent(CPed* ped, CEntity* responsible, eWeaponType weaponType, int damagePerHit, ePedPieceTypes hitZone, std::uint8_t dir) { return ((bool(__thiscall*)(CWeaponSAInterface*, CPedSAInterface*, CEntitySAInterface*, eWeaponType, int, ePedPieceTypes, std::uint8_t))FUNC_CWeapon_GenerateDamageEvent)(this, ped ? ped->GetPedInterface() : nullptr, responsible ? responsible->GetInterface() : nullptr, weaponType, damagePerHit, hitZone, dir); }
```
TheNormalnij, line 45: It's better to not use MTA types in SAInterface methods

Client/game_sa/CWeaponSA.cpp
```diff
@@ -264,145 +116,89 @@ bool CWeaponSA::FireBullet(CEntity* pFiringEntity, const CVector& vecOrigin, con
...
-    bool bReturn = pGame->GetWorld()->ProcessLineOfSight(vecStart, vecEnd, colCollision, CollisionEntity, flags, pBuildingResult);
-    _asm
-    {
-        add esp, 10h
-    }
+    // Call CBirds::HandleGunShot
+    ((void(__cdecl*)(CVector*, CVector*))FUNC_CBirds_CheckForHit)(const_cast<CVector*>(vecStart), const_cast<CVector*>(vecEnd));
+
+    // Call CShadows::GunShotSetsOilOnFire
+    ((void(__cdecl*)(CVector*, CVector*))FUNC_CShadows_CheckForHit)(const_cast<CVector*>(vecStart), const_cast<CVector*>(vecEnd));
```
TheNormalnij, line 168: Could you define these function parameters as const instead of using const_cast?

Dutchman101: If agreed on, please address in a follow up PR, because there are schedule reasons to merge now.

Client/game_sa/CWeaponSA.h
Dutchman101: If agreed on, please address in a follow up PR, because there are schedule reasons to merge now.

Client/game_sa/CWeaponSA.cpp
Dutchman101: I've discussed the situation around "Is it possible that (x) is null?" cases with him on Discord, there are particular security reasons for additional checks that may seem redundant.

Client/game_sa/CWeaponSA.cpp
Dutchman101: If agreed on, please address in a follow up PR, because there are schedule reasons to merge now.

Client/game_sa/CWeaponSA.cpp
Dutchman101: I've discussed the situation around "Is it possible that (x) is null?" cases with him on Discord, there are particular security reasons for additional checks that may seem redundant.

Client/game_sa/CWeaponSA.cpp
TheNormalnij: These check are useless and do nothing

TheNormalnij:
This PR needs a refactor. In this state it creates more problems than it solves

PR #4049, opened 2025-02-21
The `onPlayerDamageCancelled` event for improved server-side cheat detection

ghost:
This pull request introduces a new event, called `onPlayerDamageCancelled`, which allows for reliable server-side detection of cancelled `onClientPlayerDamage` events.

There was no way to directly detect the cancellation of the `onClientPlayerDamage` event server-side. Developers had to rely on custom events for this purpose. However, custom events are much easier for cheat developers to bypass, as they can prevent the event from being sent to the server. In contrast, preventing the packet itself from being sent is far more difficult, making this new event a more reliable solution for detecting cheat related behavior, such as "god mode".

Here's an example that kicks a player if suspicious activity is detected:
```lua
addEventHandler("onPlayerDamageCancelled", root,
    function(attacker, cause, bodypart, loss)
        if getElementData(source, "admin") and getElementData(source, "godmode") then
            return
        end

        kickPlayer(source, "Yeet!")
    end
)
```

> [!IMPORTANT]
> This event cannot prevent all types of "god mode" cheats. For example, some cheats set the player's health to maximum in a given interval, instead of cancelling the event. However, you can detect such behavior with the `onPlayerDamage` event and using timers to track health changes over time.

> [!WARNING]  
> This may work with most cheats, but experienced cheat developers are able to prevent the packet from being sent to the server. Regardless, when sent, `source` will always be the player element sending the packet.


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4307,6 +4315,32 @@ void CGame::Packet_PlayerWorldSpecialProperty(CPlayerWorldSpecialPropertyPacket&
     player->CallEvent("onPlayerChangesWorldSpecialProperty", arguments, nullptr);
 }
 
+void CGame::Packet_PlayerDamageCancelled(CPlayerDamageCancelledPacket& packet) noexcept
```
FileEX: Remove noexcept

Server/mods/deathmatch/logic/packets/CPlayerDamageCancelledPacket.cpp
```diff
@@ -0,0 +1,22 @@
...
+ *  FILE:        mods/deathmatch/logic/packets/CPlayerDamageCancelledPacket.cpp
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#include "StdInc.h"
+#include "CPlayerDamageCancelledPacket.h"
+
+bool CPlayerDamageCancelledPacket::Read(NetBitStreamInterface& stream) noexcept
```
FileEX: Remove noexcept

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -4412,6 +4412,25 @@ bool CClientGame::ApplyPedDamageFromGame(eWeaponType weaponUsed, float fDamage,
...
             pDamagedPed->GetGamePlayer()->SetArmor(fPreviousArmor);
+
+            if (IS_PLAYER(pDamagedPed) && pDamagedPed->IsLocalPlayer() && g_pNet->CanServerBitStream(eBitStreamVersion::PlayerDamageCancelled))
+            {
+                if (auto stream = g_pNet->AllocateNetBitStream())
+                {
+                    if (pInflictingEntity)
+                        stream->Write(pInflictingEntity->GetID());
+                    else
+                        stream->Write(0);
```
FileEX:
```cpp
stream->Write(pInflictingEntity ? pInflictingEntity->GetID() : 0);
```

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -4412,6 +4412,25 @@ bool CClientGame::ApplyPedDamageFromGame(eWeaponType weaponUsed, float fDamage,
             }
             pDamagedPed->GetGamePlayer()->SetHealth(fPreviousHealth);
             pDamagedPed->GetGamePlayer()->SetArmor(fPreviousArmor);
+
+            if (IS_PLAYER(pDamagedPed) && pDamagedPed->IsLocalPlayer() && g_pNet->CanServerBitStream(eBitStreamVersion::PlayerDamageCancelled))
+            {
+                if (auto stream = g_pNet->AllocateNetBitStream())
```
FileEX:
```cpp
if (auto* stream = g_pNet->AllocateNetBitStream())
```

FileEX (changes requested):

FileEX:
In my opinion, this event doesn't make sense. If we want to detect event cancellations, we should do it once for all events rather than creating separate packets and events for everything. Using the same logic, one could cancel vehicle damage, ped damage, object damage, weapon shots, etc.

A potentially better (though still not ideal) solution could be an event like ``onEventCancelled``, triggered when any event is canceled on the client side. However, I believe this should be an option in mtaserver.conf because some server owners might prioritize performance and not want the server to constantly process packets about canceled events, such as decorative peds damage, weapon fire, etc.

ghost:
> In my opinion, this event doesn't make sense. If we want to detect event cancellations, we should do it once for all events rather than creating separate packets and events for everything. Using the same logic, one could cancel vehicle damage, ped damage, object damage, weapon shots, etc.
> 
> A potentially better (though still not ideal) solution could be an event like `onEventCancelled`, triggered when any event is canceled on the client side. However, I believe this should be an option in mtaserver.conf because some server owners might prioritize performance and not want the server to constantly process packets about canceled events, such as decorative peds, weapon fire, etc.

That's too complicated, because there's no need to detect cancellation of every event on the server-side, `onClientPlayerDamage` is a highly exploitable event on the client-side, which can't really be detected otherwise, except with custom events, which are quite easy for a cheater to prevent from sending, even from Lua.

starfleet001:
So if gamemode cancels this event by default (what most advanced servers do) this will add unnecessary network / CPU overhead? Is this correct?

Wouldn't it be better to make it toggleable via server config for example?
Players can receive a lot of damage and this will flood server with useless packets if gamemode cancels this event by default.

FileEX:
> 
> That's too complicated, because there's no need to detect cancellation of every event on the server-side, `onClientPlayerDamage` is a highly exploitable event on the client-side, which can't really be detected otherwise, except with custom events, which are quite easy for a cheater to prevent from sending, even from Lua.

If cheaters can cancel the onClientPlayerDamage event, they can just as easily cancel other events like onClientVehicleDamage, onClientObjectDamage, or onClientPedDamage. I understand your point that they mostly focus on themselves and possibly their vehicle, but as I mentioned earlier, I think adding a toggle option in mtaserver.conf would be a good approach. Many servers naturally cancel this event depending on the game mode.

ghost:
Then I'll add an option to mtaserver.conf @starfleet001 & @FileEX

ghost:
This pull request is being closed in favor of a better implementation. Read more: https://discord.com/channels/801330706252038164/1300970029562597457/1342576783064694786

PR #4052, opened 2025-02-23
Add ability to replace clump model with atomic and vice-versa

### LINKED ISSUE #3699: After the model replacement, the frames are not properly loaded
### Describe the bug

When we load a model that has more than one frame, all the remaining frames are ignored. Only one frame appears in the game.

image

In the game, there is only one frame; the remaining frames are not loaded.

### Steps to reproduce

1. Replace, for example, the windmill (3425), and the model will be incomplete. It's enough to load the default model—no edits are necessary.

### Version

Multi Theft Auto v1.6-release-22650

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

FileEX:
Fixed #3699 

Until now, when replacing an atomic model with a clump model, only one atomic was visible in the game. On the other hand, when replacing a clump model with an atomic model, the function would either return false or cause a crash. 

Now MTA also supports replacing clump-type objects. However, to maintain backward compatibility, models are not automatically converted based on the number of atomics. If you want to replace an atomic-type object with a clump-type object, you need to convert it using the function below

## Syntax
```lua
bool engineConvertModelToType(number modelId, string modelType)
```
- modelType: **object (atomic)** | **damageable-object** | **timed-object** | **clump**. 

Once this PR is merged, the next item on my to-do list is object animations.

image

```lua
-- ATOMIC -> CLUMP
local dff = engineLoadDFF('nt_windmill.dff');
engineConvertModelToType(1337, 'clump');

engineReplaceModel(dff, 1337);

-- CLUMP -> ATOMIC
local dff = engineLoadDFF('binnt07_la.dff');
engineConvertModelToType(3425, 'object');

engineReplaceModel(dff, 3425);
```


Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1530,6 +1532,16 @@ void CModelInfoSA::RestoreOriginalModel()
         pGame->GetStreaming()->RemoveModel(m_dwModelID);
     }
 
+    // Restore original interface if model was converted
+    if (MapContains(m_convertedModelInterfaces, m_dwModelID))
+    {
+        CBaseModelInfoSAInterface* currentInterface = ppModelInfo[m_dwModelID];
+        ppModelInfo[m_dwModelID] = MapGet(m_convertedModelInterfaces, m_dwModelID);
```
TheNormalnij, line 1552: m_convertedModelInterfaces[m_dwModelID] can have invalid state here
Invalid loading state
Invalid pointer to rwObject
Wrong references counter
..

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1799,6 +1811,67 @@ void CModelInfoSA::MakeClumpModel(ushort usBaseID)
...
+bool CModelInfoSA::ConvertToClump()
+{
+    // Get current interface
+    CAtomicModelInfoSAInterface* currentAtomicInterface = static_cast<CAtomicModelInfoSAInterface*>(ppModelInfo[m_dwModelID]);
+    if (!currentAtomicInterface)
+        return false;
+
+    // Create new clump interface
+    CClumpModelInfoSAInterface* newClumpInterface = new CClumpModelInfoSAInterface();
+    MemCpyFast(newClumpInterface, currentAtomicInterface, sizeof(CClumpModelInfoSAInterface));
```
TheNormalnij: Don't read out of buffer

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1799,6 +1811,67 @@ void CModelInfoSA::MakeClumpModel(ushort usBaseID)
     CopyStreamingInfoFromModel(usBaseID);
 }
 
+bool CModelInfoSA::ConvertToClump()
+{
+    // Get current interface
+    CAtomicModelInfoSAInterface* currentAtomicInterface = static_cast<CAtomicModelInfoSAInterface*>(ppModelInfo[m_dwModelID]);
+    if (!currentAtomicInterface)
```
TheNormalnij: Check that the current model is atomic

Client/game_sa/CModelInfoSA.h
```diff
@@ -252,9 +261,19 @@ class CClumpModelInfoSAInterface : public CBaseModelInfoSAInterface
...
+    void SetClump(RpClump* clump) { ((void(__thiscall*)(CClumpModelInfoSAInterface*, RpClump*))FUNC_CClumpModelInfo_SetClump)(this, clump); }
 };
 
-class CTimeModelInfoSAInterface : public CBaseModelInfoSAInterface
+class CAtomicModelInfoSAInterface : public CBaseModelInfoSAInterface
+{
+public:
+    void DeleteRwObject() { ((void(__thiscall*)(CAtomicModelInfoSAInterface*))FUNC_CAtomicModelInfo_DeleteRwObject)(this); }
+    void SetAtomic(RpAtomic* atomic) { ((void(__thiscall*)(CAtomicModelInfoSAInterface*, RpAtomic*))FUNC_CAtomicModelInfo_SetAtomic)(this, atomic); }
+};
```
TheNormalnij, line 284: You can remove macroses

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -496,30 +497,70 @@ bool AtomicsReplacer(RpAtomic* pAtomic, void* data)
...
-            RpClumpForAllAtomics(pCopy, AtomicsReplacer, &data);
+        char name[24];
+        bool damage = false;
+        GetNameAndDamage(GetFrameNodeName(frame), name, damage);
 
-            // Get rid of the now empty copied clump
-            RpClumpDestroy(pCopy);
-        }
+        if (damage)
+            static_cast<CDamageableModelInfoSAInterface*>(currentInterface)->SetDamagedAtomic(atomic);
```
TheNormalnij: OOB and causes random crashes if the model is atomic

Client/game_sa/CModelInfoSA.cpp
```diff
@@ -1799,6 +1811,67 @@ void CModelInfoSA::MakeClumpModel(ushort usBaseID)
...
+
+    // Set new interface for ModelInfo
+    ppModelInfo[m_dwModelID] = newClumpInterface;
+
+    // Store original interface
+    MapSet(m_convertedModelInterfaces, m_dwModelID, static_cast<CBaseModelInfoSAInterface*>(currentAtomicInterface));
+    return true;
+}
+
+bool CModelInfoSA::ConvertToAtomic()
```
TheNormalnij: Why we need that?

Client/game_sa/CModelInfoSA.cpp
FileEX: 1. How can m_convertedModelInterfaces[m_dwModelID] be invalid here?
2. The pointer to pRwObject is not relevant here because when this function is called, the game already calls DestroyRwObject and recreates it through the streamer

Client/game_sa/CModelInfoSA.cpp
FileEX: ``ConvertToAtomic``? So what do you propose?

Client/game_sa/CRenderWareSA.cpp
FileEX: I don't understand what do you mean. CDamageableAtomic inherit CAtomicModelInfo

TheNormalnij: And has an extra field. You can't safely cast CAtomicModelInfo to CDamageableAtomic and use that field.
You write out of buffer here and override random data. This causes random crashes.
This is why 'object-damageable' exists in engineRequestModel

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Old references counter in m_convertedModelInterfaces can cause memory leak or random model unloading with a crash.
There is also the pColModel pointer that can reference to unloaded custom collision

Client/game_sa/CRenderWareSA.cpp
FileEX: So I think it would be safer to simply leave SetAtomic without considering SetDamagedAtomic because it's better than checking the vtbl, right?

Client/game_sa/CModelInfoSA.cpp
TheNormalnij: Remove this function if it doesn't add a new functionality.

FileEX: This function is used to convert a clump into an atomic. Instead of replacing and removing atomics within a clump, a new CAtomicModelInfo interface can simply be created, and that's exactly what this function does to replace a clump-type model with an atomic

Client/game_sa/CRenderWareSA.cpp
TheNormalnij: Check if it works.

Client/mods/deathmatch/logic/CClientDFF.cpp
```diff
@@ -332,6 +332,14 @@ bool CClientDFF::ReplaceObjectModel(RpClump* pClump, ushort usModel, bool bAlpha
...
     CModelInfo* pModelInfo = g_pGame->GetModelInfo(usModel);
 
+    // If new model is clump or atomic then we need to convert existing model
+    // pClump->object.type is always RP_TYPE_CLUMP so check number of atomics
+    // to check if new model is clump or atomic
+    if (g_pGame->GetRenderWare()->GetClumpNumOfAtomics(pClump) > 1)
+        pModelInfo->ConvertToClump();
+    else if (g_pGame->GetRenderWare()->GetClumpNumOfAtomics(pClump) == 1)
+        pModelInfo->ConvertToAtomic();
+
```
TheNormalnij: Does it break ~vehicles and~ another model types?

TheNormalnij:
Is it possible to use a general solution?
```
engineConvertModelToType(modelId, "ped" | "vehicle" | "object" | "timed-object" | "clump" | "object-damageable")
```

It looks like automatic conversation may break some models.
And i don't think that somebody will be happy, when 'timed-object' converts to 'object' or 'clump' silently.

FileEX:
> Is it possible to use a general solution?
> 
> ```
> engineConvertModelToType(modelId, "ped" | "vehicle" | "object" | "timed-object" | "clump" | "object-damageable")
> ```
> 
> It looks like automatic conversation may break some models. And i don't think that somebody will be happy, when 'timed-object' converts to 'object' or 'clump' silintly.

Yes, you're probably right. This could be backwards incompatible in some way, and additionally, automatically 'guessing' the model type is quite problematic and results in more complex code. I think this solution is appropriate, allowing the user to define which model type should be created

FileEX:
I added engineConvertModelToType to convert the model to another type (atomic, clump, damageable-atomic, timed object). I tested this in many different variations to eliminate as many crashes as possible. I think the PR is ready, as I haven't observed any new crashes. The default behavior is the same as before, meaning only one atomic is replaced

```lua
bool engineConvertModelToType(number modelId, string newType = "object" | "timed-object" | "damageable-object" | "clump")
```

nekjutsu:
Merge this and #4555, please

PR #4058, opened 2025-03-01
Fix CVehicleSA::BlowUp crash

TheNormalnij:


Dutchman101:
Thanks, this was introduced in 61d1caf and quickly followed up on.

PR #4061, opened 2025-03-02
Refactor Int Types

G-Moris:


Shared/sdk/SharedUtil.IntTypes.h
```diff
@@ -1,38 +1,41 @@
...
     Simplification of some 'unsigned' types
 *************************************************************************/
 //                                                  VS      GCC
 //                      Actual sizes:      32bit   64bit   64bit
-typedef unsigned long  ulong;             //  32      32      64
-typedef unsigned int   uint;              //  32
-typedef unsigned short ushort;            //  16
-typedef unsigned char  uchar;             //  8
+using ulong = unsigned long;           //  32      32      64
+using uint = std::uint32_t;            //  32
```
TheNormalnij, line 19: This change doesn't look safe for x64 server

G-Moris: > This change doesn't look safe for x64 server

In general, everything is fine, if you have accurate thoughts about a possible problem, I will take note.

botder:
I am not going to merge this because it might cause unforeseen bugs. I would instead welcome PRs that replace usage of these types with types from `<cstdint>`.

PR #4041, opened 2025-02-20
Fix #4029: Expire vector display instances after SVG destruction

Lpsd:
Resolves #4029 by using the functionality in `CClientDisplayManager::DoPulse` which cleans up expired elements.

After the underlying vector graphic has been destroyed, an expiration of 1ms is set on the vector display instance so the manager can clean it up automatically.

Display instances are already destroyed/cleared upon destruction of the display manager.

Tested lightly by creating multiple SVGs, modifying them (updating, destroying) and manually tracking the lifetime of display instances in debugger. Seems to work as intended through restarting resources, disconnecting & exiting the client.

PR #4067, opened 2025-03-04
Update from lunasvg 2.3.8 to 3.2.0

Lpsd:
Whilst there are many improvements and new features in lunasvg, this PR simply updates the version from 2.3.8 to 3.2.0

We can probably make use of some new features in another PR, see release history since 2.3.8: https://github.com/sammycage/lunasvg/releases

#4041 also needs merging after this

I've attached a test resource which creates a bunch of SVGs, and allows you to manipulate them.

Start the resource, use `j` and `k` to select elements prev/next (hand icon shows currently selected element). Use F1 to randomly reposition current element, F2 to randomly resize. Use F3 to spawn a new random SVG shape.

Resource: [svg.zip](https://github.com/user-attachments/files/19077314/svg.zip)


Dutchman101:
Is there any reason to believe that this is unlike any other vendor updates (when it comes to expected stability)? Asking you, @Lpsd as you're most familiar with the SVG implementation.

Because if there's not, it can just be merged as-is and tested in a practical setting; nightlies. At this point, the test resource passes, after all.

Lpsd:
@Dutchman101 the existing behaviour should be unchanged, and if anything it will hopefully be a little faster at generation due to optimized code since the last version. It also includes support for some previously missing SVG features, and improved support for existing ones

From the release of 3.0:

> The API for rendering and rasterization—the core of what LunaSVG is most often used for—remains mostly unchanged. However, there have been significant updates to the API for document manipulation.

We'll probably want to add support for the improved document manipulation in another PR, which will make it easier for people to update their SVG data.

Dutchman101:
Ok, thanks! Then we merge it now, as we're in the process of vendor updates at large, which this is meant to be part of.

PR #4068, opened 2025-03-04
Fix model types to CHandlingManager

G-Moris:
I forgot to change std::size_t to std::uint32_t.
I also changed GetHandilgID to static.


Server/mods/deathmatch/logic/CHandlingManager.h
```diff
@@ -29,12 +29,12 @@ class CHandlingManager
...
     CHandlingEntry*       GetModelHandlingData(std::uint32_t model) const noexcept;
 
-    eHandlingTypes GetHandlingID(std::uint32_t model) const noexcept;
+    static eHandlingTypes GetHandlingID(std::uint32_t model) noexcept;
 
     // Helper functions
     eHandlingProperty GetPropertyEnumFromName(const std::string& name) const noexcept;
     bool              HasModelHandlingChanged(std::uint32_t model) const noexcept;
-    void              SetModelHandlingHasChanged(std::uint32_t model, bool bChanged) const noexcept;
+    void              SetModelHandlingHasChanged(std::uint32_t model, bool changed) const noexcept;
```
botder: A setter function shouldn't be marked `const`.

G-Moris: This function uses global variables, if you’d like, I can change it to static instead.

tederis: If it uses global variables then it has to be static for sure.

botder (dismissed):

G-Moris:
My system is broken, I'll fix it as soon as I fix it.

G-Moris:
I will take everything into account and create a new PR in the next 2-3 days (I have almost restored the system)

PR #4076, opened 2025-03-06
Fix custom weapon issue #4072

### LINKED ISSUE #4072: Bullet/Car Sync issues after v23046
### Describe the bug

Since update 23046, players massively report issue with Client-weapons (custom weapons)

If weapon hits the car, cars rotation and velocity being reset:
https://youtu.be/yh4W0ZDvSmE
https://youtu.be/zwqQlgvC8qw

When such weapons shoot in player - sometimes players gets teleported to the ocean (Can be server logic related effect).
as a bonus, car wheels dissapear :)

Words of other DayZ server owner:
`Samr46 — Today at 22:42`
`I can confirm what is on the video. However we didn't notice any far away warps or any effect with players (only vehicles seems to be affected in our case)`

### Steps to reproduce

1. create client-side weapon
2. shoot in player / car

### Version

Client: 1.6.0-9.23046.0
Server: 1.6-release-22951

### Security Policy

FileEX:
Fixes #4072 

- Clang format
- Reviews from #4048


Dutchman101:
This was requested, thanks for doing it as quickly as you did. It's tested to fix the issue.
Note to others: the patch looks bigger than it really is, due to clang format segmenting long lines.

About the 'missing' defines:

`FileEX:`
`I used these addresses directly into the CWeaponSAInterface functions instead of using #define. It's just a small cleanup while I was at it`

PR #4077, opened 2025-03-06
Remove useless checks

TheNormalnij:
CWeaponSA has no state when m_interface is nullptr. All these check are useless.

### DIFF
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -1126,7 +1126,7 @@ void CGameSA::DisableVSync()
 }
 CWeapon* CGameSA::CreateWeapon()
 {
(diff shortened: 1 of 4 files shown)


Dutchman101:
@TheNormalnij I already explained you a couple of things in private, including that we added security-related checks.

Weapon and bullet cheats (as well as "player crashers" related to weapons handling) can cause code to execute in unexpected ways, from unexpected places, causing unexpected results and abnormal states. Checks that don't make much sense to people that don't have all details can still be there for a reason, and can also be highly speculative (precautions for the unexpected). Additionally, these checks aren't expensive.

I'd like to repeat, but unfortunately more starkly and as you are public here, in public - that we have a hierarchy in the project, and the final decision is always to project management/team. I hope that you start to understand to have more respect for it. I'd appreciate it if you let it go and not re-open this, or argue further, my explanations to you previously should be enough and they are final.

PR #3656, opened 2024-08-17
CPedSA class cleanup & optimization

FileEX:
This PR is intended to clean up the CPedSA class.

- Refactoring of ``CPedSA.cpp``, ``CPedSA.h``, and ``CPed.h`` according to the latest coding guidelines and C++ standards
- Cleanup of the aforementioned files
- Completion of missing fields in the ``CPedSAInterface`` class and cleanup of their types
- Added classes for interfaces related to ped sounds
- Minor optimization by reducing unnecessary operations and redundancies.

PR #4093, opened 2025-03-12
Fix stealth kill #4092

### LINKED ISSUE #4092: Complete Knife Disability since version 23016 (Client)
### Describe the bug

The entire aspect of knife-killing (stealth kill) is completely absent.

### Steps to reproduce

1. Give yourself a knife
2. Stay behind a player standing still
3. Aim at the player using mouse2
4. Fire with mouse1
>Stealth kill doesn't happen

### Version

(Tested versions)
23077 - Affected
23051 - Affected
23024 - Affected
23022 - Client Integrity Error
23016 - Affected
23015 - Not Affected
23014 - Not Affected
23002 - Not Affected
22999 - Not Affected
22997 - Not Affected
22915 - Not Affected
22875 - Not Affected

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

FileEX:
Fixed #4092

### DIFF
--- a/Client/game_sa/CBuildingsPoolSA.cpp
+++ b/Client/game_sa/CBuildingsPoolSA.cpp
@@ -19,6 +19,7 @@
 #include "MemSA.h"
 #include "CVehicleSA.h"
 #include "CBuildingRemovalSA.h"
+#include "CPlayerPedSA.h"
 
 extern CGameSA* pGame;
 
@@ -372,7 +373,8 @@ void CBuildingsPoolSA::RemovePedsContactEnityLinks()
             ped->pLastContactedEntity[2] = nullptr;
             ped->pLastContactedEntity[3] = nullptr;
             ped->m_ucCollisionState = 0;
-            ped->pTargetedEntity = nullptr;
+
+            reinterpret_cast<CPlayerPedSA*>(pedLinks->pEntity)->SetTargetedEntity(nullptr);
         }
     }
 }
--- a/Client/game_sa/CPedSA.cpp
+++ b/Client/game_sa/CPedSA.cpp
@@ -380,20 +380,6 @@ CEntity* CPedSA::GetContactEntity() const
     return pGame->GetPools()->GetEntity(reinterpret_cast<DWORD*>(contactInterface));
 }
 
-CEntity* CPedSA::GetTargetedEntity() const
-{
-    CEntitySAInterface* targetInterface = GetPedInterface()->pTargetedEntity;
-    if (!targetInterface)
-        return nullptr;
-
-    return pGame->GetPools()->GetEntity(reinterpret_cast<DWORD*>(targetInterface));
-}
-
-void CPedSA::SetTargetedEntity(CEntity* targetEntity)
-{
-    GetPedInterface()->pTargetedEntity = targetEntity ? targetEntity->GetInterface() : nullptr;
-}
-
 void CPedSA::RemoveBodyPart(std::uint8_t boneID, std::uint8_t direction)
 {
     // char __thiscall CPed::RemoveBodyPart(CPed *this, int boneID, int localDir)
--- a/Client/game_sa/CPedSA.h
+++ b/Client/game_sa/CPedSA.h
@@ -324,7 +324,7 @@ public:
     int                              unk_75C;
     std::uint8_t                     lastWeaponDamage;
     std::uint8_t                     unk_761[3];
-    CEntitySAInterface*              pTargetedEntity;
+    CEntitySAInterface*              lastDamagedEntity;
     std::int16_t                     unk_768;
 
     CVector                          vecTurretOffset;
@@ -342,6 +342,7 @@ public:
     int                              unk_798;
 };
 static_assert(sizeof(CPedSAInterface) == 0x79C, "Invalid size for CPedSAInterface");
+
 
 class CPedSA : public virtual CPed, public virtual CPhysicalSA
 {
@@ -423,9 +424,6 @@ public:
     CEntity* GetContactEntity() const override;
 
     int GetRunState() const override { return GetPedInterface()->moveState; }
-
-    CEntity* GetTargetedEntity() const override;
-    void     SetTargetedEntity(CEntity* targetEntity) override;
 
     bool GetCanBeShotInVehicle() const override{ return GetPedInterface()->pedFlags.bCanBeShotInVehicle; }
     bool GetTestForShotInVehicle() const override { return GetPedInterface()->pedFlags.bTestForShotInVehicle; }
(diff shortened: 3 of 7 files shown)


Dutchman101:
Thanks for the quick fix

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -372,7 +373,8 @@ void CBuildingsPoolSA::RemovePedsContactEnityLinks()
             ped->pLastContactedEntity[2] = nullptr;
             ped->pLastContactedEntity[3] = nullptr;
             ped->m_ucCollisionState = 0;
-            ped->pTargetedEntity = nullptr;
+
+            reinterpret_cast<CPlayerPedSA*>(pedLinks->pEntity)->SetTargetedEntity(nullptr);
```
TheNormalnij, line 377: Possible memory bug

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -372,7 +373,8 @@ void CBuildingsPoolSA::RemovePedsContactEnityLinks()
             ped->pLastContactedEntity[2] = nullptr;
             ped->pLastContactedEntity[3] = nullptr;
             ped->m_ucCollisionState = 0;
-            ped->pTargetedEntity = nullptr;
```
TheNormalnij, line 375: This line exists to prevent crashes

Client/game_sa/CBuildingsPoolSA.cpp
FileEX: So what do you propose?

TheNormalnij: You can change reinterpret_cast to dynamic_cast and add a nullptr check.
But
This pr does something nasty and you have a problem in CPedSA or CPedSAInterface class.
Check original issue.  #3656 causes the problem. It looks like your refactor changes  field offset and pTargetedEntity points to a new memory address.

FileEX: Yes, the cause was PR #3656. The pTargetedEntity field was located at offset 0x79C, but the CPed class does not have a field at that offset because the size of CPedSAInterface is 0x79C. Additionally, there are two classes that inherit from CPed: CCopPed and CPlayerPed. The field that was used at offset 0x79C (pTargetedEntity) is actually located in the CPlayerPed class (https://github.com/gta-reversed/gta-reversed/blob/5e8a6f72ac2774ff1272078ec54f00e9b86a23d6/source/game_sa/Entity/Ped/PlayerPed.h#L20).

Therefore, to maintain consistency between CPedSAInterface and the original CPed, the targetedEntity field should be placed in CPlayerPed, just like in the original SA code. So I don't understand what is "nasty" about this.

TheNormalnij: Yeah, you are right.

You should fix reinterpret_cast only. You can't cast CPedSA to CPlayerPedSA and call a virtual method without checks.

Client/game_sa/CBuildingsPoolSA.cpp
TheNormalnij: if pTragetEntity is lastDamagedEntity now, you should replace it here too.
This code should remove links in lastDamagedEntity and mouseTargetEntity
lastDamagedEntity causes crashes
I didn't notice that you renamed that field too

Client/game_sa/CBuildingsPoolSA.cpp
TheNormalnij: Hmm, no you aren't right. You changed fields offsets in your refactor.
Your new struct ends with:
```c++
    void*                            pCoverPoint;            // CCoverPoint*
    void*                            pEnex;                  // CEntryExit*
    float                            fRemovalDistMultiplier;
    std::int16_t                     specialModelIndex; // StreamedScriptBrainToLoad

    std::int16_t                     unk_796;
    int                              unk_798;
```

reverse-modern ends with
```c++
    CCoverPoint*        m_pCoverPoint;
    CEntryExit*         m_pEnex; // CEnEx *
    float               m_fRemovalDistMultiplier;
    int16               m_StreamedScriptBrainToLoad;
    int32               field_798;
 ```

You removed a fields in the middle of CPedSAInterface and compensated with a new in the end of the class.
The class is broken in this state.
So your fix implements a new behavior based on another field while your needed fix CPedSAInterface

FileEX: I don't understand what you're talking about. The previous offsets and structure were incorrect, and now they are correct and match the actual implementation. I haven't removed any fields from the middle of the class, so I don't know what you mean.

If you're referring to some parts of the code using the wrong fields after this change, it means they were already using the wrong fields before, like how instead of setting the bChosenWeapon field, the sprintEnergy field was being set (I think that was in CPlayerDataSAInterface?).

The bugs that arise from this are quickly fixed, so I don’t understand your approach. Would you rather keep an incorrect, incomplete class with incorrect offsets just so that incorrectly mapped fields continue to be used in various parts of the code?

image

Client/game_sa/CBuildingsPoolSA.cpp
FileEX: What? Until now, pTargetedEntity pointed to offset 0x79C, and it still does, so I don’t understand what lastDamagedEntity has to do with this.

Your implementation is from June 2, 2024, while my PR was merged two weeks ago. So pTargetedEntity was pointing to offset 0x79C, and it still does—just in CPlayerPedSAInterface now.

pTargetedEntity isn't lastDamagedEntity now. That was my mistake in that PR (which even caused the stealth kill bug since it was reading lastDamagedEntity instead of pTargetedEntity).

pTargetedEntity is exactly the same as it was before my PR and still points to the same offset as before

TheNormalnij: Look at your refactor PR closelly.
You renamed pTargetedEntity to pLookAtEntity there and added a new field with name pTargetedEntity

TheNormalnij: What a mess here.
So, after refactor it should be `ped->pLookAtEntity = nullptr`

TheNormalnij: And your GetTargetedEntity should reuse this field, to restore old behavior.
Now it uses something related to mouse.

FileEX: Old CPedSAInterface
image

New CPedSAInterface
image

New CPlayerPedSA
image

The field named ``pTargetedEntity`` has always pointed to the same offset 🤔

TheNormalnij: ah, ok, i didn't notice pad[104] there, sorry

Issue #3681, opened 2024-08-26
Add setElementCollidableWith server-side


Proxy-99:
### Is your feature request related to a problem? Please describe.

I want to make the calls by the server not the client

### Describe the solution you'd like

add new function 

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

PR #4101, opened 2025-03-14
Add setElementCollidableWith server-side

Proxy-99:
#3681

test code
```lua
vehicle1 = createVehicle ( 411, 2044.65942, 1526.35352, 10.67188)
object = createObject( 3578, 2044.40405, 1510.24036, 10.67188)

setTimer(function()
  res = setElementCollidableWith(vehicle1,object,false)
  print(res)
end, 2000, 1)

```


Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
```diff
@@ -765,3 +766,15 @@ void CElementRPCs::SetElementOnFire(CClientEntity* pSource, NetBitStreamInterfac
 {
     pSource->SetOnFire(bitStream.ReadBit());
 }
+
+void CElementRPCs::SetElementCollidableWith(CClientEntity* pSource, NetBitStreamInterface& bitStream)
+{
+    ElementID ElementID;
+    bool      bCollidable;
```
FileEX: Hungarian notation

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
```diff
@@ -765,3 +766,15 @@ void CElementRPCs::SetElementOnFire(CClientEntity* pSource, NetBitStreamInterfac
...
+
+    bitStream.Read(ElementID);
+    bitStream.ReadBit(bCollidable);
```
FileEX: bitStream.Read can return false

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
```diff
@@ -765,3 +766,15 @@ void CElementRPCs::SetElementOnFire(CClientEntity* pSource, NetBitStreamInterfac
...
+
+   CClientEntity* pCollidableWith = CElementIDs::GetElement(ElementID);
```
FileEX: Hungarian notation + GetElement can return nullptr

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
```diff
@@ -765,3 +766,15 @@ void CElementRPCs::SetElementOnFire(CClientEntity* pSource, NetBitStreamInterfac
...
+   pSource->SetCollidableWith(pCollidableWith, bCollidable);
```
FileEX:
```cpp
pSource->SetCollidableWith(pCollidableWith, bitStream.ReadBit());
```

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -12566,3 +12565,39 @@ bool CStaticFunctionDefinitions::SpawnVehicleFlyingComponent(CVehicle* const veh
 
     return true;
 }
+
+bool CStaticFunctionDefinitions::SetElementCollidableWith(CElement* element, CElement* withElement, bool canCollide)
+    {
```
FileEX, line 12675: Indention

FileEX (changes requested):

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
TheNormalnij: It cannot. The protocol guarantees that the message is fully received.

TheNormalnij:
The function doesn't work for a new  connected players and after reconnect

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
```diff
@@ -765,3 +766,19 @@ void CElementRPCs::SetElementOnFire(CClientEntity* pSource, NetBitStreamInterfac
...
+
+      if (!bitStream.Can(eBitStreamVersion::SetElementCollidableWith_Serverside))
+        return;
+
+    bitStream.Read(ElementID);
+
+   CClientEntity* collidableWith = CElementIDs::GetElement(ElementID);
+
+   if (collidableWith == nullptr)            // validity check  
+       return;
```
FileEX, line 872:
```cpp
if (!collidableWith)
    return;
```

FileEX (changes requested):

Client/mods/deathmatch/logic/rpc/CElementRPCs.cpp
FileEX: > It cannot. The protocol guarantees that the message is fully received.

So what is the purpose of this function being a bool, and why does most of the code check if it returns true?

TheNormalnij: Code == documentation
It has sense on the server side. The client reads server packets only.
Checks on client could be used to add a new data to a packet without increasing bitstream version, i think

Proxy-99:
> The function doesn't work for a new connected players and after reconnect
 
What packet file should I write the info?
is it   `Server\mods\deathmatch\logic\packets\CPlayerJoinCompletePacket.cpp`

FileEX:
> > The function doesn't work for a new connected players and after reconnect
> 
> What packet file should I write the info? is it `Server\mods\deathmatch\logic\packets\CPlayerJoinCompletePacket.cpp`

I think you should store informations on the server side and then send it in the CMapInfoPacket. Check how it's done, for example, with setJetpackWeaponEnabled (CGame::SetJetpackWeaponEnabled & CGame::GetJetpackWeaponEnabled)

PR #4096, opened 2025-03-12
Add CPU affinity setting & fix process related settings

FileEX:
Resolved #4088 

This PR adds a checkbox that allows setting CPU affinity to either CPU 0 or, by default, to all cores. By default, the checkbox is checked, meaning the affinity is set to CPU 0.

Additionally, this PR fixes a bug where process-related settings, such as process priority, were not applied upon starting MTA but only after entering the settings.

image

PR #4110, opened 2025-03-16
Reimplement CPU Affinity

### LINKED ISSUE #4088: CPU 0 affinity

The problem is very known since the game is old and was designed to not run on multi-threaded platform, if you disable CPU 0 affinity the game will stop having FPS-drops and lags.

I've tried to implement this which is working as expected, however if I build the project on my own I am not allowed anymore to join normal servers, so I thought that MTA developers could implement this as a toggable option from the settings menu. I'll leave the code in the solution description.

/* Disable CPU 0 affinity to let the game be more responsive and fix the known issue on single-threade games
by eP!toxiic
*/

BOOLEAN affinity = SetProcessAffinityMask(GetCurrentProcess(), 1 << 0); // 0 = CORE
if (!affinity)
{
    std::wstring message = L"Could not set process affinity.";
    DisplayErrorMessageBox(MakeLauncherError(message), L"Error");
    return;
}

_No response_

### Additional context

_No response_

### Security Policy

FileEX:
Better version of #4096 
Resolved #4088 

- Added information that the setting is experimental when hovering over it
image

- Added a warning when enabling the setting, indicating that it is experimental
image

- The setting is now disabled by default

I moved the checkbox for saving screenshots in userfiles to the multiplayer tab (imo, this is a more appropriate place for this setting than the advanced tab). The advanced tab no longer has space, which caused the informational text at the bottom of the window (the one that appears when hovering over various elements) to be cut off and not visible.

### DIFF
--- a/Client/core/CClientVariables.cpp
+++ b/Client/core/CClientVariables.cpp
@@ -357,7 +357,7 @@ void CClientVariables::LoadDefaults()
     DEFAULT("discord_rpc_share_data", false);                                         // Consistent Rich Presence data sharing
     DEFAULT("discord_rpc_share_data_firsttime", false);                               // Display the user data sharing consent dialog box - for the first time
     DEFAULT("browser_enable_gpu", true);                                              // Enable GPU in CEF? (allows stuff like WebGL to function)
-    DEFAULT("process_cpu_affinity", true);                                            // Set CPU 0 affinity to improve game performance and fix the known issue in single-threaded games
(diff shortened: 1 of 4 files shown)


Dutchman101:
The exact implementation in this PR was requested, because to be clear, certain system configurations (thought to be CPU) it gives heavy input lag and general system instability if using CPU affinity 0. An example of this are certain modern AMD Ryzen (Mobile) processors.

Thanks!

ghost:
Fixes: #4119

PR #4112, opened 2025-03-17
Improved checks for vehicle in & out packet

ghost:


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2973,7 +2973,7 @@ void CGame::Packet_Vehicle_InOut(CVehicleInOutPacket& Packet)
             // Grab the ped with the chosen ID
             ElementID PedID = Packet.GetPedID();
             CElement* pPedElement = CElementIDs::GetElement(PedID);
-            if (pPedElement && IS_PED(pPedElement))
+            if (pPedElement && IS_PED(pPedElement) && (pPedElement == pPlayer || !IS_PLAYER(pPedElement)))
```
Lpsd, line 2976: Did you mean to use `||` instead of `&&`? This change also makes the `|| !IS_PLAYER(pPedElement)` part redundant.

If `IS_PED(pPedElement)` is true, then surely `pPedElement == pPlayer` can't be true

Lpsd: Overall I think you want `if (pPedElement && IS_PED(pPedElement) || (pPedElement == pPlayer)` ?

ghost: It can be true, because IS_PED also returns true if it's a CPlayer, not only if it's a CPed

Lpsd: Oh okay, then the condition makes sense. Thanks for confirming

Lpsd (approved):

ghost:
#4129

PR #4119, opened 2025-03-22
Fix CPU affinity

### LINKED ISSUE #4118: CPU 0 Affinity implemented .. wrongly!
### Describe the bug

Hello, 2 weeks ago I requested the implementation of an option to automatically disable CPU 0 affinity, and I’m very happy that you added it.

However, instead of disabling only CPU 0, the game is currently setting affinity to only CPU 0, disabling all other cores. This makes the game unplayable (I believe this was my mistake, as I likely provided the wrong code I used for testing)

The correct implementation should be:

```
HANDLE process = GetCurrentProcess();
DWORD_PTR mask, sysMask;

if (GetProcessAffinityMask(process, &mask, &sysMask) && mask > 1)  
    SetProcessAffinityMask(process, mask & ~1); // Disable CPU 0
```

I've included an example of how it should work correctly. This should be the result.

Image

Thank you.

### Steps to reproduce

Go Options->Advanced and enable CPU 0 affinity feature, than open task manager and check it yourself.

### Version

_No response_

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

ghost:
Fixes #4118

### DIFF
--- a/.github/workflows/crowdin-auto-merge.yml
+++ b/.github/workflows/crowdin-auto-merge.yml
@@ -13,7 +13,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          fetch-depth: 0
+          token: ${{ secrets.POT_CI_PAT }}
 
       - name: Setup Git
         run: |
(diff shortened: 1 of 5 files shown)


Dutchman101:
Nice, and the reported issues make sense now.

PR #4114, opened 2025-03-20
Fix vehicle upgrades for assigned vehicles

ghost:
Fixes #1613, #2198
Successor of #3261

PR #4128, opened 2025-03-26
Fix vehicle upgrades after #4114

ghost:
Fixes vehicle upgrades


tederis:
Could you elaborate on what exactly it fixes?

ghost:
> Could you elaborate on what exactly it fixes?

Everything, because the previous pull request broke it all

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -42,115 +38,99 @@ static const SFixedArray<SUpgradeName, 17> UpgradeNames = {{{"Hood"},
...
 bool CVehicleUpgrades::IsUpgrade(unsigned short usModel)
 {
     return (usModel >= 1000 && usModel <= 1193);
 }
 
-bool CVehicleUpgrades::IsUpgradeCompatible(const std::uint16_t upgrade)
+bool CVehicleUpgrades::IsUpgradeCompatible(const std::uint16_t upgrade) noexcept
 {
-    eClientVehicleType type = m_pVehicle->GetVehicleType();
+    const auto type = m_pVehicle->GetVehicleType();
```
tederis, line 133: It's better to avoid replacing small types with `auto`.

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -167,31 +147,24 @@ bool CVehicleUpgrades::IsUpgradeCompatible(const std::uint16_t upgrade)
         return false;
     }
 
-    std::uint16_t model = m_pVehicle->GetModel();
```
tederis, line 170: It's better to avoid replacing small types with `auto`.

Client/mods/deathmatch/logic/CVehicleUpgrades.cpp
```diff
@@ -167,31 +147,24 @@ bool CVehicleUpgrades::IsUpgradeCompatible(const std::uint16_t upgrade)
...
-    }
-    catch (const std::out_of_range&)
+    if (it == m_slots.end())
     {
         return false;
     }
+
+    const auto& set = it->second;
+
+    return set.find(upgrade) != set.end();
```
tederis, line 167: What do you think about `it->second.count(upgrade) > 0`;

tederis:
Why do we need `m_slots` for every spawned vehicle? Couldn't it be global?

FileEX:
> > Could you elaborate on what exactly it fixes?
> 
> Everything, because the previous pull request broke it all

So you opened a PR without even testing its changes or checking if it works at all? 👀

tederis:
> > > Could you elaborate on what exactly it fixes?
> > 
> > 
> > Everything, because the previous pull request broke it all
> 
> So you opened a PR without even testing its changes or checking if it works at all? 👀

Why do I need that? Before any testing I want to understand the logic of PR.

ghost:
> So you opened a PR without even testing its changes or checking if it works at all? 👀

Mistake on my end, sorry. I did it as a favour, and I was in a rush, I just implemented the same logic with the old reviews.

ghost:
> Why do we need `m_slots` for every spawned vehicle? Couldn't it be global?

Thx for the quick review, I'll solve the issues soon.

Dutchman101:
> So you opened a PR without even testing its changes or checking if it works at all? 👀

He adopted the original PR (#3261) just to comply with the last pending code reviews from TEDERIs it had. It should've been matured at that point. Btw, even after #4114 (by nightly testing) i can't really find issues, but no doubt that Nico found them due to this here.

It's not a big deal, no recent nightly was released to users.

Dutchman101:
#4114 is now reverted, so that we can get stable master again

@Nico8340 please apply this fix on top of reintroducing what was reverted

PR #4140, opened 2025-04-04
CPU Affinity description

FileEX:
I'm opening this PR to set the CPU affinity to enabled by default and to change the description of this setting


tederis (approved):

Dutchman101:
I should add that this was requested after it became clear that the affinity tweak is one of the biggest performance enhancers MTA has received in recent years. It eliminates various types of microstutter-freezes, and makes the game smoother overall.

This is what happens with older, single-core games running on modern, multi-thread, multi-core CPU's.

We will present this as a large achievement during the MTA 1.6.1 announcement. I'd like to extend special thanks to @toxiicdev for the original findings (it began as a suggestion).

PR #4141, opened 2025-04-04
Fix texture destroy to TxdForceUnload

G-Moris:
Fix for double deletion of an element


tederis:
There's no double deletion. `RwTextureDestroy` internally decreases refs count. The existing code decreases refs by one at a time and deletes a texture when the counter gets zeroed.

G-Moris:
During ForceUnload testing, a failure occurred in this part of the code.
If I can find a proper solution that accounts for all the edge cases, I might open a new PR.

PR #4145, opened 2025-04-06
Fix SetPedAimTarget from generating NaN's.

### LINKED ISSUE #492: Peds shooting incorrectly
**Describe the bug**
Peds are shooting away from the player while they're not supposed to (aimed directly at player's coordinates). This can be "fixed" by alt+tabbing out of SA and back in. A sample script to reproduce the bug included.

**To Reproduce**
[peds.zip](https://github.com/multitheftauto/mtasa-blue/files/2403084/peds.zip)

I've simplified his test resource, just do `/peds` and the ped will start shooting at you, but their aim is completely wrong, I added a line between start aim and start end to prove he's meant to be aiming there but it's like he's shooting at you, but isn't.

I've also experienced this problem when making peds that are meant to shoot, this bug basically makes it impossible to have a game mode that relies on fighting peds.

Edit: I've been working with peds again and keep getting this bug so I debugged it some more and turns out that what it's meant to be aiming at is completely wrong to what it should be aiming at, the X Y and or Z randomly become NaN which is why they're usually aiming direction is all wrong. It doesn't happen if stood still but as soon as you start moving the XYZ start becoming NaN.

I was able to successfully mitigate this bug by having the ped stop shooting for 100ms every 1000ms. After he stops shooting, the XYZ stops being a NaN until it happens again in a few seconds time.

**Expected behavior**
To aim at the desired target.

**MTA Client:** 1.5.3

**MTA Server:** 1.5.3

**Additional context**
From https://bugs.mtasa.com/view.php?id=4504

Ambassador64:
- Fix #492 aim direction and collision vector

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -1016,12 +1016,66 @@ void CClientPed::SetTargetTarget(unsigned long ulDelay, const CVector& vecSource
         float fRadius = DistanceBetweenPoints3D(m_vecTargetSource, m_vecTargetTarget);
 
         // Grab the angle of the source vector and the angle of the target vector relative to the source vector that applies
-        m_vecBeginTargetAngle.fX = acos((m_vecBeginTarget.fX - m_vecBeginSource.fX) / fRadius);
-        m_vecBeginTargetAngle.fY = acos((m_vecBeginTarget.fY - m_vecBeginSource.fY) / fRadius);
-        m_vecBeginTargetAngle.fZ = acos((m_vecBeginTarget.fZ - m_vecBeginSource.fZ) / fRadius);
-        m_vecTargetTargetAngle.fX = acos((m_vecTargetTarget.fX - m_vecTargetSource.fX) / fRadius);
-        m_vecTargetTargetAngle.fY = acos((m_vecTargetTarget.fY - m_vecTargetSource.fY) / fRadius);
-        m_vecTargetTargetAngle.fZ = acos((m_vecTargetTarget.fZ - m_vecTargetSource.fZ) / fRadius);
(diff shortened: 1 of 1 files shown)


Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -1016,12 +1016,66 @@ void CClientPed::SetTargetTarget(unsigned long ulDelay, const CVector& vecSource
         float fRadius = DistanceBetweenPoints3D(m_vecTargetSource, m_vecTargetTarget);
 
         // Grab the angle of the source vector and the angle of the target vector relative to the source vector that applies
-        m_vecBeginTargetAngle.fX = acos((m_vecBeginTarget.fX - m_vecBeginSource.fX) / fRadius);
-        m_vecBeginTargetAngle.fY = acos((m_vecBeginTarget.fY - m_vecBeginSource.fY) / fRadius);
-        m_vecBeginTargetAngle.fZ = acos((m_vecBeginTarget.fZ - m_vecBeginSource.fZ) / fRadius);
-        m_vecTargetTargetAngle.fX = acos((m_vecTargetTarget.fX - m_vecTargetSource.fX) / fRadius);
-        m_vecTargetTargetAngle.fY = acos((m_vecTargetTarget.fY - m_vecTargetSource.fY) / fRadius);
-        m_vecTargetTargetAngle.fZ = acos((m_vecTargetTarget.fZ - m_vecTargetSource.fZ) / fRadius);
+        m_vecBeginTargetAngle.fX = (m_vecBeginTarget.fX - m_vecBeginSource.fX) / fRadius;
```
tederis: `m_vecBeginTargetAngle.fX = std::acosf(Clamp(-1.0f, (m_vecBeginTarget.fX - m_vecBeginSource.fX) / fRadius, 1.0f));`
An so on.

tederis (approved):

PR #4149, opened 2025-04-08
Fix invisible weapon particles at high FPS

FileEX:
This PR fixes the bug reported on the old (now unavailable) GitHub project called "Framerate issues." At high FPS (> 74), weapon particles such as the spraycan, fire extinguisher, or flamethrower were either invisible or very rare and sporadic.


Dutchman101:
That's great, thanks

PR #4154, opened 2025-04-10
Final patch for game launch failure

ghost:
Finally patches the short lived, but really annoying bug which prevented the game from launching when it's being installed for the first time


MegadreamsBE (approved):
Look good to me

PR #4178, opened 2025-04-26
Fix animation list and add some small checks in game_sa

ghost:
Read the title :)


Dutchman101:
This was requested (and tested), thanks!

Issue #4165, opened 2025-04-17
Players are visible in all dimensions


ghost:
### Describe the bug

Players are sometimes visible in all dimensions, we haven't found a correlation as to when this happens. If I go to the dimension they're in, I can hit them and they'll get the damage, but if I go to another dimension, they won't get damaged.

### Steps to reproduce

We haven't figured out how to reproduce it, sometimes it happens randomly.

### Version

Client: v1.6-release-23147 (Windows 64-bit)
Server: v1.6-release-22951 (Ubuntu 64-bit)

### Additional context

https://imgur.com/a/P3vYkcZ

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

ghost:
In reality, the player shown in the video was in the 42th dimension. But for other players, they were seen in every dimension.

tederis:
This can be related to #1245

Xenius97:
We're noticed this issue too, sometimes player saw others who aren't in their dimension.

ArranTuna:
FileEX's PR to fix this was merged.

PR #4184, opened 2025-04-28
Fix elements sometimes being visible from other dimensions in the current dimension

### LINKED ISSUE #1245: Vehicles from other dimensions may appear in the current one
**Describe the bug**
In some situations vehicles from other dimensions may appear in the current one. The series of calls setElementDimension along with some specific conditions is resulting in breaking of streaming out process. The more complex(like more vertices or bigger size of .txds) a model the more likely the bug appears. But it's only when models don't exist in the model cache.

**To reproduce**
Steps to reproduce the behaviour:
1. **Restart the game** to clean up the model cache.
2. Download and run the resource **testcase_stream**
3. Type the command **_begintest_**
4. You will warp to the airport
5. You can probably see the bug

[testcase_stream.zip](https://github.com/multitheftauto/mtasa-blue/files/4171034/testcase_stream.zip)

**Expected behaviour**
Vehicles from other dimension should not appear in the current one(except _-1_ dimension)

**Screenshots**
mta-screen_2020-02-07_20-35-41
**Version**
Multi Theft Auto v1.5.7-release-20399

FileEX:
Fixes #1245 (and maybe #4165?)

The bug is that the player's dimension changes after one second, but during that time not all vehicles have been streamed-in yet. As a result, some vehicles get streamed while the player is already in the new dimension. If for example, increase the delay in the test script attached to the mentioned issue, the problem no longer occurs because all vehicles have already been streamed.
The solution is to cancel the streaming of an element if the element's dimension is different from its streamer's dimension (e.g., vehicle streamer)

### DIFF
--- a/Client/mods/deathmatch/logic/CClientStreamElement.cpp
+++ b/Client/mods/deathmatch/logic/CClientStreamElement.cpp
@@ -61,7 +61,7 @@ void CClientStreamElement::InternalStreamIn(bool bInstantly)
     }
 }
 
-void CClientStreamElement::InternalStreamOut()
+void CClientStreamElement::InternalStreamOut(bool ignoreSendingEvent)
 {
     if (m_bStreamedIn)
     {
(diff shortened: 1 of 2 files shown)


ghost:
This pull request breaks the -1 dimension for client-sided objects

ghost:
cc @FileEX

Issue #4037, opened 2025-02-19
Crash when player is on fire


haron4igg:
### Describe the bug

My players report massive issue since recent mta updates, causing crash when player is on fire and dying.

[client_1_6_release_22977_0_000_gtasa_00171a00_5_DNM_C269DE53_55F3.dmp.zip](https://github.com/user-attachments/files/18863263/client_1_6_release_22977_0_000_gtasa_00171a00_5_DNM_C269DE53_55F3.dmp.zip)
[client_1_6_release_22977_0_000_gtasa_00139405_5_DNM_5874DE53_55F3.dmp.zip](https://github.com/user-attachments/files/18863262/client_1_6_release_22977_0_000_gtasa_00139405_5_DNM_5874DE53_55F3.dmp.zip)

### Steps to reproduce

Ignite your-self till death

### Additional context

** -- Unhandled exception -- **

Version = 1.6-release-22977.0.000
Time = Wed Feb 19 10:55:54 2025
Module = C:\ProgramData\MTA San Andreas All\1.6\GTA San Andreas\gta_sa.exe
Code = 0xC0000005
Offset = 0x00171A00

EAX=00B9BAF0  EBX=0044215C  ECX=00000000  EDX=00B9BAE8  ESI=246FA018
EDI=246FA5A4  EBP=00B7CD98  ESP=0177F3C0  EIP=00571A00  FLG=00010293
CS=0023   DS=002B  SS=002B  ES=002B   FS=0053  GS=002B

** -- End of unhandled exception -- **

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

botder:
I can't reproduce this.

1. How did you ignite your player?
2. Does it happen with a single player?
3. What is the player-on-fire doing when it crashes?
4. Is the player in a vehicle?

botder:
I figured out one way: `run me.onFire = true ; me.health = 0 ; me.health = 100`

TheNormalnij:
If the crash is introduced by #3930, we can use an alternative way to fix the bug without assembly hooks.

botder:
I am going to rollback the changes, but besides placing two hooks in `CFire::ProcessFire` around `CPlayerPed::DoStuffToGoOnFire`, what did you have on your mind, @TheNormalnij?

TheNormalnij:
[...]

PR #4188, opened 2025-05-02
Fix setPedOnFire(ped, false) doesn't cancel TASK_SIMPLE_PLAYER_ON_FIRE (again)

### LINKED ISSUE #3249: setPedOnFire(ped, false) doesn't cancel TASK_SIMPLE_PLAYER_ON_FIRE

FileEX:
Fixes #3249 

The previous PR #3930 was reverted due to an alleged crash, which later turned out not to be caused by PR #3930. See issue #4037


Dutchman101:
> The previous PR #3930 was reverted due to an alleged crash, which later turned out not to be caused by PR #3930. See issue #4037

We'll see about that in the upcoming nightly (to test)

Dutchman101:
> > The previous PR #3930 was reverted due to an alleged crash, which later turned out not to be caused by PR #3930. See issue #4037
> 
> We'll see about that in the upcoming nightly (to test)

@FileEX
Awesome! Look at this:
error

(It will be reverted again..)
869d739

PR #4195, opened 2025-05-04
Fix ignorefirestate behaviour

FileEX:
This PR fixes a bug where, upon entering fire while aiming, the crosshair would reset its size, the character's hand would briefly shake, and the chainsaw would turn off if entering the fire while holding LMB with the chainsaw


tederis (approved):
Tested

PR #4199, opened 2025-05-05
Revert "Fix dimension -1 after #4184"

tederis:
#4198 was merged by mistake. It should be reimplemented due to the invalid comparison between `unsigned short` and `signed integer`. `m_bVisibleInAllDimensions` value has to be taken into account instead.


psydmg:
i have downloaded update with this fix, currently it's shows as Multi Theft Auto v1.6-release-23182.
the problem seems not solved.

at least Multi Theft Auto v1.6-release-23147 doesn't have such problem and works as intended.

`<object interior="0" dimension="-1" alpha="255" model="14783" posX="616.17188" posY="-76.42969" posZ="998.96094" rotX="0" rotY="0" rotZ="0"></object>`

and after warp to this location with the object - it's pitch black inside and has no collision.

Dutchman101:
Superseded by c68f458

PR #4201, opened 2025-05-06
Streaming fix

tederis:
The reimplementation of #4184(thanks to @FileEX) and includes fix #4198(thanks to @Nico8340) with changes which make the whole thing a little bit more performant(it prevents models from loading instead of unloading them straight after the load).


FileEX (approved):

ghost (approved):

Dutchman101:
Thanks! This will be merged in one of the next rounds of builds that are intended for extensive testing. Streamer is always a sensitive part to modify, as we also just got reminded of.

(Please don't merge it before i do)

Dutchman101:
I've requested a review from TheNormalnij as well, he has experience with streamer stuff.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -4041,6 +4041,13 @@ void CClientPed::ReCreateGameEntity()
 
 void CClientPed::ModelRequestCallback(CModelInfo* pModelInfo)
 {
+    // The model loading may take a while and there's a chance of ped being moved to other dimension.
+    if (!IsVisibleInAllDimensions() && GetDimension() != m_pStreamer->GetDimension())
```
TheNormalnij, line 4045:
```c++
if (!ShouldBeVisibleInCurrentDimension())
```
And you can keep `m_pStreamer` private

TheNormalnij:
The logic looks correct

tryangelic:
Up

Client/mods/deathmatch/logic/CClientPed.cpp
tederis: I don't think that a dedicated method is appropriate here. `m_pStreamer` is absolutely okay here. It had to be **protected** originally.

botder (approved):

PR #4203, opened 2025-05-06
Fix bone related functions

ghost:
Fixes functions related to bones, until now, 201, 301 and 302 were not useable. Also refactors the mentioned functions.


Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -995,110 +995,147 @@ int CLuaPedDefs::CanPedBeKnockedOffBike(lua_State* luaVM)
     return 1;
 }
 
-bool CLuaPedDefs::SetElementBonePosition(lua_State* const luaVM, CClientPed* entity, std::uint32_t boneId, CVector position)
+std::variant<bool, CLuaMultiReturn<float, float, float>> CLuaPedDefs::GetElementBonePosition(CClientPed* ped, const std::uint16_t bone)
 {
-    CEntity* theEntity = entity->GetGameEntity();
-    if (!theEntity)
+    if (bone < BONE_ROOT || bone > BONE_LEFTBREAST)
+        throw std::invalid_argument("Invalid bone: " + std::to_string(bone));
```
tederis, line 1001: The exceptions make functions backward incompatible. Because the existing code can rely on return value, but exceptions will interrupt the execution and potentially break the code. Maybe better to use `ArgumentParserWarn<false, ...>`?

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -995,110 +995,147 @@ int CLuaPedDefs::CanPedBeKnockedOffBike(lua_State* luaVM)
...
+
+    CEntity* entity = ped->GetGameEntity();
+    CVector position;
+
+    if (!entity || !entity->GetBonePosition(static_cast<eBone>(bone), position))
         return false;
-    return theEntity->SetBonePosition(static_cast<eBone>(boneId), position);
+
+    return std::make_tuple(position.fX, position.fY, position.fZ);
```
FileEX: If you used CLuaMultiReturn in the function definition in the variant, then why are you returning a plain std::tuple here instead of CLuaMultiReturn? Even though CLuaMultiReturn is std::tuple, the return type should match the definition

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
ghost: Thanks!

Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
ghost: Fixed

botder (approved):

PR #4206, opened 2025-05-07
Potential crash fix

### LINKED ISSUE #4202: Crash 0x00012AE8
### Describe the bug

Hi! Discovered one crash, i assume this is one happens pretty often now to my players (approx last month)

```
Version = 1.6-release-23183.0.000
Time = Tue May  6 14:39:34 2025
Module = C:\Games\NS-DayZ\Multi Theft Auto1.6\mta\game_sa.dll
Code = 0xC0000005
Offset = 0x00012AE8

EAX=B12A1588  EBX=25581198  ECX=B12A1588  EDX=25581198  ESI=00000022
EDI=00000000  EBP=0177FCA0  ESP=0177FC94  EIP=5AD22AE8  FLG=00010246
```

[client_1.6-release-23183.0.000_gamesa_00012ae8_5_CGNM_C269DE53_55F3_0F8_A6EA6_20250506_1439.rsa.dmp](https://github.com/user-attachments/files/20062197/client_1.6-release-23183.0.000_gamesa_00012ae8_5_CGNM_C269DE53_55F3_0F8_A6EA6_20250506_1439.rsa.dmp)

### Steps to reproduce

From what i heard from player and way i reproduced it:

1. die during PvP / PvE (could be related to fire )
2. few seconds after respawn you crash 

probably related to 'streaming out' after respawn. 
I managed to reproduce it by repsawning at place of death and flying away immideately. 
when normaly players respawn far away from death place.

If someone could check crash-dump and tell me to what its releated, i could get more insight.

### Version

Version = 1.6-release-23183.0.000

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

tederis:
According to provided crash reports there is an issue with the building removal system. It incorrectly handles dummy objects which are about to be streamed out.

Targets cases like #4202 (Technical details: [comment below](https://github.com/multitheftauto/mtasa-blue/pull/4206#issuecomment-2858003162)) where for any reason pLODInterface is passed as nullptr.

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -6443,7 +6443,12 @@ void _declspec(naked) HOOK_CWorld_Remove_CPopulation_ConvertToDummyObject()
     TIMING_CHECKPOINT("+RemovePointerToBuilding");
     RemovePointerToBuilding();
     StorePointerToBuilding();
-    RemoveObjectIfNeeded();
+
+    // pLODInterface contains a dummy object's pointer
+    // And as follows from CPopulation::ConvertToDummyObject this pointer can be nullptr
+    if (pLODInterface)
+        RemoveObjectIfNeeded();
+
     TIMING_CHECKPOINT("-RemovePointerToBuilding");
     _asm
     {


FileEX (approved):

tederis:
As it turned out, the actual reason was related to introduced by #4208 issues. And there's the dedicated PR #4208. But the current PR is still viable as it prevents from potential problems in future.

Dutchman101:
Note: subject is game_sa.dll crash at offset `0x00012AE8` (Bug report #4202 - also see [comment](https://github.com/multitheftauto/mtasa-blue/issues/4202#issuecomment-2856429094))

We were collaborating on investigating it (me, FileEX and TEDERIs) so PR #4208 and #4206 are a conclusive fix.

What was happening is that the whole interior object breaking on explosions like a grenade, the main bug, caused the interior's LOD (going through pLODInterface) to be nullptr as player was streaming out of it when they fell into the sky (interior-less) or died.

> As it turned out, the actual reason was related to introduced by #4208 issues. And there's the dedicated PR #4208. But the current PR is still viable as it prevents from potential problems in future.

Yes. As generally speaking, pLODInterface can be nullptr, and even GTA’s own code checks for that—but CBuildingRemovalSA didn’t up until now with PR #4206.

Both PR #4206 and #4208 will be merged now at the same time, so the whole situation is resolved. Thanks for the help!

PR #4208, opened 2025-05-07
Fix interiors bug introduced by #4187

tederis:
Fixes a crash and interior disappearing when throwing granades inside interiors.

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Explosions.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Explosions.cpp
@@ -48,7 +48,7 @@ static void _declspec(naked) HOOK_CWorld_TriggerExplosionSectorList()
     {
         // check entity->m_nScanCode == CWorld::ms_nCurrentScanCode
         mov ecx, dword ptr ds:[0xB7CD78]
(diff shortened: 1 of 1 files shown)


Dutchman101:
Note: subject is game_sa.dll crash at offset `0x00012AE8` (Bug report #4202 - also see [comment](https://github.com/multitheftauto/mtasa-blue/issues/4202#issuecomment-2856429094))

We were collaborating on investigating it (me, FileEX and TEDERIs) so PR #4208 and #4206 are a conclusive fix.

What was happening is that the whole interior object breaking on explosions like a grenade, the main bug, caused the interior's LOD (going through pLODInterface) to be nullptr as player was streaming out of it when they fell into the sky (interior-less) or died.

> As it turned out, the actual reason was related to introduced by #4208 issues. And there's the dedicated PR #4208. But the current PR is still viable as it prevents from potential problems in future.

Yes. As generally speaking, pLODInterface can be nullptr, and even GTA’s own code checks for that—but CBuildingRemovalSA didn’t up until now with PR #4206.

Both PR #4206 and #4208 will be merged now at the same time, so the whole situation is resolved. Thanks for the help!

PR #4210, opened 2025-05-07
Fix crash when player dies on fire

### LINKED ISSUE #4037: Crash when player is on fire

FileEX:
Fixes #4037 

The crash was introduced in https://github.com/multitheftauto/mtasa-blue/commit/e09b85fd4fc424f5258cc20aa8e9a385bfc6524c

The function CPed::RestorePlayerStuffDuringResurrection calls Fire::Extinguish, and then executes a cmp instruction with the ebx register, which got overwritten inside CFire::Extinguish during the call to our hook. That’s why the condition always evaluated as true—even when the value was actually nullptr.


Client/game_sa/CFireSA.cpp
```diff
@@ -242,13 +242,16 @@ static void _declspec(naked) HOOK_CFire_Extinguish()
     _asm
     {
         mov [eax+730h], edi
+
+        push ebx
         mov ebx, [esp+8]
```
botder: Since you modify `esp` with the `push ebx` above now, don't you have to fix the `+8` offset here?

FileEX: Yeah, you're right i forgot about it.

PR #4224, opened 2025-05-14
Allow using std::variant with several pointers (Lua parser)

tederis:
Currently the Lua parser doesn't support constructions such as 
```cpp
static void SomeFn(std::variant<CClientPlayer*, CClientSound*> value)
```
This PR makes it possible. This opens new possibilities for the refactoring. With this PR it's now possible to rewrite function [setSoundEffectEnabled](https://github.com/multitheftauto/mtasa-blue/blob/f1ede5db824cfb33493585c34e99f54f87620382/Client/mods/deathmatch/logic/luadefs/CLuaAudioDefs.cpp#L1293) using the new parser, for example.


sbx320:
Doesn't feel ideal to read the pointer multiple times, but I don't see an easy alternative either.

sbx320 (approved):

PR #4225, opened 2025-05-16
Fix Camera shoots multiple times if ammo > 1

### LINKED ISSUE #4166: Camera shoots multiple times if ammo > 1.
### Describe the bug

So, i discovered while creating a script that camera in MTA shoots multiple times if you press the button. It doesn't matter how much FPS client have and seems like other resources doesn't affect it. Also it doesn't matter if client have toggled "saving photos in GTA SA folder" - it will save one file of the first frame. Seems like resolution or window mode doesn't matter, just like any other video options. 

I have tested that on my own server and also with default MTASA resources (at least the ones that goes with MTA Map Editor in Testing mode (F5 button)).

It works fine if client will have 1 ammo available in camera, the sound and effect works as intended. 

I've got the video to show that issue: [YT video here](https://youtu.be/-yTQLIcpp30)

### Steps to reproduce

1. Join any server or Map Editor.
2. Take Camera.
3. Shoot it once or hold the attack button.
4. Try this with 1 ammo as well to see if it's fine.

### Version

Multi Theft Auto v1.6-release-23147

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

FileEX:
Fixes #4166


Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -586,6 +586,10 @@ static void __declspec(naked) HOOK_CWeapon_Update()
 {
     _asm
     {
+        // Temp fix for camera
```
tederis, line 589: It tells that this fix is temporary. Can we anticipate subsequent improvements?

FileEX: The proper fix will come when the function is rewritten using our own implementation, which probably won’t happen anytime soon, so this is a temporary solution

PR #4237, opened 2025-05-27
Fix ALT + F4 not working

lopezloo:
Fixes ALT + F4 not working.


tederis:
I guess it would be better to disconnect first and exit the game gracefully.

lopezloo:
> I guess it would be better to disconnect first and exit the game gracefully.

It actually should end game gracefully:
https://github.com/multitheftauto/mtasa-blue/blob/c3d494667a1dced51f427310a1d6b852aefb4bd7/Client/core/CMessageLoopHook.cpp#L190-L194

Just like clicking close icon on game window.

Fernando-A-Rocha (approved):

tederis (approved):

PR #4238, opened 2025-05-29
Add new world special property (vehicle_engine_autostart)

samr46:
This property toggles automatic vehicle engine state behavior (`true` by default).

By default game turns on the engine automatically on driver enter and turns off the engine on driver exit for aircraft. This property allows us to disable this behavior and allow full control of vehicle engine state by using [setVehicleEngineState](https://wiki.multitheftauto.com/wiki/SetVehicleEngineState).

In addition to disabling automatic engine state behavior I modified aircraft logic to allow rotors to spin without driver (if engine is on) and to prevent spinning if engine is off with driver inside. Existing hook is used to add additional changes to rotors state.
So this way the driver can leave helicopter with engine running. And it's possible to fly away instantly because rotors will keep spinning all the time while the engine is on.
At this moment we can achieve this kind of behavior using [setVehicleRotorState](https://wiki.multitheftauto.com/wiki/SetVehicleRotorState) and [setVehicleRotorSpeed](https://wiki.multitheftauto.com/wiki/SetVehicleRotorSpeed). This property works differently because it makes it based on the vehicle engine state instead of rotor state (which should be easier to use and more common).

> setWorldSpecialPropertyEnabled("vehicle_engine_autostart", false)

> [!NOTE]
> The game will not spin up rotors to maximum speed instantly if unoccupied helicopter with running engine just streamed in (like player teleporting to a location with helicopter that has engine on and no driver). It can be mitigated using setVehicleRotorSpeed if necessary although not really needed in most use cases.

> [!WARNING]
> To avoid possible engine state desynchronization between server and client it's recommended to use it on _server side only_ with appropriate minclientversion (like some other existing world properties).


Client/game_sa/CGameSA.cpp
```diff
@@ -930,6 +930,20 @@ void CGameSA::SetVehicleBurnExplosionsEnabled(bool isEnabled)
     m_isVehicleBurnExplosionsEnabled = isEnabled;
 }
 
+void CGameSA::SetVehicleEngineManualModeEnabled(bool isEnabled)
```
tederis: This sort of functions should be placed in `Multiplayer SA` module.

Client/game_sa/CVehicleSA.cpp
```diff
@@ -54,13 +54,39 @@ void _declspec(naked) HOOK_Vehicle_PreRender(void)
...
 static bool __fastcall CanProcessFlyingCarStuff(CAutomobileSAInterface* vehicleInterface)
 {
     SClientEntity<CVehicleSA>* vehicle = pGame->GetPools()->GetVehicle((DWORD*)vehicleInterface);
     if (!vehicle || !vehicle->pEntity)
         return true;
 
-    return vehicle->pEntity->GetVehicleRotorState();
+    if (vehicle->pEntity->GetVehicleRotorState())
+    {
+        if (*(unsigned char*)0x64BC03 == 0x75) // keep default behavior
```
tederis: It's would be better to add new method `IsVehicleEngineManualModeEnabled`.

tederis (changes requested):

Fernando-A-Rocha:
I love this addition. It truly gives more control and freedom. Thank you!

PlatinMTA:
Ok I didn't know I needed this but I need this good job and thanks!

Client/game_sa/CGameSA.h
```diff
@@ -262,6 +262,8 @@ class CGameSA : public CGame
     bool IsVehicleBurnExplosionsEnabled() const noexcept override { return m_isVehicleBurnExplosionsEnabled; }
     void SetVehicleBurnExplosionsEnabled(bool isEnabled) override;
 
+    bool IsVehicleEngineManualModeEnabled() const noexcept override { return *(unsigned char*)0x64BC03 == 0x90; }
```
tederis: This method should be in `Multiplayer SA` as well. Because it's logically connected to memory manipulations inside `SetVehicleEngineManualModeEnabled`.

samr46: It's much easier to call it from Game SA. This is just makes code more complex. I'd actually keep setter in Game SA as well along with other word special properties (like in my initial version).

tederis: See https://wiki.multitheftauto.com/wiki/Coding_info
Game SA is not the right place for low level modifications. And simplicity isn't the main principle when it takes to code organization questions. There must be also coherency and predictability.

samr46: Yet we have many low level modifications in Game SA (including recently added). Also CVehicleSA is part of Game SA. So my initial version is actually more logical.

tederis: Those modifications are obviously erroneous. So please, follow the instructions.

samr46: I want to move setter to Game SA like in my original implementation. I think this is the right way to do it because hook that is used to modify aircraft logic is in Game SA (CVehicleSA.cpp).

tederis: Game SA isn't intended to be used for the modifications. The main purpose of Game SA is holding wrappers around GTA classes (vehicles, peds and so on).

samr46: I understand what you are saying. But this doesn't make sense to make more spaghetti out of it. I'm not even sure that it's possible to include Multiplayer SA into Game SA without additional changes.

I will restore my original implementation that I consider consistent with existing codebase and logical. If it's not going to be accepted - it's fine. We can close this PR and someone else is free to use the code to reimplement this feature in proposed form.

samr46: > Game SA isn't intended to be used for the modifications

By the way, it means that most if not all hooks from CVehicleSA.cpp should be moved to Multiplayer SA. And it kinda doesn't make any sense. It's all VehicleSA related code. Including the feature of this PR.

tederis: I can understand your position. But I prefer adhere to guidlines.
And there's another link for your information: https://github.com/multitheftauto/mtasa-blue/wiki/Dev-Tips

samr46: Then I guess we should close this PR and someone else can use the code to reimplement this feature. I don't want to be the author if it's implemented in proposed form.

Regarding guidelines - it doesn't forbid using Game SA for low level modifications. It's actually mentioned there that it is used for low level modifications of game element related entities (which this feature relies on).

The fact that multiple similar PRs were merged without issues recently makes me wonder why this one resulted in a problem. In any case, it's up to you to either close this PR or keep it open.

tederis: I'm not going to close this PR. The words above are just my opinion. Hopefully there are other developers with write access and maybe they will take a different view.

samr46: Ok. I will keep this conversation not resolved for more feedback.

Also I did check if it's easy or not to directly include MultiplayerSA into CVehicleSA.cpp - it's not, there are conflicts like I previously guessed.

tederis: > Ok. I will keep this conversation not resolved for more feedback.
> 
> Also I did check if it's easy or not to directly include MultiplayerSA into CVehicleSA.cpp - it's not, there are conflicts like I previously guessed.

What do you mean by including MultiplayerSA? Isn't it enough to just `const CMultiplayer* const multiplayer = g_pCore->GetMultiplayer();`? You can extend CMultiplayer interface if needed.

samr46: > Isn't it enough to just `const CMultiplayer* const multiplayer = g_pCore->GetMultiplayer();`

It should be enough. I quickly tested it but it didn't compile due some #include related conflicts. I didn't dig into it. It's possible that full rebuild would fix it.

samr46:
I'm debating the property name. Maybe we should start naming properties in more readable format instead of keeping it consistent with existing properties? It's fine for short names but the more complex property names feel kind of unreadable.

In this case for example `vehenginemanualmode` could become `vehicle_engine_manual_mode` or `VehicleEngineManualMode` (the first alternative I think is the best one). Although manual mode could be confused with "manual transmission" (manual gearbox) so it's also debatable.

Fernando-A-Rocha:
How about `vehenginestatecontrol` ?

tederis:
I would suggest `vehicle_engine_autostart` and make it `true` by default.

Client/game_sa/CVehicleSA.cpp
```diff
@@ -54,13 +57,39 @@ void _declspec(naked) HOOK_Vehicle_PreRender(void)
     }
 }
 
+float& fTimeStep = *(float*)(0xB7CB5C);
```
tederis: Better to make it static to avoid possible collisions in other translation units.

tederis (changes requested):

tederis (approved):
LGTM. Thanks for the efforts.

PR #4245, opened 2025-06-06
Fix anim progress & speed

FileEX:
This PR fixes several behaviors related to animation progress and speed. It is a patch for #3923 and #3520.

The PR not only fixes synchronized animations but also unsynchronized ones on the client side. Previously, when teleporting to a ped with an animation, the animation progress and speed were not saved, causing animations to restart from the beginning. The same happened, for example, after reconnecting.

Now, animation progress and speed are cached and used during stream-in.


Fernando-A-Rocha:
This is an awesome step in the right direction. Thank you 🤗

syncc01:
3D sounds (playSound3D) are affected similarly — when a player teleports to another player or reconnects, the sound restarts instead of continuing from its previous position

Dutchman101:
Thanks, this will be tested in nightly channel.

PR #4253, opened 2025-06-12
Hide game window on quit to make it look instant

### LINKED ISSUE #4242: Quit is slow
### Describe the bug

It takes few seconds to quit MTA client, even when only main menu is shown and game world was never loaded.

What is interesting is that it doesn't happen on custom builds build as `VERSION_TYPE_CUSTOM` where quit is nearly instant even when client is connected to the server.

### Steps to reproduce

Click quit in main menu or press ALT + F4.

### Version

Happens on both `mtasa-1.6-rc-23237-20250605` and `mtasa-1.6-rc-22388-20240203` so probably it's not regression.

### Additional context

Caused by something in netc. `m_pNet->Shutdown();` call is causing that slowdown:
https://github.com/multitheftauto/mtasa-blue/blob/527f873aa0d11d55ec457c25a968af0f1eac03b3/Client/core/CCore.cpp#L1138

### Relevant log output

```shell

```

### Security Policy

lopezloo:
Resolves #4242.

Works best on windowed/borderless video mode.

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -1452,6 +1452,9 @@ void CCore::Quit(bool bInstantly)
 
         WatchDogBeginSection("Q0");            // Allow loader to detect freeze on exit
 
+        // Hide game window to make quit look instant
+        ShowWindow(GetHookedWindow(), SW_HIDE);
+
         // Destroy the client
         CModManager::GetSingleton().Unload();
 


nekjutsu:
"Hide a problem to avoid solving it"

Fernando-A-Rocha:
> "Hide a problem to avoid solving it"

Developer of the anticheat "netc" would need to solve the long execution of the code that makes MTA wait on quit... And I don't think that's gonna happen soon, so this is realistic 😅

nekjutsu:
> > "Hide a problem to avoid solving it"
> 
> Developer of the anticheat "netc" would need to solve the long execution of the code that makes MTA wait on quit... And I don't think that's gonna happen soon, so this is realistic 😅

Let anticheat developer always "feel" the long exit. It will serve as a reminder to him 😉 

P.S. What if I just want to restart the client? I’ll have to guess when the invisible game process will close before launching the client again

AlexTMjugador:
> P.S. What if I just want to restart the client? I’ll have to guess when the invisible game process will close before launching the client again

Given that #4242 states that "it takes few seconds to quit MTA client", it's likely that the time most users would take to launch the game again is greater than the time the game takes to properly close, rendering this workaround a non-issue. Moreover, I don't think launching MTA again tenths of seconds after closing it is a common use case, and if it ever happens, people can just try again.

I agree that the proper fix to this issue would be to locate what's causing that slowdown and optimize that, but given that such slowdown is said to be located on the net module, which for all intents and purposes is a closed-source, binary blob for the vast majority of MTA developers, I reckon getting it fixed there would be a slow process at best.

Edit: it was also said in the linked issue that the net module has been tweaked in [r23238](https://buildinfo.multitheftauto.com/?Revision=23238&Branch=), so maybe the situation with this slowdown is better now?

lopezloo:
> it was also said in the linked issue that the net module has been tweaked in [r23238](https://buildinfo.multitheftauto.com/?Revision=23238&Branch=), so maybe the situation with this slowdown is better now?

Yes, it should be slighty better.

botder (approved):

PR #4255, opened 2025-06-19
Add engineIsModelValid

sousateew:
This PR adds a new utility function **engineIsModelValid**.

Some scripts may need to check whether a model ID corresponds to a valid and loaded model. This function provides a simple way to verify if a model is within valid bounds and has a valid model info entry.
It checks whether the model ID is above the base DFF limit and ensures that the corresponding model info exists and is valid.

```lua
bool engineIsModelValid(number modelID)
```


PlatinMTA:
What about CJ models (do we even need to do a check if ``GetCountOfAllFileIDs`` is used in ``GetModelInfo`` anyways)? Related #4282 somewhat

Also I'm not that good with C but wouldn't it make more sense to first check if the ID is larger than 20k and then try to get the model info?

Anyways I could use this function yes, maybe for a hard clean of all models or something similar.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2594,13 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+bool CLuaEngineDefs::EngineIsModelValid(uint uiModelId)
```
FileEX: uint -> `std::uint32_t`

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2594,13 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+{
+    CModelInfo* pModelInfo = g_pGame->GetModelInfo(uiModelId);
```
FileEX:
```cpp
CModelInfo* modelInfo = g_pGame->GetModelInfo(modelId);
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2594,13 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+
+    if (uiModelId >= 20000 || !pModelInfo)
+        return false;
+
+    return true;
```
FileEX:
```cpp
return (modelId < 20000 && modelInfo);
```

FileEX (changes requested):

G-Moris:
Maybe `return modelInfo && modelInfo->IsValid();`?

PR #4284, opened 2025-07-04
Update dockerimage.yaml

donturb0:


botder:
The current state of this configuration file is straightforward, compact and each line already conveys the necessary information about what it does. There is no need to inflate it with these descriptions.

WUBE1234:
Okay

PR #4295, opened 2025-07-13
Make onVehicleExplode event cancellable

### LINKED ISSUE #3392: Add more utility to deal with blowVehicle abuse

Hello,

With the recent increase of bad people abusing with custom code inside MTA:SA, one of the most used functions to cause mayhem is blowVehicle. Since this is a shared function, clientside execution gets synced with the server, resulting in easy way to blow up all server vehicles.

I propose a few ideas to improve the abusive situation by expanding upon blowVehicle:

1) Add an event like onPlayerVehicleBlowThreshold which gets triggered when the client tries to blow up too many vehicles at given time. Source would be player, vehicle the event parameter.

2) Prevent client from using blowVehicle on vehicle elements which are in a different dimension or streamed out. This could be a problem regarding backwards compatibility but could greatly reduce abusive ways if simply looping all vehicle elements.

3) Add an event like onPlayerVehicleBlowUp which gets triggered when client uses blowVehicle. Source is the player, vehicle the parameter. That way server could write its own checks against abusive use and deal with malicious player.

4) Change clientside blowVehicle to be only used if the client is the active syncer of to be blow up vehicle. Again this might be impossible to do due to backwards compatibility reasons but hey I just throw it in.

_No response_

### Additional context

_No response_

### Security Policy

ghost:
Closes #3392


MegadreamsBE:
LGTM!

We'll probably want to make sure the behavior remains the same with the current expected behavior when not cancelled. Changes are significant enough for that to be a concern imo.

MegadreamsBE (approved):

Dutchman101:
Thanks, this will be tested in nightly channel.

Issue #3160, opened 2023-08-15
'onPlayerCommand' event for client-side


PlatinMTA:
### Is your feature request related to a problem? Please describe.

We don't have an official way to cancel commands from the clientside. ``onClientConsole`` will show all commands being executed, regardless of where they were executed (chat or console), but this information is only useful to know which commands are being executed because you can't cancel them.

My case use in this case would be to make a safe-state in which users can't execute clientsided commands. I understand that cancelling serversided commands should be done on the serverside, and that's already possible.

Some of my commands are made on the clientside, in which the client checks a bunch of things and, if all goes well, then triggers an event on the server which checks whatever it needs to check.

Instead of implementing a switch in every single resource in which a clientsided command is added to determine if the user can or cannot use the command, it would be easier to just disable clientsided commands entirely with an event whenever it's necessary.

Just to be clear, my use case can be done already with the tools given, but it's more hassle than it's worth.

### Describe the solution you'd like

Add an event called ``onClientCommand`` or ``onClientPlayerCommand`` with a possibility of cancelling the event so it won't execute any clientsided commands. Serversided commands should be mantain like before.

### Describe alternatives you've considered

Add the cancel event to ``onClientConsole``, altought I think that name is misleading since it won't just execute for console input but for all input (chat & console).

### Additional context

#2615 is related

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

JeViCo:
Kinda related to #3014

PR #4303, opened 2025-07-16
Add onClientCommand event to cancel client-side commands

Dryxio:
Adds onClientCommand event to intercept and cancel client-side commands before execution.

#3160

* [x] [your code follows the coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines)
* [x] your commit messages are informative ("update file.cpp" is unhelpful. "fix missing model in getVehicleNameFromModel" is helpful)

If your work is incomplete, **do not prefix your pull request with "WIP"**, instead
create a _draft_ pull request: https://github.blog/2019-02-14-introducing-draft-pull-requests/

Thank you!


Client/mods/deathmatch/ClientCommands.cpp
```diff
@@ -45,6 +45,18 @@ bool COMMAND_Executed(const char* szCommand, const char* szArguments, bool bHand
             }
         }
 
+        // Give scripts a chance to cancel this command before processing
+        CClientPlayer* localPlayer = g_pClientGame->GetLocalPlayer();
+        if (localPlayer != nullptr)
```
FileEX:
```cpp
if (localPlayer)
```

Client/mods/deathmatch/ClientCommands.cpp
```diff
@@ -45,6 +45,18 @@ bool COMMAND_Executed(const char* szCommand, const char* szArguments, bool bHand
...
+        {
+            CLuaArguments cancelArguments;
+            cancelArguments.PushString(szCommandBufferPointer);
+            if (!localPlayer->CallEvent("onClientCommand", cancelArguments, true))
+            {
+                return true;
```
FileEX, line 56: Shouldn't this return false?

Dutchman101: @ColombuxMaximus Please check this code review too, i think you missed it after addressing the other comment.

Dryxio:
Let me know if there's something else. Not sure why some of my comments/answers are not showing.
( @Dutchman101 )
image

Dryxio:
Hello @FileEX just in case you missed my comment to your review because Github says it's 'Pending'

Client/mods/deathmatch/ClientCommands.cpp
FileEX: @ColombuxMaximus You're right, it will show 'unknown command'. In that case, maybe it's better to trigger the event in ``CCommands::Execute``? That way, we preserve the correct logic – returning false if the command ultimately doesn't get executed – and at the same time we're able to handle it properly.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -2722,6 +2722,7 @@ void CClientGame::AddBuiltInEvents()
     // Console events
     m_Events.AddEvent("onClientConsole", "text", NULL, false);
     m_Events.AddEvent("onClientCoreCommand", "command", NULL, false);
+    m_Events.AddEvent("onClientCommand", "command", NULL, false);
```
FileEX, line 2725: ``nullptr``

Dryxio:
Thank you @FileEX 
Can you check [here](https://github.com/multitheftauto/mtasa-blue/compare/master...ColombuxMaximus:mtasa-blue:cancel-command-clean) if everything is correct?

I'll test it tomorrow in game

MohabCodeX (changes requested):
Good job , i have made some changes that will enhance command handling by adding argument support to the event , but i don't have changing permission to your branch , anyway the final usage would be like this : 
```lua
function onClientCommand(command, arguments)
    outputChatBox("Command: " .. command, 255, 255, 0)
    outputChatBox("Arguments: " .. arguments, 0, 255, 255)

    if command == "blocked" then
        outputChatBox("Command blocked!", 255, 0, 0)
        return false
    end

    return true
end

addEventHandler("onClientCommand", root, onClientCommand)
```
image

ColombuxMaximus:
Hello, thank you! Can you do it [here](https://github.com/multitheftauto/mtasa-blue/compare/master...ColombuxMaximus:mtasa-blue:cancel-command-clean) please? It's a new clean PR after FileEx's review

FileEX:
> Thank you @FileEX Can you check [here](https://github.com/multitheftauto/mtasa-blue/compare/master...ColombuxMaximus:mtasa-blue:cancel-command-clean) if everything is correct?
> 
> I'll test it tomorrow in game

Okay, I didn’t realize that `CCommands::Execute` is part of the Core project. So let's keep everything within Deathmatch.

My suggestion is that `COMMAND_Executed` should return two booleans: something like `wasCanceled` and `wasExecuted`. Then, in `CCommands::Execute`, we display `Unknown command or cvar:` only if `wasCanceled` is false. The event can stay inside `COMMAND_Executed`.

Also, I noticed a bug. `COMMAND_Executed` should return false if:

```cpp
g_pClientGame->GetRegisteredCommands()->ProcessCommand(szCommandBufferPointer, szArguments);
```

returns false – because that means the command doesn’t exist. Right now, it always returns true, even if the command doesn’t exist. As a result, when you type something like /foo, there’s no “unknown command” message – it just silently fails. You could fix this as well while you're at it.

Lastly, regarding `executeCommandHandler` – I think this function should also trigger `onClientCommand`, but with an additional parameter like `executedByFunction` to indicate whether the command was triggered manually (by typing) or programmatically (via function). 

You could also consider having the `CRegisteredCommands::ProcessCommand` function trigger the event instead of `COMMAND_Executed`, and have it return two booleans – whether the command was canceled and whether it was executed (i.e., if it exists). That way, `executeCommandHandler`, which directly calls `CRegisteredCommands::ProcessCommand`, would benefit from the same logic. This way, we avoid triggering the event in two different places.

Dryxio:
> > Thank you @FileEX Can you check [here](https://github.com/multitheftauto/mtasa-blue/compare/master...ColombuxMaximus:mtasa-blue:cancel-command-clean) if everything is correct?
> > I'll test it tomorrow in game
> 
> Okay, I didn’t realize that `CCommands::Execute` is part of the Core project. So let's keep everything within Deathmatch.
> 
> My suggestion is that `COMMAND_Executed` should return two booleans: something like `wasCanceled` and `wasExecuted`. Then, in `CCommands::Execute`, we display `Unknown command or cvar:` only if `wasCanceled` is false. The event can stay inside `COMMAND_Executed`.
> 
> Also, I noticed a bug. `COMMAND_Executed` should return false if:
> 
> ```c++
> g_pClientGame->GetRegisteredCommands()->ProcessCommand(szCommandBufferPointer, szArguments);
> ```
> 
> returns false – because that means the command doesn’t exist. Right now, it always returns true, even if the command doesn’t exist. As a result, when you type something like /foo, there’s no “unknown command” message – it just silently fails. You could fix this as well while you're at it.
> 
> Lastly, regarding `executeCommandHandler` – I think this function should also trigger `onClientCommand`, but with an additional parameter like `executedByFunction` to indicate whether the command was triggered manually (by typing) or programmatically (via function).
> 
> You could also consider having the `CRegisteredCommands::ProcessCommand` function trigger the event instead of `COMMAND_Executed`, and have it return two booleans – whether the command was canceled and whether it was executed (i.e., if it exists). That way, `executeCommandHandler`, which directly calls `CRegisteredCommands::ProcessCommand`, would benefit from the same logic. This way, we avoid triggering the event in two different places.

Ok I'll do that next.
Closing this PR, I'll open a clean one with those changes.

PR #4311, opened 2025-07-26
Fixes #4299 - Isolate Client Resource Cache by Server

### LINKED ISSUE #4299: Allow client sided resource cache to be isolated per server

Currently, when a player joins a server, all client sided resources are cached in the shared resources/ directory based on the resource name. This leads to several problems:

- If multiple servers use resources with the same name but different contents or versions, the client must redownload them every time they switch servers.

- Developers are forced to prefix names (e.g., server1_inventory, server2_inventory) to avoid conflicts.

Introduce a way for the client to cache resources in a server specific subdirectory, using a unique identifier. This would isolate resource storage per server and eliminate name conflicts or redundant downloads.

There is already a working internal identification system via the server-ids.xml file, which assigns a unique identifier to each server. I propose leveraging this existing solution to define per server cache directories.

For example, instead of saving resources under:
```
resources/inventory/
```

They could be stored as:
```
resources/<id>/inventory/
```

This follows the same structure already used by the priv/ directory, which isolates data per server.

Resource name prefixing, which is widely used but leads to inconsistent resource names, bloated exports.

### Additional context

_No response_

### Security Policy

MohabCodeX:
For: #4299

the implementation includes automatic migration for legacy cached resources, allowing a smooth transition without user interaction or data loss.
While legacy support is currently maintained for backward compatibility, it may be deprecated in the future to simplify the system further. Removing legacy support would reduce maintenance overhead .. but if we didn't implement it for now .. it would require clients to re-download resources when connecting to servers for the first time after the change.


tzwer:
I like the idea. 
I have a question, what will happen to the resources shared between the servers? such as hedit and xhedit. The drift servers use the same resource to let the players have their handlings across all the servers.

MohabCodeX:
>  what will happen to the resources shared between the servers?

With the current automatic migration, shared resources will be copied into each server's folder the first time the client connects.
for example :
```
resources/
│
├── hedit/
│
├── D432TE/          # server ID
│   └── hedit/       # copied from main resources
│
└── DFG523/          # another server ID
    └── hedit/       # another copy
 ```
but if he don't have the resource, for sure it will be downloaded

tzwer:
> > what will happen to the resources shared between the servers?
> 
> With the current automatic migration, shared resources will be copied into each server's folder the first time the client connects. for example :
> 
> ```
> resources/
> │
> ├── hedit/
> │
> ├── D432TE/          # server ID
> │   └── hedit/       # copied from main resources
> │
> └── DFG523/          # another server ID
>     └── hedit/       # another copy
> ```
> 
> but if he don't have the resource, for sure it will be downloaded

Will it only be copied after joining for the first time after this update?
What will happen if the player saves new handlings? will it appear if they play in another server?

MohabCodeX:
> Will it only be copied after joining for the first time after this update?
> What will happen if the player saves new handlings? will it appear if they play in another server?

Yes it , it will only be copied once when you join a server for the first time after the update, as the migration only runs if the server's specific path doesn’t already have the file. After that, the resource in each server’s folder is independent, so if a player saves new handlings in one server, those changes will not appear in another server

For real example , `hud` resource for many RP servers , it changes it's data each time you join new RP (OWLg) server , and all servers are saving their imgs and files in the same hud folder , without differentiating between the files of each server.

So, no the player’s handlings will not be shared across servers since each server has it's own directory. But the player could still do it manually by copying the data, or the server could use a database instead of relying on file based settings.

blonter:
Just asking: What happens with the default resources, especially if every server uses them? There will be countless of copies of the admin resource for example? What about race maps?

MohabCodeX:
Yes, in this approach there will be separate copies of default resources (like admin or race maps) for each server.
I don’t really see this as a problem even if you have the same resource across servers, it’s very common for them to have small differences, like a single line of code or configuration change. That still means the script is different.

tzwer:
What would happen to the servers that have multiple servers? such as Drift Paradise and Dayz?

MohabCodeX:
As I said, for now, to support backward compatibility, downloaded scripts will be copied to the server directory that uses them. But in the future, each server will have its own directory, which means scripts will be re-downloaded. 
**By the way, the suggestion you mentioned is good 👍🏻 I’d like to hear more opinions on the usability and possible methods to configure scripts so we can come up with better solutions.**

samr46:
> which means scripts will be re-downloaded

We have 21 servers with the exact same gamemode (cache size is ~800 MB). 60% of permanent player base play on at least 4 servers on average. Some players play on 10+ servers. And most of players connect to different servers on a daily basis (due server specific events).

So you want to tell me that our players will have to re-download cache files? Is this a joke? And I'm not even talking about wasted disk space on player PCs.

This feature isn't feasible unless you add a way to use the same cache folder for multiple servers.
It can be done in a similar way private per-server cache files work (the key is saved in `server-id.keys` on the server and transferred to the client on connection).

Xenius97:
By default this is a good thing, however there are also servers like the ones samr46 mentioned. That’s why there could be an extra option for servers that overrides this. So if I disable it, the files will be downloaded to one location.

PR #4312, opened 2025-07-26
Fixes #4249 - `isGuestAccount` : return nil on invalid input

### LINKED ISSUE #4249: ``isGuestAccount`` will return false for values that aren't accounts
### Describe the bug

Maybe not actually a bug, but isGuestAccount shouldn't accept values that arent accounts, or if it does it should have a different return value.

``isGuestAccount("something")`` will return false, same with any other value that is not an account
There is a warning being thrown out, but **the return value should be something different than false**, maybe nil

so isGuestAccount should return another value to differentiate
Maybe it should return true, since probably the only reason you are checking if an account is guest is to negate this value and cancel the method/function.

### Steps to reproduce

1. start runcode
2. /run isGuestAccount("something")

### Version

_No response_

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

MohabCodeX:
Fixes #4249


FileEX:
~~This isn't really backwards-incompatible~~, because false and nil are handled the same way in Lua, typically with 
```lua
if (not something) then
```

So existing code will still behave the same. The chance that anyone did something like:
```lua
local b = isGuestAccount(acc)
if (type(b) == 'boolean') then
```

...is minimal

FileEX (approved):

jjustns:
> So existing code will still behave the same. The chance that anyone did something like:
> 
> ```lua
> local b = isGuestAccount(acc)
> if (type(b) == 'boolean') then
> ```
> 
> ...is minimal

I've seen cases of where people will do something like;
```lua
local b = isGuestAccount(acc)
if  b == false then
```
So the backwards incompatibility does apply here.

ghost (approved):

botder:
Thanks for this pull request, but even with the upcoming 1.7 release, I don't think we should touch the return value. It's just not worth it, for correctness sake.

PR #4313, opened 2025-07-26
Fixes #3655 - prevent duplicate command registration

### LINKED ISSUE #3655: addCommandHandler should disallow to create commands with duplicated names
### Describe the bug

As in title.

### Steps to reproduce

addCommandHandler("foo", function()end)
addCommandHandler("foo", function()end) -- command is added but it should not, error should be printed

iprint(getCommandHandlers()) -- prints two above commands, commands are registered

### Version

/

### Additional context

It mostly for security reasons as bad actor can create command with duplicated name to avoid detection. It for example can happen in polish gamemode "pystories" where certain commands such a "admins" are excluded from logs. Because often owners of such servers are not aware of danger and they have not trusted resources someone can create mailicious command with the same name to do not raise suspicious.

It is not only good anticheat that provide protection for servers, but also good api that prevent weird behaviours.

I know this is breaking change, but it is very easy to update resources.

### Relevant log output

_No response_

### Security Policy

MohabCodeX:
Fixes #3655 
Edit : read comments for more updated info

This PR makes addCommandHandler reject any duplicate command names whether they're in the same resource or across different ones. So if a command like "test" is already registered in one resource, trying to register it again (even in the same resource) will fail with a warning in debugscript 3. For example, if Resource A registers "kick", Resource B (or even A again) won’t be able to override it. It's a solid security improvement that ensures predictable command behavior and prevents hijacking or conflicts with minimal changes.
⚠️ Heads up: any existing resources with duplicate commands will now fail to register them so devs may need to audit and rename conflicting ones.
image


Server/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -35,9 +37,15 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (CommandExists(szKey, NULL))
+    {
+        m_pScriptDebugging->LogWarning(pLuaMain->GetVM(), "addCommandHandler: Attempt to register duplicate command '%s'", szKey);
+        return false;
+    }
```
FileEX: Use `g_pGame->GetScriptDebugging`

Client/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -27,6 +28,12 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (CommandExists(szKey, NULL))
+    {
+        CLuaDefs::m_pScriptDebugging->LogWarning(pLuaMain->GetVM(), "addCommandHandler: Attempt to register duplicate command '%s'", szKey);
+        return false;
+    }
+
```
FileEX: Use 
```cpp
g_pClientGame->GetScriptDebugging()->LogWarning(pLuaMain->GetVM(), "addCommandHandler: Attempt to register duplicate command '%s'", szKey);
```

FileEX (changes requested):

ffsPLASMA:
I am against this PR, there are valid reasons to bind multiple functions to a single command.

MegadreamsBE:
> I am against this PR, there are valid reasons to bind multiple functions to a single command.

Should be configurable at the very least

MohabCodeX:
I agree that sometimes multiple functions may need to be bound to the same command, but this can be handled better by combining them inside a single handler instead of registering several handlers for the same command which is bad practice in most cases.

Making this configurable per server is a good idea, and I can adjust the PR to support that. 
Also, something like this should at least have a warning when duplicate handlers are added I know it's a breaking change if it refused the duplicated commands, but the warning would be important for catching accidental duplications.

Thanks for the feedback, I’ll make improvements.

Fernando-A-Rocha:
Please make it configurable as this is trying to solve a problem that doesn't exist, imo.

I used to have a chatbox resource that, if I remember correctly, made use of multiple command handlers per command. Also there are cases in which you want the same command to be handled clientside and also serverside (idk if this PR considers this).

Fernando-A-Rocha:
A warning on duplicate command handlers could be useful, I agree.
The config could be:
<allow_multi_command_handlers>
Values: 0, 1, 2
Where 0 is disallow, 1 is allow but with warning, and 2 is allow but no warnings.

MohabCodeX:
Thank you for all the feedback

---

### ✅ **Final Outcome: Configurable Duplicate Command Handler Behavior**

A new server configuration option has been added:

```xml
<allow_multi_command_handlers>1</allow_multi_command_handlers>
```

This option controls how the system handles multiple `addCommandHandler` registrations for the same command.


###  **Available Levels:**

| Value           | Behavior                                  | Debug Output                                                                                      |
| --------------- | ----------------------------------------- | ------------------------------------------------------------------------------------------------- |
| `0`             | ❌ **Disallow** duplicate command handlers | `addCommandHandler: Duplicate command registration blocked for '%s' (multiple handlers disabled)` |
| `1` *(default)* | ⚠️ **Allow with warning**                 | `addCommandHandler: Attempt to register duplicate command '%s'`                                   |
| `2`             | ✅ **Allow silently**                      | *(No debug output)*                                                                               |


###  **Default Behavior:**

* The default value is **`1`**, which **allows** duplicates but emits a **warning**, helping developers identify unexpected duplicate registrations.

Any other feedback would be appreciated.

Fernando-A-Rocha:
Nice
When blocking duplicate command handlers I think it should throw a Lua error instead of a warning

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -120,6 +120,7 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
 
     m_bCursorEventsEnabled = false;
     m_bInitiallyFadedOut = true;
+    m_eAllowMultiCommandHandlers = MULTI_COMMAND_ENABLED;
```
FileEX: We no longer use Hungarian notation (not even for enums).

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6137,6 +6138,16 @@ bool CClientGame::GetBirdsEnabled()
...
 
+void CClientGame::SetAllowMultiCommandHandlers(CClientGame::eMultiCommandHandlerPolicy policy)
+{
+    m_eAllowMultiCommandHandlers = policy;
+}
+
+CClientGame::eMultiCommandHandlerPolicy CClientGame::GetAllowMultiCommandHandlers() const
+{
+    return m_eAllowMultiCommandHandlers;
+}
```
FileEX: This can be moved to the header and marked as noexcept

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -192,6 +192,12 @@ class CClientGame
         QUIT_CONNECTION_DESYNC,
         QUIT_TIMEOUT,
     };
+    
+    enum eMultiCommandHandlerPolicy
+    {
+        MULTI_COMMAND_DISABLED = 0,
+        MULTI_COMMAND_ENABLED = 1
+    };
```
FileEX: What do you think about enum class?

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -5567,6 +5575,10 @@ void CPacketHandler::Packet_SyncSettings(NetBitStreamInterface& bitStream)
     if (bitStream.Can(eBitStreamVersion::ShotgunDamageFix))
         bitStream.Read(ucAllowShotgunDamageFix);
 
+    uchar ucAllowMultiCommandHandlers = 1;
+    bitStream.Read(ucAllowMultiCommandHandlers);
+    g_pClientGame->SetAllowMultiCommandHandlers(static_cast<CClientGame::eMultiCommandHandlerPolicy>(ucAllowMultiCommandHandlers));
```
FileEX: ou need to use BitStream.Can to check whether the client/server is capable of handling this functionality. If a newer client connects to an older server build, it will try to read non-existent data, which may result in a crash or a protocol error

Client/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -27,6 +28,23 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (CommandExists(szKey, NULL))
```
FileEX: nullptr

Client/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -27,6 +28,23 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
...
+        if (allowMultiHandlers == CClientGame::MULTI_COMMAND_DISABLED)
+        {
+            g_pClientGame->GetScriptDebugging()->LogWarning(pLuaMain->GetVM(), "addCommandHandler: Duplicate command registration blocked for '%s' (multiple handlers disabled)", szKey);
+            return false;
+        }
+        // If allowing with warning (default), log warning and proceed
+        else if (allowMultiHandlers == CClientGame::MULTI_COMMAND_ENABLED)
+        {
+            g_pClientGame->GetScriptDebugging()->LogWarning(pLuaMain->GetVM(), "addCommandHandler: Attempt to register duplicate command '%s'", szKey);
+        }
```
FileEX: You **can** remove braces (but don't have to)

Server/mods/deathmatch/logic/CMainConfig.h
```diff
@@ -150,6 +150,7 @@ class CMainConfig : public CXMLConfig
 
     int GetPlayerTriggeredEventInterval() const { return m_iPlayerTriggeredEventIntervalMs; }
     int GetMaxPlayerTriggeredEventsPerInterval() const { return m_iMaxPlayerTriggeredEventsPerInterval; }
+    int GetAllowMultiCommandHandlers() const { return m_iAllowMultiCommandHandlers; }
```
FileEX: noexcept & hungarian notation

Server/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -35,11 +38,28 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (CommandExists(szKey, NULL))
```
FileEX: nullptr

Server/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -35,11 +38,28 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
...
+        if (iAllowMultiCommandHandlers == 0)
+        {
+            // Block silently
+            return false;
+        }
+        else if (iAllowMultiCommandHandlers == 1)
+        {
+            // Allow with warning (default behavior)
+            g_pGame->GetScriptDebugging()->LogWarning(pLuaMain->GetVM(), "Attempt to register duplicate command '%s'", szKey);
+        }
```
FileEX: You can remove braces

Server/mods/deathmatch/mtaserver.conf
```diff
@@ -94,6 +94,11 @@
          Available values: special detection (SD) codes ; (e.g. To enable special detection #15 use: 15) -->
     <enablesd></enablesd>
 
+    <!-- This parameter controls how duplicate command handlers are handled.
+         Some resources may need to register multiple handlers for the same command.
+         Values: 0 - disallow duplicate handlers, 1 - allow with warning, 2 - allow silently ; default value: 1 -->
+    <allow_multi_command_handlers>1</allow_multi_command_handlers>
+
```
FileEX, line 101: Also edit `mtaserver.conf.template`

FileEX (changes requested):

Client/mods/deathmatch/logic/CClientGame.h
MohabCodeX: That makes sense. I'll consider replacing it with enum class and refactor the usages accordingly for better type safety

Server/mods/deathmatch/logic/packets/CSyncSettingsPacket.h
```diff
@@ -36,4 +36,5 @@ class CSyncSettingsPacket final : public CPacket
     uchar                 m_ucAllowFastSprintFix;
     uchar                 m_ucAllowDrivebyAnimationFix;
     uchar                 m_ucAllowShotgunDamageFix;
+    uchar                 m_allowMultiCommandHandlers;
```
FileEX: std::uint8_t?

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -193,6 +193,12 @@ class CClientGame
         QUIT_CONNECTION_DESYNC,
         QUIT_TIMEOUT,
     };
+    
+    enum class MultiCommandHandlerPolicy
+    {
+        DISABLED = 0,
+        ENABLED = 1
+    };
```
FileEX: I think this enum is unnecessary. Wouldn’t a simple bool be better here?

FileEX:
I’m missing the ability to change this setting via ``setServerConfig``.
Having to stop the server just to change such a minor setting is imo unacceptable

MohabCodeX:
@FileEX 
> I’m missing the ability to change this setting via setServerConfig.
> Having to stop the server just to change such a minor setting is imo unacceptable

Thanks for the heads-up, the suggestion is implemented (also client side doesn't need to reconnect to have the new changes of the server config) and reviews applied, but I’d like opinions on one feature. Currently, it warns on duplicate commands on the same side (client or server), but not if the same command exists on both. I think that’s fine since they’re separate, but if needed we could add it as an optional check (though I don’t think we’ll need it, but I’d like to hear opinions).

Server/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -35,6 +38,26 @@ bool CRegisteredCommands::AddCommand(CLuaMain* pLuaMain, const char* szKey, cons
     assert(pLuaMain);
     assert(szKey);
 
+    if (CommandExists(szKey, nullptr))
+    {
+        const auto policy = static_cast<MultiCommandHandlerPolicy>(g_pGame->GetConfig()->GetAllowMultiCommandHandlers());
```
FileEX: const here is unnecessary

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -4701,9 +4701,10 @@ void CGame::SendSyncSettings(CPlayer* pPlayer)
     uchar ucAllowFastSprintFix = true;
     uchar ucAllowDrivebyAnimFix = true;
     uchar ucAllowShotgunDamageFix = true;
+    const std::uint8_t multiCommandHandlerPolicy = static_cast<std::uint8_t>(m_pMainConfig->GetAllowMultiCommandHandlers());
```
FileEX: const here is unnecessary

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6139,6 +6140,7 @@ bool CClientGame::GetBirdsEnabled()
     return m_bBirdsEnabled;
 }
 
+
```
FileEX: extra space

FileEX (changes requested):

FileEX:
> @FileEX
> 
> > I’m missing the ability to change this setting via setServerConfig.
> > Having to stop the server just to change such a minor setting is imo unacceptable
> 
> Thanks for the heads-up, the suggestion is implemented (also client side doesn't need to reconnect to have the new changes of the server config) and reviews applied, but I’d like opinions on one feature. Currently, it warns on duplicate commands on the same side (client or server), but not if the same command exists on both. I think that’s fine since they’re separate, but if needed we could add it as an optional check (though I don’t think we’ll need it, but I’d like to hear opinions).

It's fine imo. But i'm still missing the ability to change this setting via `setServerConfig`.

MohabCodeX:
> It's fine imo. But i'm still missing the ability to change this setting via setServerConfig.

I’ve already added this feature and tested it. You can use setServerConfig / getServerConfig.

FileEX (approved):

FileEX:
Let’s wait for the restoration of #3857, because for now we don’t know what will happen with the mtaserver.conf_template file.

MohabCodeX:
Yeah, I was just about to say that. 👍🏻

PR #4317, opened 2025-07-28
Fix isPedOnGround

### LINKED ISSUE #534: isPedOnGround issues

FileEX:
Fixed #534

Original author: #1603

In addition to fixing the precision of `IsOnGround`, this PR adds a `checkVehicles` argument that allows specifying whether the function should return true if the ped is standing on a vehicle (by default it returns false, as before).


Shared/mods/deathmatch/logic/Utils.cpp
botder, line 1: Don't we already have functions that do these operations?

FileEX: I didn't find any

Shared/mods/deathmatch/logic/Utils.h
```diff
@@ -336,3 +336,9 @@ void DeletePointersAndClearList(T& elementList)
         delete *iter;
     }
 }
+
+// Comparing floating point numbers
+bool approximatelyEqual(float a, float b, float epsilon = FLOAT_EPSILON);
```
tederis: Why do these names start with lower case letters?

Shared/mods/deathmatch/logic/Utils.cpp
```diff
@@ -1102,3 +1102,23 @@ CVector ConvertEulerRotationOrder(const CVector& a_vRotation, eEulerRotationOrde
         return a_vRotation;
     }
 }
+
+bool approximatelyEqual(float a, float b, float epsilon)
```
tederis: I think they should be marked `noexcept` and `inline`.

Shared/mods/deathmatch/logic/Utils.h
tederis: `std::numeric_limits<float>().epsilon()`?

ArranTuna:
I have tested this PR and confirms it fixes isPedOnGround being unreliable, when you were moving on a slope isPedOnGround would flicker between true/false now it stays on true.

This might not seem like an important fix, but it might be the reason why there is significant player movement desync where players can be seen moving forwards and then snapping back when the next puresync packet arrives because IsOnGround is used in puresync so if 1 puresync update sends you as not on ground it can cause this: "they’re flagged as airborne. While flagged airborne, puresync enables velocity-based prediction with little or no friction, so the remote player keeps sliding forward even after their real movement stops"

This desync is brought up a lot by players, so if this does fix it, they'd be very happy. And at the very least it'd fix this function which is also itself quite a useful fix.

Shared/mods/deathmatch/logic/Utils.cpp
botder: I found one function in SharedUtil.Math.h named `IsNearlyEqual`

FileEX: Yes, but as you can see, this function works differently from those.

Shared/mods/deathmatch/logic/Utils.h
FileEX: I copied it from the mentioned PR and didn’t pay attention

Dutchman101:
Nice, this will likely be backported to the next 1.6.0 build that'll also be circulated, so it can be of use before the release of 1.7

PR #2533, opened 2022-01-30
Allow allocating models server-side

TheNormalnij:
Fixes #2427
```lua
-- Get all allocated models
table getAllocatedModels([string modelType])

-- Get free id's list
table getFreeModels()

-- Allocate model
bool allocateModelFromParent(number parentId)

-- Unload model, works for custom models
bool unloadModel(nuber modelId)
```

PR #4324, opened 2025-07-31
Remove bad vehicle model offset for 1.7

G-Moris:
uint16_t or uint32_t?


MegadreamsBE:
Is there any benefit to this change? I don't see a reason why the current method of fitting the model into a char isn't fine.

G-Moris:
> Is there any benefit to this change? I don't see a reason why the current method of fitting the model into a char isn't fine.

This will help with #2533 and potential future changes.

MegadreamsBE:
> > Is there any benefit to this change? I don't see a reason why the current method of fitting the model into a char isn't fine.
> 
> This will help with #2533 and potential future changes.

Seems fair. Thanks for the clarification!

botder (approved):

Issue #413, opened 2018-09-05
Add "aiming" to setCameraFieldOfView


qaisjp:
**Describe the solution you'd like**
It has not been implemented yet. See https://github.com/multitheftauto/mtasa-blue/commit/2f892d649af212e0a998a78695c96c09e0ee03a0#commitcomment-11994347

**Additional context**
From https://bugs.mtasa.com/view.php?id=8985

Einheit-101:
Is there any progress on this?

qaisjp:
no

PR #4325, opened 2025-07-31
setCameraFieldOfView improvements

### LINKED ISSUE #4275: setCameraFieldOfView: "aiming"

Currently there is no easy and convenient way to adjust FOV while aiming a weapon. One of purposes could be creating various weapon scopes (with different zoom levels), or changing how zooming in (while aiming) works. The game already utilises different FOV based on weapon you aim with, e.g: pistol/assault rifle/rifle - each of them have different fov applied, so i assume it'd be possible to do.

cameraMode: the camera mode to set the field of view of:

"player": whilst walking/running
"vehicle": whilst in vehicle
"vehicle_max": the max the field of view can go to when the vehicle is moving at a high speed (must be higher than "vehicle")
**"aiming": while aiming a weapon**

/

### Additional context

/

### Security Policy

FileEX:
Fixes #4275 & #413

This PR adds several camera modes for modifying FOV:

- **aiming** – aiming with a weapon

- **sniper_aiming** – aiming with a sniper rifle (with increased zoom range)

- **1stperson_aiming** – aiming with a weapon using the flag_aim_1st_person

- **vehicle_bump** – bumper camera view in vehicles

This PR also adds a ``resetCameraFieldOfView`` function to restore default values.

### Important notes:

- The aiming, sniper_aiming, and 1stperson_aiming modes should be used within the render loop, as their FOV values are overwritten every frame.

- resetCameraFieldOfView does not reset the FOV for the aiming mode while actively aiming. The FOV will automatically reset once the player stops aiming.

- Low values, as documented on the wiki, may cause visual glitches but do not cause crashes (tested).

- Using setCameraFieldOfView with the sniper_aiming mode increases the sniper zoom range from the default to the full function range of 0–179. This extended range remains until resetCameraFieldOfView is called, which then restores the default range of 15–70."

Thanks to @sacr1ficez  for the ideas and testing.


Client/game_sa/CSettingsSA.cpp
```diff
@@ -510,6 +522,96 @@ float CSettingsSA::GetFieldOfViewVehicleMax()
     return ms_fFOVCarMax;
 }
 
+float CSettingsSA::GetFieldOfViewAiming()
+{
+    CCamera* camera = pGame->GetCamera();
```
tederis, line 527: What do you think about placing assertations here and below? Just to be sure that these pointers are always valid.

FileEX: In theory, cam or camera should never be nullptr, but I think it won’t hurt to check anyway

Issue #1861, opened 2020-11-22
engineRequestModel("vehicle") doesn't load default GTA scripts


Xenius97:
**Describe the bug**
engineRequestModel("vehicle") doesn't load default GTA scripts
I've tested with Towtruck, Tractor, Police Maverick (light doesn't work), Seasparrow (can't fire), Packer etc.

**To reproduce**
engineRequestModel("vehicle", 525)

**Version**
Multi Theft Auto v1.5.8-release-20732

AlexTMjugador:
#1701 explicitly states that this kind of behaviours do not work:

> Does not work:
> * Сhassis
> * Turrets
> * ZR-350 opening headlights
> * Rhino, Hydra, Skimmer, Seasparrow
> * Some specific effects like a dust effect from some helicopters, smoke effects from bikes.
> * Another hardcoded stuff

I guess this is caused because somewhere in GTA code they are limited to a specific vehicle ID. If that's the case, for these to work, the affected methods would need to be patched.

qaisjp:
The wiki should be updated to include this information. That article is also missing correct versioning usage

Xenius97:
Trains doesn't work too.

Fernando-A-Rocha:
🤔  Is it possible to fix this at this stage?

Pirulax:
Would require to patch tons of code, so no, not really.
Special features are hard coded based on modelID.

TheNormalnij:
I think it's still hackable. When i added vehicle allocation i thought  about a `setVehicleModelFeatureState` function

lopezloo:
I think it could be implemented like that: https://github.com/multitheftauto/mtasa-blue/blob/eedbdc5fbfdd5d5bed8454728d8e224a4b94838e/Client/multiplayer_sa/CMultiplayerSA.cpp#L5755-L5794

Hooking in place of these ID checks and calling vehicle entity method to check if specific feature is enabled for specific vehicle element.

So MTA API might look like this:
```lua
setVehicleFeatureEnabled(vehicle theVehicle, string feature, bool enabled)
```

In some places there might be no vehicle entity available though in theory. It would be great to document all the missing features.

Fernando-A-Rocha:
Bumping this. A lot of people have messaged me in regards to my resource newmodels saying that added vehicles can't have features like attaching trailers, having water hose etc... It's a popular demand

Pirulax:
It's possible to fix all these things, and it's not even that hard.
[...]

PR #4330, opened 2025-08-01
New vehicle features functions (setVehicleSpecialFeatureEnabled)

FileEX:
This PR is the first step toward resolving #1861.

It introduces a new set of functions that allow manipulation of vehicle features. For now, only the water_cannon functionality is supported - additional features will be added in separate future PRs. These functions allow enabling or disabling the water cannon support.

Importantly, the functionality works both on a model level and on specific vehicles - this means, for example, you could implement a system where a firetruck can't use its water cannon if its tank is empty!

It works for both allocated vehicles and vanilla game vehicles.

### Syntax
```lua
bool setVehicleSpecialFeatureEnabled(vehicle/int modelOrVehicle, string feature, bool enabled)
bool isVehicleSpecialFeatureEnabled(vehicle/int modelOrVehicle, string feature)
```

Available features:
- **water_cannon**

image


Dryxio:
Huge W
Aim FOV aswell 🫡

Fernando-A-Rocha:
Wow, thank you for this contribution!
This fix and addition has been awaited for the past years. I can say that I will continue working on the community resource "newmodels" that is a tool/library to easily manage added models and their features in servers, with full sync.
This feature will be included :)

Fernando-A-Rocha:
Towing ability for tow truck and other trucks like Roadtrain will be super cool. Afaik for this to work the DFF of the vehicle needs to have a special dummy that is positioned at the tow hitch (where the trailer/vehicle will attach), right?

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -27,7 +27,7 @@ void CAutomobileSAInterface::SetPanelDamage(std::uint8_t panelId, bool breakGlas
     if (nodeId < 0)
         return;
 
-    eCarNodes node = static_cast<eCarNodes>(nodeId);
+    CarNodes::Enum node = static_cast<CarNodes::Enum>(nodeId);
```
tederis, line 33: How about `using namespace CarNodes` and then just use `WINDSCREEN`, `WING_LF` and `WING_RF` directly?

Client/game_sa/CAutomobileSA.cpp
```diff
@@ -87,3 +87,79 @@ void CAutomobileSAInterface::SetPanelDamage(std::uint8_t panelId, bool breakGlas
...
+    auto* vehicle = pGame->GetPools()->GetVehicle(reinterpret_cast<DWORD*>(vehicleInterface));
+    if (!vehicle || !vehicle->pEntity)
+        return;
+
+    // Support for default vehicles is still in the GTA code, so we don't need to do it again
+    if (vehicleInterface->m_nModelIndex == 407 || vehicleInterface->m_nModelIndex == 601)
+        return;
+
+    // Simple turret like in fire truck
+    if (vehicle->pEntity->IsSpecialFeatureEnabled(VehicleFeatures::Enum::WATER_CANNON) && !vehicle->pEntity->IsSpecialFeatureEnabled(VehicleFeatures::Enum::TURRET))
```
tederis: How about `using namespace VehicleFeatures` and then just use `WATER_CANNON` and `TURRET` directly?

Client/game_sa/CGameSA.h
```diff
@@ -322,6 +323,13 @@ class CGameSA : public CGame
 
     bool SetBuildingPoolSize(size_t size);
 
+    bool SetVehicleModelSpecialFeatureEnabled(std::uint16_t model, const VehicleFeatures::Enum& feature, bool enabled) override { return CVehicleSA::SetModelSpecialFeatureEnabled(model, feature, enabled); };
```
tederis: Can be shortened to `VehicleFeatures::Enum feature` because the underlying type is numerical. By passing it by ref you can potentially slow down the execution.

Client/game_sa/CVehicleSA.cpp
```diff
@@ -2423,3 +2421,74 @@ void CVehicleSA::ReinitAudio()
...
         audioInterface->SoundJoin();
 }
+
+bool CVehicleSA::SetSpecialFeatureEnabled(const VehicleFeatures::Enum& feature, bool enabled)
+{
+    CVehicleSAInterface* vehicleInterface = GetVehicleInterface();
+
+    switch (feature)
+    {
+        case WATER_CANNON:
```
tederis, line 2431: It looks like the `if` would fit better here.

Client/game_sa/CVehicleSA.cpp
```diff
@@ -2423,3 +2421,74 @@ void CVehicleSA::ReinitAudio()
...
+}
+
+bool CVehicleSA::SetModelSpecialFeatureEnabled(std::uint16_t model, const VehicleFeatures::Enum& feature, bool enabled)
+{
+    CModelInfo* modelInfo = pGame->GetModelInfo(model);
+    if (!modelInfo)
+        return false;
+
+    if (feature == VehicleFeatures::Enum::WATER_CANNON)
+        if (static_cast<VehicleClass>(modelInfo->GetVehicleType()) != VehicleClass::AUTOMOBILE)
```
tederis, line 2454: I dislike this style of nesting. Please, arrange the brackets around the outer if statement.

Client/game_sa/CVehicleSA.h
```diff
@@ -222,6 +222,13 @@ class CAutoPilot
     BYTE pad[56];
 };
 
+enum class eComponentRotationAxis
```
tederis: `enum class eComponentRotationAxis : uint8_t`

tederis (changes requested):

Client/game_sa/CGameSA.h
FileEX: You’re right, an enum is just a regular number, so there’s no point in passing it by reference.

Client/game_sa/CVehicleSA.cpp
FileEX: I think a switch is appropriate here since more types will be added soon. For now, there’s only one, so an if would work better, but considering future development, I believe keeping the switch is fine.

Client/game_sa/CVehicleSA.h
FileEX: CVehicle::SetComponentRotation (0x6DBA30) expect int (4 bytes). Should i use static_cast for it?

tederis: Then maybe better to use uint32_t and add a comment about the reason of choosing it.

Client/game_sa/CVehicleSA.cpp
tederis: Okay then.

PR #4332, opened 2025-08-02
Make cross compiling the project using msvc-wine on Linux possible

vahook:
This PR aims to streamline compiling (mainly the Client) on Linux using msvc-wine. I plan on submitting a full write-up about the process to the wiki.

This PR replaces CertUtil with 7za.exe inside `utils/buildactions/utils.lua`. This is needed, because Wine (as of writing) doesn't implement CertUtil, hence `os.sha256_file` would fail otherwise. On Windows, this shouldn't break anything.

The PR also introduces a helper build script `./wine-build.sh`. `./win-build.bat` can't be used, since the `msbuild` wrapper script that's shipped by msvc-wine fails to correctly set the paths in the environment variables if it's invoked from withing Wine. Also, `vswhere.exe` would just never work on Wine. This script is not strictly necessary for things to work though, as the individual commands can be invoked directly:
```sh
wine win-create-projects.bat < /dev/null
/path/to/msvc-wine-install-dir/bin/x64/msbuild" Build/MTASA.sln -property:Configuration=Release -property:Platform=Win32 -maxCpuCount
```
Thus it can be removed from this PR, if requested.

### DIFF
--- a/utils/buildactions/utils.lua
+++ b/utils/buildactions/utils.lua
@@ -96,7 +96,7 @@ function os.sha256_file(path)
 	local windows = os.host() == "windows"
 	local s, errc
 	if windows then
-		s, errc = os.outputof(string.format("utils\\7z\\7za.exe h -scrcSHA256 \"%s\"", path))
+		s, errc = os.outputof(string.format("call \"utils\\7z\\7za.exe\" h -scrcSHA256 \"%s\"", path))
 	else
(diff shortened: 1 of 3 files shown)


Fernando-A-Rocha:
Congrats on the initiative!
We are slowly migrating documentation of things from the Wiki to the [new wiki](https://github.com/multitheftauto/wiki.multitheftauto.com) (stepping away from MediaWiki), so please reach out on Dev Discord #data-driven-docs channel so we can decide the best way and format to document this.

botder (approved):

PR #4353, opened 2025-08-11
Reimplement projectiles

FileEX:
This PR is one of the steps towards starting a new approach to the project - refactoring and removing ASM and hooks by implementing GTA functionality ourselves.

This PR contains rewritten and improved projectile logic, modified to produce cleaner code than the original, while keeping the behavior exactly the same.

With this PR, we can increase the projectile limit from 32 to any number, and gain full control over projectile behavior. Additionally, the onClientProjectileCreation event can now be canceled, thus eliminating the need for hooks and ASM instructions.

At the moment, the PR is somewhat messy - there are many references to internal functions and raw addresses but I believe the PR is already quite large, so I don’t want to address that here. Everything will be cleaned up in the next PR so that this one doesn’t become too huge.

I have tested numerous scenarios: all projectiles, thrown manually and from vehicles, projectiles thrown by remote players, and based on my tests I haven’t observed any bugs other than those already present in the original. In my opinion, the code reproduces the original functionality and behavior 100%.


tzwer:
Thanks for your efforts!
Will these changes make it possible to calculate how much a projectile traveled? or if a projectile has entered a collision shape?

FileEX:
> Thanks for your efforts! Will these changes make it possible to calculate how much a projectile traveled? or if a projectile has entered a collision shape?

Yeah

Issue #4171, opened 2025-04-20
Request: Stop overriding `gta_sa.set` file.


blonter:
### Is your feature request related to a problem? Please describe.

Hello.

I recently noticed that MTA uses the singleplayer's settings file despite it has its own.
The `gta_sa.set` file which is located in your Documents folder will be overriden each time you launch MTA.

This is bad if you use different settings for singleplayer and MTA. I don't know what settings are being overridden, the screen resolution is one of them but it could be more.

I personally use a 4:3 screen resolution (1400x1050) in singleplayer because this game doesn't really have a widescreen support. The toggleable option in settings is broken, it'll make things worse. Widescreen fix mod could be an option but if you are a speedrunner you can't really install all fixing mods like silentpatch etc. Or if you just dislike having such mods at all.

So you either stick with FHD in singleplayer as well and live with the streched vehicles/peds/hud or you set up a 4:3 screen resolution.

If you first install MTA, it should still read the contents of `gta_sa.set` so all your binds will be there, but after having the `coreconfig.xml` I think that the game should not touch `gta_sa.set` file anymore. 

Installing the game twice for singleplayer and for MTA separatedly is not an option, since the settings file will be the same for every instance of installation. 

### Describe the solution you'd like

If MTA needs to keep using the settings file, then somehow make it to not to override the screen resolution as well. Most settings will stay unchanged, even if you use different keybindings in MTA, they won't be synced back to singleplayer only the screen resolution (based on my experiences)

Totally getting rid of the usage of the settings file would be preferrable. However it should still use it once (for  `reading` only) when `coreconfing.xml` is absent.

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

Issue #1304, opened 2020-03-20
Add setting to load from separate `gta_sa.set` file


Lordmau5:
**Is your feature request related to a problem? Please describe.**
So this is mostly coming from a speedrunner point of view - we value our `gta_sa.set` file in the `Documents\GTA SA User Files` folder
Some are speedrunning the game in 4:3 or some other specific settings enabled that we would much rather otherwise like to disable when playing MTA.

Would there be a chance for a config option ingame to have the game write to / load from a separate `gta_sa.set` file?
If that's even possible with how MTA works - I'm not sure about the specifics of what the game can and is hooking into

**Describe the solution you'd like**
An option or setting to allow writing to and loading from a separate `gta_sa.set` file in the User Files (or any other) directory. Perhaps have it be called `mta_sa.set`?

**Describe alternatives you've considered**
Can't think of any alternatives to write and load the settings file...

PR #4354, opened 2025-08-11
Change `GTA San Andreas User Files` to `MTA San Andreas User Files`

G-Moris:
This will help isolate files like `gta_sa.set` from other GTA:SA builds.
It may not be the most proper way to do this in the `loader` module, but it’s already too late to do it in the `game_sa` and `multiplayer_sa` modules.

This partially solves:
#1304
#4171


botder:
That's like the worst place to do it. Early patching can be done in Client Core project, because the core initialization runs before the  game's WinMain, even before static initialization.

G-Moris:
> That's like the worst place to do it. Early patching can be done in Client Core project, because the core initialization runs before the game's WinMain, even before static initialization.

Do you have any idea where can do this?

botder:
> > That's like the worst place to do it. Early patching can be done in Client Core project, because the core initialization runs before the game's WinMain, even before static initialization.
> 
> Do you have any idea where can do this?

`void CCore::ApplyHooks()` is currently the best place

nweb:
In current process it is better to use VirtualProtect (with sizeof(newDocumentPath)) and direct memcpy, or just replace first byte of string with 'M'.

Client/core/CCore.cpp
```diff
@@ -871,6 +880,8 @@ void CCore::ApplyHooks()
 {
     WriteDebugEvent("CCore::ApplyHooks");
 
+    PatchMemory(0x8747A9, 'M');
```
qaisjp, line 883: This is a really simple solution to make GTA save to a different location, nice!

If we're creating a new folder, I would prefer we save to an existing folder like `Saved Games/MTA San Andreas`, to [avoid this frustration](https://www.reddit.com/r/pcgaming/comments/8ry11w/dont_you_hate_when_a_game_saves_to_my_documents/).

If it's not possible to do `Saved Games/MTA San Andreas` (e.g. it's too long), I would accept `My Games/MTA San Andreas`, as that should fit within the existing buffer:

```
GTA San Andreas User Files
Saved Games/MTA San Andreas
My Games/MTA San Andreas
```

Client/loader/CInstallManager.cpp
qaisjp, line 1: What is the migration plan?

If we merge this and someone runs MTA, they will start off with no settings and will have to reconfigure everything.

I think we should copy the existing `gta_sa.set` if it exists to the new location as early as possible.

Client/loader/CInstallManager.cpp
```diff
@@ -1604,7 +1604,7 @@ SString CInstallManager::_ProcessLangFileChecks()
     const char* langFileNames[] = {"american.gxt", "french.gxt", "german.gxt", "italian.gxt", "spanish.gxt"};
 
     // Get language that will be used
-    SString strSettingsFilename = PathJoin(GetSystemPersonalPath(), "GTA San Andreas User Files", "gta_sa.set");
+    SString strSettingsFilename = PathJoin(GetSystemPersonalPath(), "MTA San Andreas User Files", "gta_sa.set");
```
qaisjp, line 1607: I think this function uses any complete language file as a fallback to fix missing ones:

1. If the user picks German language in GTA, and `german.gxt` exists, then do nothing and return.
1. If the user picks German language in GTA, but `german.gxt` doesn't exist on disk, then auto-fix.
1. If the user has no GTA settings, and not all languages exist on disk, then auto-fix.

And auto-fixing will: find the first complete file that exists (e.g. `spanish.gxt`) and overwrite all the other non-complete languages with `spanish.gxt`. Where complete is > 10kb.

With your current patch, on first install, the third code branch will always be reached. I think if you copy `gta_sa.set` from the old directory first, then this issue will be resolved.

qaisjp (changes requested):
Thanks for the pull request. I agree that we should solve this problem. Let me know what you think of this feedback

PR #4356, opened 2025-08-12
Add function to change clothing cache time

W3lac3:
This function will allow you to change the cache time for CJ's clothes. Currently, the time is fixed and short, which can affect performance if you use custom models brought in by the new `addClothingModel` function.


Client/multiplayer_sa/CMultiplayerSA_ClothesCache.cpp
```diff
@@ -16,6 +16,8 @@
 
 #define CLOTHES_REF_TEST    1       // Debug clothes geometry refs
 
+static constexpr uint DEFAULT_CACHE_TIME = 1000;
```
tederis: std::uint32_t

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -852,6 +853,14 @@ bool CLuaEngineDefs::EngineAddClothingModel(CClientDFF* pDFF, std::string strMod
     return true;
 }
 
+bool CLuaEngineDefs::EngineSetClothingCacheTime(uint timeInMs)
+{
+    if (!g_pMultiplayer->SetClothingCacheTime(timeInMs))
+        return false;
+
+    return true;
```
FileEX:
```cpp
return g_pMultiplayer->SetClothingCacheTime(timeInMs);
```

Client/multiplayer_sa/CMultiplayerSA.h
```diff
@@ -345,6 +345,8 @@ class CMultiplayerSA : public CMultiplayer
 
     virtual void GetRwResourceStats(SRwResourceStats& outStats);
     virtual void GetClothesCacheStats(SClothesCacheStats& outStats);
+    virtual void ResetClothingCacheTime();
+    virtual bool SetClothingCacheTime(uint timeInMs);
```
FileEX: std::uint32_t

Client/multiplayer_sa/CMultiplayerSA_ClothesCache.cpp
```diff
@@ -402,6 +408,34 @@ void CMultiplayerSA::GetClothesCacheStats(SClothesCacheStats& outStats)
     outStats = ms_clumpStore.m_Stats;
 }
 
+//////////////////////////////////////////////////////////////////////////////////////////
+//
+// CMultiplayerSA::SetClothingCacheTime
+//
+//
+//////////////////////////////////////////////////////////////////////////////////////////
+bool CMultiplayerSA::SetClothingCacheTime(uint timeInMs)
```
FileEX: std::uint32_t

Client/sdk/multiplayer/CMultiplayer.h
```diff
@@ -455,6 +455,8 @@ class CMultiplayer
 
     virtual void GetRwResourceStats(SRwResourceStats& outStats) = 0;
     virtual void GetClothesCacheStats(SClothesCacheStats& outStats) = 0;
+    virtual void ResetClothingCacheTime() = 0;
+    virtual bool SetClothingCacheTime(uint timeInMs) = 0;
```
FileEX: std::uint32_t

FileEX (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
FileEX: Why was this conversation marked as resolved?

FileEX (dismissed):

Client/multiplayer_sa/CMultiplayerSA_ClothesCache.cpp
```diff
@@ -402,6 +408,34 @@ void CMultiplayerSA::GetClothesCacheStats(SClothesCacheStats& outStats)
...
+// CMultiplayerSA::SetClothingCacheTime
+//
+//
+//////////////////////////////////////////////////////////////////////////////////////////
+bool CMultiplayerSA::SetClothingCacheTime(std::uint32_t timeInMs)
+{
+    if (timeInMs == ms_clumpStore.m_minCacheTime)
+        return false;
+
+    ms_clumpStore.savedClumpList.clear();
```
tederis: Why do we need to clear it? I doesn't look safe because we still keep references to these clumps.

W3lac3: > Why do we need to clear it? I doesn't look safe because we still keep references to these clumps.

I put it up for cleaning because of the way it removes clumps that are no longer in use. And as for the reference, it really went unnoticed, thank you for giving that review.

Fernando-A-Rocha:
Nice!

PR #4357, opened 2025-08-12
Add functionality to re-stream models separately

Lamaaar:
`boolean engineRestreamModel( int modelId )` - re-stream a specific model using its ID

`boolean engineRestream( [ string restreamOption = "world" ] )` - re-stream a specific group of models (can be: "world", "vehicles", "peds", "objects")

These functions can be used instead of `engineRestreamWorld` (thus improving stability) if you only need to re-stream a specific (group of) model(s).

For example, instead of using `engineRestreamWorld` after using `engineImageLinkDFF` you can use `engineRestreamModel` to re-stream that single replaced model or `engineRestream("vehicles")` if you have used `engineImageLinkDFF` in bulk.


Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6748,7 +6748,7 @@ bool CClientGame::TriggerBrowserRequestResultEvent(const std::unordered_set<SStr
     return GetRootEntity()->CallEvent("onClientBrowserWhitelistChange", Arguments, false);
 }
 
-void CClientGame::RestreamModel(unsigned short usModel)
+bool CClientGame::RestreamModel(unsigned short usModel)
```
ghost:
```suggestion
bool CClientGame::RestreamModel(const std::uint16_t model)
```

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6748,7 +6748,7 @@ bool CClientGame::TriggerBrowserRequestResultEvent(const std::unordered_set<SStr
...
 {
     // Is this a vehicle ID?
     if (CClientVehicleManager::IsValidModel(usModel))
```
ghost:
```suggestion
    if (CClientVehicleManager::IsValidModel(model))
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -96,6 +96,8 @@ class CLuaEngineDefs : public CLuaDefs
     static eModelLoadState EngineStreamingGetModelLoadState(std::uint16_t modelId);
 
     static void EnginePreloadWorldArea(CVector position, std::optional<PreloadAreaOption> option);
+    static bool EngineRestreamModel(std::uint16_t modelId);
```
ghost, line 99:
```suggestion
    static bool EngineRestreamModel(const std::uint16_t modelId);
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2596,15 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+bool CLuaEngineDefs::EngineRestreamModel(std::uint16_t modelId)
```
ghost, line 2600:
```suggestion
bool CLuaEngineDefs::EngineRestreamModel(const std::uint16_t modelId)
```

ghost (changes requested):

ghost (approved):
Good job

tederis (dismissed):
Looks good

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6748,47 +6748,49 @@ bool CClientGame::TriggerBrowserRequestResultEvent(const std::unordered_set<SStr
     return GetRootEntity()->CallEvent("onClientBrowserWhitelistChange", Arguments, false);
 }
 
-void CClientGame::RestreamModel(unsigned short usModel)
+bool CClientGame::RestreamModel(const std::uint16_t model)
```
FileEX: What’s the point of making this function return a bool if it always returns true anyway?
void would be more appropriate here.

FileEX: The const here is unnecessary.

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -465,8 +465,9 @@ class CClientGame
     bool    IsHighFloatPrecision() const;
 
     bool TriggerBrowserRequestResultEvent(const std::unordered_set<SString>& newPages);
-    void RestreamModel(unsigned short usModel);
+    bool RestreamModel(const std::uint16_t model);
```
FileEX: The const here is unnecessary.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2596,15 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+bool CLuaEngineDefs::EngineRestreamModel(const std::uint16_t modelId)
```
FileEX: The const here is unnecessary.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2596,15 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+{
+    return g_pClientGame->RestreamModel(modelId);
+}
+
+bool CLuaEngineDefs::EngineRestream(std::optional<RestreamOption> option)
+{
+    g_pClientGame->Restream(option);
+
+    return true;
+}
```
FileEX: There’s no point in making the function return bool if it always returns the same value.
The function should be void, so it won’t return anything (just nil).

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2596,15 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+bool CLuaEngineDefs::EngineRestreamModel(const std::uint16_t modelId)
+{
+    return g_pClientGame->RestreamModel(modelId);
```
FileEX, line 2602: There’s no point in making the function return bool if it always returns the same value.
The function should be void, so it won’t return anything (just nil).

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -96,6 +96,8 @@ class CLuaEngineDefs : public CLuaDefs
     static eModelLoadState EngineStreamingGetModelLoadState(std::uint16_t modelId);
 
     static void EnginePreloadWorldArea(CVector position, std::optional<PreloadAreaOption> option);
+    static bool EngineRestreamModel(const std::uint16_t modelId);
```
FileEX: The const here is unnecessary.

Client/sdk/core/CClientBase.h
```diff
@@ -24,7 +24,7 @@ class CClientBase
     virtual void PreHUDRenderExecutionHandler(bool bDidUnminimize, bool bDidRecreateRenderTargets) = 0;
     virtual void PostFrameExecutionHandler() = 0;
     virtual void IdleHandler() = 0;
-    virtual void RestreamModel(unsigned short usModel) = 0;
+    virtual void RestreamModel(const std::uint16_t model) = 0;
```
FileEX: The const here is unnecessary.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -6808,6 +6810,55 @@ void CClientGame::RestreamWorld()
...
+                g_pClientGame->GetModelCacheManager()->OnRestreamModel(model->GetModelID());
+            }
+        }
+
+        m_pManager->GetObjectManager()->RestreamAllObjects();
+    }
+
+    if (option == RestreamOption::ALL)
+    {
+        m_pManager->GetPickupManager()->RestreamAllPickups();
```
FileEX: A pickup is an object, so maybe pickups should be restreamed together with objects?

FileEX (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2593,3 +2596,13 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+void CLuaEngineDefs::EngineRestreamModel(std::uint16_t modelId)
```
FileEX: What I meant is that a function which always returns the same value doesn’t make sense and it’s better to make it void.
However, in this case you can use bool, but you need to adjust `CClientGame::RestreamModel` so that it doesn’t always return true, but instead returns true or false depending on whether the model is valid.

Lamaaar: I originally had some plans that required the bool, but I dropped them, and it just accidentally stayed that way. You’re right - in its current form, it should be a void.

FileEX: > I originally had some plans that required the bool, but I dropped them, and it just accidentally stayed that way. You’re right - in its current form, it should be a void.

I’m not sure. I think this function should be a bool, because what if the passed ID turns out to be invalid?
Imo, in that case the function should return false or display some kind of warning, instead of just doing nothing.

Lamaaar: I've made it return bool.

FileEX (approved):

tederis (approved):

Meine1:
lgtm

Fernando-A-Rocha:
Missing wiki documentation for the new functions

PR #4360, opened 2025-08-14
Fixes #2680 - GetGarageSize returns wrong values

### LINKED ISSUE #2680: Function GetGarageSize returns wrong values
### How this actually works: ###
local **sizeZ**, sizeX, sizeY = getGarageSize (ID)
> Yes, sizeZ is being returned first than sizeX and sizeY.

> Also, sizeZ considers the garage Z position to 0 instead of its floor Z position. So if you try to create a colShape using 
getGaragePosition and getGarageSize, the colshape height will become higher than the garage.

### How this should work: ###
local sizeX, sizeY, **sizeZ** = getGarageSize (ID)
> Also:
> sizeZ = (current sizeZ) - posZ

### Steps to reproduce

1. Enable development mode with this (client-side):
```lua
setDevelopmentMode (true)
```
2. Run this script (client-side):
```lua
function getGarageColShape (id, fromZero)
    local sizeZ, sizeX, sizeY = getGarageSize (id) -- This function sucks and its first value is Z size. The Z size starts from 0 to the garage ceil.
    local posX, posY, posZ = getGaragePosition (id) -- This function uses the garage's floor Z position. When using getGarageSize() to create a colShape, set Z position to 0.
    if (fromZero) then
        return createColCuboid (posX, posY, 0, sizeX, sizeY, sizeZ) -- If you want to set colShape Z position to 0.
    else
        return createColCuboid (posX, posY, posZ, sizeX, sizeY, sizeZ - posZ) -- If you want to set colShape on garage's floor.
    end
end

setGarageOpen (9, true) -- Open CJ House Garage
getGarageColShape (9, true) -- Change to false if you wan't the colshape to start from the garage floor.
```
3. As you can see, the colshape is created perfectly with the code above. Now try with this code below (client-side):
```lua
local sizeX, sizeY, sizeZ = getGarageSize (9) -- How this function is supposed to be used.
local posX, posY, posZ = getGaragePosition (9)

setGarageOpen (9, true) -- Open CJ House Garage
createColCuboid (posX, posY, posZ, sizeX, sizeY, sizeZ) -- What's going on here? This should work well...
```
### HOW TO FIX: ###
- [ ] Change the returned values order. The first value should be the third value.
- [ ] Change how the Z size value is calculated. Just get the current value and subtract the garage Z position from it. (sizeZ - posZ)

### Version

Client: v1.5.9-release-21261 (Windows 10 Home, 64-bit)
Server: v1.5.9-release-21261 (Windows 10 Home, 64-bit)

### Additional context

_No response_

### Relevant log output

_No response_

### Security Policy

MohabCodeX:
Fixes #2680 

```
-- Before
Garage ID: 9
Size (X, Y, Z): 16.566600799561, 6.300048828125, 8.3499755859375
Position (X, Y, Z): 2502.3100585938, -1699.3599853516, 12.432299613953

-- After 
Garage ID: 9
Size (X, Y, Z): 6.300048828125, 8.3499755859375, 4.1343011856079
Position (X, Y, Z): 2502.3100585938, -1699.3599853516, 12.432299613953
```

- [x] Change the returned values order. The first value should be the third value.

- [x] Change how the Z size value is calculated. Just get the current value and subtract the garage Z position from it. (sizeZ - posZ)


Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -6676,6 +6676,11 @@ bool CStaticFunctionDefinitions::GetGarageSize(unsigned char ucGarageID, float&
     if (pGarage)
     {
         pGarage->GetSize(fHeight, fWidth, fDepth);
+        
+        CVector vecPosition;
+        pGarage->GetPosition(vecPosition);
+        fHeight = fHeight - vecPosition.fZ;
```
FileEX:
```cpp
fHeight -= vecPosition.fZ;
```

FileEX (changes requested):

FileEX (approved):

Issue #3915, opened 2024-12-31
Add Support for using Partial Animations and also being able to move


RughCuttle:
### Is your feature request related to a problem? Please describe.

MTA SA doesn't support you playing partial animations or any other animations and also be able to move at the same time. SA-MP has it implemented. 

### Describe the solution you'd like

A way to enable movement (WASD) while using partial animations with setPedAnimation or engineReplaceAnimation. Be able to still walk while using animations like gangtalk, handsign.

### Describe alternatives you've considered

Create a shader that does the animation for us (option based on Denihe approach, make it an MTA official resource), it would also solve another problem, which is max bones, therefore this would allow us to create more bones. It's more efficient/optimized as well. 

Make an function that works with partial animations in a task level. 
### Additional context

What would the end result ingame be like:
https://www.youtube.com/watch?v=ja-C9VcNyKw&t=85s

Example 2: 
https://youtu.be/akOPpQZ0b5A?t=309

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

Xenius97:
https://github.com/multitheftauto/mtasa-blue/issues/2733

RughCuttle:
I think this should be added as a project. Maybe it would get more attention.

FileEX:
Fixed in #5101

Issue #3055, opened 2023-06-03
Rifle_fire and all other weapon_fire errors


OguzArslan0000:
### Describe the bug

I wanted to change the weapon spin animation. When I changed the animation, other body functions were not working except the arms.

I tried hard to fix it but I couldn't. Only Weapon_Crouch works fine, others are corrupt.

Example video;  https://www.youtube.com/watch?v=gqyxviWVcJE

### Steps to reproduce

1-Create aim animation
2-Generate a code with EngineReplaceAnimation
3-Start the game and get the bug

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

PlatinMTA:
As stated in the wiki, you can't change partial animations.

FileEX:
Fixed in #5101

Issue #2733, opened 2022-09-11
Cannot replace weapon animations


Xenius97:
### Describe the bug

You can't replace any weapon animation, it will be bugged.

### Steps to reproduce

Start the resource:
[pistolanim.zip](https://github.com/multitheftauto/mtasa-blue/files/9542324/pistolanim.zip)

### Version

Multi Theft Auto v1.5.9-release-21261

### Additional context

Bug:
https://streamable.com/z9o4b1

It should be work like this:
https://youtu.be/Wm9UwrvvJEg

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

Haxardous:
> It should be work like this:
https://youtu.be/Wm9UwrvvJEg

According to that video it seems like it uses a mod that modifies the camera rotation matrix "GTA V aiming (Credits: DK22Pac and Astitva)" but the animation is working. 

Attempt to install the mod in singleplay without the required mod and see if it works to confirm.

Xenius97:
@Haxardous I would't be upload if not bugged. Works fine in Singleplayer without camera mod.
https://streamable.com/tq13bs

Xenius97:
Basic GTA IV animations converted to SA, works perfectly in Singleplayer.
https://streamable.com/dp9k8u
Mod: https://www.mediafire.com/file/996ziu5bgi2scuc/GTA+4+UPDATE+WEAPON.7z

PlatinMTA:
Pretty sure it's because that animation is a partial one, and you can't really change those as intended.

``It should be noted that partial animations are not supported, you can still replace them, but they won't work as intended, for example, "FightA_block" animation from "ped" block is a partial animation, you can't replace it properly, only a few animations are partial, rest of them are not, so it shouldn't be a problem.``
https://wiki.multitheftauto.com/wiki/EngineReplaceAnimation

TetsuDev:
![image](https://user-images.githubusercontent.com/101634036/189761684-d318b9a8-dc9f-44e0-a278-4ac060c41ed4.png)

Same issues with setWeaponProperty - "flag_aim_arm".

PR #4364, opened 2025-08-16
Fix no footsteps during walking animations (#4359)

### LINKED ISSUE #4359: onClientPedStep doesn't trigger when animation is applied
### Describe the bug

As in title.

### Steps to reproduce

```lua
local function onClientPedStep()
	iprint(eventName, source, getTickCount())
end
addEventHandler("onClientPedStep", root, onClientPedStep)
```

run this code, then do:
```lua
setPedAnimation(localPlayer, "CARRY", "crry_prtial", 0, true, true, false, true)
```
then try to walk around; cancel animation, and event will start to trigger.

### Version

/

### Additional context

/

### Relevant log output

```shell
/
```

### Security Policy

FileEX:
Fixed #4359

Some animations allow walking (e.g., `crry_partial`, `silence_reload`) or are entirely based on walking (e.g., `wuzi_walk`, `wuzi_cs_pt1`). During these animations, there are no footstep sounds, no footprints, no blood trails, and the `onClientPedStep`event is not triggered.

This PR, following the new guidelines, is based on reverse engineering - the `CPed::PlayFootSteps` function has been reversed and appropriately modified to fix the bug, instead of using an ASM hook.


tederis:
I just have tried it. And I have a question. Based on what did you choose animations `wuzi_walk` and `wuzi_cs_pt1`? And why do you spare other animations (from `ped` block, for example)? This solution is kinda... incomplete currently.

FileEX:
> I just have tried it. And I have a question. Based on what did you choose animations `wuzi_walk` and `wuzi_cs_pt1`? And why do you spare other animations (from `ped` block, for example)? This solution is kinda... incomplete currently.

It’s about partial animations that allow walking, and I only know of two: crry_prtial and silence_reload. There may be more, but I don’t know them.

Also, there are animations that consist only of walking and can be looped, namely wuzi_walk and wuzi_cs_pt1. I don’t know of any other animations that allow looping walking.

We’re not talking about animations where the character takes two steps and then sits down, but strictly those that are continuously about walking. There are probably more of these, but I only know those two. If you know more such animations, list them here and I’ll add them.

tederis:
> If you know more such animations, list them here and I’ll add them.

Some of looped animations from `ped` block:
```
run_1armed
run_armed
run_civi
run_csaw
run_fat
run_fatold
run_gang1
run_left
run_old
run_player
run_right
run_rocket
run_stop
run_stopr
run_wuzi
sprint_civi
sprint_panic
sprint_wuzi
swat_run
walk_armed
walk_civi
walk_csaw
walk_doorpartial
walk_drunk
walk_fat
walk_fatold
walk_gang1
walk_gang2
walk_old
walk_player
walk_rocket
walk_shuffle
walk_start
walk_start_armed
walk_start_csaw
walk_start_rocket
walk_wuzi
woman_run
woman_runbusy
woman_runfatold
woman_runpanic
woman_runsexy
woman_walkbusy
woman_walkfatold
woman_walknorm
woman_walkold
woman_walkpro
woman_walksexy
woman_walkshop
```

And that's not all the animations, you can find more.

RughCuttle:
Question, Would this draft solve #3915 , #2733 , #3055 ?

FileEX:
> Question, Would this draft solve #3915 , #2733 , #3055 ?

No

Issue #4363, opened 2025-08-16
/showmemstat does not display svgCreate objects


Serius41:
### Describe the bug

I don't know if this is a problem or if it was done intentionally, but images created with SVG are not reflected in the /showmemstat textures panel.

### Version

Multi Theft Auto v1.6-release-23324

### Additional context
https://streamable.com/g8wsqs

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/273e2dff-a738-4d02-b98d-d7bc7b837e6b" />

FileEX:
Fixed in https://github.com/multitheftauto/mtasa-blue/commit/7b236543b74fe77666321b060bc03b2bc2e7a50b

PR #4366, opened 2025-08-16
Update CLuaVectorGraphicDefs.cpp

Serius41:
https://github.com/multitheftauto/mtasa-blue/issues/4363

fixed


Client/mods/deathmatch/logic/luadefs/CLuaVectorGraphicDefs.cpp
```diff
@@ -156,7 +156,7 @@ CClientVectorGraphic* CLuaVectorGraphicDefs::SVGCreate(lua_State* luaVM, CVector
 
             if (!didLoad)
             {
-                delete vectorGraphic;
+                g_pClientGame->GetManager()->GetRenderElementManager()->Remove(vectorGraphic);
```
FileEX, line 159: There’s no need to do this here, since the CClientVectorGraphic destructor already handles it.

PR #4369, opened 2025-08-18
Fix ped sync flags order

FileEX:
Reimplement of #3965


botder:
Basically just a minor refactor and nothing really changed?

PR #4381, opened 2025-08-22
Refactor bulletsync and add checks

ghost:
Refactors bulletsync related code, fixes some related methods, adds new checks (ammo, range)


Dutchman101:
This was requested, code had more eyes on it than his, and it'll be tested further in nightlies.
Thanks!

PR #4384, opened 2025-08-22
New event onDamageEventCancelled

FileEX:
This PR adds a new event: `onDamageEventCancelled`.

```
element attacker, element damagedEntity, int weapon, float damage, string resourceName
```
**resourceName** - The name of the resource in which the event was cancelled.

The event is triggered when a client cancels (via `cancelEvent`) events such as `onClientPedDamage`, `onClientPlayerDamage`, and `onClientVehicleDamage`.

The **source** of the event is the source of the packet, i.e. the client where the event was cancelled.

## 1. Why?
It’s meant to help fight cheats related to godmode. Cheaters using executors either call `setElementHealth` on the localPlayer (for example in `onClientRender` - this soon can be blocked through PR #4376), or they cancel damage events such as `onClientPedDamage` or `onClientPlayerDamage`. I’m not entirely sure whether this event will be useful for vehicles - if not, it can be removed from `onClientVehicleDamage`.

## 2. Unnecessary packet sending / event flood
That’s why there’s an interval - one packet per second. This interval can be increased if one second is too short. The goal is mainly to prevent massive spam, e.g. when a player is on fire and takes constant damage. In one “cycle” of burning, the event is triggered exactly six times with the current interval. Maybe it should be increased?
For the same reason, the event is not triggered while a player is drowning. When oxygen drops to zero, the game attempts to subtract HP every frame, which would otherwise cause event flooding and unnecessary packet spam.

## 3. This doesn’t solve the problem
That’s true - this is not a direct fix for cheats. But we need to take into account that a huge portion of cheats aren’t memory modifications. They’re malicious Lua code executed through Lua executors. That’s exactly what cheaters use to modify vehicle handling, weapon properties, or to enable godmode.


ghost:
Would be great to be able to customise this event through [setServerConfigSetting](https://wiki.multitheftauto.com/wiki/SetServerConfigSetting).
- "cancelled_damage_send_frame_based_dmg" - enable/disable reporting about damage type/weapons which are called every frame (such as chainsaw, fire, fire extinguisher, teargas, lack of oxygen etc.); you might want to prioritize default weapons over that (for obvious reason, or simply reduce event call count)
- "cancelled_damage_send_interval" - how often client could notify server about damage being cancelled; i believe the hardcoded value wouldn't satisfy everyone, therefore it should be adjustable from Lua itself, so scripter could set it to own needs

Additionally, each type (player/ped/vehicle) should have own cooldown. Otherwise, with simple glue resource you will be lacking some of the reported data.

FileEX:
Three new settings have been added in mtaserver.conf:

- **cancelled_damage_for_vehicles** – Some servers may not need this, and to prevent unnecessary spam this option is disabled by default. It specifies whether onDamageEventCancelled should be triggered for vehicles (cancelling onClientVehicleDamage).

- **cancelled_damage_send_frame_based_dmg** – Specifies whether onDamageEventCancelled should be triggered by cancelling onClientPed/PlayerDamage when damage is applied every frame, e.g. from a minigun, fire, lack of oxygen underwater, chainsaw cuts, etc.

- **cancelled_damage_send_interval** – Specifies how often the onDamageEventCancelled event can be triggered.

Additionally, the interval is now checked per element, rather than globally since the last damage to a vehicle or ped.

sbx320:
It seems like a terrible idea to add even more events that get triggered by scripting.

FileEX:
> It seems like a terrible idea to add even more events that get triggered by scripting.

The current event system is terrible, that’s true, and I also have it on my list. I could also add an option in mtaserver.conf to completely disable this event.

samr46:
> The current event system is terrible, that’s true, and I also have it on my list. I could also add an option in mtaserver.conf to completely disable this event.

I think this is a good idea to add new server config option to disable it. Many projects cancel damage events by default (to implement custom damage system). So avoiding unnecessary overhead from this event will be useful.

FileEX:
Added the **cancelled_damage_for_peds** setting to mtaserver.conf. It’s disabled by default — I figured it’s better not to have such an event enabled by default.

PR #4385, opened 2025-08-23
Refactor framerate limiter to reduce CPU consumption

PerikiyoXD:
Extract and refactor FPSLimiter to use a better FPS limiting logic while reducing CPU consumption

~Queue-only ApplyFrameRateLimit, centralized pacing in ApplyQueuedFrameRateLimit~

~Caveats: Main menu runs with arbitrary FPS limit now (until we figure out hot to fps limit that codepath)~
- ~EnsureFrameRateLimitApplied no longer calls ApplyFrameRateLimit() from Present path to avoid skewed pacing; now only queues the target rate for the timer hook.~
- ~ApplyFrameRateLimit simplified to just queue the requested rate.~
- ~ApplyQueuedFrameRateLimit is now the single enforcement point: uses high-precision timing, adaptive sleep/yield calibration, and minimal spin to achieve stable pacing without wasting CPU.~
- ~Added detailed comments clarifying responsibilities and caveats (e.g. Main Menu not using timer hook).~

---

* [x] [your code follows the coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines)
* [x] your commit messages are informative ("update file.cpp" is unhelpful. "fix missing model in getVehicleNameFromModel" is helpful)


PerikiyoXD:
Tests on 5800X + 3080Ti

FPS Cap | 1.6 Release Build (CPU) | #4385 changes Debug Build (CPU)
-- | -- | --
144 | 7.2% ± 0.3% | 2.0% ± 0.2%
60 | 5.0% ± 0.2% | 1.0% ± 0.2%
30 | 2.8% ± 0.2% | 0.9% ± 0.2%
Uncapped | 9.3% @ 450fps | 7.8% @ 922fps

ArranTuna:
Your chart got my interest, so I took some readings of my own before and with your PR and... holy shit with FPS limited, CPU usage was approximately halved.

| FPS | CPU | GPU |
|--------|--------|--------|
| Normal unlimited (197) | 11% | 34% |
| PR unlimited (197) | 11% | 29% |
| Normal limited 60 | 9.2% | 25% |
| PR limited 60 | 4.7% | 18% | 

So for me didn't make any difference when using unlimited FPS. But amazing reduction in CPU and GPU usage when limited.

Is this CPU usage now freed up so that players with low performance systems are now potentially going to get a much better FPS because that CPU time is now available to actually compute things, or no? I dunno just seems to good to be true because so many MTA players don't have good PCs so this change would be monumental for them.

ArranTuna:
I've done some further testing, with some high CPU load:

crun setTimer(function() for i=1, 10000 do getElementData(localPlayer, "a") end end, 50, 0)

CPU and FPS were the same with and without this PR when under heavy load, when FPS was limited. However when FPS was set to a max of 60, it remained 60 FPS with the PR, but was 50 FPS on standard. Though is it possible although FPS may say it's higher it might not be better in the sense that the 10 extra frames may have tiny gaps between them and so some could end up being unbalanced in terms of feeling of smoothness.

PerikiyoXD:
> Is this CPU usage now freed up so that players with low performance systems are now potentially going to get a much better FPS because that CPU time is now available to actually compute things, or no?

Yes. The old limiter wasted CPU cycles in a busy-wait loop after finishing a frame. That didn’t steal time from MTA's or SA’s own logic/rendering, but it did keep one core artificially busy doing nothing.



TL;DR: I replaced a spinlock (Tight loop calling Sleep) with a mixed approach to not burn so many CPU cycles

> Your chart got my interest, so I took some readings of my own before and with your PR and... holy shit with FPS limited, CPU usage was approximately halved.
> FPS 	CPU 	GPU
> Normal unlimited (197) 	11% 	34%
> PR unlimited (197) 	11% 	29%
> Normal limited 60 	9.2% 	25%
> PR limited 60 	4.7% 	18%

It'd be nice to know CPU + GPU combo

ArranTuna:
> It'd be nice to know CPU + GPU combo

i7-10700K + GTX 1070

My readings were based on a rough average of what task manager said.

PerikiyoXD:
> Is this CPU usage now freed up so that players with low performance systems are now potentially going to get a much better FPS because that CPU time is now available to actually compute things, or no?

### Short answer: kinda, but it's more about not wasting CPU cycles than giving you free performance.

### Long answer:

The old frame limiter was genuinely awful (_Sorry OG code dev_). Look at this mess:

```cpp
while (true) {
    double dSpare = dTargetTimeToUse - m_FrameRateTimer.Get();
    if (dSpare <= 0.0) break;
    if (dSpare >= 10.0)
        Sleep(1);  // Only sleep if >= 10ms left, otherwise just spin
}
```

So let's walk through what happened on a typical frame. You want 30fps (33.3ms), game renders in, let's say 4ms. Now you need to kill 29ms somehow to fit the frame target. 

First it burns through 19ms calling `Sleep(1)` over and over... that's 19 separate system calls. Then when you're down to the last 9ms? It just sits there spinning the CPU, constantly checking `m_FrameRateTimer.Get()` in a tight loop until time runs out. This is the CPU burning I'm fixing.

Also, for lower FPS target, the more wasted cycles because you're waiting longer. 30fps wastes way more CPU than 144fps.

It's true that on lower end systems "render" time can be longer ms and it'd need to fill less, but the fact remains true for the remaining time.

The new approach is actually intelligent:

```cpp
// Learn the sleep overhead, then sleep for most of the wait in ONE call
if (remaining_ms > sleep_overhead + 0.5) {
    Sleep(remaining_ms - learned_overhead);
}
// For medium waits, yield instead of spinning  
else if (remaining_ms > yield_overhead + 0.05) {
    Sleep(0);
}
// Only spin for the final microseconds where precision matters
if (still_need_to_wait) {
    if (remaining > 0.02ms) {
        for (int i = 0; i < 10; i++) _mm_pause();  // Batched pauses
    } else {
        _mm_pause();  // Ultra-tight final loop
    }
}
```

So, ONE bulk Sleep, then Sleep(0) for medium waits, and finally _mm_pause() for the last few microseconds.

One of the key insights is that it measures how long Sleep() actually takes on your machine and it adapts. Instead of guessing "`Sleep(1)` takes 1ms", it tracks the real overhead and somewhat compensates.

Plus it uses `QueryPerformanceCounter` for microsecond precision instead of that janky millisecond timer.

Will this give you +90 FPS? Nah. But it stops your CPU from pointlessly burning cycles just to wait around. The freed up CPU time goes back to the OS for other stuff like better multitasking, less heat, more headroom for background processes.

If you were already CPU bottlenecked somewhere else, this won't magically fix that. But if frame limiting was eating cycles for no reason, yeah those cycles are available now.

> I dunno just seems to good to be true because so many MTA players don't have good PCs so this change would be monumental for them.

I get the skepticism but honestly this was just really bad code that got fixed.

Think about it: if you're running 30fps on a potato PC, the old limiter was literally doing 30+ unnecessary system calls per frame just to wait around, then spinning the CPU for whatever was left. 

That's like 900+ wasted Sleep() calls per second plus constant busy-waiting IN A SECOND.

On weak hardware it actually matters more because every wasted cycle hurts.

The improvement scales with how bad your FPS cap is...

So yeah, players stuck at 30fps because their PC sucks will see the biggest benefit. Not necessarily higher FPS, but way less pointless CPU waste that was competing with the actual game.

PerikiyoXD:
On the other part, there is this issue with how MTA tried to frame-limit... It had two codepaths and I disabled one, with no real impact BUT the main menu rendering limit, which doesn't use the CTimer as the game is "technically" not using the timer until it's connected. 

I need to find a proper way to fix the other codepath...

PlatinMTA:
More benchmarks if you care, ran at 1600x900 windowed
CPU: Ryzen 7 7800X3D
GPU: RTX 5070

| FPS Cap | 1.6 | 1.7 with PR |
| ------------- | ------------- | ------------- |
| Unlimited | 12%~ @ 550fps | 12%~ @ 550fps |
| 144 | 9,8%~ | 4,0%~ |
| 100 | 9,3%~ | 3,0%~ |
| 60 | 7,0%~ | 2,0%~ |
| 30 | 4,5%~ | 1,2%~ |

NOTE: I had to wait like 5 to 10 minutes before the CPU usage dropped on 1.6 release (before it was at least 10% usage all the time), meanwhile this didn't happen with the PR. More over this dropped my unlimited FPS from 550fps to 480fps.

No difference in FPS disabling the cap, but the CPU usage is noticeable smaller for the capped results.

ArranTuna:
That 10% constant usage you speak of is something I've seen before, it seems to be a random thing.

If this PR is complete I think you need to mark it as not a draft, so that it will get looked at.

PerikiyoXD:
> If this PR is complete I think you need to mark it as not a draft, so that it will get looked at.

Still a draft as it has a huge BUT. The main menu don't get proper limitations.

PerikiyoXD:
> That 10% constant usage you speak of is something I've seen before, it seems to be a random thing.

Probably _MTA doing might and magic checks for it's own safety_ related to anticheat measures or release nitpicks... Can't tell really as I didn't really want to deep dive and risk a ban.

For a more fair comparison you can try:

1. **Git clone and Build master**: Save as baseline. Keep the commit hash for reference.
2. **Apply patch ([4385.diff](https://github.com/multitheftauto/mtasa-blue/pull/4385.diff))**: Rebuild & save as patched.
3. **Run both under identical conditions**: Same server, same map. I used Map Editor for ease of use.
4. **Measure CPU usage**: idle + load. I used System Informer for coarse graph plot. 
> [!CAUTION]
> (**WARNING: System Informer triggers AC# 4! Adjust editor.conf if pertinent `<disableac>4</disableac>`**. Don't use on Release or you'll get kicked automatically or risk a permaban)
5. **Compare baseline vs patched**: And see if it's improved. Must be relatively near to current measurements.

PerikiyoXD:
I've finished the relevant changes. 
I added some goodies and TODOs that don't really harm and might need a little inspection for later refactors along the way.

# FPS metrics
> Tested idling in Ganton with no resources and no vehicles around. "stopall" basically.
## Main Menu
| FPS | CPU Usage |
|-----|-------------|
| 144FPS | 1.55%~ |
| 60FPS | 0.97%~ |
| 30FPS | 0.55%~ |

## Ingame
| FPS | CPU Usage |
|-----|-------------|
| 144FPS | 3.2%~ |
| 60FPS | 1.6%~ |
| 30FPS | 1.2%~ |

**Test Request:** @PlatinMTA @ArranTuna and anyone not mentioned that wants to help

PerikiyoXD:
## New Features

### VSync Frame Limiting
The game can now synchronize its frame rate with your monitor’s refresh rate.  

- **Enable:** `vsync = 1`  
- **Disable:** `vsync = 0`  
- **Default:** `1` (enabled)  

### New CVar: `vsync`
- **Purpose:** Locks the frame rate to the monitor’s refresh rate when enabled.  
- **Default:** `1`  

---

## Frame Limiter

Frame rate limits are applied in this priority order:

1. **Server Enforced** – Hard cap set by the server  
2. **Client Enforced** – Explicit cap set by the client  
3. **Client Preferred** – Preferred cap via `fps_limit` CVar  
4. **VSync** – Falls back to monitor refresh rate if no stricter cap applies  

**Rule:** The **lowest active limit** always takes precedence.  

### API Functions
- `setFPSLimit(limit)` – server or client enforced _(unchanged)_  
- `fps_limit` CVar – client preferred _(unchanged)_  
- `vsync` CVar – enables/disables monitor sync **_(new)_**  

---

## Examples
## Frame Limiter Examples

| Server | Client | Preferred | VSync (Hz) | Final FPS | Notes |
|--------|--------|-----------|------------|-----------|-------|
| 60     | -      | 144       | 144        | 60        | Server limit overrides all |
| 90     | 120    | 144       | 144        | 90        | Server (90) is lowest active limit |
| -      | 120    | 144       | 144        | 120       | Client enforced active, no server limit |
| -      | -      | 200       | 144        | 144       | VSync limits FPS to monitor refresh |
| -      | -      | -         | 75         | 75        | Only VSync active |
| -      | -      | 300       | Disabled   | 300       | Only client preferred limit active |
| 30     | 60     | 120       | 144        | 30        | Extreme server limit overrides all |
| 120    | 90     | 144       | 144        | 90        | Client enforced (90) is lowest |
| -      | -      | 90        | 60         | 60        | VSync lower than client preferred → VSync wins |
| -      | 150    | 120       | 144        | 120       | Client preferred (120) is lowest |
| 200    | 180    | 150       | 144        | 144       | VSync (144) is lowest among all limits |
| -      | -      | -         | Disabled   | Unlimited | No limits active, VSync off |

PerikiyoXD:
## Possible later improvements to add:

- Settings toggle for `vsync`

Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -2362,9 +2362,13 @@ void CPacketHandler::Packet_MapInfo(NetBitStreamInterface& bitStream)
...
-    unsigned short usFPSLimit = 36;
-    bitStream.ReadCompressed(usFPSLimit);
-    g_pCore->RecalculateFrameRateLimit(usFPSLimit);
+    auto readAndSetFPSLimit = [&]()
+    {
+        unsigned short usFPSLimit = 36; // Default to 36
+        bitStream.ReadCompressed(usFPSLimit); // See also: Client\mods\deathmatch\logic\rpc\CWorldRPCs.cpp:216
+        g_pCore->GetFPSLimiter()->SetServerEnforcedFPS(usFPSLimit);
+    };
+    readAndSetFPSLimit();
```
FileEX: There’s no need for that lambda here.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -1107,22 +1109,27 @@ void CClientGame::DoPulsePostFrame()
...
+// TODO: (pxd) Refactor this function to reduce its size and complexity
+// Pulse subsystems on post frame: Part 2
 void CClientGame::DoPulses()
 {
-    TIMING_CHECKPOINT("-CClientGame::DoPulsePostFrame");
-
-    g_pCore->ApplyFrameRateLimit();
+    auto defer = [&]() {
+        TIMING_CHECKPOINT("-CClientGame::DoPulsePostFrame2");
+    };
```
FileEX: Unrelated changes. For refactoring purposes, a separate PR should be opened. These changes go beyond the intended scope of this PR.

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -1024,7 +1026,7 @@ void CClientGame::DoPulsePostFrame()
 
                     if (zoneName == "Unknown")
                     {
-                        zoneName = _("Area 51");
+                        zoneName = _("Area 51");            // Easter egg!
```
FileEX: ??

Client/core/CCore.cpp
```diff
@@ -146,18 +146,16 @@ CCore::CCore()
...
     m_bDidRecreateRenderTargets = false;
     m_fMinStreamingMemory = 0;
     m_fMaxStreamingMemory = 0;
     m_bGettingIdleCallsFromMultiplayer = false;
     m_bWindowsTimerEnabled = false;
     m_timeDiscordAppLastUpdate = 0;
-    m_CurrentRefreshRate = 60;
+
+    // Initialize FPS limiter
+    m_pFPSLimiter = new FPSLimiter::FPSLimiter();
```
FileEX: std::unique_ptr?

Client/core/CCommands.cpp
```diff
@@ -83,160 +82,221 @@ bool CCommands::Execute(const char* szCommandLine)
...
+        return true;
+
+    // Handle nick command
+    // TODO: Make a proper builtin command entry for this
+    HandleNickCommand(command.c_str(), parameters.c_str(), bIsScriptedBind);
+
+    // Try external handler
+    if (ExecuteExternalHandler(command.c_str(), parameters.c_str(), bHandleRemotely, bIsScriptedBind))
+        return true;
+
```
FileEX: Unrelated changes. For refactoring purposes, a separate PR should be opened. These changes go beyond the intended scope of this PR.

Client/core/CClientVariables.h
```diff
@@ -28,8 +28,9 @@
 class CClientVariables : public CCVarsInterface, public CSingleton<CClientVariables>
 {
     // Sanity macros   << Who ever did this is idiot
-    #define SAN     if(!m_pStorage) return
-    #define SANGET  if(!Node(strVariable)) return false
+    // TODO: (pxd) Remove these macros and ODR violations
+    #define SAN     if(!m_pStorage) { OutputDebugLine("CClientVariables::Get: m_pStorage is NULL"); return; }
+    #define SANGET  if(!Node(strVariable)) { OutputDebugLine(SString("CClientVariables::Get: Node for cvar %s is NULL", strVariable.c_str())); return false; }
```
FileEX: Unrelated changes. For refactoring purposes, a separate PR should be opened. These changes go beyond the intended scope of this PR.

Client/core/CClientVariables.cpp
```diff
@@ -234,7 +235,20 @@ void CClientVariables::ValidateValues()
...
+    // So do it with our specialized FPS validation function
+    {
+        // Re-validate fps_limit cvar to ensure it's within valid range
+        // This is to catch users manually editing the config file with invalid values
+        // We don't clamp it here as we want to allow 0 (unlimited)
+        uint uiTemp;
+        CVARS_GET("fps_limit", uiTemp);
+        uiTemp = FPSLimiter::ValidateFPS(uiTemp);
+        CVARS_SET("fps_limit", uiTemp);
+    }
```
FileEX: Please clean up this

ghost:
> ## Possible later improvements to add:
> 
>     * Settings toggle for `vsync`

Would be quite useful and shouldn't be too hard to implement. That said, the average user likely isn't very familiar with CVARs or how to use them.

Client/core/CClientVariables.h
```diff
@@ -119,15 +120,15 @@ class CClientVariables : public CCVarsInterface, public CSingleton<CClientVariab
...
+    bool IsLoaded() { return m_bLoaded; }            // TODO: (pxd) Remove if unused? Seems unused
     int  GetRevision() { return m_iRevision; }
     void ValidateValues();
 
 private:
     CXMLNode* Node(const std::string& strVariable);
     void      LoadDefaults();
 
-    bool      m_bLoaded;
+    bool      m_bLoaded;            // TODO: (pxd) Remove if unused? Seems unused
```
FileEX: Unrelated changes. For refactoring purposes, a separate PR should be opened. These changes go beyond the intended scope of this PR.

Client/core/CCommands.cpp
```diff
@@ -247,25 +307,21 @@ void CCommands::Delete(const char* szCommand)
...
 {
-    // Delete all the commands
     list<COMMANDENTRY*>::iterator iter = m_CommandList.begin();
     for (; iter != m_CommandList.end(); iter++)
     {
         delete *iter;
     }
     m_CommandList.clear();
 
-    // Re-register our commands
```
FileEX: Unrelated changes. For refactoring purposes, a separate PR should be opened. These changes go beyond the intended scope of this PR.

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto v1.7
```
FileEX: We no longer use versioning in comments.
```
 ```

File: Client/core/FPSLimiter.cpp [hunk 11]
```diff
@@ -0,0 +1,442 @@
...
+
+#include "StdInc.h"
+#include "FPSLimiter.h"
+#include "CGraphStats.h"
+#include <core/FPSLimiterInterface.h>
+
+namespace FPSLimiter
+{
+
+    FPSLimiter::FPSLimiter(uint32_t initialPreferredFPS) : m_data{0, 0, ValidateFPS(initialPreferredFPS), ValidateFPS(initialPreferredFPS), 0, 0, false, false}
```
FileEX: std::uint32_t

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+    {
+        OutputDebugLine("FPSLimiter::Reset");
+        m_data.serverEnforcedFPS = 0;
+        m_data.clientEnforcedFPS = 0;
+        m_data.userDefinedFPS = 0;
+        // m_data.displayRefreshRate = 0; // We do not reset display refresh rate
+        CalculateCurrentFPSLimit();
+    }
+
+    uint32_t FPSLimiter::GetFPSTarget()
```
FileEX: std::uint32_t

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+            case EnforcerType::UserDefined:
+                return "UserDefined";
+            case EnforcerType::Client:
+                return "Client";
+            case EnforcerType::Server:
+                return "Server";
+            default:
+                return "Unknown";
+        }
+    }
```
FileEX: Use EnumToString instead

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+            double qpcTicks = (double)(end.QuadPart - start.QuadPart);
+            double tscTicks = (double)(tscEnd - tscStart);
+            tscFrequency = (tscTicks / qpcTicks) * m_frequency.QuadPart;
+        }
+
+        // Calculate target frame time in TSC ticks
+        const double targetFrameTime = tscFrequency / m_data.activeFPSTarget;
+
+        // Calculate elapsed time since last frame
+        double elapsedTicks = (double)(currentTSC - m_lastFrameTSC);
```
FileEX: static_cast

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+            }
+
+            // Final precision loop
+            do
+            {
+                _mm_pause();
+                lastMeasuredTSC = __rdtsc();
+            } while (lastMeasuredTSC < targetTSC);
+
+            // OutputDebugLine(SString("FPSLimiter: Finished %s RDTSC spin at %llu TSC, expected %llu TSC, overshoot %lld TSC", context, lastMeasuredTSC,
```
FileEX: Please clean up this

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+                        if (SetWaitableTimer(m_hTimer, &dueTime, 0, nullptr, nullptr, FALSE))
+                        {
+                            WaitForSingleObject(m_hTimer, INFINITE);
+                            frameEndTSC = rdtscSpinWait(targetTSC);//, "precision finish");
+
+                            // uint64_t postTimerTSC = __rdtsc();
+                            // int64_t  overshoot = (int64_t)(postTimerTSC - targetTSC);
+                            // double   overshootMs = (double)overshoot * 1000.0 / tscFrequency;
+                            // OutputDebugLine(SString("FPSLimiter: Finished precision wait finished at %llu TSC, expected %llu TSC, overshoot %lld TSC (%.2f ms)",
+                            //                        postTimerTSC, targetTSC, overshoot, overshootMs));
```
FileEX: Please clean up this

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+        // m_data.displayRefreshRate = 0; // We do not reset display refresh rate
+        CalculateCurrentFPSLimit();
+    }
+
+    uint32_t FPSLimiter::GetFPSTarget()
+    {
+        return m_data.activeFPSTarget;
+    }
+
+    EnforcerType FPSLimiter::GetEnforcer()
```
FileEX: const noexcept?

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,442 @@
...
+    {
+        // If unlimited, no enforcer
+        if (m_data.activeFPSTarget == FPS_LIMIT_UNLIMITED)
+            return EnforcerType::None;
+
+        // Find which source provided the lowest FPS (the active target)
+        uint32_t     minFPS = UINT32_MAX;            // Set to max for comparison
```
FileEX: std::numeric_limits

FileEX (changes requested):

FileEX:
> ## Possible later improvements to add:
> * Settings toggle for `vsync`

Since you've added such a CVAR, it's your responsibility to add the corresponding setting in the menu. I think the PR is incomplete without it.

Client/mods/deathmatch/logic/CPacketHandler.cpp
PerikiyoXD: I was pointing out diff between ReadCompressed and Read usage between CWorldRPCs and CPacketHandler

Client/mods/deathmatch/logic/CClientGame.cpp
PerikiyoXD: QoL that should be just added

Client/core/CClientVariables.h
PerikiyoXD: Future work comments. One PR for noting this is excessive?

Client/core/FPSLimiter.cpp
PerikiyoXD: Then it should be changed in all touched files for cohesion

PerikiyoXD: For now I'll assume new files get this treatment

Client/core/CCommands.cpp
PerikiyoXD: QoL stuff.

Client/core/CClientVariables.h
PerikiyoXD: QoL stuff.

Client/mods/deathmatch/logic/CPacketHandler.cpp
PerikiyoXD: Removed

Client/core/CCommands.cpp
PerikiyoXD: CCommands refactoring will be extracted, will replace with foul dirty hack instead of cleaner CCommands.

Client/core/CCommands.cpp
PerikiyoXD: CCommands refactoring will be extracted, will replace with foul dirty hack instead of cleaner CCommands.

Client/core/CClientVariables.h
PerikiyoXD: Pointing out AOI shouldn't be removed

Client/mods/deathmatch/logic/CClientGame.cpp
PerikiyoXD: Deferred to #4428

Client/core/CClientVariables.h
PerikiyoXD: Deferred to #4427

Client/core/CClientVariables.h
PerikiyoXD: Deferred to #4427

Client/core/CCommands.cpp
PerikiyoXD: Deferred to #4427

Client/core/CCommands.cpp
PerikiyoXD: Deferred to #4427

PerikiyoXD:
Added VSync option to settings:
imagen

Client/core/CCommands.cpp
```diff
@@ -172,10 +172,19 @@ bool CCommands::Execute(const char* szCommand, const char* szParametersIn, bool
...
-                    m_FpsLimitTimer.Reset();
+                    int fpsVal = 0;
+                    CVARS_GET("fps_limit", fpsVal);
+                    CCore::GetSingleton().GetFPSLimiter()->SetUserDefinedFPS(fpsVal);
+                }
+
+                // HACK: Foul dirty hack to force vsync (Rework on #4427)
+                if (key == "vsync")
+                {
+                    bool bVSync;
```
FileEX: Hungarian notation

Client/core/CCore.cpp
```diff
@@ -1831,141 +1831,16 @@ void CCore::ApplyCoreInitSettings()
...
-    TIMING_GRAPH("");
+    // NOTE: (pxd) We are handling the frame limiting updates
+    // earlier in the callpath (CModManager::DoPulsePreFrame, CModManager::DoPulsePostFrame)
 }
 
-//
-// Frame rate limit (wait) is done here.
-//
-void CCore::ApplyQueuedFrameRateLimit()
+void CCore::OnFPSLimitChange(uint32_t uiFPS)
```
FileEX: std::uint32_t & hungarian notation

Client/core/CCore.h
```diff
@@ -107,6 +108,7 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
     CTrayIconInterface*                GetTrayIcon() { return m_pTrayIcon; };
     std::shared_ptr<CDiscordInterface> GetDiscord();
     CSteamClient*                      GetSteamClient() { return m_steamClient.get(); }
+    FPSLimiter::FPSLimiterInterface*   GetFPSLimiter() { return m_pFPSLimiter.get(); }
```
FileEX: const noexcept?

Client/core/CCore.h
```diff
@@ -220,13 +222,9 @@ class CCore : public CCoreInterface, public CSingleton<CCore>
...
 
-    uint GetFrameRateLimit() { return m_uiFrameRateLimit; }
-    void RecalculateFrameRateLimit(uint uiServerFrameRateLimit = -1, bool bLogToConsole = true);
-    void ApplyFrameRateLimit(uint uiOverrideRate = -1);
-    void ApplyQueuedFrameRateLimit();
-    void EnsureFrameRateLimitApplied();
-    void SetClientScriptFrameRateLimit(uint uiClientScriptFrameRateLimit);
-    void SetCurrentRefreshRate(uint value);
+    // FPS Limiter
+    void OnFPSLimitChange(uint32_t uiFPS);            // Called when FPS limit changes
```
FileEX: std::uint32_t & hungarian notation

Client/core/CModManager.cpp
```diff
@@ -55,30 +55,46 @@ bool CModManager::TriggerCommand(const char* commandName, size_t commandNameLeng
...
+                Stop();
+                break;
+            default:
+                break;
+        }
+    };
+
+    TIMING_GRAPH("+DoPulsePostFrame");
+
+    handleStateChange();            // Handle state changes before pulse
```
FileEX, line 90:
```cpp
if (m_state == State::PendingStart)
    Start();
else if (m_state == State::PendingStop)
    Stop();
```

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,399 @@
...
+#pragma region Interface
+
+    void FPSLimiter::Reset()
+    {
+        OutputDebugLine("FPSLimiter::Reset");
+        m_data.serverEnforcedFPS = 0;
+        m_data.clientEnforcedFPS = 0;
+        m_data.userDefinedFPS = 0;
+
+        bool bVSync;
```
FileEX: Hungarian notation

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,399 @@
...
+        TIMING_GRAPH("LimiterStart");
+        auto defer = [&]() { TIMING_GRAPH("LimiterEnd"); };
+
+        if (m_data.activeFPSTarget == FPS_LIMIT_UNLIMITED)
+        {
+            return defer();
+        }
+        else if (m_data.appliedThisFrame)
+        {
+            OutputDebugLine("Warning: FPSLimiter::SetFrameRateThrottle called when already applied this frame");
```
FileEX, line 194: If this kind of information is needed, shouldn’t it be displayed only in the debug build? I mean all outputDebugLine in this PR

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,399 @@
...
+            return lastMeasuredTSC;
+        };
+
+        // Set timer resolution once per session
+        auto ensureTimerResolution = []()
+        {
+            static bool resolutionSet = false;
+            if (!resolutionSet)
+            {
+                typedef LONG(NTAPI * NtSetTimerResolution)(ULONG, BOOLEAN, PULONG);
```
FileEX: using

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,399 @@
...
+    ADD_ENUM(EnforcerType::None, "none")
+    ADD_ENUM(EnforcerType::VSync, "vsync")
+    ADD_ENUM(EnforcerType::UserDefined, "user-defined")
+    ADD_ENUM(EnforcerType::Client, "client")
+    ADD_ENUM(EnforcerType::Server, "server")
+    IMPLEMENT_ENUM_END("EnforcerType");
+
+};            // namespace FPSLimiter
+
+std::uint32_t FPSLimiter::ValidateFPS(std::uint32_t uiFPS)
```
FileEX: Hungarian notation

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -7044,19 +7044,21 @@ bool CStaticFunctionDefinitions::SetMoonSize(int iSize)
     return false;
 }
 
-bool CStaticFunctionDefinitions::SetFPSLimit(int iLimit)
+bool CStaticFunctionDefinitions::SetClientFPSLimit(int iFPS)
```
FileEX: Hungarian notation

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -7044,19 +7044,21 @@ bool CStaticFunctionDefinitions::SetMoonSize(int iSize)
...
-    return false;
+    g_pCore->GetFPSLimiter()->SetClientEnforcedFPS(iFPS);
+    return true;
+}
+
+bool CStaticFunctionDefinitions::GetFPSLimit(int& iFPS) 
+{
+    iFPS = g_pCore->GetFPSLimiter()->GetFPSTarget();
+    return true;
 }
```
FileEX: There’s no reason for this function to be bool since it always returns true.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -7044,19 +7044,21 @@ bool CStaticFunctionDefinitions::SetMoonSize(int iSize)
...
 
-bool CStaticFunctionDefinitions::GetFPSLimit(int& iLimit)
+bool CStaticFunctionDefinitions::SetServerFPSLimit(int iFPS)
```
FileEX: Hungarian notation

Client/core/FPSLimiter.h
```diff
@@ -0,0 +1,82 @@
...
+        // API
+        void Reset();
+
+        std::uint32_t GetFPSTarget() const noexcept;
+        EnforcerType  GetEnforcer() const noexcept;
+
+        void SetServerEnforcedFPS(std::uint32_t frameRateLimit);
+        void SetClientEnforcedFPS(std::uint32_t frameRateLimit);
+        void SetUserDefinedFPS(std::uint32_t frameRateLimit);
+        void SetDisplayVSync(bool enabled);
```
FileEX: override?

Client/core/FPSLimiter.cpp
```diff
@@ -0,0 +1,399 @@
...
+            if (!resolutionSet)
+            {
+                typedef LONG(NTAPI * NtSetTimerResolution)(ULONG, BOOLEAN, PULONG);
+                HMODULE ntdll = GetModuleHandleA("ntdll.dll");
+                if (ntdll)
+                {
+                    auto proc = GetProcAddress(ntdll, "NtSetTimerResolution");
+#pragma warning(push)
+#pragma warning(suppress : 4191)
+                    auto setRes = reinterpret_cast<NtSetTimerResolution>(proc);
```
FileEX: std::bit_cast and remove pragma rules

FileEX (changes requested):

Client/core/CCore.h
PerikiyoXD: This is a bit pedantic.
Wasn't required for CSteamClient? 1 month ago? okay, where is consistency here?

Client/core/CCore.cpp
PerikiyoXD: Surrounding context asks for hungarian notation. Will change anyway.

Client/core/CCore.h
FileEX: Steam support wasn’t introduced via a PR; if it had been, the same suggestion would have appeared there as well.

PerikiyoXD: Bad things happen when coding guidelines aren’t enforced outside PRs. Remember a week ago? I’ll comply, but this is my warning about the issue.

Client/core/CCore.cpp
FileEX: This is your code; I only marked the parts edited by you. When writing new code, we follow the guidelines, so you shouldn’t argue, “Hungarian notation is used everywhere else, so it’s fine here too.”. I’m not asking you to edit the entire file, just adjust your changes to comply with the guidelines and current standards. You may not like it, but if you want your changes to be accepted, you have to follow the rules of this repository. They might not appeal to you, but imo, you must respect them.

PerikiyoXD: Understood. New code don't care about preexisting coding style.

Client/core/CCore.h
FileEX: If a function can be marked as const and noexcept, it should be. Why isn’t this the case for Steam Client? Perhaps it was forgotten or considered unimportant at the time. I’m sure that if such a suggestion were made, Botder wouldn’t object to adding the const and noexcept specifiers.

Also, I posed this as a question, so you could simply decline the change and wait for review/merge from other developers. Generally, we don’t block good PRs over minor details like this, but I don’t think this is such a difficult or costly change that it even warrants discussion.

Client/core/CCore.cpp
Fernando-A-Rocha: > Understood. New code don't care about preexisting coding style.

Yeah man it's true. The goal is to get rid of Hungarian notation altogether. But writing new code u aren't require to refactor other things unrelated to your changes. If u wish to fix preexisting code you'd need to make a separate PR afaik

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
PerikiyoXD:
```
    static void GetFPSLimit(int& iLimit);
    static void SetClientFPSLimit(int iFPS);
    static void SetServerFPSLimit(int iFPS);
```
    Too?
    Or do I hardcode `return true;`?

PerikiyoXD:
[At this point, as the server spits a short](https://wiki.multitheftauto.com/wiki/SetFPSLimit)
> Starting from version [[r21313](https://buildinfo.mtasa.com/?Revision=21313&Branch)] and above fpsLimit range is 25-32767. In older MTA releases it was 25-100.

Should frame limiter use `std::uint16_t` instead of `std::uint32_t`?

(Also why signed? Negative values why?)

Client/core/FPSLimiter.cpp
PerikiyoXD: It's done automatically:
```cpp
#ifdef MTA_DEBUG
void OutputDebugLine(const char* szMessage);
void SetDebugTagHidden(const SString& strTag, bool bHidden = true);
bool IsDebugTagHidden(const SString& strTag);
#else
inline void OutputDebugLineDummy() {}
inline void SetDebugTagHiddenDummy() {}
inline bool IsDebugTagHiddenDummy() { return false; }
    #define OutputDebugLine(x) OutputDebugLineDummy ()
    #define SetDebugTagHidden(x) SetDebugTagHiddenDummy ()
    #define IsDebugTagHidden(x) IsDebugTagHiddenDummy ()
#endif
```

Client/core/CModManager.cpp
botder: I don't really understand why you had to wrap a simple if-else-if with a lambda

PerikiyoXD: Deduplication? idk

PerikiyoXD:
Rebased branch to provide a clean linear history on top of latest master

PerikiyoXD:
Tested locally, compiles successfully, server runs normally, and all modified functionality works as expected... It feels done.

ArranTuna:
I've tested it.

Saves a lot of CPU when minimized. CPU usage is 1% instead of 4%.

Dutchman101:
Thanks @PerikiyoXD, this is huge. I've asked for an additional MTA team opinion and we'll be merging it now, so it can be tested before the release of 1.7

PR #4390, opened 2025-08-25
Fix argument parser not handling noexcept functions

sbx320:


sbx320:
yay msvc... I guess noexcept is not supported by the arg parser for now

PR #4392, opened 2025-08-25
Add onElementPositionChange event to SetElementPosition for anti-cheat compatibility

frqher:
This PR adds a server-side "onElementPositionChange" event to the SetElementPosition function.

**Reason:**
Some anti-cheat scripts detect large position changes as cheating, but do not distinguish between server-initiated and client-initiated changes. This causes false positives when the server teleports players (e.g., for elevators or scripted events).

**How it works:**
By triggering the "onElementPositionChange" event server-side, anti-cheat scripts can listen for this event and safely ignore position changes that originate from the server.

This improves compatibility with anti-cheat resources and prevents legitimate server-side teleports from being flagged as cheats.

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -1672,6 +1672,7 @@ void CGame::AddBuiltInEvents()
     m_Events.AddEvent("onElementInteriorChange", "oldInterior, newInterior", nullptr, false);
     m_Events.AddEvent("onElementAttach", "attachSource, attachOffsetX, attachOffsetY, attachOffsetZ, attachOffsetRX, attachOffsetRY, attachOffsetRZ", nullptr, false);
     m_Events.AddEvent("onElementDetach", "detachSource, detachWorldX, detachWorldY, detachWorldZ, detachWorldRX, detachWorldRY, detachWorldRZ", nullptr, false);
+    m_Events.AddEvent("onElementPositionChange", "oldX, oldY, oldZ, newX, newY, newZ", nullptr, false);
 
     // Radar area events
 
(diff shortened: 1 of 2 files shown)


MohabCodeX:
I’m not sure, but isn’t [onPlayerTeleport](https://wiki.multitheftauto.com/wiki/OnPlayerTeleport) enough?
But maybe `onElementPositionChange` would be useful for general cases to deal with `SetElementPosition`

frqher:
> I’m not sure, but isn’t [onPlayerTeleport](https://wiki.multitheftauto.com/wiki/OnPlayerTeleport) enough? But maybe `onElementPositionChange` would be useful for general cases to deal with `SetElementPosition`

onPlayerTeleport is indeed sufficient for detecting sudden (teleport-like) position changes, especially those initiated by the client.
However, an event like onElementPositionChange can be useful for more general cases, as it allows you to detect all position changes made via server-side scripts (e.g., when using setElementPosition). This is not limited to players, but can also apply to vehicles, peds, objects, etc.

tederis:
Gamemode scripts shall not allow cheaters to call server-side setElementPosition. Otherwise it indicates that a gamemode was designed with mistakes. I personally don't like the idea of adding tons of events in trying to fix poorly designed scripts.

blonter:
There are multiple issues with `onPlayerTeleport`. It doesn’t trigger when it should, and when it does, it can still produce false positives. More info can be found on Discord. There’s been a lot of discussion about these problems there. 

It’s important to ensure this event only triggers for element position changes initiated on the client side, since we trust the server.

Once everything is fixed, the current `onPlayerTeleport` should be marked as deprecated.

frqher:
> Gamemode scripts shall not allow cheaters to call server-side setElementPosition. Otherwise it indicates that a gamemode was designed with mistakes. I personally don't like the idea of adding tons of events in trying to fix poorly designed scripts.



dasd

tzwer:
> dasd

Check these out
- https://github.com/multitheftauto/mtasa-resources/pull/663
- https://github.com/multitheftauto/mtasa-resources/pull/664

Issue #1282, opened 2020-03-08
getCameraMatrix does not return up-to-date FoV value


quiret:
**Describe the bug**
When calling `local sx, sy, sz, tx, ty, tz, roll, fov = getCameraMatrix()` and inspecting the fov value, then this value is not up-to-date.

Symptoms:
1) setCameraFieldOfView does not synchronize with getCameraMatrix FoV
2) accelerating in a vehicle increases the FoV, but not the value returned by getCameraMatrix
3) camera calculation code is wrong if FoV changes from default value

**To reproduce**
1) the scripting way (if not in vehicle):
```
setCameraFieldOfView("player", 90)

setTimer(function()
    local _,_,_,_,_,_,_,fov = getCameraMatrix()
    assert(fov == 90)
end, 2000, 1)
```

2) the gameplay way: enter a vehicle, accelerate very fast, then execute the following command handler (clientside):
```
addCommandHandler("getfov", function()
    local _,_,_,_,_,_,_,fov = getCameraMatrix()
    outputChatBox("current field of view is " .. fov);
end)
```

**Expected behaviour**
The field of view is supposed to adjust to the camera's real value instead of being default. If FoV is changed by script to 90 getCameraMatrix should also return the interpolating value.

**Version**
MTA:SA 1.5.7

qaisjp:
Good point. I'd say [[set](https://wiki.multitheftauto.com/wiki/SetCameraFieldOfView)|[get](https://wiki.multitheftauto.com/wiki/GetCameraFieldOfView)]CameraFieldOfView should return the static setting, but [getCameraMatrix](https://wiki.multitheftauto.com/wiki/SetCameraMatrix) the actual 'realtime' FOV.

Quick research shows that, client side, the `fov` param does this:

```cpp
fFOV = m_pCamera->GetFOV();
```

And `CClientCamera::GetFOV()` is this:

```cpp
float GetFOV() { return m_fFOV; }
```

Doesn't seem like we use `CCam::GetFOV()` anywhere (but we do use `CCam::SetFOV()`, so I guess this might be enough:

```cpp
CCam* pCam = m_pCamera->GetCam(m_pCamera->GetActiveCam());
pCam->GetFOV();
```

Server side only works for fixed cameras, so not relevant.

quiret:
Good idea, qaisjp! If you can provide me with a nightly build with the fix inserted then I am willing to test it.

qaisjp:
[...]

Issue #1284, opened 2020-03-08
Camera Matrix inaccurate when entering/leaving vehicle


quiret:
**Describe the bug**
If you enter or exit a vehicle then GTA:SA performs camera interpolation for a smooth transition. During this camera state, calling getElementMatrix on the camera element returns a wrong matrix. Instead of returning the interpolated matrix, it returns the result of interpolation.

**Video**
https://www.youtube.com/watch?v=orM45pHPiXA

**To reproduce**
1) join a local server and start this resource: 
[screen-border-detect.zip](https://github.com/multitheftauto/mtasa-blue/files/4303635/screen-border-detect.zip)
2) teleport to the middle of the map (close to 0, 0, 0)
3) spawn a vehicle near you
4) look at the white rectangle and enter the vehicle. you should see the white rectangle leave the hayroll object as long as the camera is interpolating.

**Expected behaviour**
The white rectangle should not leave the center of the hayroll object. Then the camera matrix is the same as used by GTA:SA at all times.

**Version**
MTA:SA 1.5.7

**Remarks**
The issue about the rectangle leaving the hayroll on vehicle acceleration has been addressed by https://github.com/multitheftauto/mtasa-blue/issues/1282 already.

qaisjp:
Maybe we could add an extra argument `realtime=false` to `getCameraMatrix` that, when set to `true`, will return the CCam results rather than 'target values'

quiret:
Well, the getCameraMatrix function is not mentioned in this issue, but it could suffer from the same problem, as demonstrated by #1282. Do you have a plan how to return this 'realtime' matrix? Also I think introducing this realtime parameter would just overcomplicate things for the scripter. Like what justification does non-realtime have if it is just worse then realtime.

qaisjp:
> Like what justification does non-realtime have if it is just worse then realtime.

getCameraTargetLookAt/TargetPosition does not exist — this is what getCameraMatrix is supposed to provide (or _currently_ provides)

changing fov return value is OK because we can get the 'target' fov using get/setCameraFieldOfView

> getCameraMatrix function is not mentioned in this issue

Ah, now I see, your example uses `getElementMatrix(getCamera())`. I suppose we can update the camera object matrix getter to get the realtime version.

The question now is whether it makes sense for `getElementMatrix(getCamera())` to return different underlying data as `getCameraMatrix`.

quiret:
[...]

PR #4393, opened 2025-08-26
Fix camera behavior and `getCameraMatrix` return values (#1282 & #1284)

MohabCodeX:
This PR fixes issue #1282 by reading the actual rendering FOV instead of using the individual camera instance, and addresses #1284 by returning the interpolated matrix instead of the target matrix during transitions.

`getCameraMatrix()` now reports FOV values (including Roll) that match the actual rendering, and the camera matrix during vehicle entry is smooth and accurate. However, on vehicle exit there is a \~1 second delay (still better than the current behavior), which seems consistent with normal game behavior (until now i see it's maybe not a bug due to directional timing differences and different interpolation fractions between entering and exiting a vehicle, try to reproduce #1284 with the script in same issue to understand what i mean .. I can provide a comparison video later). I don’t consider this a bug, though replacing the hooks might eliminate the delay entirely (so hard to refactor camera behavior :D)

It also removes server-side restrictions to ensure `getCameraMatrix()` works correctly across all camera states.


Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include <StdInc.h>
 #include <game/CCam.h>
+#include "../../game_sa/CCameraSA.h"
```
FileEX: I don’t like this. In general, we try not to mix projects with each other. The overall layer is like this:

Client Deathmatch -> Client SDK (virtual) -> CGameSA

So in `CCamera.h` you can add a virtual method that:

Returns an interface or a bool indicating whether it exists.

The same goes for FOV - instead of `pCameraInterface->FOVDuringInter;` you can create a method and use it.

Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -644,3 +648,65 @@ void CClientCamera::ResetShakeCamera() noexcept
 {
     m_pCamera->ResetShakeCamera();
 }
+
+bool CClientCamera::IsInCameraTransition() const
+{
+    if (!m_pCamera)
+        return false;
+    
+    return m_pCamera->GetTransitionState() != 0;
```
FileEX:
```cpp
return m_pCamera ? m_pCamera->GetTransitionState() != 0 : false;
```

Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -644,3 +648,65 @@ void CClientCamera::ResetShakeCamera() noexcept
...
+        return 70.0f;
+    
+    CCameraSAInterface* pCameraInterface = static_cast<CCameraSA*>(m_pCamera)->GetInterface();
+    if (!pCameraInterface)
+        return 70.0f;
+    
+    if (IsInCameraTransition())
+        return pCameraInterface->FOVDuringInter;
+    
+    return *(float*)0x8D5038;
```
FileEX: Always document such addresses. Either define them as `constexpr std::uintptr_t` or add a comment. We don’t want such magic values.

https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md#avoid-magic-numbers.

Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -644,3 +648,65 @@ void CClientCamera::ResetShakeCamera() noexcept
...
+}
+
+float CClientCamera::GetAccurateFOV() const
+{
+    if (!m_pCamera)
+        return 70.0f;
+    
+    CCameraSAInterface* pCameraInterface = static_cast<CCameraSA*>(m_pCamera)->GetInterface();
+    if (!pCameraInterface)
+        return 70.0f;
```
FileEX: What do you think about defining this as DEFAULT_FOV?

FileEX (changes requested):

Client/game_sa/CCameraSA.cpp
```diff
@@ -15,6 +15,9 @@
 #include <mutex>
 #include <cmath>
 
+// Camera constants
+constexpr float DEFAULT_FOV = 70.0f;
```
FileEX: You can define this in CCamera.h, then it will be available both in Client Deathmatch and in Game_SA.

Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -13,6 +13,10 @@
 
 #define PI_2 6.283185307179586476925286766559f
 
+// Camera FOV constants
+constexpr float DEFAULT_FOV = 70.0f;
```
FileEX: You can define this in CCamera.h, then it will be available both in Client Deathmatch and in Game_SA.

Client/mods/deathmatch/logic/CClientCamera.cpp
```diff
@@ -13,6 +13,10 @@
...
+constexpr std::uintptr_t VAR_CurrentCameraFOV = 0x8D5038; // CCamera::CurrentFOV
```
FileEX: constexpr const?

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "StdInc.h"
 #include <game/CAnimManager.h>
+#include <game/CCam.h>
```
FileEX: Do you need this?

Client/game_sa/CCameraSA.h
```diff
@@ -414,6 +414,9 @@ class CCameraSA : public CCamera
     void                ShakeCamera(float radius, float x, float y, float z) noexcept override;
     void                ResetShakeCamera() noexcept override;
     std::uint8_t        GetTransitionState() override;
+    bool                IsInTransition() override;
+    float               GetTransitionFOV() override;
+    bool                GetTransitionMatrix(CMatrix& matrix) override;
```
FileEX: const

Client/game_sa/CCameraSA.cpp
```diff
@@ -566,3 +569,49 @@ std::uint8_t CCameraSA::GetTransitionState()
...
+
+float CCameraSA::GetTransitionFOV()
+{
+    CCameraSAInterface* pInterface = GetInterface();
+    return pInterface ? pInterface->FOVDuringInter : DEFAULT_FOV;
+}
+
+bool CCameraSA::GetTransitionMatrix(CMatrix& matrix)
+{
+    CCameraSAInterface* pInterface = GetInterface();
```
FileEX: pInterface -> cameraInterface

Client/game_sa/CCameraSA.cpp
```diff
@@ -566,3 +569,49 @@ std::uint8_t CCameraSA::GetTransitionState()
...
 }
+
+bool CCameraSA::IsInTransition()
+{
+    return GetTransitionState() != 0;
+}
+
+float CCameraSA::GetTransitionFOV()
+{
+    CCameraSAInterface* pInterface = GetInterface();
```
FileEX: pInterface -> cameraInterface

FileEX (changes requested):

Client/game_sa/CCameraSA.h
```diff
@@ -381,7 +381,7 @@ class CCameraSA : public CCamera
     CCameraSA(CCameraSAInterface* cameraInterface);
     ~CCameraSA();
 
-    CCameraSAInterface* GetInterface() { return internalInterface; };
+    CCameraSAInterface* GetInterface() const { return internalInterface; };
```
FileEX: noexcept?

Client/game_sa/CCameraSA.cpp
```diff
@@ -15,6 +15,7 @@
 #include <mutex>
 #include <cmath>
 
+
```
FileEX: Extra line

Client/sdk/game/CCamera.h
```diff
@@ -17,6 +17,9 @@ class CMatrix;
 class CCam;
 struct RwMatrix;
 
+// Camera constants
+constexpr float DEFAULT_FOV = 70.0f;
```
FileEX: constexpr const

FileEX (changes requested):

FileEX (approved):

ghost (approved):

PR #4403, opened 2025-08-30
Add new format string function for non-constant string

G-Moris:
Error:
```c++
message = std::format(_("Connecting to {0}:{1} ..."), m_strHost, m_usPort);
```

Decision:
```c++
message = mtasa::format(_("Connecting to {0}:{1} ..."), m_strHost, m_usPort);
```

I modified `CConnectManager.cpp` to include an example of how it works. If that’s a problem, I can remove the changes I made to it.


botder:
With the introduction of placeholders `{}` you should start using the index, for example `{0}` to refer to arguments, when you're formatting a translatable string. This allows translators to reposition the arguments without breaking the formatting operation.

PR #4411, opened 2025-09-03
Add ability to make client-side vehicles damageable

MegadreamsBE:
This PR makes it possible to mark a client-side vehicle to be damageable. By default client-side vehicles cannot receive damage, this PR does not change that default behavior.

When **false** is passed to **SetVehicleDamageProof** for a local vehicle, it will now become damageable.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicle.cpp
+++ b/Client/mods/deathmatch/logic/CClientVehicle.cpp
@@ -35,7 +35,7 @@ std::set<const CClientEntity*> ms_AttachedVehiclesToIgnore;
 #define VEHICLE_INTERPOLATION_WARP_THRESHOLD            15
 #define VEHICLE_INTERPOLATION_WARP_THRESHOLD_FOR_SPEED  10
 
-CClientVehicle::CClientVehicle(CClientManager* pManager, ElementID ID, unsigned short usModel, unsigned char ucVariation, unsigned char ucVariation2)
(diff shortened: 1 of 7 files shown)


Fernando-A-Rocha:
I think the createVehicle function has enough arguments. Wouldn't it be wise to make them be damageable with like setVehicleDamageProof false?

MegadreamsBE:
> I think the createVehicle function has enough arguments. Wouldn't it be wise to make them be damageable with like setVehicleDamageProof false?

It crossed my mind, but I'd still need to have two variables internally anyways... felt silly in that regards. I could of course change the default behavior, but many servers probably wouldn't appreciate that.

I'll see what I can do.

MegadreamsBE:
> > I think the createVehicle function has enough arguments. Wouldn't it be wise to make them be damageable with like setVehicleDamageProof false?
> 
> It crossed my mind, but I'd still need to have two variables internally anyways... felt silly in that regards. I could of course change the default behavior, but many servers probably wouldn't appreciate that.
> 
> I'll see what I can do.

I could just set the variable that handles damage proof to true by default for a local vehicle.... Okay, I'll make the change

MegadreamsBE:
@Fernando-A-Rocha Updated! Your message made me realize the easy solution to making it work through SetVehicleDamageProof. Thanks!

Fernando-A-Rocha (approved):
This seems ok to me, less complicated than the initial concept 😅 but other devs can share their opinion

PR #4413, opened 2025-09-04
Add engineFramerateFixingSetProperty/GetProperty/ResetProperties

thisdp:
For adjustable vehicle physics time step and air resistor physics when glitch "vehicle_rapid_stop" is disabled. In this way, DM players can play maps in any FPS.

Usage:
engineFramerateFixingSetProperty("vehicle_physics", 50.0/30.0) which is the default one.

Now we have only "vehicle_physics", but in future, there maybe more.

I notice that "vehicle_rapid_stop" acually includes vehicle turning (passive and active).
Passive turning includes terrain changing and colliding.
Active turning includes air control and normal on-road turning.
Thus, description of "vehicle_rapid_stop" should be "framerate_related_vehicle_physics" for better naming.

Maybe the name and usage of this function need to be discussed.

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -15,6 +15,7 @@
 #include <game/CStreaming.h>
 #include <game/CPtrNodeSingleLinkPool.h>
 #include <lua/CLuaFunctionParser.h>
+#include <lua/CLuaFunctionParseHelpers.h>
 #include "CLuaEngineDefs.h"
 #include <enums/VehicleType.h>
 
@@ -2612,33 +2613,27 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
 }
 
 
-bool CLuaEngineDefs::EngineFramerateFixingResetProperties()
+void CLuaEngineDefs::EngineFramerateFixingResetProperties()
 {
     g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(0);   //use default, Should we reset this when player disconnects?
-    return true;
 }
 
-bool CLuaEngineDefs::EngineFramerateFixingSetProperty(std::string strPropertyName, float timestep)
+void CLuaEngineDefs::EngineFramerateFixingSetProperty(eFramerateFixingProperty propertyName, float timestep)
 {
-    if (strPropertyName.compare("vehicle_physics") == 0)
+    switch (propertyName)
     {
-        g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(timestep);
+        case eFramerateFixingProperty::FFP_VEHICLE_PHYSICS:
+            g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(timestep);
+            break;
     }
-    else
-    {
-        throw std::invalid_argument("Unsupported property name at argument 1");
-    }
-    return true;
 }
 
-float CLuaEngineDefs::EngineFramerateFixingGetProperty(std::string strPropertyName)
+std::variant <bool, float> CLuaEngineDefs::EngineFramerateFixingGetProperty(eFramerateFixingProperty propertyName)
 {
-    if (strPropertyName.compare("vehicle_physics") == 0)
+    switch (propertyName)
     {
-        return g_pMultiplayer->FramerateFixingGetPhysicsTimeStep();
+        case eFramerateFixingProperty::FFP_VEHICLE_PHYSICS:
+            return g_pMultiplayer->FramerateFixingGetPhysicsTimeStep();
     }
-    else
-    {
-        throw std::invalid_argument("Unsupported property name at argument 1");
-    }
+    return false;
 }
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -207,7 +207,6 @@ void CLuaEngineDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "framerateFixingGetProperty", "engineFramerateFixingGetProperty");
     lua_classfunction(luaVM, "framerateFixingResetProperties", "engineFramerateFixingResetProperties");
 
-
     lua_registerstaticclass(luaVM, "Engine");
 
     // `EngineStreaming` class
@@ -2618,21 +2617,21 @@ void CLuaEngineDefs::EngineFramerateFixingResetProperties()
     g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(0);   //use default, Should we reset this when player disconnects?
 }
 
-void CLuaEngineDefs::EngineFramerateFixingSetProperty(eFramerateFixingProperty propertyName, float timestep)
+void CLuaEngineDefs::EngineFramerateFixingSetProperty(FramerateFixingProperty propertyName, float timestep)
 {
     switch (propertyName)
     {
-        case eFramerateFixingProperty::FFP_VEHICLE_PHYSICS:
+        case FramerateFixingProperty::FFP_VEHICLE_PHYSICS:
             g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(timestep);
             break;
     }
 }
 
-std::variant <bool, float> CLuaEngineDefs::EngineFramerateFixingGetProperty(eFramerateFixingProperty propertyName)
+std::variant <bool, float> CLuaEngineDefs::EngineFramerateFixingGetProperty(FramerateFixingProperty propertyName)
 {
     switch (propertyName)
     {
-        case eFramerateFixingProperty::FFP_VEHICLE_PHYSICS:
+        case FramerateFixingProperty::FFP_VEHICLE_PHYSICS:
             return g_pMultiplayer->FramerateFixingGetPhysicsTimeStep();
     }
     return false;
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -2611,7 +2611,6 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
 
-
 void CLuaEngineDefs::EngineFramerateFixingResetProperties()
 {
     g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(0);   //use default, Should we reset this when player disconnects?
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -150,6 +150,9 @@ void CLuaEngineDefs::LoadFunctions()
         {"engineGetPoolUsedCapacity", ArgumentParser<EngineGetPoolUsedCapacity>},
         {"engineSetPoolCapacity", ArgumentParser<EngineSetPoolCapacity>},
         {"enginePreloadWorldArea", ArgumentParser<EnginePreloadWorldArea>},
+        {"engineFramerateFixingSetProperty", ArgumentParser<EngineFramerateFixingSetProperty>},
+        {"engineFramerateFixingGetProperty", ArgumentParser<EngineFramerateFixingGetProperty>},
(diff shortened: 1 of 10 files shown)


Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -742,11 +745,23 @@ static void __declspec(naked) HOOK_VehicleRapidStopFix()
...
+    // Just change time step, will be automatically applied when related hook is installed
+    // If time step is not set, kOriginalTimeStep will be used as default;
+    if (timestep == 0)
+        timestep = kOriginalTimeStep;
+    kPhysicTimeStep = timestep;
+}
+
+float CMultiplayerSA::FramerateFixingGetPhysicsTimeStep(){
+    return kPhysicTimeStep;
+}
```
FileEX: const noexcept

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -742,11 +745,23 @@ static void __declspec(naked) HOOK_VehicleRapidStopFix()
...
     }
 }
 
+void CMultiplayerSA::FramerateFixingSetPhysicsTimeStep(float timestep){
+    // Just change time step, will be automatically applied when related hook is installed
+    // If time step is not set, kOriginalTimeStep will be used as default;
+    if (timestep == 0)
+        timestep = kOriginalTimeStep;
+    kPhysicTimeStep = timestep;
+}
```
FileEX: Please use the format:
```cpp
void foo()
{
}
```

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
```diff
@@ -97,6 +97,10 @@ class CLuaEngineDefs : public CLuaDefs
 
     static void EnginePreloadWorldArea(CVector position, std::optional<PreloadAreaOption> option);
 
+    static bool EngineFramerateFixingResetProperties();
+    static bool EngineFramerateFixingSetProperty(std::string strPropertyName, float timestep);
```
FileEX: These functions should be void (returning nil), since they always return true anyway. There’s no reason for them to be bool.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2602,3 +2610,35 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
...
+bool CLuaEngineDefs::EngineFramerateFixingSetProperty(std::string strPropertyName, float timestep)
+{
+    if (strPropertyName.compare("vehicle_physics") == 0)
+    {
+        g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(timestep);
+    }
+    else
+    {
+        throw std::invalid_argument("Unsupported property name at argument 1");
+    }
```
FileEX: Instead of std::string, use an enum here. Create the enum and add it to the definition in CLuaFunctionParseHelper. Then the parser will handle everything and return an error if the type is invalid.

FileEX (changes requested):

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.h
thisdp: these are lua functions, should return true if function executes successfully

FileEX: > these are lua functions, should return true if function executes successfully

In this case, the function always completes successfully, so it doesn’t need to return true, because using it in a condition makes no sense. With the new parser, all new functions that always return true are simply void and return nil. There’s no reason for a function to be bool if it always returns the same value.

thisdp: > > these are lua functions, should return true if function executes successfully
> 
> In this case, the function always completes successfully, so it doesn’t need to return true, because using it in a condition makes no sense. With the new parser, all new functions that always return true are simply void and return nil. There’s no reason for a function to be bool if it always returns the same value.

ooh, that makes sense. I didn't know that. Thanks

thisdp: > > these are lua functions, should return true if function executes successfully
> 
> In this case, the function always completes successfully, so it doesn’t need to return true, because using it in a condition makes no sense. With the new parser, all new functions that always return true are simply void and return nil. There’s no reason for a function to be bool if it always returns the same value.

But I found it returns nil instead of "true" when it success.

FileEX: And that’s exactly how it should be.

thisdp: > And that’s exactly how it should be.

oh, ok. now I fixed everything

Shared/sdk/enums/FramerateFixingProperty.h
```diff
@@ -0,0 +1,17 @@
...
+ *  FILE:        sdk/FramerateFixingProperty.h
+ *  PURPOSE:     Header for common definitions
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
+
+#pragma once
+
+enum class eFramerateFixingProperty
```
FileEX: eFramerateFixingProperty -> FramerateFixingProperty

Server/mods/deathmatch/logic/CCommon.h
```diff
@@ -1068,6 +1068,11 @@ enum eStats
     STAT_MAYBE_WU_ZI_MEETING = 342
 };
 
+enum class eFramerateFixingProperty
+{
+    FFP_VEHICLE_PHYSICS = 0
+};
+
```
FileEX: Delete this

Client/sdk/multiplayer/CMultiplayer.h
```diff
@@ -465,6 +465,10 @@ class CMultiplayer
     virtual eAnimID    GetLastStaticAnimationID() = 0;
     virtual DWORD      GetLastAnimArrayAddress() = 0;
 
+    virtual void        FramerateFixingSetPhysicsTimeStep(float timestep) = 0;
+    virtual float       FramerateFixingGetPhysicsTimeStep() const noexcept = 0;
+
+
```
FileEX: Extra line

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -742,11 +745,26 @@ static void __declspec(naked) HOOK_VehicleRapidStopFix()
...
+
+float CMultiplayerSA::FramerateFixingGetPhysicsTimeStep() const noexcept
+{
+    return kPhysicTimeStep;
+}
+
+
```
FileEX: Extra line

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -742,11 +745,26 @@ static void __declspec(naked) HOOK_VehicleRapidStopFix()
...
         jmp RETURN_VehicleRapidStopFix
     }
 }
 
+void CMultiplayerSA::FramerateFixingSetPhysicsTimeStep(float timestep)
+{
+    // Just change time step, will be automatically applied when related hook is installed
+    // If time step is not set, kOriginalTimeStep will be used as default;
+    if (timestep == 0)
+        timestep = kOriginalTimeStep;
```
FileEX: Please add new line below

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -12,9 +12,12 @@
 
 static bool         bWouldBeNewFrame = false;
 static unsigned int nLastFrameTime = 0;
-
 constexpr float kOriginalTimeStep = 50.0f / 30.0f;
 
+//
+static float kPhysicTimeStep = kOriginalTimeStep;
+
```
FileEX, line 21: Extra line

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -12,9 +12,12 @@
 
 static bool         bWouldBeNewFrame = false;
 static unsigned int nLastFrameTime = 0;
-
 constexpr float kOriginalTimeStep = 50.0f / 30.0f;
 
+//
```
FileEX: ?

Client/multiplayer_sa/CMultiplayerSA.h
```diff
@@ -365,6 +365,11 @@ class CMultiplayerSA : public CMultiplayer
     unsigned int EntryInfoNodePool_NoOfUsedSpaces() const noexcept override;
     unsigned int PtrNodeDoubleLinkPool_NoOfUsedSpaces() const noexcept override;
 
+    //Framerate Fixing Property
+    void  FramerateFixingSetPhysicsTimeStep(float timestep) override;
+    float FramerateFixingGetPhysicsTimeStep() const noexcept override;
+
+
```
FileEX: Extra line

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -199,6 +203,11 @@ void CLuaEngineDefs::AddClass(lua_State* luaVM)
     lua_classfunction(luaVM, "setModelTXDID", "engineSetModelTXDID");
     lua_classfunction(luaVM, "resetModelTXDID", "engineResetModelTXDID");
 
+    lua_classfunction(luaVM, "framerateFixingSetProperty", "engineFramerateFixingSetProperty");
+    lua_classfunction(luaVM, "framerateFixingGetProperty", "engineFramerateFixingGetProperty");
+    lua_classfunction(luaVM, "framerateFixingResetProperties", "engineFramerateFixingResetProperties");
+
+
```
FileEX: Extra line

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2602,3 +2611,29 @@ void CLuaEngineDefs::EnginePreloadWorldArea(CVector position, std::optional<Prel
     if (option == PreloadAreaOption::ALL || option == PreloadAreaOption::COLLISIONS)
         g_pGame->GetStreaming()->LoadSceneCollision(&position);
 }
+
+
+void CLuaEngineDefs::EngineFramerateFixingResetProperties()
+{
+    g_pMultiplayer->FramerateFixingSetPhysicsTimeStep(0);   //use default, Should we reset this when player disconnects?
```
FileEX: Yes, you should also reset it when the player joins the server

FileEX (changes requested):

tzwer:
Does this make it possible to have the same vehicle_rapid_stop behaviour when it's not enabled?

thisdp:
> Does this make it possible to have the same vehicle_rapid_stop behaviour when it's not enabled?

Only vehicle_rapid_stop glitch is disabled.
And maybe we need to change the name "vehicle_rapid_stop" to "vehicle_physics_framerate_dependent" later.

Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
thisdp: done

Lpsd:
something like `engineSetPhysicsProperty` would be much nicer. if you need more words/explanation then use the individual properties/keys for that.

thisdp:
> something like `engineSetPhysicsProperty` would be much nicer. if you need more words/explanation then use the individual properties/keys for that.

This may be not just physics.

Lpsd:
if it will always be used for timestep related properties, then `engineSetTimestep(string property, float step)`?

thisdp:
> if it will always be used for timestep related properties, then `engineSetTimestep(string property, float step)`?

Maybe a good idea for short naming

thisdp:
To achieve this, maybe we need more hooker. From the test of @xLive , this PR is not enough for DM physics.
https://github.com/gta-reversed/gta-reversed/blob/c2df58461d2df0f455530132ebefb63348bb9275/source/game_sa/Entity/Physical.cpp
ProcessCollision
ProcessShift
ApplyTurnSpeed
ApplyMoveSpeed
ApplyGravity
ApplySoftCollision
ApplySpringCollision
ApplySpringCollisionAlt
ApplySpringDampening
ApplyCollisionAlt
ApplyFriction
PlacePhysicalRelativeToOtherPhysical
PositionAttachedEntity
ApplySpeed

Issue #2312, opened 2021-08-19
onPlayerWasted receiving incorrect parameters


stoneage-mta:
**Describe the bug**
Currently, if we use `setElementHealth(player, 0)`, the parameters of `onPlayerWasted` will be incorrect. 
(similar to #2273)

**To reproduce**
<details><summary>Code used to test:</summary>
<p>

#### Client-side:
```lua
addEventHandler('onClientPlayerWasted', root, function(killer, weapon, bodypart, stealth)
    outputConsole(inspect{
        'onClientPlayerWasted:',
        "killer:", killer,
        "weapon:", weapon,
        "bodypart:", bodypart,
        "stealth:", stealth,
    })
end)
```
#### Server-side:
```lua
addEventHandler('onPlayerWasted', root, function(totalAmmo, killer, weapon, bodypart, stealth)
    outputConsole(inspect{
        'onPlayerWasted:',
        "totalAmmo:", totalAmmo,
        "killer:", killer,
        "weapon:", weapon,
        "bodypart:", bodypart,
        "stealth:", stealth
    })
end)
```
</p>
</details>

Results: 
![image](https://user-images.githubusercontent.com/38073795/130007317-462114a0-eb9e-46d1-ad91-9d8802b7ce13.png)

**Expected behaviour**
I would expect the parameters to be the same as those resulting in `onClientPlayerWasted`.

**Version**
Server: MTA:SA Server v1.5.8-release-20927
Client: Multi Theft Auto v1.5.8-release-20927

Dutchman101:
It claims that the player has died by Explosion (damage type 63) rather than being killed by a script for any reason using setElementHealth, and this is more severe than the inconsistency between bodypart that's also part of this issue.

stoneage-mta:
and everything gets a little more confusing, here are the parameters returned in the following situations...

### killPed(player)
```lua
onClientPlayerWasted:
    killer: false, weapon: 63, bodypart: 7, stealth: false
onPlayerWasted:
    totalAmmo: 1, killer: false, weapon: false, bodypart: false, stealth: false
```

### setElementHealth(player, 0)
```lua
onClientPlayerWasted:
    killer: false, weapon: false, bodypart: false, stealth: false
onPlayerWasted: 
    totalAmmo: 0, killer: false, weapon: 63, bodypart: 7, stealth: false
```

MohabCodeX:
Fixed #4418

Issue #2273, opened 2021-07-05
onClientPlayerWasted doesn't returns valid parameters when dying in vehicle explosion


ghost:
**Describe the bug**
Client-side event doesn't return valid data, while it's counter-part at server-side does it correctly.

**To reproduce**
1. Paste code (from below) in client & server-script.
2. Die in vehicle explosion.
3. Profit.

Code:
```lua
-- Client

function onClientPlayerWasted(...)
	local t = {...}

	iprint("Client: "..inspect(t))
end
addEventHandler("onClientPlayerWasted", root, onClientPlayerWasted)

-- Server

function onPlayerWasted(...)
	local t = {...}

	iprint("Server: "..inspect(t))
end
addEventHandler("onPlayerWasted", root, onPlayerWasted)
```

**Expected behaviour**
It should return valid data.

**Screenshots**
![obraz](https://user-images.githubusercontent.com/51768772/124499815-f3476800-ddbe-11eb-85b8-ca380e3682e1.png)

**Version**
Client/server version: v1.5.8-release-20884

**Additional context**
/

MohabCodeX:
Fixed #4418

PR #4418, opened 2025-09-04
Death Event Parameters Synchronization Fix

MohabCodeX:
This PR fixes issues #2312  and #2273  where `onPlayerWasted` and `onClientPlayerWasted` events received inconsistent parameters when players died via `setElementHealth` or vehicle explosions.

**Root Cause**: Race condition between health RPC packets and client-side death events caused `DoWastedCheck()` to fire with stale damage data, resulting in inconsistent parameters between server and client death events.

**Solution**: Added client-side flag `m_bServerProcessedDeath` to prevent stale data usage and implemented dynamic weapon data retrieval.

<details>
<summary><strong>Results</strong></summary>

```lua
-- Client

function onClientPlayerWasted(...)
	local t = {...}

	iprint("Client: "..inspect(t))
end
addEventHandler("onClientPlayerWasted", root, onClientPlayerWasted)

-- Server

function onPlayerWasted(...)
	local t = {...}

	iprint("Server: "..inspect(t))
end
addEventHandler("onPlayerWasted", root, onPlayerWasted)
````

```lua
crun setElementHealth(localPlayer,"0")
```

```
[13:15:37] [debug] mark\client.lua:4 - "Client: { false, 0, 3, false, 0, 15 }"
[13:15:37] [debug] mark\server.lua:4 - "Server: { 0, false, 0, 3, false, 0, 15 }"
```

**Vehicle Explosion**

```
mark\client.lua:4 - "Client: { false, 51, 3, false, 0, 15 }"
mark\server.lua:4 - "Server: { 2972, false, 51, 3, false, 0, 15 }"
```

</details>

### This also incidentally fixed an additional bug:

The inconsistent value of `63` that appeared when dying from a vehicle explosion. It was hardcoded and unrelated to the correct ID, which is `51`. Please note that the information on the wiki will need to be updated accordingly. **I will provide the full list of damage types later.**

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -108,7 +108,7 @@ CClientGame::CClientGame(bool bLocalPlay) : m_ServerInfo(new CServerInfo())
     m_ucDamageWeapon = WEAPONTYPE_INVALID;
     m_ulDamageTime = 0;
     m_bDamageSent = true;
-    m_bServerProcessedDeath = false;
+    m_serverProcessedDeath = false;
     m_bShowNetstat = false;
     m_bShowFPS = false;
     m_bHudAreaNameDisabled = false;
@@ -1467,7 +1467,7 @@ void CClientGame::DoPulses()
         {
             // Only call DoWastedCheck if server hasn't already processed this death
             // This prevents duplicate events when server processes death via unified context system
-            if (!m_bServerProcessedDeath) {
+            if (!m_serverProcessedDeath) {
                 DoWastedCheck(m_DamagerID, m_ucDamageWeapon, m_ucDamageBodyPiece);
             }
         }
@@ -4513,7 +4513,7 @@ bool CClientGame::ApplyPedDamageFromGame(eWeaponType weaponUsed, float fDamage, 
                     GetDeathAnim(pDamagedPed, pEvent, animGroup, animID);
 
                     // Check if we're dead
-                    if (!m_bServerProcessedDeath) {
+                    if (!m_serverProcessedDeath) {
                         DoWastedCheck(damagerID, weaponUsed, hitZone, animGroup, animID);
                     }
                 }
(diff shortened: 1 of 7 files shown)


MohabCodeX:
This is likely to be backward-incompatible because of the explosion type ID change.

Fernando-A-Rocha:
Perfect fix for 1.7 😄

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -770,6 +811,7 @@ class CClientGame
     unsigned char  m_ucDamageBodyPiece;
     unsigned long  m_ulDamageTime;
     bool           m_bDamageSent;
+    bool           m_bServerProcessedDeath = false;  // Flag to track server-processed deaths
```
FileEX:
```cpp
bool m_serverProcessedDeath{false}; // Flag to track server-processed deaths
```

Client/mods/deathmatch/logic/CClientGame.h
```diff
@@ -405,6 +405,44 @@ class CClientGame
...
+    void SetScriptedDeathData() noexcept {
+        const auto* pLocalPlayer = GetLocalPlayer();
+        if (!pLocalPlayer) {
+            m_DamagerID = INVALID_ELEMENT_ID;
+            m_ucDamageWeapon = WEAPONTYPE_INVALID;
+            m_ucDamageBodyPiece = BODYPART_INVALID;
+            m_ulDamageTime = CClientTime::GetTime();
+            m_bServerProcessedDeath = false;
+            return;
+        }
```
FileEX: GetTime and GetLocalPlayer aren’t noexcept, so this function shouldn’t be either. If a function marked as noexcept calls other functions that aren’t noexcept, any potential crash will provide no information (app will be terminate)

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -7144,3 +7165,22 @@ void CClientGame::AudioZoneRadioSwitchHandler(DWORD dwStationID)
...
+//
+//////////////////////////////////////////////////////////////////
+unsigned char CClientGame::TryGetCurrentWeapon(const CClientPlayer* pPlayer) const noexcept
+{
+    if (!pPlayer)
+        return WEAPONTYPE_UNARMED;
+        
+    // Remove const to call non-const method
+    auto* player = const_cast<CClientPlayer*>(pPlayer);
+    const auto weaponType = player->GetCurrentWeaponType();
```
FileEX: There’s no point in using const in such cases. This variable only exists until the function exits. Additionally, the type isn’t explicitly specified here, for example via static_cast, so the full type should be used instead of auto (according to guidelines).

FileEX (changes requested):

FileEX (approved):

PR #4419, opened 2025-09-04
Fix crash when requesting CJ clothing model

W3lac3:
Sometimes when requesting a clothing model, the game would crash due to an infinite loop. This PR solves this problem. The system was functioning normally, but this error could eventually occur. After receiving reports, I reviewed the code and noticed an incorrect use of static that had gone unnoticed.

I recorded a short video demonstrating a system I made using this new feature. Watch the video below:

https://github.com/user-attachments/assets/fa0e72fe-bfa4-4dd3-ad20-00e567db53e3


PlatinMTA:
Ok so this might be what I was experiencing while testing (mouse became extremely laggy after allocating, game froze and crashed without dump), right?

botder (approved):

Fernando-A-Rocha:
Just a note of appreciation: it's really cool to finally see the CJ skin changing body and clothing models... a much awaited feature 😜

Issue #3830, opened 2024-10-28
getPedOccupiedVehicle and isPedInVehicle not working as wanted


Xenius97:
### Describe the bug

When you use setPedEnterVehicle and ped isn't sitting in vehicle, getPedOccupiedVehicle and isPedInVehicle will return element/true.

### Steps to reproduce

Lock your vehicle doors
Spawn a ped on serverside
Call setPedEnterVehicle

https://streamable.com/0vdefx

### Version

Multi Theft Auto v1.6-release-22789

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

Xenius97:
I think same for players too

TracerDS:
Please provide a minimal reproducible example to diagnose the issue

Xenius97:
> Please provide a minimal reproducible example to diagnose the issue

You can check easily with runcode, start entering to vehicle and getPedOccupiedVehicle and isPedInVehicle will return incorrect values. This is a very old bug. (only on server side)

TracerDS:
> You can check easily with runcode, start entering to vehicle and getPedOccupiedVehicle and isPedInVehicle will return incorrect values. This is a very old bug. (only on server side)

> `Server side IsPedInVehicle returns whether the ped is entering, inside or exiting a vehicle.`

The same goes for `getPedOccupiedVehicle`.
Its not that its not working as intended but more like it should be the same as the client counterpart

PR #4420, opened 2025-09-05
Fix IsPedInVehicle synchronization issue between server and client

Serius41:
Fix IsPedInVehicle synchronization issue between server and client

- Server was returning true for IsPedInVehicle when ped was only trying to enter vehicle
- Client correctly returned false until ped was actually in vehicle
- Added check for VEHICLEACTION_NONE to ensure ped is actually in vehicle
- Fixes desync where server reported ped as in vehicle during entry process

Fixes issue where setPedEnterVehicle would cause server-side IsPedInVehicle 
to return true before ped actually entered the vehicle, while client-side 
correctly returned false until ped was physically in the vehicle.

bug:

fixed:

### DIFF
--- a/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
@@ -781,7 +781,9 @@ int CLuaPedDefs::IsPedInVehicle(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        lua_pushboolean(luaVM, CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed) != NULL);
+        CVehicle* pVehicle = CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed);
+        bool bInVehicle = (pVehicle != NULL) && (pPed->GetVehicleAction() == CPed::VEHICLEACTION_NONE);
+        lua_pushboolean(luaVM, bInVehicle);
         return 1;
     }
     else
--- a/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
+++ b/Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
@@ -781,9 +781,9 @@ int CLuaPedDefs::IsPedInVehicle(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        CVehicle* pVehicle = CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed);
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -781,7 +781,9 @@ int CLuaPedDefs::IsPedInVehicle(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        lua_pushboolean(luaVM, CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed) != NULL);
+        CVehicle* pVehicle = CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed);
+        bool bInVehicle = (pVehicle != NULL) && (pPed->GetVehicleAction() == CPed::VEHICLEACTION_NONE);
```
FileEX: bInVehicle -> inVehicle

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
```diff
@@ -781,7 +781,9 @@ int CLuaPedDefs::IsPedInVehicle(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        lua_pushboolean(luaVM, CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed) != NULL);
+        CVehicle* pVehicle = CStaticFunctionDefinitions::GetPedOccupiedVehicle(pPed);
```
FileEX: pVehicle -> vehicle

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
FileEX: Remove parenthesis

FileEX (changes requested):

Server/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
FileEX:
```cpp
bool inVehicle = vehicle && pPed->GetVehicleAction() == CPed::VEHICLEACTION_NONE;
```

FileEX:
Hmmm...
image

Is this a side effect (bug) or an intended effect? It seems to me that on the server side it should work the same way as on the client side.

Related #3830

Dutchman101:
@Serius41 Do you have plans to respond to the feedback, or was this PR abandoned the day you opened it?

FileEX:
PR abandoned for over a year.

PR #4426, opened 2025-09-07
Fixes #3160 - New `onClientCommand` event

### LINKED ISSUE #3160: 'onPlayerCommand' event for client-side

MohabCodeX:
### Summary of Changes

Adds the `onClientCommand` event to MTA client-side logic, enabling Lua resources to intercept, inspect, and cancel client commands prior to local handler execution or server transmission. (Fixes #3160)

### Motivation & Semantics

- **Client vs. Server Semantics:** Server-side commands can already be intercepted via `onConsole` and `addCommandHandler` logic. Previously, client-side commands executed locally or transmitted over the network without giving client resources an early interception hook.
- **Goal:** `onClientCommand` provides client-side parity, allowing resource event handlers to invoke `cancelEvent()` to block execution locally and prevent sending cancelled commands across the network to the server.

### Testing & Verification

**Client Lua:**
<details>
<summary>Testing Script</summary>

   ```lua
   addEventHandler("onClientCommand", root, function(command, ...)
       if command == "restricted" then
           cancelEvent() 
           outputChatBox("This command is disabled!", 255, 100, 100)
       end
   end)

   addEventHandler("onClientCommand", root, function(command, ...)
       local args = {...}
       local executedByFunction = args[#args]
       table.remove(args, #args) 
       
       local blockedCommands = {"run", "exec", "execute", "lua", "script"}
       for _, blocked in ipairs(blockedCommands) do
           if command == blocked then
               cancelEvent()
               triggerServerEvent("onPlayerUsedExecutor", localPlayer, command)
               return
           end
       end
       
       if not executedByFunction then
           local currentTime = getTickCount()
           if currentTime - (lastCommandTime or 0) < 100 then
               triggerServerEvent("onPlayerUsedExecutor", localPlayer, "rapid_execution")
           end
           lastCommandTime = currentTime
       end
   end)

</details>


Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -2730,6 +2730,7 @@ void CClientGame::AddBuiltInEvents()
     // Console events
     m_Events.AddEvent("onClientConsole", "text", NULL, false);
     m_Events.AddEvent("onClientCoreCommand", "command", NULL, false);
+    m_Events.AddEvent("onClientCommand", "command, ..., executedByFunction", NULL, false);
```
FileEX: nullptr - Additionally, the correct order of arguments is ``command, executedByFunction, ...``

Client/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -150,13 +150,36 @@ bool CRegisteredCommands::CommandExists(const char* szKey, CLuaMain* pLuaMain)
...
+        
+        for (std::string arg; stream >> arg;)
+        {
+            arguments.PushString(arg.c_str());
+        }
+    }
+    
+    arguments.PushBoolean(executedByFunction);
+    
+    g_pClientGame->GetRootEntity()->CallEvent("onClientCommand", arguments, true);
```
FileEX: Why root together with children? I prefer the event source to be localPlayer, and there’s no need for callOnChildren.

Client/mods/deathmatch/ClientCommands.cpp
```diff
@@ -57,7 +58,12 @@ bool COMMAND_Executed(const char* szCommand, const char* szArguments, bool bHand
         strClumpedCommandUTF = strClumpedCommandUTF.substr(0, MAX_COMMAND_LENGTH);
         strClumpedCommand = UTF16ToMbUTF8(strClumpedCommandUTF);
 
-        g_pClientGame->GetRegisteredCommands()->ProcessCommand(szCommandBufferPointer, szArguments);
+        CommandExecutionResult commandResult = g_pClientGame->GetRegisteredCommands()->ProcessCommand(szCommandBufferPointer, szArguments, false);
```
FileEX, line 81: So now, when you type /foo, do you see the message "Unknown command or cvar"?

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
```diff
@@ -17,6 +17,7 @@ class CLuaFunctionDefinitions;
 #include "CLuaTimerManager.h"
 
 class CRegisteredCommands;
+struct CommandExecutionResult;
```
FileEX: Do you need this?

Client/mods/deathmatch/logic/CRegisteredCommands.cpp
```diff
@@ -150,13 +150,36 @@ bool CRegisteredCommands::CommandExists(const char* szKey, CLuaMain* pLuaMain)
...
 
+    CommandExecutionResult result;
+
+    CLuaArguments arguments;
+    arguments.PushString(szKey);
+    
+    if (szArguments && *szArguments)
+    {
+        const std::string_view argsView{szArguments};
+        std::istringstream stream{std::string{argsView}};
```
FileEX: By creating an std::string here you lose the benefits of string_view, since you’re still making a copy.

What do you think about using something like `std::views::split` and `std::views::filter` instead?

FileEX (changes requested):

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
MohabCodeX: yes, CLuaFunctionDefs.Commands.cpp uses `CommandExecutionResult`

Client/mods/deathmatch/ClientCommands.cpp
MohabCodeX: No, I fixed this .. they get silently sent to the server. My implementation preserves this original behavior:

1. If command is cancelled by onClientCommand → return true (don't send to server, don't show unknown command)
2. If command is not cancelled → continue normal flow (send to server, return true)

MohabCodeX: oops .. wait i need to check your suggestion again

Client/mods/deathmatch/logic/lua/CLuaFunctionDefs.h
FileEX: You already have #include "CRegisteredCommands.h" there, so this seems redundant

Client/mods/deathmatch/ClientCommands.cpp
MohabCodeX: done @FileEX 

https://github.com/user-attachments/assets/c0351147-8260-4d2e-9c64-d035b9448ac5

MohabCodeX: also if handled by script & `onClientCommand`, it will not return `unknown`
image

MohabCodeX: found new bug while using editor 😅, it also made heavy load than original behavior & network trouble when starting/restarting editor resource 
image

MohabCodeX: Fixed ✅

FileEX (dismissed):

PR #4431, opened 2025-09-07
Fix remote player fist/melee strafing sync

### LINKED ISSUE #395: Remote player fist/melee strafing is not properly synchronized
**Describe the bug**

@fedorms (Maccer) said:

> In GTA, the player is able to strafe with certain weapons. With fists and melee weapons, you are able to strafe by simply making the player attack while aiming, and then moving from left to right while he's in a post-fighting animation.
>
> MTA does not appear to synchronize the aiming action when you are strafing whilst holding these non-firearm weapons.

**To Reproduce**
> Ask a remote player to strafe by doing the following:
>
> - They must throw a punch while aiming.
> - While in the post-punch animation, they must move left and right while aiming to keep the strafe animation.

@ArranTuna noted:

> Can be easily reproduced on debug mode with setmimic 1

**Expected behavior**
Strafing state should be synced.

**Screenshots**

> FPS was controlled for both clients: capped at 25.
>
> #### With deagle
> ###### Local player strafing
> image
> 
> ###### What remote player sees
> image
> 
> #### With fists
> ###### Local player strafing
> image
> 
> ###### What remote player sees
> image
> 

 - 1.4.0

 - 1.4.0

**Additional context**
> What leads me to this conclusion is that what appears to happen after a player begins to strafe, to a remote player it looks like as if the player had punched, but then the animation blends out and the player begins to sprint and jerk weirdly because the rotation/position/velocity of the player don't match what key sync is being sent (just W/A/S/D jogging.)

From https://bugs.mtasa.com/view.php?id=8326

FileEX:
Fixed #395


ArranTuna:
Tested it with /setmimic 1 and it appears to work good.

image

ArranTuna:
https://streamable.com/8opepy

Note that I was using a keyboard and mouse that was controlling a screen-less PC, so I don't know if that jerking was caused by that rather than the PR, but aiming with a gun was fine so maybe it is a bug in the PR. Also note that later on in the video I was moving around fast with the strafing animation, but maybe I was moving without the strafing on the other client.

Dutchman101:
Be open for feedback on the results when we do the 1.7 public beta

Dutchman101 (approved):

PR #4437, opened 2025-09-09
Fix chat message word wrapping after scale change (#4425)

### LINKED ISSUE #4425: Color-coded chat messages are rendered incorrectly after changing the scale of the chat
### Describe the bug

Image

Same message, different spacing between each word.

### Steps to reproduce

1. /start runcode
2. /crun crun outputChatBox("#FF3232test #FFFFFFtest #32FF32test #3232FFtest #323232test", 0, 0, 0, true)
3. Go to settings, interface, chat and change the scaling to something bigger (for example instead of 1.0 use 1.2)
4. Profit

### Version

Client v1.6-release-23324

### Additional context

Probably just recalculating the chat after changing the scale would fix the issue.
This does not happen with messages that are not color-coded. Neither does it happen if there wasn't any color change between the same line.

### Relevant log output

```shell

```

### Security Policy

MohabCodeX:
The issue was caused by cached width calculations that weren't being invalidated when the chat scale changed.
Fixes #4425


botder:
Apparently there is only a single place where `m_vecScale` can change?

MohabCodeX:
> Apparently there is only a single place where `m_vecScale` can change?

Yes.

Dutchman101 (approved):

Issue #4157, opened 2025-04-12
a function to abort downloads initiated with downloadFile


ffsPLASMA:
### Is your feature request related to a problem? Please describe.

Currently, there is no way to stop the download of a file initiated via downloadFile from lua.
This can be an issue if downloading large files like IMG containers.
Unless you disconnect from server...

### Describe the solution you'd like

Add a new function like abortDownload(handler) to be used if download was started manually in lua via downloadFile.

local downloadHandler = downloadFile("foo.bin")
-- if it takes too long...
abortDownload(downloadHandler)

### Describe alternatives you've considered

Disconnect from server or stop resource.

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

PR #4438, opened 2025-09-09
New `abortDownload` function for `downloadFile` (#4157)

MohabCodeX:
>[!WARNING]
**~This feature has not been fully tested. Please test thoroughly and report any issues or unexpected behaviors. Your feedback is highly appreciated~.**

> Sorry if I may have disappointed you, but this will take time and will need a different solution 🫠

### New Functionality

* **`abortDownload(handlerId)`**
  Cancels an ongoing download using a unique handler ID.

The function returns `false` in these cases:
- Invalid handler ID (download doesn't exist)
- Download already completed
- Download already cancelled
- HTTP cancellation failed
- It returns `true` when the download was successfully cancelled.

###  Usage Example

```lua
local download, handlerId = downloadFile("file.txt")
-- ... later ...
local success = abortDownload(handlerId)
```

### 🛠️ Changes

* Added download tracking using unique handler IDs.
* Integrated proper cancellation logic with the HTTP manager.
* Implemented input validation for both `downloadFile` and `abortDownload`.
* memory safety and error handling throughout.

###  Breaking Changes

* **`downloadFile()`** now returns **two values** (`download object`, `handler ID`) instead of a single boolean.

### 🔗 Resolves

* Issue **#4157**

### DIFF
--- a/Client/mods/deathmatch/logic/CSingularFileDownload.cpp
+++ b/Client/mods/deathmatch/logic/CSingularFileDownload.cpp
@@ -12,7 +12,7 @@
 #include <StdInc.h>
 
 CSingularFileDownload::CSingularFileDownload(CResource* pResource, const char* szName, const char* szNameShort, SString strHTTPURL, CResource* pRequestResource,
-                                             CChecksum serverChecksum)
+                                             CChecksum serverChecksum, std::uint32_t handlerId)
 {
     // Store the name
     m_strName = szName;
@@ -28,6 +28,10 @@ CSingularFileDownload::CSingularFileDownload(CResource* pResource, const char* s
     m_ServerChecksum = serverChecksum;
 
     m_bBeingDeleted = false;
+    m_bCancelled = false;
+    m_handlerId = handlerId;
+    m_pHTTPManager = nullptr;
+    m_downloadMode = EDownloadModeType::NONE;
 
     GenerateClientChecksum();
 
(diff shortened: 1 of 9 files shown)


Fernando-A-Rocha:
I think new feature PRs are **on-hold** for now. 1.7 release is waiting for fixes first

Either way, this is cool

ffsPLASMA:
Thank you for adding this feature, much needed.
If it isnt too  much work, could you also add something like:

local numBytesDownloaded, numTotalBytesToDownload = getDownloadProgress(downloadHandler);

Very good addition for UIs to display the download progress.

MohabCodeX:
The main problem is that practical usage is limited due to a fundamental issue with `downloadFile`.
Currently, downloadFile is blocking (synchronous), it halts script execution until the file is fully downloaded.
This prevents proper asynchronous handling and makes progress tracking ineffective during the actual download.

ffsPLASMA:
How feasible would it be to create something like asyncDownloadFile function?

Fernando-A-Rocha:
> How feasible would it be to create something like asyncDownloadFile function?

Why not just make downloadFile not blocking? I think that wouldn't break scripts. It returns true if file to download was queued successfully - is this the slow operation?

ffsPLASMA:
> > How feasible would it be to create something like asyncDownloadFile function?
> 
> Why not just make downloadFile not blocking? I think that wouldn't break scripts. It returns true if file to download was queued successfully - is this the slow operation?

Yes I see no problem making downloadFile async in general. No one sane would put lua logic after downloadFile call.

Issue #1034, opened 2019-07-31
ShowCursor does not disable controls


Cyber52:
**Describe the bug**
When we set cursor to be showd without disabling controls and after it we do the same but with disabling controls, controls wont be disabled. To get around it you have to hide cursor and then show cursor. I think it should not happen when we set disableControls to true.

**To reproduce**
Steps to reproduce the behaviour:
```lua
showCursor(true, false)
showCursor(true, true)
```

**Expected behaviour**
Like the wiki says: "toggleControls: A boolean value determining whether to disable controls whilst the cursor is showing. true implies controls are disabled, false implies controls remain enabled."

When we do use showCursor twice in 2 resources and the first one disables controls and the second doesnt the `showCursor(false)` command should know if the resource disabled controls or not so when we hide cursor using only first resource the cursor still will be visible but controls will not be disabled.

**Version**
1.5.6, windows 8.1

**Additional context**
I was bringing it up on Discord server, and this Is what staff sad:
![image](https://user-images.githubusercontent.com/29693226/62227122-ab8e5980-b3bb-11e9-8854-e1f3957035bd.png)
This is the link: https://github.com/multitheftauto/mtasa-blue/blob/4244ec52a5155d624cd5ec8dca1bdca067309358/Client/mods/deathmatch/logic/CResource.cpp#L379

FileEX:
Fixed in https://github.com/multitheftauto/mtasa-blue/commit/117f804cb2aa844608f939ab749bf280929a003a

PR #4448, opened 2025-09-10
Fix `showCursor` not updating controls state when cursor already visible (#1034)

MohabCodeX:
> * Fixes **#1034**

Now `bToggleControls` parameter is always applied regardless of cursor visibility state.

Tested.


Client/mods/deathmatch/logic/CResource.cpp
```diff
@@ -410,10 +410,19 @@ void CResource::ShowCursor(bool bShow, bool bToggleControls)
...
+    if (m_iShowingCursor > 0)
+    {
+        g_pCore->ForceCursorVisible(true, bToggleControls);
+        g_pClientGame->SetCursorEventsEnabled(true);
+    }
+    else
+    {
+        // Hide cursor if no resources want it visible
+        g_pCore->ForceCursorVisible(false, bToggleControls);
+        g_pClientGame->SetCursorEventsEnabled(false);
```
FileEX: How does this solve the problem, since the logic of this code is the same as the original one, just needlessly longer?

```cpp
g_pCore->ForceCursorVisible(m_iShowingCursor > 0, bToggleControls);
g_pClientGame->SetCursorEventsEnabled(m_iShowingCursor > 0);
```
This is exactly the same as
```cpp
    if (m_iShowingCursor > 0)
    {
        g_pCore->ForceCursorVisible(true, bToggleControls);
        g_pClientGame->SetCursorEventsEnabled(true);
    }
    else
    {
        // Hide cursor if no resources want it visible
        g_pCore->ForceCursorVisible(false, bToggleControls);
        g_pClientGame->SetCursorEventsEnabled(false);
   }
```

FileEX: Okay, I see that the condition just ends earlier. In that case, leave this code as it was in the original, now it’s unnecessarily lengthened.

FileEX (approved):

Issue #1519, opened 2020-06-13
Using removePedFromVehicle in onVehicleStartEnter causes network trouble


zo3n:
**Describe the bug**
If you use removePedFromVehicle(enteringPlayer, source) in onVehicleStartEnter, it will cause a network trouble after a few seconds.

**To reproduce**
I've provided a resource. It also comes with instructions when you run it.
1. Use /vehicle
2. Press F
3. Wait for a few seconds.

Please paste reproduction resources (zip files) directly into this section.
-->[vbug.zip](https://github.com/multitheftauto/mtasa-blue/files/4774534/vbug.zip)

**Expected behaviour**
To not happen

**Version**
Multi Theft Auto v1.5.7-release-20582

**Additional context**
I know it's not normal to use this in correctly working scripts, but if your mind goes woof like mine did, and you copy paste lines of code, if you use this part of code, it'll cause a network trouble.

Unde-R:
Yeah I just tried It, you get network trouble

Unde-R:
I guess here we can just document that It wont work when entering the vehicle.
Also Instead of `removePedFromVehicle` you can just use `cancelEvent` here

FileEX:
Fiixed in https://github.com/multitheftauto/mtasa-blue/commit/3ec9b5c5a4e6f2a3783a3ae2c1c9eb9802678f31

PR #4450, opened 2025-09-10
Fix network desync when `removePedFromVehicle` called during vehicle enter (#1519)

MohabCodeX:
* Fixes **#1519**

Calling `removePedFromVehicle()` while a player is entering a vehicle left the client-server handshake incomplete, causing network desync, so i added handling to detect mid-enter states and send proper abort packets to clients to prevent the desync while maintaining backward compatibility.

Tested using the same script of #1519.


FileEX (commented):
Will network trouble also occur if, while exiting a vehicle, we use warpPedIntoVehicle on the same vehicle?

MohabCodeX:
* discussed in PR **#4458**

FileEX (approved):

PR #4452, opened 2025-09-11
Fix missing mtaserver.conf setting added from template not receiving attributes

### LINKED ISSUE #4444: <rule> is being added to mtaserver.conf multiple times
### Describe the bug

Every time I start the server these lines are being added despite being already added:

    <rule></rule>
    
    
    <rule></rule>
    
    
    <rule></rule>
    
    
    <rule></rule>
    
    
    <rule></rule>
    
    
    <rule></rule>

Also they are not being added properly as this is  how they look in a fresh install:

    
    <rule name="description" value="" />
    
    
    <rule name="languages" value="" />
    
    
    <rule name="tags" value="" />
    
    
    <rule name="website_url" value="" />
    
    
    <rule name="social_url_1" value="" />
    
    
    <rule name="social_url_2" value="" />
    
    
    <rule name="social_url_3" value="" />

### Steps to reproduce

Upgrade a 1.6 server to 1.7

### Version

1.7-untested-25543 Windows 64 bit.

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

Fernando-A-Rocha:
Fixes #4444

Previously, when a missing config node was added from the template, only its tag name, content, and comment were copied, leaving out all attributes.

This fix ensures that all attributes from the template node are copied into the new node, so added settings fully match the template. Currently only used in the new <rule> nodes.

### DIFF
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -916,7 +916,7 @@ bool CMainConfig::AddMissingSettings()
                 CXMLAttribute* templateAttribute = *it3;
                 const SString& attrName = templateAttribute->GetName();
 
-                // Don't check value attribute which is intended to be different
+                // Don't check value attribute which is intended to be customized by the server
                 if (attrName == "value")
                     continue;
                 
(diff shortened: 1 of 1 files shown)


ArranTuna:
Tested and confirmed that this PR fixes the bug.

FileEX (approved):

PR #4453, opened 2025-09-11
Fix position desync with dead players (#928)

MohabCodeX:
Dead players stopped sending position updates to server, causing `getElementPosition()` to return frozen death location instead of current position (e.g., after falling). 
Fixed by allowing position sync for dead players on both client and server sides.

**Results** : 
```
Your position: X=-710.00, Y=957.00, Z=12.40
Your position: X=-710.63, Y=957.88, Z=42.75 -> In air
Your position: X=-710.63, Y=957.88, Z=12.40 -> Dead player position after falling
```

* Fixes **#928**


Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -36,8 +36,8 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
             return false;
 
         // Only read this packet if it matches the current time context that
-        // player is in.
-        if (!pSourcePlayer->CanUpdateSync(ucTimeContext))
+        // player is in. Allow position updates for dead players
+        if (!pSourcePlayer->CanUpdateSync(ucTimeContext) && !pSourcePlayer->IsDead())
```
FileEX: I don't think dead players can be synchronized if the time context doesn't match

Dutchman101: @MohabCodeX Please respond to the feedback

FileEX (approved):

Issue #2798, opened 2022-10-23
Jump control triggers zoom in


lopezloo:
### Describe the bug

Jump control triggers weapon zoom in action.

### Steps to reproduce

```lua
run giveWeapon(me, 34, 69, true) setControlState(me, "aim_weapon", true) setControlState(me, "jump", true)
```

### Additional context

That doesn't happen in single player.

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

bum8hj:
✅ **I was able to reproduce it** _(v1.5.9-release-21342)_

Setting the `"jump"` control state to `false` stops the zooming-in, and setting it back to `true` starts zooming-in again.

Another runcode line for the lazy :skull:
`run p = getRandomPlayer() giveWeapon(p, 34, 69, true) setControlState(p, "aim_weapon", true) setControlState(p, "jump", true)`

MohabCodeX:
already fixed. #4454

PR #4454, opened 2025-09-11
Fix jump control triggering weapon zoom while aiming (#2798)

MohabCodeX:
* Fixes **#2798**

#### Before:
https://github.com/user-attachments/assets/6b57d1b1-3340-404e-a09f-ad7e7a1f388e

---

#### After:
https://github.com/user-attachments/assets/6635922f-9920-4cc8-9e10-5338216dfe16


Client/core/CKeyBinds.cpp
```diff
@@ -1950,7 +1950,7 @@ void CKeyBinds::DoPostFramePulse()
 
             // * Enter Exit
             // * Change View
-            cs.ButtonSquare = (!bEnteringVehicle && g_bcControls[11].bState) ? 255 : 0;            // Jump
+            cs.ButtonSquare = (!bEnteringVehicle && !bAimingWeapon && g_bcControls[11].bState) ? 255 : 0;            // Jump
```
FileEX, line 1953: Are you sure this doesn’t remove the ability to jump while holding other weapons without zoom capability, e.g., can you still jump while holding a Desert Eagle? It seems to me that this issue only affects sniper and camera, since holding shift actually lets you zoom with them. But generally using bAiming seems like it will block jump for all weapons while aiming. Is that the intended effect?

MohabCodeX: Thanks for the observation. I also tested it, and I can still jump while holding other weapons normally.
`bAimingWeapon` only becomes true when actively aiming with specific weapons that create a `TASK_SIMPLE_USE_GUN` task (snipers/camera). When aiming with regular weapons like Desert Eagle, no such task is created, so `bAimingWeapon` remains false and jumping works normally. so fix is specifically targets the zoom-capable weapons

Dutchman101:
To be tested in 1.7 public beta; any issues with weapon aiming will surely surface if any.

Dutchman101 (approved):

PR #4456, opened 2025-09-11
Sort the screen resolutions in the in-game video settings

MohabCodeX:
This PR sorts screen resolutions in the video settings dialog by size (largest first, descending).
It collects all available resolutions, sorts them by ~total pixel count (width × height)~ width first approach, then height, in descending order

image

More time and effort are needed for external dialogue.

~It's ordered by total pixel, not **width first**, so i think i will change this to match other game patterns.~

### DIFF
--- a/Client/core/CSettings.cpp
+++ b/Client/core/CSettings.cpp
@@ -1771,14 +1771,12 @@ void CSettings::PopulateResolutionComboBox()
     if (resolutions.empty())
         return;
 
-    // Sort resolutions by total pixels (descending), then by width, then by depth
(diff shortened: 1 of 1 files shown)


Client/core/CSettings.cpp
```diff
@@ -1683,6 +1685,25 @@ void CSettings::UpdateVideoTab()
...
+    
+    bool operator==(const ResolutionData& other) const
+    {
+        return width == other.width && height == other.height && depth == other.depth;
+    }
+    
+    int getPixelCount() const
+    {
+        return width * height;
+    }
```
FileEX: Unused

Issue #4466, opened 2025-09-15
Limit "but event is not added serverside" to one error per second.


ArranTuna:
### Is your feature request related to a problem? Please describe.

ERROR: Client (name) triggered serverside event eventName, but event is not added serverside

This debug error is useful to find bugs in scripts, but there is no reason for it to debug hundreds of times a second. Limiting it to one output per second would prevent servers from being impacted by... buggy scripts.

### Describe the solution you'd like

I was thinking at first a setting you have to enable to ignore these errors, but instead just protect all servers by default with a limit of 1 per second.

### Describe alternatives you've considered

There is no better alternative than this.

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

ArranTuna:
I have made a PR trying my best to comply with all the coding guidelines - https://github.com/multitheftauto/mtasa-blue/pull/4467

PR #4467, opened 2025-09-15
Limit non-added event logging (#4466)

ArranTuna:
Only throw an error once per second to prevent flooding from clients triggering non-added server side events.


Fernando-A-Rocha:
Servers should auto-punish a client when they trigger an invalid event (with https://wiki.multitheftauto.com/wiki/OnPlayerTriggerInvalidEvent) so they don't spam at all.
But considering that some servers might not still be doing this, it's probably a good idea to rate-limit the errors thrown on the sever 👍🏼

ArranTuna:
I use that event, but sometimes auto-punish doesn't work and I won't say more on that as it shouldn't be discussed publicly. But this limit will help when auto-punish doesn't work, but also this is good for scripters, say if you've got 100 players on a server and your server side script failed to load, or you forgot to add an event, you could be getting spammed with so many of these errors that you didn't get a chance to see the initial load failure and it's quite a stressful experience. Also onPlayerTriggerInvalidEvent is bad in that case as you could end up auto-punishing innocent players due to a scripting mistake.

Fernando-A-Rocha:
Ok you made valid points. Hundreds of players could trigger the error at the same time.

👍🏼

Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2673,15 +2674,22 @@ void CGame::Packet_LuaEvent(CLuaEventPacket& Packet)
             }
 
         }
-            else
+        else
+        {
+            // Limit this error to prevent debug / log files being flooded.
+            static std::uint64_t s_lastClientEventErrorLog = 0;
+            auto now = GetTickCount64_();
```
FileEX, line 2681: auto -> std::uint64_t

FileEX (changes requested):

FileEX (approved):

PR #4475, opened 2025-09-17
Fix map editor bug

omar-o22:
There was a bug when you join map editor and rejoin again game will crash
In this PR this bug got fixed

image

### DIFF
--- a/Client/core/CModManager.cpp
+++ b/Client/core/CModManager.cpp
@@ -122,6 +122,8 @@ void CModManager::Unload()
 
 bool CModManager::Start()
 {
+    if (m_library != nullptr || m_client != nullptr)
+        Stop();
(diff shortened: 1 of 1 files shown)


ArranTuna:
I've tried launch map editor twice in a row but no crash, what are the steps to reproduce?

omar-o22:
> I've tried launch map editor twice in a row but no crash, what are the steps to reproduce?

1. Join map editor
2. When u join, join it again

[edited]:
What is your version? I have tried on 1.6 it works correctly but I currently can't test on 1.7, The bug appears in my own build but i have tried also on latest branch on master where the bug also appears

omar-o22:
I tested it in 1.7 it works correctly
maybe my own build has something wrong, idk..
anyway i'll close this PR since its useless

PR #4477, opened 2025-09-18
onElementDataChange can be cancelled.

ArranTuna:
onElementDataChange can now be cancelled when a client, setElementData, or removeElementData changes an element data. This is useful because right now you have to do setElementData to cancel a change, which broadcasts to all players and also triggers the event again. Being able to just cancel the event is the cleanest way to deal with unauthorized data changes.

I tested it as many ways as I could think of to try and break it but everything works as intended.

### DIFF
--- a/Client/core/DXHook/CProxyDirect3D9.cpp
+++ b/Client/core/DXHook/CProxyDirect3D9.cpp
@@ -963,6 +963,7 @@ bool CreateDeviceSecondCallCheck(HRESULT& hOutResult, IDirect3D9* pDirect3D, UIN
     if (pPresentationParameters->BackBufferWidth == 0)
     {
         WriteDebugEvent(SString(" Passing through call #%d to CreateDevice because size is invalid", uiCreateCount));
+        hOutResult = pDirect3D->CreateDevice(Adapter, DeviceType, hFocusWindow, BehaviorFlags, pPresentationParameters, ppReturnedDeviceInterface);
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CGame.cpp
```diff
@@ -2738,25 +2738,45 @@ void CGame::Packet_CustomData(CCustomDataPacket& Packet)
...
+                        m_pPlayerManager->BroadcastOnlySubscribed(CElementRPCPacket(pElement, SET_ELEMENT_DATA, *BitStream.pBitStream), pElement, szName,
+                                                                  pSourcePlayer);
+
+                    CPerfStatEventPacketUsage::GetSingleton()->UpdateElementDataUsageRelayed(szName, m_pPlayerManager->Count(),
+                                                                                             BitStream.pBitStream->GetNumberOfBytesUsed());
+                }
+            }
+            else
+            {
+                // Event was cancelled; sync the authoritative value back to the source player
```
tederis: Why do we need to send it? In case `::SetCustomData` returned **false** there weren't any data transfer between server and client.

ArranTuna: This is the bit where client is telling the server they've changed an element data. I decided to add this so that if a client side element data change is rejected by the server, send the servers version of the element data back to the client so the client is synced with the server.

tederis: Ah, I see. It's a bit of overhead, but I don't see other way tho.

tederis (approved):

PR #4409, opened 2025-09-03
Fix ACL request detection when using refresh (#1824)

MohabCodeX:
The issue occurred because the system skipped creating the necessary ACL structures when permissions already existed elsewhere. Additionally, it now ensures consistent cleanup of _temporary_ ACL entries and prevents any permission leakage.
Fixes #1824

Issue #1824, opened 2020-11-11
Refresh does not work on aclrequest list


CrosRoad95:
**Describe the bug**
/refresh command does not refresh aclrequest list in meta.xml

**To reproduce**
1. create resource with meta.xml content eg.
```xml
<aclrequest>
<right name="function.startResource" access="true" />
<right name="function.stopResource" access="true" />
<right name="function.setElementPosition" access="true" />
</aclrequest>
```
2. type /refresh and `aclrequest list X` where X is your resource name
3. modify aclrequest list, eg remove one right
4. type /refresh and `aclrequest list X` again - nothing changed
5. rename resource X to Y
6. type /refresh and `aclrequest list Y` again - you will see you changes

**Expected behaviour**

**Screenshots**
/

**Version**
1.5.8 release 20698

**Additional context**
/

PerikiyoXD:
```ini
refreshall
[07:59:07] Resource 'resources1824' changed, reloading
[07:59:07] Resource 'resources1824' requests some acl rights. Use the command 'aclrequest list resources1824'
[07:59:07] New resource 'resources1824' loaded
[07:59:07] Resources: 216 loaded, 0 failed
aclrequest list resources1824
[07:59:13] aclrequest: resources1824 [pending] for function.startResource
[07:59:13] aclrequest: resources1824 [pending] for function.stopResource
refreshall resources1824
[07:59:26] Resource 'resources1824' changed, reloading
[07:59:26] Resource 'resources1824' requests some acl rights. Use the command 'aclrequest list resources1824'
[07:59:26] New resource 'resources1824' loaded
[07:59:26] Resources: 216 loaded, 0 failed
aclrequest list resources1824
[07:59:36] aclrequest: resources1824 [pending] for function.startResource
[07:59:36] aclrequest: resources1824 [pending] for function.stopResource
[07:59:36] aclrequest: resources1824 [pending] for function.setElementPosition
```

```cpp
// Load the complete list of resources and create their objects
// DOES NOT reload already loaded resources, we need a special function for lua for that (e.g. reloadResource)
// Talidan: yes it did, noob.  Not as of r897 - use bRefreshAll to do already loaded resources.
bool CResourceManager::Refresh(bool bRefreshAll, const SString strJustThisResource, bool bShowTiming)
```

`Not as of r897 - use bRefreshAll to do already loaded resources.`

So "refreshall" triggers resource loading, so this is nonsensical?

[...]

PR #4483, opened 2025-09-21
Simplify ACL request refresh detection (improves #4409)

MohabCodeX:
This PR replaces the complex ACL fingerprinting from PR #4409 with a much simpler and more efficient approach.

Thanks for @ArranTuna for mentioning a proper solution.


MohabCodeX:
related #4484

Server/mods/deathmatch/logic/CResource.h
```diff
@@ -349,8 +347,7 @@ class CResource : public EHS
     void CommitAclRequest(const SAclRequest& request);
     bool FindAclRequest(SAclRequest& result);
 
-    std::string CalculateACLRequestFingerprint();
-    bool        HasACLRequestsChanged();
+protected:
```
FileEX: ?

PerikiyoXD:
Is this really a fix? Or this became an oversimplification refactor over a bug?

ArranTuna:
Yes it fixes #4484 I have tested it.

PerikiyoXD:
I've written my concerns about this and the prior PR:
https://github.com/multitheftauto/mtasa-blue/issues/1824

ArranTuna:
> I've written my concerns about this and the prior PR: #1824

This PR basically does revert that previous PR and replaces it with a simpler fix for the original bug. How it works now is that if a resource has ACL requests, it doesn't have to be started to output the requests to the server console, if you changed them and did refresh, the changes wouldn't be registered, but now when doing refresh, the changes are registered to these non started resources. Currently the only way to get the ACL requests to get registered is to rename the resource and execute refresh. Now you don't have to rename a resource.

PerikiyoXD:
refreshall triggers the acl reload check

ArranTuna:
That's true, but you may not want to restart all changed resources, it can be very slow and freeze the server for 1 minute, or you just might not want some changes to be applied.

PerikiyoXD:
We can agree that you brought into the table that the culprit of the preference, is a performance issue.

`CMD5Hasher` is the slowness when `refreshall` is triggered.
imagen

This is to be addressed sometime using CryptoPP implementation instead of handrolled and see if the performance is improved (Which I assume will improve at least 40%)

Server/mods/deathmatch/logic/CResource.h
MohabCodeX: For accessing members, but i made a better solution now 👍🏻

Fernando-A-Rocha:
this fix is missing for 1.7

FileEX (approved):

Issue #4490, opened 2025-09-23
Using ``processLineAgainstMesh`` with building elements can result in Crash 0x00084B86


PlatinMTA:
### Describe the bug

For some reason, using ``processLineAgainstMesh`` with an array of buildings will crash the game, but only if the amount of buildings tested is "high". I managed to make it crash with under 25 buildings, but on the test I uploaded I am using 100 buildings just to be sure it works.

This doesn't happen with objects.

[test_crash.zip](https://github.com/user-attachments/files/22481482/test_crash.zip)

### Steps to reproduce

1. Download the test resource and run it, you should crash.

Otherwise, run this code

```lua
local x, y, z = getElementPosition(localPlayer)

local buildings = {}
for i=1,25 do
	buildings[i] = createBuilding(1337, x, y+2*i, z)
end

setTimer(function()
	local px, py, pz = getElementPosition(localPlayer)
	for i=1,#buildings do
		local building = buildings[i]
		local hit, u, v, texture, frame, wx, wy, wz = processLineAgainstMesh(building, px, py, pz, x, y+2*i, z)
		iprint(i, hit, u, v, texture, frame, wx, wy, wz)
	end
end, 1000, 1)
```

### Version

Client: Multi Theft Auto v1.7-untested-25596

### Additional context

```
** -- Unhandled exception -- **

Version = 1.7-untested-25596.0.000
Time = Mon Sep 22 21:19:27 2025
Module = D:\MTA San Andreas 1.7\mta\game_sa.dll
Code = 0xC0000005
Offset = 0x00084B86

EAX=0177F48C  EBX=0177F458  ECX=00000000  EDX=0177F474  ESI=0177F48C
EDI=250B1EB0  EBP=0177F450  ESP=0177F2C0  EIP=53564B86  FLG=00010282
CS=0023   DS=002B  SS=002B  ES=002B   FS=0053  GS=002B

** -- End of unhandled exception -- **
```

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

tederis:
Thanks for the report. The crash is eliminated in #4491.

PR #4491, opened 2025-09-23
Crashfix

tederis:
Fixes a crash described in #4490.

### DIFF
--- a/Client/game_sa/CWorldSA.cpp
+++ b/Client/game_sa/CWorldSA.cpp
@@ -10,6 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <numbers>
 #include <multiplayer/CMultiplayer.h>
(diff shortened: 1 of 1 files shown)


FileEX:
Doesn’t the same problem occur in ``TestSphereAgainstWorld``? Maybe it’s worth fixing it right away in this PR.
https://github.com/multitheftauto/mtasa-blue/blob/e165f6dbe6621475e0ab1956ede8d9629ef71256/Client/game_sa/CWorldSA.cpp#L544

FileEX (dismissed):

tederis:
> Doesn’t the same problem occur in `TestSphereAgainstWorld`? Maybe it’s worth fixing it right away in this PR.

That's interesting that testSphereAgainstWorld doesn't cause the crash. But you're right, it should also be fixed.

FileEX (approved):

PR #4492, opened 2025-09-23
The second iteration of element data optimization

tederis:
This is the second iteration of the element data optimization challenge. The first one see in there: #4169. This is a part of the reimplementation of #3287. This PR is aimed to eliminating unnessesary string copying. It also brings a faster element data lookup inside `std` containers.


Dutchman101 (approved):

Dutchman101:
Nice. We can still test this in 1.7

PR #4507, opened 2025-10-03
add `isLoadingCircleActive` function

### LINKED ISSUE #4505: Add loading circle function

It is not related to any problem.
But it would be good since in my case it would be used to show the Login after finishing downloading models from the server.

A function that detects when the download of the circle below the download bar is finished, something like isTransferBoxActive() but with the circle below the download bar.

_No response_

### Additional context

_No response_

### Security Policy

omar-o22:
Fixes #4505
New function
```lua
bool isLoadingCircleActive( )
```

### DIFF
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -2147,11 +2147,11 @@ void CGraphics::RestoreGTARenderStates()
 ////////////////////////////////////////////////////////////////
 void CGraphics::DidRenderScene()
 {
-    if (m_bProgressVisible)
+    if (IsLoadingCircleActive())
     {
         // Min display time before turning off spinner
         if (m_FirstDrawnProgressTimer.Get() > DUMMY_PROGRESS_MIN_DISPLAY_TIME)
-            m_bProgressVisible = false;
+            SetLoadingCircleActive(false);
         else
             DrawProgressMessage(false);
     }
@@ -2173,13 +2173,13 @@ void CGraphics::DidRenderScene()
 ////////////////////////////////////////////////////////////////
 void CGraphics::SetProgressMessage(const SString& strMessage)
 {
-    if (!m_bProgressVisible)
+    if (!IsLoadingCircleActive())
     {
         if (m_LastRenderedSceneTimer.Get() < DUMMY_PROGRESS_INITIAL_DELAY)
             return;
 
         // Enable progress drawing
-        m_bProgressVisible = true;
+        SetLoadingCircleActive(true);
         m_FirstDrawnProgressTimer.Reset();
     }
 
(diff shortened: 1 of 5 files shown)


Fernando-A-Rocha:
Doesn't the client freeze when loading models most of the time?😅

omar-o22:
> Doesn't the client freeze when loading models most of the time?😅

Yup, but the function I think can be used on the render or something..
After finishing I realized the function is kind of useless :D, but I opened this PR to hear other opinions.

Fernando-A-Rocha:
Yeah I had never heard of anyone needing to detect if the client has the loading circle haha which is a weird feature anyway

omar-o22:
Yeah, anyway it was just an experiment haha, and I learned a few things from it.

luchodev10:
It is not useless, since it would serve to show a system after the models are loaded. During the loading of models it is true that the screen freezes, but not completely. During the loading process, the login process is blocked, for example. Therefore, the function would serve to prevent this from happening. It could also be used to not load any resources until the models are loaded. Visually, it would look much better. Or if not, MTA should make it not load any resources until the models are loaded, since it affects the visual aspect during the loading process.

omar-o22:
The function is returns true maybe for 450ms only then always false because while loading models the game is freezes.
```lua
local loadingStart = 0

    if isLoadingCircleActive() then
        if loadingStart == 0 then
            loadingStart = getTickCount()
            outputChatBox("Loading started...")
        end
    elseif loadingStart ~= 0 then
        local duration = getTickCount() - loadingStart
        outputChatBox("Loading took "..duration.." ms.")
        loadingStart = 0
    end
end)
```

Fernando-A-Rocha:
> It is not useless, since it would serve to show a system after the models are loaded. During the loading of models it is true that the screen freezes, but not completely. During the loading process, the login process is blocked, for example. Therefore, the function would serve to prevent this from happening. It could also be used to not load any resources until the models are loaded. Visually, it would look much better. Or if not, MTA should make it not load any resources until the models are loaded, since it affects the visual aspect during the loading process.

You should make the model loading resource start first

luchodev10:
> The function is returns true maybe for 450ms only then always false because while loading models the game is freezes.
> 
> ```lua
> local loadingStart = 0
> 
> addEventHandler("onClientRender", root, function()
>     if isLoadingCircleActive() then
>         if loadingStart == 0 then
>             loadingStart = getTickCount()
>             outputChatBox("Loading started...")
>         end
>     elseif loadingStart ~= 0 then
>         local duration = getTickCount() - loadingStart
>         outputChatBox("Loading took "..duration.." ms.")
>         loadingStart = 0
>     end
> end)
> ```

I understand

luchodev10:
> > It is not useless, since it would serve to show a system after the models are loaded. During the loading of models it is true that the screen freezes, but not completely. During the loading process, the login process is blocked, for example. Therefore, the function would serve to prevent this from happening. It could also be used to not load any resources until the models are loaded. Visually, it would look much better. Or if not, MTA should make it not load any resources until the models are loaded, since it affects the visual aspect during the loading process.
> 
> You should make the model loading resource start first

It wouldn't work for me since I use the Login as a loadscreen. When the resources download is finished, the login is shown. For a new player, if the models have more priority than the login, the loadscreen would not work properly.

PR #4508, opened 2025-10-04
Crashfix (uninitialized ptr)

tederis:
`m_pCurrentLang` pointer is uninitialized currently which is UB and causes random crashes during a startup process. This PR fixes it.


Fernando-A-Rocha:
Did someone experience the random crashes before?

tederis:
> Did someone experience the random crashes before?

I personally have experienced it today for the first time. And it's quite strange that it had never happened before. Though... I had encountered weird crashes on game start some time ago, but didn't explore it further.

PR #4527, opened 2025-11-09
Resolve CLocalization / CGUI memory leaks

Lpsd:
Raised in Discord https://canary.discord.com/channels/801330706252038164/801411628600000522/1436840277175308299

CEGUI wasn't being properly cleaned up when GUI recreated (on locale and skin change).
We also have no reason to store 'old locales' in the map in CLocalization, so these are cleared up automatically

### DIFF
--- a/Client/core/CLocalization.cpp
+++ b/Client/core/CLocalization.cpp
@@ -41,7 +41,7 @@ CLocalization::CLocalization(const SString& strLocale, const SString& strLocaleP
 
 CLocalization::~CLocalization()
 {
-    // m_LanguageMap now uses unique_ptr so cleanup is automatic
+    m_pCurrentLang = nullptr;
 }
 
 //
--- a/Client/core/CLocalization.cpp
+++ b/Client/core/CLocalization.cpp
@@ -41,10 +41,7 @@ CLocalization::CLocalization(const SString& strLocale, const SString& strLocaleP
 
 CLocalization::~CLocalization()
 {
-    for (auto iter : m_LanguageMap)
-    {
-        delete iter.second;
-    }
+    // m_LanguageMap now uses unique_ptr so cleanup is automatic
 }
 
 //
@@ -99,29 +96,32 @@ void CLocalization::SetCurrentLanguage(SString strLocale)
 CLanguage* CLocalization::GetLanguage(SString strLocale)
 {
     strLocale = ValidateLocale(strLocale);
-    CLanguage* pLanguage = MapFindRef(m_LanguageMap, strLocale);
-    if (!pLanguage)
+    auto iter = m_LanguageMap.find(strLocale);
+    if (iter != m_LanguageMap.end())
     {
-        Language Lang = Language::from_name(strLocale);
-        Lang = Lang ? Lang : Language::from_name("en_US");
-
-        try
-        {
-            pLanguage = new CLanguage(m_DictManager.get_dictionary(Lang, MTA_LOCALE_TEXTDOMAIN), Lang.str(), Lang.get_name());
-            MapSet(m_LanguageMap, strLocale, pLanguage);
-        }
-        catch (const std::exception& ex)
-        {
-            WriteDebugEvent(SString("Localization failed to load dictionary for '%s': %s", strLocale.c_str(), ex.what()));
-            return (strLocale != "en_US") ? GetLanguage("en_US") : nullptr;
-        }
-        catch (...)
-        {
-            WriteDebugEvent(SString("Localization failed to load dictionary for '%s': unknown error", strLocale.c_str()));
-            return (strLocale != "en_US") ? GetLanguage("en_US") : nullptr;
-        }
+        return iter->second.get();
     }
-    return pLanguage;
+
+    Language Lang = Language::from_name(strLocale);
+    Lang = Lang ? Lang : Language::from_name("en_US");
+
+    try
+    {
+        std::unique_ptr<CLanguage> pLanguage = std::make_unique<CLanguage>(m_DictManager.get_dictionary(Lang, MTA_LOCALE_TEXTDOMAIN), Lang.str(), Lang.get_name());
+        CLanguage* pLanguagePtr = pLanguage.get();
+        m_LanguageMap.emplace(strLocale, std::move(pLanguage));
+        return pLanguagePtr;
+    }
+    catch (const std::exception& ex)
+    {
+        WriteDebugEvent(SString("Localization failed to load dictionary for '%s': %s", strLocale.c_str(), ex.what()));
+        return (strLocale != "en_US") ? GetLanguage("en_US") : nullptr;
+    }
+    catch (...)
+    {
+        WriteDebugEvent(SString("Localization failed to load dictionary for '%s': unknown error", strLocale.c_str()));
+        return (strLocale != "en_US") ? GetLanguage("en_US") : nullptr;
+    }
 }
 
 //
--- a/Client/gui/CGUI_Impl.cpp
+++ b/Client/gui/CGUI_Impl.cpp
@@ -330,28 +330,31 @@ bool CGUI_Impl::GetGUIInputEnabled()
             break;
         case INPUTMODE_NO_BINDS_ON_EDIT:
         {
-            CEGUI::Window* pActiveWindow = m_pTop->getActiveChild();
-            if (!pActiveWindow || pActiveWindow == m_pTop || !pActiveWindow->isVisible())
+            if (m_pTop)
             {
-                return false;
-            }
-            if (pActiveWindow->getType() == "CGUI/Editbox")
-            {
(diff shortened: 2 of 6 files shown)


Client/gui/CGUI_Impl.h
```diff
@@ -361,4 +364,7 @@ class CGUI_Impl : public CGUI, public CGUITabList
     bool         m_HasSchemeLoaded;
     SString      m_CurrentSchemeName;
     CElapsedTime m_RenderOkTimer;
+    const char*  m_currentSkin;
```
Dutchman101: Pointer never used, and uninitialized

Client/gui/CGUI_Impl.cpp
```diff
@@ -144,8 +144,17 @@ CGUI_Impl::~CGUI_Impl()
...
+void CGUI_Impl::CreateRootWindow()
+{
+    // Create dummy GUI root
+    m_pTop = reinterpret_cast<CEGUI::DefaultWindow*>(m_pWindowManager->createWindow("DefaultWindow", "guiroot"));
+    m_pSystem->setGUISheet(m_pTop);
+}
+
 void CGUI_Impl::SetSkin(const char* szName)
 {
+    m_currentSkin = szName;
```
Dutchman101: Unused 'm_currentSkin'

Client/gui/CGUI_Impl.cpp
```diff
@@ -165,9 +174,8 @@ void CGUI_Impl::SetSkin(const char* szName)
     // Destroy any windows we already have
     CEGUI::WindowManager::getSingleton().destroyAllWindows();
 
-    // Create dummy GUI root
-    m_pTop = reinterpret_cast<CEGUI::DefaultWindow*>(m_pWindowManager->createWindow("DefaultWindow", "guiroot"));
-    m_pSystem->setGUISheet(m_pTop);
+    // Clean up CEGUI - this also re-creates the root window
+    Cleanup();
```
Dutchman101, line 176: Double window destruction > Dangling pointer/UAF
(See line 175)

Client/gui/CGUI_Impl.cpp
```diff
@@ -1797,3 +1805,22 @@ CEGUI::Window* CGUI_Impl::GetMasterWindow(CEGUI::Window* wnd)
...
+            m_pWindowManager->destroyAllWindows();
+
+        // Clear redraw structures that may reference old elements
+        m_RedrawQueue.clear();
+        m_RedrawRegistry.clear();
+
+        // Recreate the root window (destroyed above via destroyAllWindows)
+        CreateRootWindow();
+    }
+    catch (...) {}
```
Dutchman101: Swallowing everything in `Cleanup()` > hides bugs, **'s debugging

Client/core/CLocalization.cpp
```diff
@@ -41,10 +41,7 @@ CLocalization::CLocalization(const SString& strLocale, const SString& strLocaleP
 
 CLocalization::~CLocalization()
 {
-    for (auto iter : m_LanguageMap)
```
Dutchman101, line 44: `m_pCurrentLang` became dangling (and a potential UAF).. you didn't account for it in the destructor after converting to unique_ptr. If accessed between destructing `m_LanguageMap` and `m_DictManager`, you will also have an UAF.

Client/gui/CGUI_Impl.cpp
```diff
@@ -165,9 +174,8 @@ void CGUI_Impl::SetSkin(const char* szName)
     // Destroy any windows we already have
```
Dutchman101: `m_pTop` UAF + race condition.. line 174-177 are executing it while invalid. Becomes valid again at line 178: Cleanup()

Dutchman101 (changes requested):

Client/gui/CGUI_Impl.cpp
Lpsd: lines 174-177 only contain the destroyAllWindows call which you mentioned as duplicated. Where exactly is the UAF

Client/gui/CGUI_Impl.cpp
Lpsd: I have no idea what you mean by "swallowing everything and hiding bugs", we're ensuring windows get cleared up where they previously were not. Can you elaborate on how this screws with debugging?

Client/core/CLocalization.cpp
Lpsd: became dangling, or was already dangling? I didn't touch m_pCurrentLang

Client/gui/CGUI_Impl.cpp
Lpsd: now I've cleared up the duplicate call, can you explain where the UAF or race condition is here? 

Considering when calling `Cleanup`, the object contained within `m_pTop` is destroyed on the CEGUI side (when calling `m_pWindowManager->destroyAllWindows()`) and we immediately re-allocate `m_pTop` with a new window in `CreateRootWindow()` ?

Client/gui/CGUI_Impl.cpp
Lpsd: Cleaned up the duplicate call, but again please can you explain where exactly the UAF or dangling pointer is here?

Client/core/CLocalization.cpp
Dutchman101: > became dangling, or was already dangling? I didn't touch m_pCurrentLang

@Lpsd

The old code manually deleted all CLanguage* objects.. `m_pCurrentLang` pointed to one of those objects. With the old manual deletion it was already dangling, but that didnt really matter because the destrctor was about to finish anyways.

You went from manual deletion to unique_ptr. When m_LanguageMap destructs, it destroys all of the unique ptr objects, but m_pCurrentLang still points to one of them

Between line 42 (destructor starts) and when m_LanguageMap destructs (line 44), m_pCurrentLang points to valid memory. After m_LanguageMap destructs, m_pCurrentLang becomes dangling

If anything touches m_pCurrentLang during shutdown, there you got it.

Client/gui/CGUI_Impl.cpp
Dutchman101: > Considering when calling `Cleanup`, the object contained within `m_pTop` is destroyed on the CEGUI side (when calling `m_pWindowManager->destroyAllWindows()`) and we immediately re-allocate `m_pTop` with a new window in `CreateRootWindow()` ?

@Lpsd 

Yes, you re-allocate it. But there's a window of time where m_pTop points to freed memory. That's the UAF. That you eventually fix it doesn't prevent what can happen during the destruction.

1. Line 174: destroyAllWindows() - destroys old windows, m_pTop becomes dangling

2. Line 175-177: Create new window, m_pTop now points to valid memory again
3.Later, line 178: Call Cleanup()

4. Inside Cleanup(): destroyAllWindows() - destroys the window m_pTop is pointing to
5. m_pTop is now dangling but hasn't been updated yet

6. Still inside Cleanup(): CreateRootWindow() - creates new window and assigns to m_pTop

Between step 4 and step 6, m_pTop is dangling.

Access m_pTop during that timeframe (event handlers, callbacks, other threads), there you have it, UAF

About the race condition: CGUI can fire events during window destruction. If any code tries to access m_pTop during the destroyAllWindows() call, it's reading freed memory

Client/gui/CGUI_Impl.cpp
Dutchman101: > I have no idea what you mean by "swallowing everything and hiding bugs", we're ensuring windows get cleared up where they previously were not. Can you elaborate on how this screws with debugging?

@Lpsd 

`catch (...) {}`:
This catches every possible exception and does nothing

What could go wrong inside Cleanup():

CleanDeadPool() throws
destroyAllWindows() throws
CreateRootWindow() throws
Memory allocation fails
CEGUI has an internal error

Furthermore, effects on debugging:

No log entry saying "Cleanup() failed"
No exception message explaining what broke
No stack trace to follow
Continues with potentially corrupted state
You get a crash 5 minutes later in completely unrelated code

Client/gui/CGUI_Impl.cpp
Dutchman101: > Cleaned up the duplicate call, but again please can you explain where exactly the UAF or dangling pointer is here?

@Lpsd 

The UAF is in Cleanup(). After destroyAllWindows() frees the window m_pTop points to, there's a gap before CreateRootWindow() creates a new one. During that gap, m_pTop points to freed memory. If any code touches m_pTop (callbacks, event handlers, redraw code), you get your UAF.

If you do `m_pTop = nullptr`, it won't point to freed memory.

Dutchman101 (approved):

Dutchman101:
All code reviews were addressed, and we worked together on some remaining minor issues.
Heap profiling confirmed the leak is fixed, nice

PR #4533, opened 2025-11-12
Fix isMTAWindowFocused & onClientMTAFocusChange

### LINKED ISSUE #4233: isMTAWindowFocused() returns false if you join a server while minimized
### Describe the bug

I recently started using isMTAWindowFocused() to reduce CPU usage of clients while they're minimized.

I join a server via a shortcut on my desktop like this: mtasa://94.23.158.180:22003/

I then maximize MTA once the server has been joined.

And noticed that it was as if my game was minimized (not rendering stuff and lowered FPS)

Image

Because even though the game is visible as you can see from the screenshot, isMTAWindowFocused() returns false when it should return true.

Minimizing and then maximizing the game does fix it.

I also added an event handler to see if onClientRestore triggers when first maximizing and it doesn't.

### Steps to reproduce

Add a shortcut like: mtasa://94.23.158.180:22003/
Double click on it, do something else so that it doesn't get maximized when it opens.
Once joined the server: crun isMTAWindowFocused()

### Version

Multi Theft Auto v1.6-release-23198

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

FileEX:
Fixed #4233 

This PR fixes several bugs related to the isMTAWindowFocused function:
- Bug #4233
- If the window is minimized or the user clicks outside the window after clicking "join," isMTAWindowFocused incorrectly returns true.
- Sometimes, when returning to the game after encountering the above bug, isMTAWindowFocused returns false even though the window has focus and full control.

Known issue: Occasionally, when applying the above fix (to prevent the bug) and returning to the window, the event gets duplicated — focus, lost focus, focus. This isn’t a major problem since the final state is correct, but I haven’t been able to determine why it happens (Windows seems to behave strangely).


ArranTuna:
I've tested your PR, it works, I see you have access so why don't you merge it?

FileEX:
I can't approve my own PRs

qaisjp (approved):
Static feels a bit unclean but otherwise looks fine

qaisjp:
Sorry that sounded a bit unenthusiastic: thanks for fixing this!!!

PR #4534, opened 2025-11-12
Fix button misalignments in the settings menu

Xenius97:
This PR fixes button misalignments in the settings menu that have been present for a while. The same issue currently exists in the 1.6 nightly as well.

Before:
image

image

-----

After:
image
image


Dutchman101 (approved):
Looks good across large and small resolutions, thanks

PR #4535, opened 2025-11-12
Add new function getSystemInfo

FileEX:
Original author @Pirulax (#2267)

> This function returns basic information about the client's system in a table.
> Example:
> 
> ```lua
> {
>     CPU = {
>         MaxClockSpeed = 3200,
>         NaNumberOfCores = 6,
>         Name = "AMD Ryzen 5 1600 Six-Core Processor",
>         NumberOfLogicalProcessors = 12
>     },
>     GPU = {
>         Name = "Radeon RX 570 Series",
>         RAM = 419328
>     },
>     TotalPhysicalMemory = 16727152 -- in bytes. I actually think it should be in MB (as the VRAM size)
> }
> ```
> Note: First call to this function is veeeeeery slow, as it has to query WMI.
> Subsequent calls are fast, as static variables are used to cache the data.
> 
> To clarify, this PR adds:
> 
> - CPU info (table under CPU key)
> - TotalPhysicalMemory
> - GPU info is just here for convenience (already available from dxGetStatus)

I’m not sure why the original PR was closed - probably due to being abandoned and not updated after review. As for the “privacy” aspect (though I don’t know if anyone still believes it actually exists), if it’s really necessary, we could add a checkbox in the settings. If someone wants to “maintain privacy,” they could simply uncheck it, and then ``getSystemInfo`` for that client would just return false. I’m waiting for feedback on whether this is actually needed.

This function will allow server owners to properly adjust their server for older computers, whose users experience FPS drops during rain, weapon/fx effects rendering. Without it, server creators cannot ensure smooth gameplay for everyone, even on a low-end PC.

### DIFF
--- a/Client/core/Graphics/CRenderItemManager.cpp
+++ b/Client/core/Graphics/CRenderItemManager.cpp
@@ -852,6 +852,19 @@ void CRenderItemManager::UpdateMemoryUsage()
 
 ////////////////////////////////////////////////////////////////
 //
+// CRenderItemManager::GetBasicGPUInfo
+//
+//
+//
+////////////////////////////////////////////////////////////////
+void CRenderItemManager::GetBasicGPUInfo(BasicGPUInfo& outInfo)
+{
+    outInfo.name = m_strVideoCardName;
+    outInfo.memoryKB = m_iVideoCardMemoryKBTotal;
+}
+
+////////////////////////////////////////////////////////////////
+//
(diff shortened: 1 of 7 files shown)


Fernando-A-Rocha:
This is awesome.
Like you said, to make everyone happy you can make a client setting to disable this function, but servers should be able to disallow players from joining if they have it disabled (like the screenshot feature) for whatever purposes they deem necessary.

PR #4536, opened 2025-11-12
Skip antivirus warning when running under Wine

### LINKED ISSUE #4459: Remove "no antivirus found" info box on startup for linux/wine users

Once in a while, when I start MTA on linux/wine, I get this popup saying no antivirus found and I tick the checkbox and click on OK.

This is fine for windows people but does not make sense on wine environment.

Check if client is running on linux/wine and if so do not show this info box.

_No response_

### Additional context

_No response_

### Security Policy

Xenius97:
Closes #4459

This PR skips antivirus warnings when running under Wine.
(It needs to be tested, but based on my research, it should be fine.)


qaisjp (changes requested):
Nice improvement.

We have a helper here already: https://github.com/multitheftauto/mtasa-blue/blob/1b4e5a2c093a102b94b5d554f126efa47fc93fc1/Client/cefweb/CWebApp.cpp#L24-L29

Maybe you can move this code to a shared location so both projects can share the code

qaisjp (approved):

Xenius97:
> Maybe you can move this code to a shared location so both projects can share the code

As i see you already did it, thanks.

gamelaster:
Huge thanks to both of you guys, that prompt was always super annoying!

PR #4539, opened 2025-11-13
Fix for the crouch key behavior

### LINKED ISSUE #488: Can't make ped crouch
**Describe the bug**
Cannot make ped crouch. When using 'setPedControlState' he will start crouching for a few milliseconds and stops.

**To Reproduce**
```lua
-- CLIENTSIDE SCRIPT
local ped1 = createPed(101, 0, 0, 3)
addCommandHandler('crouch_ped', function()
    setPedControlState(ped1, 'crouch', true)
end)
```

Also you can reproduce it by set control 'crouch' state to 'true' and set it to 'false' after 100 ms.

**Expected behavior**
Ped should stay crouched as there is no way to keep a ped crouched.

**MTA Client:** 1.5.3

**MTA Server:** 1.5.3

**Additional context**
From https://bugs.mtasa.com/view.php?id=9495

FileEX:
Fixed #488 

This PR fixes an old bug where holding the crouch key (default: C) causes the ped to stand up after a while, even though the key is still pressed. This behavior prevents the ped from staying crouched and requires manually releasing the key to keep the ped from standing.

image


qaisjp:
> This behavior prevents the ped from staying crouched and requires manually releasing the key to keep the ped from standing.

Do you mean:

This behavior prevents the ped from **standing up** and requires manually releasing the key to keep the ped from crouching.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -3124,14 +3124,10 @@ void CClientPed::ApplyControllerStateFixes(CControllerState& Current)
...
-            {
-                // Disable double crouching (another anim cut)
-                if (g_pClientGame->IsUsingAlternatePulseOrder())
-                    Current.ShockButtonL = 255;  // Do this differently if we have changed the pulse order
-                else
-                    Current.ShockButtonL = 0;
-            }
+
+            // Disable double crouching (another anim cut)
+            if (ulNow - m_ulLastTimeBeganCrouch < 500 * fSpeedRatio)
```
qaisjp, line 3129: 👍 

```
m_ulLastTimeBeganCrouch >= ulNow - 400.0f * fSpeedRatio
```
is equivalent to
```
ulNow - m_ulLastTimeBeganCrouch < 500 * fSpeedRatio
```

If you're going to change this logic you should change the others too, to be consistent. I like the style you picked



Can you also explain the change from 400ms to 500ms?  (Please make sure it's a float literal and not an integer literal.)

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -3124,14 +3124,10 @@ void CClientPed::ApplyControllerStateFixes(CControllerState& Current)
             // Disable the fire keys whilst crouching as well
             Current.ButtonCircle = 0;
             Current.LeftShoulder1 = 0;
-            if (m_ulLastTimeBeganCrouch >= ulNow - 400.0f * fSpeedRatio)
-            {
-                // Disable double crouching (another anim cut)
-                if (g_pClientGame->IsUsingAlternatePulseOrder())
```
qaisjp, line 3130: I don't really know how alternate pulse order works but it looks like this code was introduced here: https://github.com/multitheftauto/mtasa-blue/commit/cd9473bbe8ffc3097a54317d7cf7d54f60c08001

According to https://web.archive.org/web/20200809114050/https://bugs.mtasa.com/view.php?id=7250 

> **Description**  
> When the server's configuration parameter <[vehext_percent](https://wiki.multitheftauto.com/wiki/Server_mtaserver.conf#vehext_percent:~:text=running%20with%20setServerConfigSetting-,vehext_percent,-%3Cvehext_percent%3E0%3C/vehext_percent)> is greater than zero, then all players can not crouch always
>
> **Steps to reproduce**
> - Add or set in mtaserver.conf parameter <vehext_percent> to 1-100
> - Start the server and to try crouch

**However those repro steps will not work today** because the implementation of `IsUsingAlternatePulseOrder` has changed.

Today, `IsUsingAlternatePulseOrder` is managed by the [`latency_reduction`](https://wiki.multitheftauto.com/wiki/Server_mtaserver.conf#vehext_percent:~:text=500.%20Default%20%2D%20150-,latency_reduction,-%3Clatency_reduction%3E0%3C/latency_reduction) setting (default 0).

----

Please can you confirm that this code does not break crouching when using `<latency_reduction>1</latency_reduction>`?

qaisjp (changes requested):

Client/mods/deathmatch/logic/CClientPed.cpp
qaisjp: I am testing this now

qaisjp: | Your patch? | latency_reduction = 0 | latency_reduction = 1 |
| --- | --- | --- |
| Not applied | Holding crouch immediately stands up | Holding crouch works properly |
| Applied | Holding crouch works properly | Holding crouch works properly |

Which means that your patch works with any  `latency_reduction` state, great.

But we should also test to make sure we've haven't reintroduced this exploit: 

https://github.com/multitheftauto/mtasa-blue/blob/a665e0144e7f71a4c7898f024262393f019f6b7b/Client/mods/deathmatch/logic/CClientPed.cpp#L3108-L3112


Does anyone know how to trigger this exploit?

qaisjp: Found info on https://gta.fandom.com/wiki/Desert_Eagle

> It is possible to use a bug to make the gun shoot faster by doing this: Aim, fire, release both fire and aim, click the crouch button, and repeat. This bug also works via the "free aim" option in the controls menu. Choosing this, the player can rapidly tap the fire button over and over, making the weapon fire far faster. This applies to the other pistols as well.

I can't test this right now but someone should test to make sure we haven't unintentionally restored that exploit

Issue #3894, opened 2024-12-21
It is not possible to apply a shader to the vehiclegrunge256 texture after the latest update.


ZbyK77:
### Describe the bug

After the latest update (probably “Fix setVehicleDirtLevel (#3837)” r22823), it is no longer possible to apply shaders to the vehiclegrunge256 texture.
This blocks many opportunities for shader use.

### Steps to reproduce

1.  Run example code (Based on shader_tex_names)
`shader = dxCreateShader([[ float4 GetColor() { return float4( 1,0,1, 1 ); } technique tec0 { pass P0 { MaterialEmissive = GetColor(); } } ]])
engineApplyShaderToWorldTexture(shader,"vehiclegrunge*")`
2. Compare the results in the release after 22823 and before 22823.

### Version

Client: Multi Theft Auto v1.6-release-22829 (>22823)
Server: -

### Additional context

r22789
![Image](https://github.com/user-attachments/assets/eb8b4bd8-0827-43b7-b0ba-eadfbc3add51)

r22829
![Image](https://github.com/user-attachments/assets/d6906b9c-42f6-4297-99f7-be18fea4cc3d)

### Relevant log output

```shell
-
```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

tederis:
Duplicate of #3893

Issue #3893, opened 2024-12-21
Shaders used on vehicles no longer have backward compatibility and the components have incorrect textures applied [after #3837]


C0DELEVEL:
### Describe the bug

This PR probably broke backward compatibility of some shaders. Also after recent update (ver. 22829) texture of vehiclegrunge256 are not correctly displayed on vehicles and components.

Vehicle components probably use now different materials or texture and they are not shown up on shader_tex_names (displayed only on 'chassis' of ZR-350, missing on doors, bonnet, rear bumper, extra1 and whole Premier)
![Image](https://github.com/user-attachments/assets/4a1816aa-fc72-49a2-b255-d93d05d149a2)
![Image](https://github.com/user-attachments/assets/be876307-2221-48d2-b870-6a72616f512d)

ver. 22789 - shader works perfectly fine
![Image](https://github.com/user-attachments/assets/928d9551-c230-4700-a60c-c058c654f099)

ver. 22829 - shader does not work on vehicle components as bonnet and boot resulting in different shades of paint
![Image](https://github.com/user-attachments/assets/1347f19c-c88f-45e1-a85f-a77c2838de99)

### Steps to reproduce

- self-explained above with screenshoots

### Version

Client - r22829

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

Dutchman101:
Thanks, ff84958

PR #4541, opened 2025-11-13
Fix setVehicleDirtLevel again

### LINKED ISSUE #3233: Fix setVehicleDirtLevel

FileEX:
Fix #3233 

This PR is an improved version of #3837 - it’s not based on SilentPatch but on the original SA code. Additionally, the PR introduces a new special world property, ``vehicles_vanilla_dirt``, which defaults to **false** to avoid breaking backward compatibility like in the previous PR (#3894, #3893). The ``setVehicleDirtLevel`` function only works if ``vehicles_vanilla_dirt`` is set to **true**.

The PR uses a minimal number of hooks and is based on a clean approach, implementing the GTA function directly (other functions will be reversed later to keep this PR as small as possible).


Client/game_sa/CVehicleModelInfoSA.cpp
```diff
@@ -0,0 +1,148 @@
...
+static void CVehicleModelInfo__SetClump()
+{
+    _asm
+    {
+        mov eax, [esp+8]
+        push eax
+        call CVehicleModelInfoSAInterface::SetClump
+        add esp, 4
+        retn 4
+    }
```
qaisjp: Is it possible to rewrite this to not use `__asm`? 

GitHub Copilot says this:
```
// Assuming SetClump takes a single pointer argument and uses cdecl
using SetClumpFn = void(__cdecl*)(void*);

void ForwardSetClump(void* arg)
{
    const auto fn = reinterpret_cast<SetClumpFn>(&CVehicleModelInfoSAInterface::SetClump);
    fn(arg);
}
```

qaisjp (changes requested):
**This is not a full code review, but I have a question**

Instead of a special world property, why not just enable the patch when someone calls `setVehicleDirtLevel` for the first time?

Or is the case that this code does more than just make `setVehicleDirtLevel` work?

Client/game_sa/CVehicleModelInfoSA.cpp
FileEX: I can use VTBL hook to avoid asm

Client/game_sa/CModelInfoSA.h
```diff
@@ -288,24 +294,51 @@ class CVehicleModelVisualInfoSAInterface  // Not sure about this name. If somebo
...
+#define MAX_PER_VEHICLE_MATERIALS      48
+#define MAX_PER_VEHICLE_MATERIALS2     32
+#define MAX_PER_VEHICLE_MATERIALS3     16
+#define MAX_PER_VEHICLE_MATERIALS4     16
+#define MAX_PER_VEHICLE_DIRT_MATERIALS 32
+#define NUM_DIRT_LEVELS                16
+#define MAX_POSSIBLE_COLOURS           8
+#define MAX_POSSIBLE_UPGRADES          18
+#define MAX_NUM_REMAPS                 4
+#define NUM_DIRT_LEVELS                16
```
AlexTMjugador, line 307: `NUM_DIRT_LEVELS` is defined twice, seemingly not on purpose.

PR #4542, opened 2025-11-14
Implement `getResources` on client side

Xenius97:
This PR enables the usage of the `getResources()` function on the client side as well.  
Currently, the only way to list resources on the client is through workarounds for example, manually writing every resource name into a Lua table and iterating over them using `getResourceFromName`.  
This change provides a simpler and more transparent method, primarily useful for debugging purposes.

```lua
addCommandHandler("c_listresources", function()
    for i, res in ipairs(getResources()) do
        outputChatBox(i .. ": " .. getResourceName(res))
    end
end)
```

> [!IMPORTANT]
> This only lists resources that are running on the client side.


Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -391,6 +392,20 @@ int CLuaResourceDefs::GetResourceExportedFunctions(lua_State* luaVM)
     return 1;
 }
 
+int CLuaResourceDefs::GetResources(lua_State* luaVM)
```
qaisjp: Does `ArgumentParser<GetResources>` with `std::vector<CResource*> CLuaResourceDefs::GetResources()` not work?

qaisjp (changes requested):
Seems reasonable.

Can you flesh out a specific reason why someone would need this though?

You mentioned debugging purposes but I'm struggling to think of an actual reason to use this client side.

Xenius97:
> Can you flesh out a specific reason why someone would need this though?

A few months ago we had memory issues, and I ran into this while working on it as well. I had to solve it with a workaround trick, where I sent all resources to the client and ran the code there. Specifically, the problem is that performancebrowser does not show which elements are created per resource separately. Because of this, I wrote a script that prints how many elements are created for a given element type per resource. Other than that, I can’t really think of any other use case for it.


```lua
addCommandHandler("elementinfo", function(player, cmd, elementType)
	if Admin:isScripter(player) then
		if not elementType then
			Core:OutputToPlayer("Használd: /elementinfo [element típus]", player, "red", "Developer")
			return
		end

		local resources = {}
		for _, value in ipairs(getResources()) do
			if getResourceState(value) == "running" then
				table.insert(resources, getResourceName(value))
			end
		end
		triggerClientEvent(player, "devtools->elementInfo", resourceRoot, resources, elementType)
	end
end)
```

```lua
addEvent("devtools->elementInfo", true)
addEventHandler("devtools->elementInfo", root, function(resources, elementType)
	local totalCount = 0
	for _, name in ipairs(resources) do
		local resource = getResourceFromName(name)
		if resource and getResourceState(resource) == "running" then
			local count = (#getElementsByType(elementType, getResourceRootElement(resource)))
			if count > 0 then
				totalCount = totalCount + count
				Core:OutputToPlayer("Resource: "..name.." | Element típus: "..elementType.." | Darabszám: "..count, _, "green", "Developer")
			end
		end
	end

	Core:OutputToPlayer("Összes "..elementType.." elem darabszám: "..totalCount, _, "yellow", "Developer")
end)
```

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
Xenius97: I've changed to it (this was one of my early PR's and didn't use ArgumentParser)

dmi7ry:
This can be used for cheaters' purposes. Perhaps it would be better if such a possibility were allowed by a flag in the server config.

Xenius97:
> This can be used for cheaters' purposes. Perhaps it would be better if such a possibility were allowed by a flag in the server config.

I dont think so. This will only list running resources, you can make a list of them from client cache too.
If they want to abuse it they could now too.

dmi7ry:
> I dont think so. This will only list running resources, you can make a list of them from client cache too.
> If they want to abuse it they could now too.

They can use this specifically to find out what hidden resources are on the client (that are not in the cache).

Xenius97:
> They can use this specifically to find out what hidden resources are on the client (that are not in the cache).

(i'm was a bit wrong, but you can get now too)

samr46:
> They can use this specifically to find out what hidden resources are on the client (that are not in the cache).

They can get a list of all client side resources now by checking resource elements (element tree).

Xenius97:
In my opinion, even now there are plenty of ways to find out what’s running on the server, this is just an easier way. I wouldn’t consider it a potential attack surface, but if the management sees it that way, then just close the PR.

qaisjp:
> They can get a list of all client side resources now by checking resource elements (element tree).

My thoughts align with this. This data is already in the client memory so we're not leaking more than is necessary. We're only making it available to script.

If an attacker is sophisticated enough to run a Lua script client side, then they could:
- read it out of the element tree instead 
- read it straight from memory 
- look at the http server instead

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -411,6 +412,16 @@ int CLuaResourceDefs::GetResourceState(lua_State* luaVM)
...
 }
 
+std::vector<CResource*> CLuaResourceDefs::GetResources()
+{
+    std::vector<CResource*> resources;
+    for (auto iter = m_pResourceManager->IterBegin(); iter != m_pResourceManager->IterEnd(); ++iter)
+    {
+        resources.push_back(*iter);
+    }
+    return resources;
```
qaisjp: I can make the change but it would be good for this to use a modern looping mechanism -- will make the change when I'm free unless you get to it first

Xenius97: What do you think about this? I've simplified
https://github.com/multitheftauto/mtasa-blue/pull/4542/commits/bad35dcff063bd04e77a1c24b8b654a6ccc1c8d0

qaisjp (approved):
Elegant, thanks

PR #4544, opened 2025-11-14
Fix memory leak in engineImageLinkDFF

### LINKED ISSUE #4288: Image streaming memory leak
### Describe the bug

This only occurs when engineSetAsynchronousLoading(true, true) is used

The error occurs when the player has some txd/dff loaded via an IMG block and he reconnects, disconnects or the resource that loads the IMG block is interrupted. I've tried a few possible solutions, such as taking the vehicle out of rendering when the 'onClientResourceStop' event 
is fired, I've also tried using the "engineStreamingFreeUpMemory" functions with high values, I've also tried the "engineStreamingReleaseModel" function.

Nothing seems to work.

### Steps to reproduce

1 - download the attached resource
2 - start it
3 - restart a few times or simply reconnect a few times
4 - you will see that the streaming memory usage will not decrease, it will always increase.

### Version

Client: Multi Theft Auto v1.6-release-23247
Server: MTA:SA Server v1.6-release-23219

### Additional context

Below is the video demonstrating what happened and the resource for testing

[ImageBug.zip](https://github.com/user-attachments/files/21151971/ImageBug.zip)

https://www.youtube.com/watch?v=StmRB-dp8A0

### Relevant log output

```shell

```

### Security Policy

FileEX:
Fixed #4288 

After updating the archive information and then calling the streaming ``CStreamingSA::ReInit`` (``engineRestreamWorld``), the ``RwObject`` associated with the original archive was “lost”, and a new one was loaded for the new archive.

Sequence:
```
link dff -> restream -> keep old RwObject -> create RwObject for new archive

Restart resource:
restore dff (unlink) -> restream -> keep previously loaded new RwObjects -> create RwObject from original archive
```

With every subsequent call to ``engineImageLinkDFF``, new ``RwObjects`` were created without removing the previous ones. As a result, after each resource restart or reconnect, the number of ``RwObjects`` kept increasing, depending on how many models were inside the IMG. This is what caused the memory leak.

### DIFF
--- a/Client/game_sa/CStreamingSA.cpp
+++ b/Client/game_sa/CStreamingSA.cpp
@@ -336,9 +336,9 @@ void CStreamingSA::SetStreamingInfo(uint modelid, unsigned char usStreamID, uint
     CStreamingInfo* pItemInfo = GetStreamingInfo(modelid);
 
     // We remove the existing RwObject because, after switching the archive, the streamer will load a new one.
-    // ReInit doesn’t delete all RwObjects unless certain conditions are met.
+    // ReInit doesn't delete all RwObjects unless certain conditions are met.
     // In this case, we must force-remove the RwObject from memory, because it is no longer used,
-    // and due to the archive change the streamer no longer detects it and therefore won’t delete it.
+    // and due to the archive change the streamer no longer detects it and therefore won't delete it.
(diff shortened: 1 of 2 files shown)


Dutchman101:
Good job, but fingers crossed it won't break anything that wasn't immediately obvious in your local tests.

Dutchman101 (approved):

PR #4546, opened 2025-11-15
Add new `vehicleburnblowup` special property

### LINKED ISSUE #4522: vehicleburnexplosions does nothing
### Describe the bug
Disabling `vehicleburnexplosions` special property currently does nothing, vehicle still exploding when health is bellow 250

### Steps to reproduce
Create a vehicle and set it's health to 0-50-100

### Version
1.6.0-9.23324.0

### Security Policy

Xenius97:
Fixes #4522

This PR makes it possible to disable the default vehicle explosion.
In the base code, when a vehicle’s health reaches zero, it explodes after 5 seconds. This special property completely disables that mechanism, so there’s no need to struggle with `setTimer` and `damageProof`.

https://github.com/user-attachments/assets/cf687e1d-8757-4ded-b5ce-5f3ed3db07a6


samr46:
Hi. This isn't what "vehicleburnexplosions" property is for. This property shouldn't prevent vehicles from exploding. We have other functions for this. It disables additional explosions created for aircraft during burning (it's in the description of this property). The aircraft has to explode in the end like any other vehicle.

This fixes not existing issue.

There is a good argument about renaming this property since the name can be misleading (although "burn explosions" is quite literally about explosions during burning).

Xenius97:
> There is a good argument about renaming this property since the name can be misleading (although "burn explosions" is quite literally about explosions during burning).

Hi, i've separated it from "vehicleburnexplosions"

qaisjp:
Please could you update the pull request description to be clearer that this is a brand new feature and not a bug fix?

Xenius97:
> Please could you update the pull request description to be clearer that this is a brand new feature and not a bug fix?

It’s basically a new feature that prevents vehicles from exploding. It removes the check from the original code that causes the vehicle to explode 5 seconds after its health reaches zero. If this is enabled, the car will keep burning indefinitely. At the moment, this can only be achieved using a `setTimer` + `damageProof`.

PR #4547, opened 2025-11-15
Add `getPedAnimation` for serverside too

### LINKED ISSUE #1546: Server side for getPedAnimation
Server side for getPedAnimation

**Need for**
https://github.com/multitheftauto/amx/issues/64

**Additional context**
https://wiki.multitheftauto.com/wiki/GetPedAnimation

Xenius97:
Fixes #1546

This PR adds missing `getPedAnimation` for serverside too. Every return is same as the client.

```
Executed command: setPedAnimation(getRandomPlayer(), "ped", "WOMAN_walknorm")
Executed command: getPedAnimation(getRandomPlayer())
Command results: ped [string], WOMAN_walknorm [string], -1 [number], true [boolean], true [boolean], true [boolean], true [boolean], 250 [number], false [boolean]
```


ArranTuna:
Doesn't this just return the last used server side setPedAnimation data? So most of the time, wouldn't accurately reflect their true animation. But I suppose as long as the wiki page warns of that.

Xenius97:
> Doesn't this just return the last used server side setPedAnimation data? So most of the time, wouldn't accurately reflect their true animation. But I suppose as long as the wiki page warns of that.

This will return animation data only when it's active. If player is not in animation anymore it will return false.

ArranTuna:
Tested and works.

FileEX:
> Tested and works.

Animation synchronization was introduced in PR #3520 and #4326.

FileEX (approved):

tederis (approved):

PR #4548, opened 2025-11-16
Implement new `getResourceFiles` function

### LINKED ISSUE #4543: Add getResourceFiles function

Resources have files listed in their meta.xml. Files need to be listed there in order for certain functions to work like downloadFile (clientside) or fileGetContents (shared).

Developers have no way of knowing via scripting if a file (using its path) is in a resource's meta.xml.

Problem 1: you want to make a file download system. You use pathListDir to obtain files in a folder. These files should be listed in meta.xml but you don't know for sure. You want to ask the client to download only certain files that you know are listed, to avoid errors. You have no function to find them, other than manually parsing the resource's meta.xml and sending that info to the client. If you do that, you will have even more difficulty parsing the newly added globs used to load files dynamically like models/*.dff.

New shared function `getResourceFiles(resource theResource [, bool includeAttributes = false ])`
Returns table with file paths.
If includeAttributes is true it returns a table with file paths as key and the <file/> node attributes, like `download = "false"`.

x

### Additional context

PR #3603 was almost ready but it was abandoned unfortunately 

### Security Policy

Xenius97:
Closes #4543

This PR makes possible to list files within a given resource along with specific filtering conditions.

Usage:
```
getResourceFiles(resource theResource [, bool includeAttributes = false, string filter = "all/map/script/config/file" ]])
```

```lua
iprint(getResourceFiles())
iprint(getResourceFiles(getResourceFromName("cam2rt"), false, "script"))
```

> [!IMPORTANT]
> The client-side function will only be able to see the files that are actually sent to the client.


Fernando-A-Rocha:
Thank you!
I will use this function for my "new models" resource:
- in meta.xml I have file globs like models/*.dff 
- I will use getResourceFiles to find all file paths starting with "models/"
- i will be able to determine if they have the download="false" attribute,
- then perform some logic to store the new models info in a table
- and send the list to clients so they can do downloadFile or access the file directly to load the mod

Client/mods/deathmatch/logic/luadefs/CLuaResourceDefs.cpp
```diff
@@ -391,6 +396,113 @@ int CLuaResourceDefs::GetResourceExportedFunctions(lua_State* luaVM)
     return 1;
 }
 
+namespace
+{
+    CLuaResourceDefs::eResourceFileFilter ParseFilterString(const std::string& strFilter)
```
Pirulax:
```suggestion
    auto ParseFilterString(const std::string& strFilter) -> eResourceFileFilter
```

PR #4549, opened 2025-11-16
Fix area name showing for a second when joining to server for first time

### LINKED ISSUE #560: Zone name appears for a moment when joining the server.
**Describe the bug**
When joining a server you can sometimes see "Ganton" zone name in the bottom of the screen and after a second it disappears.

**To reproduce**
It appears most often when joining the server for the first time after opening MTA.

**Expected behaviour**
Maybe MTA can hide the HUD until joined which is the point at which the server can control HUD visibility.

**Screenshots**
image

**Version**
Unstated at mantis.

**Additional context**

From https://bugs.mtasa.com/view.php?id=8334

Xenius97:
Fixes #560


Fernando-A-Rocha (approved):
HUD_AREA_NAME will be made visible later in the process when the camera fades in (`CCameraRPCs::FadeCamera`), so this is good!

Xenius97:
> HUD_AREA_NAME will be made visible later in the process when the camera fades in (`CCameraRPCs::FadeCamera`), so this is good!

Yes, that’s exactly why I thought of this solution. I tested it and it didn’t break anything, the script functions worked the same, and in normal freeroam I could also see the area names after connecting.
It shouldn’t break anything, since it turns it back on later once everything has loaded.

Pirulax:
Really should add that as a comment, because that was my first thought.

FileEX (approved):

PR #4551, opened 2025-11-17
Fix img crash when game shutting down

Xenius97:
This PR fixes the following crash when game shuts down:
```
Fatal Exception: Yes
Stack trace:
  0x49646550 [0x49646550]
  0x409172 [lookup failed] [0x409172]
  0x156D017 [lookup failed] [0x156D017]
  CClientIMG::StreamDisable (F:\Git\mtasa-blue\Client\mods\deathmatch\logic\CClientIMG.cpp:195) [0x65BD1411]
```

When the game shuts down there is no reason to do anything with IMG containers, because the entire process is exiting and all memory is freed anyway.


Client/mods/deathmatch/logic/CClientIMG.cpp
```diff
@@ -174,7 +174,8 @@ bool CClientIMG::StreamEnable()
 
 bool CClientIMG::StreamDisable()
 {
-    if (!IsStreamed())
+    // If game is shutting down or not streaming is not available, do nothing
+    if (g_bClientShuttingDown || !IsStreamed())
```
FileEX: The crash is caused by calling `CClientGame::RestreamWorld` during game shutdown, so that’s where you should add the condition to prevent future crashes for the same reason

FileEX (approved):

PR #4560, opened 2025-11-19
Return Event ID from triggerLatentClient/ServerEvent

### LINKED ISSUE #3998: Return Event ID from triggerLatentClient/ServerEvent

Currently, `triggerLatentClientEvent` and `triggerLatentServerEvent` only return true, making it impossible to directly obtain the event ID. As a workaround, developers must compare the list of active latent events before and after calling triggerLatentClientEvent, which is inefficient.

```lua
local currentEvents = {}
for _, eventID in pairs(getLatentEventHandles(player)) do
    currentEvents[eventID] = true
end

triggerLatentClientEvent(player, "openImage", 2500000, false, player, imageData)

for _, eventID in pairs(getLatentEventHandles(player)) do
    if not currentEvents[eventID] then
        openedImages[player].latentEvent = eventID
        break
    end
end
```

This extra logic is unnecessary when the function could simply return the event ID directly.

Modify `triggerLatentClientEvent` and `triggerLatentServerEvent` so that it returns the latent event ID instead of true.

```lua
local eventID = triggerLatentClientEvent(player, "openImage", 2500000, false, player, imageData)
if eventID then
    openedImages[player].latentEvent = eventID
end
```

_No response_

### Additional context

_No response_

### Security Policy

Xenius97:
Closes #3998

Currently, you only receive a `true/false` value indicating whether it was successful. I already mentioned in the issue some time ago that this is quite inconvenient, as you constantly have to query and compare it with the `getLatentEventHandles` table if you need the ID.
With the current solution, you get the ID directly `(SSendHandle = uint)`, so there’s no need for any workaround. If the request fails, you simply receive a false.


Server/mods/deathmatch/logic/CGame.h
```diff
@@ -465,6 +465,7 @@ class CGame
                     NetServerPacketPriority packetPriority, NetServerPacketReliability packetReliability,
                     ePacketOrdering packetOrdering = PACKET_ORDERING_DEFAULT);
     void SendPacketBatchEnd();
+    uint GetLastSentHandle() const { return m_LastSentHandle; }
```
FileEX: noexcept and `std::uint32_t`

FileEX (changes requested):

Issue #3842, opened 2024-11-04
Increase grass limit


lopezloo:
### Is your feature request related to a problem? Please describe.

Increase grass draw distance & amount limit.

### Describe the solution you'd like

-

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

PR #4567, opened 2025-11-22
Increase grass rendering distance

Xenius97:
This PR increases the grass rendering distance.
This is only one part of fully implementing the idea in #3842.
The problem is that even if we increase the view distance, it’s simply not enough. By default, the grass buffer is limited to 32, and to exceed this limit, the TriPlantBuffer would also need to be rewritten. I’m not that deep into it yet, but the view-distance increase works on its own, so if someone wants to take it further, they only need to handle increasing the grass buffer.

*getGrassDrawDistance*
```lua
closeDistance, farDistance = getGrassDrawDistance()
```

*setGrassDrawDistance*
```lua
bool = setGrassDrawDistance(closeDistance, farDistance)
```

*resetGrassDrawDistance*
```lua
bool = resetGrassDrawDistance()
```

Before:
mta-screen_2025-11-21_21-02-12

After:
mta-screen_2025-11-21_21-02-21


Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -983,6 +999,11 @@ int CLuaWorldDefs::setFogDistance(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaWorldDefs::SetGrassDrawDistance(float closeDistance, float farDistance)
+{
+    return CStaticFunctionDefinitions::SetGrassDrawDistance(closeDistance, farDistance);
```
FileEX: If the function always returns true, then there’s no reason for it to be a bool - it should be void and return nil.

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -1060,6 +1081,11 @@ int CLuaWorldDefs::resetFogDistance(lua_State* luaVM)
     return 1;
 }
 
+bool CLuaWorldDefs::ResetGrassDrawDistance()
+{
+    return CStaticFunctionDefinitions::ResetGrassDrawDistance();
```
FileEX: If the function always returns true, then there’s no reason for it to be a bool - it should be void and return nil.

Server/mods/deathmatch/logic/packets/CMapInfoPacket.cpp
```diff
@@ -264,6 +268,14 @@ bool CMapInfoPacket::Write(NetBitStreamInterface& BitStream) const
         BitStream.Write(m_fFogDistance);
     }
 
+    // Grass draw distance
+    BitStream.WriteBit(m_bOverrideGrassDrawDistance);
+    if (m_bOverrideGrassDrawDistance)
+    {
+        BitStream.Write(m_fGrassCloseDistance);
+        BitStream.Write(m_fGrassFarDistance);
+    }
```
FileEX: You should never add new data in the middle of a stream. Any new data should always be added at the end

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -797,6 +802,17 @@ int CLuaWorldDefs::getFogDistance(lua_State* luaVM)
     return 1;
 }
 
+std::variant<CLuaMultiReturn<float, float>, bool> CLuaWorldDefs::GetGrassDrawDistance()
+{
+    float fCloseDistance, fFarDistance;
+    bool  bSuccess = CStaticFunctionDefinitions::GetGrassDrawDistance(fCloseDistance, fFarDistance);
+
+    if (bSuccess)
+        return std::make_tuple(fCloseDistance, fFarDistance);
```
FileEX: Maintain consistency with the client-side version of the function and return CLuaMultiReturn

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -404,6 +404,20 @@ class CGame
     float GetFogDistance() { return m_fFogDistance; }
     void  SetFogDistance(float& fFogDistance) { m_fFogDistance = fFogDistance; }
 
+    bool HasGrassDrawDistance() { return m_bOverrideGrassDrawDistance; }
```
FileEX: const noexcept

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -404,6 +404,20 @@ class CGame
...
+    void SetHasGrassDrawDistance(bool bOverrideGrassDrawDistance) { m_bOverrideGrassDrawDistance = bOverrideGrassDrawDistance; }
```
FileEX: noexcept

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -404,6 +404,20 @@ class CGame
...
+
+    void  GetGrassDrawDistance(float& fCloseDistance, float& fFarDistance)
```
FileEX: const noexcept

Server/mods/deathmatch/logic/CGame.h
```diff
@@ -404,6 +404,20 @@ class CGame
...
+    {
+        fCloseDistance = m_fGrassCloseDistance;
+        fFarDistance = m_fGrassFarDistance;
+    }
+    void SetGrassDrawDistance(float fCloseDistance, float fFarDistance)
```
FileEX: noexcept

Client/multiplayer_sa/CMultiplayerSA.h
```diff
@@ -192,6 +192,9 @@ void                InitHooks();
     float GetFogDistance();
     void  SetFogDistance(float fDistance);
     void  RestoreFogDistance();
+    void  SetGrassDrawDistance(float fCloseDistance, float fFarDistance);
+    void  GetGrassDrawDistance(float& fCloseDistance, float& fFarDistance) const;
+    void  ResetGrassDrawDistance();
```
FileEX: override

Client/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -1818,6 +1823,25 @@ int CLuaWorldDefs::ResetFogDistance(lua_State* luaVM)
...
+{
+    g_pMultiplayer->SetGrassDrawDistance(closeDistance, farDistance);
+    return true;
+}
+
+bool CLuaWorldDefs::ResetGrassDrawDistance()
+{
+    g_pMultiplayer->ResetGrassDrawDistance();
+    return true;
+}
```
FileEX: If the function always returns true, then there’s no reason for it to be a bool - it should be void and return nil.

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2064,6 +2064,24 @@ void CMultiplayerSA::RestoreFogDistance()
     }
 }
 
+void CMultiplayerSA::SetGrassDrawDistance(float fCloseDistance, float fFarDistance)
+{
+    MemPutFast<float>(0xC02DBC, fCloseDistance);
```
FileEX: Look at this https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md#avoid-magic-numbers

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -10686,6 +10686,17 @@ bool CStaticFunctionDefinitions::GetFogDistance(float& fFogDist)
     return false;
 }
 
+bool CStaticFunctionDefinitions::GetGrassDrawDistance(float& fCloseDistance, float& fFarDistance)
```
FileEX: Here and in many other places there is Hungarian notation
https://github.com/multitheftauto/mtasa-docs/blob/main/mtasa-blue/CODING_GUIDELINES.md#naming-conventions

FileEX (changes requested):

Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
Xenius97: Thanks, i will check your reviews shortly and fix everything.

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -2064,6 +2067,61 @@ void CMultiplayerSA::RestoreFogDistance()
...
+    // 0 = low, 1 = medium, 2 = high, 3 = very high
+    CGameSettings* pSettings = pGameInterface ? pGameInterface->GetSettings() : nullptr;
+    const unsigned int fxQuality = pSettings ? pSettings->GetFXQuality() : 2;
+    float actualFarDistance = farDistance;
+    
+    // Keep base GTA:SA logic
+    if (fxQuality < 2)
+    {
+        actualFarDistance = farDistance / 2.0f;
+    }
```
FileEX:
```cpp
if (fxQuality)
    farDistance /= 2.0f;
```

Server/mods/deathmatch/logic/luadefs/CLuaWorldDefs.cpp
```diff
@@ -797,6 +802,17 @@ int CLuaWorldDefs::getFogDistance(lua_State* luaVM)
...
 
+CLuaMultiReturn<float, float> CLuaWorldDefs::GetGrassDrawDistance()
+{
+    float closeDistance, farDistance;
+    bool  bSuccess = CStaticFunctionDefinitions::GetGrassDrawDistance(closeDistance, farDistance);
+
+    if (bSuccess)
+        return {closeDistance, farDistance};
+
+    return {};
```
FileEX: Will this return nil/false? I’m not sure how the parser will interpret it

Xenius97: fixed, thanks

Client/multiplayer_sa/CMultiplayerSA.cpp
Xenius97: done, thanks

FileEX (dismissed):

PR #4568, opened 2025-11-22
Add timestamp display option for chatbox/console

### LINKED ISSUE #1029: Add a timestamp option in the chat configuration
A lot of games nowadays that have a chat system included have the option to see in which hour, minute and second that message was sent. I think MTA could have this and benefit from it.

We could add an option on Config -> Interface -> Options that adds a timestamp to the chat.

**Additional context**
I know that is possible to make this in your own server with resources, but I don't think this is worth it. Besides, what happens if you want to have a timestamp on a server that just didn't made one?

Also I think we should see if adding this could break something in the chat because of the maximum lenght of each message. I don't think this would be a problem tho.

Xenius97:
Closes #1029

This PR adds timestamp display option for chatbox/console

image

image

image


omar-o22:
Loved it.

SpeedyFolf:
It'd be nice if chat and console were separate options, I'd like to have timestamp shown in console but not chat, I don't want to clutter chat too much but console is a bigger window

Client/core/CGUI.cpp
qaisjp, line 1: Toggling the setting should show/hide the time in the chat and the console.

The current implementation conditionally modifies the chat string to include the time.

Instead I suggest you save the timestamp string + message string separately, and include the timestamp string conditionally at render time.

Client/core/CConsole.cpp
qaisjp, line 1: No need to change Console to support toggling. (mainly because the console is implemented as simply appending raw text to `m_pHistory` / a `CGuiMemo`, and I don't want to introduce another copy of the buffer)

Client/core/CClientVariables.cpp
```diff
@@ -323,6 +323,7 @@ void CClientVariables::LoadDefaults()
     DEFAULT("chat_text_alignment", Chat::Text::Align::LEFT);                          // chatbox horizontal text alignment
     DEFAULT("server_can_flash_window", true);                                         // allow server to flash the window
     DEFAULT("allow_tray_notifications", true);                                        // allow scripts to create tray balloon notifications
+    DEFAULT("show_time_in_chat", false);                                              // show time prefix in chat messages
```
qaisjp, line 326: @SpeedyFolf's feedback is valid:

> It'd be nice if chat and console were separate options, I'd like to have timestamp shown in console but not chat, I don't want to clutter chat too much but console is a bigger window

qaisjp (changes requested):
thanks for the pull request, here's some feedback

Pieter-Dewachter:
Maybe also a good idea to add it to the `getChatboxLayout()` function in Lua (`CLuaGUIDefs.cpp`)?

PR #4573, opened 2025-11-23
Add seat parameter to `setPedEnterVehicle`

MohabCodeX:
This PR adds an optional seat parameter to `setPedEnterVehicle`.

The implementation **reuses the existing seat validation logic** from `warpPedIntoVehicle`.
When a seat is specified, it validates that:
- The seat number doesn't exceed the vehicle's maximum passenger count
- The vehicle model supports passengers (some models return 255 for invalid passenger counts)

### Syntax:
```lua
bool setPedEnterVehicle ( ped thePed [, vehicle theVehicle = nil, bool passenger = false, int seat = nil ] )
```

Optional Arguments:
- theVehicle: The vehicle to enter. If not specified, the ped will enter the nearest vehicle
- passenger: If set to true and no seat is specified, the ped will enter as a passenger
- seat: The seat number to enter (0-8). If specified, this overrides the passenger parameter

> [!NOTE]
> The **seat** parameter takes priority over the **passenger** parameter.

Examples:
```lua
-- Enter driver seat
setPedEnterVehicle(ped, vehicle, false)

-- Enter as passenger (auto-determines seat)
setPedEnterVehicle(ped, vehicle, true)

-- Enter specific seats
setPedEnterVehicle(ped, vehicle, nil, 0)  -- Driver seat
setPedEnterVehicle(ped, vehicle, nil, 1)  -- Front passenger
setPedEnterVehicle(ped, vehicle, nil, 2)  
```

https://github.com/user-attachments/assets/fad40f17-716b-493a-935d-5ca302496fa6

#### Testing
[hit.zip](https://github.com/user-attachments/files/23696102/hit.zip)


qaisjp:
This is really cool. Great pull request description too.

What do you think about just letting an int be provided as the third argument? I think this would be cleaner and remove the need for a dummy/unused argument.

You should be able to use a `std::variant` to achieve this, as demonstrated here: https://github.com/multitheftauto/mtasa-blue/blob/3d02b86df9f26934c81a96f1feacdc3e112bffc0/Server/mods/deathmatch/logic/luadefs/CLuaColShapeDefs.cpp#L825-L843

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -6641,17 +6621,36 @@ bool CClientPed::EnterVehicle(CClientVehicle* pVehicle, bool bPassenger)
...
-        uiSeat = 1;
+        // Explicit seat specified - validate it (reuse validation logic from warpPedIntoVehicle)
+        uiSeat = optSeat.value();
+        const unsigned char ucMaxPassengers = CClientVehicleManager::GetMaxPassengerCount(pVehicle->GetModel());
+        
+        if (uiSeat > ucMaxPassengers)
+            return false;
+
+        if (uiSeat > 0 && ucMaxPassengers == 255)
+            return false;
```
qaisjp: If this is copied from the `warpPedIntoVehicle` method, then I think it'd make sense to extract to a function that can be used by both. E.g. `IsValidSeat`

Client/mods/deathmatch/logic/CClientVehicleManager.cpp
```diff
@@ -486,6 +486,24 @@ unsigned char CClientVehicleManager::GetMaxPassengerCount(unsigned long ulModel)
     return 0xFF;
 }
 
+bool CClientVehicleManager::IsValidSeat(unsigned long ulModel, unsigned char ucSeat)
+{
+    // Camper only has 3 seats (0-2)
+    if (static_cast<VehicleType>(ulModel) == VehicleType::VT_CAMPER && ucSeat > 2)
+        return false;
```
qaisjp, line 493: I wonder why this is a special case not accounted for by `GetMaxPassengerCount` but I won't argue with it since it was there before 😆

qaisjp (approved):
Great job, thanks for contributing this.

qaisjp:
I'm moderately confident in this change, so I'm going to let this merge into 1.7 despite the feature freeze.

qaisjp:
Thanks for sending this patch @MohabCodeX, you're churning out lots of great code!

Issue #3860, opened 2024-11-19
Ability to launch a secondary client


Xenius97:
### Is your feature request related to a problem? Please describe.

The current solution is not very energy-efficient. You can connect to a server multiple times using a VM or, if you have a notebook, through that. I think there could be a better solution: for instance, if I specify `-cl2` or `-second` (or something similar) through the launcher, it could start a secondary client, similar to how FiveM handles it.

The program would store this setting and only allow access to servers where it is explicitly enabled (e.g., `<allow_second_client>1</allow_second_client>` -> doesn't allow it by default).
If this is not enabled, you would only be able to connect using the primary (main) client.

### Describe the solution you'd like

This solution would make much better use of hardware and remove the need to struggle with VMs.
Second instance should have a different path for resource-priv cache, eg: `mods/deathmatch/cl2`

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not about a cheat or security vulnerability.

TracerDS:
Whats the purpose of it except debugging?
Also its conflicting with #3854

Xenius97:
> Whats the purpose of it except debugging?

Honestly, I think it wouldn’t be useful for anything other than debugging, but it would greatly assist server developers in their work. There are surely people among us whose computers can’t handle a VM or who don’t have another machine nearby.

I have the option to run a VM, for example, but even with a shared GPU, MTA becomes unmanageable, achieving only 10-15 FPS even with a 3060 card.

ghost:
> Whats the purpose of it except debugging? Also its conflicting with [#3854](https://github.com/multitheftauto/mtasa-blue/pull/3854)

#3854 is opt-in in mtaserver.conf

ffsPLASMA:
Im not a fan of this as this will open doors for people to abuse. This will add another weak point to client to the already weak AC.

Fernando-A-Rocha:
There is no conflict with #3584 which will allow clients with same serial on a server.

I support this. For debugging purposes, using Windows VMs is a huge pain in the ass for developers. It would be awesome to run multiple MTA clients on the same machine. If impossible to run more than 1 MTA client with the same GTA game, it could require multiple GTA installations, that's fine I think?

TracerDS:
> I have the option to run a VM, for example, but even with a shared GPU, MTA becomes unmanageable, achieving only 10-15 FPS even with a 3060 card.

[...]

PR #4574, opened 2025-11-23
Ability to launch a secondary client

### LINKED ISSUE #3970: Unable to control game in VM
### Describe the bug

When i'm using Windows Sandbox or Hyper-V game control is almost impossible due mouse is extremely sensitive compared to normal.

### Steps to reproduce

Use Windows Sandbox

My configuration:
```wsb
<Configuration>
  <GpuAcceleration>Enable</GpuAcceleration>
  <MappedFolders>
    <MappedFolder>
      <HostFolder>F:\Games\Sandbox</HostFolder>
      <Writable>true</Writable>
    </MappedFolder>
  </MappedFolders>
</Configuration>
```

### Version

22916

### Security Policy

Xenius97:
Closes #3970, #3860

This PR allows you to launch a secondary client. Similar to FiveM, it can be accessed using the “-cl2” command line argument.
The secondary client uses its own configuration files and resource cache, avoiding any conflicts. Your serial remains the same, nothing changes there.
To use it, you need to disable the double-serial check in mtaserver.conf.

```xml
<check_duplicate_serials>0</check_duplicate_serials>
```

From a development perspective, if you have two monitors, it can be extremely helpful to test things like synchronization even before potentially publishing a script.
You can also be in two places at the same time, which is an advantage when working with multiple servers.
I hope this feature can be included, and that MTA server developers will be able to make use of the potential it offers.

Please let me know if you have any requests or ideas related to the change, and I’ll do my best to accommodate them.


FileEX:
https://github.com/multitheftauto/mtasa-blue/compare/master...feature/multi-client-support

Xenius97:
> [master...feature/multi-client-support](https://github.com/multitheftauto/mtasa-blue/compare/master...feature/multi-client-support)

Yes i saw but that looks unfinished or idk whats with that now. Isn't merged, has no update since june.

G-Moris:
How about running an unlimited number of clients? Or maybe up to 5-10?

Xenius97:
> How about running an unlimited number of clients? Or maybe up to 5-10?

I think it’s doable if we completely scrap this whole “mutex” and “GUID” thing. Although for such a step, feedback from the MTA leadership would also be necessary.
However, I would definitely separate the configs and the resource cache for each one to avoid any issues.

Client/core/CVersionUpdater.cpp
```diff
@@ -564,6 +564,18 @@ void CVersionUpdater::DoPulse()
 ///////////////////////////////////////////////////////////////
 void CVersionUpdater::InitiateUpdate(const SString& strType, const SString& strData, const SString& strHost)
 {
+    // Disable server-initiated updates for secondary client
+    if (CCore::GetSingleton().IsSecondaryClient())
+        return;
+    
+    // Don't allow update if CL2 is running
```
qaisjp, line 574: smart

Client/loader/Utils.cpp
```diff
@@ -321,6 +332,14 @@ std::vector<DWORD> GetGTAProcessList()
 ///////////////////////////////////////////////////////////////////////////
 bool IsGTARunning()
 {
+    // Don't report GTA as running if CL2 is active (to allow coexistence)
+    HANDLE hCL2Mutex = OpenMutexA(SYNCHRONIZE, FALSE, MTA_GUID_CL2);
+    if (hCL2Mutex)
+    {
+        CloseHandle(hCL2Mutex);
+        return false;
+    }
```
qaisjp: Should abstract this to a common `IsSecondaryClientRunning`

Client/mods/deathmatch/logic/CServerIdManager.cpp
```diff
@@ -234,7 +246,13 @@ DWORD WINAPI CServerIdManagerImpl::StaticThreadProc(LPVOID lpdwThreadParam)
 ///////////////////////////////////////////////////////////////
 void CServerIdManagerImpl::StaticSaveServerIdMap()
 {
-    CXMLFile* pConfigFile = g_pCore->GetXML()->CreateXML(PathJoin(g_pClientGame->GetFileCacheRoot(), MTA_SERVERID_LOOKUP_XML));
```
qaisjp, line 237: Non-blocking, we can address this in a future PR -- what if we just change the `GetFileCacheRoot` to return a different directory for cl2?

Client/loader/Utils.cpp
```diff
@@ -1245,6 +1291,9 @@ bool CreateSingleInstanceMutex()
 ///////////////////////////////////////////////////////////////////////////
 void ReleaseSingleInstanceMutex()
 {
+    if (IsSecondaryClient())
+        return;
+
```
qaisjp: Since `CreateSingleInstanceMutex` holds a different mutex based on cl1 vs. cl2, shouldn't we delete this code?

Xenius97: We could delete i think

Client/core/CVersionUpdater.cpp
```diff
@@ -564,6 +564,20 @@ void CVersionUpdater::DoPulse()
 ///////////////////////////////////////////////////////////////
 void CVersionUpdater::InitiateUpdate(const SString& strType, const SString& strData, const SString& strHost)
 {
+    // Disable server-initiated updates for secondary client
+    if (CCore::GetSingleton().IsSecondaryClient())
+        return;
+
+#ifdef MTA_CL2
```
Xenius97, line 573: i think we can remove these checks too in CVersionUpdater since is blocked already with `CCore::GetSingleton().IsSecondaryClient`
@qaisjp

Xenius97: or we need this? 
```cpp
#ifdef MTA_CL2
    // Don't allow update if CL2 is running
    HANDLE hCL2Mutex = OpenMutexA(SYNCHRONIZE, FALSE, "Global\\{4962AF5F-5D82-412D-9CCA-AB8BB9DBD354}");
    if (hCL2Mutex)
    {
        CloseHandle(hCL2Mutex);
        return;
    }
#endif
```

qaisjp: I think we need to check both IsSecondaryClient and IsSecondaryClientRunning

Xenius97: Okay, but here is a small typo, `AB8BB9DBD354` and you changed in Utils to `AB8BB9DBD352` as i see

qaisjp: ah good point

guibzo:
Any updates on this?

The teams I’ve worked with and I have been using a custom build that includes the changes from this PR pretty much ever since it was opened, and honestly, it has made a huge difference to our productivity, especially when testing and developing resources.

I now need to switch back to the standard MTA build, and I really miss having this feature available. It would be great to see it included in the official MTA client.

MegadreamsBE:
> Any updates on this?
> 
> The teams I’ve worked with and I have been using a custom build that includes the changes from this PR pretty much ever since it was opened, and honestly, it has made a huge difference to our productivity, especially when testing and developing resources.
> 
> I now need to switch back to the standard MTA build, and I really miss having this feature available. It would be great to see it included in the official MTA client.

We are currently in a feature freeze due to MTA 1.7. Maybe 1.7.1, or later, may include this. No promises of course.

PR #4576, opened 2025-11-24
Fix `isPedDead` didn't check network dead state for players

### LINKED ISSUE #4147: isPedDead client-side bug
### Describe the bug

Calling this function with _onClientPlayerWasted_ will return false (ped is not dead) if you die by falling (**client-side**), However on **server-side**, calling it with _onPlayerWasted_ will return true (ped is dead) if you also die for the same reason.

This bug happens only if you die by falling as far as I know but, if you delay the call to _isPedDead_ function when _onClientPlayerWasted_, it returns true (player is dead)

It happens even for the localPlayer who is always considered as a streamed-in GTA SA entity.

### Steps to reproduce

```lua
-- client (no call delay)
addEventHandler("onClientPlayerWasted", root, 
function()
      -- player dies due to falling (localPlayer included)
      iprint(isPedDead(source))
      -- returns false (ped not dead but he triggered the onClientPlayerWasted event?)
end)


-- client (with call delay)
addEventHandler("onClientPlayerWasted", root, 
function()
      -- player dies due to falling (localPlayer included)
      local player = source
      setTimer(
           function() iprint(isPedDead(player)) end
      100, 1)
      -- sometines 50ms of delay still causes it to out put false
      -- returns true (ped is dead)
end)


-- server
addEventHandler("onPlayerWasted", root, 
function()
      -- player dies due to falling.
      iprint(isPedDead(source))
      -- returns true (ped is dead)
end)
```

### Version

Latest version (not nightly)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

Xenius97:
Fixes #4147

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp
@@ -1804,6 +1804,13 @@ int CLuaPedDefs::IsPedDead(lua_State* luaVM)
     {
         // Grab his dead state and return it
         bool bDead = pPed->IsDead() || pPed->IsDying();
+
+        // Check player is already dead on network (#4147)
+        if (auto pPlayer = dynamic_cast<CClientPlayer*>(pPed))
+        {
+            bDead = bDead || pPlayer->IsDeadOnNetwork();
+        }
+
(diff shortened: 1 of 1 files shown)


ghost (approved):
Nice

FileEX (approved):

PR #4582, opened 2025-11-28
Add new `getElementsSyncedByPlayer`

### LINKED ISSUE #2400: Add getPlayerSyncerElements
It's not really a problem. It would be just nice to have.

A function like **getPlayerSyncerElements** that returns a table of elements (vehicles and peds) that the player is currently sycning.

When I set a player as a syncer of somthing I store that infomation and try to keep it up to date as much as possible.

> If anybody is wanting something like we are talking about here you go:
> 
> ```
> function getElementsSyncedByPlayer( player,  theType )
>     local elementsByType = getElementsByType( theType )
>     local elementsSyncedByPlayer = {}
>     for i, v in pairs(elementsByType) do
>         local syncer = getElementSyncer( v )
>         if syncer == player then
>             table.insert( elementsSyncedByPlayer, v )
>         end
>     end
>     return elementsSyncedByPlayer
> end
> ```
> 
> This is not a nice way of doing it but it's good enough for now I guess

Xenius97:
Closes #2400

This PR adds new `getElementsSyncedByPlayer` which returns element table synced by player.

Shared:
```lua
table getElementsSyncedByPlayer()
```


tederis (dismissed):
LGTM

FileEX (dismissed):

qaisjp (changes requested):
Thanks for the pull request!

### `isElementSyncer` feedback
 
- I have concerns around the argument order of `isElementSyncer` server-side being confusing. Should the element be first, or the player be first? Currently in your pull request, the player is first.
- What's the benefit of `isElementSyncer` server-side when you can already do `if element.syncer == player`? This is more concise and doesn't introduce confusion around the argument order.

I'd prefer we remove the `isElementSyncer` portion of this pull request.

### `getElementsSyncedByPlayer` feedback

The filter feels strange to me. Do we have other examples of filters being parsed as comma-separated strings?

The linked issue didn't really explain why they want to filter by type, so for now I'd suggest just returning all the elements. We can always add filtering in a future pull request.

Xenius97:
Thanks for pointing this out!
I’ve removed the isElementSyncer part as suggested, and also removed the filter from getElementsSyncedByPlayer.

qaisjp (dismissed):
Thanks, this looks good. Btw it would be good if you merged the CStaticFunctionDefinitions directly into the luadefs.cpp file. (And removed the assertion.)

The CStaticFunctionDefinitions exists mainly for historical reasons and is not necessary for new functions

Xenius97:
> The CStaticFunctionDefinitions exists mainly for historical reasons and is not necessary for new functions

done

PR #4583, opened 2025-11-28
Fixed Polish translations

SOCZEKDEV:
Added translation for Post FX and shortened translations for some options that didn’t fit in the window.


Fernando-A-Rocha:
@SOCZEKDEV I think you need to use translate.multitheftauto.com (Crowdin platform) for translations, instead of opening a PR manually.

Xenius97:
Translations are managed on crowdin, needs to be changed there.

PR #4587, opened 2025-11-29
Add PostFX Lua functions

omar-o22:
Add PostFX functions
```lua
float GetPostFXValue ( int type )
    Types:
        0: Gamma
        1: Brightness
        2: Contrast
        3: Saturation
     Returns:
         The current value of the selected type
```
```lua
int GetPostFXMode ( )
    Returns: 
        0: Disabled
        1: Enabled in fullscreen mode
        2: Enabled in windowed/borderless mode
```
```lua
bool IsPostFXEnabled ( int type )
    Types:
        0: Gamma
        1: Brightness
        2: Contrast
        3: Saturation
     Returns:
        True if the selected type is enabled, otherwise false
```
Do you think adding `SetPostFXMode`/`SetPostFXValue` is needed, or not really?

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaFunctionParseHelpers.cpp
@@ -988,6 +988,13 @@ ADD_ENUM(VehicleAudioSettingProperty::VEHICLE_TYPE_FOR_AUDIO, "vehicle-type-for-
 IMPLEMENT_ENUM_CLASS_END("vehicle-audio-setting")
 
 
+IMPLEMENT_ENUM_CLASS_BEGIN(PostFXType)
+ADD_ENUM(PostFXType::GAMMA, "gamma")
+ADD_ENUM(PostFXType::BRIGHTNESS, "brightness")
+ADD_ENUM(PostFXType::CONTRAST, "contrast")
+ADD_ENUM(PostFXType::SATURATION, "saturation")
+IMPLEMENT_ENUM_CLASS_END("postfx-type")
+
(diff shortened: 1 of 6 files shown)


Fernando-A-Rocha:
Nice! For the type of FX I would prefer a string instead of an integer because it's more friendly for the developer.

Setter functions make sense as long as it is reset to client defaults when they disconnect from the server (AKA it doesn't override the esc settings - but it could display "The server you are currently on is modifying these values" or something)

Kinimel:
> Nice! For the type of FX I would prefer a string instead of an integer because it's more friendly for the developer.
> 
> Setter functions make sense as long as it is reset to client defaults when they disconnect from the server (AKA it doesn't override the esc settings - but it could display "The server you are currently on is modifying these values" or something)

Your point of view is quite interesting, I liked it.

omar-o22:
About your idea
> Setter functions make sense as long as it is reset to client defaults when they disconnect from the server (AKA it doesn't override the esc settings - but it could display "The server you are currently on is modifying these values" or something)

I'm not sure how to implement this yet, as I'm a beginner C++ developer so feel free to create an issue for it.

Fernando-A-Rocha:
> About your idea
> 
> > Setter functions make sense as long as it is reset to client defaults when they disconnect from the server (AKA it doesn't override the esc settings - but it could display "The server you are currently on is modifying these values" or something)
> 
> I'm not sure how to implement this yet, as I'm a beginner C++ developer so feel free to create an issue for it.

I don't know if there is demand for those features yet, so your PR is perfect with getter functions for now

tederis (approved):

PR #4592, opened 2025-12-03
Fix login censor and add register censor in chatbox history

C0DELEVEL:
This code fixes the unexpected behaviour of the [previously introduced](https://github.com/multitheftauto/mtasa-blue/pull/3725/files) `/login` censor in chatbox history, as the `/login` command was appearing in chatbox history even if it wasn't used, one of examples was to write `/` in chat.

\+ extend command censoring logic to also include the `/register` command, similar to `/login`.


Dutchman101:
How about detecting common typo's as well?

For example:
-login (-register)
\login (\register)
.login (.register)

(These aren't valid commands)

And also, it's out of scope, but ideally we can avoid such typo's from being printed to chatbox as well. It's a very common way that players get their server accounts compromised.

PR #4593, opened 2025-12-03
Advanced F11 map control

### LINKED ISSUE #4591: Change the map image serverside or stop updating the radar map on every update

Before [this](https://github.com/multitheftauto/mtasa-blue/pull/3823) pull request got merged, players could locally change the radar image to a custom one with server specific information to show. The locally changed now gets replaced after every MTA update.

Either don't update the radar map in the MTA installation folder after every update or implement a Lua function to change the map image.

A serverside reimplementation is not desired for a "feature" which could've been used in the past.

### Additional context

https://discordapp.com/channels/278474088903606273/278521065435824128/1445478223210086482

### Security Policy

Xenius97:
Closes #4591

This PR adds new functions to take full control of F11 map and adds ability to use custom radar textures (currently if you update MTA then it will overwrite everytime).
To use custom textures you need to place into new "cegui/radar" folder with `-custom` suffix.
This can be overwritten by lua scripts.

New functions:

---

```lua
bool setPlayerMapImage(string/texture, size = 1024/2048)
```
This will replace radar image based on size (compatible with new user resolution settings)

---

```lua
bool resetPlayerMapImage((optional) size = 1024/2048)
```
This will restore default radar image.

---

```lua
bool setPlayerMapOpacity(0-255)
```
This will set custom opacity for radar when opened.

```lua
bool resetPlayerMapOpacity()
```
This will reset custom opacity.

---

Test resource:
[radar.zip](https://github.com/user-attachments/files/23917852/radar.zip)


Fernando-A-Rocha:
Awesome!

Just a problem tho: F11 map is purposely called Player Map on MTA to distinguish from the minimap (Radar),  do the disable function should be disablePlayerMap ?

Fernando-A-Rocha:
Also, bool resetPlayerMapImage should not take any parameters in my opinion, to restore to the map size defined by the client in their settings automatically

Xenius97:
> Awesome!
> 
> Just a problem tho: F11 map is purposely called Player Map on MTA to distinguish from the minimap (Radar), do the disable function should be disablePlayerMap ?

i've renamed to this



> Also, bool resetPlayerMapImage should not take any parameters in my opinion, to restore to the map size defined by the client in their settings automatically

if you don't add a parameter for it, it will reset both, argument for this are optional

Fernando-A-Rocha:
To discuss: what about replacing the blip images? I believe for the F11 map they are .png but for the radar minimal they are textures part of HUD.txd that you can replace with a shader...?

Xenius97:
> To discuss: what about replacing the blip images? I believe for the F11 map they are .png but for the radar minimal they are textures part of HUD.txd that you can replace with a shader...?

F11 blips are managed here too, so i think we can make a function for replacing those. I think default radar images can be replaced via shader.

Fernando-A-Rocha:
> > To discuss: what about replacing the blip images? I believe for the F11 map they are .png but for the radar minimal they are textures part of HUD.txd that you can replace with a shader...?
> 
> F11 blips are managed here too, so i think we can make a function for replacing those. I think default radar images can be replaced via shader.

To me it's weird that F11 blips represent the blips with separate PNG images. These MTA blips are created using createBlip function, drawn using GTASA textures in the minimap, so I don't think it has been a good choice to make separate images for them since they are already in the game files (hud.txd).

Could you make it so the F11 blips use the GTASA textures so they can only be replaced using shaders, affecting F11 map and minimap at the same time?
I don't see why anyone would want separate blip images does the same blip in minimap and full map... But correct me if I'm wrong.

FileEX:
> This will completely disable F11 radar map, prevoriusly this was only possible with onClientKey workaround.

```lua
toggleControl('radar', false)
```

Client/mods/deathmatch/logic/CPlayerMap.cpp
```diff
@@ -752,3 +807,209 @@ SString CPlayerMap::GetBoundKeyName(const SString& strCommand)
         return strCommand;
     return pCommandBind->boundKey->szKey;
 }
+bool CPlayerMap::SetCustomMapImage(const std::string& strTexturePath, uint uiSize, CResource* pResource)
+{
+    if (uiSize != 1024 && uiSize != 2048)
```
tederis: How about using a namespace instead of uint?
`
enum class ECustomMapResolution
{
    Res_1024,
    Res_2048
}
`
Because `uint uiSize` argument takes just two possible values. This is a sign that something is wrong with the current implementation.

tederis (changes requested):

Xenius97:
> > This will completely disable F11 radar map, prevoriusly this was only possible with onClientKey workaround.
> 
> ```lua
> toggleControl('radar', false)
> ```

omg ty, i'm really blind, i've removed `disablePlayerMap`

Fernando-A-Rocha:
> > > This will completely disable F11 radar map, prevoriusly this was only possible with onClientKey workaround.
> > 
> > 
> > ```lua
> > toggleControl('radar', false)
> > ```
> 
> omg ty, i'm really blind, i've removed `disablePlayerMap`

I know it's just naming and it doesn't affect anything, but the F11 map (player-map) strings/commands really should not be called "radar" because that refers to the minimap 😔 

image

denjifb:
Is there any progress?

Xenius97:
> Is there any progress?

Last time i've checked this worked, and blips could be added in future PR if this gets merged.

Xenius97:
When new texture-related fixes are done this needs to be checked before merge, as i see there are a few changes in master branch.

PR #4600, opened 2026-01-05
Add ability to make client-side vehicles damageable

MegadreamsBE:
This PR makes it possible to mark a client-side vehicle to be damageable. By default client-side vehicles cannot receive damage, this PR does not change that default behavior.

When false is passed to **SetVehicleDamageProof** for a local vehicle, it will now become damageable.

PREVIOUS: https://github.com/multitheftauto/mtasa-blue/pull/4411


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1176,6 +1176,9 @@ void CClientVehicle::CalcAndUpdateCanBeDamagedFlag()
     if (m_bSyncUnoccupiedDamage)
         bCanBeDamaged = true;
 
+    if (IsLocalEntity())
+        bCanBeDamaged = m_bScriptCanBeDamaged;
+
     // Script override
     if (!m_bScriptCanBeDamaged)
         bCanBeDamaged = false;
```
qaisjp, line 1185: I think it'd be clearer to set `bCanBeDamaged = true;` here, and let the script override on L1183/1184 remain as the override. That way we don't have to think too much about how the multiple variables interact.


```suggestion
    if (IsLocalEntity())
        bCanBeDamaged = true;

    // Script override
    if (!m_bScriptCanBeDamaged)
        bCanBeDamaged = false;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1218,6 +1221,9 @@ void CClientVehicle::CalcAndUpdateTyresCanBurstFlag()
     if (m_bSyncUnoccupiedDamage)
         bTyresCanBurst = true;
 
+    if (IsLocalEntity())
+        bTyresCanBurst = m_bScriptCanBeDamaged;
```
qaisjp:
```suggestion
        bTyresCanBurst = true;
```

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -135,7 +135,7 @@ CClientVehicle::CClientVehicle(CClientManager* pManager, ElementID ID, unsigned
     memset(&m_ucLightStates[0], 0, sizeof(m_ucLightStates));
     m_bCanBeDamaged = true;
     m_bSyncUnoccupiedDamage = false;
-    m_bScriptCanBeDamaged = true;
+    m_bScriptCanBeDamaged = !IsLocalEntity();
```
qaisjp, line 139: Can we add a comment to this line to the effect of

```cpp
// Local vehicles are damage proof by default due to backwards compatibility. Before 1.7 it wasn't possible for client side vehicles to be damaged.
```

qaisjp (changes requested):

qaisjp (approved):
Thanks! Sorry for the delay with the review. Gonna sort out my GitHub notifications...

PR #4605, opened 2026-01-06
Update copyright year

omar-o22:
I've created an old pr for it and added unrelated commits, so I closed the old one


FileEX (approved):

Xenius97:
Fact: According to the wiki, MTA:SA Race will turn 20 years old on this January 22.

PR #4609, opened 2026-01-07
train-tracks: fix track IDs shifting when you delete a default track

qaisjp:
Fix an issue where if you use `destroyElement(getDefaultTrack(2))`, calling `getDefaultTrack(2)` again would return what should have been track 3.

This is because `m_Tracks` is a `std::vector`, and we shift the entire vector when we destroy a train track. This is resolved by storing the default tracks in a separate `m_DefaultTracks` array and _not_ doing any shifting.

I've also merged `Reset()` into the constructor, since `Reset` isn't safe to be called on demand. (It simply empties the list without deleting any objects, which is a memory leak.)
 
### Backwards compatibility

This change affects only unreleased Lua APIs.

#### Before

```lua
t2 = getDefaultTrack(2) -- userdata(0x002)
t3 = getDefaultTrack(3) -- userdata(0x003)
destroyElement(t2)
getDefaultTrack(2) == t3 -- true         !!!! incorrect, expected false
getDefaultTrack(2) -- userdata(0x003)    !!!! incorrect, expected nil
getDefaultTrack(3) -- nil                !!!! incorrect, expected 0x003
```

#### After

```lua
t2 = getDefaultTrack(2) -- userdata(0x002)
t3 = getDefaultTrack(3) -- userdata(0x003)
destroyElement(t2)
getDefaultTrack(2) -- nil
getDefaultTrack(3) -- userdata(0x003) 
```

i.e. getDefaultTrack(2):destroy() followed by getDefaultTrack(2) returns the original track 3.

### DIFF
--- a/Server/mods/deathmatch/logic/CTrainTrack.cpp
+++ b/Server/mods/deathmatch/logic/CTrainTrack.cpp
@@ -19,8 +19,8 @@ CTrainTrack::CTrainTrack(CTrainTrackManager* pManager, const std::vector<STrackN
     m_iType = CElement::TRAIN_TRACK;
     SetTypeName("train-track");
 
-    m_LinkLastNodes = linkLastNodes;
     m_Nodes = nodes;
+    m_LinkLastNodes = linkLastNodes;
     m_DefaultTrackId = defaultTrackId;
 }
 
--- a/Server/mods/deathmatch/logic/CTrainTrack.h
+++ b/Server/mods/deathmatch/logic/CTrainTrack.h
@@ -57,6 +57,5 @@ private:
 
     std::vector<STrackNode> m_Nodes;
     bool                    m_LinkLastNodes;
-    bool                    m_Default;
     uchar                   m_DefaultTrackId;
 };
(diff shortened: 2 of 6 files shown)


Server/mods/deathmatch/logic/CTrainTrack.h
```diff
@@ -57,6 +57,5 @@ class CTrainTrack final : public CElement
 
     std::vector<STrackNode> m_Nodes;
     bool                    m_LinkLastNodes;
-    bool                    m_Default;
```
qaisjp, line 60: Linters complained that this wasn't being used, so I've removed this as well.

Server/mods/deathmatch/logic/CTrainTrack.cpp
```diff
@@ -19,8 +19,8 @@ CTrainTrack::CTrainTrack(CTrainTrackManager* pManager, const std::vector<STrackN
     m_iType = CElement::TRAIN_TRACK;
     SetTypeName("train-track");
 
-    m_LinkLastNodes = linkLastNodes;
     m_Nodes = nodes;
+    m_LinkLastNodes = linkLastNodes;
     m_DefaultTrackId = defaultTrackId;
```
qaisjp, line 24: Reordered for consistency with headers

FileEX (approved):

Issue #2507, opened 2022-01-09
Setup automatic clang format/tidy for pull requests


patrikjuvonen:
**Is your feature request related to a problem? Please describe.**
Pull requests are not often run through the clang formatter which causes inconsistent code formatting to poison the codebase, causing [lots of manual formatting](https://github.com/multitheftauto/mtasa-blue/search?q=clang&type=commits) work to take place.

**Describe the solution you'd like**
Create a GitHub action that automatically formats pull request code and commits the formatted code into the PR.

**Describe alternatives you've considered**
1. Doing it periodically through a GitHub action on master and generating a pull request out of it. **This is still probably good to implement anyway on top of all of this.**
2. Continuing to do it manually. **Probably need to do this sometimes anyway.**
3. Fail the action if formatting has issues. Aka linting and not fixing anything automatically, leaving the cleanup to the dev.
4. A local Git hook that will automatically run formatting on pre-commit/push and commit the formatting fixes. This is however dependent on the user installing and using the Git hooks, so probably not good enough on its own. **Good as an addition on top of the action.**

**Additional context**
- See #772 - clang-tidy?
- mtasa-resources is battling with the same issue at https://github.com/multitheftauto/mtasa-resources/issues/365

Disinterpreter:
> Create a GitHub action that automatically formats pull request code and commits the formatted code into the PR.

IMHO It's a bad idea. It's better to fail an action when the code on PR does not match our style guide.

patrikjuvonen:
> IMHO It's a bad idea. It's better to fail an action when the code on PR does not match our style guide.

I've added that to the list of alternatives.

Keep in mind that formatting **by default** takes place globally, not just in the files that you have changed.

Disinterpreter:
> > IMHO It's a bad idea. It's better to fail an action when the code on PR does not match our style guide.
> 
> I've added that to the list of alternatives.
> 
> Keep in mind that formatting by default takes place globally, not just in the files that you have changed.

it's just a typical linter/static analysator. we experiment with it in https://github.com/multitheftauto/mtasa-resources/issues/365 (lua) for making uniform style guide.

You have to make good settings for it, and it can be implemented.

AlexTMjugador:
[...]

PR #4611, opened 2026-01-07
codemod: disable clang-format for `__asm` directives

qaisjp:
Use [fastmod](https://github.com/facebookincubator/fastmod) to wrap `// clang-format off/on` around all our `__asm` directives:

```bash
$ fastmod '^( +?)(__asm$\n(?:\s+)\{$[^}]+})' '${1}// clang-format off
$1$2
${1}// clang-format on' --accept-all --print-changed-files | wc -l
102
```

I'm confident in this regex because VS Code showed it's selecting the right blocks of code:

{FF157373-DA22-4BA0-BBEC-97883C25F7B5}

Additionally, fastmod says it changed 1021 instances, and VS Code reported the same number when I plugged in the regex.

Searching for "__asm" reports 1030 instances. The remaining 9 comes from inline `__asm {` (i.e. the open brace is on the same line), which I've manually changed.

This pull request essentially makes this change from the original clang-format script: https://github.com/multitheftauto/mtasa-blue/blob/81677d9e1a3632bd5e03b660c0bb37fe4234c7f8/utils/win-apply-clang-format.bat#L59-L60

**Motivation**

Unblocks https://github.com/multitheftauto/mtasa-blue/issues/2507 and possibly unblocks https://github.com/multitheftauto/mtasa-blue/issues/772.

Essentially, we can't currently enforce clang-tidy across our codebase because it doesn't work correctly with `__asm`. This addresses that problem by disabling clang-tidy for all `__asm`  directives.

**Does this make the codebase ugly?**

Maybe, but this is standard operating procedure for large codebases. This pull request allows people to enable format-on-save in their editors without having to manually undo uglification to `__asm` directives. I think it's an acceptable tradeoff.

**Test plan**

All changes are to comments, so they are safe. `fastmod`'s manual mode let me verify that this worked correctly.

### DIFF
--- a/Client/game_sa/CAnimBlendAssociationSA.cpp
+++ b/Client/game_sa/CAnimBlendAssociationSA.cpp
@@ -32,25 +32,29 @@ CAnimBlendAssociationSAInterface* CAnimBlendAssociationSA::Constructor(CAnimBlen
 {
     DWORD DwFunc = 0x4CF080;
     DWORD DwThisInterface = reinterpret_cast<DWORD>(m_pInterface);
+    // clang-format off
     __asm
     {
         mov     ecx, DwThisInterface
         push    staticAssociationByReference
         call    DwFunc
-    };
+    }
+    // clang-format on;
 }
 
 CAnimBlendAssociationSAInterface* CAnimBlendAssociationSA::Constructor(RpClump* pClump, CAnimBlendHierarchySAInterface* pAnimHierarchy)
 {
     DWORD DwFunc = 0x4CEFC0;
     DWORD DwThisInterface = reinterpret_cast<DWORD>(m_pInterface);
+    // clang-format off
     __asm
     {
         mov     ecx, DwThisInterface
         push    pAnimHierarchy
         push    pClump
         call    DwFunc
-    };
+    }
+    // clang-format on;
 }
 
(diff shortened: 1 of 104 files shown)


Client/game_sa/CAnimBlendAssociationSA.cpp
```diff
@@ -32,25 +32,29 @@ CAnimBlendAssociationSAInterface* CAnimBlendAssociationSA::Constructor(CAnimBlen
...
+    // clang-format off
     __asm
     {
         mov     ecx, DwThisInterface
         push    pAnimHierarchy
         push    pClump
         call    DwFunc
-    };
+    }
+    // clang-format on;
```
Lpsd, line 57: Does this slight difference matter at all, compared to the rest where there's no semi-colon? (surely it's fine, just double checking!)

qaisjp: Oh, nice catch! 

Looking at https://learn.microsoft.com/en-us/cpp/assembler/inline/asm?view=msvc-170 it says the semicolon is optional:

```
__asm { assembly-instruction-list } ;opt
```

And [`opt` is described as](https://learn.microsoft.com/en-us/cpp/preprocessor/grammar-summary-c-cpp?view=msvc-170#conventions) "Nonterminals followed by opt are always optional."

Lpsd (approved):

qaisjp:
Thank you @lpsd!

Issue #2869, opened 2023-01-12
Running out of memory (and crash) by loading content on demand


RatajVaver:
### Describe the bug

This bug has been already documented in this issue: #966. Unfortunately author of that issue marked it as completed (because he found a workaround), although the bug was never fixed.

This bug is persisting for years and happens for every player on some servers that have this issue after longer sessions of playtime (i.e. 3 hours, but it differs mainly server to server and also client to client). I'm convinced this is not an issue for most servers but only a few, because otherwise I suppose it would have been solved a long time ago. However it's not a small issue, because for those few servers every single player experiences these crashes (besides for a few very unusual cases).

The issue is, when you use `dxCreateTexture` and similar functions very often and you're also unloading these textures properly later on (to avoid memory leaks), there are probably some leftovers in the memory that in the end cause memory leak by re-creating new and new textures, sounds, importing models, etc.

Only workaround (althought not very viable) is doing a memory leak intentionally (by not unloading textures and keeping them in memory even when they're not in use), which helps some players (crash happens later on), but also harms other players (crashes even more often).

There might be a memory leak in my server, althought I doubt it, since I was trying to find it for a very long time and I wasn't able to replicate this bug on my dev server with same scripts but different data (there just wasn't 100 players driving around with textured vehicles, as it is on my production server).

I've tried to find the problem in both showmemstat and performancebrowser, all with no success. Therefore I believe this is bug in MTA and how it handles freed memory. Maybe it's something with memory fragmentation? I have no idea, I don't know much about that. There's probably someone who can track this down and explain it better than me.

### Steps to reproduce

Might be hard to reproduce, because it usually happens only on servers with a lot of custom player content. For example on RolePlay servers with a lot of custom interior textures, vehicle textures (remap), player radio stations, ped models. These things are loaded only when needed (e.g. vehicle streams in -> shader+texture is created, vehicle streams out -> shader+texture is deleted). When you do this a lot of times, you get to a point where debug outputs "not enough memory" and no new textures are being loaded, no new sounds from URL are being played, no new models are imported, even screenshots can't be taken. Then after few minutes there's the crash (well documented in #966).

It might however be reproducable by creating lots of textures, unloading them and loading them again for a while.

I've did some testing year ago, but I wasn't able to reproduce it on dev server at all. Maybe someone can. I'm creating this issue mainly to get some attention to it, since I can't find out enough information that could lead to solution.

### Version

Every version to date (including nightly).

### Security Policy

[...]

PR #4612, opened 2026-01-07
Fix unbounded virtual memory growth by switching textures to D3DPOOL_…

### LINKED ISSUE #4554: Textures Memory Leak (Virtual)
### Describe the bug

Memory leak that causes crashes on servers that use many textures, even if it destroys the elements and manages resources correctly

### Steps to reproduce

1 - Download the resource below
2 - Start the resource
3 - Use the “ctextures” command a few times
4 - Check the virtual memory usage (located in the lower left corner of the screen)
5 - Use the “dtextures” command and check virtual memory usage again
6 - You can repeat the process as many times as you want; it seems to be related to the number of textures already created.
7 - If you get close to 4096 MB, start walking around the map, do anything that increases RAM usage a little more, and that's it, you'll crash
8 - This behavior does not occur when using “dxCreateTexture(Width, Height)”.

### Version

Client: Multi Theft Auto v1.6-release-23586

### Additional context

Video of the incident:
https://www.youtube.com/watch?v=nAyrUDaesdQ

[TexturesMemoryLeak.zip](https://github.com/user-attachments/files/23609688/TexturesMemoryLeak.zip)

Edit: If you create textures with your RGBA pixels, there will be no memory leak.

### Relevant log output

```shell

```

### Security Policy

ZeusHay:
Fixed a critical issue where loading textures, caused unbounded "virtual Memory" growth, leading to crashes.

I just replaced the D3DPOOL_MANAGED with D3DPOOL_DEFAULT for file-based textures, to avoid d3dx system memory retention.

I implemented OnLostDevice and OnResetDevice to handle texture restoration (when user use full screen) for ALT TAB Support.

I added forced texture unbinding too in ReleaseUnderlyingData to ensure the immediate VRAM cleanup on destruction

I updated file reading to use VirtualAlloc, verifying full memory release to the os.

Just a note:

if user load like, 30K textures (its a impratical situation, just example), it will took the time for reloading this textures, it's very lightweight and highly optimized.)

This Fixes: https://github.com/multitheftauto/mtasa-blue/issues/4554 & https://github.com/multitheftauto/mtasa-blue/issues/2869

### DIFF
--- a/Client/core/Graphics/CRenderItem.FileTexture.cpp
+++ b/Client/core/Graphics/CRenderItem.FileTexture.cpp
@@ -25,13 +25,34 @@ void CFileTextureItem::PostConstruct(CRenderItemManager* pManager, const SString
     m_TextureType = textureType;
     m_TextureAddress = textureAddress;
 
+    // @ZeusHay - Store for recreation
+    m_bMipMaps = bMipMaps;
+    m_uiCreateSizeX = uiSizeX;
+    m_uiCreateSizeY = uiSizeY;
+    m_CreateFormat = format;
+
     // Initial creation of d3d data
     if (pPixels)
+    {
+        m_bIsFromFile = false;
+        m_bIsBlank = false;
+        m_PixelData.SetSize(pPixels->GetSize());
+        memcpy(m_PixelData.GetData(), pPixels->GetData(), pPixels->GetSize());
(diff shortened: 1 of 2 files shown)


Dutchman101:
Closed because of factors to do with the author.
We will independently assess if this (the issue described) is really a bottleneck. and if so, will come up with a redesigned fix soon.

PR #4633, opened 2026-01-13
Move Lpsd to programming section

qaisjp:
#### Motivation

@Lpsd is now a member of the MTA team 🎉


Client/core/CCredits.cpp
qaisjp, line 101: Separate PR -- someone should update these with the active contributors to the default resources

Lpsd (approved):

PR #4641, opened 2026-01-18
Move maetro patches to master behind MTA_MAETRO env var

qaisjp:
#### Summary

Instead of maintaining two branches that causes problems when they diverge, we incorporate maetro directly into the master branch behind the `MTA_MAETRO` environment variable.

Try it in powershell using:
```
$env:MTA_MAETRO="true"
.\win-create-projects.bat
```

#### Motivation

Reduces maintenance burden

#### Test plan

created projects as above, old version of CEF was downloaded, confirmed that it shows up in preprocessor defs now:

image

builds on my machine ✔️ 

#### Build server

Added an env-based parameter to the maetro projects:

{5FB2BEF7-E455-4040-8F86-B93416301528}

### DIFF
--- a/.github/workflows/sync-master-to-maetro.yaml
+++ b/.github/workflows/sync-master-to-maetro.yaml
@@ -24,14 +24,10 @@ jobs:
 
           git checkout release/maetro
           git merge master --no-ff --no-edit --no-commit || true
-          
+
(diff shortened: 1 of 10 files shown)


qaisjp:
Whoops, I forgot that auto-merge ignores code review

PR #4644, opened 2026-01-18
Store maetro d3dcompiler_47.dll in master branch too

qaisjp:
#### Summary

Not ideal storing binary files in git, but it's already there.

This pull requests stores the dll currently in the `release/maetro` branch directly in `master` at `d3dcompiler_47.maetro.dll`.

#### Motivation

Eliminate the `release/maetro` branch

#### Test plan

I'll merge this and generate an installer on the build server. Then run Windows 7 using https://archive.org/details/Windows7-iso. Then run the maetro installer in the VM.

Edit: I installed maetro in a VM and confirmed the hash of d3dcompiler_47.dll is what I expected (`C2C7CA6505AD10826E6B92319CE7AA355392B0CBD092A0FB8D4381C2D31268BF`, sha256)


ghost:
Why not host it on MTA's CDN instead?

qaisjp:
@AlexTMjugador replied with a good explanation on Discord:

> The harm has been done already, and it's probably more reliable in the event the CDN goes down during build. Also no code required to download it before installer build if the file is already checked out

I think it would be a good idea to store these files in a GitHub release on a separate repo, so that we don't have to self-host. But I didn't want to expand the scope of this change.

I don't feel bad about storing this particular DLL in git because it was already in the `release/maetro` branch; so it's already there in the git db

(Can always keep a backup elsewhere in case something happens to GitHub.)

PR #4651, opened 2026-01-19
minor organisational tweaks to CBulletsyncPacket

qaisjp:
#### Summary (I recommend reading commit by commit)

* Prefer member initializer over `ctor-initializer` (we should have one, not both)
* Instead of calling `ResetDamageData` before every `return false` in `ReadOptionalDamage`, just have the caller do it when `ReadOptionalDamage` returns false
* remove dead code

#### Motivation

We were talking about packet architecture in the dev Discord and noticed that there was some dead code, so I thought I'd fix it. While I was here, I noticed other minor improvements that could be made.

#### Test plan

CI

### DIFF
--- a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
@@ -189,6 +189,9 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
 
     if (!ReadOptionalDamage(stream))
     {
+        // todo: do we really need to reset damage data when we're returning
+        // false? returning false deletes the packet, and other packets don't
+        // reset internal data based on validation failures.
         ResetDamageData();
         return false;
     }
--- a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
@@ -141,6 +141,7 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
     if (!m_pSourceElement)
         return false;
 
+    // todo: is this condition always true?
     CPlayer* pPlayer = static_cast<CPlayer*>(m_pSourceElement);
     if (pPlayer)
     {
@@ -207,6 +208,7 @@ bool CBulletsyncPacket::Write(NetBitStreamInterface& stream) const
     if (id == INVALID_ELEMENT_ID)
         return false;
 
+    // why?
     if (id == 0)
         return false;
 
--- a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
@@ -141,14 +141,10 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
     if (!m_pSourceElement)
         return false;
 
-    // todo: is this condition always true?
-    CPlayer* pPlayer = static_cast<CPlayer*>(m_pSourceElement);
-    if (pPlayer)
-    {
-        // Check if player is spawned and alive
-        if (!pPlayer->IsSpawned() || pPlayer->IsDead())
-            return false;
-    }
+    // Check if player is spawned and alive
+    CPlayer* pPlayer = GetSourcePlayer();
+    if (!pPlayer->IsSpawned() || pPlayer->IsDead())
+        return false;
 
     if (!ReadWeaponAndPositions(stream))
         return false;
--- a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
@@ -18,7 +18,6 @@
 #include "CWeaponNames.h"
 
 CBulletsyncPacket::CBulletsyncPacket(CPlayer* player)
-    : m_weapon(WEAPONTYPE_UNARMED), m_start(), m_end(), m_order(0), m_damage(0.0f), m_zone(0), m_damaged(INVALID_ELEMENT_ID)
 {
     m_pSourceElement = player;
 }
--- a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
@@ -100,41 +100,28 @@ bool CBulletsyncPacket::ReadWeaponAndPositions(NetBitStreamInterface& stream)
     return true;
 }
 
+// Returns false when there's a validation error.
+// Make sure to reset damage data to defaults when that happens.
 bool CBulletsyncPacket::ReadOptionalDamage(NetBitStreamInterface& stream)
 {
     if (!stream.ReadBit())
-    {
-        ResetDamageData();
         return true;
-    }
 
     stream.Read(m_damage);
     stream.Read(m_zone);
     stream.Read(m_damaged);
 
     if (IsNaN(m_damage))
-    {
-        ResetDamageData();
         return false;
-    }
 
     if (m_damage < 0.0f || m_damage > MAX_DAMAGE)
-    {
-        ResetDamageData();
         return false;
-    }
 
     if (m_zone > MAX_BODY_ZONE)
-    {
-        ResetDamageData();
         return false;
-    }
 
     if (m_damaged == 0)
-    {
-        ResetDamageData();
         return false;
-    }
 
     // Check that target element exists (if specified)
     // Note: m_damaged can be INVALID_ELEMENT_ID when shooting at ground/world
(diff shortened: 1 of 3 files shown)


Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
```diff
@@ -205,7 +183,13 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
         return false;
 
     if (!ReadOptionalDamage(stream))
+    {
+        // todo: do we really need to reset damage data when we're returning
+        // false? returning false deletes the packet, and other packets don't
+        // reset internal data based on validation failures.
+        ResetDamageData();
```
qaisjp, line 186: I think we could get rid of `ResetDamageData` but not sure if it was added for a specific reason.

I asked AI to double check whether I am right and it agrees:

> Looking at the code, when Read() returns false, the packet is indeed dropped and won't be processed further.
>
> #### **1. Packet is deleted immediately after Read() returns false**
> 
> In [`CPacketTranslator.cpp` (lines 252-256)](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CPacketTranslator.cpp#L252-L256):
> ```cpp
> // Attempt to read the content, if we fail, delete the packet again
> if (!pTemp->Read(BitStream))
> {
>     delete pTemp;
>     pTemp = NULL;
> }
> ```
> 
> This confirms that when `Read()` returns false, the packet object is **immediately destroyed** and not used further.
> 
> #### **2. Other packets DON'T reset state on validation failure**
> 
> Checking similar packet implementations:
> - [`CKeysyncPacket.cpp` (lines 69-80)](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/packets/CKeysyncPacket.cpp#L69-L80): Multiple `return false` statements without any cleanup
> - [`CObjectSyncPacket.cpp` (lines 37-63)](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/packets/CObjectSyncPacket.cpp#L37-L63): Returns false on validation errors without resetting allocated data
> 
> **None of the other packets reset their member variables before returning false from validation.**
> 
> ### **Conclusion**
> 
> The `ResetDamageData()` call at [line 192](https://github.com/multitheftauto/mtasa-blue/blob/master/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp#L192) is **not necessary** because:
> 1. The packet is deleted immediately after `Read()` returns false
> 2. No other packet implementations reset state on validation failure
> 3. The pattern across the codebase is to simply return false and let the packet translator handle cleanup
> 
> You could safely remove the `ResetDamageData()` call to be consistent with other packet implementations.

Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
```diff
@@ -224,6 +208,7 @@ bool CBulletsyncPacket::Write(NetBitStreamInterface& stream) const
     if (id == INVALID_ELEMENT_ID)
         return false;
 
+    // why?
     if (id == 0)
         return false;
```
qaisjp, line 209: It shouldn't be possible for `Write` to be called when `m_pSourceElement` doesn't exist or is invalid - this would indicate server corruption further up the stack.

this check should probably be removed, or moved to an outer function instead of here

qaisjp: I asked the LLM `when would this return INVALID_ELEMENT_ID or 0?` and it seems to agree:

> 
> Based on my investigation, here's when `pPlayer->GetID()` would return these values:
> 
> ### `INVALID_ELEMENT_ID` (0xFFFFFFFF)
> 
> Returns when:
> - **The element ID stack is exhausted** - When [CElementIDs::PopUniqueID()](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CElementIDs.cpp#L23-L33) fails because there are no more available IDs in the stack (more than [MAX_SERVER_ELEMENTS - 2](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CElementIDs.h#L27) = 131,070 elements have been created)
> - The [CStack::Pop()](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Shared/sdk/SharedUtil.Misc.h#L901-L912) returns `false`, leaving `ID` with its default value of [INVALID_ELEMENT_ID](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Shared/sdk/Common.h#L26)
> 
> ### ID `0`
> 
> Returns when:
> - **ID 0 is actually assigned from the stack** - Looking at the [CStack implementation](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Shared/sdk/SharedUtil.Misc.h#L875-L923), IDs are added [starting from ulOldSize + 1](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Shared/sdk/SharedUtil.Misc.h#L894). With initial capacity expansion, the stack is populated with IDs from `1` to `INITIAL_MAX_STACK_SIZE`
> - However, ID `0` could theoretically be assigned if manually pushed back into the stack (e.g., via [PushUniqueID(0)](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CElementIDs.cpp#L37-L43)), though this would be unusual
> 
> The comment [// why?](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp#L210) next to the `id == 0` check is questioning why this check exists. Based on the code:
> 
> 1. Element IDs should normally start from 1 (based on the stack initialization logic)
> 2. ID 0 being reserved/invalid is likely a defensive check or legacy code
> 3. The [MAX_SERVER_ELEMENTS - 2](https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CElementIDs.h#L27) capacity suggests IDs 0 and `RESERVED_ELEMENT_ID` are intentionally excluded from the valid range
> 
> In practice, **these conditions should be extremely rare** in normal operation - they would only occur if:
> - The server has run out of element IDs (server created 131,070+ elements)
> - There's a bug causing ID 0 to be pushed back into the stack
> - Memory corruption or initialization failure

Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
```diff
@@ -158,18 +141,13 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
     if (!m_pSourceElement)
         return false;
 
+    // todo: is this condition always true?
     CPlayer* pPlayer = static_cast<CPlayer*>(m_pSourceElement);
     if (pPlayer)
     {
```
qaisjp: `CPacketTranslator.cpp` which sets up the source element like this:

https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CPacketTranslator.cpp#L230-L231

https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CPacketTranslator.cpp#L241-L242

And then calls `Read`:

https://github.com/multitheftauto/mtasa-blue/blob/954f7416d83f25683e3db9d18c687e9d7d4f243a/Server/mods/deathmatch/logic/CPacketTranslator.cpp#L251-L252

So `if (pPlayer)` should always be true.

Ideally we could make `CBulletsyncPacket` some sort of generic that holds the `CPlayer` type in `m_pSourceElement` but for now we could just remove the check

qaisjp: Ended up documenting this here https://github.com/multitheftauto/mtasa-blue/pull/4651/commits/1915561c559f6e628af5e2ed573b4bbb4a7792a1 and making the necessary change https://github.com/multitheftauto/mtasa-blue/pull/4651/commits/93fdb488e975051c2f0a26eb6b1ad4792c456803

Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
```diff
@@ -158,19 +141,10 @@ bool CBulletsyncPacket::Read(NetBitStreamInterface& stream)
     if (!m_pSourceElement)
         return false;
 
-    CPlayer* pPlayer = static_cast<CPlayer*>(m_pSourceElement);
-    if (pPlayer)
-    {
```
qaisjp, line 163: See https://github.com/multitheftauto/mtasa-blue/pull/4651#discussion_r2705803106 for more details on this code being deleted

PR #4657, opened 2026-01-20
Make server components optional

### LINKED ISSUE #4656: Allow server to be unticked in MTA's installer

Allow dedicated server component to be unticked in MTA's installer and skipped during auto-updates if the server folder isn't present in the MTA installation. It's forcibly ticked in the installer, but you can delete the server folder from your MTA installation and MTA continues to work just fine. Additionally, the Host Game and Map Editor buttons should be removed from the menu if the server folder isn't present - they don't crash the game, but they fail to create the MTA Server process because it doesn't exist so they're useless. I have a dedicated server folder on a different drive, I'd like to keep client and server separate.

Image

Allow dedicated server component to be unticked in MTA's installer and skipped during auto-updates if the server folder isn't present in the MTA installation. It's forcibly ticked in the installer, but you can delete the server folder from your MTA installation and MTA continues to work just fine. Additionally, the Host Game and Map Editor buttons should be removed from the menu if the server folder isn't present - they don't crash the game, but they fail to create the MTA Server process because it doesn't exist so they're useless. I have a dedicated server folder on a different drive, I'd like to keep client and server separate.

Image

_No response_

### Additional context

Kind of similar to #1309

### Security Policy

Xenius97:
#### Summary
This PR hides server-dependent menu options when the MTA server component is not present.
Closes #4656

#### Motivation
Users who want to keep client and server installations separate (for example, running a dedicated server on a different drive) may manually remove the server component.

Currently, even if the `server` folder is deleted, the **Host Game** and **Map Editor** buttons remain visible. Clicking them fails because the MTA Server process cannot be created.

This change ensures that server-dependent UI elements are hidden when the server component is missing, the auto-updater does not restore the deleted server files, and the server can be restored later by running the MTA Installer to update existing files or by manually recreating the `server` folder with the required files.

#### Test plan
Delete the `server` folder and start MTA.

#### Checklist
- [x] Your code should follow the [coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines).

### DIFF
--- a/Client/core/CMainMenu.cpp
+++ b/Client/core/CMainMenu.cpp
@@ -184,7 +184,7 @@ CMainMenu::CMainMenu(CGUI* pManager)
     m_menuItems.push_back(CreateItem(MENU_ITEM_BROWSE_SERVERS, "menu_browse_servers.png", CVector2D(0.168f, fBase + fGap * iMenuItemIndex++)));
 
     // Only add Host Game and Map Editor if server folder and MTA Server.exe exist
-    if (DirectoryExists(CalcMTASAPath("server")) && FileExists(CalcMTASAPath("server/MTA Server.exe")))
+    if (DirectoryExists(CalcMTASAPath("server")))
     {
         m_menuItems.push_back(CreateItem(MENU_ITEM_HOST_GAME, "menu_host_game.png", CVector2D(0.168f, fBase + fGap * iMenuItemIndex++)));
         m_menuItems.push_back(CreateItem(MENU_ITEM_MAP_EDITOR, "menu_map_editor.png", CVector2D(0.168f, fBase + fGap * iMenuItemIndex++)));
--- a/Client/core/CMainMenu.cpp
+++ b/Client/core/CMainMenu.cpp
@@ -179,13 +179,20 @@ CMainMenu::CMainMenu(CGUI* pManager)
     // Create the menu items
     // Filepath, Relative position, absolute native size
     // And the font for the graphics is ?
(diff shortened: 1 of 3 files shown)


blonter:
I have a question about this.

Great job overall!

However, considering that 99.99% of people who install MTA only intend to play the game, shouldn’t this option be unchecked by default? In other words, the installer shouldn’t install the server component unless the user explicitly chooses it.

Just add a note in the installer explaining that if someone wants to use the Map Editor or the Host Game feature, they need to enable the server components option.

It might also be worth considering adding the 64‑bit server as a separate, optional checkbox—just like the regular server component. So you don't have to manually download it from nightly page. This would also allow users who only want the 64‑bit version to install it without having to tick the 32‑bit one, and vice versa. (or installing both by ticking them) The only important thing is to ensure that all required files are included for each option, so the 64‑bit server doesn’t rely on anything installed by the 32‑bit one, and the other way around.

Xenius97:
> It might also be worth considering adding the 64‑bit server as a separate, optional checkbox

I think this isn’t possible, but maybe the pros have an idea.

Xenius97:
Update: installer changes are reverted, but if you delete `server` folder manually `Host game` and `Map Editor` buttons will be hidden and auto-updater won't restore it.
If you want to server install again you can run MTA Installer and update your existing files and server will come back. Or you can create the folder and put needed files into it.

Discord discussion: https://discord.com/channels/801330706252038164/801330706252038170/1463377381656891567

qaisjp (dismissed):
Sharing some conversation from Discord, I asked that we remove this setting from the installer because it also removes the map editor.

The reason for this is to avoid support questions around why the map editor is missing, either because the user didn't realise the server was needed for that, or because they changed their mind later.

The changes right now to the UI + updater are good and help the people that really want these files out of their disk.

qaisjp (approved):

PR #4659, opened 2026-01-21
Allow changing client file cache path in Advanced Settings

### LINKED ISSUE #4655: Allow file cache path to be changeable from MTA's menu

It displays it, so why not also let us actually change it? No setting for it in `\MTA\config\coreconfig.xml` either. Currently you have to use Registry Editor instead - `Computer\HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Multi Theft Auto: San Andreas All\Common\File Cache Path`.

Add a Change or Modify button which opens a file explorer window like the Show in Explorer button does and let us select a folder and/or add a text input box so we can type or paste a folder path.

_No response_

### Additional context

_No response_

### Security Policy

Xenius97:
#### Summary
Adds the ability to change the client-side file cache location from the Advanced Settings menu.
Two new buttons were introduced: **Change**, which allows the user to select a custom cache directory via a file explorer dialog, and **Reset**, which restores the default behavior.

By default, the client continues to use the cache path stored in the Windows Registry. If a custom cache path is defined in `coreconfig.xml`, it takes precedence. If the custom directory becomes unavailable or is deleted, the system automatically falls back to the default path.

image

Closes #4655

#### Motivation
Currently, the cache path is only displayed in the UI and cannot be modified from within the client.
Changing it requires manually editing the Windows Registry.

This change makes cache path configuration accessible from the UI while keeping backward compatibility and safe fallback behavior.

#### Test plan
1. Open **Advanced Settings** and click **Change** to select a custom cache folder.
2. Restart the client and verify that the new cache path is in use.
3. Click **Reset**, restart again, and verify that the default cache path is restored.

#### Checklist
* [x] Code follows the [coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines).
* [x] The pull request is small, focused, and easy to review.

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -2453,3 +2453,95 @@ std::shared_ptr<CDiscordInterface> CCore::GetDiscord()
 {
     return m_pDiscordRichPresence;
 }
+
+// Get File Cache Path from coreconfig.xml or fallback to registry
+SString CCore::GetFileCachePath()
+{
+    // First check coreconfig.xml
+    CXMLNode* pRoot = GetConfig();
+    if (pRoot)
+    {
+        CXMLNode* pFileCachePath = pRoot->FindSubNode("file_cache_path");
+        if (pFileCachePath)
+        {
+            SString strPath = pFileCachePath->GetTagContent();
+            if (!strPath.empty())
+                return strPath;
+        }
+    }
+
+    // Fallback to registry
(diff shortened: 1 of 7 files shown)


qaisjp (changes requested):
Lets hash out the motivation and purpose on https://github.com/multitheftauto/mtasa-blue/issues/4655 before proceeding on more implementation

PR #4565, opened 2025-11-21
Add missing cutscene bones to engineLoadIFP

FileEX:
During tests of PR #4555, @MegadreamsBE noticed that some bones have no animation and are deformed, creating something like this...
image

This is caused by engineLoadIFP reading only the basic 32 bones, and those from cutscenes are completely ignored. After this PR, the problem is resolved.
image

PR #4555, opened 2025-11-18
Object animations

FileEX:
Closes #1977, #362
Merge after #4052

This PR adds support for custom object animations. Animations can be applied both to a model ID and to a specific entity (supported for objects and buildings).

https://www.youtube.com/watch?v=wn0R0C1vLp0

## Syntax
```lua
bool engineSetModelAnimation(int model, ifp animFile, string animName, [ int flags = 0x2 ] )
bool engineSetModelAnimation(int model, bool disableAnim)
nil engineRestoreModelAnimation(int model)

bool setElementAnimation(element theElement, ifp animFile, string animName, [ int flags = 0x2 ] )
bool setElementAnimation(element theElement, bool disableAnim)
bool setElementAnimationSpeed(element theElement, float speed)
```

## Code from video
```lua
local ifp = engineLoadIFP("obj_anim.ifp", "obj_anim")

local txd = engineLoadTXD("test.txd");
local dff = engineLoadDFF("test.dff");

engineConvertModelToType(1337, "clump")

engineImportTXD(txd, 1337)
engineReplaceModel(dff, 1337)

local obj = createBuilding(3425, 2485.03027, -1676.34106, 12.33802)
setElementAnimationSpeed(obj, 10)

local objw2 = createObject(3425, 2478.41260, -1663.71875, 12.33372);

local obj2 = createObject(1337, 2492.38232, -1674.29016, 12.33595)
setElementAnimation(obj2, ifp, 'Spin')
setElementAnimationSpeed(obj2, 0.2)

local obj3 = createObject(1337, 2491.65942, -1667.57007, 12.34375)
setElementAnimation(obj3, ifp, 'Left_Right')

engineSetModelAnimation(13562, false)
local obj4 = createObject(13562, 2472.57715, -1673.12170, 13.33592)
```

I’m not sure if more complex animations work, because in #1977 it was mentioned:

> But before we can play object anims, the object must have skinned data like peds in the dff model
> engineReplaceModel crashes when trying to load skin data for object models

Unfortunately, I don’t have a more advanced model or animations to test this case, so there is no guarantee that it works. If someone encounters this issue in the future and provides me with a model for testing, I'll fix the crash.

Thanks to @1Selective for the model and animations used for testing.

PR #4664, opened 2026-01-25
Add the ability to load IFP animations without compression

FileEX:
#### Summary
This PR adds an optional uncompressedAnimations argument to the `engineLoadIFP` function, allowing you to specify which animations should be loaded without compression.

```lua
bool engineLoadIFP(string ifp_file / string raw_data, string  custom_block_name [, table uncompressedAnimations = nil ] )
```

#### Motivation
Currently, all IFP animations are loaded as compressed by default, regardless of whether compression is necessary. While testing PR #4555, @MegadreamsBE discovered several issues when using cutscene animations, including broken facial bones (fixed in #4565) and offset problems. It was determined that the issue occurs because all animations are loaded as compressed by default, whereas GTA loads some cutscene animations in a special way as uncompressed (the uncompress section in .cut files).

#### Test plan
@MegadreamsBE tested this.

Thanks to @Merlin for helping with the crash investigation.

#### Checklist


qaisjp (changes requested):
I haven't review the code but

> @MegadreamsBE tested this.

Could you go into more detail into what the actual tests were? E.g. sample resources that they tested against

FileEX:
> I haven't review the code but
> 
> > @MegadreamsBE tested this.
> 
> Could you go into more detail into what the actual tests were? E.g. sample resources that they tested against

This branch https://github.com/multitheftauto/mtasa-blue/compare/master...FileEX:mtasa-blue:megadreams/object-animations_with_clumps was used for testing, because currently MTA doesn’t include object animations or the ability to replace clump-type models (#4555 and #4052).

Test resource:
[anims2.zip](https://github.com/user-attachments/files/24841078/anims2.zip)

I personally don’t know cutscene animations very well, but MegaDreams is familiar with them, and the test resource comes from him.

Client/mods/deathmatch/logic/CClientIFP.cpp
```diff
@@ -404,6 +455,22 @@ void CClientIFP::ReadKrt0FramesAsCompressed(std::unique_ptr<CAnimBlendSequence>&
     }
 }
 
+void CClientIFP::ReadKr00FramesUncompressed(std::unique_ptr<CAnimBlendSequence>& pAnimationSequence, const std::int32_t& cFrames)
```
Lpsd, line 467: It seems like the ReadKr* methods are practically doing the same thing, couldn't an interface help cut down code-duplication here?

MegadreamsBE:
> I haven't review the code but
> 
> > @MegadreamsBE tested this.
> 
> Could you go into more detail into what the actual tests were? E.g. sample resources that they tested against

Two-fold:
I run a resource to test a specific IFP which contains uncompressed animations. The fix made it work correctly, before the coordinates were wrong. FileEX linked it in a previous comment.

I ran multiple cutscene IFPs, which rely on uncompressed animations with this new code through my Sphene project. No crashes nor issues were observed.

I can vow for the stability of this change.

Client/mods/deathmatch/logic/CClientIFP.cpp
FileEX: That's true, it could be done better, but in my opinion, that's not the topic of this PR. This PR simply adds a new argument to the function, rather than refactoring the existing code.

qaisjp (dismissed):

PR #4669, opened 2026-01-26
Add `onClientElementClicked` for client side

Xenius97:
#### Summary
`onClientElementClicked` event is now available on the client side, whereas previously it was only on server side (`onElementClicked`).

#### Motivation
During development, I often missed having this functionality on the client side, as I actively use it on the server as well. It provides a much simpler way to detect clicks on specific elements compared to manually handling it within `onClientClick`.
In my opinion, it can also be quite useful from a script optimization perspective.

#### Test plan
Click on a ped
```lua
-- create ped
local ped = createPed(0, 0, 0, 3)

-- new method
addEventHandler("onClientElementClicked", ped, function(button, state, worldX, worldY, worldZ)
    if button ~= "left" or state ~= "down" then return end
    outputChatBox("Ped clicked: " .. tostring(button) .. ", " .. tostring(state) .. ", " .. tostring(worldX) .. ", " .. tostring(worldY) .. ", " .. tostring(worldZ) .. ", Ped ID: " .. tostring(source) .. "(new method).")
end)

-- old method
-- addEventHandler("onClientClick", root, function(button, state, absoluteX, absoluteY, worldX, worldY, worldZ, clickedElement)
--     if button ~= "left" or state ~= "down" then return end
--     if clickedElement and clickedElement == ped then
--         outputChatBox("Ped clicked: " .. tostring(button) .. ", " .. tostring(state) .. ", " .. tostring(worldX) .. ", " .. tostring(worldY) .. ", " .. tostring(worldZ) .. ", Ped ID: " .. tostring(clickedElement) .. "(old method).")
--     end
-- end)

-- multiple peds
for i = 1, 5 do
    local ped = createPed(0, i * 2, 0, 3)
    addEventHandler("onClientElementClicked", ped, function(button, state, worldX, worldY, worldZ)
        if button ~= "left" or state ~= "down" then return end
        outputChatBox("Ped clicked: " .. tostring(button) .. ", " .. tostring(state) .. ", " .. tostring(worldX) .. ", " .. tostring(worldY) .. ", " .. tostring(worldZ) .. ", Ped ID: " .. tostring(source) .. "(new method).")
    end)
end
```

#### Checklist


FileEX:
Adding this event is unnecessary duplication of events - we would be firing this event alongside `onClientClick`, which already includes the clicked element as an argument. On the server side, this event exists only because there is no onClick event there. Imo, adding a note on the `onElementClicked` wiki page stating that the client-side equivalent is `onClientClick` would be sufficient.

Xenius97:
> On the server side, this event exists only because there is no onClick event there.

https://wiki.multitheftauto.com/wiki/OnPlayerClick

FileEX (approved):

PR #4682, opened 2026-02-01
Fix bullet sync bugs & crash

FileEX:
# Fix bullet sync bugs and crash; harden the bullet sync pipeline

Fixes the crash from huge/garbage bullet coordinates (cheaters can do it to other players), gets custom weapon sync working properly again in both directions, brings the threadnet (sim) path up to the same checks as the main path, and limits how fast a player can relay shots.

### What changed

- Positions use `SPositionSync` and damage uses `SFloatAsBitsSync<8>`
  (0-200), same as the rest of the netcode. The old code wrote raw floats,
  so garbage values could reach other clients. The formats are bounded by
  design now.
- The order byte (and its client-side counters) is gone from both packet
  types. Client dedup is a 50 ms window per shooter keyed on the quantized
  start/end pair. 50 ms is under
  the fastest real fire rate in 22-34 (MP5, 94 ms), so repeated shots from
  someone standing still aren't eaten and re-delivered copies are dropped.
  The state is only saved after the whole packet has parsed.
- Netcode version bumped to 0x1DF, so old and new builds fail at join
  instead of silently losing bullet sync.
- Custom weapon path fixed: the server packet now reads the same format the
  client sends, the source guard was backwards (rejected every packet), and
  unknown weapon ids are rejected. The weapon id is cached at read, so
  `Write` can't touch a weapon a script deleted. The handler gets a rate
  gate of `max(10 ms, the weapon's fire time)`.
- Main path: the range check now uses the weapon's best skill tier + 10% +
  15 m slack, instead of the shooter's skill stat (starts at 0, so legit
  shots got rejected). Added a 40 ms per-player rate gate.
- Sim path: now checks spawned/alive, weapon, ammo, range, shooter distance (50 m / 100 m) and a 40 ms gate, using a snapshot filled on the main thread. The range check measures the packet's weapon at the best skill tier, same as the main path. Doesn't relay before the first puresync, so the player id in the relay is always set. The damaged target isn't checked here because the element table isn't safe on the sim thread; the main path check and the client cast cover it. This is commented in the code.
- Client: rejects zone above 9 and drops shots whose damaged target id doesn't resolve to a player, checks positions/bounds/trajectory on both readers, caps custom shots by the weapon's own range (range * 1.1 + 15 m, 400 m floor), and the damage preserve flag now matches the server.
- Every parse point now checks the positions are inside the world bounds:
  X/Y +-8192 (the sync range) and Z +-100000 (the sync limit), so
  underground and high-altitude maps keep bullet sync.
- Small stuff: `HasWeaponBulletSync` takes `uint8_t` on both ends, and
  `CCustomWeapon` seeds its fire rate from the weapon stat (it was
  uninitialized until a script touched it, and the new rate gate reads it).
- Custom weapons now keep their range cap in sync: when a script sets `weapon_range`, the value is sent to clients with a new SET_CUSTOM_WEAPON_WEAPON_RANGE element RPC, and players who join later get it re-sent during the initial data stream. Both ends reject shots beyond range * 1.1 + 15 m using the same value.
- Custom weapon shots get the same shooter checks as the player path: the packet is dropped when the player is dead or not spawned, and when the shot starts farther than 50 m (100 m in a vehicle) from the player.
- The custom weapon path is no longer filtered by weapon type on the receiving client, so every type the server accepts is rendered by viewers (matches the send and server sides, which were already type-agnostic).
- Clients now drop bullet sync packets whose damaged target id doesn't resolve to a player, so a relayed shot that the main path would reject can't feed a bogus target into the local game.
- Server-side custom weapon stats are now initialized properly: each weapon loads the game's standard data for its type at creation (the client already did), weapon types outside the standard tables read fixed zeros instead of uninitialized memory, and the stat flag helpers no longer crash for types without original data.
- The sim path range gate now measures the packet's weapon at the best skill tier, matching the main path, using a per-slot snapshot filled on the main thread. Zone-0 viewers now accept the same shots zone-1/2 viewers get.
- The custom weapon rate gate floor is 10 ms, not 40 ms, so fast-firing weapons like the minigun aren't throttled (the gate is max(10 ms, the weapon's fire time)).

### Testing

- Still needs a 2-PC run: shooter on foot, on a server-side object, in a
  vehicle, matching builds, threadnet on/off, high-altitude
  and underground maps, custom weapons both ways. Can be done when the 1.7
  nightly is out.


ArranTuna:
I tested it with 2 PCs and bullet sync is completely broken, no bullets are being synced.

onPlayerWeaponFire was triggered, onClientPlayerWeaponFire doesn't.

FileEX:
> I tested it with 2 PCs and bullet sync is completely broken, no bullets are being synced.
> 
> onPlayerWeaponFire was triggered, onClientPlayerWeaponFire doesn't.

Did you test this after my latest commit (50 mins ago)?

ArranTuna:
I have re-tested with the newest commit and bullets still do not sync.

ArranTuna:
image
 
Just see the flashing, no bullets. I don't know if related to your PR but I don't see the remote player reloading.

FileEX:
Reloading the weapon is probably a separate issue. As for bullet sync, I’m not really sure what to say, because for me onClientPlayerWeaponFire and onPlayerWeaponFire are being triggered, and the vector values are correct.

I’m not entirely sure what exactly I’m supposed to see. What does the bullet you’re referring to look like? I thought it was the muzzle flash from the weapon.

ArranTuna:
onClientPlayerWeaponFire triggers for bullets you fire, it does not trigger for remote players. Remote player bullets are not being synced, just the animation and muzzle flash.

FileEX:
> onClientPlayerWeaponFire triggers for bullets you fire, it does not trigger for remote players. Remote player bullets are not being synced, just the animation and muzzle flash.

Are you sure?
https://streamable.com/emvkpc

Dutchman101 (changes requested):
Notes from my review of the changes - these are bugs and regressions in the current state:

Custom weapon sync is broken in both directions. The client changes the format (positions to SPositionSync, order byte dropped), but the server packet (CCustomWeaponBulletSyncPacket) is untouched - so the two ends read different wire layouts. Client-to-server hits an extra problem: the read guard is inverted (it rejects when the source is set, and the source is always set for connected players), so every packet gets deleted before parsing. Server-to-client misparses the old server write into bit-shuffled positions. This path also gets none of the new server-side validation the player path gets.

Double relay. The change adds a RelayNearbyPacket call before the event, but master already has one after the event (from #4763). Every shot is broadcast twice to the same viewers.

No version gate. The wire format changes unconditionally, but _NETCODE_VERSION stays at 0x1DE. Mixed old/new builds connect fine and then silently lose bullet sync in both directions.

Client dedup removed with nothing in its place. The old duplicate check (prev start/end + order counter) is deleted, and the server can still double-deliver - so receivers process the same shot twice.

Z floor is too strict. IsInWorldBounds uses -100 as the Z floor. Players on underground custom maps (Z below -100) have every shot silently dropped.

The two paths validate at different strengths, and the weaker one runs first. The sim path relays before the main path validates, but it has no range check, no shooter-distance check, no ammo check, no spawn/alive check, and no target-element check - so zone-0 viewers receive shots the main path rejects, including shots from dead players. The main path's own range gate has a separate problem: it measures the shooter's skill level, and server-side ped stats start at zero, so clients who never raised their weapon skill have legitimate shots rejected (the #5122 type of issue). The "10% tolerance" is also ~4.9% on distance because it's applied to the squared range.

Damage handling has three issues. The client sends with the preserve flag false and the server uses true, so small damage values round to zero and the hit data gets dropped. The main packet's m_damage field is never initialized at construction, and the damage reads don't check their results - a truncated packet reaches an indeterminate-value read. The "precision fix" claim is also backwards: damage goes from full float precision to 8-bit quantization (about 0.78 damage steps).

No per-player rate limit on either path. A shooter can emit bulletsync at line rate and every valid packet is relayed to all nearby viewers.

Minor leftovers: Dead EPSILON/EPSILON_SQ constants, unused prev-bullet fields on CClientPlayer, and a uint8_t/uint32_t signature mismatch on HasWeaponBulletSync between the two ends.

A few smaller things as well:

The clip-ammo check removal in the custom weapon handler goes against what #4381 added it for (empty-clip fire gets accepted again). It's currently hidden by the guard bug, but it becomes live once that's fixed - worth confirming the removal is on purpose.
The extra relay call sits before the event, so the shot is broadcast before Lua runs. That's fine today, but if onPlayerWeaponFire ever becomes cancellable (there's an open issue # for it), a cancel couldn't take the shot back. The sim relay has the same property.
The new damage range checks can never actually fire - the 8-bit format always decodes in range - so they're dead code. Same for the removed m_damaged == 0 check (ID 0 never exists).
Scripts see the 1/1024 position quantization in the weapon fire events now (server event start/end, client event origin for remote shots). Relevant for aim-line math.

Dutchman101:
In my code review, i said "these are bugs and regressions in the current state" because i am preparing a commit to the PR branch with fixes for all of my findings.

Dutchman101:
I've now pushed a commit with fixes for all of my [findings](https://github.com/multitheftauto/mtasa-blue/pull/4682#pullrequestreview-4937153472): https://github.com/multitheftauto/mtasa-blue/pull/4682/changes/fa5823412fa833730e4d6896907c47cc4dcf302b
I'm awaiting FileEX to review it separately, and if it holds up, we'll merge this PR.

Client/mods/deathmatch/logic/CNetAPI.cpp
```diff
@@ -2297,54 +2297,78 @@ void CNetAPI::ReadVehiclePartsState(CClientVehicle* pVehicle, NetBitStreamInterf
...
+    // window must stay below the fastest legitimate fire interval (MP5 at
+    // 94 ms) so repeated shots from a player standing still are not eaten,
+    // while the two relay copies arrive within the same server frame.
+    const CTickCount tickCountNow = CTickCount::Now();
+    if (startPosition.data.vecPosition == player->m_vecPrevBulletSyncStart && endPosition.data.vecPosition == player->m_vecPrevBulletSyncEnd &&
+        (tickCountNow - player->m_BulletSyncDedupTime) < CTickCount(50LL))
+        return;
+
+    // 200 is MAX weapon damage
+    SFloatAsBitsSync<8> damage(0, 200.0f, true, false);
```
FileEX: Are 8 bits enough? With 8 bits, you can only represent integer values, so if a weapon deals, for example, 2.64 damage, the server will read it as 2.

Server/mods/deathmatch/logic/net/CSimBulletsyncPacket.h
```diff
@@ -25,12 +25,11 @@ class CSimBulletsyncPacket : public CSimPacket
...
-        CVector      start{};
-        CVector      end{};
-        std::uint8_t order{};
-        float        damage{};
-        uchar        zone{};
-        ElementID    damaged = INVALID_ELEMENT_ID;
+        eWeaponType         weapon = eWeaponType::WEAPONTYPE_UNARMED;
+        SPositionSync       start{};
+        SPositionSync       end{};
+        SFloatAsBitsSync<8> damage{0.0f, 200.0f, true, false};
```
FileEX: CBulletsyncPacket::MAX_DAMAGE

Server/mods/deathmatch/logic/packets/CBulletsyncPacket.h
```diff
@@ -37,24 +36,20 @@ class CBulletsyncPacket final : public CPacket
...
-    CVector      m_start{};
-    CVector      m_end{};
-    std::uint8_t m_order{};
-    float        m_damage{};
-    std::uint8_t m_zone{};
-    ElementID    m_damaged{INVALID_ELEMENT_ID};
+    eWeaponType         m_weapon{};
+    SPositionSync       m_start{};
+    SPositionSync       m_end{};
+    SFloatAsBitsSync<8> m_damage{0.0f, 200.0f, true, false};
```
FileEX: CBulletsyncPacket::MAX_DAMAGE

Server/mods/deathmatch/logic/packets/CBulletsyncPacket.cpp
```diff
@@ -109,45 +82,42 @@ bool CBulletsyncPacket::ReadOptionalDamage(NetBitStreamInterface& stream)
         return true;
     }
 
-    stream.Read(m_damage);
-    stream.Read(m_zone);
-    stream.Read(m_damaged);
-
-    if (IsNaN(m_damage))
+    if (!stream.Read(&m_damage) || !stream.Read(m_zone) || !stream.Read(m_damaged))
```
FileEX, line 85: Since we already have the damage on the server side, maybe it would be worth adding it to the `onPlayerWeaponFire` event?

Server/mods/deathmatch/logic/net/CSimPlayer.h
```diff
@@ -58,6 +58,19 @@ class CSimPlayer
...
     SSimVehicleDamageInfo m_VehicleDamageInfo;
 
+    // Used in bullet sync validation (HandleBulletSync)
+    CVector      m_vecPosition;
+    CElapsedTime m_BulletSyncRateTimer;
+
+    // Bullet sync validation snapshot, refreshed by UpdateSimPlayer on the
+    // main thread. The sim thread must not read live player state, so the
+    // sim path's possession and ammo checks run against these copies.
+    static constexpr unsigned char WEAPON_SLOT_COUNT = 13;  // Matches WEAPON_SLOTS in CPed.h
```
FileEX: `std::size_t`

Server/mods/deathmatch/logic/net/CSimPlayer.h
```diff
@@ -58,6 +58,19 @@ class CSimPlayer
...
+    bool                           m_bIsSpawned;
+    bool                           m_bIsDead;
+    unsigned char                  m_WeaponTypes[WEAPON_SLOT_COUNT];
+    unsigned short                 m_WeaponTotalAmmo[WEAPON_SLOT_COUNT];
```
FileEX: Shouldn't we initialize this?

```cpp
    bool                           m_bIsSpawned{};
    bool                           m_bIsDead{};
    unsigned char                  m_WeaponTypes[WEAPON_SLOT_COUNT]{};
    unsigned short                 m_WeaponTotalAmmo[WEAPON_SLOT_COUNT]{};
```

Dutchman101:
Explanation for the latest fix-up commit (575d0ea) i pushed to this PR:

**Custom weapon range cap parity**

The server's custom weapon shot cap uses the weapon's own stat, which setWeaponProperty(weapon, "weapon_range", ...) modifies, but the client's cap used a client-side stat that no RPC updated. A script-raised range made the server accept and broadcast shots that every receiving client then dropped. Added a SET_CUSTOM_WEAPON_WEAPON_RANGE element RPC (appended above NUM_RPC_FUNCS, so no existing RPC id shifts) that the setter broadcasts, a client handler that applies it to the weapon's stat copy, and a join-time resend in CGame::InitialDataStream so late joiners measure shots the same way the server does.

**Server-side weapon stat initialization**

CCustomWeapon now loads the pristine weapon data through LoadDefault after creating its stat, so the range cap, the clip and fire-time initialization read real values instead of uninitialized memory.
Both CWeaponStat constructors zero the weapon data first, since LoadDefault only fills types 22-42; the other types now read fixed zeros instead of garbage, and the parameterless constructor also sets a sane weapon type and skill.
HandleFlagsValueChange now null-checks the original-stat lookups. Types without original data keep the anim bits off instead of deref a missing stat.

**Custom weapon path security parity**

CCustomWeaponBulletSyncPacket::Read now rejects shots from dead or not-yet-spawned players and enforces the same 50 m / 100 m shooter proximity check as the player path, so a hacked client cant report shots from anywhere on the map.
Restored the empty-clip rejection that the first pass dropped, and lowered the custom path fire-rate floor from 40 ms to 10 ms so legitimate minigun fire (14 ms loop) is not throttled.

**Sim path range gate parity**

The sim range gate measured the held weapon's skill-tier range, which can disagree with the main path's widest-tier measurement. CSimPlayer now carries a per-slot widest-tier range snapshot filled on the main thread, so zone-0 viewers accept the same shots zone-1/2 viewers get.

**Client-side hardening**

ReadBulletsync now drops shots whose target id doesn't resolve to a player (the sim relay can carry shots the main path rejects), and ReadWeaponBulletsync drops the 22-34 type gate that the custom path never applied on the send or server sides.

Server/mods/deathmatch/logic/net/CSimBulletsyncPacket.h
```diff
@@ -25,11 +27,11 @@ class CSimBulletsyncPacket : public CSimPacket
...
-        SPositionSync       end{};
-        SFloatAsBitsSync<8> damage{0.0f, 200.0f, true, false};
-        std::uint8_t        zone{};
-        ElementID           damaged = INVALID_ELEMENT_ID;
+        eWeaponType          weapon = eWeaponType::WEAPONTYPE_UNARMED;
+        SPositionSync        start{};
+        SPositionSync        end{};
+        SFloatAsBitsSync<16> damage{0.0f, CBulletsyncPacket::MAX_DAMAGE, true, false};
+        std::uint8_t         zone{};
+        ElementID            damaged = INVALID_ELEMENT_ID;
```
FileEX:
```lua
eWeaponType          weapon{eWeaponType::WEAPONTYPE_UNARMED};
ElementID            damaged{INVALID_ELEMENT_ID};
```

Dutchman101:
Note that the established convention (see [this code search](https://github.com/search?q=repo%3Amultitheftauto%2Fmtasa-blue+WEAPON_WEAPON&type=code)) is the basis for using internal names that contain "WeaponWeapon".

Dutchman101 (approved):
@FileEX Thanks for the collab, we quickly polished this PR and mutually reviewed each other's changes.

Dutchman101:
This PR somehow boosted bullet sync's overall performance: Better reproduction of shots with less noticable latency, less dropped shots (you'd always notice if you paid attention to that before). I didn't expect it, but this will make MTA 1.7's release even more legendary, as it will feel like even better netcode in combat.

Issue #888, opened 2019-04-25
Add greater control and server support to fire


MegadreamsBE:
**Is your feature request related to a problem? Please describe.**

Fire is client-side right now and has to be manually synced by your scripts if you want it to be somewhat similar for other players including manual streaming (due to the 60 fires limit). It is however not even possible to get a hold of fire that wasn't created by your script (like fire spreading or fires caused by explosions).

**Describe the solution you'd like**

For a start a fire should be represented as an element, just like vehicles, peds, markers, projectiles, etc. This would make it much easier to get a hold of all active fires and already allow more control over them.

Then, it should be possible to control how a fire spreads (even completely disabling it) and for how long it actually lasts (not just only depending on its size). 

At last it should also be possible to create fires server-side (should this also include fires created by explosion however? - discussion point) and have them properly synced among players. Just like other elements they'd be streamed (allowing for more active fires, although we could look into raising this hardcoded limit) and the fire spread will be handled by the player that is considered to be the syncer.

**Describe alternatives you've considered**

Scripting most of the above, which, let us be real, is not a complete solution.

sbx320:
I've looked at this in the past. One major problem I faced back then was that practically everything fire-related incorporates some form of randomness, which makes it complicated to sync properly. 

The fire limit is very simple to expand, as fire isn't all that complicated. After all it's just an effect with some minor logic stuck to it to spread, set things on fire and extinguish.

MegadreamsBE:
The "easiest" thing to do then I suppose would be incorporating the logic into MTA directly and disallowing GTA SA itself from deciding things like fire spreading. From what I see there doesn't seem to be a lot of logic related to fire.

qaisjp:
This sounds good and reasonable.

There is some prior work done ~10 years ago, here: https://github.com/multitheftauto/mtasa-blue/commits/feature/fire-sync. The branch was brought up to date last year and, last I checked, still works. 

Feel free to start over or fork that branch.

Some questions to discuss:
- If I recall correctly, in that branch a created fire is more like a "fire source". Fire from the source may spread freely, and destroying the fire object will just extinguish the source. The fire that spreads just naturally dies out (eventually). Should we keep this behaviour?
- How do you think "naturally occurring fire" should behave? This is related to the above.
    - i.e. fire from / resulting from a burning vehicle; or as you mentioned - from explosions. 
    - Should naturally occurring fire be their own objects?
[...]

PR #4707, opened 2026-02-14
Fire reimplementation

FileEX:
#### Summary
This PR is similar to #4353 because it also implements its own fire handling, effectively disabling the default GTA functions. Reimplementing a specific feature is generally a better approach than adding more ASM hooks just to change or add something, although it obviously requires knowledge and time. This PR is not yet fully complete, but a large portion of the work - i.e., reverse engineering the original code (gta-reversed often contains errors/mistakes/bugs, so verification is always necessary) and its own implementation.. has already been done. Ultimately, this PR is intended only to pave the way for subsequent PRs. Due to its size, it will not contain everything.

## Benefits of the reimplementation:
- No 60-fire limit
- Full control over fire: custom lifespan, position, size (particle), and ability to enable/disable fire sound
- Fires created via `createFire` will return a CClientFire element that can be moved with `setElementPosition` or destroyed via `destroyElement` allowing individual flames to be extinguished without affecting the surrounding area
- Functions enabling fire immunity (`setElementFireProof` / `isElementFireProof`) - useful for things like fire-resistant firefighter outfits, etc.
- Server synchronization; `createFire` and other fire functions available as shared
- Ability to enable/disable fire spreading, e.g., via a special world property. Currently this is permanently disabled in MTA for unknown reasons.
- Introduction of events to notify about fire creation (if deemed necessary)

There’s potentially much more that can be done; with a custom implementation, virtually anything is possible. If anyone has interesting ideas, they can share them here.

## TODO for this PR to be ready:
~~- Create the `CClientFire` class so fires created via `createFire` generate this element and can be moved or destroyed~~
~~- Fix bugs originally handled via hooks in **CMultiplayerSA**~~
~~- Implement `extinguishAllFires`~~

#### Motivation
#888 

Very little control over fire. createFire only returns a bool, and until recently you couldn’t even extinguish fires. This is probably the most underdeveloped and neglected feature in MTA..

#### Test plan
The PR is currently stable. I’ve been testing it over the past two days, fixing numerous bugs. At this point, the fire behavior in the MTA matches the default exactly and I haven’t observed any bugs or deviations from the original behavior.

#### Checklist


FileEX:
Ready for review, cleanup will follow later in separate PRs.

Pirulax:
> gta-reversed often contains errors/mistakes/bugs, so verification is always necessary

If you notice any, please let us know, or, even better, make a PR.

T-MaxWiese-T:
Great idea!
But in my opinion, what's still missing is an event that triggers when a fire is extinguished, PLUS a `cancelEvent` so you can prevent a fire from being modified.

PR #4714, opened 2026-02-20
Allow the Rhino to render headlight cones and coronas

rxyyy:
#### Summary
This pull request allows the Rhino (model ID 432) to use its headlight and brakelight textures and dummies by patching `CAutomobile::Render` and `CAutomobile::PreRender`. The vanilla model already has markers in place for these lights, and they are roughly in the correct positions. No model changes or Lua API changes are required.

#### Motivation
The vanilla model already has markers in place for these lights, but they were disabled by code. This pull request makes it so the base model and any potential replacement models of the Rhino can use the lights like other vehicles.

#### Test plan
The change was tested locally by spawning a Rhino on a default server setup and observing that the lights now work.

an9w84t

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1571,8 +1571,8 @@ void CMultiplayerSA::InitHooks()
     MemSet((void*)0x6C4453, 0x90, 0x68);
 
     // Allow model ID 432 (Rhino) to render headlight cones and coronas
-    MemSet((void*)0x6A2EAF, 0x90, 2);
-    MemSet((void*)0x6ABC85, 0x90, 2);
+    MemSet((void*)0x6A2EAB, 0x90, 6);
+    MemSet((void*)0x6ABC81, 0x90, 6);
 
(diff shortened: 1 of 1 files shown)


qaisjp (dismissed):
lgtm

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1570,6 +1570,10 @@ void CMultiplayerSA::InitHooks()
     MemSet((void*)0x6C444B, 0x90, 6);
     MemSet((void*)0x6C4453, 0x90, 0x68);
 
+    // Allow model ID 432 (Rhino) to render headlight cones and coronas
+    MemSet((void*)0x6A2EAF, 0x90, 2);
+    MemSet((void*)0x6ABC85, 0x90, 2);
```
qaisjp: thanks @rxyyy — a teammate also says:

> could've NOPed the `cmp` too

is that helpful feedback?

rxyyy: Thank you, indeed it is helpful feedback. I have done so in the latest commit of this PR.

qaisjp (approved):
thank you!

FileEX (approved):

PR #4721, opened 2026-02-23
add script to track differences between release/1.6.0 and master

qaisjp:
#### Summary

This script can help us keep track of which changes happened to master which weren't backported to release/1.6.0, and vice-versa.

#### Motivation

Ideally we backport as many changes we can to release/1.6.0, except the changes that involve versioning

#### Test plan

Run `ruby diff.rb` (works out of the box without additional dependencies, inside WSL)


utils/master-diff/diff.rb
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env ruby
+require 'open3'
+require 'fileutils'
+require 'set'
+require 'time'
+require 'yaml'
+
+MASTER = 'master'
+FORK   = 'release/1.6.0'
```
sbx320, line 9: BRANCH?

Fork seems slightly odd here.

sbx320 (commented):
I suppose  `output/` should be added to a .gitignore.

> Run ruby diff.rb (works out of the box without additional dependencies, inside WSL)

`bash: ruby: command not found` 😢

PR #4728, opened 2026-02-26
Add check bytes for voice packet

G-Moris:


Server/mods/deathmatch/logic/packets/CVoiceDataPacket.cpp
```diff
@@ -24,6 +24,9 @@ bool CVoiceDataPacket::Read(NetBitStreamInterface& BitStream)
 
     if (BitStream.Read(voiceBufferLength) && voiceBufferLength <= m_voiceBuffer.capacity())
     {
+        if (!BitStream.CanReadNumberOfBytes(voiceBufferLength))
```
DmitriyColeman, line 27: This check is actually unnecessary and essentially redundant. The logic drops invalid packets in 
`if (BitStream.Read(voiceBufferLength) && voiceBufferLength <= m_voiceBuffer.capacity())`

G-Moris: Will it handle a scheme like this?

```c++
Write(2048);
Write("", 0);
```

I doubt it.

DmitriyColeman: Yes, it does. If you attempt to read more from the BitStream than is available, it simply returns false.

G-Moris: > Yes, it does. If you attempt to read more from the BitStream than is available, it simply returns false.

Calling `resize` is more expensive than performing a single check.

DmitriyColeman: If you call resize when the buffer is already that size, it doesn't consume significant CPU time, as no reallocations occur..

G-Moris: > If you call resize when the buffer is already that size, it doesn't consume significant CPU time, as no reallocations occur..

If you're that confident about it - good.

PR #4744, opened 2026-03-06
Fix streamer bug where objects can constantly stream in and out

### LINKED ISSUE #4743: Streamer bug where an object will constantly stream in / out
### Describe the bug

This is a very old bug, I remember reporting it many years ago, but couldn't find the bug report, so reporting again as it still exists.

I discovered this when debugging low FPS, turned out it was sometimes an object will constantly stream in and out many times per second:

object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0
object streamed in userdata: 7C2D34E0

### Steps to reproduce

crun addEventHandler("onClientElementStreamIn", root, function() outputChatBox(getElementType(source).." streamed in "..tostring(source)) end)

You need a map with lots of (possibly over 500) objects.

Freecam around until eventually you get this bug where the same object keeps streaming in.

### Version

Any client in the last few years.

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

ArranTuna:
#### Summary
Fixes https://github.com/multitheftauto/mtasa-blue/issues/4743

Worked with codex to develop a fix for this bug, first it added "Avoid swap ping-pong when two candidates are almost the same distance" but that wasn't enough to fix it on it's own, so a last stream out time was added.

#### Motivation
A very old MTA bug that can cause FPS instability when a player is affected by it.

#### Test plan
Was a little hard to test this, but could reproduce the bug in under 1 minute, by constantly moving slightly with freecam. However after adding the fix I could not reproduce it when testing for at least 5 minutes and observed no other issues.

#### Checklist

Made sure to comply with the coding guidelines such as using std::uint32_t

### DIFF
--- a/Client/mods/deathmatch/logic/CClientStreamElement.cpp
+++ b/Client/mods/deathmatch/logic/CClientStreamElement.cpp
@@ -20,6 +20,7 @@ CClientStreamElement::CClientStreamElement(CClientStreamer* pStreamer, ElementID
     m_fExpDistance = 0.0f;
     m_bStreamedIn = false;
     m_bAttemptingToStreamIn = false;
+    m_lastStreamOutTime = 0u;
     m_usStreamReferences = 0;
     m_usStreamReferencesScript = 0;
     m_pStreamer->AddElement(this);
@@ -67,6 +68,7 @@ void CClientStreamElement::InternalStreamOut()
     {
         StreamOut();
         m_bStreamedIn = false;
+        m_lastStreamOutTime = static_cast<std::uint32_t>(CClientTime::GetTime());
 
         // Stream out attached elements
         CClientObject* thisObject = DynamicCast<CClientObject>(this);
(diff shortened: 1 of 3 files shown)


FileEX:
I don't think this is the right solution. We should fix the root cause of the issue instead of just limiting the time between streaming in and streaming out.

ArranTuna:
I know it's not a perfect fix but the root cause is that the 2 objects are virtually identical in distance so the tiny frame by frame changes in position and camera rotation are causing it to switch back and forth, I don't know how else that could be fixed.

Dutchman101 (approved):

PR #4745, opened 2026-03-07
Add removeElementData client side

### LINKED ISSUE #594: Client-side removeElementData
Always if i want removeElementData i must use triggerServerEvent and send element from which i want remove data.

Adding removeElementData too on client, (shared)

**Additional context**

ArranTuna:
#### Summary
Fixes https://github.com/multitheftauto/mtasa-blue/issues/594

removeElementData can now be called on the client which actually deletes (rather than just sets to nil or false) an element data. It also sends a new packet type to the server and syncs it with clients.

#### Motivation
To actually delete element data so that nil or false isn't being unnecessarily synced to future clients.

#### Test plan
Although I didn't test with 2 PCs I used it on an object then reconnected to verify that new clients are no longer being synced a removed element data.

Edit: tested with 2 PCs and the remote players have the element data removed correctly.


ArranTuna:
Tested everything with 2 PCs, fixed that client variable wasn't being sent to onElementDataChange when a client used removeElementData.

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,7 +1062,31 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
...
+    if (currentVariable)
+    {
+        if (isSynced && !Entity.IsLocalEntity())
+        {
+            NetBitStreamInterface* pBitStream = g_pNet->AllocateNetBitStream();
+            // Write element ID, name length and name for server-side removal handling
+            pBitStream->Write(Entity.GetID());
+            std::uint16_t nameLength = static_cast<std::uint16_t>(name->length());
+            pBitStream->WriteCompressed(nameLength);
+            pBitStream->Write(name.ToCString(), nameLength);
```
FileEX: WriteString?

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,73 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
...
+
+void CRPCFunctions::RemoveElementData(NetBitStreamInterface& bitStream)
+{
+    CLOCK("NetServerPulse::RPC", "RemoveElementData");
+
+    if (!m_pSourcePlayer->IsJoined())
+    {
+        UNCLOCK("NetServerPulse::RPC", "RemoveElementData");
+        return;
+    }
```
FileEX:
```cpp
if (!m_pSourcePlayer->IsJoined())
    return;

CLOCK("NetServerPulse::RPC", "RemoveElementData");
```

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,73 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
...
+                const bool changesAllowed = clientChangesMode == eCustomDataClientTrust::UNSET ? !g_pGame->GetConfig()->IsElementDataWhitelisted()
+                                                                                               : clientChangesMode == eCustomDataClientTrust::ALLOW;
+                if (!changesAllowed)
+                {
+                    CLogger::ErrorPrintf("Client trying to change protected element data %s (%s)\n", m_pSourcePlayer->GetNick(), customDataName);
+
+                    CLuaArguments arguments;
+                    arguments.PushElement(element);
+                    arguments.PushString(customDataName);
+                    arguments.PushArgument(CLuaArgument());
```
FileEX: I don’t think passing nil as the new value is appropriate in the case of removal. Maybe we should add an additional parameter to onPlayerChangesProtectedData and onElementDataChanged that indicates whether the element data is being removed.

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,73 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
     }
     UNCLOCK("NetServerPulse::RPC", "RequestStealthKill");
 }
+
+void CRPCFunctions::RemoveElementData(NetBitStreamInterface& bitStream)
```
FileEX: early-returns

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,7 +1062,31 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
 
 bool CStaticFunctionDefinitions::RemoveElementData(CClientEntity& Entity, CStringName name)
 {
-    // TODO
+    assert(name);
+    assert(name->length() <= MAX_CUSTOMDATA_NAME_LENGTH);
+
+    bool          isSynced;
+    CLuaArgument* currentVariable = Entity.GetCustomData(name, false, &isSynced);
+    if (currentVariable)
```
FileEX: early-returns

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,7 +1062,31 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
...
+            NetBitStreamInterface* pBitStream = g_pNet->AllocateNetBitStream();
+            // Write element ID, name length and name for server-side removal handling
+            pBitStream->Write(Entity.GetID());
+            std::uint16_t nameLength = static_cast<std::uint16_t>(name->length());
+            pBitStream->WriteCompressed(nameLength);
+            pBitStream->Write(name.ToCString(), nameLength);
+
+            // Send RPC and deallocate
+            g_pClientGame->GetNetAPI()->RPC(REMOVE_ELEMENT_DATA_RPC, pBitStream);
+            g_pNet->DeallocateNetBitStream(pBitStream);
```
FileEX: Shouldn’t the onClientElementDataChanged event be triggered here?

FileEX (changes requested):

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
ArranTuna: It is, it gets called below here in Entity.DeleteCustomData(name);

Server/mods/deathmatch/logic/CRPCFunctions.cpp
ArranTuna: Sure, but as that would require changing the server side removeElementData I feel it'd be outside the scope of this PR to make that change here.

ArranTuna:
I've made the requested changes and re-tested it.

Server/mods/deathmatch/logic/CRPCFunctions.cpp
FileEX: We can do it later

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,8 +1062,28 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
...
+
+    bool          isSynced;
+    CLuaArgument* currentVariable = Entity.GetCustomData(name, false, &isSynced);
+    if (!currentVariable)
+        return false;
+
+    if (isSynced && !Entity.IsLocalEntity())
+    {
+        NetBitStreamInterface* pBitStream = g_pNet->AllocateNetBitStream();
+        // Write element ID, name length and name for server-side removal handling
```
FileEX: "name length" - outdated

FileEX (dismissed):

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,8 +1062,28 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
...
+    CLuaArgument* currentVariable = Entity.GetCustomData(name, false, &isSynced);
+    if (!currentVariable)
+        return false;
+
+    if (isSynced && !Entity.IsLocalEntity())
+    {
+        NetBitStreamInterface* pBitStream = g_pNet->AllocateNetBitStream();
+        // Write element ID and name for server-side removal handling
+        pBitStream->Write(Entity.GetID());
+        pBitStream->WriteString(name.ToCString());
```
tederis: I would suggest you to use `ToString` method instead of `ToCString`. `WriteString` accepts string_view, so std::string will nicely be converted to string_view without a string length caculation.

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,71 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
...
+    CElement* element = CElementIDs::GetElement(elementId);
+    if (!element)
+    {
+        UNCLOCK("NetServerPulse::RPC", "RemoveElementData");
+        return;
+    }
+
+    ESyncType              lastSyncType = ESyncType::BROADCAST;
+    eCustomDataClientTrust clientChangesMode{};
+    element->GetCustomData(customDataName.c_str(), false, &lastSyncType, &clientChangesMode);
```
tederis: Please, use `CStringName(customDataName)` instead. `.c_str()` implies overhead costs

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,71 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
...
+            else
+                m_pPlayerManager->BroadcastOnlySubscribed(CElementRPCPacket(element, REMOVE_ELEMENT_DATA, *outBitStream.pBitStream), element,
+                                                          customDataName.c_str(), m_pSourcePlayer);
+
+            CPerfStatEventPacketUsage::GetSingleton()->UpdateElementDataUsageRelayed(customDataName.c_str(), m_pPlayerManager->Count(),
+                                                                                     outBitStream.pBitStream->GetNumberOfBytesUsed());
+        }
+
+        if (lastSyncType == ESyncType::SUBSCRIBE)
+            m_pPlayerManager->ClearElementData(element, customDataName.c_str());
```
tederis: Please, use `CStringName(customDataName)` instead. `.c_str()` implies overhead costs

Server/mods/deathmatch/logic/CRPCFunctions.cpp
```diff
@@ -366,3 +368,71 @@ void CRPCFunctions::RequestStealthKill(NetBitStreamInterface& bitStream)
...
+
+    const bool changesAllowed = clientChangesMode == eCustomDataClientTrust::UNSET ? !g_pGame->GetConfig()->IsElementDataWhitelisted()
+                                                                                   : clientChangesMode == eCustomDataClientTrust::ALLOW;
+    if (!changesAllowed)
+    {
+        CLogger::ErrorPrintf("Client trying to change protected element data %s (%s)\n", m_pSourcePlayer->GetNick(), customDataName.c_str());
+
+        CLuaArguments arguments;
+        arguments.PushElement(element);
+        arguments.PushString(customDataName.c_str());
```
tederis: Please, use `CStringName(customDataName)` instead. `.c_str()` implies overhead costs

tederis (changes requested):

ArranTuna:
Requested changes have been added and re-tested.

tederis (dismissed):

FileEX (dismissed):

Client/mods/deathmatch/logic/CNetAPI.h
```diff
@@ -34,6 +34,7 @@ enum eServerRPCFunctions
     KEY_BIND,
     CURSOR_EVENT,
     REQUEST_STEALTH_KILL,
+    REMOVE_ELEMENT_DATA_RPC,
 };
```
qaisjp, line 37: I wonder why there are so few RPCs here. I feel like there are other ways of defining RPCs that perhaps we're missing?

Like surely there should be a regular `setElementData` RPC here?

Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp
```diff
@@ -1062,8 +1062,28 @@ bool CStaticFunctionDefinitions::SetElementData(CClientEntity& Entity, CStringNa
...
 {
-    // TODO
-    return false;
+    assert(name);
+    assert(name->length() <= MAX_CUSTOMDATA_NAME_LENGTH);
+
+    bool          isSynced;
+    CLuaArgument* currentVariable = Entity.GetCustomData(name, false, &isSynced);
+    if (!currentVariable)
+        return false;
```
qaisjp: We should be able to skip this check if the entity is a local entity, as DeleteCustomData checks if the field already exists: https://sourcegraph.com/r/github.com/multitheftauto/mtasa-blue@7d8fd38976d6e9fc1edf767697cde2e05b937631/-/blob/Client/mods/deathmatch/logic/CClientEntity.cpp?L501-506

Client/mods/deathmatch/logic/CNetAPI.h
qaisjp: Okay, an LLM figured it out. It uses `PACKET_ID_CUSTOM_DATA`, which is simultaneously used to broadcast to all clients.

I think we should consider reusing that packet, if possible.

qaisjp: For example, here:

https://github.com/multitheftauto/mtasa-blue/blob/7d8fd38976d6e9fc1edf767697cde2e05b937631/Server/mods/deathmatch/logic/packets/CCustomDataPacket.cpp#L36-L39

We could do something like this:

```cpp
if (m_Value.ReadFromBitStream(BitStream)) {
  return true;
} else {
  m_bDelete = true;
  return true;
}
```

qaisjp (changes requested):
Left some feedback about how we can preserve ordering + reuse the same packet, instead of using a new packet.

qaisjp:
Sorry! I know it can be frustrating to go through multiple rounds of review :(

ArranTuna:
Made it so the new RPC isn't needed, checked for compliance with coding guidelines, and re-tested.

Issue #678, opened 2018-10-29
onClientElementEnterScreen


ArranTuna:
**Is your feature request related to a problem? Please describe.**
Trying to make scripts more efficient, such as only applying a shader on an element when it comes on screen.

**Describe the solution you'd like**
An event that is triggered when an element comes on screen.

**Describe alternatives you've considered**
Which would save having to check isElementOnScreen hundreds of times every frame which is so inefficient that any savings made are lost in this constant checking.

**Additional context**
I guess onClientElementExitScreen could be useful for some too.

ArranTuna:
getElementsByType("player", root, "onScreen") would be useful for some scripts.

Pirulax:
> getElementsByType("player", root, "onScreen") would be useful for some scripts. 

For sure, we can do that.

PR #4746, opened 2026-03-07
Add on screen only option to getElementsByType

ArranTuna:
#### Summary
I suggested onClientElementEnterScreen https://github.com/multitheftauto/mtasa-blue/issues/678 but in the comments suggested that an option in getElementsByType to return on screen only would be good.

#### Motivation
There are scripts that try to be more efficient by applying things to only elements that are on screen, for example a script that renders text above a players head currently would have to do getElementsByType("player", root, true) and call isElementOnScreen on that list, as this is being done every frame that's inefficient. This change allows you to just do: getElementsByType("player", root, true, true) to get only on screen elements.

#### Test plan
This is an easy to test change, at first I just compared the sizes of #getElementsByType("ped", root, true) and #getElementsByType("ped", root, true, true) while turning camera to see it worked correctly, I then used createBlipAttachedTo to make sure the actually on screen elements were the ones in the table.


Dutchman101:
I like the idea, ill code-review it soon

CArg22:
It just pollutes api with unnecessary parameters

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -1456,12 +1460,16 @@ void CClientEntity::GetEntitiesFromRoot(unsigned int uiTypeHash, lua_State* luaV
             // Only streamed in elements?
             if (!bStreamedIn || !pEntity->IsStreamingCompatibleClass() || reinterpret_cast<CClientStreamElement*>(pEntity)->IsStreamedIn())
             {
-                if (!pEntity->IsBeingDeleted())
+                bool bVisibleOnScreen = false;
+                if (!bOnScreenOnly || (CStaticFunctionDefinitions::IsElementOnScreen(*pEntity, bVisibleOnScreen) && bVisibleOnScreen))
```
FileEX: You should first check whether the entity is being deleted to avoid unnecessary calculations for entities that will be skipped anyway because they are being deleted.

FileEX (changes requested):

Client/mods/deathmatch/logic/CClientEntity.cpp
ArranTuna: Fixed. Note that clang format was ran, but still says it failed.

FileEX (dismissed):

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -1015,18 +1015,22 @@ void CClientEntity::FindAllChildrenByTypeIndex(unsigned int uiTypeHash, lua_Stat
         // Only streamed in elements?
         if (!bStreamedIn || !IsStreamingCompatibleClass() || reinterpret_cast<CClientStreamElement*>(this)->IsStreamedIn())
         {
-            // Add it to the table
-            lua_pushnumber(luaVM, ++uiIndex);
-            lua_pushelement(luaVM, this);
-            lua_settable(luaVM, -3);
+            bool bVisibleOnScreen = false;
+            if (!bOnScreenOnly || (CStaticFunctionDefinitions::IsElementOnScreen(*this, bVisibleOnScreen) && bVisibleOnScreen))
```
tederis: Can be reduced to
`if (!bOnScreenOnly || IsOnScreen())`

Client/mods/deathmatch/logic/CClientEntity.cpp
```diff
@@ -1453,15 +1457,20 @@ void CClientEntity::GetEntitiesFromRoot(unsigned int uiTypeHash, lua_State* luaV
...
+                // Only streamed in elements?
+                if (!bStreamedIn || !pEntity->IsStreamingCompatibleClass() || reinterpret_cast<CClientStreamElement*>(pEntity)->IsStreamedIn())
                 {
-                    // Add it to the table
-                    lua_pushnumber(luaVM, ++uiIndex);
-                    lua_pushelement(luaVM, pEntity);
-                    lua_settable(luaVM, -3);
+                    // Only on screen elements?
+                    bool bVisibleOnScreen = false;
+                    if (!bOnScreenOnly || (CStaticFunctionDefinitions::IsElementOnScreen(*pEntity, bVisibleOnScreen) && bVisibleOnScreen))
```
tederis: The same. Please, use `IsOnScreen` method

tederis (changes requested):

ArranTuna:
Fixed and re-tested.

tederis (approved):

FileEX (approved):

qaisjp:
I wonder if we should instead have a new function that accepts a filter spec as the first parameter 

```lua
getElements({searchRoot = root, filterScreen = true, filterStreamedIn = true, filterTypes = ["vehicle", "player"]})
```

qaisjp:
That way you can set `filterScreen = false` to find elements off screen, `filterStreamedIn = false` to find elements that are not streamed in, and this can be easily expanded to future use cases that people care about.

qaisjp (changes requested):
The current implementation of this pull request treats `false` as a free variable, when we should be treating `nil` as the free variable, and `false` as a fixed variable.

That way we don't need to add an extra "off screen only" parameter in the future.

I'm curious what you think about my proposal for a more general `getElements` function too.

qaisjp:
I've started a discussion here: https://github.com/multitheftauto/mtasa-blue/discussions/4797

Lets discuss there how this API could look like and what the potential issues could be

ArranTuna:
Guess I'll close this then.

PR #4751, opened 2026-03-09
Fix objects with alpha below 141 are invisible

### LINKED ISSUE #425: Objects with alpha below 141 are invisible
**Describe the bug**
Devan_LT said:
> setElementAlpha on objects with the alpha set to 140 or less will cause the object to be invisible.

**To Reproduce**
```
run o = Object(1337, me.position)
run -- walk out of obj, look at obj
run o.alpha = 140 -- obj now invisible
```

**Expected behavior**
It should not be invisible.

**Additional context**
TheGTA said:
> The problem was that MTA did not adjust render states (ALPHAREF stayed 100). I fixed this alpha issue for vehicles and objects, currently eir fork only, by rewriting RenderEntity.

From https://bugs.mtasa.com/view.php?id=5623

FileEX:
#### Summary
This PR fixes a bug where objects with alpha < 141 are not rendered.

This issue has two stages: the first is the alpha threshold (`rwRENDERSTATEALPHATESTFUNCTIONREF`) set to 140 in the function `CRenderer::RenderEverythingBarRoads` - this is the direct cause of #425. The next stage is in the function `CVisibilityPlugins::RenderEntity`, where the alpha threshold is set to 100, and then objects with alpha < 101 are not rendered.

## Is it safe to just set this to 0?
These settings were R*’s attempts to fix the halo effect around trees and bushes, which, as can be seen in the game, did not work. 
```cpp
            //!PC - required to minimize alpha halo problems outside (bushes & trees mainly)
            if ((CGame::currArea==AREA_MAIN_MAP) && (!pModelInfo->GetDontWriteZBuffer()))
            {
                RwRenderStateSet(rwRENDERSTATEALPHATESTFUNCTIONREF, (void*)100);
            }
            else
            {
                RwRenderStateSet(rwRENDERSTATEALPHATESTFUNCTIONREF, (void*)0);
            }
```
I did not observe any new or worse rendering bugs with plants after setting these values to 0.

Alpha halo bugs "fixed" by R*:
image
image
image

#### Motivation
Fixes #425

#### Test plan
1. ``crun o = createObject(1337, me.position)``
2. ``crun o.alpha = 140``

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1588,7 +1588,7 @@ void CMultiplayerSA::InitHooks()
 
     // Fix objects with alpha below 141 are invisible (#425)
     MemPut<BYTE>(0x553AD9, 0);
-    MemPut<BYTE>(0x732C2E, 0);
+    MemPut<BYTE>(0x732C2F, 0);
 
     InitHooks_CrashFixHacks();
     InitHooks_DeviceSelection();
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1586,6 +1586,10 @@ void CMultiplayerSA::InitHooks()
     MemCpy((void*)0x53A23F, "\x33\xC0\x90\x90\x90", 5);
     MemCpy((void*)0x53A00A, "\x33\xC0\x90\x90\x90", 5);
 
+    // Fix objects with alpha below 141 are invisible (#425)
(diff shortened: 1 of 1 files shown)


Dutchman101 (approved):

Dutchman101:
This would be ok, unless we get a bunch of user reports from different trees (or tree mods, or any unchecked scenario) that there's worsened visual glitches. Time will tell and we got to pay attention.

rifleh700:
FYI
http://gta.rockstarvision.com/skygfx/skygfx.html#SAatest
https://github.com/aap/skygfx/blob/master/src/main.cpp#L1569

Issue #2485, opened 2021-12-25
Issues with player movement synchronization on high altitude locations, when bandwidth_reduction is maximum


haron4igg:
Not a critical issue, but just to keep a track of

When bandwidth_reduction set to maximum, players will experience issues with synchoriation of movement, 
on locations built high in the 'sky'. 
On video altitude is ~3500 meters.

https://youtu.be/EJlEXf97Pnc

PlatinMTA:
could this also happen in somewhere like xyz:0,5000,0?

haron4igg:
> could this also happen in somewhere like xyz:0,5000,0?

Yes, so it happens outside the ~4000 coordinate. by X, Y

haron4igg:
Some clarification about Z. Indepenently from Z coordinate value, we noticed that some small sync movement issues happen when one player stands on custom mapping object and observer is in freecam being higher then player. if observer will move down to the same Z coordinate as player - issue dissapears.  When standing on GTA Map, such issue doesen't appear.

Another observation - on test env its hard to find that issue, so looks like number of players online affects this.

Z axis limit seems to be ~3000 for this issue to become noticable when observer and player moving over the same ground.

ArranTuna:
Might not be the height but because of #4460

ArranTuna:
@haron4igg when a new nightly is built, this bug may be fixed.

Issue #4675, opened 2026-01-29
Bullet sync doesn't work when stood on a server side object


ArranTuna:
### Describe the bug

onPlayerWeaponFire does trigger, but onClientPlayerWeaponFire does not trigger, you see a player doing shoot animation but no bullets are being fired, when stood on a server side object, the bug doesn't occur on a client side object.

This could be related to https://github.com/multitheftauto/mtasa-blue/issues/4460 how when you are stood on a server side object, instead of using your game world position it uses an offset of your contact element.

### Steps to reproduce

Have 2 PCs, give the other PC a gun to shoot, have PC 1 stand on a server side object eg:
srun a = createObject(10244, -487.415, -1234.009, 38.577, 0, 52)
Have other player shoot around and many bullets will not sync.
Then do these 2 lines to replace the object with a client side copy:
crun a = createObject(10244, -487.415, -1234.009, 38.577, 0, 52)
srun destroyElement(a)
Player starts shooting.

Video proof, where I switch between a server side, then client side, then back to server:

https://cit.gg/arran/vids/bsync2.mp4

### Version

MTA:SA Server v1.6-release-23685

Client:
Multi Theft Auto v1.6-release-23776

### Additional context

These 2 commits are what changed bullet sync and is when we first started getting issues with bullet sync:
https://github.com/multitheftauto/mtasa-blue/commit/eab7ec8e6be63e71ffde704e51adca37e1383bb5
https://github.com/multitheftauto/mtasa-blue/commit/b9d5249563fab573af550beea6bf475f38d75344

2 commits have already been made to try fix issues:
https://github.com/multitheftauto/mtasa-blue/commit/ef963645c4767f80c88587ef4423a6fb9b7f3a3d
https://github.com/multitheftauto/mtasa-blue/commit/e91077ee65f8ef5f4115521f16eeda260d6385b3

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

FileEX:
```cpp
    if (start == player->m_vecPrevBulletSyncStart && end == player->m_vecPrevBulletSyncEnd)
        duplicate = true;
```

ArranTuna:
After further testing I found that it's not about moving, but about whether you're stood on a server side object, or client side / no object.

[...]

PR #4752, opened 2026-03-10
Fix #4460 and #4675

### LINKED ISSUE #4460: Bandwidth reduction zones aren't accurate when stood on a mapped object.
### Describe the bug

For a long time players have reported de-sync with players on mapped objects, I thought the isPedOnGround de-sync fix would fix it, but after testing 1.7 with someone it did not. So I tested it further using a custom build where I added this to CNetAPI::ReadPlayerPuresync

    // Trigger event with received position
    {
        CLuaArguments Arguments;
        Arguments.PushNumber(position.data.vecPosition.fX);
        Arguments.PushNumber(position.data.vecPosition.fY);
        Arguments.PushNumber(position.data.vecPosition.fZ);
        pPlayer->CallEvent("onClientPlayerReceivePureSync", Arguments, true);
    }

I then outputted these to chatbox, I was hoping to have it draw in the 3D world the updated positions, however when stood on a mapped object (I assume server side only mapped objects) instead of it using your world co-ordinates it uses an offset from the object, so that was of no use.

However the outputChatBox was useful as I noticed it was only coming in about 1 or 2 per second, despite player_sync_interval being set to 50ms. After some trial and error I discovered it was caused by bandwidth_reduction being set to "maximum" however "medium" also has some de-sync but less noticeable.

So it seems that the bandwidth reduction zone remote players are placed in, you can see the zones here: https://github.com/multitheftauto/mtasa-blue/blob/545972afef171cc3bbf993725054a09d67b648bd/Server/mods/deathmatch/logic/CBandwidthSettings.h are sometimes being placed in a higher zone than they should be. I noticed that sometimes it works fine (10 messages per second), you move a little bit and suddenly the player is only sending a few a second.

How it looks stood on a game world roof: 

https://github.com/user-attachments/assets/da41f585-612b-4a9f-8934-1e6eaa68a9d1

How it looks stood on the exact same roof but mapped:

https://github.com/user-attachments/assets/c32bbb60-0985-4f03-b95f-d2e351d791fd

Notice how at the end of the video I moved which then suddenly caused the sync packets to come in normal.

Note the difference in frequency of the sync packets.

### Steps to reproduce

Have 2 players
srun a = createObject(3624, 2424.612, -2500.337, 13.655)
crun setElementPosition(localPlayer, 2424.612, -2500.337, 20)
srun setServerConfigSetting("bandwidth_reduction", "maximum")
Watch other player move around on the roof

### Version

1.6 and 1.7 newest

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

ArranTuna:
#### Summary
Fixes:
Bullet sync doesn't work when stood on a server side object #4675
and
Bandwidth reduction zones aren't accurate when stood on a mapped object. #4460

Which were both caused by the same issue. The problem was that when stood on a mapped object, MTA thought you were somewhere much further away at certain camera rotations which caused bandwidth reduction to not sync you as frequently as it should and bullets to not sync either.

#### Motivation
Both of these issues caused major player complaints which is why I eventually investigated them enough to have bugs to report and now thanks to help from codex investigated them further and developed a fix.

#### Test plan
I tested with 2 PCs, first of all verified that I could easily reproduce the bug.

Then codex added some debug outputs like so:

when shots sync:

[syncdbg] zone src=a **zone=0** route=sim dist=2999.2 dot=0.564 recvPos=(2484.81,-1665.73,15.74) cam=(4971.46,-3330.33,30.24) recvObj=130 recvRel=(0.81,-0.73,3.74) srcPos=(2477.54,-1664.47,13.33) srcObj=0 srcRel=(0.00,0.00,0.00)

when shots don't sync:

[syncdbg] zone src=a **zone=2** route=main dist=2997.2 dot=-0.122 recvPos=(2484.81,-1665.73,15.74) cam=(4970.46,-3328.33,30.24) recvObj=130 recvRel=(0.81,-0.73,3.74) srcPos=(2477.54,-1664.53,13.33) srcObj=0 srcRel=(0.00,0.00,0.00)

Then a bug fix was developed and I have tested it with 2 PCs and all bullets are syncing correctly regardless of camera rotation.

Also I tested just walking around the same as I did in the video on the bug reports and sync is now flawless with bandwidth reduction on maximum.

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
@@ -81,6 +81,7 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
         // Player position
         SPositionSync position(false);
         bool          positionRead = BitStream.Read(&position);
+        const CVector vecRelativePosition = position.data.vecPosition;
 
         if (positionRead && pContactElement != nullptr)
         {
@@ -204,7 +205,10 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
 
         // Read the camera orientation
         CVector vecCamPosition, vecCamFwd;
-        ReadCameraOrientation(position.data.vecPosition, BitStream, vecCamPosition, vecCamFwd);
+        // Camera orientation is encoded against the same position basis the client wrote.
+        CVector vecCameraBasePosition = pContactElement ? vecRelativePosition : position.data.vecPosition;
+
+        ReadCameraOrientation(vecCameraBasePosition, BitStream, vecCamPosition, vecCamFwd);
         pSourcePlayer->SetCameraOrientation(vecCamPosition, vecCamFwd);
 
         if (flags.data.bHasAWeapon)


ArranTuna:
In fact I reckon this will also fix #2485

MegadreamsBE (approved):
Thank you!

PR #4754, opened 2026-03-11
Fix #4233 isMTAWindowFocused() being unreliable

### LINKED ISSUE #4233: isMTAWindowFocused() returns false if you join a server while minimized

ArranTuna:
#### Summary
If you launch MTA using the mtasa:// URI to automatically join a server, isMTAWindowFocused() returns false even when you're maximized, because m_bFocused can't be relied upon for this function.

#### Motivation
Because this bug affects me every day, because I launch MTA with mtasa:// so it's very annoying that once I login to the server I have to minimize and maximize my game to get this function to return the correct value, I use this function to save players CPU usage when they're minimized to not render stuff.

#### Test plan
Tested with and without launching via mtasa://
Made sure it does return false when minimized using a timer.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -6044,6 +6044,21 @@ bool CClientGame::OnFocusLoss(CGUIFocusEventArgs Args)
     return true;
 }
 
+bool CClientGame::IsWindowFocused() const
+{
+    HWND foregroundWindow = GetForegroundWindow();
+    if (!foregroundWindow)
+        return false;
+
+    HWND hookedWindow = g_pCore->GetHookedWindow();
+    if (foregroundWindow == hookedWindow || GetAncestor(foregroundWindow, GA_ROOTOWNER) == hookedWindow)
+        return true;
+
+    std::uint32_t processId = 0;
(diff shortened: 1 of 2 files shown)


FileEX:
https://github.com/multitheftauto/mtasa-blue/pull/4533

ArranTuna:
Damn... I've tested your PR and it works.

PR #4766, opened 2026-03-26
Expand vehicle variants

Xenius97:
#### Summary
This PR expands the possibilities for vehicle variants, increasing the available base limit from 6 to 253.
(254 = random default variant, 255 = reset variants)
Extras are indexed starting from 1, so for example, extra40 will be variant 39.
By default, GTA:SA can handle extras from `extra1` to `extra5`. With this update, modelers can now use `extra6` through `extra253`, and MTA will handle these natively. These components should be added to `chassis_dummy` like other variants.

#### Motivation
Currently, vehicle variants are quite limited by default. We use many vehicles where multiple types of the same model are available as separate variants, but often the default 6 variants are not enough, forcing us to rely on custom element data. This solution is much more performance-oriented and, incidentally, can natively handle 253 extra components without scripts.
Furthermore, in our case, there is no need to rely on custom element data, as the built-in synchronization manages this automatically.

#### Test plan
Start the attached resource and run `setVehicleVariant(vehicle, 99, 39)`
[replace.zip](https://github.com/user-attachments/files/26265614/replace.zip)

#### Checklist


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2940,10 +2940,32 @@ void CClientVehicle::Create()
...
-                    // apple our new scale
+                    // apply our new scale
                     SetComponentScale(strTemp, (*iter).second.m_vecComponentScale);
                 }
             }
+            // Handle custom variation components. These are components that are named "extraX" where X is the variation number.
+            // We hide all of them by default and only show the ones that match the current variation.
+            if (strTemp.rfind("extra", 0) == 0)
+            {
+                unsigned int id = atoi(strTemp.substr(5).c_str());
```
FileEX: Please use std::atoi or std::from_chars

FileEX (changes requested):

Client/mods/deathmatch/logic/CClientVehicle.cpp
tederis: `std::stoi` would fit better because it accepts `std::string` directly.

Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -2940,10 +2940,32 @@ void CClientVehicle::Create()
...
+
+                // Hide extras > 6 by default
+                if (id > 6)
+                {
+                    (*iter).second.m_bVisible = false;
```
tederis: Please, give `*iter` a name via ref

tederis (approved):

PR #4771, opened 2026-04-01
Fix isElementOnScreen for markers

### LINKED ISSUE #2029: isElementOnScreen() does not work for markers
**Describe the bug**
Like said in https://bugs.mtasa.com/view.php?id=3510, isElementOnScreen() will never return true for markers, even if they are on the screen.

**Expected behaviour**
isElementOnScreen() should work with markers.

**Version**
Every version

**Additional context**
If anyone wants to fix this, it would be cool if you could check out this related issue as well #678

FileEX:
#### Summary
This PR adds the ability to use the `isElementOnScreen` function for markers.

### Known issues
When rendering 3D markers, GTA checks visibility using a fixed radius of 2.0. This causes a bug where large markers disappear/flicker when moving the camera.

A second bug concerns this PR itself - `isElementOnScreen` returns true even when the marker is not visible but is close to the edge of the camera. The reason is simple: the marker is not visible, but GTA still renders it because the camera is within the 2-unit radius. This can be fixed in a future PR.

#### Motivation
Fixes #2029

#### Test plan
```lua
local m = createMarker(x, y, z + 1, 'checkpoint', 1.0, 255, 0, 0, 255);

    dxDrawText('Marker visible: '..(isElementOnScreen(m) and 'true' or 'false'), 300, 300, 0, 0, 0xFFFFFFFF, 1, 'default-bold');
end);
```

#### Checklist


ArranTuna:
I have tested this with all 5 marker types and it works.

PR #4776, opened 2026-04-05
Run clang-format

qaisjp:


qaisjp:
cc @botder

The clang-format script wouldn't finish on my machine, I had to revert your most recent commit to the script (5b11c3810ec48995ae0e37ed2e177e70be51766b), run the script, and then unrevert

Do you have more details on what was broken so that I can make sure that when I try to fix the issue, I don't cause a regression?

PR #4781, opened 2026-04-05
Add ability for the new parser to return nil on invalid arguments

FileEX:
#### Summary
This PR adds the ability for the new parser to return nil on invalid arguments.

#### Motivation
Some functions that return a boolean will return nil for invalid arguments. For example, `isElementOnScreen` returns false if the element is not on the screen, but if no argument is provided, it returns nil. The new parser also needs this ability in order to fully maintain backward compatibility with old functions.

#### Test plan
```cpp
        {"isElementOnScreen", ArgumentParserWarn<nullptr, IsElementOnScreen>},
```

#### Checklist


qaisjp (dismissed):
Is the first type parameter of ArgumentParserWarn the value to return? And it just didn't support nil? Surprised we never supported nil after all this time haha.

FileEX:
It looks like ``lua::Push`` does handle nullptr. I don’t know why it didn’t work for me before.

PR #4784, opened 2026-04-09
fix: dxDrawModel3D flickering issue

### LINKED ISSUE #4596: dxDrawModel3D flickering
### Describe the bug

DxDrawModel3D is not working properly anymore since version 23625 -> Flickering all over the place when being called

### Steps to reproduce

1. Use dxDrawModel3D in conjunction with some default SA object
2. Start up resource see flickering

### Version

Multi Theft Auto v1.6-release-23625

### Additional context

https://www.youtube.com/watch?v=cB1Q5YWyZSg

### Security Policy

naouh:
#### Summary
Fix `dxDrawModel3D` flickering when rendering a model

#### Motivation
Commit `e5e308de1` ("GTA streamer bug-fix round #1") added `ModelAddRef`/`RemoveRef` inside `RenderModel()` to protect the `RwObject` from being GC'd during rendering. The problem is that `RemoveRef` does more than just decrement a counter, when MTA's `m_dwReferences` hits 0, it calls `Remove()`, which decrements GTA's `usNumberOfRefs` and if that also reaches 0, fires `RemoveModel()` and unloads the model.

With `dxDrawModel3D`, no persistent MTA-side reference is held, so every single frame:
1. `ModelAddRef` bumps refs 0 → 1
2. Model renders
3. `RemoveRef` drops refs 1 → 0 → `Remove()` → model unloaded
4. Next frame: model gone, `EnqueueModel` fails, nothing renders
5. `engineStreamingRequestModel` re-requests it, model reloads
6. Next frame: renders, then unloads again → repeat

Hence the flickering every other frame.

The fix replaces `ModelAddRef`/`RemoveRef` with a direct increment/decrement of GTA's `usNumberOfRefs`, which still blocks the streaming GC from freeing the model mid-render without triggering MTA's unload logic on release.

#### Test plan
Test with the Maybe v4 resource, objects that were flickering should now render stable.

[[DM]_Maybe_v4_-_Saints_Boulevard.zip](https://github.com/user-attachments/files/26604757/DM._Maybe_v4_-_Saints_Boulevard.zip)

#### Checklist


Client/game_sa/CRendererSA.cpp
```diff
@@ -32,25 +32,16 @@ void CRendererSA::RenderModel(CModelInfo* pModelInfo, const CMatrix& matrix, flo
...
-        pModelInfo->RemoveRef();
         return;
-    }
+
+    // Prevent GC from freeing RwObject during rendering by bumping GTA's own
+    // ref count directly.  We must NOT use ModelAddRef/RemoveRef here because
+    // RemoveRef triggers Remove() → RemoveModel() when the MTA ref count drops
+    // to 0, which unloads the model every frame for dxDrawModel3D models that
+    // have no persistent MTA-side reference -- causing visible flickering.
+    pModelInfoSAInterface->usNumberOfRefs++;
```
FileEX: You should call the AddRef/RemoveRef method from CBaseModelInfo to prevent the TXD from being unloaded. This function modifies the reference count for the model as well as for its textures.

FileEX (changes requested):

Client/game_sa/CRendererSA.cpp
```diff
@@ -32,25 +32,21 @@ void CRendererSA::RenderModel(CModelInfo* pModelInfo, const CMatrix& matrix, flo
...
+    // Prevent GC from freeing RwObject/TXD during rendering by calling GTA's
+    // native CBaseModelInfo::AddRef/RemoveRef.  These manage usNumberOfRefs and
+    // texture dictionary refs without going through MTA's ref-counting (which
+    // triggers Remove() when m_dwReferences drops to 0, unloading the model
+    // every frame for dxDrawModel3D models with no persistent MTA-side reference
+    // -- causing visible flickering).
+    using BaseModelInfoMethod = void(__thiscall*)(CBaseModelInfoSAInterface*);
+    auto CBaseModelInfo_AddRef = reinterpret_cast<BaseModelInfoMethod>(FUNC_AddRef);
+    auto CBaseModelInfo_RemoveRef = reinterpret_cast<BaseModelInfoMethod>(FUNC_RemoveRef);
+
```
FileEX: It would be better if you simply add these methods to the CBaseModelInfoSAInterface class, and then just do it here
```cpp
pModelInfoSAInterface->AddRef();
pModelInfoSAInterface->RemoveRef();
```
Example: https://github.com/multitheftauto/mtasa-blue/blob/53521fa8382632e8010ff1910724ff856ef2028f/Client/game_sa/CAutomobileSA.h#L40-L48

FileEX (changes requested):

FileEX (approved):

Lpsd:
Thank you, will be implemented in #4807

PR #4786, opened 2026-04-09
Add engineSetRenderingListSize

### LINKED ISSUE #4755: Undesired quantity limit for dxDrawModel3D when alphaTransparency is enabled on the model.
### Describe the bug

When a model has the alphaTransparency property enabled, the number of displayed objects becomes limited (~200). The issue does not occur when alphaTransparency is set to false.
This applies to both default models and replaced ones.

This is mainly a problem with all types of grasses and plants.

### Steps to reproduce

1. Run resource [test_grass.zip](https://github.com/user-attachments/files/25943224/test_grass.zip)
2. Look at the grass on the farm
3. Switch 'alphaTransparency' var to true in c.lua
4. Restart resource

...Or

Run code:

```
startPosition = Vector3(-1193.58, -1058.74, 128)
engineStreamingRequestModel(806,true,true) -- Plant with alpha
engineStreamingRequestModel(1337,true,true) -- Bin without alpha
addEventHandler('onClientRender',root,function()
    for x=0,50 do
        for y=0,50 do
            dxDrawModel3D(806, startPosition.x + x, startPosition.y + y, startPosition.z, 0,0,0, 1,1,1,1)
        end
    end
    for x=50,100 do
        for y=0,50 do
            dxDrawModel3D(1337, startPosition.x + x, startPosition.y + y, startPosition.z+1, 0,0,0, 1,1,1,1)
        end
    end
end)
```
Look at the number of objects.

### Version

Multi Theft Auto v1.6-release-23953

### Additional context

Two model combinations (same quantity)
Image

Grass model when alphaTransparency is false.
Image

alphaTransarency is true
Image

### Relevant log output

```shell

```

### Security Policy

FileEX:
#### Summary
By default, render lists have a fixed size, and for alphaEntitiesList it is 200 elements, and for underwater it is 100. These lists can be safely modified, so this PR adds the ability to change the size of such a list in order to avoid the issue encountered by the author of the referenced issue if necessary. If more objects are rendered than the list capacity, the oldest ones disappear randomly when moving the camera, and this is normal behavior.

```lua
nil engineSetRenderingListSize(string list, int count)
```
- **list** - The list to be modified. It can be `alpha` or `underwater`.
- **count** - The new size for the list. It cannot be smaller than the default size.

```lua
engineSetRenderingListSize("alpha", 1000)
```

image

## Issues
For a large number of entities inside the list rendered simultaneously (> 3500), there are FPS drops. This is caused by O(n²) lists and may be fixed in a future PR.

#### Motivation
Fixes #4755 

#### Test plan
Resource attached to the referenced issue.

#### Checklist


Fernando-A-Rocha:
Wow so this fixes the limit of rendered objects that start disappearing when too many near you??!
It's a common issue when you load a map made for SAMP which sometimes has a huge amount of objects for decoration

FileEX:
> Wow so this fixes the limit of rendered objects that start disappearing when too many near you??! It's a common issue when you load a map made for SAMP which sometimes has a huge amount of objects for decoration

Probably yes. Underwater entities and alpha entities are limited. Opaque entities have no quantity limit.

tederis (dismissed):

PR #4793, opened 2026-04-12
Drop asm usages in game_sa

sbx320:
Some initial work on automatically rewriting game_sa's inline asm to proper C++ calls via a custom clang-tidy plugin. 

See https://github.com/sbx320/mtaasmrewriter for source

Rough approach is: 

- Find all `__asm` statements
- Collect all `push` instructions (arguments)
- Look for a `mov ecx, [something]` instruction (__thiscall indicator)
- Look for a `mov [something], [something]` instruction (return value)

Then rewrite the statement to a function type declaration and a function call. 

Current issues:

- Does not support other asm instructions (e.g. `lea` is used in some places)
- ~~Does not properly escape prefix-less hexadecimal numbers (e.g. `push 0B73710h` is mapped to a function argument `0B73710h` which is a base-2 number~~
- If multiple function calls are used, the type declaration collides due to duplicate names
- clang formatting is disabled
- needs to be on master, not on 1.6.0


sbx320:
Turns out github apparently broke branch retargeting... https://github.com/multitheftauto/mtasa-blue/pull/4794 for doing things to master

PR #4798, opened 2026-04-12
Guard recursive Lua table bitstream reads

Zephkek:
## Summary
- add a 64-level recursion limit when reading nested Lua tables from bitstreams
- apply the guard on both client and server deserialization paths
- preserve existing call sites by threading depth through defaulted parameters

## Why
The public source currently allows unbounded recursive LUA_TTABLE reads from the bitstream serializer. A crafted packet can drive recursive descent until deserialization fails from stack exhaustion or excessive recursion.

Current shipped binaries already enforce a 64-level limit in this path. This change brings the public source in line with that behavior so downstream forks and public builds do not keep the unbounded recursion path.

## Testing
- validated the touched files with editor diagnostics
- manually reviewed the nested-table bitstream read path to confirm inputs deeper than 64 levels now fail deserialization
- reviewed the patch scope to confirm it only changes the bitstream table read path


qaisjp (dismissed):
Nice find, thanks!

qaisjp (approved):

Issue #3569, opened 2024-07-15
Fix weapon switching (scroll and Q+E) while using Jetpack


Dutchman101:
### Describe the bug

As the title says; you can't switch weapons as expected when flying a Jetpack.
If you have multiple weapons in inventory, it should work as it does on foot.

Of course, it should only switch to weapons allowed in Jetpack by default, or otherwise whitelisted through [setJetpackWeaponEnabled](https://wiki.multitheftauto.com/wiki/SetJetpackWeaponEnabled)

### Steps to reproduce

1. Give yourself jetpack
2. Give yourself multiple weapons that are allowed in Jetpack (by default, or modified by `SetJetpackWeaponEnabled`
3. Try to switch weapons; you can't. The only way to equip another weapon is typically to spawn it again.

### Version

Client r22616
Server r22616

Dutchman101:
Fixed in 180fbc0

Issue #508, opened 2018-09-20
Texture bug with setJetpackWeaponEnabled


ArranTuna:
**Describe the bug**
If you use setJetpackWeaponEnabled to make weapons usable on jetpack and then wield them on jetpack, the weapon can become invisible. There is only one way to circumvent this, explained in steps to reproduce.

**To Reproduce**
The bug occurs when you set an jetpack while the weapon you're using is already equipped.

```lua
-- We will enable the MP5 that's not jetpack-usable by default, for reproduction.
addEventHandler("onResourceStart", resourceRoot, function()
     setJetpackWeaponEnabled("29", true)
end)
```

1) run above script
2) wield MP5
3) set yourself an jetpack

Now the muzzle flash is the only thing seen because the weapon texture misses and your weapon becomes invisible, both to yourself and other players.

Circumventing it and getting the weapon visible on jetpack requires you to first get the jetpack, and spawning the weapon after that. However, you must atleast then not have the weapon already (invisibly) equipped as it won't change then, and have either no weapon equipped or one different from the jetpack-enabled weapon that would otherwise bug.

**Expected behavior**

**Screenshots**
![image](https://user-images.githubusercontent.com/923242/45840300-7a121f80-bd0e-11e8-9512-9fcc003cc892.png)

**Additional context**
From https://bugs.mtasa.com/view.php?id=8827

PR #4804, opened 2026-04-13
Fix jetpack behavior

FileEX:
#### Summary
In GTA, when you equip a jetpack, the weapon is automatically switched to a pistol, tec-9, or uzi if you have one of these weapons. Otherwise, the weapon is hidden from the player’s hands, but the slot remains unchanged, which causes two issues:
- Issue #508, where weapons are invisible
- The state described in PR #3573 - if you pick up a jetpack while holding a e.g chainsaw/molotov, the weapon model disappears, but the weapon sound and its HUD icon remain

In GTA, when you remove the jetpack, the slot is restored to the one you had before equipping it. For example, if you are holding an AK-47, then equip a jetpack and it switches to an uzi, after removing the jetpack you get the AK-47 back. PR #3573 changed this behavior so that after removing the jetpack, the slot is always reset to 0.

From the GTA code, it appears that Rockstar also intended to include the sawed-off shotgun for jetpack use, but a trivial mistake prevented it from working.
```cpp
			// this is to let the player use other 1 handed weapons when flying a jetpack!
			else if(nVehicleType==1 && m_WeaponSlots[WEAPONSLOT_TYPE_SHOTGUN].m_eWeaponType==WEAPONTYPE_SAWNOFF_SHOTGUN
			&& m_WeaponSlots[WEAPONSLOT_TYPE_SHOTGUN].m_nAmmoTotal>0)
			{
				nDoDrivebySlot = WEAPONSLOT_TYPE_HANDGUN;
			}
			else if(nVehicleType==1 && m_WeaponSlots[WEAPONSLOT_TYPE_HANDGUN].m_eWeaponType==WEAPONTYPE_PISTOL
			&& m_WeaponSlots[WEAPONSLOT_TYPE_HANDGUN].m_nAmmoTotal>0)
			{
				nDoDrivebySlot = WEAPONSLOT_TYPE_HANDGUN;
			}
```

This PR resolves the issues described above.

#### Motivation
https://github.com/multitheftauto/mtasa-blue/pull/3510#issuecomment-2573837673

#### Test plan
See issues: #508, #3569
1. Get chainsaw
2. Get jetpack
3. Remove jetpack
4. The slot should be reset to the chainsaw, but it is instead reset to 0.

**TL;DR**
- Removing the jetpack restores the previous slot, exactly like in GTA:SA
- While using a jetpack, you can now also use the sawed-off shotgun in addition to pistol, Uzi, and Tec-9

#### Checklist

### DIFF
--- a/Client/game_sa/CPedSA.cpp
+++ b/Client/game_sa/CPedSA.cpp
@@ -213,8 +213,8 @@ void CPedSA::ClearWeapons()
 
 void CPedSA::RemoveWeaponModel(std::uint32_t model)
 {
-    // void __thiscall CPed::RemoveWeaponModel(CPed *this, int modelID)
-    ((void(__thiscall*)(CEntitySAInterface*, std::uint32_t))FUNC_RemoveWeaponModel)(m_pInterface, model);
+    if (auto* pedInterface = GetPedInterface())
+        pedInterface->RemoveWeaponModel(model);
 }
 
 void CPedSA::ClearWeapon(eWeaponType weaponType)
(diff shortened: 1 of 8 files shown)


Lpsd:
The proposed solution in https://github.com/multitheftauto/mtasa-blue/pull/3510#issuecomment-2573837673 sounds good - how come this wasn't implemented?

FileEX:
> The proposed solution in [#3510 (comment)](https://github.com/multitheftauto/mtasa-blue/pull/3510#issuecomment-2573837673) sounds good - how come this wasn't implemented?

Due to the ongoing feature freeze, I want to do this in a separate PR so that the bugfix one doesn’t sit indefinitely just because it adds a new feature.

PR #4814, opened 2026-04-17
Fix incorrect alpha blending in render targets

pedrohenryz:
Semi-transparent elements rendered to RTs were too faint due to alpha being
applied twice (RT + screen), resulting in incorrect opacity.

Use separate alpha blending when rendering to RTs and introduce RT_TEXTURE
blend mode (ONE, INVSRCALPHA) for correct compositing. Automatically apply
RT_TEXTURE when drawing render targets.

Files:
- CGraphicsInterface.h
- CRenderItemManager.h
- CGraphics.cpp

Before:
- Semi-transparent elements in RTs appear almost invisible
image

After:
- Correct transparency matching direct screen rendering
image 11


pedrohenryz:
Why is it marked as "backwards-incompatible"?

FileEX:
I don't fully understand the issue described here - I've never encountered it, even when using transparent dxDraw calls inside a render target (is this a new issue in 1.7?).
Also, please follow the PR template in the future.

pedrohenryz:
This issue already exists in 1.6, the “before” image is from 1.6. 

Notice that the rectangle on the left (inside the render target) is almost invisible, while the one outside of it is clearly visible. Both have the same color and transparency.

You can reproduce the problem in your MTA using this script.

```local rtSize = 300
local renderTarget = dxCreateRenderTarget(rtSize, rtSize, true)

    dxSetRenderTarget(renderTarget, true)
    dxDrawRectangle(0, 0, rtSize, rtSize, tocolor(255, 255, 255, 51)) -- 20% alpha
    dxSetRenderTarget()

    -- Draw RT
    dxDrawImage(50, 50, rtSize, rtSize, renderTarget)

    -- Direct draw
    dxDrawRectangle(400, 50, rtSize, rtSize, tocolor(255, 255, 255, 51))
end)```

FileEX:
Hmm, I'm not sure this is the correct way to draw render targets. The appropriate blend mode should always be set https://wiki.multitheftauto.com/wiki/DxSetBlendMode
Won't this PR cause issues with scripts that currently use dxSetBlendMode as shown in the wiki examples?

pedrohenryz:
I need to run specific tests. But I don’t think `dxSetBlendMode` will solve the issue described. The main problem is that the render target changes how the transparency of the rectangle rendered inside it behaves.
The render target blends the color of the rectangle with its transparent background, which results in reduced alpha.

pedrohenryz:
Before
rtissues

After
rtissues_fix

I noticed that the render target also affects text quality. My fix corrects that

syncc01:
amazing🚀

FileEX:
Before
image

After
image

Tested, works. Thank you!

FileEX (approved):

Issue #3464, opened 2024-06-12
The getPedAnimation function doesn't work in IFP.


W3lac3:
### Describe the bug

When applying an IFP animation to the player, the getPedAnimation function cannot return the animation that the player is using. It will always return an animation from the game, which is run_wuzi. The same does not happen with in-game animations, only in the case of IFP.

### Steps to reproduce

1. Load an IFP animation;
2. Apply this animation using setPedAnimation on the client-side;
3. Use the getPedAnimation function and you'll see the bug happen.

### Version

Client: v1.6-release-22512 (Windows 11, 64-bit)
Server: v1.6-release-22512 (Windows 11, 64-bit)

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

nekjutsu:
Issue still exists in version 25994+

UPD:
`getPedAnimationProgress`
`getPedAnimationLength`

`setPedAnimationSpeed`
`setPedAnimationProgress`

These functions also doesn't work for custom animations

PR #4816, opened 2026-04-17
Fix client-sided animation progress

FileEX:
#### Summary
Because `CStaticFunctionDefinitions::SetPedAnimation` returned true immediately after setting the animation, startTime was never set to a timestamp. As a result, it always remained 0, and progress was calculated as timestamp / 1000, producing large values (clamped to 1).

#### Test plan
```
crun ped = createPed(50, getElementPosition(me)) setPedAnimation(ped, "ped", "ko_shot_face", -1, false, false, false, true)
```

#### Checklist


nekjutsu:
@FileEX Could you also please fix #3464?

PR #4818, opened 2026-04-18
Fix follow-car camera NaN instability

### LINKED ISSUE #3979: Game freezes in recent updates
### Describe the bug

Some players have issue with recent updates, i guess started with r22880, when driving/spectating a vehicle/player sometimes game freezes for a sec and a lot happens within this second, camera get reset, vehicles can go through objects, game brightness increased for a sec

i have no idea what causes this issue, im not affected, tried both stable and nightly and couldn't reproduce this bug

### Steps to reproduce

no idea

### Version

Multi Theft Auto v1.6-release-22894

### Additional context

Videos from players affected by this bug:
https://www.youtube.com/watch?v=WyHvCtFeumY
https://streamable.com/dfxeib
https://streamable.com/xoyg9j
https://streamable.com/dlyzy0
https://youtu.be/cuyc6OH-SQU?t=113

### Relevant log output

```shell

```

### Security Policy

Lpsd:
#### Summary
This change hardens the vehicle follow-camera path against non-finite camera angle state that can occur during `Process_FollowCar` and cause brief camera flash/fall-through behavior.

The patch:
- Detects non-finite `beta`, `betaSpeed`, and `alpha` before look-direction reconstruction.
- Recovers to safe finite values instead of allowing NaNs to propagate into the camera direction vector.
- Preserves live `beta` when only `betaSpeed` is invalid (to avoid input snapping/jitter).
- Resets poisoned `betaSpeed` to `0.0f`.
- Rewrites target beta only when `beta` itself is invalid.
- Keeps the active camera mode unchanged (no mode-switch workaround).

#### Motivation
Resolves #3979.

On Windows 11, we observed intermittent follow-camera instability where the camera could flash/fall through. Investigation provided strong evidence that non-finite angle values were entering the native follow-camera flow and then being propagated into look-direction math.

We did not fully isolate a single native root-cause instruction with high confidence, but we narrowed the failure mode enough to apply a targeted mitigation at the point where invalid values become user-visible. This patch addresses the observed issue while preserving expected camera behavior and controls.

#### Test plan
Manual validation:
- Reproduced the scenario from #3979 in vehicle follow-camera conditions on Windows 11.
- Verified camera no longer flashes/falls through with this patch.
- Verified normal follow-camera behavior remains intact (no forced camera mode changes and no persistent control jitter after refinement).
- Re-ran camera interaction patterns that previously triggered the issue and confirmed stable behavior.

Regression checks:
- Confirmed mitigation path activates only when non-finite angle state is detected.
- Confirmed finite-state behavior follows the normal path.

#### Checklist

#### Note

The underlying root cause is still being investigated -- this workaround allows us to continue with the 1.7 release in the meantime.

Also includes some unrelated crash fix cherry-picks that were missing from the revert


Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -16,6 +16,8 @@
 #include <game/CPedDamageResponse.h>
 #include <game/CEventList.h>
 #include <game/CEventDamage.h>
+#include "../game_sa/CCamSA.h"
```
qaisjp, line 19: Not sure if we should be adding game_sa import to multiplayer_sa

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -806,7 +808,7 @@ void CMultiplayerSA::InitHooks()
     // Disable CPopulation::RemovePed
     MemPut<BYTE>(0x610F20, 0xC3);
 
-    // Temporary hack for disabling hand up
+    // Legacy disabled hand-up patch retained as commented reference.
```
qaisjp, line 811: What is hand up?

Client/game_sa/CHandlingEntrySA.cpp
```diff
@@ -83,24 +83,32 @@ void CHandlingEntrySA::SetSuspensionHighSpeedDamping(float fDamping) noexcept
 
 void CHandlingEntrySA::SetSuspensionUpperLimit(float fUpperLimit) noexcept
 {
+    if (!std::isfinite(fUpperLimit)) [[unlikely]]
+        return;
```
qaisjp, line 87: There's quite a few of these so I'd be weary to make sure this doesn't cause performance issues

Client/game_sa/CVehicleSA.cpp
```diff
@@ -1913,7 +1925,14 @@ void CVehicleSA::RecalculateSuspensionLines()
...
     DWORD       dwModel = GetModelIndex();
     CModelInfo* pModelInfo = pGame->GetModelInfo(dwModel);
-    if (pModelInfo && pModelInfo->IsMonsterTruck() || pModelInfo->IsCar())
+    if (!pModelInfo)
+        return;
+
+    // Parenthesise the precedence to match the original intent: only proceed
+    // for monster trucks or cars. The bareword `pModelInfo && X || Y` parsed
+    // as `(pModelInfo && X) || Y`, which let a null pModelInfo fall through
+    // to the IsCar() deref below on pure-car paths.
```
qaisjp, line 1934: nit: the extra paras here seems moot now that we put the guard earlier?

qaisjp (changes requested):

qaisjp:
Wow it let.me request changes post merge :laughing:

Client/multiplayer_sa/CMultiplayerSA.cpp
Lpsd: massive brainfart

Client/game_sa/CHandlingEntrySA.cpp
qaisjp: If we figure out the root cause we should consider removing these, or wrapping in compile time checks so we can measure time difference maybe? I'm really not sure

Client/multiplayer_sa/CMultiplayerSA.cpp
Lpsd: Some value in the Process_FollowCar_SA function that I can't even remember what it does now -- it might not even be in Process_FollowCar_SA that's how much I remember about it. This is a leftover comment change from when I was investigating that angle.

Client/multiplayer_sa/CMultiplayerSA.cpp
Lpsd: this was from testing/debugging btw, cleared up now

Client/game_sa/CHandlingEntrySA.cpp
Lpsd: Technically these aren't even necessary they're remnants of my investigation, I just assumed they wouldn't be that harmful - but if performance is a concern they can be removed at any time

qaisjp:
(I didn't genuinely want to request changes, just was testing to see if it'd let me!)

PR #4837, opened 2026-04-20
Protocol error fixes

### LINKED ISSUE #4832: 1.7 protocol error 60 on join
### Describe the bug

I tested newest 1.7 64 bit server (mtasa_x64-1.7-untested-26530-20260415) and I got protocol error 60 when joining the server (tried with the newest nightly (mtasa-1.7-untested-26546-20260419) and a matching nightly (mtasa-1.7-untested-26530-20260415) both fail.

I then tried with no resources, and it works fine, started starting resources to see which causes it and it seems that if in-game when starting, you don't get the error, I reconnected and got the error.

I then tried constantly reconnecting to see when I'd start get protocol error during resource starts. When I got the error, I stopped the starting of more resources, started stopping recently started resources, still couldn't join despite stopping all the resources that had been started since my last successful join, so I've no idea.

### Steps to reproduce

Unknown, if any devs would benefit from joining the server to get the protocol error I can tell you the IP/port/password.

### Version

Client: v1.7-untested-26530
Server: v1.7-untested-26530 64 bit windows (also tested 32 bit and same error)

### Additional context

Image

### Relevant log output

```shell

```

### Security Policy

NanoBob:
#### Summary
- Fix ped animation protocol error
  - Animation start time was changed to be sent as timedelta (to avoid casting uint64 to uint32), but the client side counterpart was never changed. (Fixes #4832)
- Fix picking health and armour
  - These were being sent with wrong precision, this wouldn't cause crashes but was a mismatch.
- Fix building protocol error
  - Interior was being sent a second time for buildings, but not being read.
- Fix building memory leak
  - Building would not be destroyed on resource end
- Fix buildings not appearing on startup
  - Buildings were not being sent in the initial entity add packet

### DIFF
--- a/Client/mods/deathmatch/logic/CPacketHandler.cpp
+++ b/Client/mods/deathmatch/logic/CPacketHandler.cpp
@@ -4198,8 +4198,6 @@ retry:
                     if (!CClientBuildingManager::IsValidModel(modelId))
                         modelId = 1700;
 
-                    bitStream.ReadCompressed(ucInterior);
-
                     bitStream.Read(LowLodObjectID);
                     CClientBuilding* pBuilding = new CClientBuilding(g_pClientGame->m_pManager, EntityID, modelId, position.data.vecPosition,
                                                                      rotationRadians.data.vecRotation, ucInterior);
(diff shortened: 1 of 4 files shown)


Lpsd (approved):
Thank you, LGTM! :shipit:

PR #4843, opened 2026-04-23
use correct map keys in SetResponseCookie

### LINKED ISSUE #4806: httpSetResponseCookie is broken
### Describe the bug

`httpSetResponseCookie` is not working - it doesn’t set the cookie. `document.cookie` is empty, even though the function returns true.

``cookies`` variable returns empty table.

### Steps to reproduce

Try example from https://wiki.multitheftauto.com/wiki/HttpSetResponseCookie

Or use this
```html
<*
    local timestamp = getElementData(getPlayerFromName(getAccountName(user)), 'loginTime') or ''
    local date = os.date("%Y-%m-%d %H:%M:%S", timestamp);

    print(httpSetResponseCookie("login_time", tostring(date)))
*>

<!DOCTYPE html>
<html>
    <head>
        <title>Hello</title>
    </head>

    <body onLoad="getLoginTime();">
        <p>Logged into the server: <span id='time_span'></span></p>

        <script>
        // getCookie function from w3schools.com
        function getCookie(cname) {
          let name = cname + "=";
          let decodedCookie = decodeURIComponent(document.cookie);
          let ca = decodedCookie.split(';');
          for(let i = 0; i <ca.length; i++) {
            let c = ca[i];
            while (c.charAt(0) == ' ') {
              c = c.substring(1);
            }
            if (c.indexOf(name) == 0) {
              return c.substring(name.length, c.length);
            }
          }
          return "";
        }

        function getLoginTime()
        {
            var time = document.getElementById('time_span');
            var login_time = getCookie('login_time');

            time.innerHTML = (!login_time || login_time == '') ? 'Unknown' : login_time;
        }
        </script>
    </body>
</html>
```

### Version

Multi Theft Auto v1.6-release-24060
Multi Theft Auto v1.7

### Additional context

[testresource.zip](https://github.com/user-attachments/files/26712054/testresource.zip)
```
crun setElementData(me, 'loginTime', os.time())
```

### Relevant log output

```shell

```

### Security Policy

Elabyad247:
#### Summary

Fix httpSetResponseCookie so it actually sets cookies correctly.

CResourceHTMLItem::SetResponseCookie was building the CookieParameters map with the cookie name as the key (params[szCookieName] = data). But HttpResponse::SetCookie looks for two fixed keys "name" and "value" and skips the cookie if either is empty. So no Set-Cookie header was ever sent.

The fix puts the cookie name and value under the "name"` and "value" keys.

#### Motivation

Fixes #4806. httpSetResponseCookie is the only way for a resource to attach cookies to an HTTP response, so while it's broken there's no way to do anything cookie-based.

#### Test plan

Tested on a server build with:

1. The repro resource from #4806.
2. The example on https://wiki.multitheftauto.com/wiki/HttpSetResponseCookie

Before the fix neither sent a Set-Cookie header. After the fix the header shows up and the browser stores the cookie.

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
+++ b/Server/mods/deathmatch/logic/CResourceHTMLItem.cpp
@@ -219,7 +219,8 @@ void CResourceHTMLItem::SetResponseCookie(const char* szCookieName, const char* 
     CookieParameters params;
     Datum            data;
     data = szCookieValue;
-    params[szCookieName] = data;
+    params["name"] = szCookieName;
+    params["value"] = data;
     m_currentResponse->SetCookie(params);
 }
 


FileEX (approved):
Good job 😃

Issue #3598, opened 2024-07-22
createPed server-side with the parameter synced = false, the health is still synchronized


T-MaxWiese-T:
### Describe the bug

The server-side function createPed with the parameter synced = `false` synchronizes the health anyway. However, the desired behaviour is that no synchronization takes place at all.

### Steps to reproduce

1. Execute this code on the server: 
    `createPed (0, 2226.1, -2218.7, 12, 0, false)`
3. Two players must be on the server.
4. One player beats or shoots the ped until he is dead.
5. The health is synchronized between the clients.

### Version

Client: Multi Theft Auto v1.6-release-22628
Server: MTA:SA Server v1.6-release-22628

### Additional context

None

### Relevant log output

```shell
None
```

### Security Policy

- [X] I have read and understood the Security Policy and this issue is not security related.

Fernando-A-Rocha:
So you're saying that non synced peds shouldn't take damage? Or by not syncing the health then it would only die for the client that killed it?

IIYAMA12:
Health does not seems to be synced. At least if there are no other players around. Not sure if having another remote player in the server matters in the outcome.

To reproduce:
1. Create a ped with synchronized disabled
2. Damage the ped
3. Rejoin
4. The ped health still 100%

<hr>

But I can confirm that deaths are synced, which is indeed bugged.

To reproduce:
1. Create a ped with synchronized disabled
2. Kill the ped
3. Rejoin
4. Ped is still dead

MohabCodeX:
Fixed in #4845

PR #4845, opened 2026-04-24
Fix server side peds death sync despite sync being disabled

jjustns:
#### Summary
Peds which were created on the serverside with sync disabled still had their death synchronized to all players. 
This PR should resolve the issue

#### Motivation
#3598

#### Test plan
(Solo)
- Create a serversided ped with sync disabled
- Kill the ped on client
- Reconnect (Before this change, the ped will remain dead. After this change the ped is now alive as it should be)

(Multiple Clients)
- Create a serversided ped with sync disabled
- Have the ped in different positions on each client
- Kill the ped on client 1 (Before this change, the ped dies on all clients. After this change it only dies on client 1)
- Reconnect on any client (Before this change, the ped will remain dead. After this change the ped is now alive as it should be)


Lpsd (approved):
Thanks for the PR!

PR #4846, opened 2026-04-24
remove duplicate Ryder skin 271 from player models

shadylua:
This change removes support for the duplicate Ryder skin ID 271 and keeps 300 as the only supported Ryder entry.

### DIFF
--- a/Client/game_sa/CGameSA.cpp
+++ b/Client/game_sa/CGameSA.cpp
@@ -1015,7 +1015,6 @@ void CGameSA::SetupSpecialCharacters()
     ModelInfo[268].MakePedModel("DWAYNE");
     ModelInfo[269].MakePedModel("SMOKE");
     ModelInfo[270].MakePedModel("SWEET");
-    ModelInfo[271].MakePedModel("RYDER");
     ModelInfo[272].MakePedModel("FORELLI");
     ModelInfo[273].MakePedModel("MEDIATR");
     ModelInfo[289].MakePedModel("SOMYAP");
@@ -1029,6 +1028,8 @@ void CGameSA::SetupSpecialCharacters()
     ModelInfo[297].MakePedModel("MADDOGG");
     ModelInfo[298].MakePedModel("CAT");
     ModelInfo[299].MakePedModel("CLAUDE");
+    // Keep Ryder available through the canonical 300 entry only so scripts and tools no longer
+    // see 271 as a separate selectable skin.
     ModelInfo[300].MakePedModel("RYDER2");
     ModelInfo[301].MakePedModel("RYDER3");
     ModelInfo[302].MakePedModel("EMMET");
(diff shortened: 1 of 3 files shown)


qaisjp:
What's the motivation for this change? Seems like it could break existing scripts that rely on model 271?

shadylua:
intention was cleanup/consistency, not fixing a crash or exploit. If removing 271 is considered too breaking for compatibility, I'm fine with closing this or reworking it into a non-breaking approach.

qaisjp:
Yeah, I think this would break existing scripts, and for this reason the functionality doesn't need to change.

shadylua:
As mentioned in my latest comment on PR #4846, we should only remove the incorrect ID that needs to be deleted.

Following that, you can review my related PR on GitHub (mtasa-resources):


It’s directly related to this issue, and I’d appreciate your review and feedback.

dnl-k:
300 is in the list of "special" skin ids - which, according to the wiki, were only used in cutscenes / interiors.

I don't see any reason to remove skin id 271.

shadylua:
I understand your point about ID 300 being part of the "special" cutscene/interior skins.

however issue here is that ID 300 and 271 appear to reference the same model (Ryder) which creates inconsistency between wiki mtasa-blue and resources.

My intention is not to remove valid data but to eliminate duplication and ensure consistency across all related repositories.

I will double-check in-game behavior again to confirm whether both IDs are actually usable or if one of them is redundant. Based on that I’ll proceed with a proper fix.

Xenius97:
If this doesn’t improve the game’s performance, I think it’s completely unnecessary. We also use this skin for replacement, but probably many others do as well, so we’re just unnecessarily taking up a skin slot from the developers.
Of course, there’s a way around it using engineRequest, but that requires additional scripts & sync, since the server-side model allocation PR has been sitting there for years.

shadylua:
this isn’t about performance but about data consistency duplicate or unclear IDs can cause confusion for developers especially when different sources don’t match goal is simply to clean that up not to remove functionality

Xenius97:
image

These skins were introduced around version 1.1 or 1.2 at least 10 years ago, if they were new, it would make sense, but this way there’s a risk that some older scripts relying on them might break, and we’d just be creating unnecessary extra work for servers.

shadylua:
I understand the backward compatibility concern instead of removing IDs we could keep them and improve documentation/clarity to avoid confusion my goal is consistency not breaking existing scripts

shadylua:
not about performance just consistency
some ids (like 300 and 271) are confusing / duplicated
not trying to remove anything just make it clearer
can adjust for full backward compatibility if needed

qaisjp:
At this point it's too late to change, so we can't accept this change. Thanks for the suggestion and taking the time to propose this!

PR #4880, opened 2026-05-03
Fix install-root resolution failures that lead to U01 errors in updater

Dutchman101:
Commits 4200137 & c15736e fixed some cases, but not all. This should close the remaining gaps.

**Fix install-root resolution failures that lead to U01 errors in updater**

MTA's auto-update spawns multiple processes. Process B (the installed launcher) extracts the update archive into a transient directory under `\upcache\_<archive>_tmp_<n>\` and launches Process C from the new executable there. Process C runs entirely from that temp extraction directory, which doesnt contain mta\core.dll or most MTA files.

Previously, Process C had no reliable way to know where the real MTA installation lives. It fell back to reading Last Run Location from the registry - but that value can be missing, stale, or unwritable (HKLM write failures on non-admin installs). When the registry lookup failed, Process C silently degraded to treating its own temp extraction directory as the install root. This produced a domino of problems:

* InitLocalization failed to find core.dll and aborted with CL23
* UpdateMTAVersionApplicationSetting loaded the wrong (or no) version DLL, producing wrong version strings
* ConfigureWerDumpPath and _ProcessAppCompatChecks built paths against the temp dir, silently doing nothing or pointing at non-existent files
* _SwitchBackFromTempExe called PathJoin(GetInstallPathForLauncher(), MTA_EXE_NAME) with an empty install path, producing a root-relative path like \Multi Theft Auto.exe that could match an unrelated file on the current drive
SetMTASAPathSource wrote the temp directory path into Last Run Location in the registry. When CleanDownloadCache later deleted the temp dir, the registry pointed at nothing, causing U01 on the very next launch

**What this PR does**
1. Carry the install root across the process boundary via the sequencer.

In _MaybeSwitchToTempExe (Process B), before spawning Process C, we now write the current real install root into the sequencer state snapshot as INSTALL_ROOT. When Process C starts and calls CInstallManager::SetMTASAPathSource, it reads that key back, validates it with IsUsableMtasaInstallRoot and IsTemporaryUpdateLaunchPath, and immediately installs it as a SetMTASABaseDirOverride. This override is in place before UpdateSettingsForReportLog runs (which transitively calls GetMTASAPath through UpdateMTAVersionApplicationSetting), so all downstream path consumers in Process C see the real install root. The ordering of UpdateSettingsForReportLog is moved after the path source is resolved for exactly this reason.

2. Make GetInstallPathForLauncher return empty on unresolvable temp launches, and guard every caller.

Rather than silently returning the temp path as a fallback, GetInstallPathForLauncher now returns an empty string when it cannot find a usable non-temp install root. This is a cleaner contract: callers know to apply their own fallback (typically GetMTASAPath(), which consults the carried override). Every call site is updated with an explicit guard. SetMTASAPathSource additionally refuses to write any temp-directory path into the registry.

3. Add a g_strMTASAPath fast-path in GetInstallPathForLauncher.

When the far-update sequencer path has already populated g_strMTASAPath via the override, GetInstallPathForLauncher returns it directly without touching the registry. This avoids the GetMTASAPath → SetMTASAPathSource(false) → GetInstallPathForLauncher recursion that would occur if we called GetMTASAPath() from inside the function.

4. Read registry views independently with full validation.

Both GetInstallPathForLauncher (Utils.cpp) and GetMTASABaseDir (SharedUtil.Misc.hpp) now probe the 64-bit view, 32-bit view, and default view as separate candidates, each validated through IsUsableMtasaInstallRoot and IsTemporaryUpdateLaunchPath. A stale or temp-path value in one view no longer shadows a usable value in another.

5. Harden registry reads. A 1 MB size cap is applied before any allocation, preventing a malformed registry entry on the U01 recovery path from triggering a large allocation. REG_EXPAND_SZ values are now expanded so install paths containing environment variable references (e.g. %ProgramFiles%\...) pass the file-existence validation checks.

6. Promote IsUsableMtasaInstallRoot and IsTemporaryUpdateLaunchPath to public scope so CInstallManager can use them directly for sequencer-carried root validation without duplicating the logic.

7. Strengthen the parent-process check in GetMTASABaseDir to use IsUsableMtasaInstallRoot && !IsTemporaryUpdateLaunchPath instead of only checking for core.dll, preventing the parent-process branch from pointing a child process at a directory that is about to be deleted.

Impact on normal (non-update) launches:

None. When the launch path is not a temp update directory, GetInstallPathForLauncher returns it immediately as before. The g_strMTASAPath fast-path only activates when the override was explicitly set by the install manager. The registry write guard only fires when the launch path itself is a temp directory.

### DIFF
--- a/Client/loader/CInstallManager.cpp
+++ b/Client/loader/CInstallManager.cpp
@@ -304,17 +304,28 @@ void CInstallManager::InitSequencer()
 //////////////////////////////////////////////////////////
 void CInstallManager::SetMTASAPathSource(const SString& strCommandLineIn)
 {
-    // Update some settings which are used by ReportLog
-    UpdateSettingsForReportLog();
-
     InitSequencer();
     RestoreSequencerFromSnapshot(strCommandLineIn);
 
     // If command line says we're not running from the launch directory, get the launch directory location from the registry
     if (m_pSequencer->GetVariable(INSTALL_LOCATION) == "far")
+    {
+        // Process B carried the real install root through the sequencer state. Honor it here so this
+        // temp launcher process can resolve its base dir without depending only on the registry.
+        // The override must be installed before UpdateSettingsForReportLog runs, because that helper
+        // reaches GetInstallPathForLauncher / GetMTASAPath through UpdateMTAVersionApplicationSetting.
+        const SString strRealInstallRoot = m_pSequencer->GetVariable(INSTALL_ROOT);
+        if (!strRealInstallRoot.empty() && IsUsableMtasaInstallRoot(strRealInstallRoot) && !IsTemporaryUpdateLaunchPath(strRealInstallRoot))
+            SharedUtil::SetMTASABaseDirOverride(strRealInstallRoot);
+
         ::SetMTASAPathSource(true);
+    }
     else
         ::SetMTASAPathSource(false);
+
+    // Update some settings which are used by ReportLog. Must run after the path source is resolved
+    // so that report-log helpers see the real install root rather than a temp extraction directory.
+    UpdateSettingsForReportLog();
 }
 
 //////////////////////////////////////////////////////////
@@ -1042,6 +1053,11 @@ SString CInstallManager::_MaybeSwitchToTempExe()
     // If a new "Multi Theft Auto.exe" exists, let that complete the install
     if (m_pSequencer->GetVariable(INSTALL_LOCATION) == "far")
     {
+        // Carry the current real install root forward to the temp launcher so it can set a base-dir
+        // override before reading the registry. Without this, the temp launcher has only Last Run
+        // Location to fall back on, which can be missing, stale, or denied by HKLM write failures.
+        m_pSequencer->SetVariable(INSTALL_ROOT, GetMTASAPath());
+
         ReleaseSingleInstanceMutex();
         if (ShellExecuteNonBlocking("open", GetLauncherPathFilename(), GetSequencerSnapshot()))
             ExitProcess(0);  // All done here
@@ -1065,8 +1081,13 @@ SString CInstallManager::_SwitchBackFromTempExe()
     {
         m_pSequencer->SetVariable(INSTALL_LOCATION, "near");
 
-        SString strLauncherPathFilename = PathJoin(GetInstallPathForLauncher(), MTA_EXE_NAME);
-        if (!FileExists(strLauncherPathFilename))
+        // GetInstallPathForLauncher returns empty when running from a temp update directory and no
+        // usable non-temp install root could be resolved. PathJoin would then produce a root-relative
+        // probe like "\Multi Theft Auto.exe", which could match an unrelated file on the current drive
+        // and cause us to launch the wrong executable. Skip straight to the GetMTASAPath() fallback.
+        const SString strRealInstallRoot = GetInstallPathForLauncher();
+        SString       strLauncherPathFilename = strRealInstallRoot.empty() ? SString() : PathJoin(strRealInstallRoot, MTA_EXE_NAME);
+        if (strLauncherPathFilename.empty() || !FileExists(strLauncherPathFilename))
         {
             strLauncherPathFilename = PathJoin(GetMTASAPath(), MTA_EXE_NAME);
             if (!FileExists(strLauncherPathFilename))
@@ -1729,7 +1750,10 @@ SString CInstallManager::_ProcessAppCompatChecks()
     IsWow64Process(GetCurrentProcess(), &bIsWOW64);
     uint    uiHKLMFlags = bIsWOW64 ? KEY_WOW64_64KEY : 0;
     WString strGTAExePathFilename = GetGameExecutablePath().wstring();
-    WString strMTAExePathFilename = FromUTF8(PathJoin(GetInstallPathForLauncher(), MTA_EXE_NAME));
+    SString strMTAInstallPath = GetInstallPathForLauncher();
+    if (strMTAInstallPath.empty())
+        strMTAInstallPath = GetMTASAPath();
+    WString strMTAExePathFilename = FromUTF8(PathJoin(strMTAInstallPath, MTA_EXE_NAME));
     WString strCompatModeRegKey = L"SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\AppCompatFlags\\Layers";
     int     bWin816BitColorOption = GetApplicationSettingInt("Win8Color16");
     int     bWin8MouseOption = GetApplicationSettingInt("Win8MouseFix");
--- a/Client/loader/CInstallManager.cpp
+++ b/Client/loader/CInstallManager.cpp
@@ -315,8 +315,20 @@ void CInstallManager::SetMTASAPathSource(const SString& strCommandLineIn)
         // The override must be installed before UpdateSettingsForReportLog runs, because that helper
         // reaches GetInstallPathForLauncher / GetMTASAPath through UpdateMTAVersionApplicationSetting.
         const SString strRealInstallRoot = m_pSequencer->GetVariable(INSTALL_ROOT);
-        if (!strRealInstallRoot.empty() && IsUsableMtasaInstallRoot(strRealInstallRoot) && !IsTemporaryUpdateLaunchPath(strRealInstallRoot))
+        if (strRealInstallRoot.empty())
+        {
+            AddReportLog(1063, "CInstallManager::SetMTASAPathSource: no sequencer install root, falling back to registry");
+        }
+        else if (SharedUtil::IsUsableMtasaInstallRoot(strRealInstallRoot) && !SharedUtil::IsTemporaryUpdateLaunchPath(strRealInstallRoot))
+        {
+            AddReportLog(1063, SString("CInstallManager::SetMTASAPathSource: honoring sequencer install root '%s'", strRealInstallRoot.c_str()));
             SharedUtil::SetMTASABaseDirOverride(strRealInstallRoot);
+        }
+        else
+        {
+            AddReportLog(1063, SString("CInstallManager::SetMTASAPathSource: rejecting unusable sequencer install root '%s', falling back to registry",
+                                       strRealInstallRoot.c_str()));
+        }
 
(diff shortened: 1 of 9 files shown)


Client/loader/Utils.cpp
```diff
@@ -588,15 +623,66 @@ SString GetInstallPathForLauncher()
...
-    if (IsUsableMtasaInstallRoot(strSavedInstallPath64) && !IsTemporaryUpdateLaunchPath(strSavedInstallPath64))
-        return strSavedInstallPath64;
+    // No usable non-temp install root resolved. Returning empty forces callers to use their own fallbacks
+    // (such as a base-dir override carried forward from the parent launcher) instead of treating a temp
+    // update directory as the real install location.
+    return SString();
+}
 
-    return strLaunchPath;
+bool IsUsableMtasaInstallRoot(const SString& strPath)
```
Lpsd: This method already exists in SharedUtil.Misc.hpp - can we share this instead?

Client/loader/Utils.cpp
```diff
@@ -580,6 +559,62 @@ namespace
...
+        const WString wstrValueName = FromUTF8("Last Run Location");
+
+        HKEY hKey = nullptr;
+        if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, wstrSubKey.c_str(), 0, KEY_READ | viewFlag, &hKey) != ERROR_SUCCESS || !hKey)
+            return "";
+
+        DWORD valueType = REG_SZ;
+        DWORD valueSize = 0;
+        LONG  result = RegQueryValueExW(hKey, wstrValueName.c_str(), nullptr, &valueType, nullptr, &valueSize);
+        if (result != ERROR_SUCCESS || (valueType != REG_SZ && valueType != REG_EXPAND_SZ) || valueSize == 0 || valueSize > kMaxRegistryValueBytes)
```
Lpsd:
```suggestion
        if (result != ERROR_SUCCESS || (valueType != REG_SZ && valueType != REG_EXPAND_SZ) || valueSize == 0 || valueSize > kMaxRegistryValueBytes || 
        (valueSize % sizeof(wchar_t)) != 0)
```

Client/loader/Utils.cpp
```diff
@@ -580,6 +559,62 @@ namespace
...
+        {
+            RegCloseKey(hKey);
+            return "";
+        }
+
+        std::vector<wchar_t> buffer((valueSize / sizeof(wchar_t)) + 1u, L'\0');
+        result = RegQueryValueExW(hKey, wstrValueName.c_str(), nullptr, &valueType, reinterpret_cast<LPBYTE>(buffer.data()), &valueSize);
+        RegCloseKey(hKey);
+
+        if (result != ERROR_SUCCESS || (valueType != REG_SZ && valueType != REG_EXPAND_SZ) || valueSize > kMaxRegistryValueBytes)
```
Lpsd:
```suggestion
        if (result != ERROR_SUCCESS || (valueType != REG_SZ && valueType != REG_EXPAND_SZ) || valueSize > kMaxRegistryValueBytes ||
            (valueSize % sizeof(wchar_t)) != 0)
```

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -393,20 +393,44 @@ static bool IsUsableMtasaInstallRoot(const SString& strPath)
...
     HKEY hkTemp = NULL;
-    if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, wstrSubKey, 0, KEY_READ | KEY_WOW64_64KEY, &hkTemp) != ERROR_SUCCESS || !hkTemp)
+    if (RegOpenKeyExW(HKEY_LOCAL_MACHINE, wstrSubKey, 0, KEY_READ | viewFlag, &hkTemp) != ERROR_SUCCESS || !hkTemp)
         return "";
 
     DWORD dwType = REG_SZ;
     DWORD dwSize = 0;
     LONG  result = RegQueryValueExW(hkTemp, wstrValue, NULL, &dwType, NULL, &dwSize);
-    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ) || dwSize == 0)
+    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ) || dwSize == 0 || dwSize > kMaxRegistryValueBytes)
```
Lpsd:
```suggestion
    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ) || dwSize == 0 || dwSize > kMaxRegistryValueBytes ||
        (dwSize % sizeof(wchar_t)) != 0)
```

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -416,18 +440,31 @@ static SString ReadInstallRootRegistryValue64()
     result = RegQueryValueExW(hkTemp, wstrValue, NULL, &dwType, reinterpret_cast<LPBYTE>(buffer.data()), &dwSize);
     RegCloseKey(hkTemp);
 
-    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ))
+    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ) || dwSize > kMaxRegistryValueBytes)
```
Lpsd:
```suggestion
    if (result != ERROR_SUCCESS || (dwType != REG_SZ && dwType != REG_EXPAND_SZ) || dwSize > kMaxRegistryValueBytes || (dwSize % sizeof(wchar_t)) != 0)
```

Shared/sdk/SharedUtil.Misc.hpp
```diff
@@ -393,20 +393,44 @@ static bool IsUsableMtasaInstallRoot(const SString& strPath)
...
+    if (strLaunchPath.empty())
+        return false;
+
+    if (!strLaunchPath.ContainsI("\\upcache\\"))
+        return false;
+
+    // The auto-update flow creates the extraction directory in Install.cpp::CheckOnRestartCommand
+    // as MakeUniquePath("...\\upcache\\_<archiveName>_tmp_"), so the leaf always begins with '_'.
+    const SString strLeaf = ExtractFilename(strLaunchPath);
+    return strLeaf.BeginsWith("_") && strLeaf.ContainsI("_tmp_");
```
Lpsd: This matches anywhere in the path, so if any ancestor of the MTA install path is called `upcache` (unlikely but possible) it would be misclassified. It would be better to do this:

```suggestion
    // The auto-update flow creates the extraction directory in Install.cpp::CheckOnRestartCommand
    // as MakeUniquePath("...\\upcache\\_<archiveName>_tmp_"), so the leaf always begins with '_'
    // and its immediate parent directory is literally named "upcache".
    const SString strLeaf = ExtractFilename(strLaunchPath);
    if (!strLeaf.BeginsWith("_") || !strLeaf.ContainsI("_tmp_"))
        return false;

    // Anchor on the parent component rather than a substring search so an unrelated 
    // install path that simply contains the word "upcache" somewhere cannot be misclassified.
    const SString strParent = ExtractFilename(ExtractPath(strLaunchPath).TrimEnd("\\"));
    return strParent.CompareI("upcache");
```

Client/loader/CInstallManager.cpp
```diff
@@ -304,17 +304,28 @@ void CInstallManager::InitSequencer()
...
     // If command line says we're not running from the launch directory, get the launch directory location from the registry
     if (m_pSequencer->GetVariable(INSTALL_LOCATION) == "far")
+    {
+        // Process B carried the real install root through the sequencer state. Honor it here so this
+        // temp launcher process can resolve its base dir without depending only on the registry.
+        // The override must be installed before UpdateSettingsForReportLog runs, because that helper
+        // reaches GetInstallPathForLauncher / GetMTASAPath through UpdateMTAVersionApplicationSetting.
+        const SString strRealInstallRoot = m_pSequencer->GetVariable(INSTALL_ROOT);
+        if (!strRealInstallRoot.empty() && IsUsableMtasaInstallRoot(strRealInstallRoot) && !IsTemporaryUpdateLaunchPath(strRealInstallRoot))
+            SharedUtil::SetMTASABaseDirOverride(strRealInstallRoot);
```
Lpsd:
```suggestion
        if (!strRealInstallRoot.empty() && IsUsableMtasaInstallRoot(strRealInstallRoot) && !IsTemporaryUpdateLaunchPath(strRealInstallRoot))
        {
            SharedUtil::SetMTASABaseDirOverride(strRealInstallRoot);
            AddReportLog(1063, SString("SetMTASAPathSource: honoring sequencer-carried install root '%s'", strRealInstallRoot.c_str()));
        }
        else if (!strRealInstallRoot.empty())
        {
            // Carried value failed validation (e.g. install was moved/deleted between Process B and C,
            // or the sequencer string was truncated). Fall through to the registry-based resolver.
            AddReportLog(1063, SString("SetMTASAPathSource: rejecting unusable sequencer-carried install root '%s'", strRealInstallRoot.c_str()));
        }
        else
        {
            AddReportLog(1063, "SetMTASAPathSource: no sequencer-carried install root present, falling back to registry");
        }
```

Client/loader/Utils.cpp
```diff
@@ -588,15 +623,66 @@ SString GetInstallPathForLauncher()
...
+    if (strLaunchPath.empty())
+        return false;
+
+    if (!strLaunchPath.ContainsI("\\upcache\\"))
+        return false;
+
+    // The auto-update flow creates the extraction directory in Install.cpp::CheckOnRestartCommand
+    // as MakeUniquePath("...\\upcache\\_<archiveName>_tmp_"), so the leaf always begins with '_'.
+    const SString strLeaf = ExtractFilename(strLaunchPath);
+    return strLeaf.BeginsWith("_") && strLeaf.ContainsI("_tmp_");
```
Lpsd: Same as in SharedUtil.Misc.hpp

Lpsd (changes requested):

Dutchman101:
All reviews should be addressed by b5e56e6

Lpsd (approved):

PR #4887, opened 2026-05-04
Fix push sync vehicle syncer hijack

Zephkek:
#### Summary

Prevent vehicle push sync from replacing an existing unoccupied vehicle syncer.

Conditions for whether a player can claim syncer of a vehicle:

- there is no existing syncer
- there is no "persistent syncer" (set when setElementSyncer is used)
- **there is a contact element**

#### Motivation
Pushsync accepted a nearby client's vehicle ID and could call `OverrideSyncer` even when another player already owned sync. This let forged pushsync packets hijack vehicle sync authority.

#### Test plan

* Review the pushsync path to confirm it now only fills missing syncers.

#### Checklist


Lpsd (approved):
Very nice, thanks!

Server/mods/deathmatch/logic/CUnoccupiedVehicleSync.cpp
```diff
@@ -485,21 +485,22 @@ void CUnoccupiedVehicleSync::Packet_UnoccupiedVehiclePushSync(CUnoccupiedVehicle
         {
             // Convert to a CVehicle
             CVehicle* pVehicle = static_cast<CVehicle*>(pVehicleElement);
-            // Is the player syncing this vehicle and there is no driver? Also only process
-            // this packet if the time context matches.
-            if (pVehicle->GetSyncer() != pPlayer && pVehicle->GetTimeSinceLastPush() >= MIN_PUSH_ANTISPAM_RATE &&
+
+            CPlayer*   pSyncer = pVehicle->GetSyncer();
+            const bool bCanClaimSync = !pSyncer || (!IsSyncerPersistent() && pPlayer->GetContactElement() == pVehicle);
```
qaisjp, line 490: Is this the server side equivalent of this check?

https://github.com/multitheftauto/mtasa-blue/blob/6c54a2df44a4da07e43d774c5536cfd4b3458559/Client/mods/deathmatch/logic/CClientGame.cpp#L4710-L4713

qaisjp: yep -- https://github.com/multitheftauto/mtasa-blue/commit/6df10945aba9cae5085713add11019724cd89ee8

qaisjp (approved):
This affects sync when the player ped is "pushing" a vehicle, by walking into it.

`pPlayer->GetContactElement() == pVehicle` is the main change here.

This could in theory cause 1.5s of desync for a player vehicle, if the "contact" packet arrives after the vehicle push packet.

If that happens, the server will reject the syncer change and the client will stop sending syncer packets for ~1.5s (because of [the anti-span check client side](https://github.com/multitheftauto/mtasa-blue/blob/6c54a2df44a4da07e43d774c5536cfd4b3458559/Client/mods/deathmatch/logic/CClientGame.cpp#L4714-L4715))

This seems OK given that:

- if there are other players in the vicinity, the other players will have their own simulation of the push
- the player doing the pushing is likely pushing for more than 1.5s, so the next push packet is likely to succeed
- if the player is pushing a vehicle carefully (where packets being lost for 1.5s would be bad), they have likely already set foot in the vehicle before

That said, is this change strictly necessary, given that we have the `iVehicleContactSyncRadius` check?

qaisjp:
Question for you @Zephkek!

Zephkek:
@qaisjp Thanks for the review, to answer your question yes I believe this check is necessary because the radius check alone doesn't prevent a cheating client from hijacking sync (default 30 units which is about 4 car lengths). This PR fixes the ability to forge or unintended push packet from a player who is *near* a vehicle but not actually pushing it so radius alone passes that. `GetContactElement() == pVehicle` is checking if the player is actually in contact with the vehicle and gate sourced from the independent puresync stream, which the push packet can't unilaterally fake.

So the two checks aren't redundant the contact check is what specifically prevents syncer hijack and it's only enforced when there's already a syncer to take over from so the no-syncer "first one to nudge it" remains unchanged as intended.

Zephkek:
> If that happens, the server will reject the syncer change and the client will stop sending syncer packets for ~1.5s (because of [the anti-span check client side](https://github.com/multitheftauto/mtasa-blue/blob/6c54a2df44a4da07e43d774c5536cfd4b3458559/Client/mods/deathmatch/logic/CClientGame.cpp#L4714-L4715))


Also about the 1.5s desync concern, in theory yes client can lose 1.5s of push sync packets if the contact element update arrives after the first push packet but that window is bounded by exactly the case where it least matters:

- It only triggers when **another player is already the syncer** of the same vehicles otherwise (!pSyncer bypasses the contact check) meaning someone elses simulation is already authoritative so visible drift is minimal

- Persistent syncers bypass it via !IsSyncerPersistent()

- For non persistent cases with an existing syncer, the legitimate displacing player will set foot on/ collide with the vehicle within a frame or two of starting to push and the next push packet after MIN_PUSH_ANTISPAM_RATE will go through normally

Dutchman101 (approved):

PR #4890, opened 2026-05-04
Fix chat priority starvation under high load

### LINKED ISSUE #4872: Chat Delay with High Player Count
### Describe the bug

When the server player count exceeds ~950 players, the in-game chat system starts experiencing noticeable delays. Messages take time to appear for players.
Other systems are working perfectly fine even at high player counts:
Commands execute instantly
Client events trigger without delay
Server events trigger without delay
The issue appears to be isolated specifically to the chat system.

### Steps to reproduce

1. Increase player count to ~950+ players
2. Send chat messages
3. Observe delay in message delivery

### Version

_No response_

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

Zephkek:
#### Summary
Raise chat echo and clear packets from low to medium priority.

#### Motivation
Resolves #4872. At high player counts, low-priority reliable chat packets can be starved by constant sync traffic, while commands and client events remain responsive because they use higher priority packets.

#### Test plan

* Verify chat still uses `PACKET_ORDERING_CHAT` for ordered delivery.

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/packets/CChatClearPacket.h
+++ b/Server/mods/deathmatch/logic/packets/CChatClearPacket.h
@@ -18,9 +18,9 @@ class CChatClearPacket final : public CPacket
 public:
     CChatClearPacket() {};
 
-    // Chat uses low priority channel to avoid getting queued behind large map downloads #6877
+    // Keep chat on its own ordered channel, but avoid low-priority starvation under heavy sync load.
     ePacketID               GetPacketID() const { return PACKET_ID_CHAT_CLEAR; };
-    unsigned long           GetFlags() const { return PACKET_LOW_PRIORITY | PACKET_RELIABLE | PACKET_SEQUENCED; };
+    unsigned long           GetFlags() const { return PACKET_MEDIUM_PRIORITY | PACKET_RELIABLE | PACKET_SEQUENCED; };
     virtual ePacketOrdering GetPacketOrdering() const { return PACKET_ORDERING_CHAT; }
 
     bool Write(NetBitStreamInterface& BitStream) const;
--- a/Server/mods/deathmatch/logic/packets/CChatEchoPacket.h
+++ b/Server/mods/deathmatch/logic/packets/CChatEchoPacket.h
@@ -45,9 +45,9 @@ public:
         m_ucMessageType = ucMessageType;
     };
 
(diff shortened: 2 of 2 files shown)


qaisjp:
Could there be any negative impact to this change, such that other traffic could be starved by chat messages? I'm curious what other "medium priority" messages are now competing with chat.

Just imagining someone trying to flood the server with chat messages as a DOS attack, but I guess that's possible with any packet anyway.

Could you check to see whether there's anything interesting in git blame? I'm guessing it goes back to the initial commit though

Zephkek:
> Could there be any negative impact to this change, such that other traffic could be starved by chat messages? I'm curious what other "medium priority" messages are now competing with chat.
> 
> Just imagining someone trying to flood the server with chat messages as a DOS attack, but I guess that's possible with any packet anyway.
> 
> Could you check to see whether there's anything interesting in git blame? I'm guessing it goes back to the initial commit though

Chat would mainly be competing with the regular sync packets, like player/vehicle pure sync, key sync, camera sync, ped/object/unoccupied vehicle sync, and bullet sync/custom weapon bullet sync.
 

So yes, in theory, a sustained chat flood could now put pressure on medium-priority sync traffic rather than only low-priority traffic (though this is irrelevant because if an attacker can manage to send packets through RakNet he can set whatever priority he wants). That said, chat still uses PACKET_ORDERING_CHAT, so it stays on its own ordered channel and should not introduce reliable-ordered blocking on the default sync channel, and ofc high-priority traffic also still stays above it.

I also checked blame. It does not go back to the initial commit: 
- CChatEchoPacket was changed from high to low in [97f8fe4eb](https://github.com/multitheftauto/mtasa-blue/commit/97f8fe4eb47b9037f904a5c57348d4569a9a8149)
 for #6877, specifically "Chat messages don't update while a map download is in progress". 
- CChatClearPacket was added later in [600de4948](https://github.com/multitheftauto/mtasa-blue/commit/600de49482de4ed920ea410d9012260860290779)
 and copied the same low-priority behavior/comment.

The remaining tradeoff I see is that chat spam can now contend with normal sync under abuse, but that feels like something better handled with chat/output rate limiting than by keeping chat in a priority class where it can starve during normal high-load sync.

qaisjp (approved):
Thank you!

PR #4894, opened 2026-05-05
Fix hugely popular crash in RpAnimBlendAllocateData (SA 0x000D5F6F)

Dutchman101:
This PR fixes all paths i could trace for a very common crash, according to crash stats.
There's not much to say, other than that the pre-existing hook meant to avert this crash wasn't sufficient, but now it is.


Client/mods/deathmatch/logic/CClientPed.cpp
qaisjp, line 1: The code changes here seem to be guarding against `m_pModelInfo` pointing to invalid data, but do we know why it's invalid in the first place?

qaisjp: (Seems like a use after free or something?)

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -2865,15 +2868,80 @@ static void __declspec(naked) HOOK_CAnimManager_CreateAnimAssocGroups()
...
+    argMap.Set("reason", "createassoc");
+    SetApplicationSetting("diagnostics", "gta-model-fail", argMap.ToString());
+}
+
+#define HOOKPOS_CAnimBlendAssocGroup_CreateAssociations   0x4CE2F7
+#define HOOKSIZE_CAnimBlendAssocGroup_CreateAssociations  7
+#define HOOKCHECK_CAnimBlendAssocGroup_CreateAssociations 0x8B
+DWORD                 RETURN_CAnimBlendAssocGroup_CreateAssociations = 0x4CE2FE;
+DWORD                 RETURN_CAnimBlendAssocGroup_CreateAssociations_Skip = 0x4CE36F;
+void _declspec(naked) HOOK_CAnimBlendAssocGroup_CreateAssociations()
```
qaisjp, line 2918: Please could you write some comments explaining exactly what the hook does?

Client/mods/deathmatch/logic/CClientPed.cpp
Dutchman101: > The code changes here seem to be guarding against `m_pModelInfo` pointing to invalid data, but do we know why it's invalid in the first place?

@qaisjp 
No, we don't know it.. the 'fix' acts on an emerging crash as per the crash stats. There's users complaining of this crash as well, but none of them could pinpoint the exact circumstances. After i couldn't find a 100% confirmed root cause, i just looked at, and prevented, additional paths i could find for it to occur.

Averting crashes based on stats and collected dumps alone is an established practice, we don't always get much to work with.

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
Lpsd: @Dutchman101 please can you add the requested comments here and we'll get this merged

PR #4896, opened 2026-05-05
Fix SVG texture color banding (#4891)

### LINKED ISSUE #4891: SVG textures color banding
### Describe the bug

A texture with a gradient created using ``svgCreate`` shows visible color banding, which does not occur with a regular PNG texture. Probably caused by https://github.com/multitheftauto/mtasa-blue/commit/04f297b7b1aecb3753c8fbfa19fa9627abf422b4

Left SVG, right PNG
Image

### Steps to reproduce

```lua
local svg = [[<svg width="811" height="736" viewBox="0 0 811 736" fill="none" xmlns="http://www.w3.org/2000/svg"><rect width="811" height="736" rx="21" fill="url(#a)" fill-opacity=".87"/><rect x="1.5" y="1.5" width="808" height="733" rx="19.5" stroke="#657c86" stroke-width="3"/><defs><linearGradient id="a" x1="405.937" y1="278.995" x2="405.937" y2="793.625" gradientUnits="userSpaceOnUse"><stop stop-color="#2b3c4d"/><stop offset="1" stop-color="#1f2326"/></linearGradient></defs></svg>]]
```

### Version

Multi Theft Auto v1.6-release-24060

### Additional context

Image

### Relevant log output

```shell

```

### Security Policy

Zephkek:
#### Summary
SVG textures rendered with any opacity below 1.0 show visible color banding on smooth gradients. This PR keeps the lunasvg bitmap in its native premultiplied-alpha form and routes the default `dxDrawImage` blend through MTA's existing PM blend pipeline instead of unpremultiplying per pixel.

#### Motivation
Closes #4891.

The render path was doing `r = (r * 255) / a` per pixel after lunasvg wrote a premultiplied 8-bit ARGB bitmap. With alpha < 255 this is mathematically incapable of producing a smooth gradient: plutovg's gradient LUT is 8-bit per channel, so a translucent gradient (e.g. opacity 0.87, a=222) only takes ~11 distinct premul-R states across the whole gradient. The integer divide then stretches those 11 states across 0..255 and skips outputs (33 -> 37, 34 -> 39, 38 unreachable). Round-to-nearest does not help, it only shifts which output value is skipped.

PNG looks fine in the same scene because PNG stores straight alpha and never goes through the premul/unpremul roundtrip.

The fix is what every modern compositor does (Direct2D, Skia, WPF, browsers): keep the data premultiplied and use a (ONE, INVSRCALPHA) blend at draw time.

- Add `m_bPremultipliedAlpha` flag on `CMaterialItem`, default false.
- Set the flag in `CVectorGraphicItem::PostConstruct`. Drop the lossy `UnpremultiplyBitmap` pass.
- In `DrawTextureQueued` and `DrawMaterialPrimitiveQueued`, when the active blend is the default `BLEND` and the material is PM, route the queued item to `ADD`, which is the existing `(ONE, INVSRCALPHA)` mode in `CGraphics::SetBlendModeRenderStates`. Scripts that set a blend mode explicitly are not affected.

#### Test plan
- Reproduce #4891 with the SVG from the issue (rounded rect + linear gradient #2b3c4d -> #1f2326, opacity 0.87) via `svgCreate` and `dxDrawImage`. Before: visible bands. After: smooth gradient matching the PNG reference.
- Sanity check on opaque SVGs (no opacity, no gradient transparency): unchanged output, since `BLEND` and `ADD` agree when alpha is 255.
- Sanity check on `dxSetBlendMode("modulate_add")` and `dxSetBlendMode("add")` paths: untouched, the auto-route only fires when the active mode is `BLEND`.
- `dxDrawMaterialPrimitive` with an SVG texture under default blend: composites correctly (same auto-route applied).

#### Checklist

### DIFF
--- a/Client/core/Graphics/CGraphics.cpp
+++ b/Client/core/Graphics/CGraphics.cpp
@@ -27,6 +27,40 @@ extern std::atomic<bool> g_bInMTAScene;
 
 using namespace std;
 
+namespace
+{
+    // Convert a straight-alpha ARGB tint into its premultiplied form:
+    //
+    //     (R, G, B, A)  ->  (R*A/255, G*A/255, B*A/255, A)
+    //
+    // Needed when routing a translucent draw onto the (ONE, INVSRCALPHA)
+    // pipeline. That pipeline computes
+    //
+    //     out.rgb = src.rgb + dst.rgb * (1 - src.a)
+    //
+    // with src.rgb = tex.rgb * diff.rgb and src.a = tex.a * diff.a, so
+    // diff.a never enters the source rgb term. For a PM texel tex.rgb =
+    // R*A_tex, that leaves
+    //
+    //     out.rgb = R * A_tex * diff.rgb + dst.rgb * (1 - A_tex * diff.a)
+    //
+    // which is brighter than the correct PM composite by a factor of
+    // 1/diff.a during a fade. Premultiplying the diffuse here
+    // (diff.rgb *= diff.a) restores the missing factor and the equation
+    // reduces exactly to the PM "over" operator.
+    inline SColor PremultiplyAlpha(SColor c) noexcept
+    {
+        if (c.A == 0xFF)
+            return c;
+
+        const uint a = c.A;
+        c.R = static_cast<unsigned char>((c.R * a + 127) / 255);
+        c.G = static_cast<unsigned char>((c.G * a + 127) / 255);
+        c.B = static_cast<unsigned char>((c.B * a + 127) / 255);
+        return c;
+    }
+}  // namespace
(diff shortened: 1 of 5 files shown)


FileEX:
Unpremultiply was introduced as a fix for #3828 - won’t this PR reintroduce that issue?

Zephkek:
> Unpremultiply was introduced as a fix for #3828 - won’t this PR reintroduce that issue?
  
  Yes it does, fixed by premultiplying the diffuse when the auto-route fires (diff.rgb *= diff.a)

FileEX (approved):

Issue #4062, opened 2025-03-02
TXD loading uses double memory


Xenius97:
### Describe the bug

If you have a 10mb txd and load it with engineLoadTXD then it will show 20mb in memory stat
I've talk with @TheNormalnij , he said engineLoadTXD stores every TXD twice in memory

### Steps to reproduce

use engineLoadTXD

### Version

all versions

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

Meine1:
No wonder memory usage goes up so much even with optimized textures.

Xenius97:
> No wonder memory usage goes up so much even with optimized textures.

I've optimized every textures, every texture is compressed as it should, tried to disable all mipmaps too and nothing changed. This isn't texture optimizing issue, i had a lot of tests and then ask one of dev who almost knows everything about MTA/GTA.

ColombuxMaximus:
https://github.com/multitheftauto/mtasa-blue/blob/master/Client/game_sa/CRenderWareSA.cpp#L237

`RwTexDictionaryStreamRead `instead of `RwTexDictionaryGtaStreamRead` would fix it?

(@TheNormalnij )

Dryxio:
`RwTexDictionaryStreamRead` instead of `RwTexDictionaryGtaStreamRead` would fix it?

(@TheNormalnij  )

Xenius97:
> `RwTexDictionaryStreamRead` instead of `RwTexDictionaryGtaStreamRead` would fix it?
> 
> ([@TheNormalnij](https://github.com/TheNormalnij) )

I've tested, same issue. Maybe not this is the issue.

PR #4897, opened 2026-05-05
Reduce TXD load memory before import

Zephkek:
#### Summary
Delay retaining decoded TXD replacement textures until `engineImportTXD` actually applies them.

`engineLoadTXD` still validates file-path and raw-data TXDs, but it now releases the decoded RenderWare textures immediately after validation instead of keeping them on the TXD element before import.

#### Motivation
Fixes unnecessary memory usage reported in #4062 where `engineLoadTXD` could hold decoded TXD textures even before they were imported into any model.

For raw-data TXDs, the original byte buffer is still kept because clothes replacement stores pointers to that data. For normal model imports, decoded textures are loaded again on `engineImportTXD` and the existing raw-data cleanup path still frees the buffer afterward.

#### Test plan

- `engineLoadTXD(file)` validates then releases decoded textures.
- `engineLoadTXD(rawData)` validates, keeps raw bytes for clothes compatibility, and releases decoded textures.
- `engineImportTXD` reloads textures through the existing file/raw import paths.
- `engineAddClothingTXD` and clothes TXD ID imports still have valid `m_FileData`.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientTXD.cpp
+++ b/Client/mods/deathmatch/logic/CClientTXD.cpp
@@ -107,29 +107,28 @@ bool CClientTXD::Import(unsigned short usModelID)
         // Ensure loaded for replacing model textures
         if (m_ReplacementTextures.textures.empty())
         {
-            if (!m_bIsRawData)
+            // Decode from the buffer kept since Load. Falls back to a fresh
+            // disk read for the file path only if the buffer was already freed
+            // (e.g. after an earlier non-clothes import).
+            if (m_FileData.empty())
             {
+                if (m_bIsRawData)
+                    return false;
                 SString strUseFilename;
                 if (!GetFilenameToUse(strUseFilename))
                     return false;
-                g_pGame->GetRenderWare()->ModelInfoTXDLoadTextures(&m_ReplacementTextures, strUseFilename, SString(), m_bFilteringEnabled);
-                if (m_ReplacementTextures.textures.empty())
+                if (!FileLoad(std::nothrow, strUseFilename, m_FileData))
                     return false;
             }
-            else
-            {
-                g_pGame->GetRenderWare()->ModelInfoTXDLoadTextures(&m_ReplacementTextures, NULL, m_FileData, m_bFilteringEnabled);
-                if (m_ReplacementTextures.textures.empty())
-                    return false;
-            }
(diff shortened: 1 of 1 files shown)


qaisjp (dismissed):
I don't fully understand the implications of the code changes themselves, I'll dive into the code when I get chance, but I have some feedback about some of the points in the description:

- if engineLoadTXD is validating textures, but doesn't hold onto any resources, and then engineImportTXD reloads from disk; isn't there an opportunity here for the file to be rewritten on disk and to load something that hasn't been validated?
- the test plan you've written ins useful, but not actually a test plan. Was there a resource you tested with?

Zephkek:
@qaisjp I think the rewrite-on-disk window already exists in master, ```CClientDFF::LoadFromFile``` does no format validation at all, it just checks ```FileExists``` plus ```CheckFile``` and defers ```ReadDFF``` to replace time so the TXD path now just matches what DFF already does. 

The reload from disk branch inside ```Import()``` (CClientTXD.cpp L110-L117) is also reachable when a TXD is used as clothes first then re-imported on a non clothes model since the clothes path clears m_ReplacementTextures at L84-88. CheckFile runs again on every import through GetFilenameToUse and if the bytes changed to garbage in between then ```ReadTXD``` just returns NULL and Import would return false so there's no new crash surface either. Either way I took care of it it's now slurped into m_FileData on load.


As for testing, I haven't gotten the chance to test yet as the fix seemed obvious enough just from the code logic, I'm waiting on FileEX to test it the plan was just what needed to be validated code wise. The real root cause of the issue isn't what this PR fixes, D3DPOOL_MANAGED keeps a CPU side shadow copy in addition to the vram copy so a 10mb DXTn TXD becomes 10MB shadow + 10MB VRAM that's what the issue reporter sees as "2x"

What im fixing here however is a 3x path, where in the buffer path there's a *second* duplication on top CClientTXD.cpp L177 keeps m_FileData (the raw 10 mb) *and* the decoded textures live until Import() frees one of them. So the buffer pathway memory peaks at raw bytes + D3D shadow + VRAM so 3x pre this PR, so the issue being fixed here is correct.

Zephkek:
ill be writing an enhancement PR that alters D3DPOOL_MANAGED behavior, this one will need a lot of testing but it'll bring massive improvements if proven to be good.

qaisjp (approved):
LGTM but ideally we didn't call `ModelInfoTXDLoadTextures` twice (seems wasteful?) and instead just reused the same m_ReplacementTextures produced by the original load call.

PR #4918, opened 2026-05-16
Add ability to enable/disable vehicle model hierarchy typo fixes

SpeedyFolf:
#### Summary

Adds an opt-in clientside fix ([SilentPatch](https://github.com/CookiePLMonster/SilentPatch)-style) so GTA can match **canonical** vehicle component names from handling/IDE to **typo'd** frame names in stock `.dff` hierarchies (e.g. `wheel_lm` vs `wheel_lm_dummy`, `boat_moving` vs `boat_moving_hi`, transmission / taillight / misc / elevator spellings). Disclaimer: This was created with assistance from Cursor, an AI coding tool.

Implementation:

- Hooks the existing **`_stricmp` call site** used during **`CVehicleModelInfo`** setup (same approach as SilentPatch SA: `0x4C5311`), with a small sorted alias table.
- **`CMultiplayer` / `CMultiplayerSA`**: `SetVehicleHierarchyTypoFixesEnabled` / `GetVehicleHierarchyTypoFixesEnabled` (default **off**).
- Hook installed from **`CMultiplayerSA::InitHooks_Vehicles`** (new `CMultiplayerSA_VehicleHierarchyTypoFixes.cpp`).
- **Lua (client only):** `engineSetVehicleHierarchyTypoFixesEnabled`, `engineGetVehicleHierarchyTypoFixesEnabled`, and `Engine.set/getVehicleHierarchyTypoFixesEnabled`.

#### Motivation

Server/scripters today fix broken components (wheels, boat moving parts, lights, etc.) by shipping custom DFFs that only rename hierarchy nodes. That duplicates data and increases download size. Matching SilentPatch's optional behaviour in the client lets scripts enable the same matching without replacing models, while keeping vanilla behaviour for everyone else by default.

#### Test plan

1. **Default off:** Fresh connect - `engineGetVehicleHierarchyTypoFixesEnabled()` is false; stock vehicles behave as before (no regression when unused).
2. **On from start:** In a client script, `engineSetVehicleHierarchyTypoFixesEnabled(true)` early (e.g. `onClientResourceStart`), then spawn/stream affected models - verify e.g. DFT-30 wheel, Predator/Tropic moving boat part, Dumper transmission dummies, Uranus taillights, Sweeper misc, Dodo elevator, etc., match expected fixed behaviour (same as my DFF-replace resource).
3. **Mid-session toggle:** With model already loaded, toggling alone does not visually update until `engineRestreamModel(vehicleModelId)` is used.
4. **`crun`:** `engineSetVehicleHierarchyTypoFixesEnabled(true)` / `false` and getters - console shows success.

This is not reset in `ResetWorldProperties`, just like `setGlitchEnabled()` glitches; resources/maps should disable this when they're stopped like other opt-in client features.

#### Checklist


FileEX:
https://github.com/multitheftauto/mtasa-blue/pull/5084#issuecomment-5110150334

FileEX:
If we want to fix this at the code level instead of using a Lua script with corrected models, I would propose the following syntax
```lua
nil engineSetVehicleModelFrameHierarchyFixed(string name, bool enabled)
```

Where `name` identifies a specific fix. This allows us to fix, for example, only the missing wheel on the DFT-30 while keeping the sweeper's brushes disabled if we don't want them to move.

For example
```lua
engineSetVehicleModelFrameHierarchyFixed("dft_wheel", true) -- missing wheel on the DFT-30
engineSetVehicleModelFrameHierarchyFixed("sweeper_brushes", true) -- Sweeper brushes
```

Fernando-A-Rocha:
^^ it would be nice to know all of the things that can be fixed

FileEX:
We could also add `all` to simply enable all fixes.

> ^^ it would be nice to know all of the things that can be fixed

Looks like there are more of these typos, such as `bump_rear_ok` for model 464, `taillights2`, etc. https://wiki.preview.multitheftauto.com/reference/ID_Lists/Vehicle_Components

PR #4930, opened 2026-06-01
Fix rhino middle wheel rendering

### LINKED ISSUE #4929: Rhino (model 432) renders extra middle wheels at wheel_lm_dummy / wheel_rm_dummy after PR #4714
### Describe the bug

In vanilla GTA: San Andreas, the Rhino tank (model 432) carries six wheel dummies (wheel_lf, wheel_lm, wheel_lb, wheel_rf, wheel_rm, wheel_rb) but the two middle wheels (lm/rm) are intentionally not rendered. `CAutomobile::SetupModelNodes` (at game address `0x6B0CF0`) clears `rpATOMICRENDER` on the cloned wheel atomics attached to those dummies once at vehicle init.

Since PR #4714 ("Allow the Rhino to render headlight cones and coronas", merged in commit b6902774), the middle-wheel atomics become visible on every Rhino, on a clean MTA install, without any custom resources, handling changes, or replacement models. The artefact is two extra wheel sprites overlapping the tread between the front and rear wheels on each side. Headlights and coronas (the intended effect of the PR) still work and should remain working after the fix.

**Root cause.** PR #4714 NOPs 6 bytes at two addresses in `Client/multiplayer_sa/CMultiplayerSA.cpp` (lines 1564-1566):

- `0x6A2EAB` inside `CAutomobile::Render` (`0x6A2B10`): gates the standard headlight-cone / corona draw. Safe to leave NOPed.
- `0x6ABC81` inside `CAutomobile::PreRender` (`0x6AAB50`): gates a broader "standard car PreRender" block. **This block also walks wheel atomics and re-asserts `rpATOMICRENDER` every frame** (same pattern as `CVehicleModelInfo::HideAllComponentsAtomicCB` at `0x4C7790`, whose `else` branch unconditionally sets `rpATOMICRENDER` on atomics that do not match the damage mask). With the guard removed, the Rhino now flows through that block every frame, silently undoing the one-time hide done in `SetupModelNodes` for the middle-wheel atomics.

`CVehicleModelInfo::SetClump` clones the main wheel atomic onto every wheel dummy on the Rhino (see `Models/VehicleModelInfo.cpp` lines 651-680 in `gta-reversed-modern`), so the LM/RM frames already carry a real renderable atomic; only the `rpATOMICRENDER` flag was keeping them invisible.

The bug is purely visual. Suspension physics (the Rhino's 12 suspension lines computed by lerping lf↔lb and rf↔rb in `SetupSuspensionLines` near `0x6A6995`) does not read the middle dummy positions and is unaffected. All six wheel dummies must remain in place for tank physics.

**Proposed fix.** Keep both PR #4714 NOPs intact so headlights/coronas keep working. Re-clear `rpATOMICRENDER` on the LM/RM atomics from MTA's existing tail-of-PreRender hook at `0x6ABCFD` (already installed for sun glare in `Client/game_sa/CVehicleSA.cpp:39`). Because that hook is the last code MTA runs inside `CAutomobile::PreRender`, the clear wins against the generic block's per-frame re-enable. Guarded on `m_nModelIndex == 432` (`VehicleType::VT_RHINO`) so it is a no-op for every other vehicle. Cost: one compare per automobile per frame.

Patch sketch (~30 lines added in `Client/game_sa/CVehicleSA.cpp`, no header changes, no new includes — `CAutomobileSA.h` and `gamesa_renderware.h` are already included):

```cpp
static RwObject* __cdecl ClearAtomicRenderFlagCB(RwObject* object, void* /*data*/)
{
    object->flags &= ~0x04; // rpATOMICRENDER
    return object;
}

static void __fastcall RehideRhinoMiddleWheels(CAutomobileSAInterface* vehicle)
{
    if (!vehicle || vehicle->m_nModelIndex != static_cast<unsigned short>(VehicleType::VT_RHINO))
        return;

    for (auto comp : { eCarNodes::WHEEL_LM, eCarNodes::WHEEL_RM })
    {
        if (RwFrame* frame = vehicle->m_aCarNodes[static_cast<std::size_t>(comp)])
            RwFrameForAllObjects(frame, (void*)ClearAtomicRenderFlagCB, nullptr);
    }
}
```

Original Rhino from GTA SA:
Image

### Steps to reproduce

1. Install clean GTA + MTA
2. Create Local Server in MTA
3. Spawn Rhino
4. You will see this:
Image
Image

### Version

Server: v1.6.0-9.22815 (Windows 11, 64-bit)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

David-Vice:
resolves https://github.com/multitheftauto/mtasa-blue/issues/4929

#### Summary

Re-hides the Rhino's (model 432) middle wheels (`wheel_lm_dummy` / `wheel_rm_dummy`)
so the tank renders like it does in single-player GTA:SA again.

The change is a single file, `Client/game_sa/CVehicleSA.cpp`: it adds a small
`__cdecl` RenderWare callback that clears the `rpATOMICRENDER` flag on those two
wheel atomics, and calls it from the existing tail-of-PreRender hook
(`HOOK_Vehicle_PreRender` at `0x6ABCFD`), guarded so it only touches model 432.

Final result:
mta-screen_2026-06-01_05-45-27

#### Motivation

PR #4714 NOP'd a Rhino model-index guard in `CAutomobile::PreRender` so the tank
falls through the standard "car PreRender" path — needed so its headlight cones and
coronas render. A side effect: that path walks the vehicle's wheel atomics and
re-asserts `rpATOMICRENDER` on them every frame, which undoes the one-time hide that
`CAutomobile::SetupModelNodes` applies to the Rhino's middle wheels at init. The
result is that the Rhino now visibly renders its two inner road wheels, which it
should not (they are hidden in vanilla).

This re-clears the flag at the very end of PreRender, so the hide is the last write
of the frame and wins, restoring vanilla appearance while keeping #4714's
headlight/corona behavior fully intact (no changes to `CMultiplayerSA.cpp`).

#### Test plan

- Built the full client from source and confirmed `Client/game_sa/CVehicleSA.cpp`
  compiles cleanly (no new warnings/errors from the added code) and links into
  `game_sa.dll`.
- Ran the self-built client in-game:
  - Spawned a Rhino (model 432): the two middle wheels are no longer rendered,
    matching single-player GTA:SA.
  - Confirmed the Rhino's headlight cones and coronas still render at night
    (i.e. PR #4714's behavior is unaffected).
  - Spawned regular vehicles: unaffected, since the re-hide is guarded by
    `m_nModelIndex == 432`.
  - Respawned the Rhino several times: the middle wheels stay hidden every frame.

To re-verify in future: spawn a Rhino, check the inner wheels are invisible while
the headlight cones/coronas still show.

#### Checklist


Lpsd (approved):
Thanks for the PR!

CookiePLMonster:
This is 100% not caused by #4714, the explanation is hallucinated (`CVehicle::DoVehicleLights` does nothing to atomic) and this bug predates the PR it blames for the regression. I suspect this PR hides the real cause of this regression.

Issue #1962, opened 2020-12-30
Support IPv6


Disinterpreter:
**Is your feature request related to a problem? Please describe.**
The Internet evolution. Possibility to create a server for playing with friends.

**Describe the solution you'd like**
Support IPv6 (subj)

**Describe alternatives you've considered**
IPv4 but there aren't allocated addresses anymore.

**Additional context**
https://www.google.com/intl/en/ipv6/statistics.html

Statistic from above in nutshell:

- on December 2020 average percent is 33 i.e every fourth visiting google from IPv6
- adaptation in Germany is 54% in the US - 46%, in India - 56% i.e every second visit to the google from IPv6

Short movie from google: https://www.youtube.com/watch?v=-Uwjt32NvVA

https://bugs.mtasa.com/view.php?id=5700

Pirulax:
Maybe it's a stupid question, but would it be possible to support both ipv4 and v6 at the same time? It should be, right?
This is sadly a mostly netcode a netcode upgrade, for which ccw is responsible. AFAIK he isn't too active, so I presume upgrading it will take a long time.

patrikjuvonen:
Need to also make sure all necessary MTA backend services have IPv6 configured (master server list, etc.)

And with that we can enable support for IPv6 for curl too.

Disinterpreter:
> 
> 
> Maybe it's a stupid question, but would it be possible to support both ipv4 and v6 at the same time? It should be, right?

Sure. Same with the players, if the server supports ipv4 and ipv6, players can connect to both protocols. 

But there is a problem, which depend with ASE, because right now masterlist save IP as 4 char symbols
https://github.com/tederis/ase2json/blob/d368bd63c16a9e82a5b7d0a4e5de6ffd8eb6b919/main.cpp#L29-L40

prnxdev:
I also wanted to write this suggestion because I am working as much as possible with WSL (hate Windows, sorry). It is possible to start MTA server using Linux libriaries but there is one problem: no one knows why but WSL 2 is not porting `localhost` correctly so everytime I do Windows restart I have to manualy find IP address of MTA server (it's not big problem but it is). The most powerfull solution is to connect using IPv6 (typing `mtasa://[::1]:22003` in connect bar should work but it doesn't).

[...]

PR #4940, opened 2026-06-10
Replace EHS HTTP server with cpp-httplib

Lpsd:
#### Summary

This swaps out the built-in web server's underlying HTTP library. The old EHS dependency is removed and replaced with cpp-httplib (a single-header library) - https://github.com/yhirose/cpp-httplib/releases/tag/v0.47.0

Main changes:
- Removed `vendor/ehs` and added `vendor/cpp-httplib`.
- Added `Shared/httpd/Types.h`, a small standalone header providing the HTTP request/response types that the rest of the server code used to get from EHS (HttpRequest, HttpResponse, status codes, form/query/cookie maps, etc). CHTTPD and the resource code keep using the same types and method signatures, so the change stays mostly contained to the web server layer.
- Rewrote `CHTTPD` to drive cpp-httplib: a catch-all handler does resource lookup and dispatch, then translates our HttpRequest/HttpResponse to and from cpp-httplib's request/response objects. Authentication, brute-force protection and connection flood handling are carried over (mostly) unchanged in behaviour.

Things worth noting for reviewers:
- The minimum Windows version is now Windows 10. cpp-httplib requires `_WIN32_WINNT >= 0x0A00`, so the server build define was bumped from 0x601. This drops Windows 7/8 support for the server.
- The `httpthreadcount` config setting still works: the worker pool is sized from it. cpp-httplib handles per-request threading rather than per-connection like EHS did, so the connection gate runs once per connection (cached) instead of once per request to avoid normal page loads tripping the flood limit. The cache is pruned by age on the existing HTTP pulse (which also expires stale logins), so it stays bounded and each new connection is still evaluated.
- The HTTP allocation stats shown in perf stats (active requests/responses/KB) are no longer available, since cpp-httplib doesn't expose those internals. They now report zero. Total bytes sent is still tracked; worth following up.
- No change to SSL/HTTPS. The built-in server has always served plain HTTP: EHS shipped SSL code but it was never compiled in (the SSL source files weren't in the build and `COMPILE_WITH_SSL` was never set), and CHTTPD never requested an HTTPS listener. cpp-httplib does support TLS via OpenSSL, but it's intentionally left disabled here, so the behaviour is identical; also worth following up.

#### Motivation

EHS is unmaintained -- cpp-httplib is actively maintained and header-only!

#### Test plan

Tested locally on a Windows build:
- Default web resources load correctly: tested performancebrowser and resourcebrowser (includes ajax usage).
- HTTP Basic auth works: the login prompt appears for protected resources, correct credentials log in, wrong credentials are rejected, and brute-force protection still kicks in.
- AJAX calls work, including the `/call/<function>` routing used by resources.
- Form and query string data is passed through to Lua page handlers.

#### Checklist


Disinterpreter:
Good idea! I asked for it many times.

Can you check? May in work with #1962 for future?

Lpsd:
> Can you check? May in work with #1962 for future?

If the system, i.e Windows/Linux, supports the protocol then it can/will work (providing our implementation allows it).

Server/mods/deathmatch/logic/CHTTPD.cpp
```diff
@@ -52,76 +72,313 @@ CHTTPD::~CHTTPD()
...
+    httpReq.nSecure = 0;
+
+    // Copy headers, normalizing names to lowercase.
+    // cpp-httplib preserves the original header case (e.g. "Authorization"),
+    // but MTA code (CheckAuthentication, etc.) expects lowercase keys.
+    for (const auto& h : req.headers)
+    {
+        std::string key = h.first;
+        for (auto& c : key)
+            c = static_cast<char>(::tolower(static_cast<unsigned char>(c)));
```
sbx320:
```suggestion
        std::string key = h.first;
        std::transform(key.begin(), key.end(), key.begin(), ::tolower);
```

sbx320 (changes requested):
cpp-httplib states there's no official 32-bit support. Do we just go with it, or drop 32 bit server support?

> 2-bit platforms are NOT supported. Use at your own risk. The library may compile on 32-bit targets, but no security review has been conducted for 32-bit environments. Integer truncation and other 32-bit-specific issues may exist. Security reports that only affect 32-bit platforms will be closed without action. The maintainer does not have access to 32-bit environments for testing or fixing issues. CI includes basic compile checks only, not functional or security testing.

Lpsd:
> Do we just go with it, or drop 32 bit server support?

I'm fine with this, since we're already enforcing W10+ for the 1.7 release on both client and server. Nobody should be using 32-bit W10, as discussed elsewhere the last release for 32-bit W10 was 2021 (IoT). 

Furthermore the stats for client show less than 0.001% of users are on 32-bit Windows.

As for Linux, we have the 32-bit server. Unfortunately I couldn't locate any stats or useful information around usage, but I don't feel like it's worth supporting anymore.

I've added some commits which switch to using the 64-bit server in the client, and prevents 32-bit OS users from using local server functionality (shows a CEGUI dialog on clicking host game / map editor, asking them to upgrade OS). Tested and working fine.

In another PR we should clean up any other 32-bit server references, such as in nightly.msi, and also do the same with our internal build pipeline, ready for 1.7 release (no longer building/shipping 32-bit server binaries).

sbx320 (dismissed):

PR #4942, opened 2026-06-11
Replace tinyxml with tinyxml2

Lpsd:
#### Summary

Replaces the unmaintained tinyxml library with tinyxml2 11.x (master: https://github.com/leethomason/tinyxml2). 

The full XML wrapper layer (CXMLImpl, CXMLFileImpl, CXMLNodeImpl, CXMLAttributeImpl, CXMLAttributesImpl) has been updated to the new API. tinyxml2 ships as a single header and .cpp with no extra dependencies, so it drops straight into vendor/ alongside the existing premake setup.

#### Motivation

tinyxml(1) has not been updated for a long time.

tinyxml2 is actively maintained by the same original author, handles the same XML subset we use, and has a cleaner modern C++ API.

#### Notable differences from the old library

This is not a fully transparent swap -- the two libraries have meaningfully different ownership semantics:

- In tinyxml2, the XMLDocument owns all nodes. Deleting a node from inside a wrapper destructor triggers tinyxml2's own recursive child cleanup, which conflicts with our wrapper tree teardown order. The wrapper destructors now leave element cleanup to the document.
- XMLAttribute in tinyxml2 is not a heap-allocatable object -- it is an intrusive member of the element's attribute list. CXMLAttributeImpl was rewritten to hold a const XMLAttribute* pointer and refresh it after any mutation.
- Windows SDK msxml.h declares a global ::XMLDocument class. Using namespace tinyxml2 in CXMLImpl.cpp caused ambiguity at compile time, fixed by using explicit tinyxml2:: prefixes throughout that file.

#### Test plan

Start the server with default resources and confirm it launches without issues/crashing.

Also worth checking (already tested lightly):
- getResourceInfo() returns correct values (reads from the `CopyNode` settings node)
- xmlLoadFile / xmlLoadString work from Lua scripts
- XML systems (such as ACL) are working, e.g: `aclrequest allow <resource> all` successfully saves the acl.xml

#### Checklist


Shared/XML/CXMLNodeImpl.cpp
```diff
@@ -65,45 +66,46 @@ CXMLNodeImpl::~CXMLNodeImpl()
...
     // Need to delete the node?
+    // The underlying XML element is owned by its document (either a CXMLFileImpl,
+    // an SXMLStringImpl, or a m_standaloneDocument). The document will clean up
+    // all elements when it is destroyed. We only handle wrapper-level cleanup here;
+    // deleting elements from inside their wrapper destructors triggers recursive
+    // deletion in tinyxml2's ~XMLNode() which conflicts with our wrapper tree
+    // deletion order.
     if (m_pNode)
     {
-        // Grab the parent of our node and delete it using that if any to prevent crashing.
```
sbx320, line 70: There seems to be something missing here

Shared/XML/XML.cpp
```diff
@@ -18,8 +18,10 @@ MTAEXPORT CXML* InitXMLInterface(const char* szSaveFlagDirectory)
     // Initialize and do any file recovery as necessary
     CXMLFileImpl::InitFileRecovery(szSaveFlagDirectory);
 
-    // this is required during parsing, see resources project fallout/help.xml @r659
-    TiXmlBase::SetCondenseWhiteSpace(false);
+    // Whitespace is handled differently in tinyxml2 (PRESERVE_WHITESPACE is
```
sbx320: "differently" when only one implementation is present sounds weird (no one reading the code in the future will know we used to use tinyxml1).

sbx320 (changes requested):

Shared/XML/CXMLNodeImpl.cpp
Lpsd: Thanks, I forgot to clean this up - it's dead code; tinyxml2 handles its own cleanup, so this just needs removing

sbx320 (approved):

PR #4947, opened 2026-06-11
Fix crash when freeing a vehicle model while its detached parts exist

### LINKED ISSUE #4911: Crash in EngineRequestModel/EngineFreeModel with blowVehicle
### Describe the bug

It results in a crash when the vehicle is exploded, destroyed, and finally uses engineFreeModel to deallocate it. Testing on repositories is well done, and this current standard result causes a crash. I used destroyElement, setTimer, and engineFreeModel on an object; the result is no crash.

I tested with engineFreeModel before destroyElement, but it still results in a crash

### Steps to reproduce

```lua
local localX, localY, localZ = getElementPosition(localPlayer)
local a = createVehicle(400, localX, localY + 5, localZ)
local c = engineRequestModel("vehicle")
setElementModel(a, c)
setTimer(blowVehicle, 3000, 1, a)
setTimer(destroyElement, 5000, 1, a)
setTimer(engineFreeModel, 7000, 1, c)
```

### Version

Client: v1.7-untested-26582 (Windows 10, 64-bit)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary

Fixes a client crash (access violation in `CVehicleModelInfo::SetVehicleColour`, offset `0x000C84D8`) when a custom vehicle model allocated with `engineRequestModel` is freed after the vehicle exploded.

Detached car parts spawned by `CAutomobile::SpawnFlyingComponent` (doors, bonnet, etc. flying off an exploding vehicle) are temporary game objects that store the model index of the vehicle they came from (`CObject::m_nRefModelIndex` + the `bChangesVehColor` flag), so `CObject::Render` can repaint them with that vehicle's colours and `CObject`'s destructor can release its model reference. These parts outlive the vehicle element, so deallocating the model while they existed made both paths dereference the freed model info.

The fix clears those references from the game's object pool when a vehicle model info is deallocated (`CModelInfoSA::DeallocateModel`), the same way the game marks parts without a source vehicle (`-1`). It also names the two `CObjectSAInterface` fields involved, verified against the documented `CObject` layout.

#### Motivation

Fixes #4911

#### Test plan

Using the script from the issue (explode → destroy → `engineFreeModel`), which crashed reliably before:

* No crash when the model is freed while parts are still flying, nor when freeing before `destroyElement`.
* No crash when the parts despawn naturally afterwards (destructor path).
* Variants tested: freeing immediately after `blowVehicle`, several vehicles sharing the same custom model exploded at once, and re-allocating the same model ID after freeing it.
* Parts of vehicles with standard models still get repainted with the vehicle's colours as usual.

#### Checklist


Client/game_sa/CPoolsSA.cpp
```diff
@@ -381,6 +381,30 @@ void CPoolsSA::DeleteAllObjects()
...
+// Detached car parts (e.g. doors spawned by CAutomobile::SpawnFlyingComponent) keep the model
+// index of the vehicle they came from, so CObject::Render can repaint them with that vehicle's
+// colours and CObject's destructor can release its model reference. Called when a vehicle model
+// info is deallocated, so the parts can't dereference the freed model info anymore.
+//
+void CPoolsSA::ResetDetachedCarPartsRefModel(std::uint16_t usModelID) noexcept
+{
+    CPoolSAInterface<CObjectSAInterface>* pObjectPool = *m_ppObjectPoolInterface;
+
+    for (int i = 0; i < pObjectPool->m_nSize; i++)
```
Lpsd, line 394: I wonder if there is a more efficient way to do this, instead of iterating the entire object pool every time a vehicle is deallocated?

TheCrazy17: This is the cleanest way to fix it without adding complexity, since these detached parts are spawned by GTA's native code and aren't tracked as MTA entities, so the object pool is the only place to reach them. And since it only runs when a vehicle model is deallocated (engineFreeModel) and not every frame, over a pool capped at 1200, it has no measurable performance impact.

Lpsd (approved):

Issue #4873, opened 2026-05-01
Map Editor goes black after stopping the test.


NeblaStuff:
### Describe the bug

The screen turns completely black when you start a new map in the Map Editor.

### Steps to reproduce

1. Go to Map Editor
2. Create a new map
3. Save the map
5. **Test** the map
6. Stop test the map
7. Start new map
8. Screen got black

### Version

Multi Theft Auto v1.7-untested-26577

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

q8X:
~~unable to reproduce, also this issue should be in mtasa-resources~~ nvm i need to add gamemode in order to reproduce this bug

SpeedyFolf:
I don't know if I can reproduce with those exact steps, but I have issues too. I don't think this should be in resources, I'm using the same files for 1.6 and 1.7 and editor is only broken in 1.7, everything works nice and correctly in 1.6. When I open a race map I get spum with info and warnings:
```
[editor]\edf\edf.lua:143: Going to represent map <Some other map that is NOT the one I just opened>
INFO: No definition for object
INFO: No definition for ped
INFO: No definition for removeWorldObject
INFO: No definition for vehicle
WARNING: [editor]\edf\edf.lua:1013: Bad argument @ 'getMarkerType' [Expected marker at argument 1, got boolean] [string "edf/edf.lua"]
WARNING: [editor]\edf\edf.lua:1013: Bad argument @ 'setMarkerTarget' [Expected element at argument 1, got boolean] [string "edf/edf.lua"]
WARNING: [editor]\edf\edf.lua:901: Bad argument @ 'setMarkerTarget' [Expected element at argument 1, got boolean] [string "edf/edf.lua"]
WARNING: [editor]\editor_main\server\saveloadtest_server.lua:269: Bad argument @ 'getMarkerType' [Expected marker at argument 1, got boolean]
WARNING: [editor]\editor_main\server\saveloadtest_server.lua:269: Bad argument @ 'setMarkerTarget' [Expected element at argument 1, got boolean]
WARNING: [editor]\editor_main\server\saveloadtest_server.lua:846: Bad argument @ 'addEventHandler' [Expected element at argument 2, got boolean]
WARNING: Bad argument @ 'setMarkerIcon' [Expected element at argument 1, got boolean] [string "edf/edf.lua"]
```
[...]

PR #4529, opened 2025-11-09
Resolve Event Context Loss in Nested Event Cacnellation

MohabCodeX:
### Problem
When events trigger other events (nested), the parent event would lose its cancellation reason and context.

### Solution 
Replaced single event context pointer with a proper context stack, each event now maintains its own cancellation state and reason.

### Results

**Before**
```
[08:30:50] [debug] [testing]\vbug\server.lua:28 - Testing nested events...
[08:30:50] [debug] [testing]\vbug\server.lua:21 - PASS: Event B has its reason
[08:30:50] [debug] [testing]\vbug\server.lua:12 - FAIL: Expected 'Reason A', got ''
```

**After**
```
[08:30:50] [debug] [testing]\vbug\server.lua:28 - Testing nested events...
[08:30:50] [debug] [testing]\vbug\server.lua:21 - PASS: Event B has its reason
[08:30:50] [debug] [testing]\vbug\server.lua:12 - PASS: Event A kept its reason

```

<details> 
<summary>Testing script</summary>

```lua
addEvent("eventA", true)
addEvent("eventB", true)

addEventHandler("eventA", root, function()
    cancelEvent(true, "Reason A")
    triggerEvent("eventB", root)
    
    local reason = getCancelReason()
    if reason == "Reason A" then
        outputDebugString("PASS: Event A kept its reason")
    else
        outputDebugString("FAIL: Expected 'Reason A', got '" .. reason .. "'")
    end
end)

addEventHandler("eventB", root, function()
    cancelEvent(true, "Reason B")
    
    local reason = getCancelReason()
    if reason == "Reason B" then
        outputDebugString("PASS: Event B has its reason")
    else
        outputDebugString("FAIL: Expected 'Reason B', got '" .. reason .. "'")
    end
end)

addCommandHandler("testevent", function()
    outputDebugString("Testing nested events...")
    triggerEvent("eventA", root)
end)
```

</details>

PR #4965, opened 2026-06-21
Fix lost onPlayerConnect cancel reason (#4963)

### LINKED ISSUE #4963: Cancelling the onPlayerConnect event does not display the reason
### Describe the bug

In 1.7, when you cancel the onPlayerConnect event, the player does not see the custom reason; it simply says "Server refused the connection:". This does not happen in 1.6.
1.7:
Image
1.6:
Image

### Steps to reproduce

Make a resource folder with these files:
meta.xml:
```xml
<meta>
	<min_mta_version server="1.5.8-9.20957" client="1.5.8-9.20957"></min_mta_version>
	<info type="script" shortname="opc" name="opc" longname="opc" author="SpeedyFolf" version="1.0.0"></info>
	<script src="server.lua" type="server" validate="false"></script>
</meta>
```
server.lua:
```lua
addEventHandler("onPlayerConnect", root, function()
	cancelEvent(true, "string reason 1.7")
end )
```

### Version

Client: v1.7-untested-26637 (Windows 10, 64-bit)
Server: v1.7-untested-26633 (Windows 10, 64-bit)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary

Fixes the lost `onPlayerConnect` cancel reason by storing the reason captured during `CallEvent()` and using it when disconnecting the player.

#### Motivation

`CallEvent()` finishes by running `PostEventPulse()`, which restores `m_strLastError` to the value it had before the event call. Because `PlayerCompleteConnect()` reads the cancel reason after `CallEvent()` returns, it could no longer access the reason passed to `cancelEvent()`. This caused players cancelled from `onPlayerConnect` to receive the generic “Server refused the connection” message instead of the scripted reason.

Fixes #4963.

image

#### Test plan

Tested by cancelling `onPlayerConnect` with a custom reason and confirming the disconnected player receives that reason instead of the generic fallback message.

Server-side test resource:

```lua
addEventHandler("onPlayerConnect", root, function()
    cancelEvent(true, "string reason 1.7")
end)
```

(Same as the issue)

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/CEvents.cpp
+++ b/Server/mods/deathmatch/logic/CEvents.cpp
@@ -141,7 +141,8 @@ void CEvents::PostEventPulse()
     m_bWasEventCancelled = m_bEventCancelled;
     m_bEventCancelled = m_CancelledList.back() ? true : false;
     m_CancelledList.pop_back();
-    // Restore the outer event's cancel reason that was active before this pulse.
+    // Mirrors m_bWasEventCancelled above, before restoring the outer reason.
+    m_strWasLastError = m_strLastError;
     m_strLastError = m_LastErrorList.back();
     m_LastErrorList.pop_back();
 }
@@ -165,4 +166,9 @@ bool CEvents::WasEventCancelled()
 const char* CEvents::GetLastError()
 {
     return m_strLastError;
+}
+
+const char* CEvents::WasLastError()
+{
+    return m_strWasLastError;
 }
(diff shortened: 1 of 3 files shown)


Lpsd:
Are you able to confirm where/when this behaviour was broken? In which commit?

TheCrazy17:
> Are you able to confirm where/when this behaviour was broken? In which commit?

The regression appears to be caused by #4877, although #4529 and #4873 may also be involved since they modified related event handling logic and seem to be part of the sequence of changes that led to the issue.

Lpsd:
yeah #4877 was basically a revert due to issues caused by #4529 (technically only added a single line, after the revert) - but a slight oversight for onPlayerConnect (error vs cancelled). thanks!

Lpsd (approved):

PR #4966, opened 2026-06-21
Fix repeated aircraft wreck components when engine autostart is disabled

### LINKED ISSUE #4916: Blown up airplanes create excess explosions and flying components
### Describe the bug

Blowing up an airplane and touching it afterwards results in a lot of excess flying components and explosions being created.

https://github.com/user-attachments/assets/448d9225-9649-4eaa-ba68-3ec6924ec7ca

### Steps to reproduce

1. Spawn any airplane, in the video I use a Dodo.
2. Blow up the airplane.
3. Touch the airplane and observe tons of flying components being created.

### Version

Client: Multi Theft Auto: San Andreas v1.7-untested-26604
Server: Multi Theft Auto: San Andreas v1.7-untested-26604

### Additional context

This may be related to special world properties `vehicleburnexplosions` and `flyingcomponents`.

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary

Prevents blown aircraft from re-entering the unattended rotor workaround when `vehicle_engine_autostart` is disabled.

This stops wrecked airplanes from repeatedly creating extra flying components and explosion effects after being touched.

#### Motivation

Fixes #4916.

When `vehicle_engine_autostart` is set to `false`, unattended aircraft can be moved into `STATUS_PHYSICS` so rotor behavior can be handled manually. For aircraft that have already blown up, that allowed later contacts with the wreck to keep producing detached components and explosion effects.

image

#### Test plan

Server-side test resource:

```lua
    setWorldSpecialPropertyEnabled("vehicle_engine_autostart", false)
end)
```

Manual test:

1. Start a server with the test resource enabled.
2. Spawn a Dodo or another airplane.
3. Blow up the airplane.
4. Touch/push the wreck.
5. Confirm it no longer creates repeated explosions or excessive flying components.

#### Checklist

### DIFF
--- a/Client/game_sa/CVehicleSA.cpp
+++ b/Client/game_sa/CVehicleSA.cpp
@@ -95,6 +95,13 @@ static bool __fastcall CanProcessFlyingCarStuff(CAutomobileSAInterface* vehicleI
 
     if (vehicle->pEntity->GetVehicleRotorState())
     {
+        // Blown aircraft must not re-enter the custom rotor processing path. With
+        // vehicle_engine_autostart disabled this path moves unattended aircraft to
+        // STATUS_PHYSICS, which lets wreck contacts repeatedly create GTA flying
+        // components/explosions after the vehicle has already blown up.
+        if (vehicle->pEntity->GetHealth() <= 0.0f)
+            return false;
+
         if (g_pCore->GetMultiplayer()->IsVehicleEngineAutoStartEnabled())  // keep default behavior
             return true;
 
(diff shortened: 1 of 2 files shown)


Lpsd:
Thanks, however I had a different solution in mind for this, to fix the issue at the source (sync-wise): https://github.com/Lpsd/mtasa-blue/commit/c2b92c70fa9c81fd0462acbd7e0847ba32f7cc37

This way it doesn't matter where damage events come from, if a vehicle is already blown then it won't propagate the damage event, and therefore no duplicate components.

Lpsd:
~~Since your fixes are also valid I'll still accept this PR, it can simply be applied on top.~~

Lpsd:
Actually, are there any further implications to not letting blown up planes enter `STATUS_PHYSICS`? Feels like it could have unintended consequences. 

It also feels like a bit of a hacky solution, not really addressing the root cause (GTA->MTA sync logic, as per my fix).

Lpsd:
Looking into this further, blown vehicles should use `STATUS_WRECKED`, so this is fine - and is actually required on top of my changes to resolve propagation of the issue in all paths.

Lpsd (approved):

PR #4968, opened 2026-06-21
Fix crash from disabling driveby during weapon fire

### LINKED ISSUE #1913: setPedDoingGangDriveby in onClientPlayerWeaponFire crashes game
In a function handled by event **onClientPlayerWeaponFire**
Doing **setPedDoingGangDriveby**(localPlayer, false) will **_crash_** the player's game
Happens all the time

To try this out for yourself:
```lua
local count = 0
function testBug(weapon, ammo, ammoInClip, hitX, hitY, hitZ, hitElement )
	if source == localPlayer then
        
        if count == 5 then
            -- trigger it; Game will crash
            if isPedDoingGangDriveby(localPlayer) then
                setPedDoingGangDriveby(localPlayer, false)
            end
        end
        
        count = count + 1
	end
end
addEventHandler ( "onClientPlayerWeaponFire", root, testBug )
```

TheCrazy17:
### Summary

Defers the native `MakeAbortable()` call for the gang driveby task when `setPedDoingGangDriveby(ped, false)` is called from inside an `onClientPlayerWeaponFire`/`onClientPedWeaponFire` handler, instead of running it immediately.

### Motivation

Calling `setPedDoingGangDriveby(localPlayer, false)` from `onClientPlayerWeaponFire` crashes the game. The event fires synchronously from inside the native `CTaskSimpleGangDriveBy::ProcessPed()` call (itself on the stack of `CPlayerPed::ProcessControl`). `MakeAbortable()` frees/nulls a member of the task object; when execution returns into the still-running `ProcessPed()` higher up the stack, it dereferences that now-null pointer and the game crashes with an access violation.

Closes #1913 (Crash issues from 2020 lol)

### Test plan

- Reproduced the crash with the original repro script (toggling `setPedDoingGangDriveby(localPlayer, false)` from `onClientPlayerWeaponFire` while actively firing in driveby).
- Applied the fix and confirmed the same script no longer crashes; the driveby task is aborted on the next pulse instead, and the ped correctly stops doing the driveby animation.

### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -5272,6 +5272,8 @@ void CClientGame::PostWeaponFire()
                 }
                 else
                     pPed->CallEvent("onClientPedWeaponFire", Arguments, true);
+
+                // Make sure to reset the weapon fire event state
                 pPed->SetProcessingWeaponFireEvent(false);
             }
             pPed->PostWeaponFire();
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -5259,6 +5259,8 @@ void CClientGame::PostWeaponFire()
                 else
                     Arguments.PushNil();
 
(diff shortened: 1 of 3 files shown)


Lpsd (approved):
Thanks!

PR #4977, opened 2026-06-22
Remove 32-bit server support & enforce W10+

Lpsd:
#### Summary
This PR removes 32-bit server support from the build system in preparation for the EHS to cpp-httplib replacement (#4940).

Changes include dropping Linux x86 server platforms, removing Windows 32-bit server project builds, bumping the server minimum Windows version to 10, and cleaning up dead code for pre-Win10 OS compatibility paths in the client.

#### Motivation
cpp-httplib does not officially support 32-bit platforms.
Rather than maintain unsupported configurations we are dropping 32-bit server builds entirely, see additional reasoning: https://github.com/multitheftauto/mtasa-blue/pull/4940#issuecomment-4762032798
The client remains 32-bit (obviously) and the installer gate for pre-Win10 is already in place for 1.7 release.
This cleanup ensures no 32-bit server binaries are built going forward.

#### Test plan
Verified that premake5 generates correct project files for both Windows (x86 x64 arm64) and Linux (x64 arm arm64) without the x86 Linux platform.
Confirmed that Windows x86 builds still compile the client while server projects are excluded via ExcludeFromBuild.
The server launcher now checks for Windows 10 or later at startup and the installer no longer offers the 32-bit server executable or Win7 SHA2 KB patches.

#### Checklist


Client/core/CSettings.cpp
```diff
@@ -1864,33 +1861,6 @@ void CSettings::CreateGUI()
     vecTemp.fX = 22.f;
     vecTemp.fY += fLineHeight;
 
-    // Windows 8 compatibility
-    m_pWin8Label = reinterpret_cast<CGUILabel*>(pManager->CreateLabel(pTabAdvanced, _("Windows 8 compatibility:")));
```
sbx320, line 1868: I'm not entirely sure if these should be removed. Looks like it was added for _forward_ compatiiblity with win8. 

https://github.com/Lpsd/mtasa-blue/commit/e619d6f9746251814b4c476ce87615929deef4ac

Lpsd: These options are not visible for users running W10+

sbx320 (approved):

Disinterpreter:
It will introduce several issues for our MTA AMX project https://github.com/multitheftauto/amx:

1. amx module currently supports and were tested only on 32-bit mta server. Adding 64-bit binaries after possible tweaks for 64-bit environment can be done as extra work, but ->
2. ALL samp plugins ever made are 32-bit (since there were always only 32-bit samp server) and will lost their current support. And this is crucial, because MTA AMX aim is to run both unmodified pawn scripts and unmodified samp plugins, and the first usually relies on the second as on core dependencies to properly work

One more minor point is that there's currently an easy way to install MTA AMX project for any samp scripter who can just download mta client with 32-bit server built in into the shared installer. Ditching 32-bit mta server will probably require excluding it from the client installer too

Shared/installer/nightly.nsi
```diff
@@ -893,7 +878,6 @@ SectionGroup /e "$(INST_SEC_SERVER)" SECGSERVER
         SetOverwrite on
         File "${SERVER_FILES_ROOT}\core.dll"
         File "${FILES_ROOT}\mta\xmll.dll"
-        File "${SERVER_FILES_ROOT}\MTA Server.exe"
         File "${SERVER_FILES_ROOT}\net.dll"
```
sbx320: The other x86 files in SERVER_FILES_ROOT should also be removed.

sbx320 (changes requested):

Lpsd:
@Disinterpreter as discussed in the development Discord (#amx channel) we're going to 'deprecate' amx for now, it will continue to work in 1.6 versions, there are a couple of ways to move forward however they would delay 1.7 release.

Xenius97:
As far i remember, Map Editor & Host Game are using 32-bit server too. If this will be removed, those won't work too.

Lpsd:
> As far i remember, Map Editor & Host Game are using 32-bit server too. If this will be removed, those won't work too.

Already tested and working fine with the x64 server. `WOW64` does its job.

See PR #4940 for more context

Xenius97:
> Already tested and working fine with the x64 server. WOW64 does its job.

But x64 binaries are not included by default.

Lpsd:
> But x64 binaries are not included by default.

Yes they are. The client installer has server binaries selected by default, and there is no differentiation between x32 and x64, it gives you them all.

Lpsd:
Ok nevermind apparently the current client installer doesn't ship x64 you are right. But it will work that way in 1.7

PR #4978, opened 2026-06-22
Fix/restream streamout collision gap

naouh:
#### Summary

`StreamOutForABit` is used for deliberate restreams (engineReplaceModel, and
model restore via RestreamObjects / RestreamPeds / RestreamVehicles / etc).
It is documented to stream the element back in on the next frame.

Since #4744 added an anti-thrash cooldown (`minStreamInDelayAfterOutMs`, 1200ms)
before an element can stream back in after a stream-out, these deliberate
restreams were blocked for over a second too.

This clears `m_lastStreamOutTime` in `StreamOutForABit` so the restream path is
exempt from the cooldown and streams back in right away again. The distance
based stream-out keeps its timestamp, so the #4744 ping-pong fix stays intact.

#### Motivation

After #4744, replacing or restoring a model at runtime (for example
engineReplaceModel on a road/object, or destroyElement on a custom DFF) keeps
the affected objects streamed out for about 1.2 seconds. During that gap the
collision is gone, so any player standing on or driving over that model falls
through the world. It used to be instant before #4744.

This is a regression from #4744.

#### Test plan

1. Have the player stand on (or drive over) an object model, for example a road piece.
2. At runtime, call `engineReplaceModel` on that model id, then `destroyElement`
   the replacement DFF to restore the original.
3. Before this PR: the object stays streamed out for about 1.2s, and the player
   falls through the object during the gap.
4. After this PR: the object streams back in on the next frame, collision stays,
   no fall-through.

Before:
https://streamable.com/4gyu8i

After:
https://streamable.com/z3x3n4

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientStreamElement.cpp
+++ b/Client/mods/deathmatch/logic/CClientStreamElement.cpp
@@ -148,6 +148,15 @@ void CClientStreamElement::StreamOutForABit()
 {
     // Remove asap, very messy
     InternalStreamOut();
+
+    // This is a deliberate, temporary stream-out (e.g. engineReplaceModel / model restore
+    // via RestreamObjects/RestreamPeds/etc). Unlike a distance-based stream-out it must be
+    // allowed to stream back in immediately. Otherwise the anti-thrash cooldown added in #4744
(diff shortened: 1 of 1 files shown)


Lpsd (approved):
Thank you!

PR #4985, opened 2026-06-27
setObjectScale: add optional scaleCollision argument to match collision to visual scale (collision scale)

### LINKED ISSUE #3775: Re-scale object collision automatically

When using `setObjectScale(objElement, 2)`, the object's collision is not altered. It would be extremely useful to have it also scale up/down in all axis accordingly.

https://wiki.multitheftauto.com/wiki/SetObjectScale

Add argument in the function to auto scale the collision according to the values passed.

bool setObjectScale ( object theObject, float scale [, float scaleY = scale, float scaleZ = scale, bool scaleCollision = false ] )

-

### Additional context

SA-MP and many other games have had this feature for a long time now.

### Security Policy

TheCrazy17:
#### Summary

Adds an optional `scaleCollision` boolean to `setObjectScale` (client and server-side). When `true`, the collision model is cloned and scaled to match the visual.

```lua
setObjectScale(object, scale [, scaleY, scaleZ, scaleCollision = false])
```

Defaults to false for full backward compatibility, so existing scripts are unaffected.

mta-screen_2026-06-26_23-06-51
mta-screen_2026-06-26_23-07-03

#### Motivation

Closes #3775.

Objects scaled visually kept their original collision size, causing an obvious mismatch (e.g. a giant object you can walk through). The argument is opt-in rather than automatic to avoid breaking existing scripts that rely on the old behavior, for example, some DM maps scale objects to make them non-solid intentionally.

Open to feedback and improvements on the implementation approach.

#### Test plan
setObjectScale(obj, 3, 4, 1, true) -> collision matches the enlarged visual (stand on/collide with it).
setObjectScale(obj, 3, 4, 1) (omitted) -> collision unchanged, backward compatible.
Scale an object the same frame it is created -> works correctly.
Disconnect and reconnect with a scaled object -> collision is restored.
Server sets scale with scaleCollision = true -> clients receive and apply it.
Existing scripts using setObjectScale without the new argument, behavior identical to before.

#### Checklist

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1744,6 +1744,12 @@ void CModelInfoSA::RestoreColModel()
     m_originalFlags = 0;
 }
 
+CColModelSAInterface* CModelInfoSA::GetColModelInterface()
+{
+    m_pInterface = ppModelInfo[m_dwModelID];
+    return m_pInterface ? m_pInterface->pColModel : nullptr;
+}
+
 void CModelInfoSA::MakeCustomModel()
 {
     // We have a custom model?
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -1603,8 +1603,14 @@ void CModelInfoSA::SetColModel(CColModel* pColModel)
     if (!pColModelInterface)
         return;
 
-    // Skip setting if already done
-    if (m_pCustomColModel == pColModel)
+    // Skip only if this col is both already recorded as our custom one AND still actually applied to
+    // the live model interface. We must NOT early-out merely because m_pCustomColModel matches:
+    // MakeCustomModel() re-invokes us right after the model streams in specifically to re-apply the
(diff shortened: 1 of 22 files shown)


Fernando-A-Rocha:
This is in the Top 10 of most wanted MTA features in the last decade :)

Dryxio:
Could you test changing an object’s model while scaled collision is active?

Dryxio:
Also if you use it on breakable objects, they loose their breakable properties, probably same for other properties like tobj, dynamic objects etc

TheCrazy17:
> Could you test changing an object’s model while scaled collision is active?

Now it works as it should

> Also if you use it on breakable objects, they loose their breakable properties, probably same for other properties like tobj, dynamic objects etc

Fixed

RushCuttle:
Tested this and besides a small crash that was already reported (collisionless objects crashing when scale collisions is set to true) to Crazy I can say it'll be a very cool new feature to play with: 
https://youtu.be/7qv1b-DFLzc
https://www.youtube.com/watch?v=B9MotABaJuE

PR #4996, opened 2026-07-03
Fix neon texture rendering regression

### LINKED ISSUE #425: Objects with alpha below 141 are invisible

TheCrazy17:
#### Summary
Restores the original `ALPHAREF=100` value in `CVisibilityPlugins::RenderEntity` (byte at `0x732C2F`), which MTA had previously lowered to `1` to fix #425. To keep compatibility with that fix, this PR adds a per-entity override in `OnMY_CEntity_RenderOneNonRoad_Pre`: whenever an entity's material alpha is set below `100` via `setElementAlpha`, `ALPHAREF` is temporarily overridden to `0` for that specific entity before rendering.

mta-screen_2026-07-02_21-50-11

#### Motivation
Neon light textures on buildings were rendering incorrectly in MTA, appearing as wide, blurry color bands instead of the thin bright lines seen in vanilla GTA SA.

The issue comes from using `ALPHAREF=1`. With such a low threshold, gradient pixels in the neon textures with alpha values between `1` and `99` pass the alpha test and blend together into a diffuse glow. The original game uses `ALPHAREF=100`, which discards those low-alpha edge pixels and leaves only the bright center of the texture, producing the sharp neon line the game is supposed to display.

The fix for #425 lowered `ALPHAREF` globally so that objects with very low alpha values could still be rendered correctly. However, that change unintentionally affected neon rendering. This PR keeps the original threshold by default and only disables it for entities that actually need low-alpha rendering, handling both cases independently.

#### Test plan
1. Verify that neon light textures on buildings render as thin bright lines, matching vanilla GTA SA.
2. Verify that `setElementAlpha` values below `100` still make objects visible and correctly transparent.
3. Verify that `setElementAlpha` values between `100` and `255` continue to render normally.
4. Verify that fences, vegetation and other alpha-cutout world geometry render without z-buffer artifacts.

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1577,15 +1577,21 @@ void CMultiplayerSA::InitHooks()
     MemCpy((void*)0x53A00A, "\x33\xC0\x90\x90\x90", 5);
 
     // Fix objects with alpha below 141 are invisible (#425)
-    // Original values: 0x553AD9=140, 0x732C2F=100. These are passed to
+    // Original SA values: 0x553AD9=140, 0x732C2F=100. These are passed to
     // RwRenderStateSet(rwRENDERSTATEALPHATESTFUNCTIONREF, value).
-    // When value=0, RenderWare DISABLES alpha testing entirely, causing
-    // fully transparent pixels to write to the z-buffer and block objects
-    // behind transparent surfaces (fences, vegetation, etc).
-    // Using value=1 keeps alpha testing enabled (rejecting only alpha=0
-    // pixels) while preserving the fix for low-alpha entity visibility.
+    //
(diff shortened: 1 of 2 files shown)


Lpsd (approved):
Thanks!

PR #4998, opened 2026-07-03
Use RtlGetVersion() instead of VerifyVersionInfo() to bypass the manifest compatibility shim (SharedUtil::IsWindows10OrGreater)

Lpsd:
Fixes W10 check issue on server introduced by e0d8ccd


Allerek:
LGTM

PR #5000, opened 2026-07-03
Fix vehicle colours breaking on models with many editable materials

### LINKED ISSUE #4982: Vehicle colors 3 and 4 randomly disappear
### Describe the bug

Vehicle colors 3 and 4 randomly disappear after a vehicle streams out and back in, even though setVehicleColor has been executed correctly.

Colors 1 and 2 remain applied correctly, while colors 3 and 4 can randomly reset or disappear after the vehicle is streamed back to the client. The issue appears to be related to vehicle streaming and does not happen consistently every time.

Before:
Image

After:
Image

### Steps to reproduce

1. Spawn a vehicle.
2. Apply all four vehicle colors using setVehicleColor.
3. Ensure colors 3 and/or 4 are different from the default values.
4. Move far enough away for the vehicle to stream out.
5. Return so the vehicle streams back in.
6. Repeat the process several times.
7. Observe that colors 3 and 4 may randomly disappear or reset, while colors 1 and 2 remain correct.

### Version

Multi Theft Auto v1.6-release-24124 (tested on 1.7 is same)

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary

Fix vehicle colours breaking on detailed vehicle models by moving GTA's editable-material restore list to a larger buffer and bounding it.

`CVehicleModelInfo::SetEditableMaterials` saves every material it repaints (colour, remap texture, lights texture/lighting) in a restore list at `0xB4DBE8`. `ResetEditableMaterials` replays that list after the vehicle is rendered. The list holds 256 entries, has no bounds check, and is directly followed by `ms_lightsOn` (`0xB4E3E8`) and `ms_currentCol` (`0xB4E3F0`). When a vehicle writes more than 257 entries, one entry lands on `ms_currentCol` and overwrites it with a material pointer. From then on, every material repainted in that pass takes its colour slot from a pointer byte, and `HOOK_VehColCB` reads `vehColors` out of bounds. Larger overflows also overwrite `ms_pRemapTexture`, the colour table and the lights textures.

This PR:

- points `SetEditableMaterials` / `ResetEditableMaterials` at a 4096-entry list owned by multiplayer_sa (4 immediates patched)
- wraps the per-atomic callback (`0x4C83E0`) so an atomic whose materials could exceed the remaining space is skipped instead of overflowing. A material adds at most 4 entries: the alpha pass, then either remap texture + colour or lights colour + texture + lighting.
- drops the earlier `movzx` slot patch and the `SaveVehColors` change. As Dryxio pointed out, the patch was a no-op. `SetEditableMaterials` has a single caller (`CVehicle::SetupRender`, `0x6D662B`), and it runs right after `HOOK_VehCol` has set `ms_currentCol` to 0..3, so there was no other path for the patch to fix.

#### Motivation

Refs #4982.

FileEX couldn't reproduce the issue with the camper, and that matches this cause: the screenshots in the issue show a detailed custom model (a BMW X5 with a full interior), while stock models stay far below 256 entries. It also explains why the problem changes after streaming out and back in. The pointer bytes that end up in `ms_currentCol` change whenever the model is reloaded, and materials repainted before the overflow point (the body, colours 1 and 2) stay correct.

I haven't confirmed yet that the model from the issue goes over the limit, so this doesn't close the issue yet.

#### Test plan

- [x] A custom vehicle with more than 257 editable material entries: before this change, colours on parts repainted late (e.g. wheels) are wrong or change after streaming out/in. After it, all four colours stay applied.
- [x] Stock vehicles (camper with 4 RGB colours via `setVehicleColor`, palette colours, paintjobs/remaps, lights on/off) render the same as before.

#### Checklist

### DIFF
--- a/Client/core/DXHook/CDirect3DEvents9.cpp
+++ b/Client/core/DXHook/CDirect3DEvents9.cpp
@@ -752,11 +752,16 @@ HRESULT CDirect3DEvents9::OnDrawPrimitive(IDirect3DDevice9* pDevice, IDirect3DDe
     if (!pLayers)
     {
         // No shaders for this texture.
-        // Apply dual-pass alpha rendering when the mesh has alpha blending with z-write enabled,
-        // but only for 3D world geometry (not pre-transformed 2D elements like the radar/HUD).
-        // This prevents transparent pixels (e.g. fence holes, vegetation cutouts) from writing to
-        // the z-buffer and blocking objects behind them — emulating PS2 per-pixel z-write behavior.
-        if (g_pDeviceState->RenderState.ALPHABLENDENABLE && g_pDeviceState->RenderState.ZWRITEENABLE && !g_pDeviceState->VertexDeclState.PositionT)
+        // Dual-pass alpha z-write emulation for blended 3D geometry (not pre-transformed 2D).
+        // Pass 1 z-writes pixels >= threshold, pass 2 blends the rest without z-write, so fence
+        // holes don't occlude what's behind them. only runs when the game has no meaningful alpha
+        // test of its own (ref at the DefinedState baseline of 2 or below): an active test at ref
+        // 100/140 already clips transparent pixels before z-write like vanilla, and forcing the
+        // dual pass over it draws the sub-ref gradient in pass 2, which is what kept the LV neon
+        // lines smeared no matter what the game side ALPHAREF was set to.
+        const bool bGameAlphaTestActive = g_pDeviceState->RenderState.ALPHATESTENABLE && g_pDeviceState->RenderState.ALPHAREF > 2;
+        if (!bGameAlphaTestActive && g_pDeviceState->RenderState.ALPHABLENDENABLE && g_pDeviceState->RenderState.ZWRITEENABLE &&
+            !g_pDeviceState->VertexDeclState.PositionT)
         {
             // Save current alpha test state
             const DWORD dwOrigAlphaTestEnable = g_pDeviceState->RenderState.ALPHATESTENABLE;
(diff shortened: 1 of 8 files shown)


Dryxio:
1. The `Create()` hunk seems redundant. There's already an unconditional `SetColor` further down in the same function ("Restore the color"), and `Create()` is synchronous, so no `SetupRender` can run between the two calls. The scenario described in the comment can't actually happen. I'd just drop this hunk.

2. I don't think the root cause is what the description says. `HOOK_VehCol` already forces the colour args to 0/1/2/3 on the `SetupRender` path, so `ms_currentCol` is already `{0,1,2,3}` there and the `movzx` patch does nothing on that path. My guess is the fix works because it removes the out-of-bounds `vehColors` read on paths that bypass `HOOK_VehCol`.

Besides that looks good!

QueryOfficial:
Good points, thanks — dropped the redundant early Create() SetColor hunk and rewrote the root-cause comment to reflect that the movzx patch matters on non-SetupRender paths, not the SetupRender path where HOOK_VehCol already sets slot indices. Also removed the unnecessary m_colour1–4 sync in CVehicleSA::SetColor.

Zephkek:
thanks chatgpt

QueryOfficial:
Reworked this after the review and @FileEX's test.

@Dryxio you were right, the `movzx` patch did nothing. I checked the exe: `SetEditableMaterials` is only called from `CVehicle::SetupRender` (`0x6D662B`), right after `HOOK_VehCol` sets the slots to 0..3, so there is no other path it could fix. I removed it along with the `SaveVehColors` change, since that branch is effectively unreachable.

What I think is actually happening: GTA's editable-material restore list at `0xB4DBE8` has 256 entries and no bounds check, and `ms_currentCol` sits right behind it. A detailed model like the one in the issue screenshots can overflow it. After that, the materials repainted later get their colour slot from a pointer byte. This would explain why the camper doesn't reproduce it and why it changes after stream in. The PR now moves the list to a larger buffer and skips atomics that wouldn't fit. I updated the title and description to match.

I haven't reproduced it with the reporter's model yet, so I've changed "Fixes" to "Refs #4982" until it's tested. @Xenius97 could you share a model that shows the problem?

PR #5003, opened 2026-07-04
Fix getElementBoundingBox/radius for engineRequestModel custom models (#4925)

### LINKED ISSUE #4925: getElementRadius / getElementBoundingBox always returns the parent's model's results for objects created with requested / custom model ids
### Describe the bug

I was getting weird always the same result for custom objects with radius and bounding box. It appears that if you run either ``getElementRadius`` or ``getElementBoundingBox`` you will always get the parent's radius and/or bounding box instead of the custom model one.

For instance I almost always use the default value for objects as 1337, which has a radius of ``0.79595631361008`` and a bounding box of ``{ -0.30003115534782, -0.40459677577019, -0.64420366287231, 0.32838663458824, 0.3880760371685, 0.6642918586731 }``, and those are the exact same values I get everytime for model ids that were created using ``engineRequestModel``.

### Steps to reproduce

1. Load a custom model for an object (probably peds/vehicles work too)
2. Get their radius/bounding box
3. Profit

### Version

Client: v1.7-untested-26617

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary

Fix `getElementRadius` and `getElementBoundingBox` for custom models created with `engineRequestModel`.

Custom model slots no longer share the parent model’s `pColModel` bounds data. After a successful DFF replace, bounds are recalculated from RenderWare atomic geometry. Visual bounds recalculation is skipped when a custom COL is applied via `engineReplaceCOL`.

Changes are in:
- `Client/game_sa/CModelInfoSA.h`
- `Client/game_sa/CModelInfoSA.cpp`

#### Motivation

Fixes #4925.

Custom models created with `engineRequestModel` always returned the parent model’s radius and bounding box. For object models with the default parent **1337** (trash can), scripts consistently saw the same radius and bounding box values regardless of the replaced DFF.

Root cause:
- `Make*Model` shallow-copied the parent `CBaseModelInfoSAInterface`, so the custom slot shared the parent’s `pColModel`
- `engineReplaceModel` only replaced visual geometry and did not update bounds
- `getElementRadius` / `getElementBoundingBox` read stale/shared collision bounds via `CModelInfo::GetBoundingBox()`

#### Test plan

1. Request a custom object model with `engineRequestModel("object")`, replace the DFF, create an object, and verify `getElementRadius` / `getElementBoundingBox` differ from parent 1337 values.
2. Repeat with an explicit parent ID (e.g. `engineRequestModel("object", 1225)`) and confirm returned bounds match the custom DFF, not the parent.
3. Test with custom vehicle and ped models.
4. Apply `engineReplaceCOL`, verify bounds come from the COL, and confirm a later DFF replace does not overwrite them.
5. Verify parent model 1337 bounds are unchanged after a custom model replace.
6. Confirm `engineFreeModel` does not crash or leak.
7. Replace the DFF before the model is streamed in, then verify bounds are updated after load via `MakeCustomModel`.

#### Checklist

### DIFF
--- a/Client/game_sa/CModelInfoSA.cpp
+++ b/Client/game_sa/CModelInfoSA.cpp
@@ -90,6 +90,129 @@ static void CColAccel_addCacheCol(int idx, const CColModelSAInterface* colModel)
     function(idx, colModel);
 }
 
+namespace
+{
+    CColModelSAInterface* CreateOwnedColModelCopy(CColModelSAInterface* pSource)
+    {
+        if (!pSource)
+            return nullptr;
+
+        CColModelSAInterface* pOwned = new CColModelSAInterface();
+        const DWORD           dwThis = reinterpret_cast<DWORD>(pOwned);
+        const DWORD           dwFunc = FUNC_CColModel_Constructor;
+        // clang-format off
+        __asm
+        {
+            mov     ecx, dwThis
+            call    dwFunc
+        }
+        // clang-format on
+
+        // Copy only the bounding data from the parent. Collision geometry must not be shared
+        // because custom model bounds are updated independently after DFF replacement.
+        pOwned->m_bounds = pSource->m_bounds;
+        pOwned->m_sphere = pSource->m_sphere;
+        pOwned->m_data = nullptr;
+        return pOwned;
+    }
+
+    void DestroyColModelInterface(CColModelSAInterface* pColModel)
+    {
+        if (!pColModel)
+            return;
+
+        const DWORD dwThis = reinterpret_cast<DWORD>(pColModel);
+        const DWORD dwFunc = FUNC_CColModel_Destructor;
+        // clang-format off
+        __asm
+        {
+            mov     ecx, dwThis
+            call    dwFunc
+        }
+        // clang-format on
+        delete pColModel;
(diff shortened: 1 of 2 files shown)


Fernando-A-Rocha:
Suitable fix PR for 1.7 ?

FileEX:
Please resolve conflicts

QueryOfficial:
Thanks for testing this, @FileEX — you're right, and my root-cause analysis was wrong.

I re-traced it: `getElementRadius` / `getElementBoundingBox` read `CModelInfo::GetBoundingBox()`, which comes from the model's `pColModel`. `engineReplaceCOL` on a requested ID goes through `CModelInfoSA::SetColModel`, which gives that slot its own col model, so the bounds are already correct there. With a DFF-only replace the slot keeps the parent's COL — the same thing that happens when you replace only the DFF on a stock ID — so that's expected behaviour, not a bug.

Reading my patch again, it would also have regressed collision: the cloned col model is created with `m_data = nullptr`, so a requested model without `engineReplaceCOL` would lose the collision it currently inherits from its parent.

Closing this. Sorry for the noise.

PR #5004, opened 2026-07-04
Fix white flash when takePlayerScreenShot is used with CEF browsers

### LINKED ISSUE #4955: takePlayerScreenShot causes white flash when a CEF browser is rendered
### Describe the bug

When using `takePlayerScreenShot` from the server, if the client is rendering a CEF browser (`createBrowser` + `dxDrawImage`/render target), the player's screen briefly **flashes white** when the screenshot is taken.

### Expected Behavior

The screenshot should be captured without any visible screen flash.

### Actual Behavior

The screen flashes white for a brief moment whenever the screenshot is taken.

### Steps to reproduce

1. Create a CEF browser on the client.
2. Render the browser to the screen.
3. Call takePlayerScreenShot from the server.
4. Observe a white flash on the client's screen.

### Version

MTA:SA release 1.6.x
Windows 10/11

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary

Remove the pre/post screenshot hooks that cleared live CEF render targets before `takePlayerScreenShot` capture. The back buffer is now grabbed as-is after `DrawPreGUIQueue`, eliminating the visible white flash when a CEF browser is on screen.

Also removes the now-unused `OnPreScreenshot`, `OnPostScreenshot`, `ClearTextures`, and `ClearTexture` code paths.

#### Motivation

When the server calls `takePlayerScreenShot` while a client has a CEF browser displayed via `createBrowser` + `dxDrawImage`, the player's screen briefly flashes white on every capture.

The old flow cleared live CEF textures to white (`0xFF`) before reading the back buffer, then triggered an async `Invalidate` to repaint them. That touched the same textures used for on-screen rendering in the same frame, so the flash was visible to the player.

This change captures the back buffer without modifying live CEF textures. Screenshot upload continues to work; only the visible flash is removed.

Fixes #4955

#### Test plan

Create a CEF browser on the client and draw it with `dxDrawImage`. Call `takePlayerScreenShot` from the server repeatedly and confirm no white flash appears on the client.

Confirm the server still receives the screenshot with correct game content.

Test downscaled capture with `takePlayerScreenShot(player, 320, 240, ...)`.

Regression: F12 local screenshot still works.

Regression: alt-tab / minimized client still returns `MINIMIZED` / `DISABLED` responses as before.

#### Checklist

### DIFF
--- a/Client/cefweb/CWebCore.cpp
+++ b/Client/cefweb/CWebCore.cpp
@@ -776,21 +776,6 @@ bool CWebCore::GetRemoteJavascriptEnabled()
     return bIsRemoteJavascriptEnabled;
 }
 
-void CWebCore::OnPreScreenshot()
-{
-    // Clear all textures
-    ClearTextures();
-}
-
-void CWebCore::OnPostScreenshot()
-{
-    // Re-draw textures
-    for (auto& pWebView : m_WebViews)
-    {
-        pWebView->GetCefBrowser()->GetHost()->Invalidate(CefBrowserHost::PaintElementType::PET_VIEW);
-    }
-}
-
 void CWebCore::OnFPSLimitChange(std::uint16_t fps)
 {
     dassert(g_pCore->GetNetwork() != nullptr);  // Ensure network module is loaded
@@ -833,14 +818,6 @@ void CWebCore::ProcessInputMessage(UINT uMsg, WPARAM wParam, LPARAM lParam)
     }
 
     m_pFocusedWebView->InjectKeyboardEvent(keyEvent);
-}
-
-void CWebCore::ClearTextures()
-{
-    for (auto& pWebBrowser : m_WebViews)
-    {
-        pWebBrowser->ClearTexture();
-    }
 }
 
 bool CWebCore::SetGlobalAudioVolume(float fVolume)
--- a/Client/cefweb/CWebCore.h
+++ b/Client/cefweb/CWebCore.h
@@ -94,13 +94,9 @@ public:
     CWebViewInterface* GetFocusedWebView();
     void               SetFocusedWebView(CWebView* pWebView) { m_pFocusedWebView = pWebView; };
     void               ProcessInputMessage(UINT uMsg, WPARAM wParam, LPARAM lParam);
-    void               ClearTextures();
 
     bool GetRemotePagesEnabled();
     bool GetRemoteJavascriptEnabled();
-
-    void OnPreScreenshot();
-    void OnPostScreenshot();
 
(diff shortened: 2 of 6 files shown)


AlexTMjugador:
It may have been intended by the commit which originally introduced these hooks to intentionally exclude CEF webview contents from screenshots captured by this function, as they can reasonably be considered part of the GUI and may contain more sensitive information than the game scene (usernames or passwords being typed, etc.).

From your PR description alone, it's not clear to me whether you've considered that potentially intentional behavior, or whether your change causes `takePlayerScreenShot` to start capturing CEF webview contents as well (i.e., the removal of the visible flashing is the only side effect).

Could you elaborate on those considerations?

QueryOfficial:
Good point. I did consider whether the old behaviour was intentional.

To clarify my planned approach: with withCef = true, the screenshot includes the browser as it appears on screen. With withCef = false, the capture is taken as if the browser were not there, and the player never sees a flash. That follow-up is ready locally but not committed yet — I wanted to align on the intended behaviour first.

Does MTA want browser content to never appear in screenshots at all? If so, I can adjust accordingly in a follow-up commit.

ffsPLASMA:
> It may have been intended by the commit which originally introduced these hooks to intentionally exclude CEF webview contents from screenshots captured by this function, as they can reasonably be considered part of the GUI and may contain more sensitive information than the game scene (usernames or passwords being typed, etc.).
> 
> From your PR description alone, it's not clear to me whether you've considered that potentially intentional behavior, or whether your change causes `takePlayerScreenShot` to start capturing CEF webview contents as well (i.e., the removal of the visible flashing is the only side effect).
> 
> Could you elaborate on those considerations?

While the idea of censoring certain parts on the clients screen before taking a screenshot is noble, its outdated nowadays. Especially with the tools given in lua to capture all elements (including mta guis) on screen with other means, this "security" approach is kinda outdated.

And given certain servers render CEF on whole screen to draw HUD elements, I dont see any other solution than removing that censoring if this is the root cause.

QueryOfficial:
Same here — I agree the censorship should go, which is why this PR removes it. I kept the follow-up uncommitted because I still think it's worth discussing.

If preferred, I can add a `withCef` parameter:

- `withCef = true` — screenshot includes the browser as shown on screen (default behaviour after removing the old hooks)
- `withCef = false` — screenshot captures the game scene only, excluding browser content, without any visible flash to the player

Happy to proceed either way based on consensus here.

FileEX:
Closed in favor of #5040

PR #5014, opened 2026-07-05
Fix custom ped textures persisting after resource stop/disconnect

### LINKED ISSUE #1877: engineImportTXD() persists through servers
**Describe the bug**
This is a well known bug (or at least it should be because it exists since forever). Changing just the texture of a model via engineImportTXD() (like the popular zday resource does) will make the change persistent even if you connect to another server. This fixes itself after restarting the game.

**To reproduce**
1. Have two servers, one with Zday resource.
2. Load Zday resource, then join that server.
3. Quit that server and join the one without the Zday resource.
4. Change to any zombie skin, like 280
5. You should see that the model 280 is still with the TXD from Zday, even when that server doesn't have it.

**Expected behaviour**
The texture should be back to normal after disconnecting from a server.

**Version**
Since forever (or at least 1.4)

**Additional context**
This could have a potential bigger side effect apart from having the textures loaded in the wrong server, because its probably related to a memory leak and maybe it also persists when changing DFF's, COL's and TXD's at the same time, but it's shown in game because the name of the textures does not match.

I just have seen this with peds and vehicles, but probably it works with objects too. Stopping the resource before disconnecting seems to solve the issue, so maybe someone forgot to revert the changes made when you disconnect from a server.

TheCrazy17:
#### Summary
Fixes custom ped textures/skins applied via `engineImportTXD` not reverting to the original texture when the replacement is removed (resource stop, script `stopResource`, or disconnecting from the server).

Two issues in the ped texture-restream path, both specific to the local player:

- `CClientPedManager::RestreamPeds`/`RestreamAllPeds` gated the model-refresh hack behind `IsStreamedIn()`, which never becomes `true` for the local player (its ped is created via `_CreateLocalModel()`, which doesn't go through the normal `InternalStreamIn()` path that sets that flag). This meant the refresh never even attempted to run when you were wearing the modified skin yourself.
- `CRenderWareSA::ModelInfoTXDRemoveTextures` restored the original textures in the TXD dictionary, but never forced the ped model to drop its currently-loaded clump, unlike `GetModelTexturesInfo` which does this when the replacement is first applied. Without it, an already-rendered ped had no guarantee it would pick the restored textures back up.

#### Motivation
Fixes #1877.

#### Test plan

[zm-skin.zip](https://github.com/user-attachments/files/29677155/zm-skin.zip)

- Loaded a custom TXD onto a worn skin model via `engineLoadTXD` + `engineImportTXD` from a test resource.
- Confirmed the texture applies correctly (unaffected by this change).
- Stopped the resource while still connected: texture now reliably reverts to the original (previously it was inconsistent; sometimes reverted, sometimes stayed applied).
- Disconnected while the custom texture was active and reconnected to a server without the resource: skin now shows correctly instead of keeping the modified texture.
- Verified vehicles/objects are unaffected (they don't share the local-player streaming quirk that caused this).

#### Checklist

### DIFF
--- a/Client/game_sa/CRenderWareSA.TextureReplacing.cpp
+++ b/Client/game_sa/CRenderWareSA.TextureReplacing.cpp
@@ -213,6 +213,14 @@ bool CRenderWareSA::ModelInfoTXDAddTextures(SReplacementTextures* pReplacementTe
 ////////////////////////////////////////////////////////////////
 void CRenderWareSA::ModelInfoTXDRemoveTextures(SReplacementTextures* pReplacementTextures)
 {
+    // Force ped models to drop their loaded clump, mirroring the forced unload in GetModelTexturesInfo, so they pick up the restored textures below
+    for (ushort usModelId : pReplacementTextures->usedInModelIds)
+    {
+        CModelInfo* pModelInfo = pGame->GetModelInfo(usModelId);
+        if (pModelInfo && pModelInfo->GetModelType() == eModelInfoType::PED)
+            ((void(__cdecl*)(unsigned short))FUNC_RemoveModel)(usModelId);
+    }
+
     // For each using txd
     for (uint i = 0; i < pReplacementTextures->perTxdList.size(); i++)
     {
--- a/Client/mods/deathmatch/logic/CClientPedManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientPedManager.cpp
@@ -106,8 +106,8 @@ void CClientPedManager::RestreamPeds(unsigned short usModel)
     {
         pPed = *iter;
 
-        // Streamed in and same vehicle ID?
-        if (pPed->IsStreamedIn() && pPed->GetModel() == usModel)
+        // Streamed in (or the local player, whose m_bStreamedIn never gets set, see CLuaElementDefs::IsElementStreamedIn) and same model ID?
+        if ((pPed->IsStreamedIn() || pPed->IsLocalPlayer()) && pPed->GetModel() == usModel)
         {
             // Stream it out for a while until streamed decides to stream it
             // back in eventually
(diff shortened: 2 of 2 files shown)


Zephkek:
thx claude

QueryOfficial:
> thx claude

What exactly are you trying to achieve with these comments? Instead of mocking people who are actually working, contributing, and trying to improve the project, maybe you should focus on doing something useful yourself.

Trying to belittle people’s work with these passive-aggressive comments just looks petty and, honestly, pathetic. If you have a real technical criticism, say it properly. Otherwise, stop making these unnecessary remarks.

Don’t worry though, I won’t let these comments bother me.

Congratulations, @TheCrazy17.

TheCrazy17:
> thx claude

I didn't see any policy that prohibits using AI tools as assistance. What matters is that the fix is correct and tested, do you have any actual feedback on the code itself? I'm just trying to help out while testing new tools. Otherwise, you're free to open your own pull request, and no one is obligated to accept mine. lol

Zephkek:
There's a difference between using AI and knowing what you're doing when using AI, keep at it though maybe one of these days gpt or claude will get something right and it'll get merged

FileEX:
> CClientPedManager::RestreamPeds/RestreamAllPeds gated the model-refresh hack behind IsStreamedIn(), which never becomes true for the local player (its ped is created via _CreateLocalModel(), which doesn't go through the normal InternalStreamIn() path that sets that flag). This meant the refresh never even attempted to run when you were wearing the modified skin yourself.

**That's not true**
image

> CRenderWareSA::ModelInfoTXDRemoveTextures restored the original textures in the TXD dictionary, but never forced the ped model to drop its currently-loaded clump

This is also a false lead. `CPedSA::SetModelIndex` already calls `DeleteRwObject`, and changing the skin doesn't fix this issue.

This PR currently looks like blind guessing that maybe `RemoveModel` will work, without really knowing why. The correct approach is to determine why the textures are properly removed in some cases but remain in others, and fix the root cause.

PR #5015, opened 2026-07-05
Fix conformPedRotation drift while ped is moving forwards

### LINKED ISSUE #4694: Player rotation with conformPedRotation is not being persistent correctly when moving forwards
### Describe the bug

When **set the rotation** of the localPlayer with **conformPedRotation=true** and the player is **moving forwards**. The rotation is not same after each frame. Some how it is set to another value.
This incorrect value is also visible for remote players, causing the remote players to see the localPlayer rapidly switching between 2 rotations when set with onClientRender on the localPlayer's end.

```Lua
setElementRotation(localPlayer, 0, 0, newRotation, "default", true)
```

Image

### Steps to reproduce

This example puts the camera above the player, makes the ped walk non-stop and rotates the ped.

- It displays the last set rotation 
- What it has become after the frame
- The difference between the two
```Lua
local forceResetCameraRotation = false
local lastRotationSet = 0
addEventHandler("onClientPreRender", root, function()
    local _, _, rotationAfterFrame = getElementRotation(localPlayer)

    setPedControlState(localPlayer, "forwards", true)
    if math.abs(rotationAfterFrame - lastRotationSet) > 2 then
        dxDrawText(
            "diff: " ..
            tostring(math.floor(math.abs(rotationAfterFrame - lastRotationSet) + 0.5)) ..
            " rotationAfterFrame: " ..
            tostring(math.floor(rotationAfterFrame + 0.5)) ..
            " lastRotationSet: " .. tostring(math.floor(lastRotationSet + 0.5)), 400,
            400, 400,
            600,
            tocolor(255, 0, 0, 255),
            2, "default",
            "left", "center")
    end

    local newRotation = getTickCount() % 100000 / 100000 * 360
    setElementRotation(localPlayer, 0, 0, newRotation, "default", true)

    local _, _, rotationAfterSet = getElementRotation(localPlayer)
    lastRotationSet = rotationAfterSet


    if forceResetCameraRotation then
        setCameraTarget(localPlayer)
    else
        setCameraMatrix(x, y, z + 20, x, y, z)
    end
end)
```

### Version

MTA:SA Server v1.6-release-23826
Server version: 1.6.0-9.23826.0

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary

Re-apply script-set ped rotation after GTA movement tasks overwrite it each frame.

When `setElementRotation` / `setPedRotation` (or server `SET_PED_ROTATION`) sets ped rotation, the client now stores that value and restores it at the start of `PostWorldProcessHandler`, after `ProcessPedsAfterPreRender`, and before puresync/keysync writes. This keeps `conformPedRotation=true` stable while the ped is walking on foot and stops remote players from flickering between two rotations.

#### Motivation

Fixes #4694

GTA’s on-foot movement task adjusts ped heading every frame (e.g. ~5° drift while `forwards` is held). Script rotation set in `onClientPreRender` was overwritten before the next frame and before network sync, so `getElementRotation` showed a different value and remote clients saw alternating rotations.

This follows the same idea as frozen ped matrix re-apply: persist the script value and restore it after the engine has run.

#### Test plan

- Repro from #4694: `onClientPreRender`, `setPedControlState(localPlayer, "forwards", true)`, `setElementRotation(localPlayer, 0, 0, rot, "default", true)` — red `diff` text should not appear (or stay ≤1°)
- Two clients: one runs the script above; the other watches the local player — no rapid rotation flicker
- Normal walking without script rotation set — unchanged behaviour
- Enter vehicle / respawn — script override cleared, no stuck rotation
- `setElementFrozen` ped rotation sync still works (regression check for #4566)
- Ran `./utils/clang-format.ps1` on changed C++ files

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -3942,6 +3942,9 @@ void CClientGame::PreWorldProcessHandler()
 
 void CClientGame::PostWorldProcessHandler()
 {
+    if (m_pManager->IsGameLoaded())
+        m_pManager->GetPedManager()->ReapplyScriptRotations();
+
     m_pManager->GetMarkerManager()->DoPulse();
     m_pManager->GetPointLightsManager()->DoPulse();
     m_pManager->GetObjectManager()->DoPulse();
@@ -3958,6 +3961,9 @@ void CClientGame::PostWorldProcessHandler()
 
 void CClientGame::PostWorldProcessPedsAfterPreRenderHandler()
 {
+    if (m_pManager->IsGameLoaded())
+        m_pManager->GetPedManager()->ReapplyScriptRotations();
+
     CLuaArguments Arguments;
     m_pRootEntity->CallEvent("onClientPedsProcessed", Arguments, false);
 }
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -738,6 +738,27 @@ void CClientPed::SetCurrentRotationNew(float fRotation)
     SetRotationRadiansNew(CVector(0, 0, fRotation));
 }
 
+void CClientPed::SetScriptRotationOverride(const CVector& vecRotationRadians, bool bNewWay)
+{
+    m_bHasScriptRotationOverride = true;
+    m_bScriptRotationNewWay = bNewWay;
+    m_vecScriptRotation = vecRotationRadians;
+}
+
+void CClientPed::ReapplyScriptRotationIfNeeded()
+{
+    if (!m_bHasScriptRotationOverride || !m_pPlayerPed)
+        return;
+
+    if (GetRealOccupiedVehicle())
+        return;
+
+    if (m_bScriptRotationNewWay)
+        SetRotationRadiansNew(m_vecScriptRotation);
+    else
+        SetRotationRadians(m_vecScriptRotation);
+}
+
 void CClientPed::Spawn(const CVector& vecPosition, float fRotation, unsigned short usModel, unsigned char ucInterior)
 {
     // Remove us from our car
@@ -779,6 +800,7 @@ void CClientPed::Spawn(const CVector& vecPosition, float fRotation, unsigned sho
 
     // Set some states
     SetFrozen(false);
+    ClearScriptRotationOverride();
     Teleport(vecPosition);
     SetCurrentRotationNew(fRotation);
     SetHealth(GetMaxHealth());
--- a/Client/mods/deathmatch/logic/CClientPed.h
+++ b/Client/mods/deathmatch/logic/CClientPed.h
@@ -202,6 +202,11 @@ public:
     void SetRotationRadiansNew(const CVector& vecRotation);
     void SetCurrentRotationNew(float fRotation);
 
+    void SetScriptRotationOverride(const CVector& vecRotationRadians, bool bNewWay);
+    void ClearScriptRotationOverride() noexcept { m_bHasScriptRotationOverride = false; }
+    void ReapplyScriptRotationIfNeeded();
+    bool HasScriptRotationOverride() const noexcept { return m_bHasScriptRotationOverride; }
+
     void Teleport(const CVector& vecPosition);
 
     // This function spawns/respawns this ped in any location. This will force a recreation
@@ -748,6 +753,9 @@ public:
     uint                                     m_uiFrameLastRebuildPlayer;
     bool                                     m_bIsSyncing;
     bool                                     m_shouldRecreate{false};
+    bool                                     m_bHasScriptRotationOverride{};
+    bool                                     m_bScriptRotationNewWay{};
+    CVector                                  m_vecScriptRotation{};
 
     bool             m_bBulletImpactData;
     CClientEntityPtr m_pBulletImpactEntity;
--- a/Client/mods/deathmatch/logic/CClientPedManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientPedManager.cpp
@@ -75,6 +75,12 @@ bool CClientPedManager::Exists(CClientPed* pPed)
     return false;
 }
 
+void CClientPedManager::ReapplyScriptRotations()
+{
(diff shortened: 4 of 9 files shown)


Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3942,6 +3942,9 @@ void CClientGame::PreWorldProcessHandler()
 
 void CClientGame::PostWorldProcessHandler()
 {
+    if (m_pManager->IsGameLoaded())
+        m_pManager->GetPedManager()->ReapplyScriptRotations();
```
FileEX: Since we reach `PostWorldProcessHandler`, the game must already be loaded

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -3958,6 +3961,9 @@ void CClientGame::PostWorldProcessHandler()
 
 void CClientGame::PostWorldProcessPedsAfterPreRenderHandler()
 {
+    if (m_pManager->IsGameLoaded())
```
FileEX: Since we reach `PostWorldProcessPedsAfterPreRenderHandler`, the game must already be loaded.

Client/mods/deathmatch/logic/CClientGame.cpp
FileEX: There's no need to do this twice per frame. `PostWorldProcessPedsAfterPreRenderHandler` is more than sufficient.

Client/mods/deathmatch/logic/CClientPed.cpp
```diff
@@ -738,6 +738,27 @@ void CClientPed::SetCurrentRotationNew(float fRotation)
...
+    if (!m_bHasScriptRotationOverride || !m_pPlayerPed)
+        return;
+
+    if (GetRealOccupiedVehicle())
+        return;
+
+    if (m_bScriptRotationNewWay)
+        SetRotationRadiansNew(m_vecScriptRotation);
+    else
+        SetRotationRadians(m_vecScriptRotation);
```
FileEX: The issue only affects rotation using the new way.

FileEX (changes requested):

FileEX:
GTA calculates rotation using its own method, which isn't compatible with the new way used in MTA. `CPed::CalculateNewVelocity` overwrites the ped's `currentHeading`,
```cpp
		else if(fHeadingDiff < -1.0f*CMaths::Abs(m_fHeadingChangeRateAccel)*fMaxHeadingChange){
			m_fCurrentHeading -= CMaths::Abs(m_fHeadingChangeRateAccel)*fMaxHeadingChange;
			m_fHeadingChangeRateAccel -= PED_HEADING_RATE_ACCEL_ADD*CTimer::GetTimeStep();
			if(m_fHeadingChangeRateAccel < -1.0f)	m_fHeadingChangeRateAccel = -1.0f;
			bTurningAFAP = true;
```
Then `CPed::UpdatePosition` sets the matrix based on `currentHeading`, resulting in the aforementioned difference of around 5 degrees.

Client/mods/deathmatch/logic/CClientGame.cpp
QueryOfficial: Removed, you're right, the hook can't fire before the game is loaded.

Client/mods/deathmatch/logic/CClientGame.cpp
QueryOfficial: Same here, removed. The call in this handler is gone entirely now, see my reply on the "twice per frame" comment.

Client/mods/deathmatch/logic/CClientGame.cpp
QueryOfficial: Agreed, it's down to a single call per frame now. I kept this one and dropped the one in `PostWorldProcessPedsAfterPreRenderHandler` instead though, because of the order the hooks run in:

1. `HOOK_CGame_Process` (0x53C095): `CWorld::Process` -> `PostWorldProcessHandler`, which fires `onClientPreRender` at its end
2. `HOOK_Idle` then runs `IdleHandler` -> `DoPulses2(true)`, which is where puresync gets written with alt pulse order
3. `HOOK_Idle_CWorld_ProcessPedsAfterPreRender` (0x53EA03) only comes after both

The repro in #4694 reads the rotation back in `onClientPreRender`, so if we only restore after `ProcessPedsAfterPreRender` that script still sees the drifted heading, and alt pulse order still syncs it. Restoring right after `CWorld::Process`, so right after `CalculateNewVelocity` / `UpdatePosition` ran, should cover every reader with one call.

Is there anything between the two hooks that changes the heading again? If so I'll move it, or put the second call back.

Client/mods/deathmatch/logic/CClientPed.cpp
QueryOfficial: Done, only the new way stores a rotation now. The old-way branch and `m_bScriptRotationNewWay` are gone, and `setElementRotation` / `setPedRotation` only register it when `conformPedRotation` is true.

QueryOfficial:
Thanks for the review, and for tracking the cause down to `CalculateNewVelocity` / `UpdatePosition`. That made it a lot clearer where the restore belongs. Pushed 520c522 and replied in the threads.

One more thing I fixed while at it, mentioning it because it changes behaviour beyond what you asked for: in my first version the override was never cleared (only on spawn / entering a vehicle), so a single `setElementRotation` call would have pinned the ped's heading forever and the player couldn't turn anymore. It's one-shot now. It gets restored once after the next `CWorld::Process`, and any newer rotation write (sync, spawn, `setElementMatrix`, the old-way setters) drops it through `CClientPed::SetCurrentRotation`. A script that sets the rotation every frame, like the one in #4694, should still get a stable value since it re-registers it each frame.

With a single restore point the extra calls in `CNetAPI`, `CClientVehicle` and `CPedRPCs` weren't needed anymore, so the PR no longer touches those files.

I still need to run the #4694 repro against this revision in-game.

FileEX:
Tested, works.

FileEX (approved):

PR #5027, opened 2026-07-11
Fix isPlayerHudComponentVisible ignoring showhud

### LINKED ISSUE #547: isPlayerHudComponentVisible only works with setPlayerHudComponentVisible

### LINKED ISSUE #5026: isPlayerHudComponentVisible("radar") does not reflect showhud state
### Describe the bug

When the GTA HUD is hidden using the showhud console command, isPlayerHudComponentVisible("radar") still returns true, even though the radar is no longer visible on the screen.

This creates an inconsistency between the actual HUD visibility and the value returned by the API, making it impossible for scripts to reliably determine whether the radar is currently visible.

Expected behavior:
isPlayerHudComponentVisible("radar") should return false when the HUD has been hidden using the showhud command.

Actual behavior:
The function returns true even though the radar is not visible.

Additional information:

It appears that isPlayerHudComponentVisible() only reflects the visibility state set by setPlayerHudComponentVisible() and does not take the global HUD visibility controlled by the showhud console command into account.

### Steps to reproduce

Execute the following console command:

showhud

Run:

outputChatBox(tostring(isPlayerHudComponentVisible("radar")))

The function returns true even though the radar is hidden.

### Version

Client: Multi Theft Auto v1.6-release-24124

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary

`isPlayerHudComponentVisible` now returns `false` while the HUD is hidden with `showhud`.

`CHudSA::IsComponentVisible` returns early when the HUD is disabled, since `showhud` hides the entire HUD rather than individual components. There is no new argument, no signature change and no new state.

The fullscreen player map (F11) hides the HUD through the same `CHud::Disable` call, so the function also returns `false` while the map is open.

#### Motivation

Fixes #5026.
Fixes #547 (the `showhud` case from its reproduction steps).

#### Test plan

Built Game SA (Release|Win32) successfully and ran clang-format on the changed file.

Steps to verify in-game:
1. `crun isPlayerHudComponentVisible("radar")` → `true`
2. `showhud`, then the same call → `false`
3. `showhud` again → `true`
4. `setPlayerHudComponentVisible("radar", false)` → `false`; set it back to `true` → `true`
5. Open the F11 map → `false`; close it → `true`

#### Checklist

### DIFF
--- a/Client/core/CCommandFuncs.cpp
+++ b/Client/core/CCommandFuncs.cpp
@@ -368,9 +368,10 @@ void CCommandFuncs::HUD(const char* szParameters)
 {
     // Toggle based on the user's own showhud preference only, so temporary suppression
     // (e.g. the F11 player map) does not flip the toggle in an unexpected direction
-    int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
-    bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : g_pCore->GetGame()->GetHud()->HasDisableReason(HUD_DISABLE_USER);
-    g_pCore->GetGame()->GetHud()->Disable(!bShow);
+    CHud*      pHud = g_pCore->GetGame()->GetHud();
+    const int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
+    const bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : !pHud->IsEnabled();
+    pHud->Disable(!bShow);
 }
 
 void CCommandFuncs::SaveConfig(const char* szParameters)
--- a/Client/core/CCommandFuncs.cpp
+++ b/Client/core/CCommandFuncs.cpp
@@ -366,8 +366,10 @@ void CCommandFuncs::CopyGTAControls(const char* szParameters)
 
 void CCommandFuncs::HUD(const char* szParameters)
 {
+    // Toggle based on the user's own showhud preference only, so temporary suppression
+    // (e.g. the F11 player map) does not flip the toggle in an unexpected direction
     int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
-    bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : g_pCore->GetGame()->GetHud()->IsDisabled();
+    bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : g_pCore->GetGame()->GetHud()->HasDisableReason(HUD_DISABLE_USER);
     g_pCore->GetGame()->GetHud()->Disable(!bShow);
 }
 
--- a/Client/core/CCommandFuncs.cpp
+++ b/Client/core/CCommandFuncs.cpp
@@ -369,7 +369,7 @@ void CCommandFuncs::HUD(const char* szParameters)
     // Toggle based on the user's own showhud preference only, so temporary suppression
     // (e.g. the F11 player map) does not flip the toggle in an unexpected direction
     CHud*      pHud = g_pCore->GetGame()->GetHud();
-    const int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
+    const int  iCmd = (szParameters && szParameters[0]) ? std::atoi(szParameters) : -1;
     const bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : !pHud->IsEnabled();
     pHud->Disable(!bShow);
 }
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -10,7 +10,6 @@
  *****************************************************************************/
 
 #include "StdInc.h"
-#include <cstring>
 #include "CHudSA.h"
 #include "CGameSA.h"
 #include "CCameraSA.h"
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -10,6 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <cstring>
 #include "CHudSA.h"
 #include "CGameSA.h"
 #include "CCameraSA.h"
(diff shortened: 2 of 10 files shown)


TheCrazy17:
You could also add `Fixes #547` so both issues get closed by the same pull request. 😄

QueryOfficial:
> You could also add `Fixes #547` so both issues get closed by the same pull request. 😄

Thanks @Crazy17 ! Good point — I've updated the PR description to also include "Fixes #547", so both issues get closed by this pull request. 🎉

FileEX:
Conflict with #5028

Client/game_sa/CHudSA.cpp
```diff
@@ -98,6 +98,50 @@ CHudSA::CHudSA()
...
 {
+    // Persistent state requested through the showhud command; temporary suppression owners are tracked separately
+    if (m_isEnabled == !bDisabled)
+        return;
+
+    m_isEnabled = !bDisabled;
+    ApplyVisibilityState();
+}
+
+bool CHudSA::IsDisabled()
```
FileEX: const noexcept

Client/game_sa/CHudSA.cpp
```diff
@@ -179,6 +220,11 @@ void CHudSA::SetComponentVisible(eHudComponent component, bool bVisible)
         // Poke bytes
         uchar* pSrc = (uchar*)(bVisible ? &pComponent->origData : &pComponent->disabledData);
         uchar* pDest = (uchar*)(pComponent->uiDataAddr);
+
+        // Avoid unnecessary unprotect/write cycles when the component is already in the requested state
+        if (memcmp(pDest, pSrc, pComponent->uiDataSize) == 0)
```
FileEX: std::memcmp

Client/game_sa/CHudSA.cpp
```diff
@@ -207,6 +256,37 @@ bool CHudSA::IsComponentVisible(eHudComponent component)
...
+        return false;
+
+    // A child component cannot render while any containing component in its draw path is disabled.
+    // Bound the traversal so malformed component metadata can never loop forever
+    SHudComponent* pComponent = MapFind(m_HudComponentMap, component);
+    for (std::size_t depth = 0; pComponent && depth < m_HudComponentMap.size(); ++depth)
+    {
+        const uchar* pSrc = (const uchar*)(&pComponent->disabledData);
+        const uchar* pDest = (const uchar*)(pComponent->uiDataAddr);
+        if (memcmp(pDest, pSrc, pComponent->uiDataSize) == 0)
```
FileEX: std::memcmp

Client/game_sa/CHudSA.h
```diff
@@ -68,6 +69,8 @@ struct SHudComponent
     DWORD         uiDataSize;
     DWORD         origData;
     DWORD         disabledData;
+    // Components rendered inside another HUD component cannot draw while their parent is hidden
+    eHudComponent parent = HUD_ALL;
```
FileEX:
```cpp
eHudComponent parent{HUD_ALL};
```

Client/sdk/game/CHud.h
```diff
@@ -72,6 +72,15 @@ enum class eFontAlignment : std::uint8_t
     ALIGN_RIGHT,
 };
 
+// Temporary owners that hide the whole HUD without changing the persistent state requested
+// through the showhud command. Each owner releases only its own suppression.
+enum class eHudSuppressionReason : std::uint8_t
```
FileEX: Do you really need this enum?

Client/game_sa/CHudSA.h
```diff
@@ -300,6 +310,13 @@ class CHudSA : public CHud
 private:
     std::map<eHudComponent, SHudComponent> m_HudComponentMap;
 
+    // Persistent visibility requested through the showhud command
+    bool m_isEnabled = true;
```
FileEX:
``` cpp
bool m_isEnabled{true};
bool m_isAppliedDisabled{false};
```

Client/sdk/game/CHud.h
```diff
@@ -133,4 +142,11 @@ class CHud
     virtual bool GetComponentUseCustomAlpha(const eHudComponent& component) const noexcept = 0;
 
     virtual CVector2D GetComponentTextSize(const eHudComponent& component) const = 0;
+
+    // Keep new methods at the end so existing CHud virtual method positions remain stable across client modules
```
FileEX: ?

FileEX (changes requested):

QueryOfficial:
Thanks for the review @FileEX — applied your feedback:

- `IsDisabled()` is now `const noexcept`
- Replaced `memcmp` with `std::memcmp`
- Switched member/default init to brace initialization (`parent{HUD_ALL}`, `m_isEnabled{true}`, `m_isAppliedDisabled{false}`)
- Removed the `eHudSuppressionReason` enum — agreed it wasn’t needed. Temporary map suppression is now a single `bool` via `SetSuppressed(bool)`
- Dropped the virtual-method-order comment

Client/game_sa/CHudSA.cpp
```diff
@@ -10,6 +10,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <cstring>
```
FileEX: Is this include needed?

Client/core/CCommandFuncs.cpp
```diff
@@ -366,9 +366,12 @@ void CCommandFuncs::CopyGTAControls(const char* szParameters)
 
 void CCommandFuncs::HUD(const char* szParameters)
 {
-    int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
-    bool bShow = (iCmd == 1) ? true : (iCmd == 0) ? false : g_pCore->GetGame()->GetHud()->IsDisabled();
-    g_pCore->GetGame()->GetHud()->Disable(!bShow);
+    // Toggle based on the user's own showhud preference only, so temporary suppression
+    // (e.g. the F11 player map) does not flip the toggle in an unexpected direction
+    CHud*      pHud = g_pCore->GetGame()->GetHud();
+    const int  iCmd = (szParameters && szParameters[0]) ? atoi(szParameters) : -1;
```
FileEX: ``std::atoi``

QueryOfficial: Updated to std::atoi.

Client/game_sa/CHudSA.cpp
QueryOfficial: You're right, it isn't needed. StdInc already pulls SharedUtil which includes string.h, and that's enough for std::memcmp here. Removed the include.

FileEX:
This PR unnecessarily adds a new argument and overcomplicates the logic. The function should immediately return `false` without checking the component's visibility if the HUD has been hidden using `showhud`, since `showhud` hides the entire HUD rather than individual components.

QueryOfficial:
Thanks @FileEX, you're right. I've removed the new argument and all of the extra HUD state handling.

`CHudSA::IsComponentVisible` now returns `false` right away when the HUD is disabled, so `isPlayerHudComponentVisible` reflects `showhud` without any API change. The PR is down to a single 4-line change in `CHudSA.cpp`, and I've updated the title and description to match.

One thing to note: the F11 player map hides the HUD through the same `Disable` call, so the function returns `false` while the map is open too.

FileEX:
Okay, I tested a few scenarios.

1. Calling `setPlayerHudComponentVisible` while the F11 map is open returns `true`, but the component isn't actually shown.
2. Calling `setPlayerHudComponentVisible` after hiding the HUD with `showhud` also returns `true`, but the component isn't actually shown.

So in my opinion, returning `false` immediately when the HUD is disabled is completely fine, since the component isn't actually visible on the screen, for example when the F11 map is open.

FileEX (approved):

PR #5028, opened 2026-07-11
HUD component visibility state handling

### LINKED ISSUE #5026: isPlayerHudComponentVisible("radar") does not reflect showhud state

HeresHavi:
## **Summary**

Updated `isPlayerHudComponentVisible` to report whether a HUD component can actually be drawn, including the global `showhud` state and fullscreen map suppression.

Radar child components now follow the parent radar state without losing their own visibility settings.

## **Motivation**

Previously, `isPlayerHudComponentVisible` only checked the state of the requested component. This meant the radar could be hidden by `showhud` while the function still returned `true`.

The fullscreen player map also used the same global HUD state. Closing the map could re-enable a HUD that the player had already hidden with `showhud`.

For example, a resource might display a custom minimap only while the GTA radar is hidden. If the API incorrectly reports that the radar is still visible, the resource cannot reliably decide when to show its replacement.

Fixes #5026.

## Test plan

**Runtime**

- Before Fix: Running `showhud` hid the radar, but `isPlayerHudComponentVisible("radar")` still returned `true`.
- After Fix: The function returned `false` while the HUD was hidden and returned `true` after restoring it.
- Valid Case: Hiding the radar directly, toggling `showhud`, and restoring the HUD kept the radar hidden until it was explicitly restored.
- Player Map: Opening the fullscreen map temporarily hid the HUD without replacing the state requested through `showhud`.
- Radar Children: `radar_map`, `radar_blips`, and `radar_altimeter` followed the parent radar state while keeping their own visibility settings.
- Reset: Reconnecting restored the global HUD and component visibility defaults.

**Builds and Tests**

- `Debug | Win32`: Client build passed.
- `Release | Win32`: Client build passed.
- Client tests: 304 tests passed.
- Ran `clang-format`.

## Checklist

### DIFF
--- a/Client/game_sa/CHudSA.cpp
+++ b/Client/game_sa/CHudSA.cpp
@@ -96,23 +96,57 @@ CHudSA::CHudSA()
     componentProperties.wanted = MapGet(defaultComponentProperties, HUD_WANTED);
 }
 
-void CHudSA::Disable(bool bDisabled)
+void CHudSA::Disable(bool disabled)
 {
-    if (bDisabled)
-        MemPut<BYTE>(FUNC_Draw, 0xC3);
-    else
-        MemPut<BYTE>(FUNC_Draw, 0x80);
+    const bool enabled = !disabled;
+    if (m_isEnabled == enabled)
+        return;
 
-    // Also disable the radar as the above code will not hide it before the local player has spawned
-    if (bDisabled)
-        MemPut<BYTE>(FUNC_DrawRadarPlanB, 0xC3);
-    else
-        MemPut<BYTE>(FUNC_DrawRadarPlanB, 0x83);
+    m_isEnabled = enabled;
+    ApplyVisibilityState();
 }
 
 bool CHudSA::IsDisabled()
 {
-    return *(BYTE*)FUNC_Draw == 0xC3;
+    return !m_isEnabled || m_suppressionReasons.any();
+}
+
+void CHudSA::SetSuppressed(eHudSuppressionReason reason, bool suppressed)
+{
+    const std::size_t index = static_cast<std::size_t>(reason);
+    if (index >= m_suppressionReasons.size() || m_suppressionReasons[index] == suppressed)
+        return;
+
+    m_suppressionReasons.set(index, suppressed);
+    ApplyVisibilityState();
+}
+
+void CHudSA::ResetVisibilityState()
+{
+    m_isEnabled = true;
+    m_suppressionReasons.reset();
+    ApplyVisibilityState();
+}
+
+void CHudSA::ApplyVisibilityState()
+{
+    const bool disabled = IsDisabled();
+    // Avoid rewriting GTA code when overlapping owners do not change the effective HUD state.
+    if (m_isAppliedDisabled == disabled)
+        return;
+
+    if (disabled)
+        MemPut<BYTE>(FUNC_Draw, 0xC3);
+    else
+        MemPut<BYTE>(FUNC_Draw, 0x80);
+
+    // The regular HUD draw patch does not cover the fallback radar path used before the local player spawns.
+    if (disabled)
+        MemPut<BYTE>(FUNC_DrawRadarPlanB, 0xC3);
+    else
+        MemPut<BYTE>(FUNC_DrawRadarPlanB, 0x83);
+
+    m_isAppliedDisabled = disabled;
 }
 
 //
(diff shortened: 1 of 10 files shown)


Client/game_sa/CHudSA.cpp
```diff
@@ -96,23 +96,57 @@ CHudSA::CHudSA()
...
-    if (bDisabled)
+    const bool enabled = !disabled;
+    if (m_isEnabled == enabled)
+        return;
+
+    m_isEnabled = enabled;
+    ApplyVisibilityState();
+}
+
+bool CHudSA::IsDisabled()
```
FileEX: const noexcept

Client/game_sa/CHudSA.cpp
```diff
@@ -194,16 +228,25 @@ void CHudSA::SetComponentVisible(eHudComponent component, bool bVisible)
...
     {
-        // Determine if invisible by matching data with disabled values
-        uchar* pSrc = (uchar*)(&pComponent->disabledData);
-        uchar* pDest = (uchar*)(pComponent->uiDataAddr);
-        if (memcmp(pDest, pSrc, pComponent->uiDataSize) == 0)
-            return false;  // Matches disabled bytes
-        return true;
+        const auto* disabledData = reinterpret_cast<const uchar*>(&currentComponent->disabledData);
+        const auto* currentData = reinterpret_cast<const uchar*>(currentComponent->uiDataAddr);
+        if (memcmp(currentData, disabledData, currentComponent->uiDataSize) == 0)
```
FileEX: std::memcmp

Client/game_sa/CHudSA.h
```diff
@@ -300,6 +308,11 @@ class CHudSA : public CHud
 private:
     std::map<eHudComponent, SHudComponent> m_HudComponentMap;
 
+    bool m_isEnabled = true;
+    bool m_isAppliedDisabled = false;
```
FileEX:
```cpp
bool m_isEnabled{true};
bool m_isAppliedDisabled{false};
```

Client/game_sa/CHudSA.h
```diff
@@ -68,6 +69,8 @@ struct SHudComponent
     DWORD         uiDataSize;
     DWORD         origData;
     DWORD         disabledData;
+    // Components rendered inside another HUD component inherit their parent's visibility.
+    eHudComponent parent = HUD_ALL;
```
FileEX:
```cpp
eHudComponent parent{HUD_ALL};
```

Client/sdk/game/CHud.h
```diff
@@ -57,6 +57,13 @@ enum class eHudComponentProperty
     ALL_PROPERTIES,
 };
 
+enum class eHudSuppressionReason : std::uint8_t
```
FileEX: Do you really need this enum?

Client/sdk/game/CHud.h
```diff
@@ -133,4 +140,9 @@ class CHud
     virtual bool GetComponentUseCustomAlpha(const eHudComponent& component) const noexcept = 0;
 
     virtual CVector2D GetComponentTextSize(const eHudComponent& component) const = 0;
+
+    // Keep new methods at the end so existing CHud virtual method positions remain stable across client modules.
```
FileEX: ?

FileEX (changes requested):

FileEX:
What makes this PR better than #5027 ?

HeresHavi:
@FileEX Apologies, this PR was meant to stay as a draft because I wasn’t done finalizing it yet. I’ve prioritized it and pushed the remaining changes so it doesn’t become stale

I've incorporated your suggestions and made some further improvements to how HUD visibility is handled. Compared with the other PR, I think this approach is more complete because it fixes the underlying issue while preserving the existing behavior for scripts. It also keeps the requested HUD state separate from temporary suppression and effective visibility, which avoids changing the meaning of the existing API

FileEX:
At the moment, your PR replicates the behavior of #5027, including adding a new argument to the function 👀

HeresHavi:
> At the moment, your PR replicates the behavior of #5027, including adding a new argument to the function 👀

I compared it with mine just now, and at this point they provide essentially the same API and behavior. Though mine is somewhat smaller and avoids the suppression enum, bitset, and additional state methods, but that is an implementation trade off rather than a clear advantage over #5027. I don't think the remaining differences are enough to justify using my PR.

There are still a couple of smaller implementation details from this PR that may be useful there such as:
- #5028 uses std::memcmp / #5027 uses unqualified memcmp
- #5028 uses reinterpret_cast for the HUD byte pointers / #5027 uses C style pointer casts
- #5028 uses brace initialization such as parent{HUD_ALL} and m_isAppliedDisabled{false} / #5027 uses parent = HUD_ALL and m_isAppliedDisabled = false

Basically your PR suggestions.

You can close this one in favor of #5027, I don't mind it!

FileEX:
Closing in favor of #5027

PR #5030, opened 2026-07-11
Preserve client console scroll position when trimming history

HeresHavi:
## **Summary**

Preserved the visible F8 console position when new output causes older history to be removed.

The console still follows new output while viewing the bottom, and trimming now handles wrapped and multibyte text correctly.

## **Motivation**

The console keeps a limited amount of history. When that limit was reached, older lines were removed from the top while the previous scrollbar position was restored unchanged. This made the visible content jump forward even though the user had scrolled back to read earlier output.

For example, a developer may scroll up to inspect an earlier error while a busy resource continues printing messages. Once the history starts trimming, the console could move away from the lines they were reading.

## Test plan

**Runtime**

- Before Fix: Filled the console with numbered lines, scrolled to a known position, and added timed output. The visible content moved when history trimming started.
- After Fix: Repeated the same test and confirmed the same retained text stayed visible.
- Valid Case: The console continued following new output normally while scrolled to the bottom.
- Wrapped Text: Repeated the test with heavily wrapped 300-character lines.
- Unicode: Tested buffer rollover with `é` and `界`. The text stayed intact and the viewport did not move.
- Oldest Lines: Confirmed that when the currently viewed lines had to be removed, the console stayed at the top and showed the earliest retained complete line.

**Builds and Tests**

- `Debug | Win32`: Client tests built, 304 tests passed.
- `Release | Win32`: Full solution build passed, 304 client tests passed.
- `Release | x64`: Server build passed.
- Ran `clang-format`.

## Checklist


Client/sdk/gui/CGUIMemo.h
```diff
@@ -39,4 +39,8 @@ class CGUIMemo : public CGUIElement
     virtual void EnsureCaratIsVisible() = 0;
 
     virtual void SetTextChangedHandler(const GUI_CALLBACK& Callback) = 0;
+
+    // Keep new virtual methods at the end so existing slots remain stable for modules built against older interfaces.
```
FileEX: Delete this AI comment

FileEX (dismissed):

Client/sdk/gui/CGUIMemo.h
HeresHavi: Just so you understand my comment philosophy, I add comments when the reason behind something isn't clear at first glance. I don't want to assume the next person reading the code already knows why it's written that way.

I added this one because someone might move these methods next to the related ones above without knowing that it would shift the existing vtable slots.

If you disagree, sure, I can remove it, not worth discussing further as it's a comment.

FileEX: Can you explain what difference it actually makes where these methods are placed?

HeresHavi: I added it for compatibility reasons because I wasn't sure how MTA builds and ships it's DLLs. Appending the methods avoids shifting existing vtable slots for older compiled callers. If they expect matching versions, then I can remove the comment. Because if we keep it, then everyone is expected to maintain this in other scripts and wouldn't be worth it if it doesn't make sense.

So I'll just remove it if that's the case. Let me know. But I'll continue to add new virtual methods at the end just as rule of thumb, just because I'm used to it

FileEX: QueryOfficial had the exact same comment in one of his PRs.

I don't want to be rude, but if you understood how this works, you'd know that once the vtable changes and the code is rebuilt, every function using this interface is compiled again as well.

This is a thoughtless AI generated comment based on assumptions that don't apply to this project. It doesn't understand the project structure and assumes that the vtable order has to be preserved, even though there is no binary compatibility requirement here.

FileEX (approved):

FileEX:
Tested, works.

PR #5037, opened 2026-07-12
Fix runtime safety and state consistency issues

HeresHavi:
## **Summary**

Fixed nine runtime safety and persistence issues across the client, server, and shared task scheduler.

The changes cover:

- Input and packet validation for resource paths, resource-start names, join data, server queries, and no-cache scripts.
- Safe shutdown and background server-cache saving.
- Backup finalization and retrying failed account saves.

Each fix is kept in a focused commit for easier review.

## **Motivation**

These problems mostly appeared during shutdown, malformed input, temporary database failures, filesystem errors, or background cache updates. Normal behavior could appear fine while the failure path dropped work, accepted the wrong response, or left saved data in an inconsistent state.

For example, a temporary database lock could cause an account update to be lost instead of retried. A failed backup rename could also be treated as success and allow older working backups to be removed.

## Test plan

**Runtime**

- Connection: Started an isolated x64 server with 206 resources, then connected, disconnected, reconnected, and shut it down normally.
- Join Data: Confirmed the deprecated compatibility field contained exactly 32 zero bytes before a successful join.
- Resource Paths: Rejected empty and incomplete paths on both the client and server while a valid file path still worked.
- Resource Names: Confirmed a 255-byte exported function name worked and a 256-byte name was rejected without breaking the connection.
- Script Packets: Confirmed a normal no-cache script transferred and executed. Truncated, zero, oversized, and mismatched lengths were rejected.
- Server Browser: Sent a response from the wrong port before the correct response and confirmed only the requested endpoint was accepted.
- Server Cache: Saved and reloaded 3,491 entries without invalid or duplicate endpoints.
- Backups: Confirmed a successful backup opened normally. Forced a rename failure and confirmed older backups were preserved.
- Accounts: Forced an asynchronous save to fail with `database is locked`, then confirmed it retried automatically and persisted the updated password.

**Builds and Tests**

- `Debug | Win32`: Clean client and server component builds passed, 305 client tests passed.
- `Release | Win32`: Clean client and server component builds passed, 305 client tests passed.
- `Debug | x64`: Server build passed.
- `Release | x64`: Server build passed.
- Scheduler shutdown test passed 1,000 consecutive runs in both Debug and Release.
- Ran `clang-format`.

## Checklist


FileEX:
The changes are generally fine, but I was just curious and asked Copilot if there were any other things that could be improved in the files you modified, and here's the result.

Would you like to take a look at these issues? If not, that's okay, they won't block this PR from being merged.


**File: Client/mods/deathmatch/logic/CPacketHandler.cpp**
**Lines: 5251–5259**

> [High] Fix the resource path containment check. strResFilePath.rfind(strResPath, 0) == 0 accepts sibling paths that merely share the same textual prefix. For example, a resource directory such as .../resources/foo also matches .../resources/foo2/file.lua. A server-controlled path such as ../foo2/file.lua could therefore escape the intended resource directory. Require a path-component boundary after strResPath, or use a filesystem-relative-path check that rejects paths outside the resource root.

**File: Client/core/ServerBrowser/CServerCache.cpp**
**Lines: 212–240**

> [Medium] Preserve cache changes made while a background save is active. The save thread receives a snapshot and m_bListChanged is cleared once the thread starts. If the map changes during the write, the new dirty state can remain pending after a caller waits for the current save to finish. In particular, shutdown can return after the old snapshot is written while the newer changes are still only in memory. Re-check the dirty state after waiting and perform another save, or use a generation counter/explicit handoff between snapshots.

**File: Server/mods/deathmatch/logic/CGame.cpp**
**Lines: 4957–4981**

> [Medium] Propagate backup input failures before marking the backup complete. The PR checks zipMaker.Close() and FileRename(), but still ignores the return values from every InsertFile() and InsertDirectoryTree() call. An unreadable or disappearing input can result in an incomplete archive that is successfully renamed, reported through NotifyDidBackup(), and followed by pruning of older backups. Track all insertion results and only finalize, notify, and prune when every required input was successfully added

**File: Server/mods/deathmatch/logic/CGame.cpp**
**Lines: 4361–4412**

> [Medium] The screenshot reassembly state accepts unvalidated packet metadata. The new 50 MB limit protects total accumulated memory, but the server does not validate that m_usTotalParts is nonzero, that m_usPartNumber < m_usTotalParts, or that each part keeps the accumulated size consistent with m_uiTotalBytes. A client can therefore create long-lived incomplete transfer state with nonsensical part counts, repeatedly reset transfers, and consume up to the per-player buffer limit until timeout. Validate the metadata when starting a transfer and reject parts whose index or declared total is invalid before appending to info.buffer

FileEX (dismissed):

HeresHavi:
@FileEX 

For the path check, the resource path already ends with a separator. I tested the sibling directory example from the report and it was a false positive, so I left that unchanged.

I also put together separate tests to reproduce the other issues. The cache test uses real threads to check what happens when the data changes during a save. The screenshot tests use the network DLL to check valid and malformed transfers. For backups, I forced read and write failures and checked that the previous backups were preserved.

Those tests reproduced the issues and passed with the changes.

FileEX (approved):

PR #5039, opened 2026-07-12
Fix health and armor puresync validation

### LINKED ISSUE #3791: setPedArmor and setElementHealth synchronization problems from Client to Server
### Describe the bug

Creating anti-cheat checks for a server, we realized something curious.

Although the MTA wiki indicates that the vest that can be placed on a player is from 0-100, this is not real, you can set the number you want (example 1000) from the clientside, from the serverside this is not possible.

Testing this problem, I reached several conclusions.
- The clientside part is currently vulnerable to this problem, you can set setPedArmor values ​​outside the range indicated by the wiki. Therefore, a cheater can use this function to set values ​​higher than those allowed.
- The Server and the Clientside will be out of sync. In the sense that if on the Clientside my armor is 1000 on the server it will be reflected as 127.5 with the getPedArmor function. This also happens with the setElementHealth but it reaches a maximum of 255.
- Armor and life are actually usable, when a person is hurt, the damage of the armor placed on the client side will be reduced. If you shoot him with 1000 armor from a distance, it will leave him with 9990.

Note:
- MTA does a check only on life every 30-60 seconds and this returns to the real maximum. This does not happen with armor
- It is not possible to do this in Health, however when a cheater uses the menu he can edit the Health.

### Steps to reproduce

1. In clientside use setPedArmor

### Version

Client: Multi Theft Auto v1.6-release-22763

### Additional context

Image

### Relevant log output

_No response_

### Security Policy

QueryOfficial:
#### Summary

Clamp player/ped health and armor on both sides of the sync.

Server:

- `CPed::SetHealth()` clamps to `[0, GetMaxHealth()]`, mirroring `CPed::SetArmor()` (#3983)
- `CPlayerPuresyncPacket` and `CVehiclePuresyncPacket` clamp the incoming health to `[0, GetMaxHealth()]` and armor to `[0, 100]` before the damage delta is calculated, so `onPlayerDamage` and the stored values agree
- `CSimPlayerPuresyncPacket` and `CSimVehiclePuresyncPacket` clamp the same way before relaying. The max health comes from a new `CSimPlayer::m_fMaxHealth` that `UpdateSimPlayer` fills on the main thread
- `CPedSync` clamps health and armor in the received sync data, so the packet relayed to the other players carries the clamped values instead of the syncer's raw floats

Client:

- `CClientPed::SetHealth()` clamps to `[0, GetMaxHealth()]`, mirroring `CClientPed::SetArmor()`, so the value is clamped as soon as it is set
- `CNetAPI::WritePlayerPuresync` / `WriteVehiclePuresync` and `CPedSync::WritePedInformation` clamp the health and armor read from GTA memory before writing them

`setElementHealth` / `setPedArmor` already clamp on both client and server and are unchanged. No wire format change.

#### Motivation

Fixes #3791.

Since #3983 the server clamps armor on receive, so `getPedArmor` can no longer return `127.5`. Health had no equivalent: the server stored whatever the 8-bit sync carried, up to `255`, so `getElementHealth` could exceed `GetMaxHealth()` and anti-cheat scripts could not trust it. Ped sync was worse, since it sends raw floats and the server relayed them to everyone as-is. This PR closes those gaps.

Changes after review (thanks @FileEX):

- Removed the health/armor authorization tokens and the sync-increase rejection logic
- Removed the correction RPCs sent back to the syncing client
- Client sync writers read GTA memory as before, the values are only clamped before writing
- Reverted the `SPlayerHealthSync` / `SPlayerArmorSync` range change and the related test changes; `255` / `127.5` stay as chosen in `9b50a0d657`
- Armor is now clamped in the incoming puresync packets as well, and both health and armor are clamped in the client sync writers
- Health and armor are clamped when they are set on both sides, the clamping in the packets is only a safeguard
- The sim puresync packets clamp too
- The ped sync writer clamps ped health as well

Intentionally not included: armor cap 150, removing client-side `setPedArmor` / `setElementHealth`, HUD / `setPedStat` changes. Sending ped stats to players who join later is in #5434: until then, a ped whose `MAX_HEALTH` was raised with `setPedStat` is clamped to the default max on players who joined after that call.

#### Test plan

Manual:

1. `crun setElementHealth(localPlayer, 999)` → client clamps to max health; `srun getElementHealth(player)` is `<= GetMaxHealth()`, never `255`
2. A client with health/armor above the limits in memory → it syncs the clamped values, server `getElementHealth` / `getPedArmor` stay within limits, and other players receive the clamped values (sim relay on and off)
3. `crun setPedArmor(localPlayer, 9999)` → argument error; armor stays `<= 100` on both sides
4. A ped whose syncer has out-of-range health/armor → server and the other clients get the clamped values
5. Server `setElementHealth(player, 75)` / `setPedArmor(player, 75)` → both sides report ~75
6. `createPickup` health/armor pickups still work via server `Use()`
7. Damage still triggers `onPlayerDamage` / `onPedDamage` with correct deltas, on foot, in vehicle and for synced peds
8. Respawn restores max health

Automated:

- `SyncDataTypes_Tests` unchanged, the sync structures are not touched

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CPedSync.cpp
+++ b/Client/mods/deathmatch/logic/CPedSync.cpp
@@ -358,14 +358,16 @@ void CPedSync::WritePedInformation(NetBitStreamInterface* pBitStream, CClientPed
     // And health
     if (ucFlags & 0x08)
     {
-        pBitStream->Write(pPed->GetHealth());
-        pPed->m_LastSyncedData->fHealth = pPed->GetHealth();
+        const float fHealth = std::clamp(pPed->GetSyncedHealth(), 0.0f, pPed->GetMaxHealth());
+        pBitStream->Write(fHealth);
+        pPed->m_LastSyncedData->fHealth = fHealth;
     }
 
     if (ucFlags & 0x10)
     {
-        pBitStream->Write(pPed->GetArmor());
-        pPed->m_LastSyncedData->fArmour = pPed->GetArmor();
+        const float fArmor = std::clamp(pPed->GetSyncedArmor(), 0.0f, 100.0f);
+        pBitStream->Write(fArmor);
+        pPed->m_LastSyncedData->fArmour = fArmor;
     }
 
     if (flags2 & 0x01)
--- a/Server/mods/deathmatch/logic/CPed.cpp
+++ b/Server/mods/deathmatch/logic/CPed.cpp
@@ -382,6 +382,66 @@ float CPed::GetMaxHealth()
     return fMaxHealth;
 }
 
+void CPed::AuthorizeHealthChange(float fExpectedValue) noexcept
+{
+    m_healthAuth.fExpectedValue = std::clamp(fExpectedValue, 0.0f, GetMaxHealth());
+    m_healthAuth.llAuthorizeTime = GetTickCount64_();
+}
+
+void CPed::AuthorizeArmorChange(float fExpectedValue) noexcept
+{
+    m_armorAuth.fExpectedValue = std::clamp(fExpectedValue, 0.0f, 100.0f);
+    m_armorAuth.llAuthorizeTime = GetTickCount64_();
+}
+
+static bool TryConsumeHealthArmorAuth(SHealthArmorAuth& auth, float fIncomingValue, float fEpsilon) noexcept
+{
+    if (auth.fExpectedValue < 0.0f)
+        return false;
+
+    if (GetTickCount64_() - auth.llAuthorizeTime > HEALTH_ARMOR_AUTH_TIMEOUT_MS)
+    {
+        auth.fExpectedValue = -1.0f;
+        return false;
+    }
+
+    if (std::abs(fIncomingValue - auth.fExpectedValue) <= fEpsilon)
+    {
+        auth.fExpectedValue = -1.0f;
+        return true;
+    }
+
+    return false;
+}
+
+float CPed::ValidateIncomingSyncHealth(float fIncomingHealth) noexcept
+{
+    const float fOldHealth = m_fHealth;
+    fIncomingHealth = std::clamp(fIncomingHealth, 0.0f, GetMaxHealth());
+
+    if (fIncomingHealth > fOldHealth + FLOAT_EPSILON)
+    {
+        if (!TryConsumeHealthArmorAuth(m_healthAuth, fIncomingHealth, HEALTH_AUTH_EPSILON))
+            return fOldHealth;
+    }
+
+    return fIncomingHealth;
+}
+
+float CPed::ValidateIncomingSyncArmor(float fIncomingArmor) noexcept
+{
+    const float fOldArmor = m_armor;
+    fIncomingArmor = std::clamp(fIncomingArmor, 0.0f, 100.0f);
+
+    if (fIncomingArmor > fOldArmor + FLOAT_EPSILON)
+    {
+        if (!TryConsumeHealthArmorAuth(m_armorAuth, fIncomingArmor, ARMOR_AUTH_EPSILON))
+            return fOldArmor;
+    }
+
+    return fIncomingArmor;
+}
+
 const char* CPed::GetBodyPartName(unsigned char ucID)
 {
     if (ucID <= NUMELMS(BodyPartNames))
--- a/Server/mods/deathmatch/logic/CPed.cpp
+++ b/Server/mods/deathmatch/logic/CPed.cpp
@@ -394,6 +394,11 @@ void CPed::AuthorizeArmorChange(float fExpectedValue) noexcept
     m_armorAuth.llAuthorizeTime = GetTickCount64_();
 }
 
+static void InvalidateHealthArmorAuth(SHealthArmorAuth& auth) noexcept
+{
+    auth.fExpectedValue = -1.0f;
+}
+
 static bool TryConsumeHealthArmorAuth(SHealthArmorAuth& auth, float fIncomingValue, float fEpsilon) noexcept
 {
     if (auth.fExpectedValue < 0.0f)
@@ -401,13 +406,13 @@ static bool TryConsumeHealthArmorAuth(SHealthArmorAuth& auth, float fIncomingVal
 
     if (GetTickCount64_() - auth.llAuthorizeTime > HEALTH_ARMOR_AUTH_TIMEOUT_MS)
     {
-        auth.fExpectedValue = -1.0f;
+        InvalidateHealthArmorAuth(auth);
         return false;
     }
 
     if (std::abs(fIncomingValue - auth.fExpectedValue) <= fEpsilon)
     {
-        auth.fExpectedValue = -1.0f;
+        InvalidateHealthArmorAuth(auth);
         return true;
     }
 
@@ -418,6 +423,12 @@ float CPed::ValidateIncomingSyncHealth(float fIncomingHealth) noexcept
 {
     const float fOldHealth = m_fHealth;
     fIncomingHealth = std::clamp(fIncomingHealth, 0.0f, GetMaxHealth());
+
+    // A pending increase authorization must not survive meaningful damage. Use the
+    // same epsilon as auth matching so wire quantization (~0.78 per health step)
+    // does not consume the token before the authorized increase is confirmed.
+    if (fIncomingHealth < fOldHealth - HEALTH_AUTH_EPSILON)
+        InvalidateHealthArmorAuth(m_healthAuth);
 
     if (fIncomingHealth > fOldHealth + FLOAT_EPSILON)
     {
@@ -432,6 +443,9 @@ float CPed::ValidateIncomingSyncArmor(float fIncomingArmor) noexcept
 {
     const float fOldArmor = m_armor;
     fIncomingArmor = std::clamp(fIncomingArmor, 0.0f, 100.0f);
+
+    if (fIncomingArmor < fOldArmor - ARMOR_AUTH_EPSILON)
+        InvalidateHealthArmorAuth(m_armorAuth);
 
     if (fIncomingArmor > fOldArmor + FLOAT_EPSILON)
     {
--- a/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
+++ b/Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
@@ -13,6 +13,7 @@
 #include "CPlayerPuresyncPacket.h"
 #include "CElementIDs.h"
 #include "CWeaponNames.h"
+#include "CStaticFunctionDefinitions.h"
 #include "Utils.h"
 #include "CTickRateSettings.h"
 #include <net/SyncStructures.h>
@@ -185,18 +186,18 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
         SPlayerHealthSync health;
         if (!BitStream.Read(&health))
             return false;
-        float fHealth = health.data.fValue;
+        const float fIncomingHealth = health.data.fValue;
 
         // Armor
         SPlayerArmorSync armor;
         if (!BitStream.Read(&armor))
             return false;
 
-        float fArmor = armor.data.fValue;
-        float fOldArmor = pSourcePlayer->GetArmor();
-        float fArmorLoss = fOldArmor - fArmor;
-
-        pSourcePlayer->SetArmor(fArmor);
+        const float fIncomingArmor = armor.data.fValue;
+        float       fHealth = pSourcePlayer->ValidateIncomingSyncHealth(fIncomingHealth);
+        float       fArmor = pSourcePlayer->ValidateIncomingSyncArmor(fIncomingArmor);
+        const float fOldArmor = pSourcePlayer->GetArmor();
+        const float fArmorLoss = fOldArmor - fArmor;
 
         // Read out and set the camera rotation
         SCameraRotationSync camRotation;
@@ -325,6 +326,12 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
         float fOldHealth = pSourcePlayer->GetHealth();
         float fHealthLoss = fOldHealth - fHealth;
         pSourcePlayer->SetHealth(fHealth);
+        pSourcePlayer->SetArmor(fArmor);
+
+        if (fHealth != fIncomingHealth)
+            CStaticFunctionDefinitions::SendHealthCorrectionToPlayer(pSourcePlayer, pSourcePlayer, fHealth);
+        if (fArmor != fIncomingArmor)
+            CStaticFunctionDefinitions::SendArmorCorrectionToPlayer(pSourcePlayer, pSourcePlayer, fArmor);
 
         // Less than last packet's frame?
         if (fHealthLoss > 0 || fArmorLoss > 0)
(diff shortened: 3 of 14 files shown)


Dryxio:
Was this PR carefully tested for edge cases ? I can already see that the auth token can stay valid for 3 seconds and may be reused after taking damage, allowing the player to heal back
https://github.com/multitheftauto/mtasa-blue/blob/46160c32b6bb6f42f4d29c1a976435c3a2879528/Server/mods/deathmatch/logic/CPed.cpp#L422-L425

QueryOfficial:
Good catch you're right, I missed this edge case in the initial testing.

The token was only consumed on a matching increase, so taking damage in between left it valid and allowed syncing back to the authorized value within the 3s window.

Fixed in the latest commit: pending health/armor authorizations are now invalidated on a meaningful sync decrease (HEALTH_AUTH_EPSILON / ARMOR_AUTH_EPSILON), so heal-back after damage is rejected. The threshold matches the auth epsilon to avoid false invalidation from wire quantization (~0.78 health step).

Thanks for flagging this.

Client/mods/deathmatch/logic/CPedSync.cpp
```diff
@@ -358,14 +358,16 @@ void CPedSync::WritePedInformation(NetBitStreamInterface* pBitStream, CClientPed
     // And health
     if (ucFlags & 0x08)
     {
-        pBitStream->Write(pPed->GetHealth());
-        pPed->m_LastSyncedData->fHealth = pPed->GetHealth();
+        const float fHealth = std::clamp(pPed->GetSyncedHealth(), 0.0f, pPed->GetMaxHealth());
+        pBitStream->Write(fHealth);
+        pPed->m_LastSyncedData->fHealth = fHealth;
```
FileEX: Reading a custom value instead of the actual value from GTA's memory makes no sense here. If a cheater modifies GTA's memory, they can also modify this value. This change could potentially lead to another desync, since GTA itself keeps track of the actual value, just like in the case of damage and other things. So please leave it as it is.

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -325,6 +326,12 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
         float fOldHealth = pSourcePlayer->GetHealth();
         float fHealthLoss = fOldHealth - fHealth;
         pSourcePlayer->SetHealth(fHealth);
+        pSourcePlayer->SetArmor(fArmor);
+
+        if (fHealth != fIncomingHealth)
+            CStaticFunctionDefinitions::SendHealthCorrectionToPlayer(pSourcePlayer, pSourcePlayer, fHealth);
+        if (fArmor != fIncomingArmor)
+            CStaticFunctionDefinitions::SendArmorCorrectionToPlayer(pSourcePlayer, pSourcePlayer, fArmor);
```
FileEX: It's nonsense. Sending the correct value to the client would just waste bandwidth and could introduce another vulnerability that cheaters could exploit.

Server/mods/deathmatch/logic/CPed.cpp
```diff
@@ -382,6 +382,80 @@ float CPed::GetMaxHealth()
...
+        InvalidateHealthArmorAuth(m_armorAuth);
+
+    if (fIncomingArmor > fOldArmor + FLOAT_EPSILON)
+    {
+        if (!TryConsumeHealthArmorAuth(m_armorAuth, fIncomingArmor, ARMOR_AUTH_EPSILON))
+            return fOldArmor;
+    }
+
+    return fIncomingArmor;
+}
```
FileEX: Another AI mumbo jumbo....

FileEX (changes requested):
Generally, this PR is overcomplicated. The solution is simple: clamp the values in the incoming packets and clamp the values on the client side. Any special actions, such as sending the corrected value back to the offending client, are unnecessary.

Server/mods/deathmatch/logic/CPed.cpp
QueryOfficial: Changed.

Client/mods/deathmatch/logic/CPedSync.cpp
QueryOfficial: Fair point, reverted. CPedSync and CNetAPI read GetHealth()/GetArmor() straight from GTA memory again like before and I dropped the GetSyncedHealth/GetSyncedArmor getters. There are no client changes left in the PR now, setPedArmor and setElementHealth were already clamped on the client side anyway.

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
QueryOfficial: Yeah, agreed. Removed the correction RPCs and both Send*Correction functions. The server just clamps what it receives now and doesn't send anything back.

Server/mods/deathmatch/logic/CPed.cpp
QueryOfficial: Removed all of it. What's left on the server side is SetHealth clamping to GetMaxHealth() (same thing SetArmor already does since #3983) and the three sync readers clamping the incoming health before the damage delta is calculated. That's the whole change now.

QueryOfficial:
@FileEX applied the changes you asked for, the PR is down to the simple version now:

- `CPed::SetHealth` clamps to `[0, GetMaxHealth()]`, same as `SetArmor`
- `CPlayerPuresyncPacket`, `CVehiclePuresyncPacket` and `CPedSync` clamp the incoming health before the damage delta is calculated

Everything else is reverted: the auth tokens, the correction RPCs, the client sync writer changes and the SyncStructures range change (so no bitstream bump, 255/127.5 stay as they were). The client side didn't need anything since setPedArmor and setElementHealth already clamp there. This also makes @Dryxio's point moot since the token logic is gone entirely.

PR description is updated to match. Compiles fine here.

FileEX:
Now this PR doesn't do anything with armor; it only clamps health, and only on the server side. Both armor and health should be clamped on both the server and client sides. The client shouldn't be able to set 1000 HP for a ped if the server doesn't allow it.

Clamp the HP and armor values when sending the synchronization packet, as well as in functions such as `setElementHealth` and `setPedArmor` for peds.

Server/mods/deathmatch/logic/packets/CVehiclePuresyncPacket.cpp
```diff
@@ -296,7 +296,7 @@ bool CVehiclePuresyncPacket::Read(NetBitStreamInterface& BitStream)
             SPlayerHealthSync health;
             if (!BitStream.Read(&health))
                 return false;
-            float fHealth = health.data.fValue;
+            float fHealth = std::clamp(health.data.fValue, 0.0f, pSourcePlayer->GetMaxHealth());
```
FileEX, line 299: Why are you clamping the vehicle's health to the maximum health of the player sending the packet? Stop blindly using an LLM and actually pay attention to what you're doing instead of just copy-pasting whatever the LLM spits out. Use your brain...

FileEX (changes requested):

Server/mods/deathmatch/logic/packets/CVehiclePuresyncPacket.cpp
QueryOfficial: That one is the player's health, not the vehicle's. It's the `// Player health` block (`SPlayerHealthSync`) that goes into `pSourcePlayer->SetHealth()` and the `onPlayerDamage` loss, the diff hunk just starts one line below that comment. The vehicle's own health is the `SVehicleHealthSync` read in the driver block ([L146](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Server/mods/deathmatch/logic/packets/CVehiclePuresyncPacket.cpp#L145-L150)) and I didn't touch it.

Armor right below it is clamped now too.

QueryOfficial:
@FileEX updated, armor is handled on both sides now as well.

Client:
- `CNetAPI::WritePlayerPuresync` and `WriteVehiclePuresync` clamp health to `[0, GetMaxHealth()]` and armor to `[0, 100]` before writing. Still the real values from GTA memory, just clamped.
- `CPedSync::WritePedInformation` clamps armor to `[0, 100]`.

Server:
- `CPlayerPuresyncPacket` and `CVehiclePuresyncPacket` clamp the incoming armor the same way as health, before the loss for `onPlayerDamage` is calculated.
- `CPedSync` clamps health and armor in the received sync data itself. Before this the server stored its own value but relayed the syncer's raw floats to everyone else, so a ped set to 1000 HP on the syncer's side showed up as 1000 for other players too. They get the clamped value now.

`setElementHealth` and `setPedArmor` already clamp on both sides, so I left them as they are:
- client: `setElementHealth` clamps peds to `GetMaxHealth()` ([here](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Client/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L1507)), `setPedArmor` rejects anything outside 0-100 ([here](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Client/mods/deathmatch/logic/luadefs/CLuaPedDefs.cpp#L2438-L2448)) and `CClientPed::SetArmor` clamps too ([here](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Client/mods/deathmatch/logic/CClientPed.cpp#L1825))
- server: `SetElementHealth` clamps to `GetMaxHealth()` ([here](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L1785)), `SetPedArmor` caps at 100 ([here](https://github.com/multitheftauto/mtasa-blue/blob/2637719fffb52109d2264d746a8bdb5c845dd602/Server/mods/deathmatch/logic/CStaticFunctionDefinitions.cpp#L3814-L3815))

One thing I left out on purpose is ped health in the client ped sync writer. Ped stats are only broadcast at the moment `setPedStat` is called and never sent to players who join later (only player stats are, via `GetPlayerStatsPacket`). So on a late joiner a ped's `GetMaxHealth()` is the default ~100. If a ped has MAX_HEALTH raised on the server and 150 HP, clamping there would sync 100 and fire a fake `onPedDamage` for 50. The server already clamps ped health with the right max and relays the clamped value, so nothing out of range gets through. If you still want it clamped on the client, I think it should come together with sending ped stats to joining players, which feels like a separate PR.

Server/mods/deathmatch/logic/packets/CVehiclePuresyncPacket.cpp
FileEX: Sorry, you're right. Although I have no idea why we synchronize the player's HP in the `VehiclePureSync` packet.

FileEX:
I think it looks reasonable now, but you should still clamp the values on both the client and server sides in `setPedArmor` and `setElementHealth`, so they're clamped immediately when being set. The clamping in the packets should only serve as an additional safeguard.

QueryOfficial:
Done. `CClientPed::SetHealth` clamps to `[0, GetMaxHealth()]` now, the same way `CClientPed::SetArmor` and the server's `CPed::SetHealth` already do, so on both sides the value is clamped as soon as it is set and the packet clamping is only the safeguard.

The script functions themselves were already clamping (client `setElementHealth` to `GetMaxHealth()` and `setPedArmor` rejecting anything outside 0-100, server the same), so they didn't need a change.

One side effect of the client setter clamp, for the record: a ped whose MAX_HEALTH was raised with `setPedStat` gets clamped to the client's default max on players that joined after that call, because ped stats are not sent at join time. If that turns out to matter in practice, the fix is shipping ped stats with the ped entity add packet, which I'd rather do in a separate PR.

FileEX:
> One side effect of the client setter clamp, for the record: a ped whose MAX_HEALTH was raised with setPedStat gets clamped to the client's default max on players that joined after that call, because ped stats are not sent at join time. If that turns out to matter in practice, the fix is shipping ped stats with the ped entity add packet, which I'd rather do in a separate PR.

Shouldn't the server send the ped's stats to the client when joining, either in the `EntityAddPacket` or the player list packet? Otherwise, this currently causes desync. Unless they're being sent somewhere else in some weird way like VehiclePureSync lmao.

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
```diff
@@ -185,14 +185,14 @@ bool CPlayerPuresyncPacket::Read(NetBitStreamInterface& BitStream)
         SPlayerHealthSync health;
         if (!BitStream.Read(&health))
             return false;
-        float fHealth = health.data.fValue;
+        float fHealth = std::clamp(health.data.fValue, 0.0f, pSourcePlayer->GetMaxHealth());
```
FileEX, line 188: I think these changes should also be applied to `CSimPlayerPureSyncPacket` and `CSimVehiclePureSyncPacket`.

FileEX (changes requested):

Server/mods/deathmatch/logic/packets/CPlayerPuresyncPacket.cpp
QueryOfficial: Done, both sim packets clamp health and armor now. The sync thread can't call `GetMaxHealth()` on the `CPlayer`, so `CSimPlayer` got an `m_fMaxHealth` that `UpdateSimPlayer` fills on the main thread, and it's passed into the packets the same way as the weapon range.

QueryOfficial:
> Shouldn't the server send the ped's stats to the client when joining, either in the `EntityAddPacket` or the player list packet?

Yes, I checked and they're not sent anywhere else, only when `setPedStat` is called. Added them to the ped entity add packet behind a new `eBitStreamVersion::PedStatsSync`, like you did for the animations in #3520. The server writes the stats that differ from the defaults, and the client applies them right after creating the ped. I put them right after armor instead of at the end of the ped block, because the client applies the health right after creating the ped and it has to know `MAX_HEALTH` by then.

Since the client knows a ped's max health now, the ped sync writer clamps ped health too, which was the part I had left out before.

Shared/sdk/net/bitstream.h
```diff
@@ -40,6 +40,10 @@ enum class eBitStreamVersion : unsigned short
     // YYYY-MM-DD
     // Name,
 
+    // Ped stats in the entity add packet
+    // 2026-09-21
+    PedStatsSync,
+
```
FileEX: Since 1.7 hasn't been released yet, it's not necessary

FileEX (changes requested):

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
```diff
@@ -871,6 +871,25 @@ bool CEntityAddPacket::Write(NetBitStreamInterface& BitStream) const
...
+                                stats.emplace_back(usStat, fValue);
+                        }
+
+                        BitStream.WriteCompressed(static_cast<unsigned short>(stats.size()));
+                        for (const auto& [usStat, fValue] : stats)
+                        {
+                            BitStream.Write(usStat);
+                            BitStream.Write(fValue);
+                        }
+                    }
```
FileEX: I'd prefer to do this in a separate PR. This one should focus only on what it's supposed to.

Shared/sdk/net/bitstream.h
QueryOfficial: Right, removed it here. #5434 writes the stats without a version check.

Server/mods/deathmatch/logic/packets/CEntityAddPacket.cpp
QueryOfficial: Moved it to #5434.

FileEX (approved):

PR #5042, opened 2026-07-13
Fix crashes and TXD slot leaks in model cleanup

HeresHavi:
## **Summary**

Added lower-bound and overflow-safe ID checks to `engineFreeModel`, `engineFreeTXD`, and the shared model lookup.

Valid TXD cleanup now uses the model manager's removal path so affected models are reset, the allocation is removed, and the pool slot can be reused.

## **Motivation**

`CClientModelManager::FindModelByID` checked only the upper boundary. A negative model ID could therefore index memory before the model array. Large TXD IDs could also wrap into a negative internal model ID before reaching the same lookup.

For example, a resource might cache custom model and TXD IDs so it can free them during a restart. If an invalid model ID or wrapped TXD ID reaches cleanup, the old lookup could crash every player running that resource.

Valid `engineFreeTXD` calls also returned `true` without fully removing the allocation. The TXD slot remained occupied for the rest of the client session instead of becoming reusable.

<details>
<summary>engineFreeModel Crash Stack</summary>

```text
Exception: Access violation
Module: client_d.dll
Offset: 0x001CCADA

engineFreeModel(-1000000)
CClientModelManager::FindModelByID (CClientModelManager.cpp:95)
```

Crash_01a

</details>

<details>
<summary>engineFreeTXD Crash Stack</summary>

```text
Exception: Access violation
Module: client_d.dll
Offset: 0x001CCADA

engineFreeTXD(4293947296)
CClientModelManager::FindModelByID (CClientModelManager.cpp:95)
```

Crash_01b

</details>

## Test plan

**Runtime**

- Before Fix: `engineFreeModel(-1000000)` caused an access violation in `CClientModelManager::FindModelByID`.
- Before Fix: `engineFreeTXD(4293947296)` wrapped into the same invalid lookup and crashed at the same client offset.
- After Fix: Both invalid calls returned `false`, and the client stayed open.
- Valid Case: Requested model ID `332` and TXD ID `3608` were freed successfully.
- TXD Cleanup: Freeing TXD `3608` reset the model using it, and the next TXD allocation reused slot `3608`.
- Repeated Free: Freeing the already released TXD returned `false`.

**Builds and Tests**

- `Debug | Win32`: Solution build passed, 304 client tests passed.
- `Release | Win32`: Solution build passed, 304 client tests passed.
- Ran `clang-format`.

## Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -983,6 +983,12 @@ int CLuaEngineDefs::EngineFreeModel(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
+        if (iModelID < 0)
+        {
+            lua_pushboolean(luaVM, false);
+            return 1;
+        }
+
         // destroyElement() only defers the actual native entity deletion to the next pulse
         // (see CElementDeleter). If a script destroys an entity using this model and calls
         // engineFreeModel in the same tick, the entity's native CEntity is still alive and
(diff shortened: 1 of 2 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2526,8 +2532,16 @@ uint CLuaEngineDefs::EngineRequestTXD(lua_State* const luaVM, std::string strTxd
...
+    const std::uint32_t uiBaseIdForCol = g_pGame->GetBaseIDforCOL();
+
+    // Validate before adding the internal TXD offset so oversized script values cannot wrap to a negative model ID.
+    if (uiBaseIdForCol <= MAX_MODEL_DFF_ID || txdID >= uiBaseIdForCol - MAX_MODEL_DFF_ID)
+        return false;
+
+    const int                     iModelID = MAX_MODEL_DFF_ID + static_cast<int>(txdID);
+    auto                          modelManager = m_pManager->GetModelManager();
+    std::shared_ptr<CClientModel> pModel = modelManager->FindModelByID(iModelID);
+    return pModel && modelManager->Remove(pModel);
```
FileEX, line 2544: Why use ``CClientModelManager::Remove`` instead of ``CClientModel::Deallocate``? You just mentioned the ID boundary checks.

HeresHavi: The ID checks and the switch to Remove() fix two separate problems in EngineFreeTXD.

CClientModel::Deallocate() intentionally skips the actual teardown for TXD models. The old function could therefore return true while the TXD pool slot remained occupied and the model was still registered with CClientModelManager. Models referencing that TXD were not reset either.

CClientModelManager::Remove() calls the existing TXD cleanup path through RestoreEntitiesUsingThisModel() and RestoreTXD(). That resets models using the TXD, removes the pool slot, and clears the manager entry. The runtime test also confirmed that the next TXD allocation reused the freed slot. That's why I used Remove() here instead of only calling Deallocate().

FileEX (approved):

PR #5046, opened 2026-07-13
Fix client crash on oversized building pool resize

HeresHavi:
## **Summary**

Added upper-bound and allocation checks to building pool resizing.

Replacement storage is now allocated before the active GTA pool is freed. Failed resize requests return `false` without removing the world or leaving stale building pointers behind.

## **Motivation**

An oversized `engineSetPoolCapacity` request could remove the game world and free the active building pool before all replacement allocations had succeeded. If a later allocation failed, the rollback path could not safely repair existing world references.

For example, a resource might increase the building pool for a large custom map. If its capacity calculation produces an extremely large value, the old resize path could leave roads and buildings missing or crash the player's client during the next world update.

The fix rejects sizes that cannot fit GTA's signed pool capacity and handles allocation failures before changing the active pool.

<details>

```text
Exception: Access violation
Module: gta_sa.exe
Offset: 0x0000620B

HOOK_FreezeFix_Misc15 (CMultiplayerSA_CrashFixHacks.cpp:683)
HOOK_CStreaming_Update_Caller (CMultiplayerSA.cpp:2904)
CGameProcessWithTiming (CMultiplayerSA.cpp:5430)
HOOK_Idle (CMultiplayerSA.cpp:5458)
HOOK_CallIdle (CMultiplayerSA_Rendering.cpp:65)
```

Crash

</details>

## Test plan

**Runtime**

- Valid Case: Increased the building pool from `13000` to `13001`, then restored it to `13000`. Both calls returned `true`.
- Before Fix: Capacity `2147483648` raised `vector too long` after removing the world. Roads and buildings disappeared even though the capacity still reported `13000`.
- Before Fix: Capacity `200000000` left stale world pointers and caused an access violation during a later GTA update.
- After Fix: Both oversized values returned `false`, retained capacity `13000`, and the client stayed open.
- World Check: Buildings, roads, collision, driving, flying, and vehicle interaction continued working after both rejected calls.

**Builds and Tests**

- `Debug | Win32`: Build passed, 304 client tests passed.
- `Release | Win32`: Build passed, 304 client tests passed.
- `Debug | x64`: Build passed.
- `Release | x64`: Build passed.
- Ran `clang-format`.

## Checklist


Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -320,7 +327,37 @@ bool CBuildingsPoolSA::Resize(int size)
     const bool skipLinkSweeps = m_bLinkSweepsDone;
     m_bLinkSweepsDone = false;
 
-    m_buildingPool.entities.resize(size);
+    try
+    {
+        m_buildingPool.entities.resize(size);
```
FileEX, line 351: Could we avoid resizing `m_buildingPool.entities` until after both replacement allocations succeed, or otherwise restore the original vector contents on failure?

When shrinking the pool, `entities.resize(size)` destroys the entries above `size`. If either allocation then fails, resizing back to `currentSize` only appends default-initialized entries; it does not restore the original `{pEntity, pClientEntity}` mappings. The GTA pool remains intact, but the parallel client-entity metadata can be corrupted. Please ensure the metadata vector is unchanged unless the resize commits successfully.

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -320,7 +327,37 @@ bool CBuildingsPoolSA::Resize(int size)
...
+        return false;
+    }
+    catch (const std::length_error&)
+    {
+        return false;
+    }
+
+    // Allocate every replacement block first so a failure leaves the old pool and its world pointers intact.
+    CBuildingSAInterface* newObjects = MemSA::malloc_struct<CBuildingSAInterface>(size);
+    if (newObjects == nullptr)
```
FileEX:
```suggestion
    if (!newObjects)
```

Client/game_sa/CBuildingsPoolSA.cpp
```diff
@@ -320,7 +327,37 @@ bool CBuildingsPoolSA::Resize(int size)
...
+    {
+        m_buildingPool.entities.resize(currentSize);
+        return false;
+    }
+
+    tPoolObjectFlags* newBytemap = MemSA::malloc_struct<tPoolObjectFlags>(size);
+    if (newBytemap == nullptr)
```
FileEX:
```suggestion
    if (!newBytemap)
```

FileEX (changes requested):

HeresHavi:
@FileEX All is good now, did thorough tests as well, and merged latest of Master into this branch to resolve conflicts.

FileEX (approved):

PR #5047, opened 2026-07-13
Fix client crash when loading malformed IFP animation data

HeresHavi:
## **Summary**

Added ANP2 and ANP3 validation before constructing animation hierarchies.

Malformed animation counts, object counts, frame counts, frame types, truncated headers, and missing frame data now return `false`. Allocation failures during the initial animation resize are also handled safely.

## **Motivation**

`CClientIFP::ReadIFPVersion2` passed the signed `TotalAnimations` value directly to `std::vector::resize`. A value of `-1` became an extremely large unsigned size and caused an uncaught `std::length_error` to terminate the client.

For example, a resource might load a custom animation file downloaded with a character pack. If the file is corrupted or has an invalid animation count, simply calling `engineLoadIFP` could crash the player before the animation was even applied.

The validation runs once when the IFP is loaded and does not add work to animation playback.

<details>

```text
RaiseException
_CxxThrowException
std::_Xlength_error
std::vector<...>::_Resize_reallocate
std::vector<...>::resize
CClientIFP::ReadIFPVersion2 (CClientIFP.cpp:128)
CClientIFP::Load (CClientIFP.cpp:58)
CIFPEngine::LoadIFP (CIFPEngine.cpp:27)
CLuaEngineDefs::EngineLoadIFP (CLuaEngineDefs.cpp:545)
```

Crash

</details>

## Test plan

**Runtime**

- Valid Case: A 36-byte ANP2 file declaring zero animations returned a valid IFP element before and after the fix.
- Before Fix: The same file with `TotalAnimations = -1` caused an uncaught `std::length_error` and terminated the client.
- After Fix: The negative animation count returned `false`, and the client stayed open.
- Truncated Header: Declaring one animation without its header returned `false`.
- Invalid Object Count: One animation with `TotalObjects = -1` returned `false`.
- Missing Frame Data: One sequence declaring a frame without its payload returned `false`.

**Builds and Tests**

- `Release | Win32`: Full build passed, 304 client tests passed.
- Ran `clang-format`.

## Checklist


Client/mods/deathmatch/logic/CClientIFP.cpp
```diff
@@ -84,8 +84,10 @@ bool CClientIFP::ReadIFPByVersion()
...
     if (bAnp2 || bAnp3)
     {
-        ReadIFPVersion2(bAnp3);
-        return true;
+        if (!ValidateIFPVersion2(bAnp3))
+            return false;
+
+        return ReadIFPVersion2(bAnp3);
     }
     else if (bAnpk)
```
FileEX, line 92: Why is there no validation for ANPK animations? Cutscene animations are ANPK.

HeresHavi: I originally handled ANPK validation separately in #5059, but I closed that PR after I was reached out by Lopsi telling me that the issue was already resolved in 1.7 and that security sensitive parser issues should be submitted privately. That's why this PR is limited to ANP2 and ANP3. The omission wasn't because ANPK doesn’t need validation. Not quite sure what the next move is because I don't know what has been fixed privately.

FileEX:
This has been patched privately

PR #5051, opened 2026-07-13
Fix client crash when reading deeply nested Lua tables

HeresHavi:
## **Summary**

Added a 64-level nesting limit to the client and server Lua table readers.

Failed reads now stop safely, clean up partially copied arguments, and return a normal Lua error instead of continuing with invalid stack data.

## **Motivation**

`toJSON` copies Lua tables into native arguments before serializing them. A deeply nested but valid table could exceed the Lua stack limit, but the failed stack check was ignored. The next `lua_pushnil` call then accessed invalid memory and crashed the client.

For example, a client resource might build or receive a deeply nested configuration table and pass it to `toJSON`. If the table contains enough nested levels, the resource could crash the player's client instead of receiving a normal error.

<details>

```text
Exception: Access violation

lua_pushnil (vendor/lua/src/lapi.c:458)
CLuaArguments::ReadTable (Client/mods/deathmatch/logic/lua/CLuaArguments.cpp:110)
CLuaArgument::Read (Client/mods/deathmatch/logic/lua/CLuaArgument.cpp:255)
CLuaArgument::CLuaArgument (Client/mods/deathmatch/logic/lua/CLuaArgument.cpp:62)
CLuaArguments::ReadTable (Client/mods/deathmatch/logic/lua/CLuaArguments.cpp:120)
... repeated for nested table levels ...
```

Crash

</details>

## Test plan

**Runtime**

- Before Fix: Passing a 4,096-level table to `toJSON` caused an access violation in `lua_pushnil`.
- After Fix: The same table returned `nil` in 0 ms, and the client stayed open.
- Valid Case: A 16-level table returned the same 177-byte JSON string as before.
- Accepted Boundary: A table with 64 total table objects returned a 647-byte JSON string.
- Rejected Boundary: A table with 65 total table objects returned `nil`, and the client stayed open.

**Builds and Tests**

- `Debug | Win32`: Full build passed.
- Ran `clang-format`.

## Checklist


FileEX:
> A deeply nested but valid table could exceed the Lua stack limit

That's not true. 

The crash is caused by a **C++ stack overflow**. Deep recursion through `CLuaArgument::Read -> CLuaArguments::ReadTable -> CLuaArgument::CLuaArgument -> CLuaArgument::Read` eventually exceeds the stack limit, which is around 1 MB on Windows.

The same issue occurs in `CLuaArguments::WriteTableToJSONObject -> CLuaArgument::WriteToJSONObject`.

This should be fixed by getting rid of the recursion and using the heap instead of limiting the user.

```lua
local rootTable = {}
local current = rootTable

for i = 1, 8000 do
    local nextTable = { level = i }
    current.nested = nextTable
    current = nextTable
end

local jsonStr = toJSON(rootTable)
print(#jsonStr);
```

image

Waiting for your fix.

HeresHavi:
@FileEX Merged Master in to resolve conflicts. Made further improvements and caught new issues during testing. All should be good now.

FileEX:
This PR changes significantly more than it should and goes beyond its scope. Fixing the crash only requires modifying 6 files: 3 on the client side and 3 on the server side. You're modifying as many as 18 files, adding Lua stack checks and modifying the json vendor code, which is completely unrelated to the main issue, which is a C++ stack overflow.

I believe continuing the review of this PR would be a waste of time, as you're clearly not following the guidelines that you agreed to yourself:

> Smaller pull requests are easier to review

I appreciate the effort, but the way you're going about it is not acceptable.

PR #5052, opened 2026-07-13
Fix client crash when decoding malformed PNG data

HeresHavi:
## **Summary**

Added recoverable error handling to `PngDecode`.

Malformed PNG data now returns `false` instead of reaching libpng's fatal error path and terminating the client. PNG reads that extend beyond the supplied buffer are also rejected safely.

## **Motivation**

MTA builds libpng without its normal setjmp recovery support. When libpng found a fatal decoding error, its default handler called `abort()` and closed the client.

For example, a resource might download an image and pass its bytes to `dxGetPixelsSize` or `dxConvertPixels`. If the image has a damaged PNG chunk or was cut off during download, processing it could crash the player's client.

<details>

```text
Exception: Stack buffer overrun

abort
png_longjmp
png_default_error
png_error
png_chunk_error
png_crc_finish_critical
png_crc_finish
png_handle_IHDR
png_handle_chunk
png_read_info
PngDecode
PngGetDimensions
CPixelsManager::GetPixelsSize
CLuaDrawingDefs::DxGetPixelsSize
```

Crash

The dialog labels this as a stack buffer overrun, but the stack shows that the client was terminated by libpng's `abort()` path.

</details>

## Test plan

**Runtime**

- Before Fix: A PNG with a corrupted IHDR CRC terminated the client through `png_longjmp` and `abort`.
- After Fix: The same malformed PNG returned `false`, and the client stayed open.
- Valid Case: A valid 1×1 RGB PNG returned the correct dimensions and converted into plain pixels.
- Corrupt Image Data: A damaged IDAT stream returned `false` after the output buffers had been allocated, and the client stayed open.
- Short Input: An IDAT chunk declaring more data than the supplied buffer returned `false`, and the client stayed open.

**Builds and Tests**

- Ran `clang-format`.

## Checklist


Client/core/CFileFormatPng.cpp
```diff
@@ -148,67 +164,69 @@ void ReadDataFromInputStream(png_structp png_ptr, png_bytep outBytes, png_size_t
...
+            return false;
+        }
+
+        ///////////////////////////////////////////////////
+        uiOutWidth = width;
+        uiOutHeight = height;
+
+        if (pOutBuffer)
         {
             pOutBuffer->SetSize(width * height * 4);
```
FileEX: The size is calculated without overflow or maximum dimension checks. A PNG with a valid signature and very large IHDR dimensions can make this 32-bit calculation wrap to a small allocation; ParseRGB/ParseRGBA then iterate over the original width and height and write past the buffer. Since this PR specifically hardens decoding of attacker-controlled malformed input, please validate the dimensions and checked allocation size before assigning uiOutWidth/uiOutHeight or allocating the output buffer.

FileEX (changes requested):

Client/core/CFileFormatPng.cpp
HeresHavi: Improved and tested.

Client/core/premake5.lua
```diff
@@ -35,6 +35,11 @@ project "Client Core"
 	filter { "files:FastFailCrashHandler/WerCrashHandler.cpp" }
 		flags { "NoPCH" }
 
+	-- PNG error recovery throws through libpng's C API; the PCH uses /EHsc.
+	filter { "files:CFileFormatPng.cpp", "system:windows" }
+		flags { "NoPCH" }
+		buildoptions { "/EHsc-" }
+
```
FileEX, line 42: The Premake change appears to be related to allowing `PngErrorException` to propagate through libpng's extern "C" API, so it may be necessary. However, the comment currently attributes the change to the PCH, while `NoPCH` and `/EHsc-` solve different problems. Could you clarify whether `NoPCH` is required and include the generated MSVC command line or a small build/runtime test proving that `/EHsc-` is needed? Also, please consider documenting that `/EHsc-` is intended to keep `/EHs` while disabling `/EHc`, not to disable exception handling.

FileEX (changes requested):

PR #5053, opened 2026-07-13
Fix crashes caused by malformed vehicle model replacements

HeresHavi:
## **Summary**

Added checks for required model frames before applying custom vehicle replacements.

Bike and BMX replacements now require both wheel frames, while Vortex replacements require the left landing gear frame. Missing plane panel frames are also disabled safely.

Rejected replacements keep the previous working model active and are no longer retried whenever the model streams in.

## **Motivation**

GTA expects certain vehicle types to contain specific model frames. A replacement DFF can load successfully while still missing the frames needed by the vehicle it replaces, causing a crash later when GTA creates or renders it.

For example, a server might use a modified Super GT model as a replacement for a bike or Vortex. When the vehicle streams in, or when a player fires a weapon nearby, GTA can try to use wheel or landing gear frames that do not exist and crash the player's client.

<details>
<summary>Crash Stacks</summary>

```text
Exception: Access violation

CBike::SetupSuspensionLines
Offset: 0x002B8AC9
```

Bike replacement crash

```text
Exception: Access violation

CMatrix::Attach
Offset: 0x0019BD39
```

Plane replacement crash

</details>

## Test plan

**Runtime**

- Before Fix: Reproduced both reported access violations with malformed vehicle replacements.
- Train and Plane Cases: Tested all 6 train models and all 11 plane models using the modified Super GT replacement.
- Bike Cases: Tested the modified Super GT replacement with all 13 bike models.
- Vortex Case: Tested the modified Super GT replacement on Vortex while firing weapons nearby.
- Plane Panel Case: Applied 22 repeated damage events to a malformed Rustler replacement without crashing.
- Replacement State: Confirmed rejected replacements kept the previous working model active and restored it correctly.

**Builds and Tests**

- `Release | Win32`: Full build passed, 304 client tests passed.

## Checklist


HeresHavi:
For additional information, this PR fixes the reported crashes from the MTA Development Discord: https://discord.com/channels/801330706252038164/801411628600000522/1526168967218200587

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -2422,6 +2422,47 @@ static void _declspec(naked) HOOK_CrashFix_DynVBCreateNull()
...
+        push    54
+        call    CrashAverted
+        jmp     RETURN_CrashFix_CPlane_PreRenderMissingBouncingPanel_Skip
+
+    valid_frame:
+        jmp     RETURN_CrashFix_CPlane_PreRenderMissingBouncingPanel
+    }
+    // clang-format on
+}
+
```
FileEX: Since you check for the frame before replacing the model, there's no point in patching it this way, as we'd have to patch every similar case as well.

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -427,6 +429,33 @@ bool CRenderWareSA::ReplaceModel(RpClump* pNew, unsigned short usModelID, DWORD
...
+            return true;
+
+        // GTA dereferences these frames without null checks while constructing or pre-rendering the vehicle.
+        // Rejecting the replacement here preserves the original model and gives scripts a clean failure instead of a delayed native crash.
+        LogEvent(852, "Model not replaced", "CRenderWareSA::ReplaceVehicleModel",
+                 SString("Vehicle model:%u class:%u is missing required frame:%s", usModelID, static_cast<unsigned int>(vehicleClass), frameName), 5432);
+        return false;
+    };
+
+    if ((vehicleClass == VehicleClass::BIKE || vehicleClass == VehicleClass::BMX) && (!requireFrame("wheel_front") || !requireFrame("wheel_rear")))
```
FileEX, line 460: Image

GTA assumes that an automobile always has a left front wheel frame.

Client/game_sa/CRenderWareSA.cpp
```diff
@@ -427,6 +429,33 @@ bool CRenderWareSA::ReplaceModel(RpClump* pNew, unsigned short usModelID, DWORD
...
+    {
+        return false;
+    }
+
+    // Vortex initializes a bouncing panel for PLANE_GEAR_L (hierarchy id 21). CPlane::PreRender passes that frame to CMatrix without checking it.
+    if (usModelID == static_cast<unsigned short>(VehicleType::VT_VORTEX) && !requireFrame("gear_l"))
+        return false;
```
FileEX, line 467: If you're patching it with a hook, why are you returning false?

FileEX (changes requested):

Issue #5054, opened 2026-07-13
Configurable draw distance limits for dense custom race maps


IOrosa:
### Is your feature request related to a problem? Please describe.

On maps with a lot of custom-object decoration (e.g. race maps that place thousands of
`.map` objects), not all objects load in dense areas, and when the camera moves the
world flickers (objects popping in and out). This is not draw-distance related — it
happens even for nearby, small areas — it's a *density* limit.

After digging into the client, the behaviour is caused by three fixed limits working
together:

1. High-detail object streaming cap = 500.
   `CClientObjectManager` sets `m_uiMaxStreamedInCount = MAX_OBJECTS_MTA / 2` and
   `m_uiMaxLowLodStreamedInCount = MAX_OBJECTS_MTA / 2`, with `MAX_OBJECTS_MTA = 1000`
   (`Client/sdk/game/Common.h`). So at most **500 high-detail** + 500 low-LOD objects
   stream in at once. Dense areas simply stop loading objects once 500 high-detail are in.

2. Render list cap = 1000.** `CRenderer::ms_aVisibleEntityPtrs` / `ms_aVisibleLodPtrs`
   are fixed 1000-entry arrays; the `Check_NoOfVisibleEntities` / `Check_NoOfVisibleLods`
   hooks in `CMultiplayerSA_Rendering.cpp` clamp the counter at 999. When more than ~1000
   entities are visible, the excess is dropped and, because *which* ones get dropped shifts
   per frame, the world flickers.

3. Supporting pools scale with `MAX_OBJECTS_MTA` (`MAX_ENTRY_INFO_NODES_MTA`,
   `MAX_POINTER_*_LINKS_MTA`).

Related: #869 (New Limit Adjuster), #2717 (object stream limit change), #905 (rendering
performance / draw-call cost).

----------------------------------------------------------------------

What I tested

As an experiment I raised `MAX_OBJECTS_MTA`, relocated the two render arrays to larger
buffers (rewriting the 7 `[reg*4 + base]` references and raising the two `cmp 999`), and
scaled the node pools. Dense maps then loaded MUCH more decoration with a notably
higher effective view — confirming these are the binding limits.

I then instrumented FPS vs. rendered-object count with a small HUD/CSV logger while driving
a dense map. Results (single machine, DX9, capped ~52 FPS):

| Rendered objects (high + low-LOD) | FPS |
|---|---|
| 0 – 1500 | 52 (cap) |
[...]

Issue #869, opened 2019-04-15
New Limit Adjuster


ghost:
**Is your feature request related to a problem? Please describe.**

SA has many limits, and we need to lift them. When you're loading mods, it takes forever to wait. Some servers will cause clients to eat 90% of MTA process RAM and crash eventually. I played modern games, even they don't consume that much RAM. This happens because objects are loaded the moment you call engine functions, and loading is done on the main thread. We should have at least 2 threads for loading mods, and here we are loading on the main one. All of these issues are possible to fix since we have optimized memory pools #480.

**Describe the solution you'd like**
- Implement a new streamer for streaming in and out entities as the current one will cause massive lags when you have too many objects and players on screen @ArranTuna 
- Loading DFF, COL, TXD, and IFP on a separate thread when using engine functions, GTA SA loads them on a separate thread using CdStream* functions, but MTA doesn't.
- Saving RAM by loading an object only when it needs to be streamed. engineReplace* functions will load it immediately, we need to only load it if the object is visible.  #748
- Maximum draw distance—ability to stream in entire map.
- Support for multiple low LODs for a single custom object or a GTA map object and set distance for each LOD. Depending on how far the object is from the camera, the low LOD will be rendered instead of the high LOD model, this would save lots of CPU. 
- Removing ID limit—ability to add new objects without replacing current ones, this includes new vehicles with driving ability.

**Optional**

- Increasing map sectors, so mappers can create objects outside of the current -3000 to 3000 range. I think that something like -10000 to 10000 would suffice. 
- Custom collisions for every object. We can make it object based rather than model based.
- general fix of collision detection, now you can affect fps drop when you're inside bounding box of collision bigger then ~300KB

obuhhh:
If someone close this issue, then he can also close MTA too :)

CrosRoad95:
add point:
- ability to load .col and .dff files bigger then 256x256x256 ( make whole map in few files )
- general fix of collision detection, now you can affect fps drop when you're inside bounding box of collision bigger then ~300KB

support to .obj files too would be nice, no more not working plugins that export invisible collision.

Einheit-101:
My biggest concern is actually the currently poor performance, but that infinite streaming sounds good

ghost:
> * ability to load .col and .dff files bigger then 256x256x256 ( make whole map in few files )
>
> make whole map in few files

I wonder how is that going to be helpful. 
> 
[...]

PR #5055, opened 2026-07-13
Add configurable object streaming limits (engineSetObjectStreamingLimits)

IOrosa:
## What this does

Implements the configurable limits discussed in #5054.

On maps with a lot of custom-object decoration (e.g. race maps placing thousands of `.map`
objects), decoration stops loading in dense areas and the world flickers. The cause is a
hard cap of **500 high-detail streamed-in objects** (`CClientObjectManager` sets
`m_uiMaxStreamedInCount = MAX_OBJECTS_MTA / 2`, with `MAX_OBJECTS_MTA = 1000`), combined
with the fixed **1000-entry render lists** that flicker when exceeded.

This PR makes those limits **opt-in and configurable**, with defaults equal to current
behaviour (fully backward compatible):

### New client functions
- `bool engineSetObjectStreamingLimits(int highDetail, int lowLod)` — how many high-detail
  (near) and low-LOD (far) objects may stream in at once. Returns `false` if invalid or the
  total exceeds the object pool.
- `int, int engineGetObjectStreamingLimits()` — current limits.

Defaults stay at **500 / 500** (stock), so servers that don't call these are unaffected.

### Supporting engine changes
- **Render list capacity**: relocate `CRenderer::ms_aVisibleEntityPtrs` (`0xB75898`) and
  `ms_aVisibleLodPtrs` (`0xB748F8`) to larger buffers and raise the `Check_NoOfVisible*`
  clamps. This lets the extra objects actually render instead of flickering (a long-standing
  ask, #869). It is inert with the default limits — a scene rarely exceeds 1000 visible
  entities — so it does not change stock behaviour.
- **Object pool ceiling**: `MAX_OBJECTS_MTA` raised to act as the maximum the runtime limits
  may reach; node pools scale with it.

## Why (draw-call trade-off)

Raising these trades FPS for density, so it must be the server/mapper's informed choice —
that's why it's opt-in rather than a global bump. I profiled FPS vs. rendered-object count
while driving a dense map (single machine, DX9, ~52 FPS frame limiter):

| Rendered objects (high + low-LOD) | FPS |
|---|---|
| 0 – 1500 | 52 |
| 1750 – 2000 | 46 |
| 2000 – 2250 | 43 |
| 2500 – 2750 | 33 |
| 3000+ | 30 |

Linear fit: **~9.4 FPS lost per 1000 rendered objects**, flat at 52 up to ~1800. Low-LOD
objects cost the same as high-detail ones despite far fewer polys, i.e. this is draw-call
bound (CPU/DX9), consistent with #905. On my content, `engineSetObjectStreamingLimits(1500, 800)`
gave a **flat ~52 FPS with ~2x the object density** of stock.

## Backward compatibility

- Defaults are identical to current behaviour (500 / 500).
- The render-array relocation and larger pools are inert unless the limits are raised.
- No new ACL rights; no server config changes.

## Testing

Built and tested locally on a dense custom race map: `/start`-ing a resource that calls
`engineSetObjectStreamingLimits(1500, 800)` visibly loads far more decoration; changing the
values at runtime updates streaming immediately; stopping it restores stock (500/500). No
crashes; FPS matches the table above.

## Open questions for reviewers

1. **Pool memory**: the object pool ceiling (`MAX_OBJECTS_MTA`) is compile-time, so raising
   it costs some memory for everyone even at default limits. Happy to tune the ceiling, make
   it configurable, or explore growing the pool on demand — guidance welcome.
2. **API surface**: free `engine*` functions here; would you also like OOP methods on an
   Engine class, and/or a getter for current streamed-in counts to help scripts auto-tune?
3. Wiki docs to follow once the API shape is agreed.

### DIFF
--- a/Client/mods/deathmatch/logic/CClientObjectManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientObjectManager.cpp
@@ -36,8 +36,10 @@ CClientObjectManager::CClientObjectManager(CClientManager* pManager)
     // Initialize members
     m_pManager = pManager;
     m_bCanRemoveFromList = true;
-    m_uiMaxStreamedInCount = MAX_OBJECTS_MTA / 2;
-    m_uiMaxLowLodStreamedInCount = MAX_OBJECTS_MTA - m_uiMaxStreamedInCount;
+    // Default streaming caps = stock behaviour (500/500). Configurable at runtime via
+    // engineSetObjectStreamingLimits (raising trades FPS for density; draw-call bound).
+    m_uiMaxStreamedInCount = 500;
+    m_uiMaxLowLodStreamedInCount = 500;
 }
 
 CClientObjectManager::~CClientObjectManager()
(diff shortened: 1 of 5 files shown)


IOrosa:
Rebased and solved issue with clang format

PR #5059, opened 2026-07-14
Fix client crash when loading malformed ANPK IFP data

HeresHavi:
## **Summary**

Added bounds and structure checks before parsing legacy `ANPK` IFP data.
Malformed chunk sizes, counts, names, and frame data are now rejected instead of being copied into fixed buffers or read past the input.

## **Motivation**

ANPK chunk sizes came directly from the file and were copied into fixed-size parser structures. An oversized INFO chunk could overwrite stack memory and terminate the client with a stack buffer overrun.

For example, a resource might include a custom animation pack and load it through `engineLoadIFP`. If that file is damaged or deliberately malformed, starting the resource could crash the player's client instead of returning a normal load failure.

<details>
<summary>Crash Information</summary>

```text
Exception: 0xC0000409 (Stack Buffer Overrun)
Module: client_d.dll
Offset: 0x0008B7D5
IDA Address: 0x1098B7D5
```

Crash

The crash was reproduced by loading an ANPK file with a declared and supplied 64-byte INFO payload. The client froze, closed, and displayed an integrity violation. The next launch reported the stack buffer overrun above.

</details>

## Test plan

**Runtime**

- Before Fix: Loading an ANPK with a 64-byte INFO payload through `engineLoadIFP` froze and terminated the Debug client.
- After Fix: Loading the same 80-byte ANPK returned `false`, and the client remained open.
- Valid Case: A normal empty ANPK and a nonempty ANPK with one named animation both loaded successfully.
- Invalid Cases: Truncated INFO and oversized NAME, DGAN INFO, and ANIM payloads returned `false` without crashing.

**Builds and Tests**

- `Debug | Win32`: Build passed, 304 client tests passed.
- `Release | Win32`: Build passed, 304 client tests passed.
- `Debug | x64`: Server build passed.
- Ran `clang-format`.

## Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientIFP.cpp
+++ b/Client/mods/deathmatch/logic/CClientIFP.cpp
@@ -90,18 +90,164 @@ bool CClientIFP::ReadIFPByVersion()
     else if (bAnpk)
     {
         m_bVersion1 = true;
-        ReadIFPVersion1();
-        return true;
+        if (!ValidateIFPVersion1())
+            return false;
+
+        return ReadIFPVersion1();
     }
     return false;
 }
 
-void CClientIFP::ReadIFPVersion1()
+bool CClientIFP::ValidateIFPVersion1() const
 {
-    SInfo Info;
+    const SString& buffer = GetBuffer();
+    size_t         offset = GetReadOffset();
+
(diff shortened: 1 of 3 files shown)


HeresHavi:
Resolved.

PR #5069, opened 2026-07-22
Improve voice packet handling

Lpsd:


MoriskaSosiska:
Wouldn't it be better to create `pInputBuffer`, `audioBuffer`, and `uiTotalBufferSize` after `if (CClientTime::GetTime() - m_ulTimeOfLastSend > 100 && m_VoiceState != VOICESTATE_AWAITING_INPUT)`?

I was wondering about that when I looked through the code previously

PR #5071, opened 2026-07-22
Use the vehicle position for explosion and fire damage in onClientVehicleDamage

### LINKED ISSUE #5068: `onClientVehicleDamage` event triggers twice for explosions
### Describe the bug

When a vehicle is damaged by a single explosion, `onClientVehicleDamage` is triggered more than once and the damage weapon/type is always `51` regardless of the actual weapon that created the explosion (RPGs, Grenades, Projectiles...)

The damage world position is somehow set at `(0.0, 0.0, 0,0)` for some event trigger, then it is set to the actual world position in the other triggers.

Loss parameter is also affected, in one of those both triggers it is set to `0`

### Steps to reproduce

Spawn a vehicle and shoot once at the vehicle with an RPG or a Grenade

### Version

Client:
Multi Theft Auto v1.6-release-24124
Copyright (C) 2003 - 2026 Multi Theft Auto

Server:
MTA:SA Server v1.6-release-23312

### Additional context

Image

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary
Report the vehicle position in `onClientVehicleDamage` when San Andreas supplies no damage position.

GTA zeroes the damage position vector before calling `CVehicle::InflictDamage` for explosion and fire damage, so scripts received `(0, 0, 0)` for those hits. When the reported position is a zero vector, the vehicle position is passed to the event instead.

#### Motivation
Follow-up to #5068.

That report is not a bug in the parts it claims: `onClientVehicleDamage` firing twice for a rocket/grenade hit is the result of two real explosions (the projectile detonating, then the vehicle blowing up), not a duplicate dispatch of a single damage event, and weapon `51` is what the game genuinely reports for explosion damage. Thanks to @FileEX for digging into this and pointing it out.

The one part worth patching is the position: `(0, 0, 0)` is intentional on the game's side but useless to scripts, since it cannot be distinguished from a real world position and gives no hint of where the vehicle was damaged. This PR only changes that, and leaves the event count, weapon type, attacker and loss untouched.

The previous approach in this PR (suppressing the second event and remapping weapon type/attacker from the last explosion) has been dropped, so the diff is now a single hunk in `CClientGame::VehicleDamageHandler`.

#### Test plan
1. Attach a debug handler to `onClientVehicleDamage` printing attacker, weapon, loss and x/y/z.
2. Hit a vehicle with an RPG or grenade: the reported position is now the vehicle position instead of `(0, 0, 0)`. Event count, weapon (`51`) and loss are unchanged from master.
3. Set a vehicle on fire (e.g. `blowVehicle` on a nearby vehicle or a molotov): fire damage also reports the vehicle position instead of `(0, 0, 0)`.
4. Ram the vehicle into another vehicle/object: collision damage still reports the real collision position, unchanged from master.
5. Shoot the vehicle with a rifle and shoot out a tyre: bullet and tyre-burst damage still report the bullet impact position, unchanged from master.
6. Cancel the event in the handler and confirm the vehicle takes no damage, unchanged from master.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientExplosionManager.cpp
+++ b/Client/mods/deathmatch/logic/CClientExplosionManager.cpp
@@ -23,6 +23,7 @@ CClientExplosionManager::CClientExplosionManager(CClientManager* pManager)
     m_pManager = pManager;
     m_LastWeaponType = WEAPONTYPE_UNARMED;
     m_pLastCreator = NULL;
+    m_vecLastPosition = CVector();
 }
 
 CClientExplosionManager::~CClientExplosionManager()
@@ -80,6 +81,9 @@ bool CClientExplosionManager::Hook_ExplosionCreation(CEntity* pGameExplodingEnti
             return false;
 
         eWeaponType explosionWeaponType = GetWeaponTypeFromExplosionType(explosionType);
+        m_LastWeaponType = explosionWeaponType;
+        m_vecLastPosition = vecPosition;
+        m_pLastCreator = localPlayer;
 
         CLuaArguments arguments;
         arguments.PushNumber(vecPosition.fX);
@@ -142,6 +146,13 @@ bool CClientExplosionManager::Hook_ExplosionCreation(CEntity* pGameExplodingEnti
             vehicle->SetBlowState(VehicleBlowState::BLOWN);
         }
 
+        if (allowExplosion)
+        {
+            m_LastWeaponType = explosionWeaponType;
+            m_vecLastPosition = vecPosition;
+            m_pLastCreator = pResponsible;
+        }
+
         return allowExplosion;
     }
 
@@ -196,8 +207,8 @@ CExplosion* CClientExplosionManager::Create(eExplosionType explosionType, CVecto
     if (pCreator)
         pGameCreator = pCreator->GetGameEntity();
 
-    // Update our records first?
     m_pLastCreator = pCreator;
+    m_vecLastPosition = vecPosition;
     if (responsibleWeapon != WEAPONTYPE_UNARMED)
         m_LastWeaponType = responsibleWeapon;
     else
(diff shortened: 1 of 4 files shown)


FileEX:
I investigated this issue and came to different conclusions than you did.

The event isn't triggered by double damage as you described, with one callback coming from `InflictDamage` and the other from `VehicleDamage`. I also wasn't able to reproduce a case where loss = 0.

`onClientVehicleDamage` is triggered twice because there are actually two explosions. One occurs when the projectile explodes, and the other occurs when the vehicle itself explodes. This is further supported by the fact that when an explosion occurs next to the vehicle, the event is triggered only once.

Several explosions near the vehicle
image

Explosion on the vehicle
image

> The x times counter simply counts all events, so don't take the fact that the event was triggered 2,466 times as an indication of anything here.

As for the position, this is intentional. In the GTA code, the position is set to (0, 0, 0) when the damage is caused by an explosion or fire. This isn't particularly intuitive, though, so it would probably make more sense to use the vehicle's position instead.

image

Ultimately, I don't think this is a bug, since two explosions are naturally occurring here. The only thing that could be patched, in my opinion, is replacing the (0, 0, 0) position with the vehicle's actual position.

QueryOfficial:
I've made the changes based on your feedback. I've also updated the PR title and description.

FileEX (approved):

gh0st2k2:
When an explosion happens, onClientVehicleDamage triggers twice if the vehicle is low hp or is going to explode, the first one is actual explosion caused by the world, and the second one is to kill the vehicle and make it explode I think. I dont think the event should trigger for the vehicle own explosion as there is an event for that onClientVehicleExplode. This happens also for vehicles that are just sitting there burning then exploding triggering this damage event, which does not make much sense because the vehicle was not really damaged. This makes it harder to filter out which explosions are really damaging the vehicle from those of self destruction explosions.

The weapon type is also always set to 51, which generic type for explosions. The damage could be originated from different sources like grenades, projectiles etc... Idk if this is patched or not

PR #5072, opened 2026-07-22
Fix settings restart prompt lost after language/skin change

### LINKED ISSUE #5063: Settings: restart prompt is lost when changing resolution then switching to Language/Interface tab
**Describe the bug**
When a user changes the screen resolution in the Graphics/Video settings tab and then switches to the Language/Interface tab to change the language (or any other setting), the client does not prompt to restart - even though changing resolution requires a restart. The restart prompt that should appear for the resolution change is lost/overwritten when switching tabs.

**To reproduce**
1. Open MTA:SA settings (ESC ? Settings)
2. Go to the Graphics/Video tab
3. Change the screen resolution to a different value
4. Without closing settings, switch to the Language/Interface tab
5. Change the language (or any other setting)
6. Close settings
7. Notice: the client does NOT prompt to restart, even though the resolution change requires one

**Expected behaviour**
If the resolution was changed, the client should always prompt to restart regardless of whether the user also changed language/interface settings. The restart requirement from the resolution change should persist across tab switches.

**Additional context**
This appears to be a bug in CSettings where the "needs restart" flag for resolution changes is cleared or not properly tracked when navigating between settings tabs.

QueryOfficial:
#### Summary
Persist the settings "restart required" prompt across Interface tab locale/skin rebuilds so changing resolution (or other restart-required options) still prompts after switching language/skin.

Previously `ShowRestartQuestion()` showed a `QuestionBox` owned by `MainMenu`. Changing language or skin destroys and recreates `MainMenu`, which dropped that prompt. The restart requirement is now tracked on `CLocalGUI` and re-shown after the UI rebuild.

#### Motivation
Fixes #5063 
When a user changed screen resolution in Video settings, then switched to the Interface tab and changed language (or skin), closing settings no longer asked to restart — even though the resolution change still required one.

Root cause: locale/skin changes call `DestroyWindows()` / `CreateWindows()`, which replaces `MainMenu` and its `QuestionBox`, so the restart dialog was lost and never restored.

#### Test plan
1. Open Settings → Video → change resolution only → OK → restart prompt appears.
2. Open Settings → Video → change resolution → Interface → change language → OK → after language reload, **RESTART REQUIRED** appears in the new language.
3. Repeat step 2 with a skin change instead of language → restart prompt appears after the UI rebuild.
4. Change language only (no restart-required setting) → no restart prompt.
5. On the restart prompt, choose **No** → client continues; choose **Yes** → client restarts.
6. Confirm dialog strings match the existing locale translations.

#### Checklist

### DIFF
--- a/Client/core/CGUI.cpp
+++ b/Client/core/CGUI.cpp
@@ -45,6 +45,7 @@ CLocalGUI::CLocalGUI()
     m_LastSettingsRevision = -1;
     m_LocaleChangeCounter = 0;
     m_bHasQueuedLocaleChange = false;
+    m_bPendingRestartPrompt = false;
 }
 
 CLocalGUI::~CLocalGUI()
@@ -104,6 +105,8 @@ void CLocalGUI::SetSkin(const char* szName)
         CreateWindows(guiWasLoaded);
         m_pConsole->SetPosition(consolePos);
         m_pConsole->SetSize(consoleSize);
+        // QuestionBox was destroyed with MainMenu; re-show if settings still need a restart
+        TryShowRestartPrompt();
     }
 
     if (CCore::GetSingleton().GetConsole() && !error.empty())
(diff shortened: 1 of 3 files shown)


DateFTP:
Thank you for fixing the bugfix, as I started encountering this bug with the latest 1.6 build. It's a good thing I added the issues to the bugfix wait list.

QueryOfficial:
> Thank you for fixing the bugfix, as I started encountering this bug with the latest 1.6 build. It's a good thing I added the issues to the bugfix wait list.

I follow the issues created in my free time and try to address them as much as I can.

FileEX (approved):

PR #5074, opened 2026-07-22
Fix vehicle headlights not lighting other vehicles

### LINKED ISSUE #4636: Vehicles textures are not reacting to light sources (headlights)
### Describe the bug

Vehicles textures are buggy and not correctly reacting to light sources as they should, like headlights from other vehicles. 

https://streamable.com/rsnyyb -> Vanilla GTA SA

https://streamable.com/4rw1dc -> MTA SA

https://streamable.com/yduogq -> MTA SA

MTASA test: light flicker happens when I press Accelerate+Reverse command controls (eg. W+S)

### Steps to reproduce

While at night or somewhere you can see shading of the vehicle change
1. Drive a Vehicle
2. Press Accelerate and Reverse at the same time
3. Have another vehicle's headlights aimed at yours
A light flicker will occour whenever you release the W or S keys. 

### Version

Client version: Multi Theft Auto v1.6-release-23690
Server version: Multi Theft Auto v1.6-release-23690

### Additional context

GTA SA vanilla only with limit adjuster and Dyom8.3 installed
MTA SA only running clump_object + freeroam, play and the basic standard resources from MTA itself. Double checked and nothing else is running that could changed the lights behaviour from the main game. 

### Relevant log output

```shell

```

### Security Policy

Dryxio:
#### Summary
Move the vehicle specular light from direct3d slot 1 to slot 7.

Slot 1 is also used for the first temporary directional light, so the vehicle pipeline was overwriting the incoming headlights before render the vehicle. Slot 7 is also added in common `LIGHT*` shader parameters to not break custom shaders.

#### Motivation
Fixes #4636.

#### Test plan
I reproduced the same scenario as in the issue video. The video below shows the result after the fix.

https://github.com/user-attachments/assets/01866dac-b1d3-43ee-b65c-5122e9c7f15b

#### Checklist

### DIFF
--- a/Client/core/Graphics/CRenderItem.EffectParameters.cpp
+++ b/Client/core/Graphics/CRenderItem.EffectParameters.cpp
@@ -579,13 +579,17 @@ bool CEffectParameters::ApplyCommonHandles()
         D3DXCOLOR   strongestDiffuse(0, 0, 0, 0);
         D3DXCOLOR   strongestSpecular(0, 0, 0, 0);
         D3DXVECTOR3 strongestDirection(0, 0, -1);
-        for (uint i = 0; i < 4; i++)
+        // The vehicle pipeline reserves slot 7 for its specular directional
+        // light. Include it without broadening the existing selection of GTA's
+        // temporary world lights in slots 0 through 3.
+        constexpr DWORD COMMON_EFFECT_LIGHT_SLOTS[] = {0, 1, 2, 3, 7};
+        for (const DWORD lightSlot : COMMON_EFFECT_LIGHT_SLOTS)
         {
             BOOL bEnabled;
-            if (SUCCEEDED(pDevice->GetLightEnable(i, &bEnabled)) && bEnabled)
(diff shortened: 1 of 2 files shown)


FileEX:
Could you explain this bug? Why doesn't it occur in vanilla SA? Why does the light only disappear/reappear when pressing W or S while driving?

Slot 1 is intentionally used
image

Dryxio:
> Could you explain this bug? Why doesn't it occur in vanilla SA? Why does the light only disappear/reappear when pressing W or S while driving?
> 
> Slot 1 is intentionally used image

GTA use slot 1 for specular, but RenderWare also gives slot 1 to first extra light, so headlight gets overwritten.
I tested again and it also happen on vanilla (the issue was misleading, he tested maybe with SilentPatch that fixes it), W+S adds brake light in next slot, so lighting comes back until one key is released.

[SilentPatch fix](https://github.com/CookiePLMonster/SilentPatch/commit/2a84d8bf7a0b34795ff49c0b2ec5d50b8a6a70f3)

FileEX (approved):

PR #5078, opened 2026-07-23
Fix getPedAnimation for short-lived partial anims

### LINKED ISSUE #5077: ``getPedAnimation`` can return false in partial anims
### Describe the bug

Executing partial anims with a short but > 0 anim time leads to this bug in which the animation is clearly playing clientside but getPedAnimation thinks its not.

Image

### Steps to reproduce

1. Execute this animation:
```lua
setPedAnimation(thePlayer, "carry", "crry_prtial", 1, true, true, true, true, 0, false)
```
2. Use ``getPedAnimation(thePlayer)``

Now if you do
```lua
setPedAnimation(jugador, "carry", "crry_prtial", -1, true, true, true, true, 0, false)
```
getPedAnimation will retrieve the correct value for getPedAnimation

### Version

Client: v1.7-untested-26679

### Additional context

Tasks also don't think you are doing any animation whatsoever

Image

### Relevant log output

```shell

```

### Security Policy

QueryOfficial:
#### Summary
Make `getPedAnimation` keep returning the cached block/anim when a short-lived partial animation has already ended its `TASK_SIMPLE_NAMED_ANIM` primary task but `loop` / `freezeLastFrame` still holds the pose.

Also set `m_AnimationCache.startTime` in `RunNamedAnimation` for locally applied animations (RPCs already set it).

```lua
setPedAnimation(localPlayer, "carry", "crry_prtial", 1, true, true, true, true, 0, false)
print(getPedAnimation(localPlayer)) -- expected: "carry", "crry_prtial", ...
```

#### Motivation
With a short but non-zero `time` (e.g. `1`), partial anims such as `carry` / `crry_prtial` finish the named-anim task almost immediately. The ped stays in the carry pose via loop/freeze, but `IsRunningAnimation` only checked `TASK_SIMPLE_NAMED_ANIM` for streamed-in peds, so `getPedAnimation` returned `false`. Using `time = -1` kept the task alive and appeared to work.

Fixes #5077.

#### Test plan
1. Build the client.
2. Run:
   `setPedAnimation(localPlayer, "carry", "crry_prtial", 1, true, true, true, true, 0, false)`
   then `getPedAnimation(localPlayer)` — expect `"carry"` / `"crry_prtial"` (not `false`).
3. Run the same with `time = -1` and confirm it still returns the correct values.
4. Clear with `setPedAnimation(localPlayer, false)` and confirm `getPedAnimation` returns `false`.

#### Checklist
```

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -5897,8 +5897,6 @@ void CClientPed::RunNamedAnimation(std::unique_ptr<CAnimBlock>& pBlock, const ch
     m_AnimationCache.bUpdatePosition = bUpdatePosition;
     m_AnimationCache.bInterruptible = bInterruptible;
     m_AnimationCache.bFreezeLastFrame = bFreezeLastFrame;
-    // Local setPedAnimation did not set this (RPCs did); needed for IsAnimationInProgress.
-    m_AnimationCache.startTime = GetTimestamp();
 }
 
 void CClientPed::KillAnimation()
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -5784,7 +5784,6 @@ bool CClientPed::GetRunningAnimationName(SString& strBlockName, SString& strAnim
 
 bool CClientPed::IsRunningAnimation()
 {
-    // Prefer the live named-anim primary task when present.
     if (m_pPlayerPed)
     {
         CTask* pTask = m_pTaskManager->GetTask(TASK_PRIORITY_PRIMARY);
@@ -5792,10 +5791,8 @@ bool CClientPed::IsRunningAnimation()
             return true;
     }
 
-    // Why: partial anims with a short time (e.g. time=1 on crry_prtial) finish the
-    // TASK_SIMPLE_NAMED_ANIM almost immediately while loop/freezeLastFrame keeps the
-    // pose. getPedAnimation must still report the cached block/anim in that case —
-    // same rule already used for streamed-out peds.
+    // Short-lived partial anims end TASK_SIMPLE_NAMED_ANIM while loop/freezeLastFrame
+    // keeps the pose; fall back to cache (same as streamed-out peds).
     return (m_AnimationCache.bLoop || m_AnimationCache.bFreezeLastFrame) && m_pAnimationBlock;
 }
 
@@ -5900,8 +5897,7 @@ void CClientPed::RunNamedAnimation(std::unique_ptr<CAnimBlock>& pBlock, const ch
     m_AnimationCache.bUpdatePosition = bUpdatePosition;
     m_AnimationCache.bInterruptible = bInterruptible;
     m_AnimationCache.bFreezeLastFrame = bFreezeLastFrame;
-    // Local setPedAnimation previously left startTime unset (only RPCs set it), which
-    // broke time-based IsAnimationInProgress checks for client-applied anims.
+    // Local setPedAnimation did not set this (RPCs did); needed for IsAnimationInProgress.
     m_AnimationCache.startTime = GetTimestamp();
 }
 
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -5784,15 +5784,18 @@ bool CClientPed::GetRunningAnimationName(SString& strBlockName, SString& strAnim
 
 bool CClientPed::IsRunningAnimation()
 {
+    // Prefer the live named-anim primary task when present.
     if (m_pPlayerPed)
     {
         CTask* pTask = m_pTaskManager->GetTask(TASK_PRIORITY_PRIMARY);
         if (pTask && pTask->GetTaskType() == TASK_SIMPLE_NAMED_ANIM)
-        {
             return true;
-        }
-        return false;
     }
+
+    // Why: partial anims with a short time (e.g. time=1 on crry_prtial) finish the
+    // TASK_SIMPLE_NAMED_ANIM almost immediately while loop/freezeLastFrame keeps the
+    // pose. getPedAnimation must still report the cached block/anim in that case —
+    // same rule already used for streamed-out peds.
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Client/mods/deathmatch/logic/CClientPed.cpp [hunk 1]
```diff
@@ -5951,6 +5897,8 @@ void CClientPed::RunNamedAnimation(std::unique_ptr<CAnimBlock>& pBlock, const ch
     m_AnimationCache.bUpdatePosition = bUpdatePosition;
     m_AnimationCache.bInterruptible = bInterruptible;
     m_AnimationCache.bFreezeLastFrame = bFreezeLastFrame;
+    // Local setPedAnimation did not set this (RPCs did); needed for IsAnimationInProgress.
+    m_AnimationCache.startTime = GetTimestamp();
```
FileEX: `getPedAnimation` doesn't call `IsAnimationInProgress`, so how do these changes fix the issue?

QueryOfficial: Thanks for the review.

This PR does two related things:

1) Fix for #5077 — getPedAnimation
getPedAnimation does not call IsAnimationInProgress. It goes through:
getPedAnimation → GetPedAnimation → GetRunningAnimationName → IsRunningAnimation

Previously, for a streamed-in ped we returned false as soon as TASK_SIMPLE_NAMED_ANIM was gone. With a short time (e.g. 1), partial anims like carry/crry_prtial end that task almost immediately, while loop/freezeLastFrame still hold the pose — so getPedAnimation returned false even though the anim was still applied. With time = -1 the task stays alive, which is why that case appeared to work.

The IsRunningAnimation change removes that early return false and falls through to the same cache check already used for streamed-out peds: (bLoop || bFreezeLastFrame) && m_pAnimationBlock.

KillAnimation clears m_pAnimationBlock, so setPedAnimation(ped, false) still makes getPedAnimation return false.

2) Separate consistency fix — startTime in RunNamedAnimation
Local setPedAnimation previously left m_AnimationCache.startTime unset, while the RPC path already sets it. We now set it in RunNamedAnimation as well, so IsAnimationInProgress / stream-in progress restoration work correctly for locally applied anims. This is not what getPedAnimation uses; it is a small related fix shipped alongside #5077.

Dryxio: `RunAnimationFromCache `also calls `RunNamedAnimation`, so doesnt this reset startTime during stream-in and restart the animation progress from 0?

QueryOfficial: Good catch, you're right.

RunAnimationFromCache calls RunNamedAnimation on stream-in, and UpdateAnimationProgressAndSpeed restores progress from the original startTime. Setting startTime inside RunNamedAnimation would reset it and restart the anim from 0 for anyone streaming back in.

I removed startTime from RunNamedAnimation. Local SetPedAnimation now stamps startTime after a successful RunNamedAnimation, same as the RPC path, so stream-in keeps the original timestamp.

I also aligned the local path with RPC by resetting speed to 1.0 and progress to 0 when a new anim starts, so a previous anim's speed doesn't leak into stream-in restore for the new one.

The #5077 fix is still only the IsRunningAnimation change.

FileEX: I'm not sure if you know what you're doing or if you're just blindly making changes without even testing them. The current changes make no sense in the context of the reported bug and don't fix it.

image

Dryxio: Feels like talking to his LLM directly, I doubt it was tested

QueryOfficial: image

I re-tested this locally on a client built from this branch.

Repro:
```lua
setPedAnimation(localPlayer, "carry", "crry_prtial", 1, true, true, true, true, 0, false)
getPedAnimation(localPlayer)
```

I tested this on a build from this branch (screenshot above) — `getPedAnimation` returns `carry` / `crry_prtial`, not `false`.

Could you have been testing on a different branch/build by mistake?

FileEX: > Could you have been testing on a different branch/build by mistake

I don't think so
image

image

The changes in this PR make no sense at all. All you've added is setting the animation cache's speed to 1 and progress to 0, but that cache is only used to restore animations after stream-in.

`GetRunningAnimationName` uses `IsRunningAnimation`, which only checks the `TASK_SIMPLE_NAMED_ANIM` task when the ped is streamed in. The changes in this PR don't affect that code path in any way, so they don't fix the reported issue.

Did you perhaps forget to push some of your changes? As it stands, this PR is invalid.
image

QueryOfficial: You're looking at a different change — agreed, those SetPedAnimation cache lines are not the getPedAnimation / #5077 fix.

While working on this I also noticed local setPedAnimation left startTime/speed/progress inconsistent with the RPC path (matters for IsAnimationInProgress / stream-in restore), so I fixed that on the same PR. That's what you're seeing in CStaticFunctionDefinitions.cpp:

- dc2bcf41bf — move startTime out of RunNamedAnimation; set startTime/speed/progress in SetPedAnimation

The actual #5077 fix is IsRunningAnimation in CClientPed.cpp (remove early return false, fall through to cache). Commits:

- b49722d3f0 — Fix getPedAnimation for short-lived partial anims
- 6360de25ad — Trim verbose comments in getPedAnimation fix

Please test those (or the full branch HEAD), not only the StaticFunctionDefinitions hunk. Same repro with the IsRunningAnimation change returns "carry", "crry_prtial" instead of false.

FileEX: #4821 changes the startTime behavior, so please remove these lines to keep this PR focused only on fixing the mentioned issue. Removing the early return is indeed the correct change.

QueryOfficial: Thanks for the review. I’ve removed the startTime / related animation cache changes from this PR so it stays focused on the getPedAnimation / IsRunningAnimation early-return fix. I’ll address the other issue in a separate PR.

FileEX (approved):

PR #5081, opened 2026-07-23
Fix for an issue where voice rejected packets

MoriskaSosiska:
Fix for an issue where `CVoiceDataPacket::Read` rejected packets and caused audio dropouts due to `while (uiSpeexFramesAvailable-- > 0)`.

~~Increased the `MAX_VOICE_PACKETS_PER_INTERVAL` limit from `5` to `32`.~~

Moved the `VOICE_PACKET_INTERVAL_MS` and `MAX_VOICE_PACKETS_PER_INTERVAL` limits to the server configuration as `voice_packets_interval` and `max_voice_packets_per_interval`.
Also increased the default value of `max_voice_packets_per_interval` from `5` to `32`.

Hopefully, there won't be any more issues.

@Lpsd

### DIFF
--- a/Server/mods/deathmatch/logic/CMainConfig.cpp
+++ b/Server/mods/deathmatch/logic/CMainConfig.cpp
@@ -1539,6 +1539,8 @@ const std::vector<SIntSetting>& CMainConfig::GetIntSettingList()
          &CMainConfig::OnPlayerTriggeredEventIntervalChange},
         {true, true, 0, 1, 1, "resource_client_file_checks", &m_checkResourceClientFiles, nullptr},
         {true, true, 0, 1, 2, "allow_multi_command_handlers", &m_allowMultiCommandHandlers, &CMainConfig::OnAllowMultiCommandHandlersChange},
+        {true, true, 50, 100, 1000, "voice_packets_interval", &m_voicePacketsInterval, nullptr},
+        {true, true, 1, 32, 200, "max_voice_packets_per_interval", &m_maxVoicePacketsPerInterval, nullptr},
     };
 
     static std::vector<SIntSetting> settingsList;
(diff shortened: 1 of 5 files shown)


Lpsd:
I will make these configurable by the server instead

MoriskaSosiska:
> I will make these configurable by the server instead

Doesn't this seem a bit excessive?

MoriskaSosiska:
> I will make these configurable by the server instead

If you consider this approach reliable, I can implement it in my PR

Lpsd:
It's not really excessive to give server owners control. We do something similar already with `player_triggered_event_interval` and `max_player_triggered_events_per_interval` for events, why not have the same for voice packets?

Lpsd:
> If you consider this approach reliable, I can implement it in my PR

Go ahead

MoriskaSosiska:
> Go ahead

Should `VOICE_PACKET_INTERVAL_MS` be controlled by the server?
I'm just not entirely sure

Lpsd:
> Should `VOICE_PACKET_INTERVAL_MS` be controlled by the server?
> I'm just not entirely sure

The interval and threshold, yes. i.e `VOICE_PACKET_INTERVAL_MS` and `MAX_VOICE_PACKETS_PER_INTERVAL`

Lpsd (approved):
I have also added `max_voice_buffer_size` as the 2048 default is excessive, new default 512 with min 128 and max 2048.

Thanks for your PR!

MoriskaSosiska:
> I have also added `max_voice_buffer_size` as the 2048 default is excessive, new default 512 with min 128 and max 2048.
> 
> Thanks for your PR!

Are you sure `512` is enough? I think `2048` would be a better default, since the goal of this PR is to ensure everything works reliably out of the box. People will most likely set this parameter without giving it much thought, which could lead to issues.

Lpsd:
2048 is an incredibly high value, even 512 is very generous. Would be safe with 128 even at wideband and highest bit rate, let alone narrowband. This is a single speex frame part

PR #5084, opened 2026-07-24
Fix missing DFT-30 middle wheel

Dryxio:
#### Summary
Fix the missing left middle wheel on DFT-30.

The model calls this component `wheel_lm`, but the game search for `wheel_lm_dummy`, so it is never found. The fix accepts `wheel_lm` when this component is searched.

Based on [SilentPatch hierarchy typo fix](https://github.com/CookiePLMonster/SilentPatch/blob/16731db9de6bd84d3ac9e84f6c6fefe3c2cb1f31/SilentPatchSA/SilentPatchSA.cpp#L1964-L2001).

PS: I planned to make one PR with all SilentPatch vehicle fixes but not sure if you prefer one fix per PR. This one only fixes DFT.

Screen before the fix : 
mta-screen_2026-07-25_00-28-50

Screen after the fix :
mta-screen_2026-07-25_00-53-01

#### Motivation
DFT-30 has one missing middle wheel in GTA SA and MTA.

#### Test plan
Spawn DFT-30, the left middle wheel is now visible.

#### Checklist

* [x] Smaller pull requests are easier to review.

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
@@ -10,6 +10,22 @@
 
 #include "StdInc.h"
 
+#define CALL_CClumpModelInfo_FindFrameFromName_Stricmp 0x4C5311
+
+static int __cdecl CompareVehicleComponentName(const char* expectedName, const char* frameName)
+{
+    const int result = _stricmp(expectedName, frameName);
+    if (result == 0)
+        return 0;
+
+    // The stock DFT-30 model names its left middle wheel "wheel_lm", while
+    // the vehicle hierarchy expects "wheel_lm_dummy".
+    if (_stricmp(expectedName, "wheel_lm_dummy") == 0 && _stricmp(frameName, "wheel_lm") == 0)
+        return 0;
+
+    return result;
+}
+
 static bool __fastcall AreVehicleDoorsUndamageable(CVehicleSAInterface* vehicle)
 {
     SClientEntity<CVehicleSA>* pair = pGameInterface->GetPools()->GetVehicle((DWORD*)vehicle);
@@ -72,4 +88,5 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
 void CMultiplayerSA::InitHooks_Vehicles()
 {
     EZHookInstall(CDamageManager__ProgressDoorDamage);
+    HookInstallCall(CALL_CClumpModelInfo_FindFrameFromName_Stricmp, (DWORD)CompareVehicleComponentName);
 }


SpeedyFolf:
Should be optional, some people might prefer to keep things vanilla, such as on our speedrunner server. I think you might as well add all of them at once. I already did #4918 but it's AI so it's probably bad.

FileEX:
Personally, I think fixes like this should be provided as an MTA resource. Similar to how the [mapfixes ](https://github.com/multitheftauto/mtasa-resources/pull/504) resource was created. The resource would simply replace these models with versions that have the corrected names.

Dryxio:
> Personally, I think fixes like this should be provided as an MTA resource. Similar to how the [mapfixes ](https://github.com/multitheftauto/mtasa-resources/pull/504) resource was created. The resource would simply replace these models with versions that have the corrected names.

fair, will work on that
what about the hunter bug? where the cockpit interior disappear when you place the camera on the right side

(https://github.com/CookiePLMonster/SilentPatch/blob/63ece88a80d666e10d70156c46a2998afd5bf3a6/SilentPatchSA/SilentPatchSA.cpp#L517)

SpeedyFolf:
> fair, will work on that

Surely already exists, I have one https://gitlab.com/SpeedyFolf/mta-resources/-/tree/main/vehiclemodelfixes?ref_type=heads and Mr. Green Gaming has one https://github.com/MrGreenGaming/MTA-Resources/tree/master/resources/%5Brace%5D/race_fix, but I figured using code instead providing models would be better, resource/folder names can vary from server to server meaning you have to download them again. Maybe adding it to `setWorldSpecialPropertyEnabled` would be better than what I did, like QueryOfficial wants to add ps2scorchedvehicles to that.

FileEX:
> > Personally, I think fixes like this should be provided as an MTA resource. Similar to how the [mapfixes ](https://github.com/multitheftauto/mtasa-resources/pull/504) resource was created. The resource would simply replace these models with versions that have the corrected names.
> 
> fair, will work on that what about the hunter bug? where the cockpit interior disappear when you place the camera on the right side
> 
> (https://github.com/CookiePLMonster/SilentPatch/blob/63ece88a80d666e10d70156c46a2998afd5bf3a6/SilentPatchSA/SilentPatchSA.cpp#L517)

Since this is a model loading issue, fixing it in the MTA code makes sense in this case

SpeedyFolf:
I just remembered in 1.7 the headlights and taillights work on the Rhino and that wasn't made optional, it's a thing by default, and they don't work in singleplayer even with SilentPatch. That may be added to a future version of SilentPatch but it probably won't be a setting, the vehicle model hierarchy typo fixes isn't a setting.

FileEX:
Closed in favor of #4918

PR #5095, opened 2026-07-26
Fix jump and climb ignoring setElementCollidableWith

### LINKED ISSUE #1758: Jump and climb controls ignore setElementCollidableWith
**Describe the bug**
Jump and climb controls/animations ignore setElementCollidableWith function (client-side).

**To reproduce**
1. Create an object
2. Make the object non-collidable with localPlayer
3. Try to jump while inside the object or climb on it

**Expected behaviour**
localPlayer shouldn't collide with the object

**Version**
1.5.8

**Video**
[https://www.youtube.com/watch?v=aNkUXZmpihk](https://www.youtube.com/watch?v=aNkUXZmpihk)

TheCrazy17:
#### Summary

Jump and climb detection now respects setElementCollidableWith. CTaskSimpleClimb's scan did its own collision check and never went through the game's existing collision handler; this hooks it into the same one setElementCollidableWith already uses.

#### Motivation

Fixes #1758. An object made non collidable with a ped via setElementCollidableWith could still be jumped on or climbed, since the climb scan bypassed the collision handler entirely.

#### Test plan

Tested in game. Created an object, disabled collision with the local player via setElementCollidableWith, then tried to jump on and climb it; the player now falls through instead of grabbing a ledge. Also confirmed normal climbing on regular objects still works.

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -168,6 +168,12 @@ DWORD**       VAR_TagInfoArray = (DWORD**)0xA9A8C0;
 #define HOOKPOS_CPhysical_ProcessCollisionSectorList 0x54BB93
 DWORD RETURN_CPhysical_ProcessCollisionSectorList = 0x54BB9A;
 
+// CTaskSimpleClimb::ScanToGrabSectorList. Hooked right after its own collision check, so climb
+// and vault scans also skip entities excluded via setElementCollidableWith.
+#define HOOKPOS_CTaskSimpleClimb_ScanToGrabSectorList 0x67DF28
+DWORD RETURN_CTaskSimpleClimb_ScanToGrabSectorList = 0x67DF36;
+DWORD SKIP_CTaskSimpleClimb_ScanToGrabSectorList = 0x67E580;
+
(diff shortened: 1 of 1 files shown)


Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -6398,6 +6406,62 @@ static void __declspec(naked) HOOK_CPhysical_ProcessCollisionSectorList()
...
+    // clang-format off
+    __asm
+    {
+        mov     pClimbScanPedInterface, ebx
+        mov     pClimbScanTargetInterface, esi
+        test    byte ptr [esi+1Ch], 1
+        mov     word ptr [esi+2Ch], cx
+        setne   al
+        mov     bClimbScanTargetUsesCollision, al
+        pushad
```
FileEX: You sure this is needed? It looks like EAX, ECX and EDX can be safely overwritten, while the remaining registers like ESI or EBX will be restored by the epilogue.

FileEX:
Was this tested after the latest commit?

TheCrazy17:
> Was this tested after the latest commit?

Of course! I tested it, and it's still working perfectly. You were right about the correction. Thanks a lot.

FileEX (approved):

PR #5097, opened 2026-07-26
Limit PDB discovery to client binary directories

### LINKED ISSUE #5096: Regression: Large client resource data (~10 GB) loads extremely slowly and triggers a false "GTA: San Andreas may not have launched correctly" dialog
### Describe the bug

When launching MTA through a custom launcher, if the client resource data becomes large (approximately **10 GB**), MTA takes an **extremely long time to start**.

After clicking the launcher, MTA begins loading the client resource data from the `mods\deathmatch\resources` folder. During this loading process, the startup time is significantly longer compared to older builds such as **r23690**.

While MTA is still loading the resource data, it incorrectly displays the following dialog:

> **"GTA: San Andreas may not have launched correctly. Terminate it?"**

This appears to be a false detection because GTA: San Andreas has not crashed. MTA is still processing and loading the large amount of client resource data.

This issue did not exist in **r23690**. Using the exact same GTA: San Andreas installation, the same client resource files, the same launcher, and the same system, MTA starts normally without any delay or warning on **r23690**. The problem only occurs on the latest **1.6** and **1.7** builds.

Additionally, if I temporarily remove or move the **`mods\deathmatch\resources`** folder before clicking the launcher, MTA starts immediately and the warning dialog never appears. As soon as the folder is restored, the issue returns.

### Tested versions

- ✅ **r23690** – Works correctly.
- ❌ **Latest MTA 1.6 (24127)** – Issue occurs.
- ❌ **Latest MTA 1.7 (26733)** – Issue also occurs.

### Expected behavior

After clicking the launcher, MTA should continue loading the client resource data normally, regardless of how long the loading process takes. It should not incorrectly detect GTA: San Andreas as failed to launch while MTA is still loading resources.

### Actual behavior

Recent MTA builds take much longer to start when launched through the launcher with a large client resource folder.

During this extended startup process, MTA incorrectly assumes that GTA: San Andreas failed to launch and displays:

> **"GTA: San Andreas may not have launched correctly. Terminate it?"**

Despite the warning, MTA eventually finishes loading and launches successfully after a very long delay.

### Additional information

- The issue happens immediately after clicking the launcher.
- Client resource data size is approximately **10 GB**.
- The issue has been reproduced on multiple computers.
- Windows was reinstalled multiple times for testing.
- The GTA: San Andreas installation has been verified and is not the cause of the issue.
- The same launcher, GTA installation, and resource files were used in all tests.
- Removing or moving the **`mods\deathmatch\resources`** folder before launching MTA makes the problem disappear immediately.
- Downgrading to **r23690** immediately resolves the issue without changing anything except the MTA version.

### Steps to reproduce

1. Prepare a client installation with approximately **10 GB** of resource data inside: mods\deathmatch\resources
2. Launch MTA using a launcher.
3. Wait while MTA loads the client resource data.
4. Notice that startup takes significantly longer on recent MTA builds.
5. Before loading completes, MTA displays:
> **"GTA: San Andreas may not have launched correctly. Terminate it?"**
6. Wait longer.
7. MTA eventually finishes loading and starts successfully.

HeresHavi:
## **Summary**

Limited crash handler PDB discovery to the two directories where client binaries and their symbols are actually placed:

- The MTA installation root
- The `MTA` binary directory

Previously, PDB discovery recursively scanned the entire installation. This included `mods/deathmatch/resources` and every cached client resource, even though these directories do not normally contain client PDBs.

The new discovery helper checks only the known binary directories without entering nested folders. Regression tests verify that PDBs beside client binaries are detected while PDBs in resources and other nested directories are ignored.

## **Motivation**

Crash handler initialization runs synchronously during client startup. `CrashHandler::GetPdbDirectories` started a recursive filesystem traversal from `SharedUtil::GetMTAProcessBaseDir`, which covers the complete MTA installation.

This means a large client resource cache adds filesystem work before the frontend appears. While this scan is running, the launcher can interpret the lack of visible startup progress as GTA being stuck and show the incorrect “GTA: San Andreas may not have launched correctly” warning described in #5096.

The amount of work scales mainly with the number of filesystem entries rather than their total size. Antivirus scanning, slower storage, or an uncached filesystem can make the same traversal considerably more expensive.

Client PDBs produced by the normal build are placed beside the launcher or directly inside `MTA`. Restricting discovery to those locations preserves local symbol support without scanning unrelated content.

This means PDBs placed in arbitrary nested directories are no longer discovered. That is intentional: nested resource and content directories are not part of the client binary layout and should not affect startup time.

Fixes #5096.

<details><summary>Profiling results</summary>

Created a resource test folder containing:

- 400 directories
- 400,000 sparse files
- 13,107,200,000 logical bytes (12.207 GiB)

Filesystem traversal results with the same dataset:

```
Before Fix: entries=405638 pdbs=0 elapsed_ms=316
After Fix:  entries=74     pdbs=0 elapsed_ms=0
```

The fixed traversal completed below the benchmark timer's one millisecond resolution.

This machine has a fast NVMe drive, so the original traversal did not reach the launcher watchdog timeout and the incorrect-launch dialog itself was not reproduced. The test did confirm that the old implementation entered the complete resource tree and that the fixed implementation does not.

Client launch validation:

- Before Fix: The original production client with the stress folder reached `L3 closed` after 14,060 ms.
- After Fix: A coherent development build with the same stress folder reached `CreateDevice success` after 6,277 ms and did not show the incorrect-launch warning.

The launch checks used different builds and milestones, so they are startup validation rather than a direct percentage comparison.

</details>

## Test plan

**Runtime**

- Before Fix: Started the production client with the 12.207 GiB, 400,000-file resource folder present. PDB discovery visited the complete installation tree, including the resource folder.
- After Fix: Started a development client with the same resource folder. It reached `CreateDevice success` after 6,277 ms, and PDB discovery inspected only 74 entries in the two binary directories.
- Local symbols: Added PDBs beside the launcher and directly inside `MTA`. Both directories were detected.
- Nested symbols: Added PDBs inside `mods/deathmatch/resources` and a nested directory under `MTA`. Neither directory was returned by PDB discovery.
- Missing symbols: Verified that discovery returns an empty result when the two binary directories contain no PDBs.

**Builds and Tests**

- `Debug | Win32`: Full build passed, 306 client tests passed (304 existing tests and 2 new regression tests).
- Ran `clang-format`.

## Checklist


Dryxio:
Is Bin/mods/deathmatch intentionally excluded? Client Deathmatch also outputs its PDB there

HeresHavi:
> Is Bin/mods/deathmatch intentionally excluded? Client Deathmatch also outputs its PDB there

Not intentional. I just missed adding it back. The original issue was that it also descended into "mods/deathmatch/resources" which we don't want.

FileEX (approved):

PR #5101, opened 2026-07-29
Fix custom partial animations leaving peds rigid

### LINKED ISSUE #2733: Cannot replace weapon animations

TheCrazy17:
#### Summary

Replacing a partial animation, like any of the weapon fire ones, with an animation loaded from a custom IFP left the ped rigid. It could still be walked around, but the legs and torso stopped animating and it stayed locked facing one direction.

Loading an IFP pads every animation out to all 32 bones and fills the ones the animation does not define with a fixed pose. A full body animation needs that padding, but a partial animation is only meant to drive part of the skeleton and leave the rest to whatever movement animation is playing, so the padding ended up driving the root, pelvis and legs as well.

The replacement is now trimmed back to the bones the original animation drives, read from its static association, so the padding can no longer reach bones the original never touched. This only runs for partial animations, so full body replacements keep taking exactly the same path as before.

**Before:** https://www.youtube.com/watch?v=gqyxviWVcJE

**After:**

https://github.com/user-attachments/assets/d2706649-f979-4060-bfdd-2791a5a95947

#### Motivation

Fixes #2733 and #3055.

Custom weapon animations have been unusable for a long time.

#### Test plan

[partialanimtest.zip](https://github.com/user-attachments/files/30487694/partialanimtest.zip)

Tested with a small resource that loads a custom shotgun.ifp and binds two keys, one calling `engineReplaceAnimation` on `"shotgun"` / `"shotgun_fire"` and the other calling `engineRestoreAnimation` to go back to the stock one.

Before the change, once the animation was replaced, aiming and firing the shotgun locked the ped facing a single direction with the legs and torso frozen, so it slid around without any walking animation.

After the change, the custom fire animation plays on the upper body while the ped keeps walking, turning and animating normally, and restoring brings the stock animation back.

#### Checklist

### DIFF
--- a/Client/game_sa/CAnimBlendAssociationSA.h
+++ b/Client/game_sa/CAnimBlendAssociationSA.h
@@ -164,6 +164,8 @@ public:
 
     float GetBlendAmount() { return m_pInterface->fBlendAmount; }
     void  SetBlendAmount(float fAmount) { m_pInterface->fBlendAmount = fAmount; }
+    bool  IsPartial() const noexcept { return m_pInterface->m_bPartial; }
+    void  RestrictToBonesOf(const CAnimBlendStaticAssociationSAInterface* pOriginalAssoc);
     void  SetCurrentProgress(float fProgress);
     float GetCurrentProgress() const noexcept { return m_pInterface->fCurrentTime; }
     float GetCurrentSpeed() const noexcept { return m_pInterface->fSpeed; }
--- a/Client/game_sa/CAnimBlendAssociationSA.h
+++ b/Client/game_sa/CAnimBlendAssociationSA.h
@@ -164,8 +164,8 @@ public:
 
     float GetBlendAmount() { return m_pInterface->fBlendAmount; }
     void  SetBlendAmount(float fAmount) { m_pInterface->fBlendAmount = fAmount; }
-    bool  IsPartial() const noexcept { return m_pInterface->m_bPartial; }
-    void  RestrictToBonesOf(const CAnimBlendStaticAssociationSAInterface* pOriginalAssoc);
+    bool  IsPartial() const override { return m_pInterface->m_bPartial; }
+    void  RestrictToBonesOf(const CAnimBlendStaticAssociationSAInterface* pOriginalAssoc) override;
     void  SetCurrentProgress(float fProgress);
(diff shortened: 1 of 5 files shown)


RughCuttle:
Tested this and it worked. I tested with all the custom animations he provided in the test script. I've even test with fighta_block and no problems.

Client/game_sa/CAnimBlendAssociationSA.h
```diff
@@ -164,6 +164,8 @@ class CAnimBlendAssociationSA : public CAnimBlendAssociation
 
     float GetBlendAmount() { return m_pInterface->fBlendAmount; }
     void  SetBlendAmount(float fAmount) { m_pInterface->fBlendAmount = fAmount; }
+    bool  IsPartial() const noexcept { return m_pInterface->m_bPartial; }
+    void  RestrictToBonesOf(const CAnimBlendStaticAssociationSAInterface* pOriginalAssoc);
```
FileEX: override

FileEX (changes requested):

Client/game_sa/CAnimBlendAssociationSA.h
FileEX: We shouldn't mark functions that read fields from SA interfaces as noexcept (IsPartial), because if `m_pInterface` somehow gets overwritten or corrupted, it will trigger `std::terminate` instead of a normal crash, making the issue harder to diagnose.

TheCrazy17: I think it's resolved now 😄

FileEX:
I think it would be good to have this in the upcoming 1.7 release. It looks good to me, and I trust that RughCuttle has tested it.

FileEX (approved):

PlatinMTA:
goated PR, thank you so much dude

PR #5103, opened 2026-07-31
Add SA-MP fast weapon strafe glitch

### LINKED ISSUE #2228: Implement new glitch for setGlitchEnabled that allows SA-MP-like strafing with an aimed weapon
The players movement while aiming is pretty different to SA-MP.
For me as an ex-SA-MP-Player this feels some kind of uncomfortable.
I know that MTA is closer to the singleplayer game than SA-MP is, but i also think that this would be a nice feature for those who want to fight like in SA-MP but want to enjoy the benefits of MTA.

It would be great if you could enable this kind of strafing like a glitch with for example setGlitchEnabled("samp_strafing", true)

None.

**Additional context**
For a better explanation what i mean, you can watch this video and you will see that in SA-MP the change between the different GUNMOVE-Animations is faster than in MTA, left is kinda MTA-like and the old SA-MP Movement, right is the current Movement in SA-MP and the Movement i try to achieve:
https://www.youtube.com/watch?v=W2MkSXFR-RU

Dryxio:
#### Summary

This PR add `fastweaponstrafe` to the existing glitch functions:

```lua
setGlitchEnabled("fastweaponstrafe", true)
isGlitchEnabled("fastweaponstrafe")
```

When enabled, changing direction while aiming react faster, similar to SA-MP. 
It is disabled by default and the normal MTA movement stay unchanged when it is off.

#### Demo

https://github.com/user-attachments/assets/c366565c-fa49-48f1-b76b-1fccbf135dd8

#### Motivation

This is not really a GTA or MTA bug, because MTA reproduce the original GTA SA behaviour here. But the direction changes while aiming feels very slow and not really enjoyable for gunplay. For players coming from SA-MP, this is also one of the complaint I heard the most about MTA gunplay.

Following the discussion in #2228, we tried for a long time to reproduce the SA-MP feeling only in Lua, by changing animation speed, blend time or replacing the gunmove animations. Some results were close, but never worked properly and not the same feeling at all.

The closest result we got was [this Lua attempt](https://www.twitch.tv/lakata__/clip/BitterGleamingDadHeyGirl-Q_0oy4f_qG_7d0mt),
but it was still not the same behaviour and was not clean enough to use.

The reason is that these Lua solutions only modify the animations, but cannot change the internal movement timestep used during the transition. This is why the behaviour cannot be reproduced cleanly only with a resource.

Closes #2228

#### Implementation

SA-MP use a separate timestep for `CTaskSimpleUseGun::ControlGunMove`, calculated from the frame rate sampled over short intervals. This PR reproduce this behaviour while the glitch is enabled and keep the current MTA code path when it is disabled.

The option is added to the existing client and server glitch lists. Its state is also synchronized in MapInfo with a versioned bit, so it is correctly applied when joining or reconnecting.

#### Test plan

Test resource: 
[faststrafe.zip](https://github.com/user-attachments/files/30567694/faststrafe.zip)

Server:

```text
refresh
start fastweaponstrafe-toggle
```

Client:

```text
/faststrafe
```

Aim with a weapon and alternate between left and right movement. The first command enable the faster direction changes and the second command restore the normal MTA behaviour.

#### Checklist


Shared/sdk/net/SyncStructures.h
```diff
@@ -1916,11 +1916,22 @@ struct SFunBugsStateSync : public ISyncStructure
     {
         BITCOUNT = 10
     };
+    // Keep post-release additions versioned and separate so older clients
+    // continue reading the following MapInfo fields at the same bit offset.
+    enum
+    {
+        BITCOUNT2 = 1
+    };
```
FileEX: Since the new version hasn't been officially released yet, I think using BITCOUNT here would be fine.

Dryxio: Done 👍

guibzo:
cool. SA combat system really seems pretty stiff.

pube1:
It looks great—this is something I've wanted for a while.

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -214,6 +289,12 @@ static void __declspec(naked) HOOK_CTimer__Update()
     // clang-format off
     __asm
     {
+        // Keep the private timestep current while disabled so enabling the
+        // glitch does not begin with a stale sample.
+        pushad
+        call UpdateFastWeaponStrafeTimeStep
+        popad
```
FileEX: pushad and the popad are likely not needed here.

Dryxio: done ty

Issue #574, opened 2018-09-24
Train carriages: setElementPosition desync


ArranTuna:
**Describe the bug**
Setting a trains serverside position will only work as long as the train is synced in. As soon as it's not synced in anymore the train will simply not move away from the last set position in the players sync radius.

Sometimes, if you set even more positions for example over a timer, you can see the train switching from frame to frame to the new position it should be and back to the position it originally spawned with its carriages. I guess the carriage position does not get updated which causes those bugs.

This only happens when you have carriages attached, if they are not attached it works as expected.

Jusonex has already reproduced it. I am just posting this bugreport to have it somewhere noted down.
**To reproduce**

**Expected behaviour**

**Screenshots**

**Version**

**Additional context**

From https://bugs.mtasa.com/view.php?id=8742

Issue #396, opened 2018-09-04
vehicle.trainPosition is only synced when it has an occupant


ArranTuna:
**Describe the bug**

@qaisjp said:

> This only affects the server.
>
> There isn't any desync with the actual vehicle.position. This also has the added affect of the trainPosition being 0 (default position) until it has been entered.
> 
> Entering the train will update the train position server side. If you leave the vehicle the trainPosition will no longer be updated

**To reproduce**
> ```
>run v = Vehicle(449, getRandomPlayer().position)
> run v.id = "test"
> run v.trainPosition -- will return 0
> crun Element.getByID("test").trainPosition -- will return non-zero value
> ```

**Additional context**
From https://bugs.mtasa.com/view.php?id=9305

Cazomino05:
most likely just some missing sync for trainPosition. I can't imagine this is difficult to fix if the vehicle is in range of someone to sync it, when I left off train sync I did consider that this is VERY easily calculated since it's just time/distance calculations and there is really no reason why a very simple "emulator" couldn't be implemented here in the server.

ModafinilCPP:
any solution for this?

Issue #747, opened 2018-12-30
Fix infinite loop if a track runs across/crosses another


Cazomino05:
**Call stack**

![image](https://user-images.githubusercontent.com/923242/50547626-d524ae00-0c3d-11e9-89f7-eb9fe8687ea9.png)

Cazomino05:
This is caused by calls to something along the line of CTrain__findclosesttracknode in the game (I don't have my IDB's anymore it's a function in the game train class) basically what it does is goes looking for the closest track node but it never takes into account
- track id
- overlapping tracks

anything within a radius (I believe from memory 0.1f) would trigger this issue.  This is so large because track position is stored as a short somewhere and divided down to a float so we lose precision and therefore it doesn't matter where it is. 

my suggestion and what I was doing when I left was modifying this function to only scan the current track (we know it, why waste time?) and implement our own find closest track function taking care to be "safe" about loops. 

The callstack above will probably contain the CTrain update method (the one called every frame), that will call the function that calculates position on the track which is likely where it stuck in a loop.

Basically the loop goes around to see if it found a new closest track node it'l keep going around but if it finds two then it gets stuck as it can't decide between the two (again due to loss of precision of the position x,y and z).

Edit: the function was either find closest track or find position on track

ghost:
> (I don't have my IDB's anymore

You can download latest GTA SA IDB from here:
[https://mirror.mtasa.com/dev/gta_sa1.0us-IDBs.zip](https://mirror.mtasa.com/dev/gta_sa1.0us-IDBs.zip)

ghost:
I talked to @qaisjp about this. Here's the reversed code for `CTrain::ProcessControl`: 
https://github.com/saml1er/gta-reversed/blob/master/source/game_sa/Entity/Vehicle/CTrain.cpp#L226

The reversed code is compilable . If we hook the original function, it should work. It's better to port the reversed code to MTA. We can create a new file for it (CTrainSA.h/.cpp) and replace it with `CTrain::ProcessControl` hook in MTA code: 
https://github.com/multitheftauto/mtasa-blue/blob/master/Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp#L1568

FileEX:
Fixed in https://github.com/multitheftauto/mtasa-blue/commit/adfbd692ed311ac13e986b078e7e445c76957708

PR #5114, opened 2026-07-31
Add support for custom train tracks

### LINKED ISSUE #388: Ability to create custom train tracks
This has an associated project, https://github.com/multitheftauto/mtasa-blue/projects/2

**Additional context**
From https://bugs.mtasa.com/view.php?id=4149

TheCrazy17:
#### Summary

Adds custom train track support to MTA:SA. Scripts can build a track from any list of points and put a train on it, instead of being limited to the game's 4 built-in tracks.

New functions:
- `createTrainTrack(nodePositions)`: creates a track element from a list of positions.
- `setTrainTrack(vehicle, trainTrack)` / `getTrainTrack(vehicle)`: now also accept custom track elements, not just the 4 default tracks.
- `getDefaultTrack(trackID)`: returns the element for one of the game's 4 built-in tracks.

A custom track runs through the game's own train code unmodified, same physics, audio, station and derail logic, just fed a different set of track data. Track IDs are a byte (0xFF reserved for "no track"), so up to 255 tracks can exist at once.

[Video](https://www.youtube.com/watch?v=4WHu3DEOnZk)

#### Motivation

GTA:SA only ships 4 built-in train tracks. Scripters wanting custom rail-based movement (roller coasters, custom subway systems, etc.) had no way to do this with real train physics, animation and audio, only workarounds like attaching vehicles to moving objects, which don't behave like an actual train.

This picks up where #250 (by qaisjp) left off; that PR aimed to do the same thing but was closed without being finished. Resolves #388.

#### Test plan

Tested manually with a local test resource. Verified: creating a custom track and spawning a train on it; a closed loop with several nodes and real curves, not just straight segments; a track with height changes along its length, to check the train handles slopes correctly; switching a train between a custom track and one of the game's default tracks; destroying a custom track while a train is on it, which derails the train; and a player joining or reconnecting mid-session still receiving existing custom tracks.

[traintracktest.zip](https://github.com/user-attachments/files/30596159/traintracktest.zip)

#### Checklist


FileEX:
What happens if the tracks intersect or cross each other? I remember there used to be an issue where GTA would end up in an infinite loop internally. Also, is it possible to switch from one track to another while the train is moving, like a railway switch?

TheCrazy17:
> What happens if the tracks intersect or cross each other? I remember there used to be an issue where GTA would end up in an infinite loop internally. Also, is it possible to switch from one track to another while the train is moving, like a railway switch?

Crossing tracks is fine, in both senses: two separate tracks crossing don't interact at all, each one is its own independent node list. And a single track crossing itself doesn't confuse the game either, since it walks the nodes in the order you gave them, not by where they sit in space.

The infinite loop you're remembering is #747; went and checked it against the actual game code. Turns out it's already fixed, since 2019 (adfbd692e), just never got closed. The native wraparound in `CTrain::ProcessControl` used to hang on a zero-length track, but that commit replaced it with a safe `fmodf`-based wrap that also resets to 0 on a degenerate track, and it covers our custom tracks too since they share the same length lookup as the default ones. Verified it myself against the exe with Ghidra, same address as the callstack in that issue. That issue should be marked as closed.

Switching tracks while moving already works, that's meant to let scripters build their own junctions out of multiple tracks.

Keeping this PR small on purpose, so it stays easy to review and merge. I originally wanted to throw in more functions, fix train sync, node editing, and a few other things, but that was turning into way too many changes for one PR. Kept it to just this for now.

Dryxio:
`CSimVehiclePuresyncPacket::Read/Write` still use the old uchar layout, shouldn't they be updated to the new train track format too?

TheCrazy17:
> `CSimVehiclePuresyncPacket::Read/Write` still use the old uchar layout, shouldn't they be updated to the new train track format too?

Done. 😀

Client/game_sa/CTrainTrackManagerSA.cpp
```diff
@@ -0,0 +1,192 @@
+/*****************************************************************************
+ *
+ *  PROJECT:     Multi Theft Auto
+ *  LICENSE:     See LICENSE in the top level directory
+ *
+ *  Multi Theft Auto is available from https://www.multitheftauto.com/
+ *
+ *****************************************************************************/
```
FileEX, line 10: Please use the standard header comment format. In every new file.
```
 *
 *  FILE:        game_sa/CTrainTrackManagerSA.cpp
 *  PURPOSE:     Something
 *
 *  Multi Theft Auto is available from https://www.multitheftauto.com/
 *
```

guibzo:
Good one!

FileEX:
Since you've made sync changes, does your PR also fix #574?

TheCrazy17:
> Since you've made sync changes, does your PR also fix #574?

No, it doesn't fix #574, but it partially fixes #396 (there's still work to be done and I didn't want to make the PR bigger).

RushCuttle:
I've tested it: 

- derail logic worked flawlessly 
- created a track with 1000 nodes (no issues)
- tried to see if SA tracks (by moving the train at the same pos/height) interfere, no they don't
- used all of the test script to test and created other functions to test
- Does allow crossed tracks (no bugs on nodes that cross each other)
- no issues or crashes
- at last node train just changes it's direction to keep following the tracks instead of stopping (checked to see if it was on the test script but it seems to be a native SA thing)

didn't check any sync issue early discussed here.

Issue #4497, opened 2025-09-29
Bullet sync / onPlayerWeaponFire is still broken


ArranTuna:
### Describe the bug

I upgraded the server to test if bullet sync was fixed, but it turns out it's still broken, just not as badly broken. Players report sometimes nobody is taking damage.

Sometimes when firing a weapon onPlayerWeaponFire is not triggering. Here is a video that shows it working sometimes, then sometimes, not, half way in I find a good spot to reproduce it:

https://streamable.com/6pax4o

### Steps to reproduce

srun addEventHandler("onPlayerWeaponFire", me, function(wep, x, y, z) outputChatBox(wep.." "..x.." "..y.." "..z, source) end)
srun setElementPosition(me, -2954.279, 443.616, 2.727) -- location I used in video
Fire M4 and see sometimes messages come up, sometimes not.

### Version

MTA:SA Server v1.6-release-23409
Multi Theft Auto v1.6-release-23441

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

ArranTuna:
Seems fixed.

PR #5115, opened 2026-07-31
Fix bullet sync validation bypass

### LINKED ISSUE #5133: Make onPlayerWeaponFire cancellable (consistency with onWeaponFire)

`cancelEvent()` inside `onPlayerWeaponFire` does nothing. A script can observe a bullet-sync shot server-side, but cannot stop it. Two things are in the way.

**1. The return value is discarded.** In [`CGame::Packet_Bulletsync`](https://github.com/multitheftauto/mtasa-blue/blob/c3f1957a4007aefa8b2114028b298105a5e17934/Server/mods/deathmatch/logic/CGame.cpp#L2612):

```cpp
player->CallEvent("onPlayerWeaponFire", args);
RelayNearbyPacket(packet);
```

`CElement::CallEvent` already returns `!WasEventCancelled()`. The custom-weapon path 26 lines below uses it (`if (!weapon->CallEvent("onWeaponFire", args)) return;`), so `onWeaponFire` is cancellable and `onPlayerWeaponFire` is not.

**2. Sim sync relays the packet before Lua runs.** [`CSimPlayerManager::HandleBulletSync`](https://github.com/multitheftauto/mtasa-blue/blob/c3f1957a4007aefa8b2114028b298105a5e17934/Server/mods/deathmatch/logic/net/CSimPlayerManager.cpp#L338-L369) broadcasts on the sync thread, and `threadnet` defaults to on. So just honouring the return value would cancel the shot only for observers sim sync did not cover — a zone-dependent partial cancel, which would be worse than today.

This is not cosmetic: receiving clients apply the shot and its damage from the relayed packet ([`CNetAPI::ReadBulletsync`](https://github.com/multitheftauto/mtasa-blue/blob/c3f1957a4007aefa8b2114028b298105a5e17934/Client/mods/deathmatch/logic/CNetAPI.cpp#L2347)). Scripts validating a shot server-side today have to let it through and undo the damage afterwards, which players can see. `onClientPlayerWeaponFire` is not a substitute, since a server cannot rely on a client-side cancel.

Make `cancelEvent()` in `onPlayerWeaponFire` suppress the relay for every observer, matching `onWeaponFire`.

That means bullets must not take the sim-sync fast path while a cancel is still possible. Proposal, and the part I would like feedback on: keep the fast path as the default, and bypass it for bullet sync only while a Lua handler for `onPlayerWeaponFire` is attached. Servers not using the event keep today's latency; servers using it pay one main-thread hop. There is precedent for gating work on handler presence (`CMapEventManager::HasEvents()`), though it is per-element and would need extending.

I am happy to write the PR — raising this first because it changes sync behaviour and I would rather agree on the approach up front.

- Honour the return value only, leaving sim sync alone — smallest diff, but gives the partial cancel described above.
- Drop the sim-sync fast path for bullets entirely — consistent, but costs latency on every server.
- An explicit config/scripting toggle instead of inferring from handler presence — more predictable, more config surface.

### Additional context

Verified on `master` (c3f1957a4007aefa8b2114028b298105a5e17934) with a local server and client build.

#852 asks for the same capability on the client-side event. I know #4989, #5115, #5116 and #4682 are open against this area; this request is about the scripting API contract rather than bullet-sync validation, and I am happy to rebase on top of them.

### Security Policy

QueryOfficial:
#### Summary

Bullet sync packets were relayed to nearby players from the sync thread by
`CSimPlayerManager::HandleBulletSync`, which validated only `CVector::IsValid()` on the
start and end positions. The main thread path (`CBulletsyncPacket::Read` →
`CGame::Packet_Bulletsync`) applies far stricter checks, but it runs *after* the sync
thread has already broadcast the packet — and a packet rejected there is dropped
silently. A shot could therefore reach every nearby client while `onPlayerWeaponFire`
never fired for it, leaving the scripting layer with no record of it.

`CVector::IsValid()` only rejects NaN and infinity. A coordinate such as `0xFFFFFFFF`
reinterpreted as a float is finite (~4.29e9) and passes, so out of range positions were
relayed and reached `CWeapon::FireBullet()` on the receiving clients.

That same split is what #5122 reports from the other side: the stricter main thread range
check was rejecting ordinary handgun shots, so `onPlayerWeaponFire` stopped firing for
them while the sync thread kept delivering the bullets to nearby players. The range check
had two problems, both fixed here:

- It measured the shot against `GetWeaponStatsFromSkillLevel(weapon, playerStat)`. Server
  ped stats start at zero (`CPed.cpp`, `memset(&m_fStats[0], 0, ...)`), so every weapon was
  measured against its **poor** entry — a 30m envelope for a pistol the firing client had
  sized with hitman data. `CWeaponStatManager::GetMaxWeaponRange` now takes the widest tier
  the weapon has, which is the only bound bullet sync actually needs.
- Its tolerance was purely proportional (`range * 1.1`). The packet reports the gun muzzle
  as the start while the game traced the shot from the camera, so the measured length picks
  up the third person camera offset — an offset that does not shrink with the weapon range.
  On a 30m pistol that left 3m of headroom; on a rifle it was never noticeable. The
  validator now adds an absolute `RANGE_SLACK` on top of the proportional tolerance.

Net effect for a handgun: the accepted envelope goes from 33m to 48m, while a sniper rifle
moves from 330m to 345m against the same 500m hard cap.

This pull request:

- adds `Shared/mods/deathmatch/logic/CBulletSyncValidation.h`, a dependency free set of
  checks used by both sides: absolute position bounds, trajectory length against the
  weapon's own range, muzzle distance from the shooter, and the optional damage payload
- moves all bullet sync validation into `CBulletsyncPacket::Read`, so nothing is relayed
  before it has passed
- stops relaying bullet sync from the sync thread (`HasSimHandler()` now returns `false`),
  so every shot goes through `CGame::Packet_Bulletsync` and fires `onPlayerWeaponFire`
  before it is broadcast
- adds `CWeaponStatManager::GetMaxWeaponRange` and uses it instead of the shooter's skill
  stat, removing the client/server skill asymmetry from the range bound
- applies the same trajectory bounds to `CCustomWeaponBulletSyncPacket` and to the
  receiving side in `CNetAPI`
- adds 27 unit tests covering the validator

#### Motivation
Fixes #5133, #5122

The two relay paths had drifted apart. Hardening the main thread path alone made the
problem harder to observe rather than fixing it: the stricter check rejected the packet,
so `onPlayerWeaponFire` stopped firing, while the sync thread had already delivered the
same packet to the closest viewers. Keeping a single path removes the class of bug
rather than one instance of it.

This is also why #5122 has to be fixed in the same change rather than separately.
Collapsing the two paths turns that bug from "the event is missing" into "the shot is
never relayed at all" — the range bound and the relay path cannot be landed independently
without a window where handgun sync is worse than it is today.

The sync thread could not perform the decisive check even in principle: `CSimPlayer` is
a snapshot that does not carry the player's position, so the muzzle origin cannot be
compared there without extending and maintaining that snapshot. Bullet sync was also
only half handled by the sim system — it relayed to zone 0 viewers while zone 1/2 already
went through the main thread, which meant the same shot reached different observers by
different paths with different latency.

#4497 ("Bullet sync / onPlayerWeaponFire is still broken", closed) reports the same
observable symptom — the event firing inconsistently depending on location — so it may be
worth revisiting once this lands.

Two notes for reviewers:

- **Thread safety.** `CBulletsyncPacket::Read()` runs on the main thread:
  `CNetServerBuffer::ProcessPacket` queues the packet on the sync thread and
  `CNetServerBuffer::ProcessIncoming` invokes the packet handler during the main pulse.
  The new `g_pGame->GetWeaponStatManager()` and `CPlayer::GetPosition()` calls are
  therefore safe. The existing checks in that function already relied on this.
- **Range tolerance.** The previous check was `distanceSq > rangeSq * 1.1`, which is a
  ~4.9% tolerance on distance despite the comment saying 10%. The shared validator
  implements the documented intent (`(range * 1.1 + slack)²`). Combined with the switch to
  the widest skill tier this widens the accepted envelope — deliberately, since the old
  envelope was rejecting legitimate shots. `RANGE_SLACK` is the knob if you want it tighter;
  say the word and I will adjust it.

#### Test plan

Automated, run by the existing `Tests` workflow:

- `Tests/client/CBulletSyncValidation_Tests.cpp` — 27 cases covering position bounds
  (NaN, infinity, huge finite values, the inclusive limit), trajectory length against
  weapon range, the fallback cap when the range is unknown, hostile `maxRange` values
  (NaN / infinity / negative) that must not widen the envelope, muzzle distance on foot
  and in a vehicle, and the damage/hit zone payload.
- `TrajectoryAcceptsShortRangeWeaponWithCameraOffset` is the regression case for #5122: a
  30m weapon firing at 30/35/40m. It fails against the previous proportional-only formula
  and passes with the slack applied.
- `TrajectoryStillRejectsShortRangeWeaponAtRifleDistance` and
  `TrajectorySlackDoesNotScaleWithRange` pin the other side of it, so the slack cannot be
  widened later without a test noticing.
- Regression cases assert that a payload built around a huge-but-finite coordinate is
  rejected, and — importantly — that an ordinary shot fired from the same distant part of
  the map is still accepted, so the bounds are relative to the shooter rather than to a
  magic number.
- One case asserts `EXPECT_TRUE(start.IsValid())` on the crafted position, documenting
  why the previous check was not sufficient.

**Reproducing #5122 without this build:** on an unmodified 1.7 server, `setPedStat(player,
69, 1000)` moves the pistol from the poor (30m) entry to hitman (35m) and
`onPlayerWeaponFire` starts firing for handguns again. That is the skill-tier half of the
bug in isolation.

**Requires a populated server (not reproducible single-client):**

Bullet sync is only ever relayed to *other* players, so the following cannot be observed
from a single client and are listed here for anyone able to run this build with real
traffic:

1. Confirm normal combat is unaffected — on foot and in a vehicle, all bullet sync
   weapons (22-34), including sniper at long range.
2. Confirm `onPlayerWeaponFire` now fires for every shot other players can see,
   handguns included.
3. Watch main thread load on a busy server: bullet relay moved off the sync thread, so
   the cost of broadcasting bullets is now paid during the main pulse.

For anyone changing this code later: the validator is pure and has no engine
dependencies, so any new bound belongs in `CBulletSyncValidation.h` with a test beside
the existing ones, rather than in the packet classes.

#### Checklist


Dryxio:
"Not covered by the above, and worth a look from someone with a live server:

Confirm normal combat is unaffected — on foot and in a vehicle, all bullet sync
weapons (22-34), including sniper at long range.
Confirm onPlayerWeaponFire now fires for every shot other players can see."

So these changes were not tested ingame?

QueryOfficial:
I did test it locally and combat behaves normally on my end, with `onPlayerWeaponFire` firing as expected.

That list is what one machine can't answer, not what I skipped. Bullet sync is only ever relayed to *other* players, and item 3 (relay cost now paid during the main pulse) scales with concurrent shooters and tick load — neither of those shows up at ~0 ms ping with a couple of clients. So it wants a populated server before anyone takes my word for it, and I'm happy to add temporary instrumentation around `CGame::Packet_Bulletsync` if that would make the comparison measurable.

I've reworded the heading of that section in the description to say exactly this; the items themselves are unchanged.

FileEX:
Is this PR still relevant after #4682 was merged?

QueryOfficial:
Yes, the points and values I mentioned have already been implemented by Dutchman and you, so they are no longer applicable. This can be closed.

PR #5123, opened 2026-08-01
Add vehicleexplosions world special property

pube1:
#### Summary

Adds a new world special property `vehicleexplosions`, which stops the game from
destroying vehicles on its own:
`setWorldSpecialPropertyEnabled("vehicleexplosions", false)`.

- Added `WorldSpecialProperty::VEHICLEEXPLOSIONS` enum in Shared.
- Synced through SWorldSpecialPropertiesStateSync in
  a new data2 block, gated on a new eBitStreamVersion entry
  so peers built before this change keep working.
- Handled sync via `CMapInfoPacket` on the server and `CPacketHandler` on the
  client.
- Enforced in `CClientGame::VehicleDamageHandler`, which holds a damaged vehicle
  at `VEHICLE_BURNING_HEALTH` (250). Below that health the game sets the vehicle
  on fire and blows it up once the delayed explosion timer expires, so keeping
  health at that floor prevents both — for every vehicle type, without patching
  game code.
- Explicit destruction is unaffected: `blowVehicle()` still works while the
  property is disabled, since the property only suppresses what the game does by
  itself.

This also fixes a pre-existing bug that the change would otherwise make hot:
`CVehicleSA::SetHealth` cast every vehicle to `CAutomobileSAInterface*` in order
to clear `m_fBurningTime`, a field that only exists on automobiles. For bikes,
boats and trains that wrote past the end of the interface. The automobile-only
field moved to a `CAutomobileSA::SetHealth` override.

#### Motivation

Lets gamemode developers keep vehicles alive instead of having them explode and
disappear once they run out of health.

#### Test plan

- Tested in-game with a Lua resource calling
  `setWorldSpecialPropertyEnabled("vehicleexplosions", false)` on resource start,
  firing an RPG and a minigun at cars, planes and boats. Health stops at 250, the
  vehicles neither catch fire nor explode, and damage and deformation still apply
  above that.
- `blowVehicle()` still destroys a vehicle while the property is disabled.
- Updated the `SWorldSpecialPropertiesStateSync` round-trip test for the extra
  bit and added `LegacyPeerOmitsSecondBlock`, which covers a peer older than the
  new bitstream version. Full client test suite passes (305 tests).

#### Checklist


guibzo:
Seems pretty useful, at least most roleplay servers have to prevent it manually
hope it won't bypass `blowVehicle` tho

Client/game_sa/CGameSA.cpp
```diff
@@ -939,6 +939,23 @@ void CGameSA::SetVehicleBurnExplosionsEnabled(bool isEnabled)
...
 }
 
+void CGameSA::SetVehicleExplosionsEnabled(bool isEnabled)
+{
+    if (isEnabled == m_isVehicleExplosionsEnabled)
+        return;
+
+    if (isEnabled)
+    {
+        MemCpy((void*)0x6B8F80, "\x56\x8B\xF1", 3);  // CVehicle::BlowUpCar
```
FileEX: 0x6B8F80 points to ``CBike::VehicleDamage``. I don't think you tested your PR. That address points to a completely different location in memory, and those bytes will simply cause a crash.

Image

Shared/sdk/net/SyncStructures.h
```diff
@@ -2012,6 +2012,7 @@ struct SWorldSpecialPropertiesStateSync : public ISyncStructure
         bool flyingcomponents : 1;
         bool vehicleburnexplosions : 1;
         bool vehicleEngineAutoStart : 1;
+        bool vehicleexplosions : 1;
```
FileEX, line 2015: Another sign that you didn't test your changes. You need to increase BITCOUNT for this to work.

FileEX (changes requested):
The corresponding changes in `SyncMisc_Tests.cpp` are missing as well.

Client/game_sa/CGameSA.cpp
pube1: You're right, thanks for identifying what that address actually is. I dropped the memory patch entirely

Shared/sdk/net/SyncStructures.h
pube1: Fixed, though with a version-gated second block rather than bumping BITCOUNT — reasoning in my comment below.

FileEX: Since version 1.7 hasn't had an official release yet, we don't have to worry about the bitstream. So just increase the BITCOUNT and revert the changes in bitstream.h and the tests.

pube1: Done, moved it back into data, BITCOUNT is 21 now, and bitstream.h is reverted. The round-trip test keeps the updated bit count and a check for the new flag; the version-gate test is gone with the gate.

FileEX:
Why don't you just disable GTA's vehicle ignition when it has low HP?

Xenius97:
I've already done this by directly disabling blow-up timers, pr: https://github.com/multitheftauto/mtasa-blue/pull/4546

FileEX:
Closed in favor of #4546

PR #5128, opened 2026-08-02
Fix custom partial animations from custom IFP banks staying rigid

### LINKED ISSUE #5121: Can't load partial animations with ``engineLoadIFP``
### Describe the bug

#5101 fixed the replacement of partial animations but it did not fix the loading of custom partial animations in custom banks.
For example if I load the gangs.ifp and run

```lua
setPedAnimation(me, "custom_bank", "prtial_gngtlkD", 1, true, true, true, true, 0, false)
```
I will be stuck in position, meanwhile if I run the original animation that already comes with GTA
```lua
setPedAnimation(me, "gangs", "prtial_gngtlkD", 1, true, true, true, true, 0, false)
```
I will be able to move freely

### Steps to reproduce

1. Load custom bank with partial animations
2. Try to use them

### Version

Client: v1.7-untested-26756

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary

Fixes custom partial animations played straight from a custom IFP bank (via `engineLoadIFP` + `setPedAnimation`) leaving the ped rigid, the same padding problem #5101 fixed for `engineReplaceAnimation`.

`CClientIFP` pads every loaded animation out to 32 bones regardless of what the source IFP actually defines. #5101 trimmed that padding by comparing against the built-in animation being replaced, but an animation played directly from a custom bank has no built-in original to compare against, so the fix never applied there and the padding kept driving the root, pelvis and legs.

`CClientIFP` now records which bones an animation actually defines while parsing it. `CAnimBlendAssociation` gains `RestrictToBones(mask)`, a sibling of `RestrictToBonesOf` driven by that record instead of a built-in original; `CClientPed` applies it once the animation is actually playing.

#### Motivation

Continuation of #5101. That fix only covered animations replacing a built-in one; partial animations loaded into their own custom bank and played directly still went rigid. Fixes #5121.

#### Test plan

Loaded a custom IFP with a partial animation (upper body only, e.g. `crry_prtial`) into a custom bank via `engineLoadIFP`, played it with `setPedAnimation` on a ped. Before the fix the ped froze in place; after, the custom animation plays on the upper body while walking, turning and leg movement keep working normally. Also re-tested #5101's original `engineReplaceAnimation` scenario to confirm that path is untouched.

Test script, thanks to @RughCuttle
[partialtest.zip](https://github.com/user-attachments/files/30626244/partialtest.zip)

#### Checklist

### DIFF
--- a/Client/game_sa/CAnimBlendAssociationSA.cpp
+++ b/Client/game_sa/CAnimBlendAssociationSA.cpp
@@ -149,6 +149,19 @@ void CAnimBlendAssociationSA::RestrictToBonesOf(const CAnimBlendStaticAssociatio
     }
 }
 
+void CAnimBlendAssociationSA::RestrictToBones(std::bitset<32> animatedBonesMask)
+{
+    // Same padding problem as RestrictToBonesOf, but for animations played directly from a custom IFP
+    // bank rather than replacing a built-in one. There's no built-in association to compare against, so
+    // the caller passes the set of bones the source IFP animation itself defines (see CClientIFP).
+    for (unsigned short i = 0; i < m_pInterface->cNumBlendNodes; i++)
+    {
+        const bool isAnimated = i < animatedBonesMask.size() && animatedBonesMask.test(i);
+        if (!isAnimated)
+            m_pInterface->pAnimBlendNodeArray[i].pAnimSequence = nullptr;
+    }
+}
+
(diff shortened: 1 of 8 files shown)


RushCuttle:
Works. Tested this and it makes all partials play after engineLoadIFP like they should. now you can freeze by using parm time at -1 and it'll 100% freeze.

FileEX (approved):

PR #5139, opened 2026-08-02
Fix setElementDimension not working on buildings

TheCrazy17:
#### Summary

`CClientBuilding` extends `CClientEntity` directly, not `CClientStreamElement`, so unlike objects, peds and vehicles it never got automatic dimension-based streaming; `SetDimension` only stored the value, nothing ever compared it to the local player's dimension to actually show or hide the building.

Added the same `RelateDimension` pattern already used by `CClientPointLights`, `CClientRadarArea`, `CClientRadarMarker` and `CClientWater`, the other element types that also aren't stream elements: `CClientBuildingManager` tracks the local player's dimension and calls `RelateDimension` on every building when it changes; `CClientBuilding` overrides `SetDimension` to re-evaluate itself the same way. Also fixed `RestoreDestroyed()`, which recreated every building unconditionally on pool resize, ignoring dimension entirely.

#### Motivation

`setElementDimension` silently did nothing on buildings; scripts had no way to hide a `createBuilding` building in a different dimension.

#### Test plan

Created a building, moved the local player between dimensions and called `setElementDimension` on the building both ways: it now shows only when its dimension matches the player's, and correctly disappears/reappears when either side changes.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientBuilding.cpp
+++ b/Client/mods/deathmatch/logic/CClientBuilding.cpp
@@ -27,7 +27,7 @@ CClientBuilding::CClientBuilding(class CClientManager* pManager, ElementID ID, u
     m_pModelInfo = g_pGame->GetModelInfo(usModelId);
     SetTypeName("building");
     m_pBuildingManager->AddToList(this);
-    Create();
+    RelateDimension(m_pBuildingManager->GetDimension());
     UpdateSpatialData();
 }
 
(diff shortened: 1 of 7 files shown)


FileEX:
Slightly off-topic, but do you think we could add buildings to MTA's streaming?

TheCrazy17:
> Slightly off-topic, but do you think we could add buildings to MTA's streaming?

It should be possible, and it shouldn't be too difficult. I'd have to look into it.

PlatinMTA:
Considering the change, is the default dimension for buildings just going to be 0 or -1? Making it backwards compatible would mean it should be -1 (stream to all dimensions) but idk if that would be confusing

FileEX:
The fact that buildings are visible in all dimensions isn't really the expected behavior. That's why imo they should default to dimension 0, just like any other element in MTA.

TheCrazy17:
> The fact that buildings are visible in all dimensions isn't really the expected behavior. That's why imo they should default to dimension 0, just like any other element in MTA.

In that case, it already works that way. The PR could be merged as it is right now.

FileEX (approved):

Fernando-A-Rocha:
So the dimension of a building is 0 or -1 by default?

Xenius97:
@FileEX @TheCrazy17 Hi! This change will be backward incompatible. By default, buildings were visible in all dimensions until now, but this change will completely break that behavior. My suggestion would also be to set the default dimension to -1 so that they remain visible everywhere, just like before.

PR #5140, opened 2026-08-02
Fix testSphereAgainstWorld returning nil hitElement for buildings

### LINKED ISSUE #5136: processLineOfSight and testSphereAgainstWorld don't return hitElement for buildings created by createBuilding
### Describe the bug

`CPoolsSA` doesn't check the building pool at all in its `GetEntity` and `GetClientEntity` methods. As a result, any function that relies on `GetEntity` (including many hooks and the SA layer) won't receive the correct `CBuildingSA`/`CClientBuilding` object. One example is `processLineOfSight` and `testSphereAgainstWorld`, which return nil as hitElement when they hit a building created with **createBuilding**.

### Steps to reproduce

1.
```lua
crun obj = createBuilding(1337, me.position)
```
2.
```lua
crun tempObj = createObject(1337, me.position) radius = getElementRadius(tempObj) destroyElement(tempObj)
```
3.
```lua
crun hit, element = testSphereAgainstWorld(obj.position.x, obj.position.y, obj.position.z, radius, nil, true) print(hit, element) -- returns true, nil
```

### Version

Multi Theft Auto v1.6-release-24129

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary
Fixes `hitElement` returning nil when `testSphereAgainstWorld` (and `processLineOfSight`) hits a building created with `createBuilding`.

#### Motivation
Resolves #5136. `CPoolsSA::GetEntity` never checked the buildings pool, so `CEntity*` resolution for a building always fell through to null. Even after that, `testSphereAgainstWorld` still returned nil because it resolves its element through `GetStoredPointer`, which `CClientBuilding::Create` never set on the native entity, unlike objects, peds and vehicles.

#### Test plan
Created a building with `createBuilding`, called `testSphereAgainstWorld` and `processLineOfSight` against it. Before the fix, both returned a hit with a nil `hitElement`; after the fix, `hitElement` correctly resolves to the building.

#### Checklist

### DIFF
--- a/Client/game_sa/CBuildingsPoolSA.cpp
+++ b/Client/game_sa/CBuildingsPoolSA.cpp
@@ -63,6 +63,16 @@ CClientEntity* CBuildingsPoolSA::GetClientBuilding(CBuildingSAInterface* pGameIn
     return m_buildingPool.entities[static_cast<size_t>(poolIndex)].pClientEntity;
 }
 
+CEntity* CBuildingsPoolSA::GetBuilding(CBuildingSAInterface* pGameInterface) const noexcept
+{
+    std::int32_t poolIndex = (*m_ppBuildingPoolInterface)->GetObjectIndexSafe(pGameInterface);
+
+    if (poolIndex == -1)
+        return nullptr;
+
+    return m_buildingPool.entities[static_cast<size_t>(poolIndex)].pEntity;
+}
+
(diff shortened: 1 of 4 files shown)


FileEX:
~~What about ``CPoolsSA::GetClientEntity`` method?~~

FileEX (approved):

PR #5141, opened 2026-08-02
Fix water staying invisible across dimensions

### LINKED ISSUE #849: [Server-side] createWater dimension problems
Hi guys,

I found a bug with water dimension.
AFAIK this was fixed some time ago, but this issue is different.

The problem is with setElementDimension using water element.
When I am in for ex. 1st dimension and use createWater and then setElementDimension and change water dimension to 1st, water is invisible and no water effects/sounds.

When I am in 0 dimension and use createWater it works fine and water is visible.
I also can change water dimension to other and it will work and visible to me (if I change my dimension to water dimension it is fine)

Water is invisible to all players who have other dimension than 0.
Water is also still invisible when player switch to 0 or water switch to player's dimension.

**Working code:**
`setElementDimension(source, 0)`
`water = createWater(-50, -50, 2, 50, -50, 2, -50, 50, 2, 50, 50, 2)`
-- Water become visible to our eyes

`setElementDimension(water, 1)`
-- Water is invisible now because different dimensions

`setElementDimension(source, 1)`
-- Water again become visible

**Code generating bug:**
`setElementDimension(source, 1)`
`water = createWater(-50, -50, 2, 50, -50, 2, -50, 50, 2, 50, 50, 2)`
-- Water is invisible because different dimensions

`setElementDimension(source, 0)`
-- Water is still invisible even if water and player are same dimension

**2nd code generating bug:**
`setElementDimension(source, 1)`
`water = createWater(-50, -50, 2, 50, -50, 2, -50, 50, 2, 50, 50, 2)`
`setElementDimension(water, 1)`
-- Water is invisible, because player was in other dimension (not 0)

`setElementDimension(source, 0)`
`setElementDimension(water, 0)`
-- Water is still invisible, because when water was created, player was in other dimension (not 0)

If any questions, I'll try to reply.

Cheers ;)

TheCrazy17:
#### Summary
Fixes water elements created in a dimension different from the local player's current dimension being deleted instead of just hidden.

#### Motivation
Resolves #849. `CClientWater`'s constructor calls `RelateDimension()` right away, comparing its own dimension (still the default 0, since the real value hasn't been read from the network packet yet) against the local player's current dimension. On a mismatch it calls `Destroy()` instead of `Create()`, leaving `m_pPoly` null.

`Packet_EntityAdd` treated that as a creation failure: `if (!pWater->Exists()) { delete pWater; pWater = NULL; }`, deleting the element right after construction, before the packet handler even reaches the line further down that applies the water's real dimension (`pEntity->SetDimension(usDimension)`). That check dates back to when water had no dimension concept and `m_pPoly == nullptr` could only mean the native call genuinely failed (invalid vertices, water pool full); it was never updated when dimension support was added, so it can no longer tell a real failure apart from water that's just correctly waiting for its dimension to be applied.

#### Test plan
Created water server-side in dimension 0 while standing in dimension 1, then switched to dimension 0. Before the fix, the water never appeared; after the fix, it shows up correctly once dimensions match.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CPacketHandler.cpp
+++ b/Client/mods/deathmatch/logic/CPacketHandler.cpp
@@ -4191,11 +4191,6 @@ retry:
                         pWater =
                             new CClientWater(g_pClientGame->GetManager(), EntityID, vecVertices[0], vecVertices[1], vecVertices[2], vecVertices[3], bShallow);
                     }
-                    if (!pWater->Exists())
-                    {
-                        delete pWater;
-                        pWater = NULL;
-                    }
                     pEntity = pWater;
                     break;
                 }


Client/mods/deathmatch/logic/CPacketHandler.cpp
```diff
@@ -4191,11 +4191,6 @@ void CPacketHandler::Packet_EntityAdd(NetBitStreamInterface& bitStream)
                         pWater =
                             new CClientWater(g_pClientGame->GetManager(), EntityID, vecVertices[0], vecVertices[1], vecVertices[2], vecVertices[3], bShallow);
                     }
-                    if (!pWater->Exists())
-                    {
-                        delete pWater;
-                        pWater = NULL;
-                    }
```
FileEX, line 4198: Okay, but what if creating the water actually fails for some reason? We shouldn't just leave it like that

FileEX (changes requested):

Client/mods/deathmatch/logic/CPacketHandler.cpp
TheCrazy17: Exists() there doesn't report a failed creation, it reports that the poly wasn't built yet. The constructor relates against the default dimension 0 before Packet_EntityAdd applies the real one after the switch, so any water in another dimension got deleted before SetDimension could relate it. That's #849.

Invalid geometry is already rejected server side by CWater::Valid(), so it never reaches this packet. What's left is pool exhaustion or no water zone, both transient, and RelateDimension retries Create() on the next dimension change, so keeping the element is what lets it recover.

That's how I understood it while digging into it, so correct me if I'm wrong or if I missed something.

FileEX (approved):

PR #5143, opened 2026-08-03
Add glob support for map files

### LINKED ISSUE #4902: Add Glob to .map

Currently, glob doesn't work with .map files; it would be interesting to add glob support for .map files as well.

```xml
<meta>
     <map src="*.map" dimension="0" />
</meta>
```

_No response_

### Additional context

_No response_

### Security Policy

efejotaese2:
#### Summary

Adds glob support for `<map>` entries in `meta.xml`, using the existing resource file glob handling.

For example:

```xml
<map src="maps/*.map" dimension="0" />
```
```xml
<map src="maps/**/*.map" dimension="0" />
```
```xml
<map src="*.map" dimension="0" />
```

All matching map files are loaded with the attributes defined on the `<map>` node.

#### Motivation

`<script>` and `<file>` entries already support glob patterns, while `<map>` entries currently require an explicit file path.

This resolves #4902.

#### Test plan

Tested:
Multiple `.map` files using `*.map`.
Recursive **/*.map
Glob with no matches
Normal paths

In all cases, the resource behaved as expected and matching map files were loaded correctly.

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/CResource.cpp
+++ b/Server/mods/deathmatch/logic/CResource.cpp
@@ -2207,25 +2207,47 @@ bool CResource::ReadIncludedMaps(CXMLNode* pRoot)
 
             if (!strFilename.empty())
             {
-                std::string strFullFilename;
                 ReplaceSlashes(strFilename);
 
-                if (IsFilenameUsed(strFilename, false))
-                {
-                    CLogger::LogPrintf("WARNING: Duplicate map file in resource '%s': '%s'\n", m_strResourceName.c_str(), strFilename.c_str());
-                }
-
-                // Grab the file (evt extract it). Make a map item resource and put it into the resourcefiles list
-                if (IsValidFilePath(strFilename.c_str()) && GetFilePath(strFilename.c_str(), strFullFilename))
-                {
-                    m_ResourceFiles.push_back(new CResourceMapItem(this, strFilename.c_str(), strFullFilename.c_str(), &Attributes, iDimension));
-                }
-                else
+                if (!IsValidFilePath(strFilename.c_str()))
                 {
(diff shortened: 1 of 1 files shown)


Server/mods/deathmatch/logic/CResource.cpp
```diff
@@ -2207,25 +2207,47 @@ bool CResource::ReadIncludedMaps(CXMLNode* pRoot)
...
+                    }
+
+                    if (GetFilePath(strFilePath.c_str(), strFullFilename))
+                    {
+                        m_ResourceFiles.push_back(new CResourceMapItem(this, strFilePath.c_str(), strFullFilename.c_str(), &Attributes, iDimension));
+                    }
+                }
+
+                if (glob::has_magic(strFilename))
+                    m_ResourceFilesCountPerDir[strFilename] = vecFiles.size();
```
FileEX: What is the purpose of having this here again?

efejotaese2: That's for tracking the match count when the glob finds files, the earlier one is just for the empty case. 
I can move it up and only set it once, i think thats change could be good

efejotaese2: Moved it before the empty check so it's only set once.

FileEX: There's still two cases

efejotaese2: i uploaded it bad, im pushing it right now

efejotaese2: done, thank u

FileEX (approved):

PR #5146, opened 2026-08-03
Fix engineImageLink not applying to weapon models

### LINKED ISSUE #5145: engineImageLink* doesn't work with weapon models
### Describe the bug
If you try to replace weapon models with engineImageLink* functions then nothing happens, you will see only default models.

### Steps to reproduce
Create an image archive
Replace weapon model (eg 346 colt model)

### Version
Both 1.6 & 1.7 version

### Security Policy

TheCrazy17:
#### Summary

`CClientIMG::LinkModel` now restreams weapon models the same way it already does for vehicle models.

mta-screen_2026-08-03_13-52-50

#### Motivation

Resolves #5145.

`engineImageLinkDFF`/`engineImageLinkTXD` link a new model through `CClientIMG::LinkModel`, which already restreams vehicle models so they pick up the new one. Weapon models were missing that call, so a weapon a ped was already holding kept its old RW clump and only picked up the replacement once something else happened to restream it. `CClientDFF`, `CClientTXD` and `CClientGame` already call `CClientPedManager::RestreamWeapon` for the same case, so `CClientIMG` now does the same.

#### Test plan

[imgweapon.zip](https://github.com/user-attachments/files/30670283/imgweapon.zip)

With a ped already holding a M4, linking a replacement dff for model 356 now swaps the held weapon model.

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientIMG.cpp
+++ b/Client/mods/deathmatch/logic/CClientIMG.cpp
@@ -224,6 +224,11 @@ bool CClientIMG::LinkModel(unsigned int uiModelID, size_t uiFileID)
     if (CClientVehicleManager::IsValidModel(uiModelID))
         g_pClientGame->GetVehicleManager()->RestreamVehicles(static_cast<unsigned short>(uiModelID));
 
+    // Weapon models already in a ped's hand keep their old RW clump until the weapon slot is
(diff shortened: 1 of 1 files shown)


Xenius97:
Wow, that was fast. Thanks!

FileEX (approved):

PR #5152, opened 2026-08-05
Fix bikes breaking when their steering lock is set to zero

### LINKED ISSUE #494: Setting steeringlock to 0 on bikes will mess up the game
**Describe the bug**
If you set steeringlock to 0 on motorcycle it will bug you until respawn, showing your location elsewhere as your ped anim movements on the bike, and turns your ped into weird angles, see video for that. Can only resume playing on respawning, movement can be forced only by melee, (and what's seen in the video)

I'm unsure if it's limited to bikes/motorcycles as i only tested it on 100% reproduction with NRG now but think so as car i tried didnt cause it after the same amount of seconds driving with steeringlock 0.

**To Reproduce**
1) For example, get an NRG
2) srun setVehicleHandling(getRandomPlayer().vehicle, "steeringLock", 0)
3) Drive awhile and have chance of getting this randomly

**Expected behavior**
It should just not break

**Screenshots**
Video: https://www.youtube.com/watch?v=234Z1s_9gGs&t=22 (must see to understand)

v1.5-release-7323

v1.5-release-7323

**Additional context**
From https://bugs.mtasa.com/view.php?id=8653

TheCrazy17:
#### Summary

Sets the bike steer and lean ratios to zero when a vehicle's `steeringLock` handling is zero, instead of dividing by it.

#### Motivation

Resolves #494. `CBike::ProcessControl` divides the steer angle by the steering lock in radians to get a ratio. The steer angle is itself the steering lock times the input, so a lock of zero makes this 0 / 0, a NaN rather than an infinity. Cars survive the same division because an infinity trips their "greater than 1" clamp, while a NaN compares false against every clamp and passes straight through. It reaches the rider lean angle, and `CBike::CalculateLeanMatrix` builds the bike and rider matrices from it, stretching the ped until respawn.

#### Test plan

Repro from the issue:

```lua
setVehicleHandling(getRandomPlayer().vehicle, "steeringLock", 0)
```

Get an NRG-500, set its steeringLock to 0, and crash into something. Before this change the ped stretches out and stays broken until respawn. After it, the bike behaves like a normal collision.

Also checked that bikes with a normal steering lock still steer and lean correctly, since only the zero case takes the new path.

#### Checklist

* [x] Your code should follow the coding guidelines (https://wiki.multitheftauto.com/index.php?title=Coding_guidelines).


FileEX:
Please resolve conflicts & fix build

TheCrazy17:
> Please resolve conflicts & fix build

Done ✅

FileEX (approved):

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -63,6 +63,90 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+
+// This one divides a value already on the FPU stack, so zero has to replace it.
+static void __declspec(naked) HOOK_CBike__ProcessControl_SteerRatio()
+{
+    MTA_VERIFY_HOOK_LOCAL_SIZE;
+
+    // clang-format off
+    __asm
+    {
+        push    eax
```
FileEX: It looks like you don't need to preserve this register, since it's modified at 0x6BAC22 anyway

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -63,6 +63,90 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+
+// This one pushes a new value onto the FPU stack, so zero is simply pushed instead.
+static void __declspec(naked) HOOK_CBike__ProcessControl_LeanRatio()
+{
+    MTA_VERIFY_HOOK_LOCAL_SIZE;
+
+    // clang-format off
+    __asm
+    {
+        push    eax
```
FileEX: It looks like you don't need to preserve this register, since it's modified at 0x6BBAE2.

FileEX (changes requested):

FileEX:
Please resolve conflicts and re-test PR after last commit

TheCrazy17:
Tested and retested just to be sure, and it’s working correctly.

https://streamable.com/04ckk4
With Linkin Park playing in the background 😂

FileEX (approved):

PR #5160, opened 2026-08-07
Implement NaN/Inf coordinate check in BulletSync packet

x6c85:
Added validation to reject NaN/Inf coordinates in BulletSync packet to prevent exploits.

#### Summary
Added an explicit `std::isfinite()` check on the `m_start` and `m_end` vectors
in `CGame::Packet_Bulletsync` before they are used in any range calculation or
relayed to nearby players via `RelayNearbyPacket`. Also added a defensive
null check on the `stats` pointer returned by `GetWeaponStatsFromSkillLevel`.

#### Motivation
Cheaters were sending BulletSync packets with `NaN`/`Inf` values in the
`m_end` (and potentially `m_start`) coordinates. The existing range check

```cpp
if (distanceSq > maxRangeSq)
    return;
```

relies on a floating point comparison, but any comparison against `NaN`
always evaluates to `false` in IEEE 754. This meant a packet with `NaN`
coordinates silently passed the range check regardless of the actual
distance, and was then relayed to nearby players and pushed into the
`onPlayerWeaponFire` Lua event with invalid coordinates, crashing clients
that received it.

Validating the vectors with `std::isfinite()` before they're used anywhere
closes this off entirely, since `NaN` and `Inf` are rejected outright rather
than relying on the range comparison to catch them.

#### Test plan
- Reviewed the code path manually: confirmed `m_start`/`m_end` are now
  validated before any arithmetic (`LengthSquared()`) or before being pushed
  into Lua args / relayed via `RelayNearbyPacket`.
- Not manually reproduced against a live cheat client — reviewers/maintainers
  with access to a repro client are encouraged to verify that a crafted
  BulletSync packet with `NaN`/`Inf` end coordinates is now dropped silently
  instead of being relayed to nearby players.
- Regression check: normal gameplay (shooting within and at the edge of a
  weapon's range) is unaffected, since finite in-range coordinates still pass
  all existing checks unchanged.

#### Checklist

### DIFF
--- a/Server/mods/deathmatch/logic/CGame.cpp
+++ b/Server/mods/deathmatch/logic/CGame.cpp
@@ -2571,26 +2571,39 @@ void CGame::Packet_Bulletsync(CBulletsyncPacket& packet)
     if (!player || !player->IsJoined())
         return;
 
+    // Reject NaN/Inf coordinates before any math or relaying happens.
+    // This is the actual exploit: NaN in a floating point comparison
+    // (distanceSq > maxRangeSq below) always evaluates to false, so a
+    // packet with NaN start/end coordinates was silently passing the
+    // range check regardless of how "far" the shot claimed to be.
+    if (!std::isfinite(packet.m_start.fX) || !std::isfinite(packet.m_start.fY) || !std::isfinite(packet.m_start.fZ) ||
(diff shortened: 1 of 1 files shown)


FileEX:
#4682

PR #5016, opened 2026-07-05
Add vehicle model special ability API part1

Allerek:
> [!IMPORTANT]
> Everything is a subject to change

#### Summary

This pull request adds a client-side vehicle special ability API for applying selected GTA SA native vehicle behaviors to vehicles whose model ID would not normally receive those behaviors.

New Lua functions:

```lua
setVehicleModelSpecialAbility(vehicle, ability)
getVehicleModelSpecialAbility(vehicle)
```

Aliases are also available:

```lua
setVehicleSpecialAbility(vehicle, ability)
getVehicleSpecialAbility(vehicle)
```

Supported ability names:

```lua
"none"
"towtruck"
"packer"
"fireTruckWater"
"fireTruckLadder"
```

Example usage:

```lua
local model = engineRequestModel("vehicle", 525)
engineImportTXD(engineLoadTXD("custom_towtruck.txd"), model)
engineReplaceModel(engineLoadDFF("custom_towtruck.dff"), model)

local vehicle = createVehicle(model, x, y, z)
setVehicleModelSpecialAbility(vehicle, "towtruck")
```

```lua
local ability = getVehicleModelSpecialAbility(vehicle)
if ability == "towtruck" then
    outputDebugString("Vehicle has towtruck behavior")
end
```

The API currently exposes these native behavior paths:

* `towtruck`
  * Enables the original towtruck hook movement, towbar lookup, tow line physics, attach behavior, and detach/drop behavior.
  * The actual tow attach point still comes from the model data used by GTA's native towbar lookup.

* `packer`
  * Enables original packer-style `misc_a` ramp rotation.
  * Uses MTA-side handling for custom moving collision/audio where native paths were unsafe for custom model layouts.

* `fireTruckWater`
  * Enables original firetruck water cannon behavior from model `407`.
  * Uses GTA's native `FireTruckControl` path for aiming, effects, audio, and hit handling.

* `fireTruckLadder`
  * Enables original firetruck ladder movement behavior from model `544`.
  * Initializes the fire ladder chassis state that GTA normally sets only when constructing a real `MODEL_FIRELA` vehicle.

* `none`
  * Explicitly disables the special ability override for that vehicle.

Getter behavior:

```lua
getVehicleModelSpecialAbility(vehicle)
```

returns one of:

```lua
"none"
"towtruck"
"packer"
"fireTruckWater"
"fireTruckLadder"
```

If no override has been set, original GTA models report their native default ability:

```lua
-- model 525
[...]

PR #5161, opened 2026-08-08
Fix trailers not attaching to custom vehicle models

### LINKED ISSUE #4525: Attach trailer to engineRequestModel truck
### Describe the bug

if you create your own truck using engineRequestModel, then it becomes impossible to attach a trailer

### Steps to reproduce

local id = engineRequestModel("vehicle", 514)

local txdData = engineLoadTXD("petro.txd")
engineImportTXD(txdData, id)

local dffData = engineLoadDFF("petro.dff")
engineReplaceModel(dffData, id)
...
setElementModel(model, id) -- model from server

### Version

_No response_

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary

Trailers now attach to trucks whose model was created with engineRequestModel, exactly as they do to the stock ones. Nothing new is exposed to scripts; it works on its own.

#### Motivation

Fixes #4525. Part of #1861. CVehicle::UpdateTractorLink asks the towing vehicle where its tow bar sits every frame, and CAutomobile::GetTowBarPos answers from a switch over the model index: the trucks built to pull a trailer return their misc_a dummy, everything else gets a generic point off the front bumper. A model created by engineRequestModel carries an ID of its own, so a cloned truck matches no case and its trailer ends up hanging off the wrong end.

The hook redirects only the model index that switch reads, to the ID the model was cloned from. The branch it then takes reads nothing but the vehicle's own misc_a dummy, so the clone resolves its tow bar from its own DFF. Stock vehicles have no parent ID, so they keep the index they had and take the branch they always did.

The tow truck and the tractor are deliberately left out: they derive the bar from their hoist angle, and the rest of their behaviour stays gated on the stock model index, so a clone would only half work.

#### Test plan

Clone a Tanker with engineRequestModel, apply petro.dff and petro.txd, and setElementModel a vehicle to it. Attach a stock trailer with attachTrailerToVehicle and confirm it sits on the tow bar and follows correctly while driving; backing the cloned truck up to a loose trailer picks it up too. Confirm a stock truck and trailer are unchanged.

[ctruck.zip](https://github.com/user-attachments/files/30848644/ctruck.zip)

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
@@ -9,6 +9,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <enums/VehicleType.h>
 
 static bool __fastcall AreVehicleDoorsUndamageable(CVehicleSAInterface* vehicle)
 {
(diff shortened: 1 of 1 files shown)


Fernando-A-Rocha:
@Allerek

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -62,6 +63,85 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+        case VehicleType::VT_TUG:
+            return true;
+        default:
+            return false;
+    }
+}
+
+static std::uint32_t __fastcall GetTowBarModelId(CVehicleSAInterface* vehicle)
+{
+    const std::uint32_t modelId = vehicle->m_nModelIndex;
```
FileEX: static_cast? I think explicit casting is good practice.

FileEX (dismissed):

FileEX (approved):

Fernando-A-Rocha:
@TheCrazy17 What's your plan for tow trucks? Did you see what Allerek and FileEX were working on?
- #5016
- #4330

TheCrazy17:
> What's your plan for tow trucks? Did you see what Allerek and FileEX were working on?

My idea is to enable the native functionality of each vehicle on the cloned vehicles; I might do the same for the tow truck and the tractor, but only as a bug fix, just enabling the existing native functionality without adding any new features, so I don’t interfere with the work Allerek and FileEX are doing. Maybe the hooks and other stuff could be useful for them as well.

PR #5162, opened 2026-08-08
Fix ZR-350 pop up headlights not working on custom vehicle models

TheCrazy17:
#### Summary

The ZR-350's pop up headlights, misc_a swinging out before the lights turn on, now work on a clone of the model created with engineRequestModel, exactly as they do on the stock 477.

#### Motivation

Part of #1861. CAutomobile::PreRender only swings the misc_a component for model 477, and CVehicle::DoVehicleLights only waits for that swing to finish before turning the headlights on for model 477 too. A model created by engineRequestModel carries an ID of its own, so a cloned ZR-350 matches neither check and its headlights stay shut and never light up.

Both checks now also accept the model the clone was cloned from, so a clone behaves exactly like the stock car and everything else is untouched.

#### Test plan

Clone model 477 with engineRequestModel, replace it with the DFF and TXD, and setElementModel a vehicle to it. Confirm the headlights swing out and light up when turned on, and swing back in when turned off, same as an unmodified ZR-350.

[zrtest.zip](https://github.com/user-attachments/files/30883512/zrtest.zip)

#### Checklist

### DIFF
--- a/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
@@ -79,7 +79,7 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
 //////////////////////////////////////////////////////////////////////////////////////////
 static bool __fastcall HasPopUpHeadlights(CVehicleSAInterface* vehicle)
 {
-    const std::uint32_t modelId = vehicle->m_nModelIndex;
+    const std::uint32_t modelId = static_cast<std::uint32_t>(vehicle->m_nModelIndex);
     if (modelId == static_cast<std::uint32_t>(VehicleType::VT_ZR350))
         return true;
 
--- a/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
@@ -9,6 +9,7 @@
  *****************************************************************************/
 
 #include "StdInc.h"
+#include <enums/VehicleType.h>
 
 static bool __fastcall AreVehicleDoorsUndamageable(CVehicleSAInterface* vehicle)
 {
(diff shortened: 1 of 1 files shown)


Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -62,6 +63,96 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+    MTA_VERIFY_HOOK_LOCAL_SIZE;
+
+    // clang-format off
+    __asm
+    {
+        pushad
+        mov     ecx, esi
+        call    HasPopUpHeadlights
+        test    al, al          // popad leaves the flags alone, so this survives it
+        popad
```
FileEX: Do you really need pushad and popad here? I think only ESI needs to be preserved.

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -62,6 +63,96 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+    MTA_VERIFY_HOOK_LOCAL_SIZE;
+
+    // clang-format off
+    __asm
+    {
+        pushad
+        mov     ecx, esi
+        call    HasPopUpHeadlights
+        test    al, al
+        popad
```
FileEX: Do you really need pushad and popad here? I think only ESI needs to be preserved.

TheCrazy17: Applied on the first one. There I can't drop it entirely though, ecx needs to survive the call there too.

TheCrazy17:
Done (and re-tested). Still working.

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -63,6 +63,98 @@ static void __declspec(naked) HOOK_CDamageManager__ProgressDoorDamage()
...
+// deployment to finish before it lets the headlights come on. A model created by engineRequestModel
+// carries an ID of its own, so a cloned ZR-350 misses both and its headlights stay shut.
+//
+// Both hooks answer the same question, whether this vehicle should behave as a ZR-350, and let the
+// model it was cloned from count. Everything else keeps the model index it already had.
+//
+//////////////////////////////////////////////////////////////////////////////////////////
+static bool __fastcall HasPopUpHeadlights(CVehicleSAInterface* vehicle)
+{
+    const std::uint32_t modelId = vehicle->m_nModelIndex;
```
FileEX: static_cast?

TheCrazy17: Oh, right, added.

FileEX (approved):

Issue #5147, opened 2026-08-04
CPU Temperature Jumps to 70°C Immediately After Launch


Katjafejleszto:
### Describe the bug

For AMD processors, as soon as you launch the game, the CPU temperature immediately rises to around 70°C, even while just sitting in the client. After a random period of 15–30 minutes, the temperature eventually returns to a relatively normal level, but it stays high until then.

My hardware:

* Ryzen 5 9600X
* DeepCool LS720 AIO

### Steps to reproduce

Start game

### Version

_Release_

### Relevant log output

```shell

```

### Security Policy

- [x] I have read and understood the Security Policy and this issue is not security related.

FileEX:
Confirmed - the temperature increases, but the usage doesn't.

<img width="1694" height="927" alt="Image" src="https://github.com/user-attachments/assets/bd109926-dcee-4f61-aecc-1fa59b048db1" />

dnl-k:
A decent jump in temperature when opening any application is fairly normal.
Though in this case it's certainly weird that it A) doesn't cool back down (or atleast not relatively quickly) and B) cools back down immediatly when unfocusing the MTA Window.
I'm seeing the same beheaviour (except for the fact that mine reaches max. 60°C) on a Ryzen 7 9800X3D paired with a bequiet Dark Rock Pro 5. This cannot be observed when running just the base game.

FileEX:
Fixed in #5169

PR #5169, opened 2026-08-10
Reduce FPSLimiter CPU usage and stabilize its frame pacing

### LINKED ISSUE #5163: Setting a FPS limit could reduce your overall FPS
### Describe the bug

I am not sure why, but sometiems capping your fps with setFPSLimit() can reduce your overall performance drastically making you unable to even hitting that limit anymore.

For example, let's say I am playing at 240FPS but I want to play at 180FPS because it has less framerate related bugs, so I limit my FPS to 180 and suddenly I am not getting 240 nor 180 anymore but 100FPS.

Tested reconnecting, restarting the game and the server but the issue is still present. I cannot reproduce this on an empty default server, but downgrading to v1.6-release-24133 gets rid of the issue (and gives me better average FPS too at least in the location I am testing)

_Version 1.7 r26770_
https://youtu.be/YQuMSFmUwaQ

_Version 1.6 r24133_
https://youtu.be/IQkPkXpfc4k

### Steps to reproduce

1. Join a relatively heavy gamemode with uncapped FPS
2. Cap your FPS to a high value (setFPSLimit)

### Version

Client  v1.7-untested-26770

### Additional context

Maybe #4385 related?

### Relevant log output

```shell

```

### Security Policy

TheCrazy17:
#### Summary
The busy-wait in `SetFrameRateThrottle` used to spin for a fixed 1ms safety margin around the waitable timer on every frame; that margin is now sized from the timer's actually observed wake-up jitter on the current machine, so it shrinks wherever the timer already wakes up on time. The PAUSE instructions in the final precision loop are also batched between TSC reads now, since `__rdtsc()` is serializing and reading it after every single PAUSE was pure overhead.

On top of that, the throttle call itself was moved; it used to run mid-frame, before the GUI (HUD, chat, CEF) was drawn and before the real `Present()` call, so none of that time counted toward the frame budget and just got added on top of every throttled frame. It now runs right after `Present()` returns, so pacing accounts for the whole frame. The spin also briefly raises the thread to `THREAD_PRIORITY_TIME_CRITICAL` for its own bounded duration, which cuts down on the OS preempting it mid-spin and causing the small overshoots that read as jitter around the cap.

#### Motivation
Helps with #5147; the adaptive margin cuts the wasted spinning that kept a core boosted, worst on AMD where PAUSE retires much faster than on Intel, though it won't remove every source of busy-wait CPU time.

Also fixes #5163, or so it looks like from here; setting a lower FPS limit was consistently producing fewer frames than the limit itself, e.g. 100 capped down to 80-90 and unstable, and that traced back to the throttle running before part of the frame's cost was even paid. Both 80 and 100 caps hold steady on my machine now, but I'd rather see it confirmed on other hardware before treating the issue as closed for good.

#### Test plan
Tested `setFPSLimit(100)` and `setFPSLimit(80)` in a demanding gamemode; before the change FPS ran well under the cap and bounced around, after it holds close to the target with only minor jitter.

#### Checklist

### DIFF
--- a/Client/core/CModManager.cpp
+++ b/Client/core/CModManager.cpp
@@ -103,9 +103,6 @@ void CModManager::DoPulsePostFrame()
 
     handleStateChange();  // Handle state changes after pulse
 
-    // TODO: ENSURE "CModManager::DoPulsePostFrame" IS THE LAST THING BEFORE THE FRAME ENDS
-    CCore::GetSingleton().GetFPSLimiter()->OnFrameEnd();  // Apply FPS limiting
-
     TIMING_GRAPH("-DoPulsePostFrame");
     TIMING_GRAPH("");
 }
--- a/Client/core/DXHook/CProxyDirect3DDevice9.cpp
+++ b/Client/core/DXHook/CProxyDirect3DDevice9.cpp
@@ -1073,6 +1073,12 @@ HRESULT CProxyDirect3DDevice9::Present(CONST RECT* pSourceRect, CONST RECT* pDes
     TIMING_GRAPH("Present");
     HRESULT hr = CDirect3DEvents9::PresentGuarded(m_pDevice, pSourceRect, pDestRect, hDestWindowOverride, pDirtyRegion);
     TIMING_GRAPH("PostPresent");
+
+    // Apply FPS limiting here, as the last thing before the frame ends, so pacing accounts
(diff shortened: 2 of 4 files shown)


FileEX:
Compared to my previous measurements in the #5147 , it's about 10-15°C lower in my case.

image

FileEX (approved):

PR #5176, opened 2026-08-11
Fix #5173: prevent server crash on invalid elementData argument

### LINKED ISSUE #5173: setElementData kills the server if the first argument isn't an element
### Describe the bug

as in title, passing element that doesn't exist to `setElementData` instantly kills the server, it dies with `c0000374` (heap corruption) no crash dump

on 1.6-r23952 the same call is harmless, it logs `Bad argument @ 'setElementData' [Expected element at argument 1, got boolean]` and the server keeps running

### Steps to reproduce

1. start runcode
2. run setElementData(false, "test", true)

### Version

master

### Relevant log output

```shell
Critical error detected c0000374

ntdll!RtlFreeHeap+0x51
deathmatch!_free_base+0x1c
deathmatch!std::basic_string<char,...>::_Tidy_deallocate+0x4c
deathmatch!__FrameHandler4::FrameUnwindToState+0x34c
deathmatch!__FrameHandler4::FrameUnwindToEmptyState+0x2f
deathmatch!__CxxFrameHandler4+0xa5
ntdll!RtlUnwindEx+0x339
ntdll!RtlUnwind+0xcd
lua5_1!__longjmp_internal+0xe9
lua5_1!luaD_throw+0x30
lua5_1!lua_error+0x9
lua5_1!luaL_error+0x4c
deathmatch!CLuaElementDefs::setElementData+0x648
deathmatch!CLuaMain::PCall+0x6f
deathmatch!CLuaTimer::ExecuteTimer+0xbe
```

### Security Policy

MohabCodeX:
## Summary

Refactor `CScriptArgReader` error handling to use `LuaFunctionError` exceptions instead of calling `luaL_error` directly.

## Motivation

Fixes #5173.

Calling `luaL_error` while C++ objects are still on the stack can skip their destructors because Lua uses `longjmp` for error handling. This can cause memory leaks, heap corruption, or crashes.

The idea here is to let C++ handle the stack unwinding first, then report the error to Lua.

In short:

1. `CScriptArgReader::ThrowError()` throws `LuaFunctionError`.
2. C++ cleans up local objects normally.
3. `CLuaFunctionParser` catches the exception and calls `luaL_error` once the stack is clean.

## Test Plan
[arg_reader_suite.zip](https://github.com/user-attachments/files/30959960/arg_reader_suite.zip)


FileEX:
This doesn't address the root cause of the problem - it's just a workaround, and an incomplete one at that. `luaL_error` is also used in many other files, so calling it in any other way will cause the same crash. Even the argument parser uses `luaL_error`.

MohabCodeX:
> This doesn't address the root cause of the problem - it's just a workaround, and an incomplete one at that. `luaL_error` is also used in many other files, so calling it in any other way will cause the same crash. Even the argument parser uses `luaL_error`.

If I add a safe helper method inside `CScriptArgReader`:
```cpp
int ThrowError()
{
    return luaL_error(m_luaVM, "%s", GetFullErrorMessage().c_str());
}
```
And replace the raw luaL_error argument error calls with argStream.ThrowError(), would that be the right direction to fix this systemically?

FileEX:
The key point isn't *where* `luaL_error()` is located, but whether there are still C++ objects on the stack that require destruction when it is called. Until the issue with MTA closing when debugging is fixed, I can't help with it

MohabCodeX:
> he key point isn't _where_ `luaL_error()` is located, but whether there are still C++ objects on the stack that require destruction when it is called. Until the issue with MTA closing when debugging is fixed, I can't help with it

Makes sense, thanks 🤍
I updated `CScriptArgReader::ThrowError(`) to throw `LuaFunctionError` instead, so the stack unwinds before luaL_error is called at the outer parser boundary.

FileEX:
Inside `ThrowError`, calling `GetFullErrorMessage` creates a temporary `SString` object on the heap, so this doesn't actually solve the problem properly (even if it doesn't crash in this particular case).

MohabCodeX:
I see what you mean, but since we're throwing a standard C++ exception here instead of calling `luaL_error` directly, exception unwinding actually guarantees that the temporary `SString`'s destructor is called right away before the stack leaves the frame. So unlike `luaL_error`'s longjmp, it won't leak or bypass any destructors.

That said, if you'd still prefer avoiding that temporary `SString` allocation inside `ThrowError()`, let me know, though I believe it isn't strictly necessary as the core issue is fully resolved.

FileEX:
FunctionParser still calls luaL_error, so the root cause isn't fully addressed. It looks like we're merely postponing the longjmp.

MohabCodeX:
Calling `luaL_error` at the `FunctionParser` boundary is intentional since the stack is already unwound by then. 

To address this systemically, I've updated all 86 `luaL_error` call sites across 10 `luadefs` files to use `argStream.ThrowError()` so every argument reader unwinds safely now.

PR #5178, opened 2026-08-14
Fix dark mode in 1.6

SpeedyFolf:
#### Summary

Porting from 1.7 (master). Parse AppsUseLightTheme registry and fix typos. Read AppsUseLightTheme with a status out param, robustly parse it as a numeric value (strtol) and set darkTitleBar when the value is 0 — replacing the previous incorrect string comparison. Also correct several instances of a spelling error: 'paramters' -> 'parameters' in debug messages and comments.

#### Motivation

Dark mode on the window titlebar formerly worked in 1.6 but stopped in or around [r24000](https://buildinfo.multitheftauto.com/index.php?Revision=24000&Branch=1.7), this fixes it.

#### Checklist

### DIFF
--- a/Client/core/DXHook/CProxyDirect3D9.cpp
+++ b/Client/core/DXHook/CProxyDirect3D9.cpp
@@ -331,8 +331,20 @@ HRESULT CProxyDirect3D9::CreateDevice(UINT Adapter, D3DDEVTYPE DeviceType, HWND 
 #endif
 
     // Set dark titlebar if needed
-    BOOL darkTitleBar =
-        GetSystemRegistryValue((uint)HKEY_CURRENT_USER, "Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", "AppsUseLightTheme") == "\x0";
+    int           themeStatus = 0;
+    const SString appsUseLightTheme =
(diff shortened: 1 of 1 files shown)


FileEX:
> this should fix it

So you didn't test this? Judging by the check results, I assume not.

SpeedyFolf:
> So you didn't test this? Judging by the check results, I assume not.

Can you still build 1.6 just fine with Visual Studio 2026 instead of 2022? My 2022 straight up broke somehow and I've hopefully fully uninstalled it. I honestly haven't tested it, but it's such a small change and like I said it's the same thing 1.7 uses (this file will now match in 1.6 and 1.7). I think the check results are normal for 1.6, it complained about clang-format even after I used it and it's complaining about Visual Studio 2022 vs. 2026. When I swapped the Hotring Racer 2 and 3 names I didn't test that either, but again it's such a small change.

SpeedyFolf:
Apparently you can build 1.6 with Visual Studio 2026 as long as you install VS 2022 v143 build tools component and MFC for v143 build tools component (or change afxres.h includes to winres.h), so yes I have tested this now and it works, it's dark when using dark theme and light when using light theme, all is well again.

Lpsd (approved):
Thanks!

PR #5179, opened 2026-08-14
Add onPedHit event

FileEX:
#### Summary
This PR adds a new server-side `onPedHit` event, which is triggered when a ped/player is hit by another player/ped with a melee weapon, a physical weapon attack (such as hitting with a rifle butt), or a vehicle.

#### Motivation
The main purpose of this event is to make it easier to detect god mode cheats. For firearms, we can use `onPlayerWeaponFire` on the server side to check whether the player actually took damage. For projectiles, we can use `onPlayerProjectileCreation`, and for the fire extinguisher and spray can, we have `isPedChoking`. However, there is currently no way to check whether a player was hit by a melee attack.

With god mode cheats, the `onPedDamage` event isn't triggered because the server keeps receiving the same HP value from the client. This event will allow us to check whether the player actually took damage after being hit.

I know it's still not perfect and can be bypassed, but if an `onElementHealthChange` event is added someday, there will be no way to bypass it, because even clever tricks like changing the HP in a timer would still be easily detectable on the server side.

#### Test plan
```lua
addEventHandler('onPedHit', root, function(attacker, weapon, bodypart, loss)
    print(source) -- victim
    print(attacker, weapon, bodypart, loss)
end)
```

#### Checklist


PlatinMTA:
a bit offtopic but yes ``onElementHealthChange`` would be really useful, not even to detect cheats but just for general purposes like disabling a "grey screen" shader when a player recovers its health. right now you basically need to know where the players can heal up and hook it up or just have a timer running to catch the changes of health. well there are other methods but its just inconvenient.

PR #5181, opened 2026-08-15
Fix aircraft not working on custom vehicle models

TheCrazy17:
#### Summary

Planes, helicopters and Vortex cloned with `engineRequestModel` now behave like the model they were cloned from. Full breakdown by vehicle below.

| Model | Fixes |
|---|---|
| Hydra | Guns, homing missiles, gunsight, missile lock-on reticle & camera, VTOL nozzle turn + thrust vectoring (plane and hover mode), ground wheel steering exemption, door, jet engine sound, landing gear, adjustable nozzle-angle readout |
| Hunter | Minigun, missile lock, gunsight, door |
| Rustler | Guns, door, landing gear |
| Cropduster | Door, smoke trail |
| Shamal | Door, jet engine sound, landing gear |
| Nevada | Door, landing gear |
| Stunt Plane | Door / rear wheel visibility, smoke trail toggle key, smoke particle spawn |
| AT-400 | Landing gear, boarding/exit stair alignment, door collision model, exit door-height exemption, G-key closest-car targeting distance, jumbo shadow, jet engine sound |
| Andromada | Loading ramp physics, door, jet engine sound, jumbo shadow, cargo door creak sound, landing gear |
| Vortex | Traction/hover feel, suspension cushion, ground-rest exemption, steering, camera distance, thrust and turn damping, flyer-collision exemption, car-style crash damage, fan FX, hovercraft audio flag, engine sound, physics keep-alive, door panel |
| RC Baron | Flight (thrust/control response), camera, engine sound |
| Sea Sparrow | Water landing physics, buoyancy, mounted gun spread |
| Leviathan | Buoyancy constants, physics keep-alive |
| Skimmer | Boat physics (capsize recovery, water-landing damping, aquaplaning, steering, water drag), buoyancy constants, wheel status, splash FX, wake trail, propeller-spin render, boat camera (mode selection + framing), first-person culling |
| Police Maverick | Native searchlight (horn key) |
| RC Bandit, RC Raider, RC Goblin | Engine sound |
| RC Tiger | Engine sound, mounted gun spread |
| Sparrow | Mounted-gun-spread resolver now recognises it (shares code with Sea Sparrow) |

Stock vehicles are untouched; every hook resolves a custom model ID back to the one it was cloned from and falls through unchanged otherwise.

#### Motivation

Part of #1861. This started much smaller: get the Hunter's minigun and the Hydra's missiles firing on a cloned model. Testing that surfaced a much bigger problem; nothing about a cloned aircraft matches native code, since every system that decides what a vehicle is (the flight model, the weapon code, the audio entity, the constructors, the cameras) does it with a hardcoded model ID comparison, and a clone made with `engineRequestModel` carries an ID of its own. Fixing only the guns would have shipped a plane that could shoot but couldn't take off right, had no engine sound, and used the wrong camera.

It wasn't the plan to turn this into a PR this size. Splitting it up was considered, but a lot of these fixes share a resolved model ID across several native compares in the same function, a seven-way door switch collapsing into one resolve, for example; doing that piecemeal across separate PRs would have meant reopening already-merged code and reverifying register liveness a second time on the same instructions, which is exactly the kind of change that's easy to get subtly wrong in inline assembly. So the PR grew to cover the whole class of bug in one pass instead, merging every hook that could be merged to keep the total line count as small as the scope allows. It ended up being several days of work, most of it disassembly verification rather than writing code.

One thing worth calling out: the Vortex is implemented natively as a `CPlane`, not a boat class, despite behaving like a hovercraft in gameplay. That's why several of its fixes live in the same plane-specific functions as the Hydra's, and why a few hooks below end up fixing the Vortex and the Skimmer together off the same shared load even though they're unrelated vehicle types on paper.

#### Test plan

Tested with the attached Lua script (spawns a clone of each vehicle behind a chat command, e.g. `/sethydra`, `/setvortex`) against a stock instance of the same model, side by side, for every vehicle listed in the table above. Confirmed stock vehicles behave exactly as before the change.

[aircraft-test.zip](https://github.com/user-attachments/files/31094182/aircraft-test.zip)

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/CClientVehicle.cpp
+++ b/Client/mods/deathmatch/logic/CClientVehicle.cpp
@@ -1357,6 +1357,21 @@ void CClientVehicle::SetLandingGearDown(bool bLandingGearDown)
     }
 }
 
+// The checks below compare against standard model IDs, so a custom model has to be read as the
+// model it was cloned from. The model info is looked up fresh rather than through m_pModelInfo,
+// since SetModelBlocking calls in here before that pointer has been updated.
+static VehicleType GetVehicleTypeForModel(unsigned short usModel)
+{
+    std::uint32_t ulModel = usModel;
+    if (ulModel < 400 || ulModel > 611)
+    {
+        if (CModelInfo* pModelInfo = g_pGame->GetModelInfo(ulModel))
+            ulModel = pModelInfo->GetParentID();
+    }
+
+    return static_cast<VehicleType>(ulModel);
+}
+
 unsigned short CClientVehicle::GetAdjustablePropertyValue()
 {
     unsigned short usPropertyValue;
@@ -1364,7 +1379,7 @@ unsigned short CClientVehicle::GetAdjustablePropertyValue()
     {
         usPropertyValue = m_pVehicle->GetAdjustablePropertyValue();
         // If it's a Hydra invert it with 5000 (as 0 is "forward"), so we can maintain a standard of 0 being "normal"
-        if (static_cast<VehicleType>(m_usModel) == VehicleType::VT_HYDRA)
+        if (GetVehicleTypeForModel(m_usModel) == VehicleType::VT_HYDRA)
             usPropertyValue = 5000 - usPropertyValue;
     }
     else
@@ -1378,7 +1393,7 @@ unsigned short CClientVehicle::GetAdjustablePropertyValue()
 
 void CClientVehicle::SetAdjustablePropertyValue(unsigned short usValue)
 {
-    if (static_cast<VehicleType>(m_usModel) == VehicleType::VT_HYDRA)
+    if (GetVehicleTypeForModel(m_usModel) == VehicleType::VT_HYDRA)
         usValue = 5000 - usValue;
 
     _SetAdjustablePropertyValue(usValue);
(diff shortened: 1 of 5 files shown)


Client/mods/deathmatch/logic/CClientVehicle.cpp
```diff
@@ -1357,14 +1357,29 @@ void CClientVehicle::SetLandingGearDown(bool bLandingGearDown)
     }
 }
 
+// The checks below compare against standard model IDs, so a custom model has to be read as the
+// model it was cloned from. The model info is looked up fresh rather than through m_pModelInfo,
+// since SetModelBlocking calls in here before that pointer has been updated.
+static VehicleType GetVehicleTypeForModel(unsigned short usModel)
+{
+    std::uint32_t ulModel = usModel;
```
FileEX: `std::uint16_t`

Client/multiplayer_sa/CMultiplayerSA_Vehicles.cpp
```diff
@@ -910,6 +910,2643 @@ static void __declspec(naked) HOOK_CMonsterTruck__PreRender_DumperSwing()
...
+// Raw model-index comparisons across the flight model, weapon code, audio and cameras don't
+// recognise a clone's own id. The three resolvers below map it back to the base aircraft:
+// ResolveAircraftWeaponModelId for individual armed aircraft, ResolveVehicleParentModelId for
+// whole vehicle classes, and ResolveJetAudioModelId for the jet engine sound.
+//////////////////////////////////////////////////////////////////////////////////////////
+static std::uint32_t __fastcall ResolveAircraftWeaponModelId(std::uint32_t modelId)
+{
+    // Standard models leave m_dwParentID uninitialised, so answer the vanilla range here instead of reading it.
+    if (modelId >= 400 && modelId <= 611)
+        return modelId;
```
FileEX, line 925: modelIndex is generally 2 bytes (`uint16_t`). Are you sure all these hooks only compare the ID and don't assign it to a 2-byte value? I want to make sure there aren't any cases where a `uint32_t` gets assigned to a `uint16_t` value.

TheCrazy17: I looked into it and it seems like you were right. It was using movsx instead of movzx when loading the model index. All the hooks had the same issue before passing it to the resolvers. Fixed.

FileEX (approved):

PR #5192, opened 2026-08-16
Fix #1073: toJSON incorrectly wraps single table in array

### LINKED ISSUE #1073: toJSON incorrectly stringifies object
**Describe the bug**

When serializing a standalone object, the returned JSON puts it into an array. This is incorrect because the object is not in an array before serialization. An argument for correct behavior should solve the problem.

**To reproduce**

```srun outputChatBox(toJSON ( { "dogs", cat = "hungry", mouse = "food", birds = 4 } ))```

**Expected behaviour**

```Command results: { "1": "dogs", "mouse": "food", "cat": "hungry", "birds": 4 } [string]```

**Actual behaviour**

```Command results: [ { "1": "dogs", "mouse": "food", "cat": "hungry", "birds": 4 } ] [string]```

This should only be triggered by:
 ```toJSON ( { { "dogs", cat = "hungry", mouse = "food", birds = 4 } } )``` 

**Version**

MTA:SA Server v1.5.6-release-16345
Multi Theft Auto v1.5.6-release-18706

### PR DESCRIPTION
**Pull request by 0xSenior**
Problem:
- toJSON({ key = value }) returned '[{...}]' instead of '{...}'
- WriteToJSONString() always called WriteToJSONArray(), wrapping all results in a JSON array even for single table arguments

Solution:
- Check if there is exactly one argument and it is a table (LUA_TTABLE)
- If so, serialize it directly using WriteToJSONObject() instead of wrapping it in an array via WriteToJSONArray()
- Otherwise, fall back to existing array behavior

Files changed:
- Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
- Server/mods/deathmatch/logic/lua/CLuaArguments.cpp

This fix maintains backward compatibility for multiple arguments and non-table single arguments.

#### Checklist

* [ ] Your code should follow the [coding guidelines](https://wiki.multitheftauto.com/index.php?title=Coding_guidelines).
* [ ] Smaller pull requests are easier to review. If your pull request is beefy, your pull request should be reviewable commit-by-commit.

### DIFF
--- a/Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
+++ b/Client/mods/deathmatch/logic/lua/CLuaArguments.cpp
@@ -536,11 +536,23 @@ bool CLuaArguments::WriteToBitStream(NetBitStreamInterface& bitStream, CFastHash
 
 bool CLuaArguments::WriteToJSONString(std::string& strJSON, bool bSerialize, int flags)
 {
-    json_object* my_array = WriteToJSONArray(bSerialize);
-    if (my_array)
+    json_object* my_object = nullptr;
+
+    // If there is exactly one argument and it is a table, serialize it directly as a JSON object
+    // This fixes toJSON({ key = value }) returning "[{...}]" instead of "{...}"
+    if (m_Arguments.size() == 1 && m_Arguments[0]->GetType() == LUA_TTABLE)
     {
-        strJSON = json_object_to_json_string_ext(my_array, flags);
-        json_object_put(my_array);  // dereference - causes a crash, is actually commented out in the example too
+        my_object = m_Arguments[0]->WriteToJSONObject(bSerialize);
+    }
+    else
+    {
+        my_object = WriteToJSONArray(bSerialize);
+    }
+
+    if (my_object)
+    {
+        strJSON = json_object_to_json_string_ext(my_object, flags);
+        json_object_put(my_object);
         return true;
     }
     return false;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by FileEX** [comment]
Please provide a test plan for both before and after the changes.

**Comment by 0xSenior** [comment]
Here is the test plan and output comparison before and after the changes:

### Test Code (Lua)
```lua
-- 1. Single table with key-value pairs (Object)
local obj = { name = "player", score = 100 }
outputDebugString(toJSON(obj))

-- 2. Single array-like table
local arr = { "item1", "item2" }
outputDebugString(toJSON(arr))

-- 3. Multiple arguments
outputDebugString(toJSON("hello", 123))
```

### Before Changes
```json
[{"name":"player","score":100}]
["item1","item2"]
["hello",123]
```

### After Changes
```json
{"name":"player","score":100}
["item1","item2"]
["hello",123]
```

FileEX:
This change will probably break scripts that use a single table, right? This includes cases where someone uses the `string.sub(data, 2, -2)` mentioned on the wiki.

ffsPLASMA:
IMO this PR is unnecessary, only there for the sake of fixing some weird quirk established a decade ago. As you said, the wiki explicitly mentions the string subtraction to "fix" this quirk.

This will break many decade old scripts at this point, I would therefore just leave the quirk as it is.

0xSenior:
You're completely right. I hadn't taken into my mind how much existing code relies on this legacy behavior and the `string.sub` workaround mentioned in the wiki.

Preserving backwards compatibility is definitely more important than fixing this quirk. Closing this PR. Thanks for the feedback!

PR #5199, opened 2026-08-17
Fix #2939: Make ped pushing vehicle physics FPS independent

### LINKED ISSUE #2939: At high FPS it's easier to move cars around by running into them
### Describe the bug

At high FPS it's easier to move cars around by running into them.

### Steps to reproduce

```lua
run setFPSLimit(30) me.position = Vector3(0, 0, 3.5) Vehicle(411, 0, 3, 4)
```

Try to move car by running into it.

```lua
run setFPSLimit(100)
```

Try to move car by running into it.

### Version
N/A

### Additional context

Parent issue:
* #929

### Relevant log output

_No response_

MohabCodeX:
Closes #2939

### Summary of Changes
Running into unoccupied vehicles at high framerates causes them to accelerate and slide unnaturally fast due to frequent collision impulses overpowering tire friction.

This PR governs pedestrian push impulses by:
- Capping push velocity using momentum conservation based on ped and vehicle mass.
- Scaling collision impulses by framerate timestep to maintain consistent vehicle weight across all FPS levels.
- Preserving vertical and roll/pitch stability to prevent vehicles from tilting or flipping during contact.

### Testing
https://streamable.com/3puhud

### DIFF
--- a/Client/mods/deathmatch/logic/CClientGame.cpp
+++ b/Client/mods/deathmatch/logic/CClientGame.cpp
@@ -4707,7 +4707,13 @@ bool CClientGame::VehicleCollisionHandler(CVehicleSAInterface*& pCollidingVehicl
 
             pVehicleClientEntity->CallEvent("onClientVehicleCollision", Arguments, true);
 
-            // Update the colliding vehicle, because it might have been invalidated in onClientVehicleCollision (e.g. fixVehicle)
+            // Update the colliding vehicle, because it might have been invalidated in onClientVehicleCollision (e.g. fixVehicle, destroyElement)
+            if (pVehicleClientEntity->IsBeingDeleted() || !pVehicleClientEntity->GetGameEntity())
+            {
+                pCollidingVehicle = nullptr;
+                return true;
+            }
+
             pCollidingVehicle = reinterpret_cast<CVehicleSAInterface*>(pVehicleClientEntity->GetGameEntity()->GetInterface());
 
             // Allocate a BitStream
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -46,6 +46,8 @@ extern CGame* pGameInterface;
 #define HOOKPOS_EndWorldColors                           0x561795
 #define HOOKPOS_CWorld_ProcessVerticalLineSectorList     0x563357
 #define HOOKPOS_ComputeDamageResponse_StartChoking       0x4C05B9
+#define HOOKPOS_ComputeDamageResponse_StealthKillTarget  0x4C035C
+#define FUNC_CTaskManager_GetActiveTask                  0x681720
 #define HOOKPOS_CAutomobile__ProcessSwingingDoor         0x6A9DAF
 
 #define FUNC_CStreaming_Update                     0x40E670
@@ -447,6 +449,7 @@ void HOOK_CObject_Render();
 void HOOK_EndWorldColors();
 void HOOK_CWorld_ProcessVerticalLineSectorList();
 void HOOK_ComputeDamageResponse_StartChoking();
+void HOOK_ComputeDamageResponse_StealthKillTarget();
 void HOOK_CollisionStreamRead();
 void HOOK_CPhysical_ApplyGravity();
 void HOOK_VehicleCamStart();
@@ -652,6 +655,7 @@ void CMultiplayerSA::InitHooks()
     HookInstall(HOOKPOS_EndWorldColors, (DWORD)HOOK_EndWorldColors, 5);
     HookInstall(HOOKPOS_CWorld_ProcessVerticalLineSectorList, (DWORD)HOOK_CWorld_ProcessVerticalLineSectorList, 8);
     HookInstall(HOOKPOS_ComputeDamageResponse_StartChoking, (DWORD)HOOK_ComputeDamageResponse_StartChoking, 7);
+    HookInstall(HOOKPOS_ComputeDamageResponse_StealthKillTarget, (DWORD)HOOK_ComputeDamageResponse_StealthKillTarget, 9);
     HookInstall(HOOKPOS_CollisionStreamRead, (DWORD)HOOK_CollisionStreamRead, 6);
     HookInstall(HOOKPOS_VehicleCamStart, (DWORD)HOOK_VehicleCamStart, 6);
     HookInstall(HOOKPOS_VehicleCamTargetZTweak, (DWORD)HOOK_VehicleCamTargetZTweak, 8);
@@ -4193,6 +4197,47 @@ static void __declspec(naked) HOOK_ComputeDamageResponse_StartChoking()
         mov     ecx, [edi]
         mov     eax, [ecx+0x47C]
         jmp     dwChokingChoke
+    }
+    // clang-format on
+}
+
+// CEventHandler::ComputeDamageResponse turns any damage into a stealth kill death while the
+// attacker still runs a CTaskSimpleStealthKill, without checking who that task is for; a knife
+// kill keeps it active for the whole animation, so a tear gas tick or a stray bullet from the
+// same attacker makes a nearby ped play the stealth death too.
+// The original code compares the task's type further down, but on any other task type this reads
+// a field that can't match the ped, so the outcome doesn't change
+bool _cdecl IsStealthKillTaskTarget(CTaskSimpleStealthKillSAInterface* pStealthKillTask, CPed* pPed)
+{
+    return pStealthKillTask && pStealthKillTask->m_target == pPed;
+}
+
+static constexpr std::uintptr_t RETURN_ComputeDamageResponse_StealthKillTarget = 0x4C0365;
+static constexpr std::uintptr_t RETURN_ComputeDamageResponse_StealthKillSkip = 0x4C03D6;
+static void __declspec(naked)   HOOK_ComputeDamageResponse_StealthKillTarget()
+{
+    MTA_VERIFY_HOOK_LOCAL_SIZE;
+
+    // clang-format off
+    __asm
+    {
+        // Replaced code: ecx already points at the attacker's task manager and eax gets his active
+        // task. Past the hook nothing reads eax, ecx or edx before writing them, on either path.
+        mov     edx, FUNC_CTaskManager_GetActiveTask
+        call    edx
+
+        mov     ecx, [edi]
+        push    ecx
+        push    eax
+        call    IsStealthKillTaskTarget
+        add     esp, 8
+        test    al, al
+        jz      notTheTarget
+
+        jmp     RETURN_ComputeDamageResponse_StealthKillTarget
+
+        notTheTarget:
+        jmp     RETURN_ComputeDamageResponse_StealthKillSkip
     }
     // clang-format on
 }
--- a/Client/multiplayer_sa/CMultiplayerSA.cpp
+++ b/Client/multiplayer_sa/CMultiplayerSA.cpp
@@ -1221,8 +1221,8 @@ void CMultiplayerSA::InitHooks()
     // Stop CPlayerPed::ProcessControl from calling CVisibilityPlugins::SetClumpAlpha
     MemSet((void*)0x5E8E84, 0x90, 5);
 
-    // Stop CVehicle::UpdateClumpAlpha from calling CVisibilityPlugins::SetClumpAlpha
-    MemSet((void*)0x6D29CB, 0x90, 5);
+    // Disable CVehicle::UpdateClumpAlpha completely (fades singleplayer traffic, crashes at 0x732B2A when m_pRwObject is nullptr)
+    MemPut<BYTE>(0x6D2980, 0xC3);
 
     // Disable CVehicle::DoDriveByShootings
     MemSet((void*)0x741FD0, 0x90, 3);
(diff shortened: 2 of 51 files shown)


FileEX:
I feel like it's significantly harder to push the vehicle at 100 FPS. Why does it behave like that?

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -853,7 +941,11 @@ void CMultiplayerSA::InitHooks_FrameRateFixes()
     // CCam::Process_FollowCar_SA
     MemSet((void*)0x524FD7, 0x90, 0x1B);
 
-    // Fixes slow boat movement on high FPS.
```
FileEX, line 856: Why did you remove this comment?

FileEX (changes requested):

MohabCodeX:
Scaling the impulse by `timeStepRatio` makes the collision force per second mathematically equal across framerates, but GTA's physics engine still applies tire friction and ground damping on every single frame. At 100 FPS, that friction runs ~3.3x more often than at 30 FPS, which absorbs the smaller per-frame impulses faster before the vehicle can build up momentum (and the per-frame penetration depth that shifts the car is also naturally smaller).

We could tweak it with something like `std::sqrt(timeStepRatio)` in `HOOK_CPhysical__ApplyCollision` to compensate for the extra friction steps and make pushing feel noticeably lighter at high FPS. I attached a video demonstrating this approach in action for comparison. You might want to test it yourself too.. personally, I wasn't entirely convinced by it compared to the clean linear ratio since it introduces an arbitrary curve into the physics, but it does make pushing easier while still preventing the launch bug. Another route could also be looking into a dedicated custom ped-push velocity system when colliding with vehicles.

https://github.com/user-attachments/assets/3a472d87-447f-43de-ac7a-bcf4de2f322e

FileEX:
It looks much better now

FileEX:
Tested and it generally works, but it introduces a noticeable side effect, even at normal FPS.

With your PR, the vehicle behaves strangely - it seems to bounce and lock its wheels, making it harder to move.

74 FPS without your PR (smooth)
https://files.catbox.moe/8m1zau.mp4

74 FPS with your PR (rough, vehicle bounces)
https://files.catbox.moe/qa060h.mp4

MohabCodeX:
**Quick update:** After further investigation into the vehicle physics, I found that the previous implementation simply scaled raw collision impulses after impact, which didn't account for directional force distribution or tire physics. The new approach isolates horizontal rolling velocity from the vertical and tilt axes, governs forward acceleration and top speed using momentum conservation, and dynamically scales the vehicle sleep threshold with delta time so suspension settling stays consistent across framerates. I'll attach a video later

Vehicle push behavior is now standardized to the 30 FPS baseline across all framerates. GTA:SA physics were tuned for 30 FPS,..  this prevents high-FPS collision impulses from breaking tire friction and causing vehicles to glide.

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -713,6 +715,65 @@ static void __declspec(naked) HOOK_CTaskSimpleSwim__ProcessEffectsBubbleFix()
...
+        fstp    dword ptr [esp + 32]
+        fmul    dword ptr [esp + 24]
+
+        fdiv    ds:[0xB7CB5C]
+        fmul    kOriginalTimeStep
+
+        jmp     RETURN_CTaskSimpleSwim__ProcessSwimmingResistance
+    }
+    // clang-format on
+}
```
FileEX, line 776: Unrelated changes

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -713,6 +715,65 @@ static void __declspec(naked) HOOK_CTaskSimpleSwim__ProcessEffectsBubbleFix()
...
+    // clang-format off
+    __asm
+    {
+        fmul    ds:[0x871DDC]           // Original constant used in code
+        fmul    ds:[0xB7CB5C]           // Multiply by current timestep
+        fdiv    kOriginalTimeStep       // Divide by desired 30 FPS timestep
+        jmp     RETURN_CVehicle__ApplyBoatWaterResistance
+    }
+    // clang-format on
+}
```
FileEX, line 736: Unrelated changes

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
MohabCodeX: Sorry, that was an accidental leftover

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -9,14 +9,15 @@
...
+#include <algorithm>
+#include <cmath>
 
 static bool         bWouldBeNewFrame = false;
 static unsigned int nLastFrameTime = 0;
 
 constexpr float kOriginalTimeStep = 50.0f / 30.0f;
 
 // Fixes player movement issue while aiming and walking on high FPS.
-// Only rescales the compare threshold; m_MoveCmd's reset is NOPed separately in InitHooks_FrameRateFixes.
```
FileEX, line 19: ?

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
```diff
@@ -909,10 +1217,6 @@ void CMultiplayerSA::InitHooks_FrameRateFixes()
     EZHookInstall(CFallingGlassPane__Update_B);
     EZHookInstall(CFallingGlassPane__Update_C);
 
-    // Fixes camera jitter while aiming and walking at high FPS.
-    // CTaskSimpleUseGun::SetMoveAnim
-    MemSet((void*)0x61E5E4, 0x90, 0x6);
```
FileEX, line 1034: ??

FileEX (changes requested):

Client/multiplayer_sa/CMultiplayerSA_FrameRateFixes.cpp
MohabCodeX: Accidental deletion during rebase .. restored now

FileEX:
Please resolve conflicts

MohabCodeX:
While testing across framerates, noticed vehicles spawned at high FPS (100–240) were freezing with stretched suspension because the game was putting them to sleep on frame 1 before gravity could settle the chassis onto the springs.

Disabled the instant-sleep bypass and stopped stationary damping from wiping small vertical gravity steps. Cars and bikes now settle naturally to their proper resting ride height across all framerates before going to sleep.

Full Testing : https://streamable.com/3puhud

Client/mods/deathmatch/logic/CClientGame.cpp
```diff
@@ -4707,7 +4707,13 @@ bool CClientGame::VehicleCollisionHandler(CVehicleSAInterface*& pCollidingVehicl
...
             pVehicleClientEntity->CallEvent("onClientVehicleCollision", Arguments, true);
 
-            // Update the colliding vehicle, because it might have been invalidated in onClientVehicleCollision (e.g. fixVehicle)
+            // Update the colliding vehicle, because it might have been invalidated in onClientVehicleCollision (e.g. fixVehicle, destroyElement)
+            if (pVehicleClientEntity->IsBeingDeleted() || !pVehicleClientEntity->GetGameEntity())
+            {
+                pCollidingVehicle = nullptr;
+                return true;
+            }
+
```
FileEX: Unrelated changes

Client/multiplayer_sa/CMultiplayerSA.cpp
```diff
@@ -1225,8 +1225,8 @@ void CMultiplayerSA::InitHooks()
     // Stop CPlayerPed::ProcessControl from calling CVisibilityPlugins::SetClumpAlpha
     MemSet((void*)0x5E8E84, 0x90, 5);
 
-    // Stop CVehicle::UpdateClumpAlpha from calling CVisibilityPlugins::SetClumpAlpha
-    MemSet((void*)0x6D29CB, 0x90, 5);
+    // Disable CVehicle::UpdateClumpAlpha completely (fades singleplayer traffic, crashes at 0x732B2A when m_pRwObject is nullptr)
+    MemPut<BYTE>(0x6D2980, 0xC3);
```
FileEX: Unrelated changes

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -4092,6 +4173,71 @@ static int _cdecl CFileLoader_LoadVehicleObject_sscanf(const char* s, const char
...
+        mov     eax, [esp+8]
+        mov     edx, ds:[0x8D6094]
+        mov     [edx+ecx+4], eax
+
+    null_clump:
+        retn
+    }
+    // clang-format on
+}
+
```
FileEX: Unrelated changes

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -4116,8 +4262,10 @@ void CMultiplayerSA::InitHooks_CrashFixHacks()
     EZHookInstall(CrashFix_Misc16);
     // EZHookInstall ( CrashFix_Misc17 );
     EZHookInstall(CrashFix_Misc18);
-    // EZHookInstall ( CrashFix_Misc19 );
     EZHookInstall(CrashFix_Misc20);
+    EZHookInstall(CrashFix_CMatrix__UpdateRwMatrix);
+    EZHookInstall(CrashFix_AllocateMatrix_Init1);
+    EZHookInstall(CrashFix_AllocateMatrix_Init2);
```
FileEX: Unrelated changes

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -4180,6 +4328,8 @@ void CMultiplayerSA::InitHooks_CrashFixHacks()
     EZHookInstallChecked(CStreaming__GetNextFileOnCd_NullTxdDef);
     EZHookInstallChecked(CStreaming__ConvertBufferToObject_NullTxdDef);
     EZHookInstallChecked(CEventScanner__ScanForEvents_ContactEntity);
+    EZHookInstallChecked(CVisibilityPlugins_GetClumpAlpha);
+    EZHookInstallChecked(CVisibilityPlugins_SetClumpAlpha);
```
FileEX: Unrelated changes

FileEX (changes requested):

FileEX:
Something is still wrong. At 74 FPS, the vehicle now feels significantly heavier than it normally does.

74 FPS without your PR
https://files.catbox.moe/8ga788.mp4

74 FPS with your PR
https://files.catbox.moe/i0tc9r.mp4

MohabCodeX:
Yes, as I mentioned, I adjusted the car’s movement to match the original 30 FPS behavior. So players at 74 FPS may feel like it’s slower, but it’s actually just behaving the same as it does at 30 FPS, keeping things consistent across all frame rates.

https://streamable.com/3puhud

MohabCodeX:
@FileEX  The unrelated changes you mentioned are fixes for crashes that actually occurred because of the changes in this PR while testing, mainly vehicles crashing when repeatedly spawned or deleted during tests, and a crash if a vehicle gets destroyed during the collision event. That's why I bundled them together here, but if you still prefer to keep this PR focused on the ped push physics and move those into a separate PR, I have no problem doing that at all

FileEX:
> @FileEX The unrelated changes you mentioned are fixes for crashes that actually occurred because of the changes in this PR while testing, mainly vehicles crashing when repeatedly spawned or deleted during tests, and a crash if a vehicle gets destroyed during the collision event. That's why I bundled them together here, but if you still prefer to keep this PR focused on the ped push physics and move those into a separate PR, I have no problem doing that at all

I admit that I don't quite understand how you're running into these issues/crashes, since your PR doesn't use any of these functions, such as `AllocateMatrix`, `UpdateRwMatrix`, or `Set/GetClumpAlpha`. So maybe your code is breaking something?

If this were an issue with `Set/GetClumpAlpha` or the matrix functions, these crashes would have been happening years ago and would be much more common among players.

MohabCodeX:
Yeah, it seems I didn't explain the crash fixes clearly enough. Basically, while testing the physics at high FPS, I also fixed an issue where spawned vehicles were going to sleep instantly on frame 1 with their suspension stretched up, so I prevented that premature sleep so they could settle naturally. But keeping them awake during those first spawn frames caused GTA to run things like UpdateClumpAlpha and matrix updates before the clump was even initialized, which immediately crashed the game at 0x732B2A and 0x59AD76. So I just added those hooks as shields to prevent those crashes from happening. But if you prefer to keep this PR strictly about the ped push physics and move the spawn-sleep fix and its crash hooks to a separate PR, I can easily split them.

MohabCodeX:
Here it's : #5429

FileEX:
Well, I'm not sure whether changing the behavior of vehicles even at 60 FPS, which everyone has been used to for years, is a good idea, but okay. Worst case, if people complain about it, we can revert it or find another solution.

FileEX (approved):

PR #5205, opened 2026-08-17
Add #4498: Expose CPU info with a detectable client opt-out

### LINKED ISSUE #4498: Add a way to retrieve CPU information

Basically #2267. Preferably with the ability to reject sending this information (but enabled by default and with a reliable way for the server to check if the function is disabled).

Either for better understanding of user specs to make better quality modes, or for user fingerprinting to catch evaders and cheaters.

Add a function that retrieves information about the CPU's name, max clock speed, number of cores, number of threads and if possible L1,L2,L3 cache memory size to check for lazy spoofed CPU names.

_No response_

### Additional context

I understand the privacy concerns, and that this information could be spoofed just as easily as serials. If you want to be a complete anonymous user in MTA you should be able to, but the servers should also be able to reject your connection if that's the case (like one can do with screenshots).

We could go the extra mile and let users overshare system information (disabled by default), like amount of devices plugged in, networks, number of drives, amount of disk space, base GTA:SA path, RAM speed, audio and microphone devices, etc. Of course not recommended if you want to stay anonymous, this information could help identify "secure" or "trusted" clients. The client should know that this, from a privacy standpoint, is not cool. 

**Of course that's not the main point of this issue, I just think it is silly that we don't share basic CPU information while also sharing specific GPU information without a way to disable it.**

### Security Policy

x6c85:
#### Summary
Adds client `getCPUInfo()` and server `getPlayerCPUInfo(player)` so scripts can read CPU name, max clock, cores, threads and L1/L2/L3 cache sizes.

Sharing is on by default (`allow_cpu_info`, Settings → Multiplayer → Allow CPU info). When the player opts out, hardware fields are omitted and the table only contains `AllowCPUInfo = false`, so a server resource cannot recover the data through a client event.

The client sends `PACKET_ID_PLAYER_CPUINFO` at join (appended after existing packet ids). The server stores it, fires `onPlayerCPUInfo`, and `getPlayerCPUInfo` returns `false` until that packet arrives (old clients stay `false`).

```lua
-- client
local info = getCPUInfo()
if info.AllowCPUInfo then
    outputChatBox(info.Name)
end

-- server
addEventHandler("onPlayerCPUInfo", root, function(allowed)
    if not allowed then
        kickPlayer(source, "Enable Allow CPU info")
    end
end)
```

#### Motivation
Fixes #4498. GPU details are already exposed via `dxGetStatus` with no opt-out. CPU info is useful for quality presets and for servers that want to reject clients who hide it, similar to screen upload.

#### Test plan
1. F8 `getCPUInfo()` - expect `AllowCPUInfo = true` plus name, MHz, cores, threads and cache.
2. Uncheck Allow CPU info, reconnect - client returns only `AllowCPUInfo = false`; server `getPlayerCPUInfo` is the same table, not `false`.
3. Connect an older client - `getPlayerCPUInfo` stays `false` (no packet).
4. Confirm `onPlayerCPUInfo` runs after join and can kick when `allowed == false`.
5. Cars/other settings tabs still work; Allow CPU info sits under Allow screen upload.

#### Checklist


PlatinMTA:
I know this is technically not in the scope of this PR, but if this gets merged we should consider an opt-out for sending GPU information from dxGetStatus().

FileEX:
> I know this is technically not in the scope of this PR, but if this gets merged we should consider an opt-out for sending GPU information from dxGetStatus().

https://github.com/multitheftauto/mtasa-blue/pull/4535

PR #5206, opened 2026-08-17
Fix engineRequestModel objects having no physics

### LINKED ISSUE #3074: engineRequestModel does not work for physical objects
### Describe the bug

When engineRequestModel('object') is used with a parent object with physical properties it doesn't copy its physical properties (not only `engineGetModelPhysicalPropertiesGroup` and `engineGetObjectGroupPhysicalProperty`!!), for example, when velocity is applied to object with requested model, it will fall through the ground.

Important note, when 
```lua
local parentGroup = engineGetObjectGroupPhysicalProperty(parentModel)
engineSetObjectGroupPhysicalProperty(model, parentGroup)
```
used, it won't have any physical properties at all

### Steps to reproduce

```lua
local model = engineRequestModel('object', 1252)

function velocityOnStreamIn()
    setElementVelocity(source, 0.1, 0, 0)
end

addCommandHandler('testme', function()
    local o1 = createObject(1252, x + 2, y, z)
    local o2 = createObject(model, x + 2, y + 2, z)
    addEventHandler('onClientElementStreamIn', o1, velocityOnStreamIn)
    addEventHandler('onClientElementStreamIn', o2, velocityOnStreamIn)
end)
```

### Version

Client: v1.6.0-release-21852(Windows 10, 64-bit)

### Additional context

preview
https://i.imgur.com/4xrHQuU.gif

### Relevant log output

_No response_

### Security Policy

TheCrazy17:
#### Summary
Cloned objects with `engineRequestModel('object')` now keep the parent's physics properties group instead of being forced static, and `engineGetModelPhysicalPropertiesGroup`/`engineSetModelPhysicalPropertiesGroup` no longer reject requested model ids.

#### Motivation
Fixes #3074.

#### Test plan
Created an object and a clone of it via `engineRequestModel`, gave both velocity on stream in; both now behave the same instead of the clone falling through the ground. Also confirmed `engineGetModelPhysicalPropertiesGroup`/`engineSetModelPhysicalPropertiesGroup` no longer error on a requested model id, and that the clone already reports the same group as its parent without calling the setter at all.

```lua
local parentModel = 1252
local model = engineRequestModel('object', parentModel)
--local model = engineRequestModel('object') -- Test with another object

function velocityOnStreamIn()
    setElementVelocity(source, 0.1, 0, 0)
end

addCommandHandler('testme', function()
    local parentGroup = engineGetModelPhysicalPropertiesGroup(parentModel)
    local cloneGroup = engineGetModelPhysicalPropertiesGroup(model)
    outputChatBox('parent group: ' .. tostring(parentGroup) .. ' | clone group (already inherited): ' .. tostring(cloneGroup))

    local success = engineSetModelPhysicalPropertiesGroup(model, parentGroup)
    outputChatBox('explicit set: ' .. tostring(success))

    local o1 = createObject(parentModel, x + 2, y, z)
    local o2 = createObject(model, x + 2, y + 2, z)
    addEventHandler('onClientElementStreamIn', o1, velocityOnStreamIn)
    addEventHandler('onClientElementStreamIn', o2, velocityOnStreamIn)
end)
```

#### Checklist

### DIFF
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -2014,9 +2014,9 @@ int CLuaEngineDefs::EngineGetModelPhysicalPropertiesGroup(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        if (modelId >= g_pGame->GetBaseIDforTXD())
+        if (modelId >= g_pGame->GetBaseIDforCOL())
         {
-            argStream.SetCustomError(SString("Expected model ID in range [0-%u] at argument 1", g_pGame->GetBaseIDforTXD() - 1));
+            argStream.SetCustomError(SString("Expected model ID in range [0-%u] at argument 1", g_pGame->GetBaseIDforCOL() - 1));
             return luaL_error(luaVM, argStream.GetFullErrorMessage());
         }
 
@@ -2045,9 +2045,9 @@ int CLuaEngineDefs::EngineSetModelPhysicalPropertiesGroup(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        if (modelId > g_pGame->GetBaseIDforTXD() - 1)
+        if (modelId > g_pGame->GetBaseIDforCOL() - 1)
         {
-            argStream.SetCustomError(SString("Expected model ID in range [0-%u] at argument 1", g_pGame->GetBaseIDforTXD() - 1));
+            argStream.SetCustomError(SString("Expected model ID in range [0-%u] at argument 1", g_pGame->GetBaseIDforCOL() - 1));
             return luaL_error(luaVM, argStream.GetFullErrorMessage());
         }
 
--- a/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
+++ b/Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
@@ -2080,9 +2080,9 @@ int CLuaEngineDefs::EngineRestoreModelPhysicalPropertiesGroup(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        if (modelId > g_pGame->GetBaseIDforTXD() - 1)
+        if (modelId > g_pGame->GetBaseIDforCOL() - 1)
         {
-            argStream.SetCustomError(SString("Expected model ID in range [0-%u] at argument 1", g_pGame->GetBaseIDforTXD() - 1));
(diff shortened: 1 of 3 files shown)


Client/mods/deathmatch/logic/luadefs/CLuaEngineDefs.cpp
```diff
@@ -2014,9 +2014,9 @@ int CLuaEngineDefs::EngineGetModelPhysicalPropertiesGroup(lua_State* luaVM)
 
     if (!argStream.HasErrors())
     {
-        if (modelId >= g_pGame->GetBaseIDforTXD())
+        if (modelId >= g_pGame->GetBaseIDforCOL())
```
FileEX: Why COL? CClientObjectManager::IsValidModel also uses 20000 as the base ID. Why are you changing it to 25000 here?

TheCrazy17: You were right, I got the approach wrong. I thought `engineRequestModel` could legitimately return IDs above 20000, which is why I raised the check to `GetBaseIDforCOL()`. I’ve looked into it more thoroughly and went through more of the code, and it turns out that range never actually worked for object models. I tested this by allocating models in a loop, and `GetFirstFreeModelID()` kept returning the same ID without ever marking it as used.

I fixed it at the root instead: `GetFirstFreeModelID()` is now capped at `MAX_MODEL_DFF_ID` (20000), so it can never hand out an ID from that broken range. I also reverted the bounds check in the three functions back to `GetBaseIDforTXD()`, as it was originally, keeping it consistent with `IsValidModel()`. There’s no need for the 25000 limit anymore.

I’m really sorry for the mistake, and thanks for pointing it out.

<details>
<summary>Test script</summary>

```lua
addCommandHandler("testmodels", function()
    local count = 0
    local lastId = nil
    local repeatCount = 0
    local startTick = getTickCount()

    for i = 0, 20000 do
        local id = engineRequestModel("object")
        if not id then
            break
        end

        if id == lastId then
            repeatCount = repeatCount + 1
        else
            lastId = id
            repeatCount = 1
        end

        outputDebugString(string.format("[%d] ID asignado: %d (van %d veces seguidas)", count, id, repeatCount))
        count = count + 1
    end

    outputChatBox(string.format("Se crearon %d modelos, el último ID fue %d y se repitió %d veces seguidas (%dms)",
        count, lastId or -1, repeatCount, getTickCount() - startTick))

    local parentGroup = engineGetModelPhysicalPropertiesGroup(1252)
    object = createObject(lastId, 0, 0, 3)
    local success = engineSetModelPhysicalPropertiesGroup(lastId, parentGroup)
    outputChatBox('explicit set: ' .. tostring(success))
end)


addCommandHandler("freemodels", function()
    for id = 0, 25000 do
        engineFreeModel(id)
    end
    outputChatBox("Liberados")
end)
```

</details>

FileEX (approved):

PR #4945, opened 2026-06-11
Add setVehicleNitroColor, getVehicleNitroColor and resetVehicleNitroColor

TheCrazy17:
#### Summary

Adds per-vehicle nitro flame colour customization, available on both client and server and synced to all players:

```lua
bool setVehicleNitroColor ( vehicle theVehicle, int r, int g, int b [, int a = 255 ] )
int, int, int, int getVehicleNitroColor ( vehicle theVehicle )  -- returns false if no custom colour is set
bool resetVehicleNitroColor ( vehicle theVehicle )
```

OOP: `vehicle:setNitroColor(...)`, `vehicle:getNitroColor()`, `vehicle:resetNitroColor()`.

[Video](https://youtu.be/WdlN7ABCqj4)

Technical approach:

* The "nitro" FX blueprint's colour keyframes are shared by every vehicle, so they are re-tinted just-in-time as each vehicle's particles are rendered (hook on the per-particle render info call in `FxEmitterBP_c::Render`). The original keyframes are cached and used as a brightness envelope, so the requested hue fully replaces the default cyan; vehicles without a custom colour keep the exact default look.
* While a custom colour is applied, the nitro emitters switch from additive to standard alpha blending (restored on reset), so dark colours - including black - render correctly. The blend state is shared by all nitro particles in a frame, so default-coloured vehicles share it while a custom-coloured nitro is on screen.
* Synced via a new element RPC (added at the end of the enum) and included in the vehicle entity-add data, so players who join or stream the vehicle in later see the correct colour.

#### Motivation

The nitro flame colour is hardcoded in the shared FX blueprint. The only workaround today is replacing the effect texture or using shaders, which affects every vehicle globally - there is no way to give each vehicle its own nitro colour. This is a frequently requested customization feature for tuning, race and DM servers.

#### Test plan

Tested with a resource that spawns 10 vehicles with random nitro colours (half of them also with random alpha), nitro added and activated:

```lua
-- Spawn a row of cars with random nitro colours
for i = 0, 9 do
    local veh = createVehicle(411, i * 5, 0, 4)
    addVehicleUpgrade(veh, 1010) -- nitro
    setVehicleNitroColor(veh, math.random(0, 255), math.random(0, 255), math.random(0, 255))
    setVehicleNitroActivated(veh, true)
end

-- /nitrocolor r g b [a]
addCommandHandler("nitrocolor", function(player, _, r, g, b, a)
    setVehicleNitroColor(getPedOccupiedVehicle(player), r, g, b, a)
[...]

PR #5226, opened 2026-08-18
Fix #5225: Apply ped alpha to the jetpack clump

### LINKED ISSUE #5225: Changing a ped's alpha does not change the alpha of their jetpack
### Describe the bug

When we change a ped's alpha, their jetpack remains visible.

Image

### Steps to reproduce

1. Get jetpack
2. crun me.alpha = 0

### Version

N/A

### Additional context

_No response_

### Relevant log output

```shell

```

### Security Policy

x6c85:
#### Summary
`CTaskSimpleJetPack::RenderJetPack` draws the jetpack from its own clump, so `setPedAlpha` / `me.alpha` never reached it. The existing RenderJetPack hook now copies the ped clump alpha onto the jetpack materials, and skips render, thruster FX and engine sound when the ped is fully hidden (alpha 0 or a different interior).

#### Motivation
Fixes #5225. Getting a jetpack and running `crun me.alpha = 0` hid the ped but left a floating pack.

#### Test plan
1. Give yourself a jetpack, then `crun me.alpha = 0`: ped, jetpack, thrusters and jetpack sound should all disappear.
2. `crun me.alpha = 128`: ped and jetpack should both be translucent.
3. `crun me.alpha = 255`: ped and jetpack should be fully visible again, sound restored.
4. Change interior while on a jetpack: pack should still hide/show with the ped as before.

#### Checklist


FileEX:
Tested, it works. Do you think you could also adjust the alpha of the FX effect?

ped.alpha = 1
image

x6c85:
Good catch, FX has no colour alpha, so the pack materials were translucent while the flames stayed full-bright.

Pushed a follow-up that scales thruster spawn rate by ped alpha and clears leftover particles when alpha changes (e.g. `me.alpha = 1`). Worth another look with the same test.

FileEX:
Unfortunately, the effect doesn't look right. With alpha = 10, the FX flickers strangely. I haven't analyzed the FX system in detail, but there's a PR that changes the nitro color, so perhaps there's a way to modify its alpha as well?

https://github.com/multitheftauto/mtasa-blue/pull/4945

FileEX:
Closed in favor of #5412

PR #5237, opened 2026-08-19
Add #5215: Add onClientMouseMove event for raw mouse delta

### LINKED ISSUE #5215: Add new client side event `onClientMouseMove`

This event should trigger whenever a raw mouse movement is detected regardless of the cursor state, with it's arguments being both the horizontal and vertical traveled distance in pixels or such. I believe this is doable. The reason for this is that, if you have the game camera fixed at some position (targeting no entity), and then move the mouse, `onClientCursorMove` event triggers but the arguments dont update, they only update if the cursor is showing, or when the camera is targeting an entity (player, vehicle, ...) regardless of cursor state.

This new event allows for mouse flicking detection, used in anti cheats, and can even be key to implementing first person camera mode in MTA.

-- Arguments: distanceX, distanceY : could be negative for left/down directions and positive for right/up directions.

_No response_

### Additional context

_No response_

### Security Policy

x6c85:
#### Summary
Adds a new `onClientMouseMove` event that fires raw mouse deltas (deltaX, deltaY) using Windows Raw Input (`WM_INPUT`). Unlike `onClientCursorMove` which only reports clamped screen coordinates, this event works regardless of cursor visibility or camera target.

#### Motivation
Fixes #5215. `onClientCursorMove` arguments don't update when the cursor is hidden and the camera isn't targeting an entity, because it relies on `WM_MOUSEMOVE` which gives absolute clamped coordinates. This new event uses Raw Input to capture actual mouse deltas, enabling mouse flick detection for anti-cheats and first-person camera implementations.

#### Test plan
1. `addEventHandler("onClientMouseMove", root, function(dx, dy) outputChatBox(dx.." "..dy) end)` - move mouse with cursor hidden, verify deltas are reported.
2. Show cursor with `showCursor(true)` - both `onClientCursorMove` and `onClientMouseMove` fire.
3. Fix camera at a position with no target entity - `onClientMouseMove` still reports deltas while `onClientCursorMove` coordinates stay clamped.
4. Verify no performance regression from Raw Input registration.

#### Checklist

### DIFF
--- a/Client/core/CMessageLoopHook.cpp
+++ b/Client/core/CMessageLoopHook.cpp
@@ -191,7 +191,7 @@ LRESULT CALLBACK CMessageLoopHook::ProcessMessage(HWND hwnd, UINT uMsg, WPARAM w
             }
         }
 
-        // When updating m_bFocused in CClientGame from CPacketHandler (to fix another bug ù see the note there),
+        // When updating m_bFocused in CClientGame from CPacketHandler (to fix another bug - see the note there),
         // the window might not actually have focus at that moment (even though Windows reports it as focused).
         // In this case, isMTAWindowFocused returns false even though the window has focus.
         // Therefore, we need to intercept the window return operation and manually set the focus in CClientGame.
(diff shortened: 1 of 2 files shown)


gh0st2k2:
Maybe change the event name to `onClientMouseRawMovement` because there is already an event with that name related to mouse moving/hover over GUI elements

gh0st2k2:
Is this still not reviewed?

FileEX:
> Is this still not reviewed?

Feature freeze is currently in effect, so only bug fixes are being merged. Additionally, the author has been inactive for two weeks.

Issue #688, opened 2018-11-06
To make sought-after domains to connect to servers


ghost:
**Is your feature request related to a problem? Please describe.**
I am very confused that few server owners use domains to connect to servers.
Most often, players need to enter an IP address to connect to the servers. It brings difficulties to each side. For example, a potential player somewhere found the IP address of a server. But this IP-address is not available for a long time, as the owner of the server changed the equipment. As a result, the player fails to connect, and the server loses him.
There will be no such problems with the domain, since it is enough to change the DNS record and players will start connecting to the new IP after the zone has been updated.
The reason for writing this issue is that in the MTA after adding the server to favorites, the connection to it is carried out by IP, not domain:
![2018-11-06_21-12-37](https://user-images.githubusercontent.com/37992851/48077585-762c5500-e209-11e8-93b1-6922eba06f05.png)
(I found in #advertisements only one working domain of two...)
This leads to the fact that the players who added the server to favorites will no longer be able to connect to the server after changing its IP.
Therefore, almost no one uses domains.
**Describe the solution you'd like**
After adding the server to your favorites by entering the domain, save it and each time you log on to the server, find out the IP address of the server.
![2018-11-06_21-40-26](https://user-images.githubusercontent.com/37992851/48079193-acb79f00-e20c-11e8-9c8b-a723b77608e7.png)

**Additional context**
[ Stop playing with IPs]

P.S. Sorry for my the best English

PR #5244, opened 2026-08-19
Save hostnames in server lists instead of resolved IPs

### LINKED ISSUE #1314: Make possible to save favourite servers with a DNS
Whenever you try to add a server to favourites with a DNS (for example: mta.example.com:22003) the IP is saved with the IP address that the DNS resolves.

The IP in MTA/coreconfig.xml under <favourite_server>, if added like a DNS, should be stored like a DNS.

None.

**Additional context**
This could be really useful for communities that, maybe in the future, would want to change their IP address for X or Y reason and don't want to (or can't) let a redirect on their old IP to the new IP with a reminder to add the new IP to their favourites again.

It would be also useful for servers that, for some reason, use a dynamic IP.

BrandSilva:
#### Summary
Servers added to favourites, recently played or connect history by hostname were stored in `coreconfig.xml` as the resolved IP, so a server changing address broke every saved entry. The hostname the player entered is now kept and resolved again on load.

`CServerListItem` already had a `strHostName` field and `SaveServerList` already preferred it over the IP when writing the config — it was simply never assigned. This fills it at the points where the typed address was discarded right after `CServerListItem::Parse()` resolved it, and makes the browser show it.

- the resolved IP is written alongside the hostname as an `ip` attribute and used as a fallback when a lookup fails, so a saved server is not lost while DNS is down
- selecting a row fills the address bar with the stored hostname instead of the resolved IP, and the favourite icon matches on it too
- `CServerList::AddUnique` now returns the item it added, mirroring `CServerListItemList::AddUnique`, so the hostname can be attached to it
- `OnFavouritesClick` no longer uses the `in_addr` uninitialised when the typed address cannot be resolved

Internal lookups still key on the resolved IP, so the address map, server cache and saved passwords are unaffected. Known limitation: two hostnames resolving to the same ip:port collapse into one entry, since the IP remains the key.

#### Motivation
Fixes #1314 and covers the favourites half of #688. Advertising a raw IP is fragile — changing host breaks every saved favourite and every address posted anywhere. With this, a server owner updates the DNS record and saved entries follow.

#### Test plan
Built from this branch on my own fork's CI and tested with the `windows-win32` artifact against a live server reached by domain.

* [x] Favourite added by domain is stored as the domain with the resolved address kept beside it: `<favourite_server host="mta.example.net" port="22003" ip="104.x.x.x"/>`
* [x] After a restart the entry is still there, the address bar shows `mtasa://mta.example.net:22003` when the row is selected, and the favourite icon is lit
* [x] Favourites added by IP are written exactly as before, with no `ip` attribute
* [x] Entries written by earlier builds load unchanged
* [ ] Connecting through a saved domain after the server's IP changes is not verified end to end: the tree builds as `VERSION_TYPE_CUSTOM`, which cannot join a public 1.6 server. Resolution goes through the same `CServerListItem::Parse` that `connect <host>` already uses today.

#### Checklist

### DIFF
--- a/Client/core/CConnectManager.cpp
+++ b/Client/core/CConnectManager.cpp
@@ -463,7 +463,7 @@ void CConnectManager::OnServerExists()
     if (m_bNotifyServerBrowser)
     {
         m_bNotifyServerBrowser = false;
-        CServerBrowser::GetSingletonPtr()->NotifyServerExists(m_Address, m_usPort);
+        CServerBrowser::GetSingletonPtr()->NotifyServerExists(m_Address, m_usPort, m_strHost);
     }
 }
 
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -1889,7 +1889,9 @@ void CCore::UpdateRecentlyPlayed()
         CServerBrowser* pServerBrowser = CCore::GetSingleton().GetLocalGUI()->GetMainMenu()->GetServerBrowser();
         CServerList*    pRecentList = pServerBrowser->GetRecentList();
         pRecentList->Remove(Address, usPort);
-        pRecentList->AddUnique(Address, usPort, true);
+        CServerListItem* pItem = pRecentList->AddUnique(Address, usPort, true);
+        if (pItem && CServerListItem::IsHostName(strHost.c_str()))
+            pItem->strHostName = strHost;
 
         pServerBrowser->SaveRecentlyPlayedList();
         if (!m_pConnectManager->m_strLastPassword.empty())
--- a/Client/core/ServerBrowser/CServerBrowser.cpp
+++ b/Client/core/ServerBrowser/CServerBrowser.cpp
@@ -1644,12 +1644,14 @@ bool CServerBrowser::OnConnectClick(CGUIElement* pElement)
     return true;
 }
 
-void CServerBrowser::NotifyServerExists(in_addr Address, ushort usPort)
+void CServerBrowser::NotifyServerExists(in_addr Address, ushort usPort, const SString& strHost)
 {
     // If the connect button was pressed, and the server exists, add it to the history
     CServerList* pHistoryList = GetHistoryList();
     pHistoryList->Remove(Address, usPort);
-    pHistoryList->AddUnique(Address, usPort);
+    CServerListItem* pItem = pHistoryList->AddUnique(Address, usPort);
+    if (pItem && CServerListItem::IsHostName(strHost.c_str()))
+        pItem->strHostName = strHost;
     while (pHistoryList->GetServerCount() > 11)
     {
         CServerListItem* pLast = *pHistoryList->IteratorBegin();
@@ -1785,7 +1787,8 @@ bool CServerBrowser::OnFavouritesClick(CGUIElement* pElement)
     {
         in_addr Address;
 
-        CServerListItem::Parse(strHost.c_str(), Address);
+        if (!CServerListItem::Parse(strHost.c_str(), Address))
+            return true;
 
         // Do we have this entry already?  If so, remove it
         if (m_ServersFavourites.Remove(Address, usPort))
@@ -1799,8 +1802,11 @@ bool CServerBrowser::OnFavouritesClick(CGUIElement* pElement)
             return true;
         }
 
-        if (m_ServersFavourites.AddUnique(Address, usPort))
+        if (CServerListItem* pItem = m_ServersFavourites.AddUnique(Address, usPort))
         {
+            if (CServerListItem::IsHostName(strHost.c_str()))
+                pItem->strHostName = strHost;
+
             SaveFavouritesList();
             RequestFilterRefresh(ServerBrowserTypes::FAVOURITES, true);
             for (std::size_t iconIndex = 0; iconIndex < std::size(m_pAddressFavoriteIcon); ++iconIndex)
@@ -2130,12 +2136,23 @@ bool CServerBrowser::LoadServerList(CXMLNode* pNode, const std::string& strTagNa
             CXMLAttribute* pPortAttribute = pSubNode->GetAttributes().Find("port");
             if (pHostAttribute && pPortAttribute)
             {
-                if (CServerListItem::Parse(pHostAttribute->GetValue().c_str(), Address))
+                const std::string strHost = pHostAttribute->GetValue();
+                iPort = atoi(pPortAttribute->GetValue().c_str());
+                if (iPort <= 0 || iPort > 0xFFFF)
+                    continue;
+
+                if (!CServerListItem::Parse(strHost.c_str(), Address))
                 {
-                    iPort = atoi(pPortAttribute->GetValue().c_str());
-                    if (iPort > 0 && iPort <= 0xFFFF)
-                        pList->AddUnique(Address, static_cast<ushort>(iPort));
+                    if (!CServerListItem::IsHostName(strHost.c_str()))
+                        continue;
+
+                    // Hostname failed to resolve, keep the entry anyway so it isn't dropped from the config on the next save
+                    Address.S_un.S_addr = 0;
                 }
+
+                CServerListItem* pItem = pList->AddUnique(Address, static_cast<ushort>(iPort));
+                if (pItem && CServerListItem::IsHostName(strHost.c_str()))
+                    pItem->strHostName = strHost;
             }
         }
     }
--- a/Client/core/ServerBrowser/CServerBrowser.h
+++ b/Client/core/ServerBrowser/CServerBrowser.h
@@ -106,7 +106,7 @@ public:
     void SetNextHistoryText(bool bDown);
 
     void OnQuickConnectButtonClick();
-    void NotifyServerExists(in_addr Address, ushort usPort);
+    void NotifyServerExists(in_addr Address, ushort usPort, const SString& strHost);
 
     void TabSkip(bool bBackwards);
 
--- a/Client/core/ServerBrowser/CServerList.h
+++ b/Client/core/ServerBrowser/CServerList.h
@@ -130,6 +130,9 @@ public:
         return true;
     }
 
+    // Anything that is not a numeric IPv4 address is treated as a hostname
+    static bool IsHostName(const char* szAddress) { return szAddress[0] != '\0' && inet_addr(szAddress) == INADDR_NONE; }
+
     bool operator==(const CServerListItem& other) const { return (Address.S_un.S_addr == other.Address.S_un.S_addr && usGamePort == other.usGamePort); }
 
     void Init()


Esteban528 (approved):
Tested by me. Works well

TridentSky:
Thanks for taking the time to build and test it!

Out of curiosity, which cases did you run through? The one I'd most like a second pair of eyes on is the unresolvable hostname at startup — a favourite saved as a domain should now survive a launch with DNS down instead of being silently dropped from coreconfig.xml on the next save.

Keeping this as draft until the CI gets approved to run, so I can go through the full test plan against that build.

vectorbasis:
In its current state, this doesn't fully solve the problem, it's more of a workaround, because the master list still only works with IP addresses, and there is no way to specify the displayed hostname in the server configuration, which the client could then use in the server browser.

Disinterpreter:
> In its current state, this doesn't fully solve the problem, it's more of a workaround, because the master list still only works with IP addresses, and there is no way to specify the displayed hostname in the server configuration, which the client could then use in the server browser.

Domain can be added by rules.
https://github.com/multitheftauto/mtasa-blue/discussions/3468

Esteban528:
> In its current state, this doesn't fully solve the problem, it's more of a workaround, because the master list still only works with IP addresses, and there is no way to specify the displayed hostname in the server configuration, which the client could then use in the server browser.

Why refuse to implement new features? We can start somewhere, and that's by adding local domain support. The master list doesn't need to implement this yet.

Esteban528:
If loyal players on the servers save the domain names instead of the IP addresses, these servers can switch hosts without needing to communicate externally with the players. In other words, it's obvious why this is beneficial.

TridentSky:
> In its current state, this doesn't fully solve the problem, it's more of a workaround, because the master list still only works with IP addresses, and there is no way to specify the displayed hostname in the server configuration, which the client could then use in the server browser.

Implementing domain names instead of relying solely on IP addresses would be extremely beneficial for all servers.

**1. They’re much easier to remember.** For example, if you want to play Minecraft and join a server like Hypixel, the first thing you remember is that you just need to type `mc.hypixel.net` and you're good to go. It simply looks cleaner and is easier to remember than something like `xxx.xx.xx.xxx`. You understand what I’m trying to say, right?

**2. Using a domain for this doesn’t even have to be mandatory.** The system could resolve and cache the numerical IP address the first time the player connects. Then, whenever the client refreshes, if the domain fails to resolve, the system could attempt to connect using the cached backup IP—as long as the IP was previously resolved through DNS.

In fact, in MTA you can already connect using a domain and port, for example: `mta.tridentsky.net:22003`. The problem is that the domain itself is not stored on the player's client; it gets converted into a numerical IP address. So let's say that, for various reasons, I change hosting providers and my IP address changes. Logistically, that's a huge headache for a server owner because they have to promote the new IP address all over again across all their social media platforms.

The goal is to give the server a permanent, recognizable address that is easier to remember and connect to. If the host changes its IP address, the server owner should only need to update the IP associated with the domain, and that's it. I honestly don't know how difficult it would be to implement something like this.

FiveM, Minecraft, and even SA-MP already work this way.

Esteban528:
The cost of this implementation is negligible in performance anyway. It's no bloat and will be useful

vectorbasis:
> Why refuse to implement new features? We can start somewhere, and that's by adding local domain support. The master list doesn't need to implement this yet.

No one said that the pull request would not be merged if it was implemented properly. The comment only represents **my personal opinion and observation**, nothing more. I also never said that we should not implement this change, only that it does not fully solve the problem yet.

@TridentSky, the main issue with your response is that it reads like an AI generated answer and does not really address the point I was making. It mostly explains things that are already known, but it does not provide a solution for the actual problem.

> FiveM, Minecraft, and even SA-MP already work this way.

Similarly, but not exactly in the same way. This is the important distinction that I was trying to point out.

The current server browser does not display custom server addresses. Even if a player can manually save a server using a domain, servers discovered through the master list are still represented using raw IP addresses. This means the actual public server discovery experience still has the same limitation.

Based on Disi's response, this does not seem like a particularly difficult improvement. A possible implementation could be adding a new optional server browser rule, similar to the existing rules such as `website_url`, `description`, or `tags`, allowing servers to provide a custom address.

For example:
```xml
<rule name="server_address" value="play.example.com" />
```

When this rule is present and valid, the server browser could display and use this address instead of the raw IP address.

Of course, this should not be treated as a simple display only value. Some validation should be involved to prevent impersonation. The provided address should be verified, for example by requiring domain ownership verification through a DNS TXT record, and then confirming that the verified address actually points to the server endpoint.

If verification fails, the server browser should simply ignore the rule and fall back to displaying the normal IP address.

This would allow community servers to have a recognizable identity while avoiding cases where someone could claim an unrelated or official looking address.

TridentSky:
> @TridentSky, the main issue with your response is that it reads like an AI generated answer and does not really address the point I was making. It mostly explains things that are already known, but it does not provide a solution for the actual problem.

To start with, I think it’s completely unnecessary to say, “your comment was made with AI.” I’m Latino, I don’t speak English, and all I do is translate my suggestions into English. That has nothing to do with the actual point I’m making. And even if it were true, what would be so wrong with that?

It seems like a lot of people take things the wrong way whenever someone tries to suggest changes that are actually useful and as simple as this one, which is absurdly easy to implement.

I’ve been working in development for years, without AI. I don’t need to prove anything to anyone. If my mistake was simply translating my messages from Spanish into English, then there’s really nothing else to say.

> Based on Disi's response, this does not seem like a particularly difficult improvement. A possible implementation could be adding a new optional server browser rule, similar to the existing rules such as `website_url`, `description`, or `tags`, allowing servers to provide a custom address.
> 
> For example:
> 
> ```
> <rule name="server_address" value="play.example.com" />
> ```
> 
> When this rule is present and valid, the server browser could display and use this address instead of the raw IP address.

I think it would be a good idea to do something like this. What are we waiting for to implement it?

Honestly, I don’t want to put time and effort into working on something only for it to be rejected later or dismissed because someone says it was made with AI. We’re already years behind on something so simple, yet genuinely useful and convenient for server advertising.

AlexTMjugador:
> What are we waiting for to implement it?

A thought experiment for you: the DNS server a player uses is determined by their network configuration and/or ISP, not by MTA server owners, and it is a widely known fact that DNS is not a tamper-proof protocol. What's your plan for handling the situation that will eventually arise when a domain name resolves to the wrong server IP?

Note that a variant of the above problem can also occur without any DNS tampering if the domain's TTL is set to such a high value that recursive DNS resolvers continue returning stale records for an extended period.

No matter whether this PR gets merged, when migrating servers to a different IP address, it's probably a good idea to keep a "stub" server listening on the previous socket address that redirects players to the new server for a transition period, economic considerations notwithstanding. That would make the migration as seamless as possible for everyone.

TridentSky:
Stepping back a bit. My last reply was defensive and didn't really answer what either of you asked, sorry about that.

@vectorbasis you're right, this doesn't touch discovery. Servers coming from the master list still show up and get saved as raw IPs, this only keeps what the player typed by hand. I looked into the rule idea: the server side is mostly there since #3761 already loads `<rule>` entries from mtaserver.conf, but the browser doesn't read rules at all yet, the master list would need to carry the field, and then there's the TXT verification on top. That's a much bigger change than this PR and it deserves its own discussion, so my proposal is to keep this one limited to the manual/favourites path and open a separate issue for the `server_address` rule where the verification part can be worked out properly. If you'd rather have it all in one PR, tell me and I'll rework it.

@AlexTMjugador this PR doesn't change the trust model, it stores the string instead of the result of gethostbyname. Resolution happens at load with the same call that `connect play.example.com` goes through today, so a tampered or stale record sends the player to the same place it would if they typed the domain. The real difference is that a saved favourite re-resolves silently on every launch, and I agree that's worth being explicit about. Stale TTL means landing on the old IP, which is what a saved IP does now anyway, so the stub server you describe is the right advice in both cases and nothing on the client side replaces it.

Two changes I think make sense for this PR based on your comments:

1. Keep the last resolved IP next to the hostname in coreconfig.xml (`host="play.example.com" ip="1.2.3.4"`). If the lookup fails the entry falls back to the last known address instead of the null address it gets now, and the browser can show the IP it actually resolved to, so there's no hidden redirect.
2. Keep resolution at launch only. It already works that way (the browser never re-resolves in the background), but I'll make sure it stays like that so a player always connects to an address they can see.

I'll hold off on pushing anything until I hear whether that's the direction you want.

Esteban528:
> > What are we waiting for to implement it?
> 
> A thought experiment for you: the DNS server a player uses is determined by their network configuration and/or ISP, not by MTA server owners, and it is a widely known fact that DNS is not a tamper-proof protocol. What's your plan for handling the situation that will eventually arise when a domain name resolves to the wrong server IP?

I dont think the DNS would be a problem, because this is just an alternative to keep favorites resolving IP automatically. In this case it makes things easier. And nowadays it's too difficult to be breached. 

> No matter whether this PR gets merged, when migrating servers to a different IP address, it's probably a good idea to keep a "stub" server listening on the previous socket address that redirects players to the new server for a transition period, economic considerations notwithstanding. That would make the migration as seamless as possible for everyone.

This is a great idea

BrandSilva:
No reply in a month, so rather than sit on this I ran the build on my own fork, since CI here has never been approved to run on this PR.

Good thing I did, because it didn't compile. I was assigning the result of `CServerList::AddUnique` to a `CServerListItem*` at four call sites, but that overload returns `bool` — only `CServerListItemList::AddUnique` hands back the item. Fixed by making `CServerList::AddUnique` forward the pointer and return `nullptr` when the endpoint is already in the list. The three existing callers are unaffected: two discard the result and the third goes through the inner list.

I force-pushed with the fix folded into the original commit rather than stacking a build-fix commit on top. windows-win32, x64 and arm64 all pass now: https://github.com/BrandSilva/mtasa-blue/actions/runs/35531976023

I also pushed the `ip` attribute fallback I described earlier. Without it, an entry whose hostname stops resolving has no address to fall back on, and every saved hostname ends up sharing the same null address. If you'd rather review the first commit on its own, say so and I'll drop it.

Keeping it as a draft until I've gone through the manual test plan against that build.

BrandSilva:
Thanks for updating the branch.

While testing the build on my own server I hit a gap that made the feature useless in practice: the hostname was stored correctly in coreconfig.xml, but clicking the row in the browser refilled the address bar from the resolved IP, so the domain disappeared from the UI the moment you touched the entry. Pushed a fix for that — the address bar now shows the stored hostname, and the favourite icon compares against it too, otherwise an entry saved by domain no longer registered as a favourite.

Confirmed working against my own server, saved as:

    <favourite_server host="mta.example.net" port="22003" ip="104.x.x.x"/>

Internal lookups still key on the resolved IP, so nothing changes for the cache or saved passwords.

PR #5245, opened 2026-08-19
Fix crash in CPathFind path node loader when allocation fails

### LINKED ISSUE #1024: SA streaming crash at 0x0116f981
**Describe the bug**
The crash at gta_sa offset 0x0116f981 has been present only since the commit 17ee332 (@sbx320) and thus probably has been introduced by it (unless an existing problem/user circumstance started exposing itself there instead of somewhere else, as a direct result of the commit).

CPathFind::LoadPathFindData
CStreaming::ConvertBufferToObject
CStreaming::ConvertBufferToObject
CStreaming::ProcessLoadingChannel
CStreaming::LoadRequestedModels

It seems to relate with/be accompanied by blocking load fails, often logging in sequences like this:
Blocking load fail: 16732 (CClientModelRequestManager::DoPulse #2)
File had NULL buffer (size:-301582388)
File had NULL buffer (size:-1759412222)
File had NULL buffer (size:-360141144)
File had NULL buffer (size:-180070956)
*Crash happens

This issue is possibly related to [https://bugs.mtasa.com/view.php?id=9449](https://bugs.mtasa.com/view.php?id=9449) considering the chain of changes the revision that seemingly introduced it, was part of.

**To reproduce**
No known reproduction

**Version**
- Build 1.5.3-r10724 until as of writing (1.5.6-r18728)

**Additional context**
I want to have this crash documented in an issue because there's a clear introduction point of the crash, and because it's in the top 10 of most common SA crashes. If it's brought on by memory issues/streaming limits, then it's possible that the referred commit (or those part of the chain of related changes) somehow made the game more prone to crashing under these circumstances than before.

Some crash dumps are attached.
[dumps.zip](https://github.com/multitheftauto/mtasa-blue/files/3388898/dumps.zip)

TheCrazy17:
#### Summary

CPathFind's per-area path node loader calls malloc() for m_pPathNodes[area] and never checks whether it succeeded before looping over the result to fix up each node's flag byte; if the allocation fails, that loop dereferences a null pointer and crashes.

This adds a native hook that checks whether m_pPathNodes[area] is null right where the node count gets loaded for the loop's bound check, forcing the count to 0 in that case so the loop is skipped instead of crashing. The stored node count itself is left untouched, and no other behaviour changes; an empty area is already a normal, safely handled case elsewhere in the same function.

One implementation note for reviewers: this hook's address sits further into the executable than any other hook in the codebase, past the range MTA's shared `IsSlowMem` check recognizes. Installing it the usual way (`EZHookInstall`) trips that check's own "should use a Mem*Fast function" assertion in Debug builds, so it's installed directly through the Fast path instead (with a matching unprotect call added for its page), the same way the ~17 other known-hot addresses already are.

#### Motivation

Fixes #1024. This has been reported as one of the more common native SA crashes, with no known way to reproduce it on demand until now. Since it comes from a plain malloc() failure with no null check, it doesn't need any specific feature to trigger; it's far more likely to show up on long sessions or heavily modded servers where GTA:SA's 32-bit address space gets fragmented, since that's exactly when this kind of allocation starts failing.

I traced the crash address from the issue back to this exact instruction, and cross-checked the struct offsets involved (m_pPathNodes, m_dwNumNodes) against plugin-sdk-sa CPathFind.h to confirm the layout.

#### Test plan

I confirmed both directions with a temporary repro hook (not part of this PR, shown below): with the repro active and the fix's install commented out, the game crashes reliably at the same address the moment path data loads; with the fix active again, it no longer crashes and the affected area just ends up with zero path nodes for that load, the same as GTA:SA's own existing empty-area case.

To reproduce it yourself, drop this in place of `Client/game_sa/CPathFindSA.cpp`, build, and try to join any server:

<details>
<summary>CPathFindSA.cpp (forces the crash, fix's own hook disabled)</summary>

```cpp
 *
 *  PROJECT:     Multi Theft Auto v1.0
 *  FILE:        game_sa/CPathFindSA.cpp
 *  PURPOSE:     Path find (CPathFind) hooks
 *
 *  Multi Theft Auto is available from https://www.multitheftauto.com/
 *

#include "StdInc.h"
#include "CPathFindSA.h"

// CPathFind's per map area path node loader calls malloc() for m_pPathNodes[area] (offset
// 0x804) without checking the result, then loops over it using m_dwNumNodes[area] (offset
// 0xfa4) as the count. If the allocation fails, that loop dereferences a null pointer and
// crashes.
//
// Right where the node count gets loaded for the loop's bound check, this hook also checks
// whether m_pPathNodes[area] is null, and if so forces the count to 0 so the loop is skipped.
#define HOOKPOS_CPathFind_LoadPathNodeCount_Mid  0x0156F966
#define HOOKSIZE_CPathFind_LoadPathNodeCount_Mid 7
DWORD                         RETURN_CPathFind_LoadPathNodeCount_Mid = 0x0156F96D;
static void __declspec(naked) HOOK_CPathFind_LoadPathNodeCount_Mid()
{
    MTA_VERIFY_HOOK_LOCAL_SIZE;

    // clang-format off
    __asm
    {
        // Replicate the overwritten instruction: EAX = m_dwNumNodes[area]
        mov     eax, dword ptr [esi + edi*4 + 0x0fa4]

        // If m_pPathNodes[area] failed to allocate, force the node count to 0 for this loop
        cmp     dword ptr [esi + edi*4 + 0x804], 0
        jne     nodesOk
        xor     eax, eax
        nodesOk:
        jmp     RETURN_CPathFind_LoadPathNodeCount_Mid
    }
    // clang-format on
}

// Test-only: deliberately reproduces the crash above by forcing every m_pPathNodes[area]
// allocation to look like it failed, the moment after it's stored.
#define MTA_DEBUG_REPRO_PATHFIND_NULL_CRASH

#ifdef MTA_DEBUG_REPRO_PATHFIND_NULL_CRASH
    #define HOOKPOS_CPathFind_ReproForceNull_Mid  0x0156F7F6
    #define HOOKSIZE_CPathFind_ReproForceNull_Mid 7
DWORD                         RETURN_CPathFind_ReproForceNull_Mid = 0x0156F7FD;
static void __declspec(naked) HOOK_CPathFind_ReproForceNull_Mid()
{
    MTA_VERIFY_HOOK_LOCAL_SIZE;

    // clang-format off
    __asm
    {
        // Replicate the overwritten instruction (unrelated field, must still run)
        mov     eax, dword ptr [esi + edi*4 + 0x1304]

        // Force the just-stored m_pPathNodes[area] to null, as if malloc() had failed
        mov     dword ptr [esi + edi*4 + 0x804], 0

        jmp     RETURN_CPathFind_ReproForceNull_Mid
    }
    // clang-format on
}
#endif

void CPathFindSA::StaticSetHooks()
{
    BYTE jumpBytes[MAX_JUMPCODE_SIZE];

    // Uncomment these 3 lines to see the fix neutralize the crash below instead:
    // MemSetFast(jumpBytes, 0x90, MAX_JUMPCODE_SIZE);
    // CreateJump(HOOKPOS_CPathFind_LoadPathNodeCount_Mid, (DWORD)HOOK_CPathFind_LoadPathNodeCount_Mid, jumpBytes);
    // MemCpyFast((PVOID)HOOKPOS_CPathFind_LoadPathNodeCount_Mid, jumpBytes, HOOKSIZE_CPathFind_LoadPathNodeCount_Mid);

#ifdef MTA_DEBUG_REPRO_PATHFIND_NULL_CRASH
    MemSetFast(jumpBytes, 0x90, MAX_JUMPCODE_SIZE);
    CreateJump(HOOKPOS_CPathFind_ReproForceNull_Mid, (DWORD)HOOK_CPathFind_ReproForceNull_Mid, jumpBytes);
    MemCpyFast((PVOID)HOOKPOS_CPathFind_ReproForceNull_Mid, jumpBytes, HOOKSIZE_CPathFind_ReproForceNull_Mid);
#endif
}
```

</details>

#### Checklist


Client/game_sa/CPathFindSA.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+// This address is further into the executable than IsSlowMem recognizes, so the normal
+// EZHookInstall path would assert in Debug builds. Unprotected up front in
+// SetInitialVirtualProtect and installed via the Fast path instead, like the other hot addresses.
+void CPathFindSA::StaticSetHooks()
+{
+    BYTE jumpBytes[MAX_JUMPCODE_SIZE];
+    MemSetFast(jumpBytes, 0x90, MAX_JUMPCODE_SIZE);
+    CreateJump(HOOKPOS_CPathFind_LoadPathNodeCount_Mid, (DWORD)HOOK_CPathFind_LoadPathNodeCount_Mid, jumpBytes);
+    MemCpyFast((PVOID)HOOKPOS_CPathFind_LoadPathNodeCount_Mid, jumpBytes, HOOKSIZE_CPathFind_LoadPathNodeCount_Mid);
```
FileEX, line 53: Use `HookInstall` or add a new method for installing fast hooks.

FileEX (changes requested):

Client/game_sa/CPathFindSA.cpp
TheCrazy17: Done, I retested it and it’s working as expected.

Client/game_sa/CPathFindSA.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+        // If m_pPathNodes[area] failed to allocate, force the node count to 0 for this loop
+        cmp     dword ptr [esi + edi*4 + 0x804], 0
+        jne     nodesOk
+        xor     eax, eax
+        nodesOk:
+        jmp     RETURN_CPathFind_LoadPathNodeCount_Mid
+    }
+    // clang-format on
+}
```
FileEX: Please move this hook into CMultiplayerSA_CrashFixes

FileEX (approved):

FileEX (changes requested):
Please fix build

Client/multiplayer_sa/CMultiplayerSA_CrashFixHacks.cpp
```diff
@@ -4191,4 +4225,6 @@ void CMultiplayerSA::InitHooks_CrashFixHacks()
 
     // Fix uninitialized wheel scale in CFileLoader::LoadVehicleObject on Win11 24H2
     HookInstallCall(CALL_CFileLoader_LoadVehicleObject_sscanf, (DWORD)CFileLoader_LoadVehicleObject_sscanf);
+
+    EZHookInstall(CPathFind_LoadPathNodeCount_Mid);
```
FileEX, line 4229: Image

What changed that makes a regular `EZHookInstall` sufficient now?

TheCrazy17: This is leftover from moving the hook, not something that changed. game_sa's HookInstall always writes through MemCpy with no check first, so at this address it asserts, that's why HookInstallFast existed there, and the call just came along when the hook moved into CMultiplayerSA_CrashFixHacks.cpp. multiplayer_sa's own HookInstall already checks IsSlowMem first and picks MemCpy or MemCpyFast on its own, so it takes the same path HookInstallFast did, just automatically, the same way CrashFix_Misc50 to 52 and FreezeFix_Misc15 already do at similar addresses.

FileEX (approved):

PR #5246, opened 2026-08-19
improve CGUI exception handling and fix some bugs

Dutchman101:
This PR makes the client a lot less likely to randomly crash when GUI-related things go wrong. The main idea is that instead of silently crashing to desktop (which often happened through std::terminate and without dialog or info), the client now shows a clear error message telling the player to reinstall MTA, and then exits cleanly. I also fixed smaller bugs I ran into while working on this, i.e. memory leaks, null pointer crashes, and broken error dialogs stacking on top of each other.

**What was happening before:**

- When a skin file was missing or corrupted, the error dialog that popped up would actually cause more crashes while it was open, because the game keeps running in the background and tries to draw GUI windows that were already destroyed. This led to either a silent crash or two error dialogs stacked on top of each other.

- When textures or web browser elements failed to load (happens with custom resources a lot), CEGUI exceptions would escape all the way up and crash the process with no message at all.

- When the D3D device got lost (Alt+Tab, fullscreen switch, screensaver), the pre-reset cleanup could throw a CEGUI exception that had no handler, causing a crash in the COM layer which is even harder to debug.

- The web browser widget would crash if you clicked on it after CEF failed to start up, because it tried to send input to a web view that didn't exist.

- Textures created by static image widgets would leak if the imageset creation step failed partway through.

- Some font-loading exceptions would go uncaught because the catch blocks were too narrow (only catching one specific type instead of the base CEGUI exception class).

**What the PR does:**

The client now catches crashes during GUI drawing (both C++ exceptions and access violations) and shows a "Rendering fault -- missing GUI/loading-screen assets" dialog instead of just disappearing. This covers both the per-frame draw and the D3D present call.

When skin loading fails and that big CC51 error dialog appears, the code now sets a flag telling everything else it's drawn and not to draw or rebuild more windows. If a second fault somehow fires during that dialog, the process just terminates immediately instead of stacking another dialog on top.

Skin and locale change functions now check whether that flag is set (or whether windows are half-destroyed) and bail out early instead of touching destroyed objects.

Console-related code in DoPulse and ApplyQueuedLocale now checks if the console object actually exists before trying to print to it, since it can be destroyed during a skin change while the error dialog is pumping messages.

Static image loading is wrapped in a try/catch so if CEGUI fails to create the imageset, the widget just goes blank and logs the error instead of crashing. The old texture/imageset cleanup was also reordered to properly detach things from the widget before destroying them (no more dangling references inside CEGUI's renderer).

Web browser loading got the same treatment, plus a bunch of null checks in the mouse/keyboard event handlers so if the browser element fails to create its web view, clicking on it doesn't crash.

The browser widget now properly tracks and frees the texture it creates (was leaking before, and the leak path was basically guaranteed if CEF failed).

The D3D device-lost handler (Invalidate) now catches CEGUI exceptions, shows a message, and terminates instead of letting the exception cross into COM.

Font-loading catch blocks were widened from the specific InvalidRequestException to the base CEGUI Exception class, so other CEGUI errors during font creation don't escape the constructor.

The ApplyConsoleSettings function got an early-out if the console isn't ready yet (happens during early startup or during skin changes).

A bunch of exception catches were changed from catching-by-value (which slices the exception object) to catching-by-const-reference.

### DIFF
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -651,9 +651,11 @@ bool CCore::IsCursorForcedVisible()
 
 void CCore::ApplyConsoleSettings()
 {
-    CVector2D vec;
     CConsole* pConsole = m_pLocalGUI->GetConsole();
+    if (!pConsole)
+        return;
 
+    CVector2D vec;
     CVARS_GET("console_pos", vec);
     pConsole->SetPosition(vec);
     CVARS_GET("console_size", vec);
--- a/Client/core/CCore.cpp
+++ b/Client/core/CCore.cpp
@@ -652,7 +652,7 @@ bool CCore::IsCursorForcedVisible()
 void CCore::ApplyConsoleSettings()
 {
     CConsole* pConsole = m_pLocalGUI->GetConsole();
-    if (!pConsole)
+    if (!pConsole) [[unlikely]]
         return;
 
     CVector2D vec;
--- a/Client/core/CGUI.cpp
+++ b/Client/core/CGUI.cpp
@@ -58,6 +58,26 @@ CLocalGUI::~CLocalGUI()
 
 void CLocalGUI::SetSkin(const char* szName)
 {
+    // Guard against re-entrant calls. MessageBoxW pumps the message loop, so the
+    // fatal CC51 dialog below would otherwise trigger pulse calls that touch
+    // m_pConsole while windows are half-destroyed (DestroyWindows called but
+    // CreateWindows not yet reached). The guard also clears the flag if a
+    // window rebuild throws, so a failed skin change cannot disable later ones.
+    static bool s_bInSetSkin = false;
+    if (s_bInSetSkin)
+        return;
+    struct SetSkinGuard
+    {
+        bool& flag;
+        explicit SetSkinGuard(bool& inFlag) : flag(inFlag) { flag = true; }
+        ~SetSkinGuard() { flag = false; }
+    } guard(s_bInSetSkin);
+
+    // A fatal fault dialog may be pumping the message loop; do not rebuild the
+    // windows inside that pump.
+    if (CLocalGUI::IsFaultDialogOpen())
+        return;
+
     CVector2D consolePos, consoleSize;
 
     bool guiWasLoaded = m_pMainMenu != NULL;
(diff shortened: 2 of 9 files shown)


Client/core/CCore.cpp
```diff
@@ -651,9 +651,11 @@ bool CCore::IsCursorForcedVisible()
 
 void CCore::ApplyConsoleSettings()
 {
-    CVector2D vec;
     CConsole* pConsole = m_pLocalGUI->GetConsole();
+    if (!pConsole)
```
FileEX: unlikely attribute?

Client/core/CGUI.cpp
```diff
@@ -115,6 +139,16 @@ void CLocalGUI::SetSkin(const char* szName)
 
 void CLocalGUI::ChangeLocale(const char* szName)
 {
+    // Guard against re-entrant calls while the windows are half-destroyed during
+    // a skin change (the fatal skin dialog pumps the message loop).
+    if (!m_pConsole)
```
FileEX: unlikely?

Client/core/CGUI.h
```diff
@@ -52,9 +52,16 @@ class CLocalGUI : public CSingleton<CLocalGUI>
...
+    void DrawInternal();
     void Invalidate();
     void Restore();
 
+    // True while a fatal GUI fault dialog is open, so a nested fault during the
+    // dialog's message-loop pump terminates without stacking more dialogs, and
+    // window-rebuild paths (skin and locale changes) do not run inside the pump.
+    static bool IsFaultDialogOpen();
+    static void SetFaultDialogOpen(bool bOpen);
+
```
FileEX: noexcept?

Client/core/CGUI.cpp
```diff
@@ -409,14 +445,80 @@ void CLocalGUI::DoPulse()
...
+    MessageBoxUTF8(0, strMsg, _("Error") + _E("CC54"), MB_OK | MB_ICONERROR | MB_TOPMOST);
+    TerminateProcess(GetCurrentProcess(), 9);
+}
+
+// SEH filter for GUI rendering faults. Access violations and C++ exceptions
+// (CEGUI errors) are the common failures with missing or corrupt GUI assets;
+// other faults are left for the wider OnPresent guard.
+static int FilterGUISehFault(unsigned int uiExceptionCode)
+{
+    if (uiExceptionCode == EXCEPTION_ACCESS_VIOLATION || uiExceptionCode == 0xE06D7363)
```
FileEX: ``CPP_EXCEPTION_CODE``

Client/core/CGUI.cpp
```diff
@@ -409,14 +445,80 @@ void CLocalGUI::DoPulse()
...
+void CLocalGUI::DrawInternal()
 {
     // Get the game interface
     CGame*      pGame = CCore::GetSingleton().GetGame();
     SystemState systemState = pGame->GetSystemState();
     CGUI*       pGUI = CCore::GetSingleton().GetGUI();
 
+    // The windows may be half-destroyed while a fatal dialog pumps the message
+    // loop during a skin or locale change, so skip drawing until they are back.
+    if (!m_pMainMenu)
```
FileEX: unlikely?

FileEX (changes requested):

FileEX (approved):

PR #5247, opened 2026-08-20
Fix #664: Prevent vehicle doors freezing after passenger exits

### LINKED ISSUE #664: Doors freeze after any player exit passenger seat
**Describe the bug**
Doors freeze after any player exit passenger seat

**To reproduce**
Drive a car with another player as passenger.
Let passenger exit car without closing door.

**Expected behaviour**
Doors shouldn't freeze.

**Screenshots**
Video:
[Video](https://www.youtube.com/watch?v=mqLpPFP-6Ng)

MohabCodeX:
Fixes #664

This problem has existed for years:
- Vehicle doors freezing in mid-air after a passenger exits while driving.
- Swinging doors never auto-latching shut when pushed closed by vehicle motion.
- Doors remaining unlatched and flapping after an occupant enters.

All are now fixed.

https://github.com/user-attachments/assets/58e7ab5b-c4f8-4af2-a79c-377fa3c1458c

### DIFF
--- a/Client/mods/deathmatch/logic/CClientPed.cpp
+++ b/Client/mods/deathmatch/logic/CClientPed.cpp
@@ -15,6 +15,7 @@
 #include <game/CCam.h>
 #include <game/CCarEnterExit.h>
 #include <game/CColPoint.h>
+#include <game/CDoor.h>
 #include <game/CPedIntelligence.h>
 #include <game/CPedSound.h>
 #include <game/CStreaming.h>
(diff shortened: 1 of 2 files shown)


MohabCodeX:
Caught another bug:

When a player tries to jack a car from the driver’s door and fails because the driver drives away, the door gets frozen. I’m fixing it now.

efejotaese2:
Could this interfere with `setVehicleDoorState`?

For example, if a script sets a closed door to state `1` or `3` while swinging doors are enabled, wouldn't this turn it back into `0`/`2` on the next pulse since the ratio is already <= 0.01?

Not sure if I'm missing something here :p

